PR #6612, opened 2024-11-01
Remove typing true flag from executorch

maggiemoss:
Summary: This is on our blocklist - it shouldn't be opted into PTT typing

Reviewed By: grievejia

Differential Revision: D65343496


zingo:
It would be nice if it is possible to make this run the ciflow/trunk tests to make sure Arm testing is executed and still work.
@digantdesai

zingo:
I have tested this a bit internally ant it seem to cause no problem, I even removed all of "typing = True," under the arm folder without problem.

zingo (approved):

PR #6620, opened 2024-11-01
Java Tensor and EValue serialization

### LINKED ISSUE #6569: RFC: Java layer EValue serialization

Context: https://github.com/pytorch/executorch/issues/6470#issuecomment-2436060471

We need to design a way to serialize an EValue (and its embedding tensor), for some IPC use case in AOSP.

This won’t be the official serialization across ET. ET uses fbs for serialization. This is only for the Java frameworks layer for AOSP.

# Basic layout

Tag (1 byte)
Bytes_of_payload (8 bytes?)
<
<
<
<
<
<
<
Payload (var)
…
…

Where payload can be one of them

None (0 byte): No value or absence of a value.
Tensor (var): A multi-dimensional array used for numerical computations.
String/uint8_array (var): A sequence of characters, often used to represent text.
Double (8): A 64-bit floating-point number, used for decimal arithmetic.
Int (4 or 8?): A 32-bit integer, used for whole numbers.
Bool (1): A boolean value, either true or false.
ListBool (var): A list of boolean values.
ListDouble (var): A list of double-precision floating-point numbers.
ListInt (var): A list of integers.
ListTensor (var): A list of tensors.
ListScalar (var): A list of scalar values (e.g., numbers).
ListOptionalTensor (var): A list of optional tensors, where each element may be present or absent.

Per Jacob, we don’t care about List types (6-11). They are internal to ET runtime. 

For those without variant length (0, 3, 4, 5), we just serialize the value directly.

For 2 String, let’s assume that it’s uint8_t[]. We just serialize the array directly.

So we will focus on the tensor type. 

# Tensor type

Scalar_type (1 byte)
Num_dim (1)
Sizes (var)
…
Dim_order (var)
…
Data (var)
…
…
…
…
…

# Questions
Do we really need a field for Bytes_of_payload?
Do we need Dim order?
Do we need TensorShapeDynamism?

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

kirklandsign:
Add serialization and deserialization for EValue (except string) and Tensor.

### Summary
Fixes #6569 

### Test plan
```
cd extension/android_test/
./gradlew build
```

### DIFF
--- a/extension/android/src/main/java/org/pytorch/executorch/EValue.java
+++ b/extension/android/src/main/java/org/pytorch/executorch/EValue.java
@@ -9,6 +9,8 @@
 package org.pytorch.executorch;
 
 import com.facebook.jni.annotations.DoNotStrip;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.Optional;
 import org.pytorch.executorch.annotations.Experimental;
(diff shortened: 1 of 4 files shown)


extension/android/src/main/java/org/pytorch/executorch/EValue.java
```diff
@@ -287,4 +289,75 @@ private void preconditionType(int typeCodeExpected, int typeCode) {
...
+    } else if (isInt()) {
+      return ByteBuffer.allocate(9).put((byte) TYPE_CODE_INT).putLong(toInt()).array();
+    } else if (isDouble()) {
+      return ByteBuffer.allocate(9).put((byte) TYPE_CODE_DOUBLE).putDouble(toDouble()).array();
+    } else if (isString()) {
+      return ByteBuffer.allocate(1 + toString().length())
+          .put((byte) TYPE_CODE_STRING)
+          .put(toString().getBytes())
+          .array();
+    } else {
```
davidlin54, line 324: to enable Tensor, can you convert it into a `byte[]` and use `put()` here?

davidlin54, line 324: will we add list support as well?

extension/android/src/main/java/org/pytorch/executorch/Tensor.java
```diff
@@ -679,4 +679,105 @@ private static Tensor nativeNewTensor(
...
+   *
+   * @return The serialized byte array.
+   * @apiNote This method is experimental and subject to change without notice. This does NOT
+   *     supoprt list type.
+   */
+  public byte[] toByteArray() {
+    int dtypeSize = 0;
+    byte[] tensorAsByteArray = null;
+    if (dtype() == DType.FLOAT) {
+      dtypeSize = 4;
```
davidlin54, line 694: can you reference the consts on L56 and add the missing ones? (uint8, int8)

extension/android/src/main/java/org/pytorch/executorch/Tensor.java
```diff
@@ -679,4 +679,105 @@ private static Tensor nativeNewTensor(
...
+  public static Tensor fromByteArray(byte[] bytes) {
+    if (bytes == null) {
+      throw new IllegalArgumentException("bytes cannot be null");
+    }
+    ByteBuffer buffer = ByteBuffer.wrap(bytes);
+    if (!buffer.hasRemaining()) {
+      throw new IllegalArgumentException("invalid buffer");
+    }
+    byte scalarType = buffer.get();
+    byte numberOfDimensions = buffer.get();
```
davidlin54, line 756: can we use same constants as the serialize function? dtype and shapeLength?

extension/android_test/src/test/java/org/pytorch/executorch/EValueTest.java
```diff
@@ -9,29 +9,25 @@
...
 
 /** Unit tests for {@link EValue}. */
 @RunWith(JUnit4.class)
 public class EValueTest {
+  private static final int TYPE_CODE_NONE = 0;
+  private static final int TYPE_CODE_TENSOR = 1;
+  private static final int TYPE_CODE_STRING = 2;
+  private static final int TYPE_CODE_DOUBLE = 3;
+  private static final int TYPE_CODE_INT = 4;
+  private static final int TYPE_CODE_BOOL = 5;
```
davidlin54, line 30: maybe we can just expose the constants in EValue as public instead of copying this here

extension/android/src/main/java/org/pytorch/executorch/EValue.java
kirklandsign: Per Jacob, list is an internal dtype within ET runtime. Maybe we should totally get rid of list in java layer. I could double check with the team

davidlin54 (approved):
could consider using switch statements instead of if, but looks good

kirklandsign:
> could consider using switch statements instead of if, but looks good

Will follow up :)

PR #6623, opened 2024-11-03
compiler.h changes for Windows MSVC support

tarun292:
Differential Revision: D65328573

### DIFF
--- a/kernels/prim_ops/register_prim_ops.cpp
+++ b/kernels/prim_ops/register_prim_ops.cpp
@@ -148,7 +148,7 @@ static Kernel prim_ops[] = {
           EValue& out = *stack[2];
           if (a.isInt() && b.isInt()) {
             const int64_t quot = a.toInt() / b.toInt();
-            if (std::signbit(a.toInt()) == std::signbit(b.toInt())) {
+            if ((a.toInt() < 0) == (b.toInt() < 0)) {
               out = EValue(quot);
               return;
             }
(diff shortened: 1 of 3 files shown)


dbort (approved):

kernels/prim_ops/register_prim_ops.cpp
```diff
@@ -148,7 +148,7 @@ static Kernel prim_ops[] = {
           EValue& out = *stack[2];
           if (a.isInt() && b.isInt()) {
             const int64_t quot = a.toInt() / b.toInt();
-            if (std::signbit(a.toInt()) == std::signbit(b.toInt())) {
+            if ((a.toInt() < 0) == (b.toInt() < 0)) {
```
malfet, line 151: This feels unrelated to PR in question, is it?

runtime/platform/compiler.h
```diff
@@ -100,22 +100,38 @@
 #endif // (__cplusplus) >= 202002L
 
 /// Define a C symbol with weak linkage.
+#ifdef _MSC_VER
+// There currently doesn't seem to be a great way to do this in Windows and
+// given that weak linkage is not really critical on Windows, we'll just leave
+// it as a stub.
+#define ET_WEAK
```
chinazhangchao, line 107: Hi, we are adding `Windows x64 Build Support` https://github.com/pytorch/executorch/pull/6979. I use `.\install_requirements.bat --pybind xnnpack` to build on windows. And we encountered duplicate symbol problem. 
image
I investigated and found that windows didn't support weak symbol. Do you have some insights how to resolve this?

PR #6637, opened 2024-11-04
Refactor pytest config + add default dump dir option

Erik-Lundell:
Introduce arm_test_options Enum to avoid having to remember string names of options.

If pytest is run with --default_dump_path=<some_dir>, .dump* functions will print to a logfile in that dir.
Additionally, pytest will clean up the folder so that only the 10 latest logs are saved.

Change-Id: I932489cf08986ae820e74cd865489ef5561a6444

### DIFF
--- a/backends/arm/test/common.py
+++ b/backends/arm/test/common.py
@@ -11,6 +11,10 @@ import shutil
 import subprocess
 import sys
 import tempfile
+from datetime import datetime
+from enum import auto, Enum
+from pathlib import Path
+from typing import Any
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


backends/arm/test/common.py
```diff
@@ -19,27 +23,46 @@
...
+
+
+_test_options: dict[arm_test_options, Any] = {}
 
 # ==== Pytest hooks ====
 
 
 def pytest_addoption(parser):
     parser.addoption("--arm_quantize_io", action="store_true")
     parser.addoption("--arm_run_corstone300", action="store_true")
```
digantdesai, line 41: drop `arm_` prefix?

digantdesai (approved):

PR #6489, opened 2024-10-24
Update Hugging Face version

guangy10:
Bump version to [v4.46.0](https://github.com/huggingface/transformers/releases/tag/v4.46.0) where more `transformers` models are compatible with ExecuTorch out-of-the-box.
Consolidate to use a single version of `transformers` in anywhere in the codebase, e.g. `examples/`, `.workflow/`

~~However, we are hardcode `lm_eval` to a very old version `0.4.2` which is not compatible with `transformers >= 4.45` (actually incompatible with `tokenizers >= 0.20`). It must be upgraded, but not in this PR. Since the eval is using the tinyllama in the CI, the version of `transformers` doesn't matter. The workaround is to force reinstall a lower version in the CI.~~

PR #6657, opened 2024-11-05
Dev weight sharing

haowhsu-quic:
### Summary
- support multiple graphs in single qnn context in runtime
- helper function in aot for generating multi-methods pte
- enable weight sharing mechanism on HTP
- test cases

### Test plan
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_multi_graphs -s $DEVICE_SN -m SM8650 -b build-android/ -a $ARTIFACTS`


backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
```diff
@@ -47,20 +47,100 @@ class PyQnnManager {
...
+      }
+      auto context = qcir::GetContext(info.ptr);
+      for (const auto& graph : *context->graphs()) {
+        std::vector<flatbuffers::Offset<qcir::Tensor>> tensors;
+        for (const auto tensor : *graph->tensors()) {
+          // flatbuffers::Offset<Tensor> ToTensor(
+          //   QnnTensor
+          //   ToTensor(flatbuffers::Vector<::flatbuffers::Offset<qcir::Tensor>>
+          //   tensor), flatbuffers::FlatBufferBuilder* builder);
+          tensors.emplace_back(ToTensor(ToTensor(tensor), &builder_));
```
cccclai, line 76: What does `ToTensor(ToTensor(tensor), &builder_)) mean?

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
```diff
@@ -47,20 +47,100 @@ class PyQnnManager {
...
+      py::buffer_info info(py::buffer(qcirs[i].cast<py::bytes>()).request());
+      flatbuffers::Verifier verifier(
+          static_cast<const uint8_t* const>(info.ptr),
+          info.size * info.itemsize);
+
+      if (!qcir::VerifyContextBuffer(verifier)) {
+        QNN_EXECUTORCH_LOG_ERROR("Fail to verify qcir format");
+        return;
+      }
+      auto context = qcir::GetContext(info.ptr);
```
cccclai, line 68: Does `context` mean `context_binary` here? Does each qcir flatbuffers combined with a context binary?

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
cccclai: I'm guessing here we'll deduplicate tensor?

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
```diff
@@ -47,20 +47,100 @@ class PyQnnManager {
...
+  }
+
   py::array_t<char> Compile(
+      const std::string& graph_name,
       std::vector<std::shared_ptr<OpWrapper>>& op_wrappers) {
     QnnExecuTorchContextBinary context_binary;
     flatbuffers::FlatBufferBuilder builder;
 
-    if (qnn_manager_->IsOnlinePrepare()) {
+    if (qnn_manager_->IsOnlinePrepare() || qnn_manager_->IsMultipleGraphs()) {
```
cccclai, line 143: Does it mean qnn manager can support both online prepare and multiple graph?

backends/qualcomm/partition/qnn_partitioner.py
```diff
@@ -63,7 +64,11 @@ def is_node_supported(self, _, node: torch.fx.Node) -> bool:
             )
             return False
 
-        if node.target in allow_list_operator:
+        if (
+            node.target in allow_list_operator
+            # bypass if custom op appears
```
cccclai, line 69: what is the custom op namespace?

backends/qualcomm/qnn_preprocess.py
```diff
@@ -104,7 +104,8 @@ def preprocess(
             else:
                 raise RuntimeError(f"{node.op} is not supported in Qnn")
         qnn_context_binary = qnn_manager.Compile(
-            [py_op_wrapper.GetOpWrapper() for py_op_wrapper in py_op_wrapper_list]
+            qnn_manager.GetGraphNames()[0],
```
cccclai, line 107: what does this line mean?

backends/qualcomm/runtime/QnnExecuTorchBackend.cpp
```diff
@@ -60,124 +59,19 @@ Result<DelegateHandle*> QnnExecuTorchBackend::init(
       "Fail to initialize Qnn Manager");
 
   if (qnn_manager->IsOnlinePrepare()) {
-    auto graph = qcir::GetGraph(qnn_context_blob.buffer);
```
cccclai, line 63: where does the online prepare option go?

backends/qualcomm/runtime/QnnManager.cpp
```diff
@@ -465,16 +475,158 @@ bool QnnManager::IsNodeSupportedByBackend(
...
+      backend_params_ptr_->qnn_context_ptr_->GetContextBinary(
+          qnn_executorch_context_binary) == Error::Ok,
+      Internal,
+      "Fail to get context binary.");
+
+  return Error::Ok;
+}
+
+Error QnnManager::CompileQcir() {
+  // check if context binary came from flatbuffer
```
cccclai, line 490: If I understand correctly, it's for compiling flatbuffers graph so we can merge the graph AOT?

backends/qualcomm/runtime/backends/QnnBackendCache.cpp
```diff
@@ -126,13 +109,16 @@ Error QnnBackendCache::Configure() {
   Error status = GetQnnGraphInfoFromBinary();
   if (status == Error::Internal) {
     // check if context binary came from flatbuffer
-    flatbuffers::FlatBufferBuilder builder;
     flatbuffers::Verifier verifier(
         static_cast<const uint8_t* const>(qnn_context_blob_.buffer),
         qnn_context_blob_.nbytes);
 
-    if (qcir::VerifyGraphBuffer(verifier)) {
+    if (qcir::VerifyContextBuffer(verifier)) {
```
cccclai, line 116: Trying to follow - are these logic for AOT or runtime?

backends/qualcomm/runtime/backends/htpbackend/HtpGraph.h
```diff
@@ -23,10 +23,9 @@ class HtpGraph : public QnnGraph {
       QnnBackend* backend,
       QnnContext* context,
       const QnnExecuTorchProfileLevel& profile_level,
-      const std::string& graph_name,
```
cccclai, line 26: interesting - what did we use this `graph_name` before?

backends/qualcomm/serialization/schema.fbs
```diff
@@ -177,7 +181,10 @@ table QnnExecuTorchOptions {
   shared_buffer:bool;
 
   /// Is model from qnn context binary
-  is_from_context_binary: bool;
+  is_from_context_binary:bool;
```
cccclai, line 184: oh is it only for the custom op solution?

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -1563,6 +1564,44 @@ def test_qnn_backend_multi_contexts_composite(self):
...
+        compiler_specs = [
+            generate_qnn_executorch_compiler_spec(
+                soc_model=self.chipset_table[TestQNN.model],
+                backend_options=backend_options,
+                multiple_graphs=True,
+                graph_name=graph_name,
+            )
+            for graph_name in graph_names
+        ]
+        exported_programs = [
```
cccclai, line 1589: Great! Did we observe smaller model size compared with no weight sharing option?

examples/qualcomm/utils.py
```diff
@@ -140,7 +140,7 @@ def push(self, inputs=None, input_list=None, files=None):
             for file_name in files:
                 self._adb(["push", file_name, self.workspace])
 
-    def execute(self, custom_runner_cmd=None):
+    def execute(self, custom_runner_cmd=None, method_index=0):
```
cccclai, line 143: Is it because we don't have the method name but just method index?

cccclai (commented):
Looks fairly clean, thanks! I just asked some questions so I can understand it better.

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
haowhsu-quic: The inner `ToTensor` is used to convert serialized tensor in `qcir` to `QnnTensor` defined in QNN SDK header. The outer `ToTensor` is used to convert `QnnTensor` to flatbuffer API compatible tensor for building `qcir`.

Looks like flatbuffer has no mechanism for merging binaries, this is the detour I can come up with so far.
Will rephrase the comment for better understanding.

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
haowhsu-quic: Yes, the concept is similar in `qcir`. The flatbuffer does not contain context binary but graph architecture and tensor data.

backends/qualcomm/aot/python/PyQnnManagerAdaptor.h
haowhsu-quic: This `Compile` method is invoked in `qnn_preprocess.py`. Once one of these two compiler specs are recognized, `qcir` will be returned instead of generating context binary.
If in `online_prepare` mode, user could directly ship the generated `pte` and let `QnnManager` compose graph on device side.
Although `multiple_graph` gives the same binary format that could be used in the same scenario of `online_prepare`. We would expect user follow the example in our test cases, because the optimization level in HTP will be different (host side will be higher and generating more computation efficient context binary).

backends/qualcomm/partition/qnn_partitioner.py
haowhsu-quic: `qaisw` for now.

backends/qualcomm/qnn_preprocess.py
haowhsu-quic: Currently we will have multiple graphs inside one `QnnManager`, the exposed APIs need graph name as an identifier to manipulate. But in this stage, there will be only one graph for processing where the graph name comes from compiler specs.

backends/qualcomm/runtime/QnnExecuTorchBackend.cpp
haowhsu-quic: If I understand the question correctly, the `online_prepare` is still working but the processing logic is refactored into `CompileQcir` which could also be leveraged in compiling `multiple_graphs`.

backends/qualcomm/runtime/QnnManager.cpp
haowhsu-quic: This is used for compiling `qcir`, but the merging of `qcir` happened earlier in `PyQnnManagerAdapter.h`'s constructor which consumes multiple flatbuffers.

backends/qualcomm/runtime/backends/QnnBackendCache.cpp
haowhsu-quic: If for `online_prepare`, this will happen in runtime.
If for `multiple_graphs` this will happen in AoT, since we will again compile the merged `qcir` in host side with higher optimization level in HTP.

backends/qualcomm/runtime/backends/htpbackend/HtpGraph.h
haowhsu-quic: Before we use it to store `graph_name` from compiler specs in AoT or from context binary in runtime.

backends/qualcomm/serialization/schema.fbs
haowhsu-quic: Yes, we need this flag to guarantee the order of graph IOs.

backends/qualcomm/tests/test_qnn_delegate.py
haowhsu-quic: Yes.

examples/qualcomm/utils.py
haowhsu-quic: I think that's related to the interface we used in `qnn_executor_runner`:
```c++
  const char* method_name = nullptr;
  {
    const auto method_name_result =
        program->get_method_name(FLAGS_method_index);
    ET_CHECK_MSG(method_name_result.ok(), "Program has no methods");
    method_name = *method_name_result;
  }
  ET_LOG(Info, "Using method %s", method_name);
```

haowhsu-quic:
It just came to my mind that the memory footprint could be still high in runtime since we will create one `QnnManager` for one method.
Will it be possible if the `method_name` proposed in #6622 could be enumerated in `BackendInitContext` if there are multiple methods map to the same processed bytes? And the actual method to be executed could be derived via `BackendExecutionContext`.

cccclai:
Hmm looks like it needs to inject runtime info, and it needs more internal discussion and feel like we may not be able to ship it on time....

Also just trying to follow, what does init look like in your case? If we can add the method name in `BackendExecutionContext`, will it work for you?

If you need to iterate all methods during init, feel like we may need to hack it via compile specs, like passing methods names as part of the compile specs...

haowhsu-quic:
We will parse the context binary and create graphs inside `init` function, all the graphs are well-prepared in that moment.
What I hope is to reuse that created context and for user to select which graph to execute, but this seems to be constrained by framework. We have to re-initialize the same context again when user allocate another method.

cccclai:
> We have to re-initialize the same context again when user allocate another method.

Is it possible to have it as part of the cache, or is it not a clean solution?

haowhsu-quic:
> > We have to re-initialize the same context again when user allocate another method.
> 
> Is it possible to have it as part of the cache, or is it not a clean solution?

I have thoughts if they are not too hacky for you:
```c++
// a map to maintain hash value of context binaries with corresponding initialized delegate handle
class QnnExecuTorchBackend final : public ::executorch::runtime::BackendInterface {
 private:
  mutable std::unordered_map<uint64_t, executorch::runtime::DelegateHandle*> delegate_map_;
};
Result<DelegateHandle*> QnnExecuTorchBackend::init(
    BackendInitContext& context,
    FreeableBuffer* processed,
    ArrayRef<CompileSpec> compile_specs) const {
  ...
  // check if the processed bytes have already been initialized
  uint64_t hash_val = calculate_hash(processed);
  auto iter = delegate_map_.find(hash_val);
  if (iter != delegate_map_.end()) {
    return iter->second;
  }
  ...
  return delegate_map[hash_val] = qnn_manager;
}
```
With this approach, current implementation and #6622 might still be leveraged.

cccclai:
Oh I think it still work! Maybe we can add a todo to remove the hack. In the meanwhile, do you need the method name during execute too?

haowhsu-quic:
Will do, and I think `method_name` in `BackendInitContext` is enough for current approach.

haowhsu-quic:
> Oh I think it still work! Maybe we can add a todo to remove the hack. In the meanwhile, do you need the method name during execute too?

I realized that we only have one `QnnExecuTorchBackend` instance during runtime, the `method_name` should be shipped via `BackendExecutionContext` instead, or the stored `method_name` will be overwritten (I wrote a simple app having multiple methods loaded to trigger that issue).
Could you help make that happen? thank you!

cccclai:
> > Oh I think it still work! Maybe we can add a todo to remove the hack. In the meanwhile, do you need the method name during execute too?
> 
> I realized that we only have one `QnnExecuTorchBackend` instance during runtime, the `method_name` should be shipped via `BackendExecutionContext` instead, or the stored name will be overwritten (I use a simple app with multiple methods loaded to trigger that issue). Could you help make that happen? thank you!

I think it should be fine to have method name for `BackendInitContext` and `BackendExecutionContext`. Will make the change

cccclai:
Hey I just add method name for execute in https://github.com/pytorch/executorch/pull/6622, can you try again? Sorry was a bit late on this.

cccclai:
Hmm let me check that too…

cccclai:
What is your command to repro? I wanted to iterate my change with your PR

haowhsu-quic:
After applying 6622.patch, I use the `pte` generated from `TestQNNQuantizedUtils.test_qnn_backend_multi_graphs`. Change `qnn_executor_runner` to load both methods `seq_conv`, `single_conv` and execute them.
To check the context binary reuse effect, the logs should contain "Use cached delegate handle for current method: ..." from following snippet in `backends/qualcomm/`runtime/QnnExecuTorchBackend.cpp`:
```cpp
  new (qnn_manager) QnnManager(qnn_executorch_options, qnn_context_blob);

  // TODO: this is a temporal solution for multi-graph support, will be
  //       removed once framework starts to accept runtime configuration
  // ---
  // check if current context binary has already been initialized
  // return cached one for reducing memory footprint
  std::string binary_hash = qnn_manager->GetBinaryHash();
  auto iter = delegate_map_.find(binary_hash);
  if (iter != delegate_map_.end()) {
    QNN_EXECUTORCH_LOG_INFO(
        "Use cached delegate handle for current method: %s",
        context.get_method_name());
    return iter->second;
  }
```
No context binary with the same md5sum would be initialized by HTP again, but I guess here is a thing need your advice: Is there a macro to roll back allocated `qnn_manager` in allocator? Since it's only used for extracting context binary hash here.

To check the effect of `weight sharing`, you could comment out function body inside `qualcomm/runtime/backends/htpbackends/x86_64/HtpContextCustomConfig.cpp` to see the difference of context binary size (55640 w/o and 51552 w/ `weight sharing`).

cccclai:
Does the latest commit in https://github.com/pytorch/executorch/pull/6622 work for you? The CI seems passing and I updated the unit test to test the method name during execute.

haowhsu-quic:
Yes, thank you for the change. It looks good on my side.
Will resubmit the final version once #6622 lands and all internal verification passed.

haowhsu-quic:
Btw, could you keep the `method_name` shipping part in init time? It seems the `BackendInitContext` related implementation has been stripped.

cccclai:
I'm landing https://github.com/pytorch/executorch/pull/6622 and the method name is available on both init and execute. Let me know if there is any issue

haowhsu-quic:
Hi @cccclai, this is the final version fully verified internally. Sorry for the huge change but it's kind of inevitable for everything to work as usual.
I do not rebase onto the latest commit because of #6489 introducing incompatible transformer module which will break `MobileBert` example so as our test cases.

cccclai:
Is this PR good for review? I think the CI is failing: `test-llama-runner-qnn-linux`, `Lint / lintrunner / linux-job`

haowhsu-quic:
> Is this PR good for review? I think the CI is failing: `test-llama-runner-qnn-linux`, `Lint / lintrunner / linux-job`

Sorry for the mistake, I just submit the fix and will add it to our internal CI.

cccclai:
There are quite a bit internal errors, can you apply the following patches, and I'll import and see the internal errors are gone? Thanks!

```
diff --git a/executorch/backends/qualcomm/aot/python/targets.bzl b/executorch/backends/qualcomm/aot/python/targets.bzl
--- a/executorch/backends/qualcomm/aot/python/targets.bzl
+++ b/executorch/backends/qualcomm/aot/python/targets.bzl
@@ -31,6 +31,7 @@
             "//executorch/backends/qualcomm/aot/wrappers:wrappers",
             "//executorch/backends/qualcomm/runtime:logging",
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/backends/qualcomm/aot/ir:qcir_utils",
             "//executorch/backends/qualcomm/runtime:runtime",
             "fbsource//third-party/qualcomm/qnn/qnn-{0}:api".format(get_qnn_library_verision()),
diff --git a/executorch/backends/qualcomm/runtime/targets.bzl b/executorch/backends/qualcomm/runtime/targets.bzl
--- a/executorch/backends/qualcomm/runtime/targets.bzl
+++ b/executorch/backends/qualcomm/runtime/targets.bzl
@@ -29,6 +29,7 @@
         ],
         exported_deps = [
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/runtime/core:core",
         ],
     )
@@ -63,6 +64,7 @@
             "fbsource//third-party/qualcomm/qnn/qnn-{0}:api".format(get_qnn_library_verision()),
             ":logging",
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/backends/qualcomm/aot/ir:qcir_utils",
             "//executorch/backends/qualcomm/aot/wrappers:wrappers",
             "//executorch/runtime/backend:interface",
diff --git a/executorch/backends/qualcomm/targets.bzl b/executorch/backends/qualcomm/targets.bzl
--- a/executorch/backends/qualcomm/targets.bzl
+++ b/executorch/backends/qualcomm/targets.bzl
@@ -16,6 +16,12 @@
 
 SCHEMA_LIRRARY_NAME = SCHEMA_NAME
 
+QC_BINARY_INFO_SCHEMA = "qc_binary_info"
+QC_BINARY_INFO_INPUT_SCHEMA = "serialization/" + QC_BINARY_INFO_SCHEMA + ".fbs"
+QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME = QC_BINARY_INFO_SCHEMA + "_generated"
+QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER = QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME + ".h"
+QC_BINARY_INFO_SCHEMA_LIRRARY_NAME = QC_BINARY_INFO_SCHEMA
+
 def generate_schema_header(rule_name, srcs, headers, default_header):
     """Generate header file given flatbuffer schema
     """
@@ -77,6 +83,33 @@
         platforms = [ANDROID],
     )
 
+    generate_schema_header(
+        QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME,
+        [QC_BINARY_INFO_INPUT_SCHEMA],
+        [QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER],
+        QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER,
+    )
+
+    runtime.cxx_library(
+        name = "qc_binary_info_schema",
+        srcs = [],
+        visibility = [
+            # Lock this down as tightly as possible to ensure that flatbuffers
+            # are an implementation detail. Ideally this list would only include
+            # //executorch/runtime/executor/...
+            "//executorch/codegen/tools/...",
+            "//executorch/runtime/executor/...",
+            "//executorch/backends/qualcomm/...",
+            "//executorch/backends/qualcomm/runtime/...",
+        ],
+        exported_headers = {
+             QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER: ":{}[{}]".format( QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME,  QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER),
+        },
+        exported_external_deps = ["flatbuffers-api"],
+        define_static_target = True,
+        platforms = [ANDROID],
+    )
+
     runtime.cxx_library(
         name = "qnn_executorch_backend",
         srcs = [],
diff --git a/xplat/executorch/backends/qualcomm/aot/python/targets.bzl b/xplat/executorch/backends/qualcomm/aot/python/targets.bzl
--- a/xplat/executorch/backends/qualcomm/aot/python/targets.bzl
+++ b/xplat/executorch/backends/qualcomm/aot/python/targets.bzl
@@ -31,6 +31,7 @@
             "//executorch/backends/qualcomm/aot/wrappers:wrappers",
             "//executorch/backends/qualcomm/runtime:logging",
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/backends/qualcomm/aot/ir:qcir_utils",
             "//executorch/backends/qualcomm/runtime:runtime",
             "fbsource//third-party/qualcomm/qnn/qnn-{0}:api".format(get_qnn_library_verision()),
diff --git a/xplat/executorch/backends/qualcomm/runtime/targets.bzl b/xplat/executorch/backends/qualcomm/runtime/targets.bzl
--- a/xplat/executorch/backends/qualcomm/runtime/targets.bzl
+++ b/xplat/executorch/backends/qualcomm/runtime/targets.bzl
@@ -29,6 +29,7 @@
         ],
         exported_deps = [
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/runtime/core:core",
         ],
     )
@@ -63,6 +64,7 @@
             "fbsource//third-party/qualcomm/qnn/qnn-{0}:api".format(get_qnn_library_verision()),
             ":logging",
             "//executorch/backends/qualcomm:schema",
+            "//executorch/backends/qualcomm:qc_binary_info_schema",
             "//executorch/backends/qualcomm/aot/ir:qcir_utils",
             "//executorch/backends/qualcomm/aot/wrappers:wrappers",
             "//executorch/runtime/backend:interface",
diff --git a/xplat/executorch/backends/qualcomm/targets.bzl b/xplat/executorch/backends/qualcomm/targets.bzl
--- a/xplat/executorch/backends/qualcomm/targets.bzl
+++ b/xplat/executorch/backends/qualcomm/targets.bzl
@@ -16,6 +16,12 @@
 
 SCHEMA_LIRRARY_NAME = SCHEMA_NAME
 
+QC_BINARY_INFO_SCHEMA = "qc_binary_info"
+QC_BINARY_INFO_INPUT_SCHEMA = "serialization/" + QC_BINARY_INFO_SCHEMA + ".fbs"
+QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME = QC_BINARY_INFO_SCHEMA + "_generated"
+QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER = QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME + ".h"
+QC_BINARY_INFO_SCHEMA_LIRRARY_NAME = QC_BINARY_INFO_SCHEMA
+
 def generate_schema_header(rule_name, srcs, headers, default_header):
     """Generate header file given flatbuffer schema
     """
@@ -77,6 +83,33 @@
         platforms = [ANDROID],
     )
 
+    generate_schema_header(
+        QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME,
+        [QC_BINARY_INFO_INPUT_SCHEMA],
+        [QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER],
+        QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER,
+    )
+
+    runtime.cxx_library(
+        name = "qc_binary_info_schema",
+        srcs = [],
+        visibility = [
+            # Lock this down as tightly as possible to ensure that flatbuffers
+            # are an implementation detail. Ideally this list would only include
+            # //executorch/runtime/executor/...
+            "//executorch/codegen/tools/...",
+            "//executorch/runtime/executor/...",
+            "//executorch/backends/qualcomm/...",
+            "//executorch/backends/qualcomm/runtime/...",
+        ],
+        exported_headers = {
+             QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER: ":{}[{}]".format( QC_BINARY_INFO_SCHEMA_GEN_RULE_NAME,  QC_BINARY_INFO_OUTPUT_SCHEMA_HEADER),
+        },
+        exported_external_deps = ["flatbuffers-api"],
+        define_static_target = True,
+        platforms = [ANDROID],
+    )
+
     runtime.cxx_library(
         name = "qnn_executorch_backend",
         srcs = [],

```

cccclai (approved):
Thanks! Looks good

cccclai:
Hmm more tests start failing now, because the `schema.fbs` and the bzl rule is renamed...

extension/llm/export/partitioner_lib.py
```diff
@@ -160,10 +160,8 @@ def get_qnn_partitioner(
             QnnPartitioner,
         )
 
-        # pyre-ignore: Undefined import [21]: Could not find a module corresponding to import `executorch.backends.qualcomm.serialization.qnn_compile_spec_schema`
-        from executorch.backends.qualcomm.serialization.qnn_compile_spec_schema import (
```
cccclai, line 164: Actually the error is 
```
ModuleNotFoundError: No module named 'executorch.backends.qualcomm.serialization.qnn_compile_spec_schema`
```
Seems like it's removed somewhere

haowhsu-quic: Many thanks for helping this!

PR #6611, opened 2024-11-01
[AOSP] add file descriptor support to file_data_loader

davidlin54:
Summary:
For Google's ODP, the model will execute in a process that does not have access to the file system. Instead, the caller must open the file then pass ownership of the FD to the model executor.

1. Added support for reading files by FD passed by the prefix "fd:///" (copied from tensorflow https://cs.android.com/android/platform/superproject/main/+/main:external/federated-compute/fcp/tensorflow/file_descriptor_filesystem.cc;l=41;bpv=1;bpt=1) into `file_data_loader`.
2. Added test cases by opening file first and passing the FD to the data loader

Test Plan:
1. run C++ tests
2. use script with following instruction:
```
$ cd aosp
$ source build/envsetup.sh
$ lunch aosp_arm64-trunk_staging-userdebug
$ m executor_runner
$ FILENAME="<model_path>"
$ exec {FD}<${FILENAME}     # open file for read, assign descriptor
$ echo "Opened ${FILENAME} for read using descriptor ${FD}"
$ out/host/linux-x86/bin/executor_runner --model_path fd:///${FD} --is_fd_uri=true
$ exec {FD}<&-    # close file
```
result:
```
Opened out/host/linux-x86/bin/add.pte for read using descriptor 10
I 00:00:00.000256 executorch:executor_runner.cpp:94] Model file fd:///10 is loaded.
I 00:00:00.000279 executorch:executor_runner.cpp:103] Using method forward
I 00:00:00.000282 executorch:executor_runner.cpp:150] Setting up planned buffer 0, size 48.
I 00:00:00.000294 executorch:executor_runner.cpp:173] Method loaded.
I 00:00:00.000310 executorch:executor_runner.cpp:183] Inputs prepared.
I 00:00:00.000336 executorch:executor_runner.cpp:192] Model executed successfully.
I 00:00:00.000345 executorch:executor_runner.cpp:196] 1 outputs: 
Output 0: tensor(sizes=[1], [2.])
```

PR #6662, opened 2024-11-05
[aosp] Fix internal build error by adding static

davidlin54:
### Summary
https://github.com/pytorch/executorch/pull/6611 causes internal builds to break due to missing static qualifier. adding it here

### Test plan
same test plan as https://github.com/pytorch/executorch/pull/6611

### DIFF
--- a/extension/data_loader/file_data_loader.cpp
+++ b/extension/data_loader/file_data_loader.cpp
@@ -76,7 +76,7 @@ FileDataLoader::~FileDataLoader() {
   ::close(fd_);
 }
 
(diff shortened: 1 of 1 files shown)


kirklandsign (approved):

shoumikhin:
@kirklandsign how shall we proceed with this?

kirklandsign:
No longer needed. Fixed already as https://github.com/pytorch/executorch/blob/main/extension/data_loader/file_descriptor_data_loader.cpp

PR #6665, opened 2024-11-05
[ExecuTorch] Proof-of-concept: use c10/macros in ExecuTorch

swolchok:
* __->__ #6665

Step 0 for code sharing: can we use c10 Macros?

This needs a guardrail to prevent breaking the ExecuTorch core
requirements before we can ship it. What's our current guard against
accidentally including streams/heap allocation/etc. in core at runtime?

Differential Revision: [D65241695](https://our.internmc.facebook.com/intern/diff/D65241695/)


swolchok:
we'll focus on kernel code sharing for now.

PR #6673, opened 2024-11-05
Refactor Init function arg

cccclai:
Summary:
The actual init function takes in backend init context instead of memory allocator directly. See https://github.com/pytorch/executorch/blob/836d5561a61877507b6d5891485725996bb6b32c/runtime/backend/interface.h#L81-L84

Update the signature in the test so it's easier to add test for exposing method name in next diff

Differential Revision: D65499276


iseeyuan (approved):
LGTM

PR #6674, opened 2024-11-05
adding suppression tags to improve autodeps noise

azad-meta:
Summary: The files updated in this diff most likely require manual fix to pass autodeps. This change is to reduce autodeps noise.

Reviewed By: Picowchew

Differential Revision: D65497022


GregoryComer (approved):
Looks like this PR is a bit old (sorry about that), but looks good to me. @azad-meta Do you still want to land this? If you can sign the CLA, we can merge it. Thanks.

PR #6682, opened 2024-11-06
[data loader] move logic for FD data loader out of file_data_loader

davidlin54:
revert changes in https://github.com/pytorch/executorch/pull/6611, and move logic to new data loader to:
1. decouple data loaders
2. fix binary size test


lucylq (approved):
thanks for placing the logic into a new data loader and fixing size test

kirklandsign (approved):

PR #6687, opened 2024-11-06
Revert changes to executor_runner

davidlin54:
Seg fault was introduced to https://github.com/pytorch/executorch/pull/6682 due to moving logic to a helper function and variable lifetimes changing.

Revert executor_runner and fix seg fault tests.


davidlin54:
@digantdesai this should unblock the trunk test that's failing

digantdesai (approved):

PR #6690, opened 2024-11-06
Search graph for quantization parameters

Erik-Lundell:
(This change was merged but had to be reverted due to type checking errors, see https://github.com/pytorch/executorch/pull/6452)

Generalizes the search for quantization parameters. The idea is to make a graph like this (part of) a valid quantized graph:

dq -> view -> transpose -> some_op
dq ------> expand -----------/^

For a subset of operations 'passable_op' that don't modify the values of the tensors it is is allowed to "pass through" the op when searching for qparams. If multiple qparams are encounterd in one search, they are asserted to be equal.

The reason for this is to unify what a quantized graph looks like, removing the need for some ad hoc exceptions (for example checking if a view is before an addmm etc.) that were in place before and in general cleaning up the handling of quantization in the ArmBackend. In particular, some decompositions in the to_edge step seem to not insert q/ dq ops which breaks the pattern of dq -> op -> q.

With this change, Arm internal passes can also in some cases skip inserting quantize ops, simplifying the passes.

### DIFF
--- a/backends/arm/_passes/annotate_channels_last_dim_order_pass.py
+++ b/backends/arm/_passes/annotate_channels_last_dim_order_pass.py
@@ -14,7 +14,7 @@ from executorch.backends.arm._passes.arm_pass_utils import (
     get_first_fake_tensor,
     insert_q_dq_pair,
 )
-from executorch.backends.arm.tosa_quant_utils import dq_op, q_op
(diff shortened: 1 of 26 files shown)


freddan80:
Rebase pls

freddan80 (approved):

PR #6692, opened 2024-11-06
Remove PATTERNLINT suppressions

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #6693
* __->__ #6692

The linter these are suppressing was removed in D63859186.

Differential Revision: [D65507718](https://our.internmc.facebook.com/intern/diff/D65507718/)

PR #6693, opened 2024-11-06
Add NOSTDINC and NOTORCHINC lintrunner

swolchok:
* __->__ #6693
* #6692

Prohibit standard C++ container, aten, and torch includes in core, at least pending further review. Note that non-core directories, such as extension/, test/, and any testing_util/ subdirectories, are exempted.

Differential Revision: [D65541332](https://our.internmc.facebook.com/intern/diff/D65541332/)


.lintrunner.toml
```diff
@@ -196,3 +196,87 @@ command = [
...
+
+[[linter]]
+code = 'NOTORCHINC'
+include_patterns = [
+    "**/*.c",
+    "**/*.cpp",
+    "**/*.h",
+    "**/*.hpp",
+]
+exclude_patterns = [
```
kimishpatel, line 249: Instead of exclude, can you just do include for runtime directory?

kimishpatel (changes requested):
Did you test that this works?

.lintrunner.toml
swolchok: > Instead of exclude, can you just do include for runtime directory?

directories should be linted for this unless we know that they shouldn't be. We want this to fail closed.

swolchok:
> Did you test that this works?

yes see internal diff test plan

mergennachin (approved):

PR #6696, opened 2024-11-06
Adding tests old lowering flow for op_abs.py

ru-m8:
Summary: Added tests with the legacy mode.

Differential Revision: D65542758


digantdesai (approved):
Thank you @ru-m8

PR #6709, opened 2024-11-07
Use lambda on all kernel prim ops

zingo:
This makes executorch_prim::et_copy_index.tensor and executorch_prim::et_view.default also use a lamba function like the rest. This fixes a problem where the functions pointers didn't get proper values for this two ops.

Tested with a non delegated view op on Arm backend.


freddan80:
@digantdesai since this is touching code outside arm delegate, will you have a look?

digantdesai (approved):

PR #6710, opened 2024-11-07
Arm backend: execute_runner link in portable_kernels

zingo:
Moving linking of libexecutorch.a to after "-Wl,--whole-archive" This adds executorch_prim::* to the build as the executorch_prim ops are added with a c++ static initialize that get optimized away by the linker as the variable in never used in the normal linking. See docs/source/runtime-build-and-cross-compilation.md

Tested on Arm backend using a non delegated view. Without this non of the executorch_prim::<OPS> can be found/used during runtime.


zingo:
Closing this, have a better way to do this that only bring in the view and not all ops

PR #6721, opened 2024-11-07
migrate passes calls to pass utils

zonglinpeng:
Summary: keep the same calling order in oss, but calling by pass util pass register

Differential Revision: D65464704

### DIFF
--- a/backends/cadence/aot/TARGETS
+++ b/backends/cadence/aot/TARGETS
@@ -48,10 +48,11 @@ python_library(
     ],
 )
 
+
 python_library(
-    name = "passes",
+    name = "pass_utils",
     srcs = [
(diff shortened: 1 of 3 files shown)


hsharma35 (approved):

backends/cadence/aot/TARGETS
```diff
@@ -48,10 +48,11 @@ python_library(
     ],
 )
 
+
 python_library(
-    name = "passes",
+    name = "pass_utils",
     srcs = [
-        "_passes.py",
+        "pass_utils.py",
```
mcremon-meta, line 55: @tarun292 do we need the _ here? I forgot the exact reason to add it

PR #6725, opened 2024-11-08
Qualcomm AI Engine Direct - Optimize QNN embedding op for llama

shewu-quic:
summary:
- Change the dtype of the token from int64 to in32 since Int32 is Qnn HTP friendly. It will allow the embedding op to run into backend optimizations to significantly speed up qnn embedding operations.

Test the PR for llama 3.2 1B instruct with seq_len=512 on SM8650
image
Test the mainline
image

### DIFF
--- a/extension/llm/runner/text_prefiller.cpp
+++ b/extension/llm/runner/text_prefiller.cpp
@@ -17,9 +17,11 @@ namespace llm {
 
 TextPrefiller::TextPrefiller(
     TextDecoderRunner* text_decoder_runner,
+    bool use_int32_token,
     bool use_kv_cache,
     bool enable_parallel_prefill)
     : text_decoder_runner_(text_decoder_runner),
+      use_int32_token_(use_int32_token),
       use_kv_cache_(use_kv_cache),
       enable_parallel_prefill_(enable_parallel_prefill) {}
 
@@ -36,12 +38,13 @@ TextPrefiller::TextPrefiller(
 
   // store the token
   uint64_t cur_token;
+  exec_aten::ScalarType token_type = use_int32_token_
+      ? exec_aten::ScalarType::Int
+      : exec_aten::ScalarType::Long;
   if (enable_parallel_prefill_ || !use_kv_cache_) {
     // initialize tensor wrappers
-    auto tokens = from_blob(
-        prompt_tokens.data(),
-        {1, num_prompt_tokens},
-        exec_aten::ScalarType::Long);
+    auto tokens =
+        from_blob(prompt_tokens.data(), {1, num_prompt_tokens}, token_type);
 
     auto start_pos_tensor =
         from_blob(&start_pos, {1}, exec_aten::ScalarType::Long);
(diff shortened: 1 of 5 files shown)


shewu-quic:
Hi @cccclai,
This PR is to change the dtype of token to optimize embedding op.
Could you help take a look?
And then, I will create the following PRs ASAP to speed up the performance with QNN Backend.
1. Delegate mutable buffer
2. Enable llama3.2 on static llama

Thank you for your effort.

cccclai (approved):

examples/models/llama/model.py
```diff
@@ -259,7 +259,7 @@ def get_example_inputs_kvcache_sdpa(self):
         else:
             return (
                 torch.tensor(
-                    [[1]], dtype=torch.long
+                    [[1]], dtype=torch.int32
```
cccclai, line 262: Can we put it in model metadata and get it in runtime to avoid misuse

cccclai:
Hi this PR breaks quite a few internal tests....Since the test are internal, I need to send you the fix patch before landing it..

examples/models/llama/model.py
shewu-quic: Oops, I miss this change. Yes, I just would like to change it with QNN Backend.

extension/llm/runner/text_token_generator.h
```diff
@@ -23,12 +23,14 @@ class ET_EXPERIMENTAL TextTokenGenerator {
   TextTokenGenerator(
       Tokenizer* tokenizer,
       TextDecoderRunner* text_decoder_runner,
+      bool use_int32_token,
```
cccclai, line 26: This is causing BC breaking and causes CI failure. Can we set a default value so it's BC compatible? Also assert when it doesn't match metadata.

cccclai: The CI breaking stack trace can be found [here](https://github.com/pytorch/executorch/actions/runs/11765104349/job/32771040544?fbclid=IwZXh0bgNhZW0CMTEAAR2LBZNme_WdX6YIVmBa83nZswo_ZGZpzE101yaAzaFIJuccgjBRkZRkFPY_aem_UQNM25k9kHiR3Ceg845sog)

extension/llm/runner/text_prefiller.cpp
```diff
@@ -17,9 +17,11 @@ namespace llm {
 
 TextPrefiller::TextPrefiller(
     TextDecoderRunner* text_decoder_runner,
+    bool use_int32_token,
```
cccclai, line 20: oh I meant more like 
```
TextPrefiller::TextPrefiller(
    TextDecoderRunner* text_decoder_runner,
    bool use_kv_cache,
    bool enable_parallel_prefill,
    bool use_int32_token=False)
```
so we don't need to update all callsite...

shewu-quic: Sorry about misunderstanding. I have updated. Thanks!

cccclai:
Is this PR still needed?

shewu-quic:
> Is this PR still needed?

I think we are focusing on static llama maybe we can close this PR.

PR #6726, opened 2024-11-08
[MTK] Add support for Llama 3.2 and code updates to align with current ET API for dynamic dim

neuropilot-captain:
This PR contains the following modifications/udpates:

- Add Llama 3.2 related files
- Fix bug for tie word embeddings weight naming
- Exclude embedding .bin file during weight checking in sanity checks
- Get Embedding layer before loading of main model if calibration dataset provided to prevent tied word embedding weight name from being removed from state dict during main model weight loading
- Add keep_in_memory during dataset mapping function to prevent disk out of space issue

### DIFF
--- a/examples/mediatek/model_export_scripts/llama.py
+++ b/examples/mediatek/model_export_scripts/llama.py
@@ -419,6 +419,9 @@ def main():
     print(f"Max Num Token: {max_num_token}")
     print(f"Max Cache Size: {max_cache_size}")
 
+    if args.dataset is not None:
+        embedding_layer = get_embedding_layer(config, weight_dir, state_dict)
+
     # Instantiate model chunks
     print("Instantiating submodels")
     models = []
@@ -437,7 +440,6 @@ def main():
     cal_dataset = None
     if args.dataset is not None:
         cal_dataset = load_dataset("text", data_files=args.dataset, split="train")
-        embedding_layer = get_embedding_layer(config, weight_dir, state_dict)
         master_rot_emb = get_master_rot_emb(config, dtype=torch.float32)
         if args.preformatter is not None:
             cal_dataset = cal_dataset.map(
@@ -460,6 +462,7 @@ def main():
                 "eos_token_id_tensor": torch.tensor(tokenizer.eos_token_id),
                 "response_cap": args.response_cap,
             },
+            keep_in_memory=True
         )
 
     for chunk_idx, chunk in enumerate(models):
--- a/examples/mediatek/model_export_scripts/llama.py
+++ b/examples/mediatek/model_export_scripts/llama.py
@@ -462,7 +462,6 @@ def main():
                 "eos_token_id_tensor": torch.tensor(tokenizer.eos_token_id),
                 "response_cap": args.response_cap,
             },
-            keep_in_memory=True
         )
 
     for chunk_idx, chunk in enumerate(models):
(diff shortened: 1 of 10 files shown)


examples/mediatek/model_export_scripts/llama.py
```diff
@@ -460,6 +462,7 @@ def main():
                 "eos_token_id_tensor": torch.tensor(tokenizer.eos_token_id),
                 "response_cap": args.response_cap,
             },
+            keep_in_memory=True
```
cccclai, line 465: Mind sharing what is keep in memory?

examples/mediatek/models/llm_models/modeling_common.py
```diff
@@ -757,9 +757,9 @@ def get_example_inputs(
             nt = Dim("num_token", max=num_token)
             cache_dims = tuple(({} for _ in range(2 * self.num_blocks)))
             dynamic_shapes = (
```
cccclai, line 759: I think I may ask in the previous PR - what does `dynamic_shape` do? Probably good to add a comment to explain

examples/mediatek/model_export_scripts/llama.py
```diff
@@ -437,7 +440,6 @@ def main():
     cal_dataset = None
     if args.dataset is not None:
         cal_dataset = load_dataset("text", data_files=args.dataset, split="train")
-        embedding_layer = get_embedding_layer(config, weight_dir, state_dict)
```
cccclai, line 440: Any specific reason to remove this line?

cccclai (approved):
Most nit comments, thank you for adding support for 3.2

cmodi-meta:
Thanks @neuropilot-captain! do we also need to make to the `examples/mediatek/shell_scripts/export_llama.sh` file to include Llama 3.2 1B and 3B model in the if-else conditions?

cccclai:
There are some lint errors. Could you send a fix?

examples/mediatek/model_export_scripts/llama.py
neuropilot-captain: Hi @cccclai, keep in memory stores the dataset in RAM instead of caching it. It was a temporary workaround for an OSError I encountered on my end. I have since resolved the issue and will remove this argument in the next commit.

examples/mediatek/models/llm_models/modeling_common.py
neuropilot-captain: Hi @cccclai, dynamic_shape is passed in to torch.export.export_for_training to indicate that the input shapes during calibration for some of the inputs may be different. After calibration, the dynamic_shape argument is not needed

examples/mediatek/model_export_scripts/llama.py
neuropilot-captain: Oh I didn't remove it. I shifted it above to line 422. The reason being that during chunk.load_weights (line 437), the weights are popped from the original state_dict and for the tie word embedding True case, this would mean that the embedding weights are not present in the state_dict anymore by the time we get the embedding layer, hence I shifted it up

neuropilot-captain:
> Thanks @neuropilot-captain! do we also need to make to the `examples/mediatek/shell_scripts/export_llama.sh` file to include Llama 3.2 1B and 3B model in the if-else conditions?

Hi @cmodi-meta, yes the script would need to be modified to include Llama 3.2 1B and 3B in the if else. I can modify the script and update it in the next commit

PR #6727, opened 2024-11-08
Qualcomm AI Engine Direct - Delegated mutable buffer

shewu-quic:
summary:
- Support copy op with QNN Reshape
- Consume mutable buffer in QNN Delegate
- Set the same memory address for I/O of mutable buffer at runtime

Test the PR for llama 3.2 1B instruct with seq_len=512 on SM8650
image
Test the mainline
image


shewu-quic:
Hi @cccclai,
This PR is to delegate mutable buffer and maintain it in QNN Backend.
I also [added a condition](https://github.com/pytorch/executorch/blob/7f236c3b7d87347bd102bbd61372027d042f8a75/backends/qualcomm/partition/qnn_partitioner.py#L162) to choose whether consuming mutable buffer or not.
Please have a look.

Thank you very much :)

cccclai (approved):
This looks very solid, thanks!

cccclai:
Hey it seems like breaking the CI ` test-llama-runner-qnn-linux`, can you take a look?

shewu-quic:
> Hey it seems like breaking the CI ` test-llama-runner-qnn-linux`, can you take a look?

It seems to be something wrong to delegate mutable buffer without quantize.
I am trying to figure out the root cause. If it is not easy to solve, I will disable delegated mutable buffer in fp mode.

shewu-quic:
It seems that [delegated mutable buffer is not removed from the output](https://github.com/pytorch/executorch/blob/dc41596b1a61ca3de7b5ad35e0692a78d7b185eb/exir/lowered_backend_module.py#L900). 
When I trace back, I found the [mutable buffer doesn't exist in original_program.state_dict ](https://github.com/pytorch/executorch/blob/dc41596b1a61ca3de7b5ad35e0692a78d7b185eb/exir/lowered_backend_module.py#L527). So, [it doesn't be added into output_specs_to_delete](https://github.com/pytorch/executorch/blob/dc41596b1a61ca3de7b5ad35e0692a78d7b185eb/exir/lowered_backend_module.py#L612). Do you have any idea for it?

cccclai:
It seems like the label thing is new...will check how to resolve it

cccclai:
Hello, is this PR still needed? Assuming yes, but we're focusing on static llama now...

shewu-quic:
> Hello, is this PR still needed? Assuming yes, but we're focusing on static llama now...

Yes, I think we can close it. Thanks.

PR #6742, opened 2024-11-08
Run slow model E2E CI on periodic only

kirklandsign:
### Summary
Some model E2E are super slow (get idea from https://hud.pytorch.org/pytorch/executorch/commit/6d6630edae0bdc37c8450e55f2fc0ef90cb5f50c). Remove DL3 and EDSR for now.
### Test plan
Reduced CPU usage

### DIFF
--- a/.ci/scripts/gather_test_models.py
+++ b/.ci/scripts/gather_test_models.py
@@ -88,8 +88,8 @@ def model_should_run_on_event(model: str, event: str) -> bool:
     if event == "pull_request":
         return model in ["mv3", "vit"]
     elif event == "push":
-        # 'emformer_predict' is running super slow. Only run it periodically
-        return model not in ["emformer_predict"]
+        # These are super slow. Only run it periodically
+        return model not in ["dl3", "edsr", "emformer_predict"]
     else:
         return True
 


huydhn (approved):

guangy10:
@kirklandsign On what backend are these models running slow? Is it on xnnpack? I observed extreme slowness today when running previously supported HF models on xnnpack. And it produces garbage outputs (see the CI job on your PR: https://github.com/pytorch/executorch/actions/runs/11750351996/job/32738448869?pr=6742#step:14:10567). I'm wondering the slowness you are seeing is same. If yes, we should not remove/disable the CI/tests.

kirklandsign:
> @kirklandsign On what backend are these models running slow? Is it on xnnpack? I observed extreme slowness today when running previously supported HF models on xnnpack. And it produces garbage outputs (see the CI job on your PR: https://github.com/pytorch/executorch/actions/runs/11750351996/job/32738448869?pr=6742#step:14:10567). I'm wondering the slowness you are seeing is same. If yes, we should not remove/disable the CI/tests.

These are for XNNPACK and portable. EDSR is very slow. Do you think we want to remove HF models on XNNPACK as well?

PR #6744, opened 2024-11-09
Correctly set _GLIBCXX_USE_CXX11_ABI pybind compile options

huydhn:
I finally understand why https://github.com/pytorch/pytorch/pull/139947#issuecomment-2465106794 passed on PyTorch nightly binaries while failed on those that were built from source on CI.  It turns out that `_GLIBCXX_USE_CXX11_ABI` is always unset when building ET pybind.  For some reason what I'm not sure, PyTorch nightly binaries unset `_GLIBCXX_USE_CXX11_ABI` while PyTorch CI sets it, maybe @malfet has some context here.  This explains why the test couldn't find these quantized symbol.  The missing torchao and torchtune was a red herring.

As ET already has `EXECUTORCH_DO_NOT_USE_CXX11_ABI` option, the fix here is just to check for this and set `_GLIBCXX_USE_CXX11_ABI` accordingly.


CMakeLists.txt
```diff
@@ -716,15 +716,27 @@ if(EXECUTORCH_BUILD_PYBIND)
...
+        -D_GLIBCXX_USE_CXX11_ABI=0
+    )
+  else()
+    set(_pybind_compile_options
+        -Wno-deprecated-declarations
+        -fPIC
+        -frtti
+        -fexceptions
+    )
+  endif()
```
malfet, line 738: Nit (avoids duplication)
```suggestion
  set(_pybind_compile_options
      -Wno-deprecated-declarations
      -fPIC
      -frtti
      -fexceptions
  )
  if(EXECUTORCH_DO_NOT_USE_CXX11_ABI)
      # libtorch is built with the old ABI, so we need to do the same for any
      # .cpp files that include torch, c10, or ATen targets. Note that PyTorch
      # nightly binary is built with _GLIBCXX_USE_CXX11_ABI set to 0 while its
      # CI build sets this to 1 (default)
    list(APPEND _pybind_compile_options -D_GLIBCXX_USE_CXX11_ABI=0)
  else()
  endif()
```

malfet (approved):

malfet:
As to reason why this is necessary: Until 2.6 PyTorch was targeting manylinux2014, which still ships with CXX library that does not support C++11 ABI, so wheel binaries where always build with pre-cxx11 ABI (but conda with it, as conda ships newer libstdc++)

huydhn:
> As to reason why this is necessary: Until 2.6 PyTorch was targeting manylinux2014, which still ships with CXX library that does not support C++11 ABI, so wheel binaries where always build with pre-cxx11 ABI (but conda with it, as conda ships newer libstdc++)

Thank you for the explanation! Now that makes sense, I vaguely remember reading about this before.

huydhn:
Going to test this out on https://github.com/pytorch/pytorch/pull/140199 to wait for a green CI signal before landing. I have been able to verified this locally

larryliu0820 (approved):

huydhn:
CI is green now https://github.com/pytorch/pytorch/actions/runs/11752828992/job/32745509211?pr=140199

PR #6745, opened 2024-11-09
Write QNN context binary to file when env variable ET_QNN_DEBUG_DIR is set

limintang:
Differential Revision: D65694854

### DIFF
--- a/backends/qualcomm/runtime/QnnManager.cpp
+++ b/backends/qualcomm/runtime/QnnManager.cpp
@@ -16,7 +16,8 @@
 #include <executorch/extension/tensor/tensor.h>
 #include <algorithm>
 #include <cstdlib>
-#include <cstring>
+
+#include <filesystem>
 #include <fstream>
 #include <string>
 
@@ -679,6 +680,39 @@ Error QnnManager::Compile(
         "Failed to finalize Qnn Graph with error: %d",
         QNN_GET_ERROR_CODE(error));
     return Error::Internal;
+  }
+
+  // Write context binary to file for debugging purpose
+  const char* et_qnn_debug_dir = getenv("ET_QNN_DEBUG_DIR");
+  if (et_qnn_debug_dir != nullptr) {
+    namespace fs = std::filesystem;
+
+    fs::path debug_dir = et_qnn_debug_dir;
+    if (!fs::exists(debug_dir) || !fs::is_directory(debug_dir)) {
+      QNN_EXECUTORCH_LOG_WARN(
+          "ET_QNN_DEBUG_DIR %s is not a valid directory", et_qnn_debug_dir);
+      return Error::Ok;
+    }
+
+    QnnExecuTorchContextBinary qnn_executorch_context_binary{nullptr, 0};
+    if (GetContextBinary(qnn_executorch_context_binary) != Error::Ok) {
+      return Error::Ok;
+    }
+    auto binary_info = GetBinaryInfo(qnn_executorch_context_binary.buffer);
+
+    std::string binary_file =
+        std::string(et_qnn_debug_dir) + "/qnn_context_binary.bin";
+    std::ofstream fout(binary_file, std::ios::binary);
+
+    fout.write(
+        reinterpret_cast<const char*>(binary_info->data()->data()),
+        binary_info->data()->size());
+
+    if (!fout) {
+      QNN_EXECUTORCH_LOG_WARN(
+          "Failed to write QNN context binary to file %s", binary_file.c_str());
+    }
+    fout.close();
   }
 
   return Error::Ok;


cccclai:
@chiwwang mind providing some suggestion on this PR?

chiwwang:
Hi @limintang ,
Could you please share what use-case this PR aim at?
I'm thinking the context binary can be obtained from the .pte file. Or, if we want a pure context-binary without wrapping it into .pte file, we can also write out the file at AOT compilation stage, i.e., this line:
https://github.com/pytorch/executorch/blob/089087b2caf4eb5eefe05fb6fe8fd216b69fb9b7/backends/qualcomm/qnn_preprocess.py#L113


The `qnn_context_binary` should be exactly the QNN context-binary loaded by the runtime.

limintang:
> Hi @limintang , Could you please share what use-case this PR aim at? I'm thinking the context binary can be obtained from the .pte file. Or, if we want a pure context-binary without wrapping it into .pte file, we can also write out the file at AOT compilation stage, i.e., add something like `fp.write` before this line:
> 
> https://github.com/pytorch/executorch/blob/089087b2caf4eb5eefe05fb6fe8fd216b69fb9b7/backends/qualcomm/qnn_preprocess.py#L113
> 
> The `qnn_context_binary` should be exactly the QNN context-binary loaded by the runtime.
We frequently use tools in QNN SDK for various debug and profiling tasks, and these tools consume QNN context binary.  We are working with QC to support functionalities provided by these tools programmably so we don't have to rely on these tool binaries in the future, but until then, saving the context binary in model export is more convenient.

qnn_preprocess.py is another place where the context binary can be written out, actually I did use this path internally for context binary saving. However, the invocation to QNN preprocess is from ExecuTorch tracing, I'm not sure where introducing a platform specific config in the platform agnostic ET module is a good idea.

backends/qualcomm/runtime/QnnManager.cpp
```diff
@@ -680,7 +680,30 @@ Error QnnManager::Compile(
...
+      backend_params_ptr_->qnn_graph_ptr_->GetProfile()->GetQnnProfileLevel();
+  // Write context binary to file for debugging purpose
+  if (qnn_profile_level == QNN_PROFILE_LEVEL_DETAILED) {
+    auto ret = backend_params_ptr_->qnn_context_ptr_->GetContextBinary(
+            qnn_executorch_context_binary);
+    if (ret != Error::Ok) {
+        QNN_EXECUTORCH_LOG_WARN("Fail to get context binary.");
+        return Error::Ok;
+    }
+    static const char* binary_file = "/tmp/qnn_context_binary.bin";
```
chiwwang, line 694: How about we specify this with an environment variable?
It that environment variable is not empty, we dump context-bianries to that folder...etc.
Then we don't need to couple `qnn_profile_level` with this functionality.

chiwwang (commented):
Hi @limintang 

Thanks for the response! It makes sense for me.

Just a further question -- do you think environment variable can be a better idea than coupling `profiling_level` with this need?

I though adding an option to `QnnExecuTorchHtpBackendOptions` but realized that it's not applicable.
Then I think it would be good if we have runtime option to control this functionality, so environment variable seems a choice.

How about defining an env variable like `ET_QNN_BE_DUMP_CNTX_BIN_DIR`, if it is not empty, then we dump the context-binaries to the directory it points to?
Is setting env variable possible in your use-case?

backends/qualcomm/runtime/QnnManager.cpp
limintang: Good idea. Let me try this out.

limintang: @chiwwang Updated to use env variable.

chiwwang (approved):
Looks good for me!

backends/qualcomm/runtime/QnnManager.cpp
```diff
@@ -681,6 +682,36 @@ Error QnnManager::Compile(
...
+
+    QnnExecuTorchContextBinary qnn_executorch_context_binary{nullptr, 0};
+    if (GetContextBinary(qnn_executorch_context_binary) != Error::Ok) {
+      return Error::Ok;
+    }
+
+    std::string binary_file = std::string(et_qnn_debug_dir) + "/qnn_context_binary.bin";
+    std::ofstream fout(binary_file, std::ios::binary);
+    fout.write(
+        reinterpret_cast<const char*>(qnn_executorch_context_binary.buffer),
```
shewu-quic, line 705: If you would like to dump qnn context binary, you should [use GetBinaryInfo to deserialize the flatbuffer](https://github.com/pytorch/executorch/blob/d679ad70a7745c60df581d6b110e6f79c389feb9/backends/qualcomm/runtime/backends/QnnContextCommon.cpp#L49) and use binary_info->data()->data().

cccclai:
Sorry a bit late on actually reading this PR...

From reading the conversation, looks like it's mainly to save the qnn_context_binary file, how about we provide a cli tool to unwrap the `.pte` file to get the qnn_context_binary`, in this case, it can be more generalized. @limintang what do you think?

limintang:
> Sorry a bit late on actually reading this PR...
> 
> From reading the conversation, looks like it's mainly to save the qnn_context_binary file, how about we provide a cli tool to unwrap the `.pte` file to get the qnn_context_binary`, in this case, it can be more generalized. @limintang what do you think?

It would be great if you can provide such tool, then this diff won't be necessary.

PR #6759, opened 2024-11-11
added instrumentation test for LlamaModule

davidlin54:
Added instrumentation test for LlamaModule. 

1. Modified `setup.sh` to include building `stories110M` model and moves it into `src/androidTest/resources`
2. Added test cases for `LlamaModule` by generating a sequence length of 32, and verifying the length. Also verifies that `stop()` works by checking output length is less than input sequence length


davidlin54:
* (to be filled)

kirklandsign (commented):
You can also add it to https://github.com/pytorch/executorch/blob/main/.github/workflows/_android.yml#L99 if you are interested

kirklandsign (approved):
You can also add it to https://github.com/pytorch/executorch/blob/main/.github/workflows/_android.yml#L99 if you are interested

davidlin54:
> You can also add it to https://github.com/pytorch/executorch/blob/main/.github/workflows/_android.yml#L99 if you are interested

will add in a follow up!

PR #6771, opened 2024-11-11
Print the number of tokens generated

helunwencser:
* __->__ #6771

This is useful for verifying the correctness of AttentionSink.

Differential Revision: [D65784095](https://our.internmc.facebook.com/intern/diff/D65784095/)

### DIFF
--- a/examples/models/llama/runner/eager.py
+++ b/examples/models/llama/runner/eager.py
@@ -91,10 +91,11 @@ def main() -> None:
         else runner.text_completion(
             prompt=args.prompt,
             temperature=args.temperature,
+            echo=True,
         )
     )
     if args.show_tokens:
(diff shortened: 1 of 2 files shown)


examples/models/llama/runner/eager.py
```diff
@@ -91,10 +91,11 @@ def main() -> None:
         else runner.text_completion(
             prompt=args.prompt,
             temperature=args.temperature,
+            echo=True,
```
davidlin54, line 94: out of curiousity, what does `echo=True` do?

davidlin54 (approved):

examples/models/llama/runner/eager.py
helunwencser: When it is true, the returned token list will include prompt tokens as well.

PR #6777, opened 2024-11-12
Skip embedding

metascroy:
Skip embedding ops in coreml partitioner.

### DIFF
--- a/extension/llm/export/partitioner_lib.py
+++ b/extension/llm/export/partitioner_lib.py
@@ -142,7 +142,8 @@ def get_coreml_partitioner(
     return CoreMLPartitioner(  # pyre-fixme[16]
         compile_specs=compile_specs,
         take_over_mutable_buffer=take_over_mutable_buffer,
-        # Skip embe
+        # Skip embedding ops in coreml delegation due to
+        # compression config conflicts inside coreml
         skip_ops_for_coreml_delegation=[
             "quantized_decomposed.embedding_4bit.dtype",
             "aten.embedding.default",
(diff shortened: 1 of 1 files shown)


extension/llm/export/partitioner_lib.py
```diff
@@ -142,6 +142,12 @@ def _validate_ios_version() -> None:
     return CoreMLPartitioner(  # pyre-fixme[16]
         compile_specs=compile_specs,
         take_over_mutable_buffer=take_over_mutable_buffer,
+        # Skip embedding ops in coreml delegation due to
```
cccclai, line 145: Is it only applicable to small models or also large models like 7b?

metascroy: I saw same error on llama 1B

cccclai: Right I understand - but this is supposed to be working for 7b

cccclai: I just want to check if this pr will break the claim in https://machinelearning.apple.com/research/core-ml-on-device-llama

cccclai: They currently link to 0.4 branch in the blog post, maybe it's okay, but still not great if we break it

metascroy: I can check the claim.  But which claim in https://machinelearning.apple.com/research/core-ml-on-device-llama specifically do we want to test?

The article is mostly about direct use of the coremltools API and says very little about ExecuTorch, except for one paragraph "Model Export via torch.export and Inference via ExecuTorch" that is light on details.

cccclai: They link to executorch page https://github.com/pytorch/executorch/tree/v0.4.0/examples/models/llama2#step-4-run-on-your-computer-to-validate with some instructions. But anyway, let me tag @YifanShenSZ on this PR

PR #6795, opened 2024-11-12
Fix flaky ET attention test

jackzhxng:
### Summary
Torch's manual seed is environment-based, ~~loosen the tol for torch.allclose~~ use assert_close instead which sets an appropriate atol and rtol based on the dtype, since the previous tol was too strict.

Fixes [flaky mha test](https://hud.pytorch.org/failure?name=pull%20%2F%20unittest%20%2F%20linux%20%2F%20linux-job&jobName=undefined&failureCaptures=%5B%22extension%2Fllm%2Fmodules%2Ftest%2Ftest_mha.py%3A%3AAttentionTest%3A%3Atest_attention_executorch%22%5D) and reverts https://github.com/pytorch/executorch/pull/6790.

### Test plan
Confirmed and reproduced locally by altering the manual seed until there was an issue. Modified unit test to pass locally.

### DIFF
--- a/extension/llm/modules/test/test_attention.py
+++ b/extension/llm/modules/test/test_attention.py
@@ -95,11 +95,7 @@ class AttentionTest(unittest.TestCase):
         et_res = self.et_mha(self.x, self.x)  # Self attention.
         tt_res = self.tt_mha(self.x, self.x)  # Self attention.
 
-        assert_close(
-            et_res,
-            tt_res,
-            msg=f"TorchTune output is not close to ET output.\n\nTorchTune: {tt_res}\nET output: {et_res}",
-        )
+        assert_close(et_res, tt_res)
 
         # test with kv cache
         self.et_mha.setup_cache(1, dtype=torch.float32, max_seq_len=20)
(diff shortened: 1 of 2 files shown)


extension/llm/modules/test/test_mha.py
```diff
@@ -92,7 +93,11 @@ def test_attention_eager(self):
         et_res = self.et_mha(self.x, self.x)  # Self attention.
         tt_res = self.tt_mha(self.x, self.x)  # Self attention.
 
-        self.assertTrue(torch.allclose(et_res, tt_res))
+        assert_close(
+            et_res,
+            tt_res,
+            msg=f"TorchTune output is not close to ET output.\n\nTorchTune: {tt_res}\nET output: {et_res}",
```
swolchok, line 99: I don't think you need the messages, but not blocking land over them

swolchok (approved):

extension/llm/modules/test/test_mha.py
dvorjackz: Would have been useful when I was debugging the error imo

swolchok: yes but I think assert_close gives you nice messages

PR #6797, opened 2024-11-12
Fix imports to use torchao public api

jainapurva:
Summary: Fixing imports to use torchao.utils API, instead of a nested call from torchao.quantization API.

Differential Revision: D65843742


jerryzh168 (approved):
is this the only one?

PR #6800, opened 2024-11-12
share embeddding and output

cccclai:
Differential Revision: D64189995


cccclai:
@chiwwang here is one thing we can do to optimize the memory. For 1b model, the embedding is shared with the last linear weight, so ideally, we can save more memory if it's enabled.

chiwwang:
@shewu-quic for awareness. This might also help us reduce memory usage.

PR #6835, opened 2024-11-13
Add PR release note label checker bot

jackzhxng:
### Summary
Add gh actions workflow for checking that there is a `Release notes:` or `topic: not user facing` label is added to each PR.

### Test plan
Testing with this PR

### DIFF
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -1,0 +1,54 @@
+name: Check Labels
+
+on:
+  # We need pull_request_target to be able to post comments on PRs from forks.
+  # Only allow pull_request_target when merging to main, not some historical branch.
+  #
+  # Make sure to don't introduce explicit checking out and installing/running
+  # untrusted user code into this workflow!
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+
+  # To check labels on ghstack PRs.
+  # Note: as pull_request doesn't trigger on PRs targeting main,
+  # to test changes to the workflow itself one needs to create
+  # a PR that targets a gh/**/base branch.
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [gh/**/base]
+
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to check labels for'
+        required: true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}-${{ github.event_name == 'workflow_dispatch' }}
+  cancel-in-progress: true
+
+jobs:
+  check-labels:
+    permissions:
+      contents: read
+      pull-requests: write
+    name: Check labels
+    if: github.repository_owner == 'pytorch'
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+	with:
+	  fetch-depth: '0'
+
+      - uses: actions/setup-python@v4
+	with:
+	  python-version: '3.10'
+
+      - name: Check labels
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/check-labels.yml
```diff
@@ -0,0 +1,52 @@
...
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+
+  # To check labels on ghstack PRs.
+  # Note: as pull_request doesn't trigger on PRs targeting main,
+  # to test changes to the workflow itself one needs to create
+  # a PR that targets a gh/**/base branch.
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [gh/**/base]
```
guangy10, line 19: I think you may want to test is on regular PR (not created via ghstack) as well by removing this temporarily. We should see it run and catch.

.github/workflows/check-labels.yml
```diff
@@ -0,0 +1,52 @@
...
+  # untrusted user code into this workflow!
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+
+  # To check labels on ghstack PRs.
+  # Note: as pull_request doesn't trigger on PRs targeting main,
+  # to test changes to the workflow itself one needs to create
+  # a PR that targets a gh/**/base branch.
+  pull_request:
```
guangy10, line 17: Can you create this PR from codev by exporting it using ghstack so you can test on a gh/ branch?

dvorjackz: https://github.com/pytorch/executorch/pull/6930 for ghstack

.github/workflows/check-labels.yml
dvorjackz: Wouldn't we want to keep `main` here as well since? Otherwise it will only run when using `ghstack`

.github/workflows/check-labels.yml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  check-labels:
+    permissions:
+      contents: read
+      pull-requests: write
+    name: Check labels
+    if: github.repository_owner == 'pytorch'
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
```
huydhn, line 40: Let me double check the behavior of `actions/checkout` a bit to make sure that this is secure

huydhn (approved):

.github/workflows/check-labels.yml
```diff
@@ -0,0 +1,54 @@
...
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+
+  # To check labels on ghstack PRs.
+  # Note: as pull_request doesn't trigger on PRs targeting main,
+  # to test changes to the workflow itself one needs to create
+  # a PR that targets a gh/**/base branch.
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [gh/**/base, main]
```
huydhn, line 19: You cannot use main here as it poses a security risk. See https://fb.workplace.com/groups/pytorch.dev/permalink/1658227814755692.
```suggestion
    branches: [gh/**/base]
```

.github/workflows/check-labels.yml
huydhn: Nope, we need to remove main here because it's a security risk.  This workflow is granted write access, so it needs to get the `check_labels.py` script from main as part of `pull_request_target` trigger.  Basically, it will grant write access to random GitHub users (privileges escalation).  You see the ghstack part here because folks who are using ghstack already have write access to the repo.

huydhn: And don't worry about the check not running in your PR, you can land this and it will start running on all PRs after

kirklandsign:
> To add a label, you can comment to pytorchbot, for example
@pytorchbot label "topic: not user facing"


Does it work?

dvorjackz:
@kirklandsign were some issues with the comment deletion - please see https://github.com/pytorch/executorch/pull/6953

PR #6837, opened 2024-11-13
update sccache to latest version

wdvr:
### Summary
Update SCCache to the latest version

This is part of the effort to get rid of AWS IMDSv1 usage. Latest SCCache uses IMDSv2, which is what we migrate to here (similar to https://github.com/pytorch/pytorch/pull/121323).

### Test plan
Check if SCCache is used and auth works to the S3 bucket for builds

### DIFF
--- a/.ci/docker/common/install_cache.sh
+++ b/.ci/docker/common/install_cache.sh
@@ -34,10 +34,6 @@ install_ubuntu() {
   rm -rf sccache
   apt-get remove -y cargo rustc
   apt-get autoclean && apt-get clean
-
-  echo "Downloading old sccache binary from S3 repo for PCH builds"
-  curl --retry 3 https://s3.amazonaws.com/ossci-linux/sccache -o /opt/cache/bin/sccache-0.2.14a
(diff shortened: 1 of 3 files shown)


.ci/docker/common/install_cache.sh
```diff
@@ -12,6 +12,30 @@ set -ex
...
 source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
 
+install_ubuntu() {
+  echo "Preparing to build sccache from source"
+  apt-get update
+  # libssl-dev will not work as it is upgraded to libssl3 in Ubuntu-22.04.
+  # Instead use lib and headers from OpenSSL1.1 installed in `install_openssl.sh``
+  apt-get install -y cargo
+  echo "Checking out sccache repo"
+  if [ -n "$CUDA_VERSION" ]; then
```
huydhn, line 22: I think we can just use upstream https://github.com/mozilla/sccache to keep it simple here as there is no CUDA

huydhn (approved):
LGTM!  We can just use https://github.com/mozilla/sccache here as there is not going to be any CUDA, so we won't need pytorch-fork sccache

PR #6851, opened 2024-11-14
TOSA specification in Arm partitioner

per:
Adds information on which TOSA specification is being used for the partitioner.

### DIFF
--- a/backends/arm/operator_support/tosa_supported_operators.py
+++ b/backends/arm/operator_support/tosa_supported_operators.py
@@ -1,0 +1,128 @@
+# Copyright 2024 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-unsafe
+
+import operator
+
+import torch.fx as fx
+from executorch.backends.arm.tosa_specification import TosaSpecification
+from executorch.exir.dialects._ops import ops as exir_ops
+from torch.fx.passes.operator_support import OperatorSupportBase
+
+
+class SupportedTOSAOperatorCheck:
+    """
+    Supported OP for TOSA lowering
+    """
+
+    # Should be populated by subclass implementation
+    tosa_specs: list[TosaSpecification] = []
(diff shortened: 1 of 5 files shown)


per:
MacOS failure since commit 2f6d64f, not related to this change.

freddan80 (approved):

backends/arm/operator_support/tosa_supported_operators.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Checks if the fx.Node node is lowerable using the TOSA specification defined by tosa_spec.
+        To be implemented by subclasses targeting
+        """
+        raise NotImplementedError("NodeVisitor must be extended.")
+
+
+# container for all SupportedTosaOperatorCheck classes
+_tosa_spec_dicts: dict[TosaSpecification, dict[str, SupportedTOSAOperatorCheck]] = {
+    TosaSpecification.create_from_string("TOSA-0.80.0+BI"): {},
```
digantdesai, line 35: Nit: Add a URL as a comment on how to find the exact version spec

backends/arm/operator_support/tosa_supported_operators.py
```diff
@@ -0,0 +1,128 @@
...
+def get_registered_tosa_support_checks(
+    tosa_spec: TosaSpecification,
+) -> dict[str, SupportedTOSAOperatorCheck]:
+
+    if tosa_spec not in _tosa_spec_dicts:
+        raise RuntimeError
+
+    tosa_support_checks = {}
+    for target, tosa_check in _tosa_spec_dicts[tosa_spec].items():
+        tosa_support_checks[target] = tosa_check()
```
lucylq, line 61: I think this should be:
```
tosa_support_checks[target] = tosa_check()
```
--> 
```
tosa_support_checks[target] = tosa_check
```

lucylq: ah, wasn't this, but a type-check. resolved in https://github.com/pytorch/executorch/pull/6918

PR #6860, opened 2024-11-14
Add shim/README.md to explain what the shim/ tree is for

dbort:
### Summary
It's not obvious why this tree is here. Add a README explaining it.

### Test plan
Previewed in github.


larryliu0820 (approved):

dbort:
Jobs like `pull / test-models-linux (cmake, mv3, portable, linux.2xlarge, 90) / linux-job (pull_request)` are succeeding, so it looks like this file doesn't confuse buck2.

PR #6863, opened 2024-11-14
Use getops to pass flags and arguments

cccclai:
Summary: To make the test_llama.sh script more scalable, such that we can cover other args from the export_llama script like `pt2e` etc

Differential Revision: D65959919


kirklandsign:
Still needed? 😆

cccclai:
> Still needed? 😆

oh https://github.com/pytorch/executorch/pull/6870 was merged instead. Closing this one

PR #6869, opened 2024-11-14
[llama-mm] Add torch.cond to replace if condition in MHA

larryliu0820:
Summary:

In torchtune's MultiHeadAttention we have this logic:

If `y` is not None, calculate the values of `k` and `v` from y and update the KVCache.

Otherwise (if `y` is None), retrieve the value of `k` and `v` from KVCache.

This logic is not able to be handled by export world. Here I'm proposing a rewrite:

If `y` does not have all values equal to nan (not a number), calculate the values of `k` and `v` from `y` and update the KVCache.

Otherwise (if all of the values of `y` are nan), retrieve the value of `k` and `v` from KVCache.

This rewrite allows the module to satisfy the requirement of `torch.cond` and avoid specialization:
* The operands to `torch.cond` should have the same shape for the true branch and the false branch.

This means we will have to change this logic in torchtune:

```
        if encoder_input is not None:
            encoder_embed = self.encoder(**encoder_input)

        output = self.decoder(
            tokens=tokens,
            mask=mask,
            encoder_input=encoder_embed,
            encoder_mask=encoder_mask,
            input_pos=input_pos,
        )
```

To be:

```
        if encoder_input is not None:
            encoder_embed = self.encoder(**encoder_input)
        else:
            encoder_embed = torch.full_like(encoder_input, torch.nan)
        output = self.decoder(
            tokens=tokens,
            mask=mask,
            encoder_input=encoder_embed,
            encoder_mask=encoder_mask,
            input_pos=input_pos,
        )
```

Test Plan: Rely on unit tests

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/extension/llm/modules/attention.py
+++ b/extension/llm/modules/attention.py
@@ -246,7 +246,6 @@ class MultiHeadAttention(nn.Module):
         # x has shape [b, s_x, d]
         # y has shape [b, s_y, d]
         b, s_x, _ = x.shape
-        s_y = y.shape[1] if y is not None else 0
 
         # q has shape [b, s_x, num_heads * head_dim]
         q = self.q_proj(x)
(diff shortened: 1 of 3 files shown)


jackzhxng (approved):
Seems fine - so instead of passing in y is None we pass in a tensor full of zeroes

PR #6873, opened 2024-11-14
Update Xcode projects to use nightly builds.

shoumikhin:
Summary: Pinning to binaries build daily.

Differential Revision: D65974580


huydhn (approved):

huydhn:
Does this mean that this would need to be updated daily too when a new nightly build becomes available?

shoumikhin:
@huydhn not really, only when we or the customers need to upgrade the pinned version.

PR #6881, opened 2024-11-15
add simply ops to oss, update fuse simply callsites

zonglinpeng:
Summary: titled

Differential Revision: D65980636

### DIFF
--- a/backends/cadence/aot/compiler.py
+++ b/backends/cadence/aot/compiler.py
@@ -10,6 +10,7 @@ import logging
 from pathlib import Path
 from typing import Callable, cast, Optional
 
+import executorch.backends.cadence.aot.ops_registrations  # noqa
 import torch
 
 from executorch.backends.cadence.aot.passes import ReplaceSafeSoftmaxWithSoftmax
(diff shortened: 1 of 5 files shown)


backends/cadence/aot/compiler.py
```diff
@@ -196,7 +197,26 @@ def export_to_edge(
 # Export the model and lower it to an EdgeProgramManager (in edge IR), and
 # apply passes specific to Cadence DSP execution. Return both to print the
 # differences.
-def export_to_cadence_edge_executorch(
+def export_to_cadence(
```
mcremon-meta, line 200: why change the name here?

backends/cadence/aot/compiler.py
```diff
@@ -196,7 +197,26 @@ def export_to_edge(
...
+    model: torch.nn.Module,
+    inputs: tuple[object, ...],
+    dump_graphs: bool = False,
+    output_dir: Optional[str] = None,
+    opt_level: int = 1,
+) -> EdgeProgramManager:
+    edge_prog_manager = export_to_edge(model, inputs)
+    cadence_passes = get_cadence_passes(opt_level)
+
+    # Run a couple required passes for quant/dequant ops
```
mcremon-meta, line 210: probably should update that comment

mcremon-meta (approved):

PR #6888, opened 2024-11-15
Arm backend: Test TOSA, Ethos-U55 and Ethos-U85 on github

zingo:
This will run some more models in the github test flow and also enable some unit tests to use Corston3x0 FVP.

This structure up the Arm backend testing into separate runable scripts in the same structure as other backends.


zingo:
Hi @digantdesai  this changes files outside Arm folders and need Meta review

freddan80 (approved):

freddan80:
lgtm, but @digantdesai should have a look as well

digantdesai:
Love it. One consideration though, we should try to set this scripts such a way that not just a CI but a human (or a tutorial doc) can use this directly without any changes. 

This way there is only one source of truth.

Stamping it to unblock you.

.github/workflows/pull.yml
```diff
@@ -354,13 +354,11 @@ jobs:
...
         # Install Arm dependencies
-        install_arm
-
-        # Run pytest with coverage
-        pytest -c /dev/null -v -n auto --cov=./ --cov-report=xml backends/arm/test
-
+        .ci/scripts/setup-arm_baremetal.sh
+        
+        # Run pytest
+        .ci/scripts/test_arm_baremetal.sh test_pytest
```
digantdesai, line 361: Do you know the time increase?

.github/workflows/trunk.yml
```diff
@@ -172,10 +173,10 @@ jobs:
         source .ci/scripts/utils.sh
         install_executorch
 
-        install_arm
+        .ci/scripts/setup-arm_baremetal.sh
 
-        # Run arm unit tests
-        pytest -c /dev/null -v -n auto --cov=./ --cov-report=xml backends/arm/test
+        # Run arm pytest tests with Corstone3x0 FVP testing
+        .ci/scripts/test_arm_baremetal.sh test_pytest_ethosu_fvp
```
digantdesai, line 179: I guess firing fvp a bunch of time is more time consuming compared to running a single large-ish model like mv2?

digantdesai (approved):

.github/workflows/trunk.yml
zingo: Maybe, we have a patch to make it run FVP faster ongoing also.

zingo: But the webare also adding more FVP testing. Lets see how it pans out a bit over time.

zingo:
The plan was to make it easy to run. We can describe them in the readme, do you want the scripts to move in that case?

.github/workflows/pull.yml
zingo: But that was before I measured the job diff and spotted that it might have been ok

malfet, line 361: Script name sounds very confusing to me, considering that this workflow is running on x86 machine

zingo: I just tried to follow the other names in that folder and are happy for new suggestions.

If something like this what you are thinking about?
- .ci/scripts/setup-arm_baremetal_build.sh
- .ci/scripts/test_arm_baremetal_delegate.sh

zingo:
> I don't have a strong preference on where the script lives, as long as a user can just run it on a shell without hacking it up would be a nice-to-have, esp for Tutorial where we can point and know it will work because it is tested by CI.

Could make since (and I kind of like  to) move at least the test part (setup could be CI/user specific, but it could also follow) to something like backends/arm/test

I look into that after some other stuff, depending on your wish of getting this is fast or not I can update this patch (a bit slower) or take it as a later patch if this get merged before.

.ci/scripts/setup-arm-baremetal-tools.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright 2024 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# NB: This function could be used to install Arm dependencies
+# Setup arm example environment (including TOSA tools)
+git config --global user.email "github_executorch@arm.com"
```
swolchok, line 9: why do we need to do this? I tried to run this script to repro a build issue and unknowingly set the name and email incorrectly on several github commits.

zingo: Ouch sorry about that.
It just followed along when moved from install_arm() in
.ci/scripts/utils.sh (in the same patch)

I'm not sure if it is needed when running in the CI nowdays, some time back the examples/arm/setup.sh did some git cloning and I assume it might have been needed then.  Ot it could be related to the patch scripts that was used for vela patches.

PR #6896, opened 2024-11-15
Use linux.4xlarge.memory instead of linux.12xlarge

huydhn:
`linux.4xlarge.memory` has 128 GB of memory with 16 CPU cores while `linux.12xlarge`, a more expensive runner, has only 96 GB of memory on a whopping 48 CPU cores.

### Testing

Example runs for:

* `dl3` https://github.com/pytorch/executorch/actions/runs/11863238941/job/33064335597?pr=6896 (20m vs 19m in trunk) 
* `emformer_predict` https://github.com/pytorch/executorch/actions/runs/11863238941/job/33064336524?pr=6896 (1h30m vs 1h20m in trunk)


kirklandsign (approved):

guangy10 (approved):
@huydhn Before merging can you link the comparison of job execution time between using `linux.4xlarge.memory` and `linux.12xlarge`? Would like to understand the actual trade-off we made, ideally we should expect it to be tiny

.ci/scripts/gather_test_models.py
```diff
@@ -20,16 +20,16 @@
 CUSTOM_RUNNERS = {
     "linux": {
         # This one runs OOM on smaller runner, the root cause is unclear (T163016365)
-        "w2l": "linux.12xlarge",
-        "ic4": "linux.12xlarge",
-        "resnet50": "linux.12xlarge",
-        "llava": "linux.12xlarge",
-        "llama3_2_vision_encoder": "linux.12xlarge",
-        # "llama3_2_text_decoder": "linux.12xlarge",  # TODO: re-enable test when Huy's change is in / model gets smaller.
```
guangy10, line 28: Can you attach the job link to this model since we re-enable it ?

kirklandsign (approved):

.ci/scripts/gather_test_models.py
huydhn: Good catch, I think I will comment it out and leave it for latter.  It doesn't OOM but take forever to export (close to 6 hours so far).  I don't have much context, so probably need help from @dvorjackz to figure this one out

Issue #4669, opened 2024-08-12
torch.nn.InstanceNorm2d fails at runtime 


wojtke:
### 🐛 Describe the bug

I found that `torch.nn.InstanceNorm2d` layer, which by default does not track running mean and var, is failing at runtime. It successfully goes through the lowering process with no errors.

I see that the error is caused by some assert assuming that `aten__native_batch_norm_legit_no_stats` is in training mode, even if it is not. The same goes for `torch.nn.BatchNorm2d` with `track_running_stats=False`.

Minimal case to reproduce:
```python
import torch
from executorch.exir import to_edge

class Model(torch.nn.Module):
  def __init__(self):
    super(Model, self).__init__()
    self.gn = torch.nn.InstanceNorm2d(3)

  def forward(self, x: torch.Tensor):
      return self.gn(x)

model = Model().eval()
sample_inputs = (
   torch.rand(1, 3, 3, 3),
)

aten_dialect = torch.export.export(model, sample_inputs)
edge_program = to_edge(aten_dialect)
print(edge_program.exported_program())
executorch_program = edge_program.to_executorch()

with open("instance_norm.pte", "wb") as file:
    executorch_program.write_to_file(file)
 ```
 The exported program at edge dialect step:
 ```
 ExportedProgram:
    class GraphModule(torch.nn.Module):
        def forward(self, x: "f32[1, 3, 3, 3]"):
             # File: /Users/woj/executorch_main/executorch/test.py:11 in forward, code: return self.gn(x)
            aten_view_copy_default: "f32[1, 3, 3, 3]" = executorch_exir_dialects_edge__ops_aten_view_copy_default(x, [1, 3, 3, 3]);  x = None
            aten__native_batch_norm_legit_no_stats = executorch_exir_dialects_edge__ops_aten__native_batch_norm_legit_no_stats(aten_view_copy_default, None, None, True, 0.1, 1e-05);  aten_view_copy_default = None
            getitem: "f32[1, 3, 3, 3]" = aten__native_batch_norm_legit_no_stats[0];  aten__native_batch_norm_legit_no_stats = None
            aten_view_copy_default_1: "f32[1, 3, 3, 3]" = executorch_exir_dialects_edge__ops_aten_view_copy_default(getitem, [1, 3, 3, 3]);  getitem = None
            return (aten_view_copy_default_1,)
            
[...]

Issue #6632, opened 2024-11-04
aten::_native_batch_norm_legit.no_stats_out Check failed (training == false): Portable kernels only support inference mode!


kiyotaka-xiaoli:
### 🐛 Describe the bug

it's happened when i run executor_runner with my pte model.
it seems failed on `torch.nn.BatchNorm2d` op.
here is the tracks:
```
./cmake-out/executor_runner --model_path 'model.pte'
I 00:00:00.007307 executorch:executor_runner.cpp:82] Model file model.pte is loaded.
I 00:00:00.007337 executorch:executor_runner.cpp:91] Using method forward
I 00:00:00.007341 executorch:executor_runner.cpp:138] Setting up planned buffer 0, size 81246208.
I 00:00:00.061899 executorch:executor_runner.cpp:161] Method loaded.
I 00:00:00.062504 executorch:executor_runner.cpp:171] Inputs prepared.
E 00:00:09.373685 executorch:op_native_batch_norm.cpp:200] Check failed (training == false): Portable kernels only support inference mode!
E 00:00:09.373724 executorch:method.cpp:1045] KernelCall failed at instruction 0:6733 in operator aten::_native_batch_norm_legit.no_stats_out: 0x12
E 00:00:09.373727 executorch:method.cpp:1051] arg 0 with type id 1
E 00:00:09.373729 executorch:method.cpp:1051] arg 1 with type id 1
E 00:00:09.373730 executorch:method.cpp:1051] arg 2 with type id 1
E 00:00:09.373732 executorch:method.cpp:1051] arg 3 with type id 5
E 00:00:09.373733 executorch:method.cpp:1051] arg 4 with type id 3
E 00:00:09.373734 executorch:method.cpp:1051] arg 5 with type id 3
E 00:00:09.373735 executorch:method.cpp:1051] arg 6 with type id 1
E 00:00:09.373737 executorch:method.cpp:1051] arg 7 with type id 1
E 00:00:09.373738 executorch:method.cpp:1051] arg 8 with type id 1
E 00:00:09.373746 executorch:method.cpp:1051] arg 9 with type id 9
F 00:00:09.373748 executorch:executor_runner.cpp:179] In function main(), assert failed (status == Error::Ok): Execution of method forward failed with status 0x12
Aborted (core dumped)
```

i am sure that i have called model.eval() functions before export model.
here is my code for exporting model:

```
model = Model().eval()
sample = (torch.rand(1, 8, 512, 81),)
[...]

PR #6929, opened 2024-11-18
Fix _native_batch_norm_legit_no_stats_out

manuelcandales:
Differential Revision: D66104138


manuelcandales:
Resolves #4669 and #6632

swolchok (approved):

PR #6933, opened 2024-11-18
Add benchmark for training training

davidlin54:
Move MethodHolder to public to be reused in internal benchmarking

Reviewed By: JacobSzwejbka

Differential Revision: D65905482

### DIFF
--- a/extension/module/module.h
+++ b/extension/module/module.h
@@ -38,6 +38,15 @@ class Module {
     MmapUseMlockIgnoreErrors,
   };
 
+  struct MethodHolder {
+    std::vector<std::vector<uint8_t>> planned_buffers;
+    std::vector<runtime::Span<uint8_t>> planned_spans;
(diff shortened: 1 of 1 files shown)


extension/module/module.h
```diff
@@ -38,6 +38,15 @@ class Module {
     MmapUseMlockIgnoreErrors,
   };
 
+  struct MethodHolder {
```
JacobSzwejbka, line 41: can you leave it private and just make a friend class for benchmarking?

davidlin54: hmm I don't want to access any data in Module (and don't even have a reference to Module, since benchmarking uses program/method), just want to use MethodHolder as a helper structure. I can also just copy and paste it into the benchmark runner code if you prefer this to be private

shoumikhin:
@JacobSzwejbka, @Tarun292, @Gasoonjia can you please take a look?

PR #6938, opened 2024-11-18
Fix error message.

hsharma35:
Summary:
Remove extra whitespace in from error message.

```
Cannot override       the -> Cannot override the
```

Differential Revision: D66105238


cccclai (approved):
Thanks for the fix!

PR #6959, opened 2024-11-19
Arm backend: Add devtools support to example

zingo:
New flags on run.sh
--etdump
  Build in etdump and profiling, the etdump base64 coded and put in  the log
--debug_build
  Build debug instead of release
--extra_build_flags
  Extra flags to pass to cmake this makes it for example possible to
  override the allocator pool size or other build time cmake flags.

The devtools build has been updated so FLATCC_EXECUTABLE can be used to point out the executable as seem to be intended to be possible but was not working.


zingo:
@digantdesai This changes files outside Arm folders and need a review from you, Thanks for your time!

backends/arm/runtime/ArmBackendEthosU.cpp
```diff
@@ -15,6 +15,17 @@
 
 #include <ethosu_driver.h>
 
+#if defined(ET_EVENT_TRACER_ENABLED)
+#include <executorch/runtime/core/event_tracer.h>
+#include <executorch/runtime/core/event_tracer_hooks.h>
+using executorch::runtime::internal::EventTracerProfileOpScope;
+#define EXECUTORCH_INTERNAL_PROF(NAME)              \
+  EventTracerProfileOpScope event_tracer_op_scope = \
+      EventTracerProfileOpScope(event_tracer, NAME);
```
digantdesai, line 24: Yeah AFAICT this is not meant to be used from outside, as you were suspecting.

So I would try using stuff from `executorch/runtime/core/event_tracer.h` like `EventTracer::[start|stop]_profiling`?

zingo: Yes that works, PR will be updated soon.

zingo: Fixed!

zingo:
Error seem unrealted to this patch e.g.

unittest / linux / linux-job] - https://github.com/pytorch/executorch/actions/runs/12255701820/job/34189395827?pr=6959#logs

++ curl --silent --show-error --location --fail --retry 3 --output /tmp/vulkansdk.tar.gz https://sdk.lunarg.com/sdk/download/1.2.198.1/linux/vulkansdk-linux-x86_64-1.2.198.1.tar.gz
curl: (22) The requested URL returned error: 404

freddan80 (approved):

PR #6973, opened 2024-11-19
Fix ARM submodule URLs

swolchok:
Looks like review.mlplatform.org is no longer hosting these things and these are the new URLs.

Test Plan: git submodule update --init (did nothing in my existing checkout)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "backends/arm/third-party/ethos-u-core-driver"]
 	path = backends/arm/third-party/ethos-u-core-driver
-	url = https://review.mlplatform.org/ml/ethos-u/ethos-u-core-driver
+	url = https://git.mlplatform.org/ml/ethos-u/ethos-u-core-driver.git/
 [submodule "backends/arm/third-party/serialization_lib"]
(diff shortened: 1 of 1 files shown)


larryliu0820 (approved):

digantdesai:
FYI @robell, @freddan80

robell:
Hi @digantdesai and @swolchok - Sorry about all of the issues relating to this. 

review.* and git.* are hosted on the same machine, and typically/by design the review.mlplatform.org software is more reliable for cloning the repositories.

There were capacity issues on the machine due to a third party github flow which are now resolved by adding capacity to the VM, so it's recommended to move back to review.mlplatform.org. This should now be reliably available again. We may prefer to leave this alone now if it's functioning, but just be aware of the option.

In addition, we'll be looking to mirror these dependencies to more scalable/reliable hosting and move the submodule fetches to that host. I'll get a PR up once those mirrors are available, I'm just clarifying how soon that can happen.

swolchok:
> recommended to move back to review.mlplatform.org

I sent https://github.com/pytorch/executorch/pull/6987, but review. URLs appear to still be down given CI failures on that PR.

PR #6977, opened 2024-11-20
Conv Combos with Simulator

skrtskrtfb:
Summary: Conv Combos that fail in simulator

Differential Revision: D66217464


AdrianLundell:
Thank you for raising this, we are looking into it!

freddan80:
@skrtskrtfb was this PR intended for merging or just to provide a reproducer? If the latter, can it be closed?

PR #6849, opened 2024-11-14
Qualcomm AI Engine Direct - Optimization in static llama

shewu-quic:
summary:
- Fuse rms norm
- Improve performance of div op
- Fixed 16a8w annotation for matmul op

PR #6978, opened 2024-11-20
Fix custom annotation

cccclai:
Summary: Looks like it's added in  https://github.com/pytorch/executorch/pull/6849, maybe it was using the old api for the default 8bit quantization

Differential Revision: D66219251


shewu-quic:
Thanks for your fix. I am also preparing the PR to fix it.

swolchok (approved):
was just looking into unbreaking lintrunner, thanks!

PR #6992, opened 2024-11-20
Add HardTanh to RemovePermutesAroundElementwiseOps

zonglinpeng:
Summary:
This diff adds `hardtanh` to the list of elementwise ops that we consider for permute elimination.

- Fixing logging issue in summary dump that results in some delegate size metrics being reported incorrectly.

Reviewed By: zonglinpeng

Differential Revision: D66187338


zonglinpeng:
dup

PR #6997, opened 2024-11-20
[Executorch] Add quantized kv cache to oss ci

kimishpatel:
* __->__ #6997
* #7038
* #6996

Fixes to make sure quantized kv cache works in oss

Differential Revision: [D66269487](https://our.internmc.facebook.com/intern/diff/D66269487/)


examples/models/llama/source_transformation/quantized_kv_cache.py
```diff
@@ -10,9 +10,36 @@
...
+        glob.glob(
+            f"{executorch_package_path}/**/libquantized_ops_aot_lib.*", recursive=True
+        )
+    )
+    assert len(libs) == 1, f"Expected 1 library but got {len(libs)}"
+    logging.info(f"Loading custom ops library: {libs[0]}")
+    torch.ops.load_library(libs[0])
+    op = torch.ops.quantized_decomposed.quantize_per_token.out
+    assert op is not None
+
```
larryliu0820, line 42: Is this needed? I feel like we can add this to `executorch/kernels/quantized/__init__.py`

larryliu0820 (approved):

examples/models/llama/source_transformation/quantized_kv_cache.py
kimishpatel: Oh I didnt realize that this was already there in quantized/__init__.py. I can try to patch that one and see if it works. Althought it does feel quite weird, TBH, to have executorch/kernels/quantized to be exposed as python package. But it does contain AOT component, so seems fair.

PR #7003, opened 2024-11-21
Qualcomm AI Engine Direct - Optimize memory usage at runtime

shewu-quic:
Qnn backend doesn't need processed data after qnn_context_create_from_binary.


shewu-quic:
Hi @cccclai,

The PR is to fix the issue of memory usage.
Could you please help take a look?

Thanks,
Hutton

cccclai (approved):
Thanks! Maybe it also helps to resolve the RAM issue we ran into before for the 8b model on 16GB? I understand you already fixed it in an alternative way.

shewu-quic:
I think spill fill buffer is still necessary for 8b due to the memory usage in HTP.
This change only reduced PSS.

PR #7017, opened 2024-11-21
Trigger wheel builds when modifying setup.py or pyproject.toml

dbort:
These two files are the core of the wheel configs, so it makes sense to try building wheels when they change.

Sort the path entries now that the list is getting longer.


swolchok:
looks like you need to rebase; test-llama-runner-linux is green on HUD

swolchok (approved):

PR #7026, opened 2024-11-22
Qaulcomm AI Engine Direct - Fix quantization annotation for per channel quant

shewu-quic:
summary:

- Fix the 8a8w config in custom annotation
- Enable to set act observer and symmetric argument for per channel quant
- Remove unuse custom annotation in llama.py


shewu-quic:
Hi @cccclai,

The PR is to enable to set the act_observer for per-channel quant which to align the observer with other nodes.
Could you help to take a look?

Thanks,
Hutton

cccclai:
@navsud can you help with verifying

navsud (approved):
LGTM

Thanks for the update... Without this, the act observers were not consistent across the graph, especially when we want to use MinMaxObserver for 16-bit activations.

cccclai (approved):
Thanks!

PR #7027, opened 2024-11-22
Update getting-started-setup.md

kirklandsign:
### Summary
Should cd to root before running cmake
### Test plan
Run these commands

### DIFF
--- a/docs/source/getting-started-setup.md
+++ b/docs/source/getting-started-setup.md
@@ -196,7 +196,7 @@ The ExecuTorch repo uses CMake to build its C++ code. Here, we'll configure it t
   ```bash
   # Clean and configure the CMake build system. Compiled programs will
   # appear in the executorch/cmake-out directory we create here.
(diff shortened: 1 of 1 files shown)


docs/source/getting-started-setup.md
```diff
@@ -196,7 +196,7 @@ The ExecuTorch repo uses CMake to build its C++ code. Here, we'll configure it t
   ```bash
   # Clean and configure the CMake build system. Compiled programs will
   # appear in the executorch/cmake-out directory we create here.
-  (rm -rf cmake-out && mkdir cmake-out && cd cmake-out && cmake ..)
+  (rm -rf cmake-out && mkdir cmake-out && cd cmake-out && cmake .. && cd ..)
```
dbort, line 199: This isn't necessary since it's in a subshell (the parentheses). That's why I put this in parens, because it won't affect the current directory of the parent shell

PR #7038, opened 2024-11-22
[Executorch][custom ops] Change lib loading logic to account for package dir

kimishpatel:
* #6997
* __->__ #7038
* #6996

Just looking at the location of the source file. In this case custom_ops.py,
can, and does, yield to wrong location depending on where you import custom_ops
from.

If you are importing custom_ops from another source file inside extension
folder, e.g. builder.py that is in extensions/llm/export, then, I think,
custom_ops gets resolved to the one installed in site-packages or pip package.

But if this is imported from say
examples/models/llama/source_transformations/quantized_kv_cache.py (Like in the
in next PR), then it seems to resolve to the source location. In one of the CI
this is /pytorch/executorch.

Now depending on which directory your filepath resolves to, you will search for
lib in that. This of course does not work when filepath resolves to source
location.

This PR changes that to resolve to package location.

Differential Revision: [D66385480](https://our.internmc.facebook.com/intern/diff/D66385480/)


larryliu0820 (approved):
I think we should avoid importing the file under `executorch/` repo but always import the one from `site-packages`. We should move our code towards that direction in later PRs.

PR #7040, opened 2024-11-23
Use c10 version of half/bfloat16 in executorch

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7042
* #7041
* #7546
* __->__ #7040

Accomplished by importing relevant files from c10 into
executorch/runtime/core/portable_type/c10, and then using `using` in
the top-level ExecuTorch headers. This approach should keep the
ExecuTorch build hermetic for embedded use cases. In the future, we
should add a CI job to ensure the c10 files stay identical to the
PyTorch ones.

Differential Revision: [D66106969](https://our.internmc.facebook.com/intern/diff/D66106969/)

PR #7041, opened 2024-11-23
Reuse GELU implementation from PyTorch core

swolchok:
* #7042
* __->__ #7041
* #7546
* #7040

kernels/optimized doesn't need to support embedded systems, so it can just take a header-only dep on PyTorch.

Note that, because we will pick up Sleef internally and ignore it
externally thanks to ATen vec, this PR gets to enable optimized GELU in OSS.

Testing: CI to make sure this doesn't break mobile build modes; happy to take advice on anything not currently covered that might break.

Differential Revision: [D66335522](https://our.internmc.facebook.com/intern/diff/D66335522/)


swolchok:
test-arm-reference-delegation failure is a known flake
test-llava-runner-linux failure might be real? CMake is failing to find Torch

kimishpatel (approved):

PR #7042, opened 2024-11-23
Remove ExecuTorch copy of Vectorized

swolchok:
* __->__ #7042

All uses are outside ExecuTorch core, so we can just use ATen Vectorized.

Differential Revision: [D66396016](https://our.internmc.facebook.com/intern/diff/D66396016/)


kimishpatel (approved):

kernels/optimized/vec/intrinsics.h
```diff
@@ -1,51 +0,0 @@
-/*
```
swolchok, line 1: hey @dbort does our bc policy require me to undelete all these files? saw you mentioned it recently but I don't know the exact text of the policy

dbort: Policy is at https://pytorch.org/executorch/main/api-life-cycle.html

The goal of the policy is to avoid breaking user code that worked with the previous release. Removing a header could break code that currently works.

Looking at the files that this PR changes, I'm not totally sure if they really qualify as part of our public API; not sure if the headers under kernels/... are meant to be used outside the repo or if they're just for the code around them.

kimishpatel: It could be if they are used to author custom ops out of the repo, but 1) I havent seen any examples in the wild and 2) I dont know if we package this header in our pip package (like pytorch) OR make it public somehow

swolchok: > haven't seen any examples in the wild

[I searched GitHub and found nothing after pruning paths within ExecuTorch and forks.](https://github.com/search?q=executorch%2Fkernels%2Foptimized%2Fvec+NOT+repo%3Apytorch%2Fexecutorch++NOT+path%3Aexecutorch+NOT+repo%3Ajustinchuby%2Fexecutorch+NOT+repo%3Akirklandsign%2Fexecutorch+NOT+repo%3ADenisVieriu97%2Fexecutorch_fork+NOT+repo%3AGregoryComer%2Fexecutorch+NOT+repo%3ACodeLinaro%2Fexecutorch+NOT+repo%3Acopyrightly%2Fexecutorch+NOT+path%3Akernels%2Foptimized%2Fcpu+NOT+path%3Aextension%2Fllm%2Fcustom_ops+NOT+path%3Akernels%2Foptimized%2Fvec+NOT+path%3Aexamples%2Fmodels+NOT+path%3Akernels%2Foptimized%2Ftest&type=code)

swolchok:
This will require https://github.com/pytorch/pytorch/pull/152366 to be included in a pin bump before landing

swolchok:
llava-runner-linux result is slightly different but wouldn't call it functionally incorrect. rerunning failed jobs but we may need to discuss whether this is landable. https://github.com/pytorch/executorch/actions/runs/15572676327/job/43851619080?pr=7042

swolchok:
yep, need to update test-llava-runner-linux output https://github.com/pytorch/executorch/actions/runs/15572676327/job/43855070881?pr=7042

swolchok:
unittest-buck failures are because we can't use PyTorch in OSS buck. short-term we need to take the affected ops out of the set we test; long-term we need to fix being unable to use PyTorch in OSS buck.

swolchok:
tests look good. failures on android / run-emulator, test-llama-runner-qnn-linux, and unittest-nxp-neutron appear to be unrelated.

PR #7043, opened 2024-11-23
Simplify cmake test invocation

kirklandsign:
### Summary
We can build the test while building the lib, just in one pass.

After this, developers can run tests while they develop. Both can be built at the same time. For developers, when they write their code, it's easy for them to test out. Just add `-DEXECUTORCH_BUILD_TESTS=ON` to their cmake config. The test target will be built automatically in cmake-out and they can run it.

NOTE: examples/models/llama/tokenizer/test and extension/llm/tokenizer/test can't use this right now. They don't have a good triggering path, and can't be triggered from root either. Use the old trigger path for them.

#### Method 1: script for all tests
```
sh test/run_oss_cpp_tests.sh
```
#### Method 2: go to cmake-out and test
```
cd cmake-out
ctest
```

#### Method 3:  test specific target
For example, we want to test kernel_resolution_test in runtime/core
```
cmake -DEXECUTORCH_BUILD_TESTS=ON . -Bcmake-out
cmake --build cmake-out
cd cmake-out
ctest -R kernel_resolution_test
```


CMakeLists.txt
```diff
@@ -854,5 +858,26 @@ if(EXECUTORCH_BUILD_VULKAN)
...
+  add_subdirectory(kernels/prim_ops/test)
+  add_subdirectory(kernels/test)
+  add_subdirectory(runtime/core/exec_aten/testing_util/test)
+  add_subdirectory(runtime/core/exec_aten/util/test)
+  add_subdirectory(runtime/core/portable_type/test)
+  add_subdirectory(runtime/core/test)
+  add_subdirectory(runtime/executor/test)
+  add_subdirectory(runtime/kernel/test)
+  add_subdirectory(runtime/platform/test)
+  add_subdirectory(test/utils)
```
larryliu0820, line 879: Can we move these into test.cmake

CMakeLists.txt
```diff
@@ -610,7 +614,7 @@ cmake_dependent_option(
 )
 
 # Add googletest if any test targets should be built
-if(EXECUTORCH_BUILD_GTESTS)
+if(BUILD_TESTING)
```
larryliu0820, line 617: Why not `EXECUTORCH_BUILD_TESTS`?

kirklandsign: `BUILD_TESTING` is from cmake system to recognize whether test config is on or not. It will be enabled when we include ctest (see https://github.com/pytorch/executorch/pull/7043/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR318-R320)

larryliu0820 (approved):

PR #7045, opened 2024-11-23
Add quantize_per_channel and dequantize_per_channel to q_dq_ops target

tarun292:
Summary: These ops are needed by the Semantic flow model and also are compilable on Windows.

Differential Revision: D66400800


iseeyuan (approved):
LGTM.

PR #7048, opened 2024-11-23
Fix test with resources

larryliu0820:
Summary: Fix test failure due to resources not handled correctly by ios tests.

Differential Revision: D66392647


swolchok (approved):
accepted with minor revision; see internal comments

PR #7049, opened 2024-11-23
[Executorch][CI] Fix qnn runner ci job scripts

kimishpatel:
* #6997
* __->__ #7049
* #7038
* #6996
* #6914
* #5715
* #5670

QNN backend's AOT artifacts were being copied in the source directory instead of
them being installed, as build artifacts, in appropriate package directory.

As a result we had to use PYTHON_PATH=".." in test_llama.sh CI script when
running CI for qnn. This results in executorch having two locations as where
the package is installed. This PR fixes that.

Differential Revision: [D66406274](https://our.internmc.facebook.com/intern/diff/D66406274/)

### DIFF
--- a/.ci/scripts/build-qnn-sdk.sh
+++ b/.ci/scripts/build-qnn-sdk.sh
@@ -12,7 +12,12 @@ build_qnn_backend() {
   echo "Start building qnn backend."
   export ANDROID_NDK_ROOT=/opt/ndk
   export QNN_SDK_ROOT=/tmp/qnn/2.25.0.240728
-  export EXECUTORCH_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../.." && pwd)"
+  export EXECUTORCH_ROOT="$(python -c 'import executorch; print(executorch.__path__[0])')"
+  if [ "$EXECUTORCH_ROOT" == "" ]; then
+    echo "Failed to find where executorch package is installed."
+    echo "import executorch failed"
+    exit -1
+  fi
 
   bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release
 }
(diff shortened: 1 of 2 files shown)


kimishpatel:
I am gonna have to abandon this because imports of model in qnn backend think that examples are part of the pip package of executorch. And to get around that we need to do PYTHON_PATH="EXECUTORCH_ROOT". I really dont like it, but dont have bandwidth to figure out better solution. So I will work around it for the purpose of this stack. cc: @cccclai @larryliu0820

cccclai:
@kimishpatel hmm how does this stack affect qnn CI job?

kimishpatel:
> @kimishpatel hmm how does this stack affect qnn CI job?

Not directly, but now we are importing quantized_kv_cache unconditionally from export_llama_lib. Because qnn CI stack has 1) PYTHON_PATH=.. which affects python imports and confuses the location of where executorch package is installed and 2) various models in examples/backends/qualcomm treat examples folder (as in `import executorch.example.models..`) as part of executorch pip package, but we pip install of executorch does not include those, so we have to reply on PYTHON_PATH.

2 in particularly is problematic because I see that pattern in a few places but it is not clear if examples is supposed be part of pip package. @larryliu0820 should confirm

larryliu0820:
> > @kimishpatel hmm how does this stack affect qnn CI job?
> 
> Not directly, but now we are importing quantized_kv_cache unconditionally from export_llama_lib. Because qnn CI stack has 1) PYTHON_PATH=.. which affects python imports and confuses the location of where executorch package is installed and 2) various models in examples/backends/qualcomm treat examples folder (as in `import executorch.example.models..`) as part of executorch pip package, but we pip install of executorch does not include those, so we have to reply on PYTHON_PATH.
> 
> 2 in particularly is problematic because I see that pattern in a few places but it is not clear if examples is supposed be part of pip package. @larryliu0820 should confirm

Ideally we shouldn’t include examples in pip package. Anything reusable should go into extension.Ilm and should not tied to any specific model. For example I’m aware a lot of code useful for llama are still in examples.models.llama. They should really go inside extension.llm for reusability

kimishpatel:
> llama

There much more widepsread use of pattern `from executorch.examples....` failures in this particular case are from `from executorch.examples.qualcomm` and were resolved by use of `PYTHON_PATH` in the past. But the way it was setup was causing executorch package directory to be listed as both source as well as what is in site-packages

cccclai:
Seems stale

PR #7054, opened 2024-11-25
Arm backend: Fix bug in ScalarsToAttributePass

oscarandersson8218:
### Summary
The pass should not modify the scalar argument if output is non-float. Also makes CastInt64ToInt32Pass more verbose and robust.


oscarandersson8218:
CI failures seem unrelated to this change.

freddan80 (approved):

PR #7055, opened 2024-11-25
Fix test-llama-runner-qnn-linux tests

davidlin54:
fix test-llama-runner-qnn-linux (fp32, qnn)
- add default value
- change pull.yml to reflect same changes in trunk.yml


swolchok (approved):
seems fine? if failures are unrelated, ship it

PR #7067, opened 2024-11-25
[ET-VK] Deserialize VkGraph in ET-VK

yipjustin:
* (to be filled)

Add logic to deserialize a VkGraph blob back python object. This allows us to get a implement debugging / visualization directly on the vulkan-exported program.  Still extra works need to be done: From the entire bundle, need to extract the specific vulkan delegate first.

Differential Revision: [D66443780](https://our.internmc.facebook.com/intern/diff/D66443780/)


swolchok:
diff was committed, closing

PR #7077, opened 2024-11-26
pip install reference_model and use pybind

Erik-Lundell:
The reference model is pip installed in setup.sh.
Also install vela similarly.
Since the installation contains serialization_lib, we don't have to include it as a package
in Executorch's setup.py. The serialization_lib
is still needed as a submodule in the arm backend
to find the tosa.fbs for deserialization.

Change-Id: I24fff6c00a3961444de5d878ab169d5ba4c9156d

### DIFF
--- a/backends/arm/test/common.py
+++ b/backends/arm/test/common.py
@@ -76,7 +76,7 @@ def get_tosa_compile_spec_unbuilt(
     if not custom_path:
         custom_path = maybe_get_tosa_collate_path()
 
-    if custom_path is not None and not os.path.exists(custom_path):
+    if custom_path is not None:
         os.makedirs(custom_path, exist_ok=True)
     compile_spec_builder = (
         ArmCompileSpecBuilder()
--- a/backends/arm/test/common.py
+++ b/backends/arm/test/common.py
@@ -74,19 +74,15 @@ def get_tosa_compile_spec_unbuilt(
     the compile spec before calling .build() to finalize it.
     """
     if not custom_path:
-        intermediate_path = maybe_get_tosa_collate_path() or tempfile.mkdtemp(
-            prefix="arm_tosa_"
-        )
-    else:
-        intermediate_path = custom_path
+        custom_path = maybe_get_tosa_collate_path()
 
-    if not os.path.exists(intermediate_path):
-        os.makedirs(intermediate_path, exist_ok=True)
+    if custom_path is not None and not os.path.exists(custom_path):
+        os.makedirs(custom_path, exist_ok=True)
     compile_spec_builder = (
         ArmCompileSpecBuilder()
         .tosa_compile_spec(tosa_version)
         .set_permute_memory_format(permute_memory_to_nhwc)
-        .dump_intermediate_artifacts_to(intermediate_path)
+        .dump_intermediate_artifacts_to(custom_path)
     )
 
     return compile_spec_builder
--- a/examples/arm/setup.sh
+++ b/examples/arm/setup.sh
@@ -88,7 +88,7 @@ ethos_u_base_rev="24.08"
 
 # tosa reference model
 tosa_reference_model_url="https://review.mlplatform.org/tosa/reference_model"
-tosa_reference_model_rev="f9ea4ab7da19318fe36b1c34d68a3e40fd6e56c5"
+tosa_reference_model_rev="c5570b79e90c3a36ab8c4ddb8ee3fbc2cd3f7c38"
 
 # vela
 vela_repo_url="https://review.mlplatform.org/ml/ethos-u/ethos-u-vela"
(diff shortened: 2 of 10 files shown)


freddan80 (approved):

setup.py
```diff
@@ -706,10 +706,6 @@ def get_ext_modules() -> List[Extension]:
         "executorch/devtools/bundled_program": "devtools/bundled_program",
         "executorch/runtime": "runtime",
         "executorch/util": "util",
-        # Note: This will install a top-level module called "serializer",
-        # which seems too generic and might conflict with other pip packages.
-        "serializer": "backends/arm/third-party/serialization_lib/python/serializer",
```
digantdesai, line 711: This is OK. I guess we have to do to two step process anyways, with `setup.sh` as a separate step.

Another option I am thinking was to overload `install_requirements.py` but not much material difference if we still have to run setup.sh. At least if we can do everything AoT with just that then it might be better. What do you think?

examples/arm/setup.sh
```diff
@@ -223,64 +223,19 @@ function patch_repo() {
...
-    VNV=$(python3 -c "import sys; print('venv') if (sys.prefix != sys.base_prefix) else print('not_venv')")
-    if [ ${VNV} == "venv" ]; then
-	pip install .
-    else
-       # if not venv, we need the site-path where the vela
-       vela_path=$(python -c "import site; print(site.USER_BASE+'/bin')")
-       echo "export PATH=\${PATH}:${vela_path}" >> ${setup_path_script}
-       pip install . --user
-    fi
+    pip install ethos-u-vela@git+${vela_repo_url}@${vela_rev}
```
digantdesai, line 238: how would we apply patches if we have to?

examples/arm/setup.sh
```diff
@@ -223,64 +223,19 @@ function patch_repo() {
...
-        n=1
-    fi
-
-    if [[ "$n" -lt 1 ]]; then
-        n=1
-    fi
+    
+    # reference_model flatbuffers version clashes with Vela.
+    # go with Vela's since it newer.
+    # Could cause issues down the line, beware..
```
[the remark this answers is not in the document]
Erik-Lundell: Vela is working on widening their flatbuffer requirement, should be done within a couple of weeks

examples/arm/setup.sh
Erik-Lundell: We are currently not applying any Vela patches and want to move away from patching so we made the decision that simplifying the code like this was worth removing patching at this point.

backends/arm/test/common.py
```diff
@@ -74,19 +74,15 @@ def get_tosa_compile_spec_unbuilt(
...
-        intermediate_path = maybe_get_tosa_collate_path() or tempfile.mkdtemp(
-            prefix="arm_tosa_"
-        )
-    else:
-        intermediate_path = custom_path
+        custom_path = maybe_get_tosa_collate_path()
 
-    if not os.path.exists(intermediate_path):
-        os.makedirs(intermediate_path, exist_ok=True)
+    if custom_path is not None and not os.path.exists(custom_path):
```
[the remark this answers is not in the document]
Erik-Lundell: We don't, this was probably just inherited from old code. But good point.

backends/arm/test/common.py
```diff
@@ -74,19 +74,15 @@ def get_tosa_compile_spec_unbuilt(
...
+        os.makedirs(custom_path, exist_ok=True)
     compile_spec_builder = (
         ArmCompileSpecBuilder()
         .tosa_compile_spec(tosa_version)
         .set_permute_memory_format(permute_memory_to_nhwc)
-        .dump_intermediate_artifacts_to(intermediate_path)
+        .dump_intermediate_artifacts_to(custom_path)
```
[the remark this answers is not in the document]
Erik-Lundell: Should be ok. If you run on FVP and have custom_path=None you will get a crash, but with a explanatory error message.

backends/arm/test/common.py
```diff
@@ -74,19 +74,15 @@ def get_tosa_compile_spec_unbuilt(
     the compile spec before calling .build() to finalize it.
     """
     if not custom_path:
-        intermediate_path = maybe_get_tosa_collate_path() or tempfile.mkdtemp(
-            prefix="arm_tosa_"
-        )
-    else:
-        intermediate_path = custom_path
+        custom_path = maybe_get_tosa_collate_path()
```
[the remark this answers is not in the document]
Erik-Lundell: Since we don't have to save files unless we want to with this change, we don't need to create a temporary dir when custom_path is not -> I viewed this as a related change.

setup.py
digantdesai: I was thinking to overload ET's `install_requirements.py` script to optionally do `pip install git+https...` for Ethos setups. That way, if we could, at least for AoT-only experiments it can be a one step process, and you don't have to run setup.sh. 

If you want to do e2e i.e. install baremetal gcc, and FVP etc., I don't see a way to do it with just pip install today, and we may have to call a setup.sh from ET's `install_requirements.py` too. Then we can be a truly single step.


I don't have a strong preference on a shell script vs. a python script, but yeah let's keep it something which is not confusing i.e. not setup.py or install_requirements.py :-p

freddan80:
@digantdesai are u ok with the changes. I suspect that the [docker-builds](https://github.com/pytorch/executorch/actions/runs/12156784071) have pushed an image to the registry including this change. Hence all our CI fails.

examples/arm/setup.sh
digantdesai: sounds good to me.

setup.py
Erik-Lundell: Ok, then I think we are on the same page! Tbh I also prefer calling it something else than install_requirements.py, I think I misunderstood you hah!

digantdesai: so we will do that as a different step?

digantdesai:
To avoid docker issue, lets merge this and then we can discuss async.

PR #7091, opened 2024-11-26
Use a template for Android test spec

huydhn:
This setup helps address the 4GB file size limit enforced by AWS and also fixes the usage of `.ci/scripts/test_llama.sh` after https://github.com/pytorch/executorch/pull/6870.

This is the first PR for Android.  I also have another PR https://github.com/pytorch/executorch/pull/7098 to enable `google/gemma-2b` that @guangy10 added a while back, which I use to test this one for large model.

### Testing

Download the export `model.zip` from S3 successfully https://github.com/pytorch/executorch/actions/runs/12040976164.  I also attempt to export `google/gemma-2b`, but it looks like it's better to do that in a separate PR https://github.com/pytorch/executorch/pull/7098 to avoid bloating this one.

* google/gemma-2b https://github.com/pytorch/executorch/actions/runs/12043268246/job/33579733419
* stories110M https://github.com/pytorch/executorch/actions/runs/12053946075


huydhn:
`google/gemma-2b` tests started fine, but it took nearly 2 hour to finish copying the 10GB model to the device.

```
2024-11-27T05:05:06.5661289Z [DeviceFarm] adb -s $DEVICEFARM_DEVICE_UDID push *.pte /sdcard
2024-11-27T05:05:06.5661948Z [  0%] /sdcard/google-gemma-2b_xnnpack_fp32.pte
...
2024-11-27T06:06:41.0614924Z [ 99%] /sdcard/google-gemma-2b_xnnpack_fp32.pte
2024-11-27T06:06:41.0615084Z [100%] /sdcard/google-gemma-2b_xnnpack_fp32.pte
2024-11-27T06:06:41.0615419Z google-gemma-2b_xnnpack_fp32.pte: 1 file pushed. 63.5 MB/s (10025204816 bytes in 150.591s)
```

@guangy10 @kirklandsign Do you know that if that makes sense?  Here is the example run https://github.com/pytorch/executorch/actions/runs/12043268246/job/33579733419

huydhn:
Looking carefully at https://github.com/pytorch/executorch/actions/runs/12043268246/job/33579733419, it looks like the test finished successfully, but the teardown done by AWS failed and I didn't see any result JSON in the spec.

huydhn:
I think the test runs sort of ok on samsung 24 https://github.com/pytorch/executorch/actions/runs/12053919612/job/33614078559.  The large model is not a blocker anymore, but I don't see any benchmark results JSON file in the output.  However, I think it would need to be investigated separately after we land this PR.  Maybe we can figure out why locally first.

.github/workflows/android-perf.yml
```diff
@@ -125,9 +117,46 @@ jobs:
...
+
+      - name: Upload the spec
+        uses: seemethere/upload-artifact-s3@v5
+        with:
+          s3-bucket: gha-artifacts
+          s3-prefix: |
+            ${{ github.repository }}/${{ github.run_id }}/artifacts/${{ matrix.model }}_${{ matrix.delegate }}
+          retention-days: 1
+          if-no-files-found: error
+          path: extension/benchmark/android/benchmark/android-llm-device-farm-test-spec.yml
```
guangy10, line 155: Remind me if this is a dest path or source path. Concern if it will cause name conflicts and end up overriding unintentionally.

.github/workflows/android-perf.yml
```diff
@@ -125,9 +117,46 @@ jobs:
...
+
   export-models:
     name: export-models
-    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
```
guangy10, line 159: What's new in _v2.yml?

.github/workflows/upload-android-test-specs.yml
```diff
@@ -45,50 +29,3 @@ jobs:
       models: stories110M
       devices: samsung_galaxy_s22
       delegates: xnnpack
```
guangy10, line 31: Now this validation step will run the `android-perf.yml` with the specialized test-spec correct?

.github/workflows/upload-android-test-specs.yml
```diff
@@ -45,50 +29,3 @@ jobs:
...
-      test_spec: https://gha-artifacts.s3.amazonaws.com/${{ github.repository }}/${{ github.run_id }}/artifacts/android-llm-device-farm-test-spec.yml
-
-  upload-android-test-spec:
```
guangy10, line 50: It seems the upload step is moved to android-perf.yml right after the test-spec is specialized and instantiated, w/o validation. Given that, do we still need to keep a validation step here? How would it catch and prevent bad test-spec being uploaded and used?

guangy10 (commented):
Left comments for clarification, other than that looks great!

.github/workflows/android-perf.yml
huydhn: It's the source path, so it's the local path to `executorch/extension/benchmark/android/benchmark/android-llm-device-farm-test-spec.yml`

.github/workflows/android-perf.yml
huydhn: It supports manywheel 2.28 (what PyTorch is moving to).  For example, arm binaries are now using this new format https://github.com/pytorch/executorch/pull/7080.  During my testing, I saw an error on this job about old v.s. new wheel format.  So, I just move it to v2.

This has been delay on PyTorch side until the next release, so this is kind of not needed right now.  Let me put it back. 

This is ok too because manywheel 2.28 is backward compatible with the current older format.

.github/workflows/upload-android-test-specs.yml
huydhn: Yup, it will test the specialized test spec

.github/workflows/upload-android-test-specs.yml
huydhn: In this new setup, the specialized test spec will just be an artifact of the current job uploaded at `https://gha-artifacts.s3.amazonaws.com/${{ github.repository }}/${{ github.run_id }}/artifacts/${{ matrix.model }}_${{ matrix.delegate }}/android-llm-device-farm-test-spec.yml`.  If the spec is bad, the subsequent benchmark job would fail.

The validation workflow here now acts just as a trigger to call android-perf workflow when the spec template changes.

Finally, the upload step here is being deleted (not moving to android-perf) because there would not be a single share spec file at `s3://ossci-android/executorch/android-llm-device-farm-test-spec.yml` anymore.  It couldn't be because each specialized spec now has different model+delegation.

guangy10: Maybe this can be simplified and merged into `android-perf.yml` by triggering a run when the test spec template is modified?

huydhn: We could do that but the job would need to use the default value in CRON_DEFAULT_MODELS, it seems overkill to me to test them all just to validate the spec.  But I think I could tweak it a bit to use a different value for pull_request and keep the full list for scheduled jobs.

huydhn: Here it the example run https://github.com/pytorch/executorch/actions/runs/12150009281 without `upload-android-test-specs.yml`

guangy10: > job would need to use the default value in CRON_DEFAULT_MODELS, it seems overkill to me to test them all just to validate the spec.

I see. This is a fair point. Maybe we could add new defaults or change the defaults for test-spec validation to `android-perf.yml`? 
I don't have strong option to merge it to the `android-perf.yml`, thinking in that direction because job `upload-android-test-specs.yml` now is essentially validating the `android-perf.yml`, the specialized test-spec generated on the fly is already part of it.

guangy10 (approved):

Issue #7083, opened 2024-11-26
Create "make clean" equivalent script that encapsulates all necessary steps


mergennachin:
### 🚀 The feature, motivation and pitch

Currently, we have to type the following command manually `rm -rf cmake-out pip-out`

Let's encapsulate into a single script that does all these necessary steps.

PR #7108, opened 2024-11-27
sh clean.sh script that does cleaning

mergennachin:
Let's encapsulate important steps inside this single script for cleaning up.

As next steps, I'd like to restart buck as part of this script.

https://github.com/pytorch/executorch/issues/7083

### DIFF
--- a/.ci/scripts/build_llama_android.sh
+++ b/.ci/scripts/build_llama_android.sh
@@ -12,7 +12,8 @@ source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
 
 install_executorch_and_backend_lib() {
   echo "Installing executorch and xnnpack backend"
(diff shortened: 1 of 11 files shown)


kirklandsign (approved):

clean.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree
+
+set -eu
+rm -rf cmake-out pip-out
```
kirklandsign, line 9: Could also add `cmake-android-out` and `cmake-out-android-*`

mergennachin: actually not creating sh.clean.sh anymore. 

Reusing `install_requirements.sh --clean` instead.

PR #7111, opened 2024-11-27
Run decompositions before the quantizer

mcremon-meta:
Summary:
In the current flow, decompositions run in `to_edge()`, long after the quantization process is done. This creates a lot of issues, since we cannot quantize any operations contained in the large operators that the graph tracer can give (e.g. aten.scaled_dot_product_attention, aten.rnn_<tanh, relu>.input, and a few others).
Any models using those will see many fp32 operators in the final graph. Running the decomps earlier solves the problem, but we need to retain a couple operators that we do rely on in the quantizer, namely `aten.linear`, `aten.conv1d` and `aten.conv2d`.

Differential Revision: D66461406

### DIFF
--- a/backends/cadence/aot/compiler.py
+++ b/backends/cadence/aot/compiler.py
@@ -28,6 +28,7 @@ from executorch.exir import (
     to_edge,
 )
 from executorch.exir.pass_base import PassResult
+from torch._inductor.decomposition import remove_decompositions
 from torch.ao.quantization.pt2e.export_utils import model_is_exported
 from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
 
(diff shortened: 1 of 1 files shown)


backends/cadence/aot/compiler.py
```diff
@@ -59,15 +60,25 @@ def convert_pt2(
     """
 
     # Export with dynamo
-    model_gm = torch.export.export_for_training(model, inputs).module()
-
-    if model_gm_has_SDPA(model_gm):  # pyre-fixme[6]
+    decomp_table = torch.export.default_decompositions()
+    ops_to_keep = [
```
zonglinpeng, line 64: would be nice to leave the same comment inline

backends/cadence/aot/compiler.py
```diff
@@ -59,15 +60,25 @@ def convert_pt2(
...
+        torch.ops.aten.conv1d.default,
+        torch.ops.aten.conv2d.default,
+        torch.ops.aten.layer_norm.default,
+        torch.ops.aten.linear.default,
+        torch.ops.aten.matmul.default,
+    ]
+    # pyre-fixme[6]: For 1st argument expected `Dict[typing.Callable[..., typing.Any
```
zonglinpeng, line 71: pyre are disabled in ET but still used internally. We need to sort it out but not in this PR

zonglinpeng (approved):

PR #7117, opened 2024-11-27
Onboard ExecuTorch to benchmark database v3

huydhn:
ExecuTorch is currently using v2 schema which doesn't have a dedicated field to store the delegate backend. Onboarding it to v3 will allow the dashboard to have this field and unblock T204741729.

* I opt to keep both v2 and v3 in the interim.  The former can be cleaned up once the [dashboard](https://hud.pytorch.org/benchmark/llms?repoName=pytorch%2Fexecutorch) is updated.
* Fix a small regex bug on Android when parsing `tinyllama_xnnpack+custom+qe_fp32` with + in the delegate backend.  This has been fixed on iOS, but I missed this on Android.

### Testing

* Android https://github.com/pytorch/executorch/actions/runs/12061313284
* Apple https://github.com/pytorch/executorch/actions/runs/12061317075

The records are available on both v2 and v3 tables, i.e. `select * from oss_ci_benchmark_v3 where workflow_id = 12061317075`

(The failure for stories110M is fixed by https://github.com/pytorch/executorch/pull/7091)

### DIFF
--- a/.github/workflows/android-perf.yml
+++ b/.github/workflows/android-perf.yml
@@ -298,15 +298,25 @@ jobs:
               --workflow-run-attempt ${{ github.run_attempt }}
           done
 
-          ls -lah benchmark-results
-
-          for BENCHMARK_RESULTS in benchmark-results/*.json; do
-            cat "${BENCHMARK_RESULTS}"
-            echo
+          for SCHEMA in v2 v3; do
+            for BENCHMARK_RESULTS in benchmark-results/"${SCHEMA}"/*.json; do
+              cat "${BENCHMARK_RESULTS}"
+              echo
+            done
           done
 
-      - name: Upload the benchmark results
+      # TODO (huydhn): Remove v2 schema once the benchmark dashboard finishes the migration
+      - name: Upload the benchmark results (v2)
         uses: pytorch/test-infra/.github/actions/upload-benchmark-results@main
         with:
-          benchmark-results-dir: 'benchmark-results'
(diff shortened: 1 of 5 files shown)


.github/workflows/android-perf.yml
```diff
@@ -298,15 +298,25 @@ jobs:
...
           done
 
-      - name: Upload the benchmark results
+      # TODO (huydhn): Remove v2 schema once the benchmark dashboard finishes the migration
+      - name: Upload the benchmark results (v2)
+        uses: pytorch/test-infra/.github/actions/upload-benchmark-results@main
+        with:
+          benchmark-results-dir: benchmark-results/v2
+          dry-run: false
+          schema-version: v2
```
guangy10, line 314: Isn't `GITHUB_TOKEN` required by v2?

guangy10 (approved):
Looks great!

.github/workflows/android-perf.yml
huydhn: Oh, this token is used to query the job ID, which isn't available as part of the job context.  In the old schema, it is supplied by `.github/scripts/extract_benchmark_results.py` script using the handmade regex in https://github.com/pytorch/executorch/blob/main/.github/scripts/extract_benchmark_results.py#L24.  In the new version, the action uses the token to do it on behalf of the caller, so `.github/scripts/extract_benchmark_results.py` would only need to extract the benchmark results from AWS, and there is no need to use any hardcoded logic.  I'll eventually clean up the script later.

PR #7136, opened 2024-12-02
Correct build instructions for the MediaTek backend

jcwtw:
### Summary
While trying to build the MediaTek backend, I stumbled on two problems that prevented successful builds. I amended the build instruction to use the right Buck2 tool and set the minimum platform level for NDK to get the build script to work.

### Test plan
Try building the MediaTek backend with the current instructions and you will hit the bumps as I did. Now following my updated instructions and you will be able to complete building the code.


cccclai:
This seems correct to me. cc: @neuropilot-captain

PR #7142, opened 2024-12-02
Introduce executorch_ops_check

larryliu0820:
Summary: Introducing `executorch_ops_check`. For any given target, find all the `executorch_generated_lib` targets in the dependent transitive closure and make sure the operators in `executorch_generated_lib` are not overlapping.

Differential Revision: D66560425


lucylq (approved):
lgtm! note linter failure

PR #7165, opened 2024-12-03
[executorch] Populate cadence cpu ops

zonglinpeng:
### Summary
quantized ops in CPU flow are not fully migrated. Adding them all in this PR
#### Custom ops
- quantized_linear_per_tensor_out
- im2row_out
- quantized_convolution_per_tensor_out

#### Native ops
- op: transpose_copy.int_out
- op: eq.Scalar_out
- op: logical_not.out
- op: any.out
- op: native_group_norm.out
- op: sum.IntList_out
- op: select_copy.int_out

### Test plan
python3 -m examples.cadence.operators.quantized_linear_op
python3 -m examples.cadence.models.babyllama


backends/cadence/reference/operators/operators.h
```diff
@@ -0,0 +1,59 @@
...
+namespace impl {
+namespace cpu {
+namespace native {
+namespace {
+using ::executorch::runtime::getLeadingDims;
+
+
+#define ET_FORALL_CADENCE_QUANTIZED_TYPES(_) \
+  _(uint8_t, Byte)                          \
+  _(int8_t, Char)
```
mcremon-meta, line 20: general comment, we should be able to add int16_t and uint16_t already, but let's ignore for now

backends/cadence/reference/operators/quantized_ops.h
```diff
@@ -0,0 +1,233 @@
...
+      for (size_t i = 0; i < X_numel; ++i) {                                  \
+        float x = kernels::dequantize<T>(X_data[i], X_scale, X_zero_point);   \
+        float z = x OP y;                                                     \
+        out_data[i] = kernels::quantize<T>(z, inv_out_scale, out_zero_point); \
+      }                                                                       \
+    } /* other broadcasting cases */                                          \
+    else {                                                                    \
+      ET_DCHECK_MSG(false, "Unsupported broadcasting");                       \
+    }                                                                         \
+  }
```
mcremon-meta, line 51: this macro is not used. It's only for quantized_add, and will need to be removed anyway with PT2 quant

mcremon-meta (approved):
LGTM with the extra macro removed. Thanks!

.lintrunner.toml
```diff
@@ -74,6 +74,7 @@ exclude_patterns = [
     # NB: Objective-C is not supported
     'examples/apple/**',
     'examples/demo-apps/apple_ios/**',
+    'examples/demo-apps/react-native/rnllama/ios/**',
```
mcremon-meta, line 77: lint change I guess?

backends/cadence/reference/operators/quantized_conv_out.cpp
```diff
@@ -1,261 +0,0 @@
-/*
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- * All rights reserved.
- *
- * This source code is licensed under the BSD-style license found in the
- * LICENSE file in the root directory of this source tree.
- */
-
-#include <executorch/backends/cadence/reference/kernels/kernels.h>
```
mcremon-meta, line 9: can we put the `convolution.cpp` stuff in this one instead, and remove the non-quant version? It should always be quantized anyway, and it will be less confusing

mcremon-meta (commented):
not sure why there are so many changes, maybe a linter error?

.lintrunner.toml
zonglinpeng: caused by rebasing to main

PR #7170, opened 2024-12-04
Remove false positive error message in the executor_runner

ykhrustalev:
### Summary

There are false positive error messages from the type check that attempts to verify several allowed types at once, but uses one function call per type leading to some of them not passing the check and emitting an error message.

```
> ./executor_runner --model-path ~/model.pte
...
I 00:00:00.519929 executorch:executor_runner.cpp:91] Using method forward
I 00:00:00.519931 executorch:executor_runner.cpp:138] Setting up planned buffer 0, size 10753549312.
I 00:00:01.212589 executorch:executor_runner.cpp:161] Method loaded.
I 00:00:01.213544 executorch:executor_runner.cpp:171] Inputs prepared.
E 00:00:02.320414 executorch:tensor_util.h:481] Expected to find Float type, but tensor has type Half
E 00:00:02.320521 executorch:tensor_util.h:487] Check failed (t.scalar_type() == dtype): Expected to find Float type, but tensor has type Half
...
I 00:00:24.911098 executorch:executor_runner.cpp:180] Model executed successfully.
...
```

Note the repeated lines
> Check failed (t.scalar_type() == dtype): Expected to find Float type, but tensor has type Half

### Test plan

I can only provide the log message from my local run for that
```
...
I 00:00:00.745945 executorch:executor_runner.cpp:138] Setting up planned buffer 0, size 13026185216.
I 00:00:02.139255 executorch:executor_runner.cpp:161] Method loaded.
I 00:00:02.139295 executorch:executor_runner.cpp:171] Inputs prepared.
I 00:00:28.297681 executorch:executor_runner.cpp:180] Model executed successfully.
...
```


runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -484,6 +486,30 @@ inline bool tensor_is_type(
   return true;
 }
 
+inline bool tensor_is_type(
+    executorch::aten::Tensor t,
+    const std::vector<executorch::aten::ScalarType>& dtypes) {
```
dbort, line 491: The runtime must not use streams, vectors or any other type that allocates memory dynamically: see https://pytorch.org/executorch/main/portable-cpp-programming.html for some commentary on this.

In this case you may be able to use a `Span<ScalarType>` since the container size is known at construction time.

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -484,6 +486,30 @@ inline bool tensor_is_type(
...
+  if (std::find(dtypes.begin(), dtypes.end(), t.scalar_type()) !=
+      dtypes.end()) {
+    return true;
+  }
+
+  std::stringstream dtype_ss;
```
dbort, line 497: It's not safe to assemble a dynamic string here. If you wanted to print all of the types, printing individual log lines would be the safest and easiest. If you wanted to assemble a string, you could create a fixed-size buffer on the stack and `strncat` strings onto it, but that can be error-prone.

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -484,6 +486,30 @@ inline bool tensor_is_type(
...
+    }
+    dtype_ss << torch::executor::toString(dtypes[i]);
+  }
+
+  ET_LOG_MSG_AND_RETURN_IF_FALSE(
+      false,
+      "Expected to find one of %s types, but tensor has type %s",
+      dtype_ss.str().c_str(),
+      torch::executor::toString(t.scalar_type()));
+  return false;
```
dbort, line 510: No need to use "return if false" when it's always false; a basic log macro and a return statement is more clear.
```suggestion
  ET_LOG(Info
      "Expected to find one of %s types, but tensor has type %s",
      dtype_ss.str().c_str(),
      torch::executor::toString(t.scalar_type()));
  return false;
```

dbort (dismissed):
It would definite be good to avoid printing log statements in the success path like this: ideally, ET will never print logs during successful execution.

@manuelcandales what do you think about this interface?

runtime/core/exec_aten/util/tensor_util.h
ykhrustalev: yes, I was thinking about using the c-ctyle buffer for that, I will update

runtime/core/exec_aten/util/tensor_util.h
ykhrustalev: noted, I will use the Span container

runtime/core/exec_aten/util/tensor_util.h
ykhrustalev: I decided to go a simple way that provides an explicit override without any loops. 

The use of strncat will lead to a warning https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83404 which I think is not desired.

@dbort , hope you find that version suitable.

ykhrustalev: @dbort surficing the thread on top.

manuelcandales (approved):

shoumikhin:
@manuelcandales can you take a look please how shall we proceed?

Issue #7019, opened 2024-11-21
macOS nightly wheel builds failing since 2024-11-19


swolchok:
### 🐛 Describe the bug

Status page: https://github.com/pytorch/executorch/actions/workflows/build-wheels-m1.yml
Note that the Python 3.9 build always fails, so even though the runs are red, they were successful through 2024-11-18.

Linking is failing with `ld: invalid use of ADRP in '_init_f32_vcopysign_config' to '_xnn_f32_vcopysign_ukernel__neon_u8’`.

### Versions

N/A

swolchok:
Inspection of PRs landed between the last good build and first bad build suggested the following:

- https://github.com/pytorch/executorch/pull/6837 (update sccache)
- https://github.com/pytorch/executorch/pull/6522 (optimized sigmoid, just because it’s one of the few PRs touching C++)
- https://github.com/pytorch/executorch/pull/6892/ (another sigmoid change)

Trial revert of #6837 in #7013 still failed the job; trialing revert of the other two PRs together

swolchok:
trial revert of #6522 in https://github.com/pytorch/executorch/pull/7020 did not fix the job

swolchok:
trial revert of #6892  in #7021 did not fix the job.

I am also not able to repro this locally, and I've inspected `git diff 8526d0a2d798658b6a6e3a42ec935b8093f355ef..04f6fcd4b3920eaf1be9905d12b449f301f89ca7` without finding anything else suspicious, so I wonder if the runners broke somehow

swolchok:
> I wonder if the runners broke somehow

I reran the last good workflow run; builds succeeded (there were some failures due to an unrelated issue).

larryliu0820:
Found a failure with the same error message in a different job (`test-llama-runner-mac`): https://github.com/pytorch/executorch/actions/runs/11959891658/job/33342737621?pr=7010

swolchok:
> Found a failure with the same error message in a different job (`test-llama-runner-mac`): https://github.com/pytorch/executorch/actions/runs/11959891658/job/33342737621?pr=7010

that job is green on trunk runs though! https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=llama-runner-mac%20(fp32%2C%20mps

kimishpatel:
am late to this so not sure my comments will help, but any change related to xnnpack upgrade? since the job fails related xnnpack

swolchok:
[...]

PR #7173, opened 2024-12-04
Fix xnn linkage issue when building wheels

mcr229:
Resolving the following issue:

https://github.com/pytorch/executorch/issues/7019

XNNPACK has split its targets to XNNPACK (core orchestration of kernels) and microkernels-prod to separate static libraries. In upstream PyTorch we updated to this change, and had to make changes throughout the PyTorch CMake to install and link both XNNPACK and microkernels-prod.

Originally I had thought the issue stemmed from a mistake in PyTorch in which we weren't correctly linking microkernels-prod to torch_cpu, which when linked with portable pybindings was causing missing symbol issues related to ADRP. This led me to spend a lot of time messing with the pytorch cmake, rebuilding the wheels locally, pip installing them and trying to rebuild the executorch wheels. However, after much troubleshooting, I realized that the issue wasn't that the microkernels-prod static library was missing, but rather because it was there. This was likely causing problems duplicate linking problems with the microkernels-prod from our xnnpack_backend target as we were looking for microkernel function pointers in torch_cpu.

Previously we explicitly linked XNNPACK along with the xnnpack_backend in order to avoid using the XNNPACK symbols libtorch_cpu.dylib(from pytorch). Since Executorch and Pytorch now have the change of split targets between XNNPACK and microkernels-prod, we also have to explicitly link microkernel-prod we built for the xnnpack_backend to avoid using the static library from upstream torch_cpu. This is likely what caused the `ld: invalid use of ADRP in ...` because the function pointer to the ukernel (in microkernel-prod) was pointing to the instance of torch_cpu.

I tested it out by finally rebuilding the wheels with the latest torch nightly, and it was able to succeed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -738,9 +738,9 @@ if(EXECUTORCH_BUILD_PYBIND)
   endif()
 
   if(EXECUTORCH_BUILD_XNNPACK)
-    # need to explicitly specify XNNPACK here otherwise uses XNNPACK symbols
-    # from libtorch_cpu
-    list(APPEND _dep_libs xnnpack_backend XNNPACK)
+    # need to explicitly specify XNNPACK and microkernels-prod
+    # here otherwise uses XNNPACK and microkernel-prod symbols from libtorch_cpu
+    list(APPEND _dep_libs xnnpack_backend XNNPACK microkernels-prod)
   endif()
 
   # compile options for pybind


huydhn:
I'm adding `ciflow/binaries` to test this out on PR. It would trigger the nightly build here.

larryliu0820 (approved):

huydhn (approved):
M1 build looks ok now https://github.com/pytorch/executorch/actions/runs/12164871356

CMakeLists.txt
```diff
@@ -738,9 +738,9 @@ if(EXECUTORCH_BUILD_PYBIND)
   endif()
 
   if(EXECUTORCH_BUILD_XNNPACK)
-    # need to explicitly specify XNNPACK here otherwise uses XNNPACK symbols
-    # from libtorch_cpu
-    list(APPEND _dep_libs xnnpack_backend XNNPACK)
+    # need to explicitly specify XNNPACK and microkernels-prod
+    # here otherwise uses XNNPACK and microkernel-prod symbols from libtorch_cpu
+    list(APPEND _dep_libs xnnpack_backend XNNPACK microkernels-prod)
```
digantdesai, line 743: surprised it didn't cause more problems TBH

digantdesai (approved):

guangy10 (approved):
Thanks for troubleshooting and fix!

guangy10:
Curious how it is related to the broken `libtorch-cpu-shared-with-deps-cxx11-abi-build` in pytorch

PR #7178, opened 2024-12-04
Fix runnning arm unittests on FVP

AdrianLundell:
- Adds _test_options to pytest namespace which has a proper global scope
- Removes test option enum which is not compatible with new approach
- Adds back expectedFailures which passed due to not running on FVP


AdrianLundell:
All reporter failures are
`ERROR: Could not install packages due to an OSError`
-> not related to patch

freddan80 (approved):

PR #7228, opened 2024-12-06
Add util function in etdump to get debug buffer size

YIWENX14:
Summary: Add util function `get_debug_buffer_size` in `ETDumpGen` to get debug buffer size in backend execution.

Differential Revision: D66855639


Olivia-liu (approved):
lgtm

PR #7266, opened 2024-12-10
Add stricter transpose condition for TOSA reshape lowering

AdrianLundell:
Removes transposes in lowered graph for reshapes in only H,W dimension and clarifies logic in the annotate_channels_last_dim_order_pass


AdrianLundell:
Docker error not related to patch

freddan80 (approved):

PR #7278, opened 2024-12-10
Support to init BundledProgram from pte file

YIWENX14:
Summary: Added an optional `pte_file_path` arg to BundledProgram's init function. This is to allow users to create bundled program with varied inputs after exporting.

Differential Revision: D67013542

### DIFF
--- a/devtools/bundled_program/core.py
+++ b/devtools/bundled_program/core.py
@@ -9,6 +9,7 @@ import typing
 from typing import Dict, List, Optional, Sequence, Type, Union
 
 import executorch.devtools.bundled_program.schema as bp_schema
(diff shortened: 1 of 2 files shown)


tarun292 (approved):

zingo (commented):
This seem to cause major problem 

from pyre_extensions import none_throws

As pyre is not installed in the github project. I kicked of a revert to test if and added the trunk flag to get more testing on it.

digantdesai:
@zingo - Wonder why the CI here is clean.

zingo:
This revert is may not be the thing you want to do, I just used it to highlight the problems, and test if it solves it.

PR #7279, opened 2024-12-11
Fix executorch kv cache incompatibility with to_executorch lowering

jackzhxng:
### Summary
Fix the Llama 3.2 vision text decoder prefill issue by marking the kv cache as an initialized mutable buffer in a custom pass

### Test plan
- Add kv_cache export tests that test accuracy against torchtune eager and verify contents of the cache after prefill and token-by-token generation
- Export and run full Llama 3.2 vision text decoder
```
> python -m examples.models.llama.export_llama --model llama3_2_vision --checkpoint /tmp/Llama-3.2-11B-Vision-Instruct/original/consolidated.pth --params examples/models/llama3_2_vision/text_decoder/params/demo_config.json  --metadata '{"append_eos_to_prompt": 0, "get_bos_id":128000, "get_eos_ids":[128009, 128001], "get_n_bos": 0, "get_n_eos": 0}' --output_name="llama3_2_vision.pte" -d fp32 --verbose --max_seq_length 64 -k
> python -m examples.models.llama3_2_vision.runner.native --model llama3_2_vision --pte llama3_2_vision.pte  --tokenizer /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model --prompt "Who's the founder of Meta?" --params examples/models/llama3_2_vision/text_decoder/params/demo_config.json --max_len 64 -kv --temperature 0
```


exir/passes/cache_pos_init_mutable_pass.py
```diff
@@ -0,0 +1,27 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from executorch.exir.pass_base import ExportPass
+
+
+class CachePosToInitializedMutableBufferPass(ExportPass):
```
tarun292, line 11: Let's make this pass more generic for people who need it in the future. Maybe call it `InitializeMutableBufferPass` that takes in a list of strings that are the names of the buffers that need to be initialized. We find those buffers and then set the node metadata on them. If we don't find the buffer of that name we can raise an Exception.

exir/passes/init_mutable_pass.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self, patterns: List[str]) -> None:
+        super().__init__()
+        self.patterns = patterns
+
+    def placeholder(self, name: str, arg, meta):
+        for pattern in self.patterns:
+            if pattern in name:
+                meta["et_init_buffer"] = True
+
```
tarun292, line 30: We should raise an exception here if we don't find the buffer.

iseeyuan (approved):

exir/passes/init_mutable_pass.py
dvorjackz: This is just for one placeholder, don't think we need to be raising the exception at this level

exir/emit/_emitter.py
```diff
@@ -1655,7 +1657,11 @@ def placeholder(
                 spec.storage = real_tensor.untyped_storage()
 
             # User inputs and mutable buffers are not constants, other buffers or parameters are.
-            spec.const = not (is_user_input or is_mutable_buffer)
+            if initialize_buffer:
+                assert is_mutable_buffer
+                spec.const = True
+            else:
+                spec.const = not (is_user_input or is_mutable_buffer)
```
dbort, line 1664: Please add unit tests for this logic; tests that would have broken before this fix, and would have caught this kv cache incompatibility

exir/passes/init_mutable_pass.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
dbort, line 1: Please add unit tests for this new file

dbort (changes requested):

exir/emit/test/test_emit.py
```diff
@@ -243,6 +246,55 @@ def forward(self, x):
...
+        ep = torch.export.export(m, example_inputs)
+        edge = to_edge(
+            ep,
+            compile_config=EdgeCompileConfig(
+                _check_ir_validity=False,
+            ),
+        )
+
+        # Save a copy of the edge program since to_executorch is
+        # stateful to sombe degree.
```
dbort, line 273:
```suggestion
        # stateful to some degree.
```

exir/emit/test/test_emit.py
```diff
@@ -243,6 +246,55 @@ def forward(self, x):
...
+        method_regular = program_regular.load_method("forward")
+
+        # Test that the mutable buffer is initialized.
+        torch.allclose(
+            method_init_pass.execute((example_inputs))[0], torch.arange(1, 11)
+        )
+        # Test that the mutable buffer is uninitialized and starts with default zeros.
+        torch.allclose(
+            method_regular.execute((example_inputs))[0],
+            torch.ones(10, dtype=torch.int64),
```
dbort, line 295: Should this be `zeros`, based on the comment? If not, please update the comment to clarify why this is `ones`.

dbort (approved):
Thanks for adding the emitter tests!

exir/emit/test/test_emit.py
dvorjackz: Oh, it's because in the forward of the model we do `self.cache_pos += 1`, I'll specify this

exir/passes/init_mutable_pass.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+from typing import List
+
+from executorch.exir.pass_base import ExportPass
+
+
+class InitializedMutableBufferPass(ExportPass):
+    """
+    If the buffer has the name "cache_pos", such as in an kv_cache
```
JacobSzwejbka, line 15: cache pos is no longer relevant to the pass description

exir/emit/_emitter.py
```diff
@@ -1575,7 +1575,8 @@ def _find_fqn_for_placeholder(
                     warnings.warn(
                         "Mutation on a buffer in the model is detected. ExecuTorch assumes "
                         "buffers that are mutated in the graph have a meaningless initial state, "
-                        "only the shape and dtype will be serialized.",
+                        "only the shape and dtype will be serialized, unless a pass which marks "
+                        "spec.const=True such as InitializedMutableBufferPass is run.",
```
JacobSzwejbka, line 1579: can you just mention meta["et_init_buffer"] instead? spec.const is more of an implementation detail now that we have the pass.

JacobSzwejbka (changes requested):

exir/emit/_emitter.py
dvorjackz: Yup that's a great point

JacobSzwejbka (approved):

PR #7290, opened 2024-12-11
Arm backend: Remove running examples/arm/setup.sh in the github docker

zingo:
This will move Arm PR testing to use an empty docker and examples/arm/setup.sh will always run from a fresh setup when testing a PR to minimize the risk of and old run of examples/arm/setup.sh done during docker creation colliding PR testing, or being out of sync with it. The different runs of examples/arm/setup.sh also used different install folders and that might break stuff even more.

### DIFF
--- a/.ci/docker/ubuntu/Dockerfile
+++ b/.ci/docker/ubuntu/Dockerfile
@@ -79,9 +79,6 @@ RUN if [ -n "${ANDROID_NDK_VERSION}" ]; then bash ./install_android.sh; fi
 RUN rm install_android.sh
 
 ARG ARM_SDK
-COPY --chown=ci-user:ci-user ./arm /opt/arm
-# Set up ARM SDK if needed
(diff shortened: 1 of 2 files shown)


freddan80 (approved):

.ci/docker/ubuntu/Dockerfile
```diff
@@ -79,9 +79,6 @@ RUN if [ -n "${ANDROID_NDK_VERSION}" ]; then bash ./install_android.sh; fi
 RUN rm install_android.sh
 
 ARG ARM_SDK
-COPY --chown=ci-user:ci-user ./arm /opt/arm
-# Set up ARM SDK if needed
-RUN if [ -n "${ARM_SDK}" ]; then git config --global user.email "ossci@example.com"; git config --global user.name "OSS CI"; bash /opt/arm/setup.sh --i-agree-to-the-contained-eula /opt/arm-sdk; chown -R ci-user:ci-user /opt/arm-sdk; fi
```
digantdesai, line 84: I guess `install_arm` will do it at runtime?

digantdesai (approved):
Thanks let's hope we can reduce frequency for the docker build issues.

PR #7303, opened 2024-12-12
Refer "./install_requirements.sh --clean" in documents

### LINKED ISSUE #7083: Create "make clean" equivalent script that encapsulates all necessary steps

mergennachin:
### Summary 

Currently, most documents will explicitly ask users to run "rm -rf"

But we have a dedicated command that does the cleanup.

Fixes #7083

### Test plan

Reproduced all the commands in the .md file. Fixed some of the existing bugs, for instance, in QNN command.

### DIFF
--- a/backends/vulkan/docs/android_demo.md
+++ b/backends/vulkan/docs/android_demo.md
@@ -81,7 +81,8 @@ First, build and install ExecuTorch libraries, then build the LLaMA runner
 binary using the Android NDK toolchain.
 
 ```shell
-(rm -rf cmake-android-out && \
+./install_requirements.sh --clean
+(mkdir cmake-android-out && \
   cmake . -DCMAKE_INSTALL_PREFIX=cmake-android-out \
     -DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK/build/cmake/android.toolchain.cmake \
     -DANDROID_ABI=$ANDROID_ABI \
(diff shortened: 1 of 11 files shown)


mergennachin:
Corresponding GH issue: https://github.com/pytorch/executorch/issues/7083

build/Utils.cmake
```diff
@@ -235,7 +235,7 @@ function(resolve_buck2)
 
   set(resolve_buck2_command
       ${PYTHON_EXECUTABLE} ${executorch_root}/build/resolve_buck.py
-      --cache_dir=${CMAKE_CURRENT_BINARY_DIR}/buck2-bin
```
guangy10, line 238: Move the buck2-bin to root dir, so that it can be cleaned up? Isn't already covered by cleaning up `CMAKE_CURRENT_BINARY_DIR` (i.e. `cmake-out`)?

guangy10 (approved):
Looks good.

build/Utils.cmake
mergennachin: Yeah, did that here

https://github.com/pytorch/executorch/pull/7305

PR #7305, opened 2024-12-12
Be able to compile in offline mode

mergennachin:
### Summary

Currently, during compilation via cmake, we can't do it in offline environments (e.g., doing development in airplane) for instance. 

There are two things we can do: 

1. Currently, we are downloading buck2 from the internet. This is fine, but we're downloading every time we `rm -rf cmake-out` file. Let's just download once and save inside `executorch/buck2-bin/` folder directly instead of `executorch/cmake-out/buck2-bin/`. That way, you will most likely have the necessary buck2 binary downloaded/cached and rarely deleted. There's little reason to delete this directory. 

2. FXDiv is getting downloaded if it is not defined. https://github.com/google/XNNPACK/blob/effe4ca0631f3fb0dac9551dab6f1d8a2eaabba2/CMakeLists.txt#L355. However, this is an oversight, since we actually have it already (https://github.com/pytorch/executorch/blob/main/.gitmodules#L19), just haven't set the environment variable appropriately. 

### Test plan

- Do installation once with internet once.
- Turn on airplane mode
- ./install_requirements.sh --clean
- Do compilation again `(mkdir cmake-out && cd cmake-out && cmake ..)`


mergennachin:
https://github.com/pytorch/executorch/issues/7306

tarun292 (approved):

PR #7313, opened 2024-12-12
Fix errors in comparison functions when dtype is uint8

YIWENX14:
Summary: Fixed error in comparison functions that were producing wrong results with uint8 dtype.

Differential Revision: D67163600


Olivia-liu (approved):
Thanks for making the fix!

Olivia-liu (approved):

PR #7322, opened 2024-12-13
[ET-VK] Don't specify memory layouts when testing

SS-JIA:
* __->__ #7322

As title. Now that we have a memory metadata tagging pass that automatically determines the optimal memory layout to use for operators, there is no need to specify what memory layout to test in the Python export tests.

There were some issues with the memory metadata tagging pass when dealing with nodes that contain tensor lists, which have been fixed as part of this diff as well.

Differential Revision: [D67180897](https://our.internmc.facebook.com/intern/diff/D67180897/)

### DIFF
--- a/backends/vulkan/test/test_vulkan_delegate.py
+++ b/backends/vulkan/test/test_vulkan_delegate.py
@@ -10,8 +10,6 @@ import ctypes
 import unittest
 from typing import Tuple
 
-import executorch.backends.vulkan.serialization.vulkan_graph_schema as vk_graph_schema
-
 import torch
 
 from executorch.backends.transforms.convert_dtype_pass import I64toI32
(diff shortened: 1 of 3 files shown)


junpi3 (approved):

backends/vulkan/test/test_vulkan_delegate.py
```diff
@@ -1722,5 +1652,33 @@ def forward(self, x):
         self.lower_module_and_test_output(
             GridPriorsModule(),
             (torch.rand(size=[1, 5, 2, 3]),),
-            memory_layouts=[vk_graph_schema.VkMemoryLayout.TENSOR_CHANNELS_PACKED],
         )
+
```
kirklandsign, line 1656: Out of sync?

PR #7323, opened 2024-12-13
Add type error suppressions for upcoming upgrade

Solumin:
Reviewed By: Solumin

Differential Revision: D67162130


Solumin:
I can't see what the actual linter error is. It may be complaining that the pyre-fixmes added by this PR aren't related to any type errors. This is expected, because they're silencing errors that are found by the upcoming release.

PR #7341, opened 2024-12-17
Arm backend: Add MM to BMM pass

Sebastian-Larsson:
aten.mm does not support input or output tensors of rank 3, which is
required by TOSA for the MM operation. Therefore, create a pass that
converts any MM nodes to BMM (which always has rank 3). The pass also
unsqueezes tensors of rank 2 to rank 3.

As a result of the new pass, op_mm.py is no longer required and has
been removed.

Signed-off-by: Sebastian Larsson <sebastian.larsson@arm.com>


freddan80:
Rebase pls

Sebastian-Larsson:
Rebase done!

freddan80:
Rebase pls

Sebastian-Larsson:
Rebase done

freddan80 (approved):

PR #7349, opened 2024-12-18
Enable composable benchmark configs for flexible model+device+optimization scheduling

guangy10:
It's not true that any combination of model + delegate can work, which make adding a new model to continuous run not easy as the workflow will run it across all delegates. Besides, each delegate may run with different configurations. For example, llama3.2 spinquant is using a prequantized checkpoitn hence it's not using the recipe for the regular fp32 checkpint. To support various combinations and optimizations, we are migrating to use `benchmark_configs` which is a set of predefined configs with combination of all optimizations that are possibly applied to the model, e.g. kv cache, embedding/activation quant, dtype, delegation, sdpa, etc.

In this PR, given a model (either a Hugging Face model ID or a in-tree model name) and a target platform ("android" vs "ios"), it's retrieving a list of supported benchmark configurations from the script `gather_benchmark_configs.py` and schedule the benchmark jobs accordingly. From the workflow dispatcher, users will just need to enter the model names, it will discover all supported benchmark configs for each model. Further more (not included in this PR), we can potentially expose `config_args` (key-value paris) from the script, if there is a way to store them in the DB and display in the dashboard. It will help understand how exactly a model is exported/lowered when discussing/debugging perf metrics.

Apple: https://github.com/pytorch/executorch/actions/runs/12404655922
Android: https://github.com/pytorch/executorch/actions/runs/12404651330

### DIFF
--- a/.ci/scripts/gather_benchmark_configs.py
+++ b/.ci/scripts/gather_benchmark_configs.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import json
+import logging
+import os
+import re
+from typing import Any, Dict
+
+from examples.models import MODEL_NAME_TO_MODEL
+
+
+# Device pools for AWS Device Farm
+DEVICE_POOLS = {
+    "apple_iphone_15": "arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/3b5acd2e-92e2-4778-b651-7726bafe129d",
+    "samsung_galaxy_s22": "arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/e59f866a-30aa-4aa1-87b7-4510e5820dfa",
+    "samsung_galaxy_s24": "arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/98f8788c-2e25-4a3c-8bb2-0d1e8897c0db",
+    "google_pixel_8_pro": "arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/d65096ab-900b-4521-be8b-a3619b69236a",
+}
+
+# Predefined benchmark configurations
+BENCHMARK_CONFIGS = {
+    "xplat": [
+        "xnnpack_q8",
+        "hf_xnnpack_fp32",
+        "llama3_fb16",
+        "llama3_spinquant",
+        "llama3_qlora",
+    ],
+    "android": [
+        "qnn_q8",
+        # TODO: Add support for llama3 htp
+        # "llama3_qnn_htp",
+    ],
+    "ios": [
+        "coreml_fp16",
+        "mps",
+        # TODO: Add support for llama3 ane
+        # "llama3_coreml_ane",
+    ],
+}
+
+
+def parse_args() -> Any:
+    """
+    Parse command-line arguments.
+
+    Returns:
+        argparse.Namespace: Parsed command-line arguments.
+
+    Example:
+        parse_args() -> Namespace(models=['mv3', 'meta-llama/Llama-3.2-1B-Instruct-QLORA_INT4_EO8'],
+                                   os='android',
+                                   devices=['samsung_galaxy_s22'])
+    """
+    from argparse import ArgumentParser
+
+    def comma_separated(value: str):
+        """
+        Parse a comma-separated string into a list.
+        """
+        return value.split(",")
+
+    parser = ArgumentParser("Gather all benchmark configs.")
+    parser.add_argument(
+        "--os",
+        type=str,
+        choices=["android", "ios"],
+        help="The target OS.",
+    )
+    parser.add_argument(
+        "--models",
+        type=comma_separated,  # Use the custom parser for comma-separated values
+        help=f"Comma-separated model IDs or names. Valid values include {MODEL_NAME_TO_MODEL}.",
+    )
+    parser.add_argument(
+        "--devices",
+        type=comma_separated,  # Use the custom parser for comma-separated values
+        help=f"Comma-separated device names. Available devices: {list(DEVICE_POOLS.keys())}",
+    )
+
+    return parser.parse_args()
+
+
+def set_output(name: str, val: Any) -> None:
+    """
+    Set the output value to be used by other GitHub jobs.
+
+    Args:
+        name (str): The name of the output variable.
+        val (Any): The value to set for the output variable.
+
+    Example:
+        set_output("benchmark_configs", {"include": [...]})
+    """
+
+    if os.getenv("GITHUB_OUTPUT"):
+        print(f"Setting {val} to GitHub output")
+        with open(str(os.getenv("GITHUB_OUTPUT")), "a") as env:
+            print(f"{name}={val}", file=env)
+    else:
+        print(f"::set-output name={name}::{val}")
+
+
+def is_valid_huggingface_model_id(model_name: str) -> bool:
+    """
+    Validate if the model name matches the pattern for HuggingFace model IDs.
+
+    Args:
+        model_name (str): The model name to validate.
+
+    Returns:
+        bool: True if the model name matches the valid pattern, False otherwise.
+
+    Example:
+        is_valid_huggingface_model_id('meta-llama/Llama-3.2-1B') -> True
+    """
+    pattern = r"^[a-zA-Z0-9-_]+/[a-zA-Z0-9-_.]+$"
+    return bool(re.match(pattern, model_name))
+
+
+def get_benchmark_configs() -> Dict[str, Dict]:
+    """
+    Gather benchmark configurations for a given set of models on the target operating system and devices.
+
+    Args:
+        None
+
+    Returns:
+        Dict[str, Dict]: A dictionary containing the benchmark configurations.
+
+    Example:
+        get_benchmark_configs() -> {
+            "include": [
+                {
+                    "model": "meta-llama/Llama-3.2-1B",
+                    "config": "llama3_qlora",
+                    "device_name": "apple_iphone_15",
+                    "device_arn": "arn:aws:..."
+                },
+                {
+                    "model": "mv3",
+                    "config": "xnnpack_q8",
+                    "device_name": "samsung_galaxy_s22",
+                    "device_arn": "arn:aws:..."
+                },
+                ...
+            ]
+        }
+    """
+    args = parse_args()
+    target_os = args.os
+    devices = args.devices
+    models = args.models
+
+    benchmark_configs = {"include": []}
+
+    for model_name in models:
+        configs = []
+        if is_valid_huggingface_model_id(model_name):
+            if model_name.startswith("meta-llama/"):
+                # LLaMA models
+                repo_name = model_name.split("meta-llama/")[1]
+                if "qlora" in repo_name.lower():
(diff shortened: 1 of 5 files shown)


guangy10:
All apple jobs are failing at "Setting up CI environment" step. The log doesn't have much useful info to tell if it's relevant. Scheduled one workflow to compare: https://github.com/pytorch/executorch/actions/runs/12387135036

.github/workflows/apple-perf.yml
```diff
@@ -181,23 +153,23 @@ jobs:
         GITHUB_RUNNER=1 PYTHON_EXECUTABLE=python ${CONDA_RUN} --no-capture-output \
           .ci/scripts/setup-macos.sh "${BUILD_TOOL}"
 
-        if [[ ${{ matrix.delegate }} == "coreml" ]]; then
+        if [[ ${{ matrix.benchmark_configs.config }} == *"coreml"* ]]; then
```
huydhn, line 156: This line and the one on line 161 looks wrong, you probably mean `${{ matrix.config }}` here.  The failure looks cryptic, but it tells something wrong in the script https://github.com/pytorch/executorch/actions/runs/12387012961/job/34575783485#step:9:8209

huydhn, line 156:
```suggestion
        if [[ ${{ matrix.config }} == *"coreml"* ]]; then
```

guangy10: Oops, forget to fix this

guangy10:
```
Dec 18 09:37:37 PDX647397983 kernel[0] <Notice>: memorystatus: killing process 602 [Benchmark] in high band FOREGROUND (100) - memorystatus_available_pages: 112541
Dec 18 09:37:37 PDX647397983 kernel[0] <Notice>: Benchmark[602] Corpse allowed 1 of 5
Dec 18 09:37:37 PDX647397983 backboardd(IOKit)[65] <Notice>: Connection removed: IOHIDEventSystemConnection uuid:2A3BA7CC-8C4A-496C-ACAA-32E7D50DC9B5 pid:602 process:Benchmark type:Passive entitlements:0x0 caller:BackBoardServices: <redacted> + 280 attributes:{
    HighFrequency = 1;
    bundleID = "org.pytorch.executorch.Benchmark";
    pid = 602;
} state:0x1 events:0 mask:0x0 dropped:0 dropStatus:0 droppedMask:0x0 lastDroppedTime:NONE
Dec 18 09:37:37 PDX647397983 kernel[0] <Notice>: 7347.894 memorystatus: killing_specific_process pid 602 [Benchmark] (per-process-limit 100) 3145733KB - memorystatus_available_pages: 112815
```

It looks like **Llama3.2 1B w/ fb16** is running around the 3GB RAM limits for iOS set by AWS.
  - OOMed: https://github.com/pytorch/executorch/actions/runs/12388301115/job/34607987766
  - Passed: https://github.com/pytorch/executorch/actions/runs/12386167313/job/34574116322
  
We will probably not include this path in the continuous run until the RAM limits is lifted.

guangy10:
The failure in [apple-perf / benchmark-on-device (llama, mps, arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d... / mobile-job (ios)](https://hud.pytorch.org/pr/pytorch/executorch/7349#34614070197) ([gh](https://github.com/pytorch/executorch/actions/runs/12399052782/job/34614070197)) is pre-existing. See details in https://github.com/pytorch/executorch/issues/7366

guangy10:
The android failure [android-perf / benchmark-on-device (llama, xnnpack_q8, arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2... / mobile-job (android)](https://hud.pytorch.org/pr/pytorch/executorch/7349#34616060368) ([gh](https://github.com/pytorch/executorch/actions/runs/12399052791/job/34616060368)) seem irrelevant to this PR since the [same model+config passed](https://github.com/pytorch/executorch/actions/runs/12386663978) earlier on this PR. But it's still important to figure out the root cause as it doesn't seem to be transient (retried and failed). @huydhn if you have any insights

huydhn:
> The android failure [android-perf / benchmark-on-device (llama, xnnpack_q8, arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2... / mobile-job (android)](https://hud.pytorch.org/pr/pytorch/executorch/7349#34616060368) ([gh](https://github.com/pytorch/executorch/actions/runs/12399052791/job/34616060368)) seem irrelevant to this PR since the [same model+config passed](https://github.com/pytorch/executorch/actions/runs/12386663978) earlier on this PR. But it's still important to figure out the root cause as it doesn't seem to be transient (retried and failed). @huydhn if you have any insights

I think the script returns the wrong device pool here for Android job.  From https://github.com/pytorch/executorch/actions/runs/12401286750/job/34620341884?pr=7349#step:4:35, it returned `arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/3b5acd2e-92e2-4778-b651-7726bafe129d` which is the iPhone device pool

.github/workflows/android-perf.yml
```diff
@@ -73,24 +53,27 @@ concurrency:
...
         shell: bash
         env:
           # Separate default values from the workflow dispatch. To ensure defaults are accessible
           # during scheduled runs and to provide flexibility for different defaults between
           # on-demand and periodic benchmarking.
-          CRON_DEFAULT_MODELS: ${{ github.event_name == 'schedule' && 'stories110M,dl3,mv3,mv2,ic4,ic3,vit' || 'stories110M' }}
-          CRON_DEFAULT_DEVICES: samsung_galaxy_s22
-          CRON_DEFAULT_DELEGATES: ${{ github.event_name == 'schedule' && 'xnnpack,qnn' || 'xnnpack' }}
+          CRON_DEFAULT_MODELS: ${{ github.event_name == 'schedule' && 'llama,mv3,mv2,ic4,ic3,resnet50,edsr,mobilebert,w2l,meta-llama/Llama-3.2-1B,meta-llama/Llama-3.2-1B-Instruct-SpinQuant_INT4_EO8,meta-llama/Llama-3.2-1B-Instruct-QLORA_INT4_EO8' || 'llama' }}
+          CRON_DEFAULT_DEVICES: apple_iphone_15
```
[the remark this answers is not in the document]
guangy10: What did I do!

guangy10:
> > The android failure [android-perf / benchmark-on-device (llama, xnnpack_q8, arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2... / mobile-job (android)](https://hud.pytorch.org/pr/pytorch/executorch/7349#34616060368) ([gh](https://github.com/pytorch/executorch/actions/runs/12399052791/job/34616060368)) seem irrelevant to this PR since the [same model+config passed](https://github.com/pytorch/executorch/actions/runs/12386663978) earlier on this PR. But it's still important to figure out the root cause as it doesn't seem to be transient (retried and failed). @huydhn if you have any insights
> 
> I think the script returns the wrong device pool here for Android job. From https://github.com/pytorch/executorch/actions/runs/12401286750/job/34620341884?pr=7349#step:4:35, it returned `arn:aws:devicefarm:us-west-2:308535385114:devicepool:02a2cf0f-6d9b-45ee-ba1a-a086587469e6/3b5acd2e-92e2-4778-b651-7726bafe129d` which is the iPhone device pool

Aha! I should probably return the device pool name as well, so that we can easily spot if it's using the wrong device pool

huydhn:
> We will probably not include this path in the continuous run until the RAM limits is lifted.

I'm still following up with AWS on this.  The confusing part is I'm not entirely sure if this is something from AWS (but let's see).  From what I read in https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.kernel.increased-memory-limit, this feels like a maybe and not a guarantee

>  If you use this entitlement, make sure your app still behaves correctly if additional memory isn’t available.

and 

> An increased memory limit is only available on some device models, call the os_proc_available_memory function to determine the amount of memory available

.github/workflows/android-perf.yml
```diff
@@ -171,9 +129,7 @@ jobs:
     needs: set-parameters
     secrets: inherit
     strategy:
-      matrix:
-          model: ${{ fromJson(needs.set-parameters.outputs.models) }}
-          delegate: ${{ fromJson(needs.set-parameters.outputs.delegates) }}
+      matrix: ${{ fromJson(needs.set-parameters.outputs.benchmark_configs) }}
```
guangy10, line 132: @huydhn do you know how I can customize the job title? It looks like it's automatically generated as part of the `matrix:` configuration. Currently it's displayed as "[export-models (meta-llama/Llama-3.2-1B, llama3_htp, arn:aws:devicefarm:us-west-2:308535385114:dev... / linux-job](https://github.com/pytorch/executorch/actions/runs/12401578767/job/34621284740#logs)", where the arn is not useful, I'd like to replace it with the human readable device pool name.

guangy10:
Core ML ANE job: https://github.com/pytorch/executorch/actions/runs/12402205998/job/34624007233

guangy10:
Qualcomm HTP job: https://github.com/pytorch/executorch/actions/runs/12402906809

.ci/scripts/gather_benchmark_configs.py
```diff
@@ -0,0 +1,204 @@
...
+    Args:
+        None
+
+    Returns:
+        Dict[str, Dict]: A dictionary containing the benchmark configurations.
+
+    Example:
+        get_benchmark_configs() -> {
+            "include": [
+                {"model": "meta-llama/Llama-3.2-1B", "benchmark_config": "hf_xnnpack_fp32", "device": "arn:aws:..."},
```
huydhn, line 137: What is the separator you are using to split the model name and the benchmark config?  I think we should use something unique , i.e. `__` (double underscore) or something similar so that the extract script can set them correctly later

.github/workflows/android-perf.yml
huydhn: AFAIK, there is no way to change this behavior from GitHub.  The name always include all the matrix dimensions.  Maybe there is a way to create a more human-friendly alias for the pool on AWS, let me take a look from that angle

huydhn (approved):

.ci/scripts/gather_benchmark_configs.py
guangy10: Can we use some more reliable than the file name? For example, can we store the benchmark config as it's serializable to json, then it's model name, benchmark_config (delegate, quant, dtype, etc.) can be safely retrieved.

.github/workflows/android-perf.yml
guangy10: Don't worry. No need to waste time on this. I can always include the device_name in the matrix

.ci/scripts/gather_benchmark_configs.py
huydhn: Yeah, being able to get away with parsing the filename is even better.  Let's have the benchmark config uploaded together with the export model and I think we can pick it up from there.

.github/workflows/android-perf.yml
guangy10: Added "device_name" and "device_arn" to the matrix, and put the "device_arn" at the end. Now we can see each job is titled with device pool name

guangy10:
@huydhn FYI, Core ML ANE and Qualcomm HTP jobs are not going to block this PR. I will merge this PR tonight to unblock you, and leave adding those paths separately if they are not working out of the box

huydhn:
> @huydhn FYI, Core ML ANE and Qualcomm HTP jobs are not going to block this PR. I will merge this PR tonight to unblock you, and leave adding those paths separately if they are not working out of the box

Sounds good!  Once this lands, I could start working on bring the benchmark config to the dashboard

guangy10:
Core ML ANE and Qualcomm HTP are not ready. Will enable in separate PRs. For now, we just hide these paths from `benchmark_configs`.

guangy10:
@kirklandsign FYI, the results of newly added llama3 benchmarkings (spintquant, qlora, bf16) are also missing from the benchmark_results.json. See the upload job here: https://github.com/pytorch/executorch/actions/runs/12404651330/job/34631545501

PR #7351, opened 2024-12-18
Revert "Add full operator to fold dq/q handling"

digantdesai:
This reverts recent commits from Arm because it is causing internal type checking errors.

@per - sorry about this. Let me work with you to remerge these ASAP.


iseeyuan (approved):
Thank you @digantdesai  for unblocking the internal diff train!

per:
@digantdesai - would be helpful to get logs or indication on where the pyre errors are. Running pyre with some default strict configuration gives the same results with or without the reverted commits. So completely in the dark here on how to approach this.

PR #7355, opened 2024-12-18
Support ceil_mode=True on maxpool2d in XNNPACK delegate

GregoryComer:
Summary: Support ceil_mode=True for MaxPool2d in XNNPACK delegate by conditionally updating padding. This works only when the input shape is static, as it requires statically computing the change in padding. As such, the partitioner constraints are also updated to reflect this.

Differential Revision: D67386151


mcr229 (approved):
This is smart, i like it

PR #7358, opened 2024-12-18
Support multiple batches (N > 1) in Arm runtime

AdrianLundell:
Extends testing of multiple batches on FVP for multiple operators. Note that currently not all operators are supported on compiler level.


AdrianLundell:
The failing tests should be marked as flaky since they are failing randomly without any related changes, looking into it.

freddan80:
Rebase pls

AdrianLundell:
We'll wait with this PR until the softmax failures are better understood

PR #7368, opened 2024-12-18
[2.6 RELEASE] Update PyTorch pin to 2.6

SS-JIA:
### Summary

Title says it all.

### Test plan

CI


SS-JIA:
All pass expect the arm test, which is to be expected. Let's merge the fix in as a later update when it is available.

lucylq (approved):

PR #7400, opened 2024-12-19
Fix torch._inductor.aoti_compile_and_package input

yushangdi:
Summary:
The inputs to `torch._inductor.aoti_compile_and_package` changed in https://github.com/pytorch/pytorch/pull/140991.

The `args` and `kwargs` do not need to be inputed anymore. The API will get them from `exported_program.example_inputs`

Differential Revision: D67436429


jackzhxng (approved):

.ci/docker/ci_commit_pins/pytorch.txt
```diff
@@ -1 +1 @@
-19eff28ff3f19b50da46f5a9ff5f4d4d213806fe
+288aa873831057b1eb7d747914ec4fdc76c23a80
```
dvorjackz, line 1: Should update this and the torch pin in a separate pr

jackzhxng (changes requested):

.ci/docker/ci_commit_pins/pytorch.txt
yushangdi: @dvorjackz but if I update the torch pin without other changes in the PR, the tests will fail due to the input to  torch._inductor.aoti_compile_and_package changed?

dvorjackz: True actually, let's use this instead though actually https://github.com/pytorch/executorch/pull/7273/, seems better for the sake of updating the pin correctly

PR #7413, opened 2024-12-20
 [ExecuTorch][BE] Split kv cache and SDPA for better code sharing

kimishpatel:
* __->__ #7413

Summary:

Why?
We have coupled SDPA with kv cache for a while. Initially this was done
as we implemented sdpa_with_kv_cache custom op to reduce multiple copy
overheads from kv cache update. (This could have been done by having
separate custom kv cache update and custom sdpa op. Recent changes
enabled this.)
As a result of SDPA module owning kv cache, we get a) non-composable
implementation and b) harder to reuse model definition and components
from repos like tune. Output of this is that we have multiple definition
of the same model, llama, lying around in ET, TorchChat and Tune. This
diff and subsequent ones will try to move in the direction where custom
kv cache and custom sdpa become decoupled and composable, making it more
module-swap friendly with tune's model definition.

How.
Earlier PRs decoupled kv cache update from sdpa. So now
1. Decouple SDPA nn.Module from KV cache.
2. Standardize on KVCache and SDPA interface. That is KVCache and SDPA
   both operate on q, k, v in [B, # heads, seq_len, head_dim] formatted
   tensors.
3. 2 will introduce multiple tranposes when KVCache and SDPA are
   replaced by custom modules, but we will write graph pass to undo
   those.

Test Plan:
Existing tests.
Make sure perf doesnt regress

Differential Revision: [D67914054](https://our.internmc.facebook.com/intern/diff/D67914054)


examples/models/llama/source_transformation/sdpa.py
```diff
@@ -47,20 +37,21 @@ def forward(
         seqlen,
         mask,
     ):
+        q = q.transpose(1, 2)  # (bs, seqlen, n_local_heads, head_dim)
```
dvorjackz, line 40: I just thought about it again, and adding this transpose here and also before in the `llama_transformer.py` so that we can share code for kv_cache.py (this is the reason right?) doesn't really make sense since we are using a custom export-friendly KV cache already anyways: https://github.com/pytorch/executorch/blob/main/extension/llm/modules/kv_cache.py#L13

extension/llm/export/builder.py
```diff
@@ -212,6 +215,13 @@ def export(self) -> "LLMEdgeManager":
 
         return self
 
+    def run_canonical_optimizations(self):
```
larryliu0820, line 218: Please also add checks to make sure `self.pre_autograd_graph_module` is not None, basically this needs to be run after `export()`.

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

examples/models/llama/source_transformation/sdpa.py
kimishpatel: So this is really standardizing what the SDPA api is. That is the input tensors q, k and v are in [bs, num heads, seq len, head dim]. This is the standard API. This actually applies to output as well, but that part i havent yet fixed.

So now imagine you did not use kv cache but just wanted to replace SDPA. This allows you to do that.

Regarding KV cache itself, the `transpose_cache` argument is also relic from the time of using sdpa_with_kv_cache. We dont really need tranpose_cache arg at all. Removing that will bring it closer to tune's kv cache impl. In fact I would like us to not have to use own own kv cache.

dvorjackz: Sure, I see you are dealing with the paired transposes later anyway. Also extra bit of info is that from my own benchmarking, these transposes here don't seem to be much overhead

extension/llm/export/export_passes.py
```diff
@@ -0,0 +1,97 @@
...
+    ndim = tensor.ndim
+    if dim_0 < 0:
+        dim_0 = ndim + dim_0
+    if dim_1 < 0:
+        dim_1 = ndim + dim_1
+    assert dim_0 < ndim and dim_1 < ndim, f"Invalid dimensions: {dim_0}, {dim_1}"
+    return dim_0, dim_1
+
+
+class RemoveRedundantTransposes(ExportPass):
```
dvorjackz, line 22: Would name something else, since this is also undoing permutes as well

jackzhxng (approved):
Looks good 👍🏻  pending comments + testing

tarun292 (approved):

extension/llm/export/export_passes.py
```diff
@@ -0,0 +1,97 @@
...
+
+                        # Check if the two permutes undo each other
+                        if dim_list == user_dim_list:
+                            graph_changed = True
+                            user.replace_all_uses_with(node.args[0])
+
+        graph_module.graph.eliminate_dead_code()
+        graph_module.recompile()
+
+        return PassResult(graph_module, graph_changed)
```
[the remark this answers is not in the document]
tarun292: https://github.com/pytorch/executorch/blob/main/exir/passes/__init__.py#L477

extension/llm/export/export_passes.py
kimishpatel: Yeah good point. I should rename it to permute as transpose is special case of permute as well

dvorjackz:
Update: synced offline and wanted to put the notes from the sync somewhere:
- TorchTune has a focus on server-side finetuning, we will use their model definitions as a starting point for when new models are released to boostrap development
  - As such, updates to TorchTune don't necessarily need to pulled in since usually they will have nothing to do with inference performance or necessarily guarantee exportability. i.e. if there are few new commits on attention, it is not necessary that we update the TorchTune pin in order to maintain parity
- We will keep separate custom copies of inference-optimized and exportable versions of major bottleneck modules such as attention, sdpa, kv_cache, which we will module swap in
- While this PR helps overall with code reuse since it lets us source swap in sdpa (although in TorchTune atm the sdpa is a callable), there are no guarantees provided moving forward that the sdpa will be swappable, for example if it moves out of the callable and the bare `nn.functional` sdpa is in the attention module itself. Then we can't source swap sdpa anymore, it's better to just swap on the attention level.

cc @kimishpatel @tarun292

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

extension/llm/export/export_passes.py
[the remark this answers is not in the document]
kimishpatel: So for now I am gonna add it here in next PR. I need to keep this here just to not break any backend delegation stuff for now.

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

kimishpatel:
@kimishpatel has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D67914054).

PR #7415, opened 2024-12-20
Add Arm conformer unit test

freddan80:
cc @digantdesai @per @zingo @oscarandersson8218

### DIFF
--- a/backends/arm/test/models/test_conformer.py
+++ b/backends/arm/test/models/test_conformer.py
@@ -1,0 +1,126 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import logging
+import unittest
+
+import torch
+from executorch.backends.arm.test import common, conftest
+
+from executorch.backends.arm.test.tester.arm_tester import ArmTester
+
+from torchaudio.models import Conformer
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class TestConformer(unittest.TestCase):
+    """Tests Torchaudio Conformer"""
+
+    # Adjust nbr below as we increase op support. Note: most of the delegates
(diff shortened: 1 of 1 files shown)


zingo (approved):

digantdesai:
Love it ❤️

digantdesai:
cc @limintang @3l1

backends/arm/test/models/test_conformer.py
```diff
@@ -0,0 +1,126 @@
...
+logger.setLevel(logging.INFO)
+
+
+class TestConformer(unittest.TestCase):
+    """Tests Torchaudio Conformer"""
+
+    # Adjust nbr below as we increase op support. Note: most of the delegates
+    # calls are directly consecutive to each other in the .pte. The reason
+    # for that is some assert ops are removed by passes in the
+    # .to_executorch step, i.e. after Arm partitioner.
```
digantdesai, line 27: > The reason for that is some assert ops are removed by passes in the .to_executorch step

One option is to partition them and delete them inside arm::delegate::preprocess()

backends/arm/test/models/test_conformer.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class TestConformer(unittest.TestCase):
+    """Tests Torchaudio Conformer"""
+
+    # Adjust nbr below as we increase op support. Note: most of the delegates
```
digantdesai, line 24: 👍

freddan80:
> cc @limintang @3l1

Just FYI. This is the TorchAudio Conformer. It doesn't fully delegate as the [sooftware conformer](https://github.com/sooftware/conformer) does.

digantdesai:
@freddan80  - why shouldn't we add https://github.com/sooftware/conformer in the CI? Since it delegates better than torchaudio? I remember you told me but can't remember why :p (may be because we prefer torchaudio?)

PR #7427, opened 2024-12-23
Add TagUnquantizedNodesPass

Aleksei-grovety:
For models with operations that are not quantized, this pass keeps unquantized operators on the CPU. For example, the [deit-tiny-patch16-224](https://huggingface.co/facebook/deit-tiny-patch16-224) network has an unquantized [scaled_dot_product_attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) operation. When compiling to Vela, invalid argument errors occur because unquantized operations are offloaded to the NPU. This pass is designed to solve this problem.

### DIFF
--- a/backends/arm/_passes/tag_unquantized_nodes_pass.py
+++ b/backends/arm/_passes/tag_unquantized_nodes_pass.py
@@ -1,0 +1,48 @@
+# Copyright 2024 Arm Limited and/or its affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-unsafe
+
+import torch
+from executorch.backends.arm.tosa_quant_utils import dq_q_ops, get_neighbour_quant_args
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class TagUnquantizedNodesPass(ExportPass):
+    """
+    Pass run before partitioning to tag unquantized nodes
+    to ensure we don't greedily partition them for device. Unquantized operations must remain on the CPU.
+    """
+
+    def is_node_quantized(self, node: torch.fx.Node) -> bool:
+        user_q_args, input_q_args = get_neighbour_quant_args(node)
+
+        # If there are no neighboring quantized nodes, then this node is not quantized except for constants,
+        # they can only have a dequantization node.
+        if (
+            len(node.all_input_nodes) > 0
+            and len(input_q_args) == 0
+            or len(user_q_args) == 0
+        ):
+            return False
+
+        return True
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        for node in graph_module.graph.nodes:
+            # Look through operations that are not quantization or dequantization
+            if node.op == "call_function" and node.target not in dq_q_ops:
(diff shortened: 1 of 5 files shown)


oscarandersson8218:
Thanks for your contribution @Aleksei-grovety! Just for clarification, do you prefer ScaledDotProductAttention (SDPA) to run on the CPU in FP32 for precision reasons, or would you prefer to run it on the NPU?

If you prefer to run it on the NPU, a different approach to this would be to decompose SDPA before quantization. An SDPA-decomposition exists in https://github.com/pytorch/executorch/blob/main/backends/transforms/decompose_sdpa.py. You can add pre-quantization decompositions to https://github.com/pytorch/executorch/blob/e2afedf935c94d7703ae964b786c87572cc6d326/backends/arm/_passes/arm_pass_manager.py#L144

**Note**: doing this might require some additional changes to https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/decompose_softmaxes_pass.py

Aleksei-grovety:
It is preferable to run SDPA on the NPU. I will consider an approach with decomposition of the operation before quantization.

What do you think of the approach of taking non-quantized operations to the CPU? Or should we prioritize a full offloading approach to the NPU?

In addition to the SDPA operation in the deit-tiny-patch16-224 network, there is also the GELU activation, which is not implemented on the NPU. Is it correct that this activation can be realized using a LUT table?

oscarandersson8218:
> What do you think of the approach of taking non-quantized operations to the CPU? Or should we prioritize a full offloading approach to the NPU?

We prioritize full offloading to NPU. In some cases it may be necessary or beneficial to run an operator on the CPU, however that should be determined by TOSASupportedOperators: https://github.com/pytorch/executorch/blob/24f0d34dcddf7e050b601b55e3e310f72568399a/backends/arm/operator_support/tosa_supported_operators.py#L69
The reasoning is that if an operator is not quantized, it should not be consumed by the delegate (there should be no need to explicitly tag nodes). I think the most common reason for these kinds of issues is that a pattern of `dq -> op -> q` gets decomposed to `dq -> op_0 -> op_1 -> ... -> op_n -> q` when converting the model to edge dialect. The way we usually tackle these issues is to apply the decomposition before quantization so that each op will have a dq-q pair around it. Alternatively you can add the aten op to https://github.com/pytorch/executorch/blob/85d274a3d20b6aae43372f2bec40c5f43535775c/backends/arm/arm_partitioner.py#L116 and do the edge decomposition yourself. If you encounter these issues, feel free to raise an issue. Also, if you would like to have a way to specify which operators to not delegate from a top level, I suggest you create an issue for that. 
 
 
> In addition to the SDPA operation in the deit-tiny-patch16-224 network, there is also the GELU activation, which is not implemented on the NPU. Is it correct that this activation can be realized using a LUT table?

That's correct. It should be possible to lower a quantized GELU with a simple modification of this pass: https://github.com/pytorch/executorch/blob/24f0d34dcddf7e050b601b55e3e310f72568399a/backends/arm/_passes/insert_table_ops.py#L30

oscarandersson8218:
Hi @Aleksei-grovety, just checking how things are going with this? Let us know if you need more guidance or if something is unclear.

Issue #2871, opened 2024-04-05
Error in Pip Install Editable Mode


Jerry-Ge:
When trying to run the following command to install in editable mode: 
`pip install --editable . --config-settings editable_mode=strict --no-build-isolation
`

Hit the following error: 
```
Building wheels for collected packages: executorch
  Building editable for executorch (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building editable for executorch (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [144 lines of output]
      [04/05/24 15:57:12] WARNING  pyproject.toml does not contain a setuptools.py:119
                                   tool.setuptools_scm section
      running editable_wheel
      creating /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info
      writing /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/PKG-INFO
      writing dependency_links to /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/dependency_links.txt
      writing entry points to /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/entry_points.txt
      writing requirements to /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/requires.txt
      writing top-level names to /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/top_level.txt
      writing manifest file '/tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/SOURCES.txt'
      adding license file 'LICENSE'
      writing manifest file '/tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch.egg-info/SOURCES.txt'
      creating '/tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch-0.1.0.dist-info'
      creating /tmp/pip-wheel-91gm0nbw/.tmp-sdag809b/executorch-0.1.0.dist-info/WHEEL
      running build_py
      creating /tmp/tmprwpw5zbw.build-lib/executorch
      creating /tmp/tmprwpw5zbw.build-lib/executorch/exir
      creating /tmp/tmprwpw5zbw.build-lib/executorch/exir/_serialize
      copying schema/scalar_type.fbs -> /tmp/tmprwpw5zbw.build-lib/executorch/exir/_serialize
[...]

PR #7443, opened 2024-12-26
[DevX] Enable editable mode

larryliu0820:
### Summary

Partially fixing #2871

Enable `pip install -e .` with local build flatc and custom ops AOT lib disabled.

### Test plan
```
EXECUTORCH_BUILD_FLATC=OFF EXECUTORCH_BUILD_KERNELS_CUSTOM_AOT=OFF pip install -e .
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -216,8 +216,13 @@ class _BaseExtension(Extension):
                 file.
         """
         # Share the cmake-out location with CustomBuild.
-        cmake_cache_dir = Path(installer.get_finalized_command("build").cmake_cache_dir)
-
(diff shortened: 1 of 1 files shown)


iseeyuan (approved):

iseeyuan:
In pip install without local, cmake build first and copy the binaries into site packages. If editable, for some reason there's no cmake build (CustomBuild), and InstallerBuildExt would error. One option is to refactor and combine CustomBuild and InstallerBuildExt.

If there's flatc in the env, no need to build flatc. Custom ops can be disabled if it's not used.

PR #7449, opened 2024-12-27
Fix NATIVE_FP16 macro in our fork of half.h

swolchok:
Simple build unbreak for Raspberry Pi 5 -- fp16 is an optional aarch64 feature.

### DIFF
--- a/runtime/core/portable_type/half.h
+++ b/runtime/core/portable_type/half.h
@@ -15,7 +15,7 @@
 #include <ostream>
 
 #if defined(__GNUC__) || defined(__clang__)
-#if defined(__aarch64__)
+#if defined(__aarch64__) && defined(__ARM_FEATURE_FP16_SCALAR_ARITHMETIC)
 #ifndef __ARM_V8_ONLY__
 #define NATIVE_FP16 1
 #endif // __ARM_V8_ONLY__


runtime/core/portable_type/half.h
```diff
@@ -15,7 +15,7 @@
 #include <ostream>
 
 #if defined(__GNUC__) || defined(__clang__)
-#if defined(__aarch64__)
+#if defined(__aarch64__) && defined(__ARM_FEATURE_FP16)
```
mergennachin, line 18: Is that the correct definition actually? 

I can see __ARM_FEATURE_FP16_SCALAR_ARITHMETIC, __ARM_FEATURE_FP16_FML and __ARM_FEATURE_FP16_VECTOR_ARITHMETIC in https://developer.arm.com/documentation/101028/0012/5--Feature-test-macros

But I don't see a standalone __ARM_FEATURE_FP16 macros in the document.

mergennachin: If it's actually a typo, it would be good to have a regression test somewhere to catch these types of errors.

kimishpatel (approved):

kimishpatel (approved):

runtime/core/portable_type/half.h
swolchok: > regression test

I guess we could disassemble half_test.o and make sure it contains the relevant aarch64 instructions. Should probably just do that in upstream PyTorch though; after #7040 is finally ready to go, this file will be powered by c10/util/Half.h anyway.

swolchok: I probably wanted `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC`, for the record.

swolchok: I manually disassembled half_test.cpp.o and confirmed that 1) it wasn't the same as baseline before I fixed the defined check to `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC` 2) it is now the same as baseline

PR #7512, opened 2025-01-04
Fixed always rebuild issue in cmake.

### LINKED ISSUE #7511: CMake generated incremental builds, rebuild more often than required.
## Issue:

So when building executorch, when running ninja/make, the rebuild builds more than required, this annoys me.

Specifically:
1. The first time call ninja-> builds everything, yay.
2. immediately call ninja again -> rebuilds 2 files.
3. immediately call ninja again -> rebuilds 12 files. 
(repeats 2-3 ad nauseum)

## Cause:

This is caused by when cmake sets up the generation of the schema headers. It generates the file: include/executorch/schema/program_generated.h

However, it expects to generate:
include/executorch/program_generated.h

When ninja/make goes to check for the file: 'include/executorch/program_generated.h' it does not exist, so it regenerates the 
file 'include/executorch/schema/program_generated.h'. 

Then the next time CMake is run, ninja sees that program_generated.h has been updated, so it regenerates the 12 other dependent files.

## Fix (In associated pull request) 

Tell CMake to look for the files that it actually generated.

## Possible drawbacks?

So this looks like someone updated the include directory for the schema headers to be in the schema subdirectory, and forgot/did not know to update the CMake dependency tracking. However, if this change occurred, it happened before the open source git history. 

I suppose that it is possible that there is some weird edge-case that requires that the schema headers are always regenerated...

cptspacemanspiff:
### Summary
Fixes #7511 

CMake generates different files than expected, thus the incremental builds always rebuild.

The generated files were located in include/executorch/schema/program_generated.h CMake was expecting files in include/executorch/program_generated.h

This causes cmake to always regenerate the files, as the expected artifacts are never there.
 
Presumably this was a change at some point, and the expected output from CMake never got updated.

### Test plan
Fixes the build issue, but I only tested it with ninja single config generator, behavior should be the same on all others though.

### DIFF
--- a/schema/CMakeLists.txt
+++ b/schema/CMakeLists.txt
@@ -15,7 +15,7 @@ endif()
 
 # The include directory that will contain the generated schema headers.
 set(_program_schema__include_dir "${CMAKE_BINARY_DIR}/schema/include")
-
+set(_program_schema__output_dir "${_program_schema__include_dir}/executorch/schema")
 # Source root directory for executorch.
 if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/..)
@@ -26,16 +26,16 @@ function(generate_program_schema _schema_srcs _schema_name)
   foreach(fbs_file ${_schema_srcs})
     string(REGEX REPLACE "[.]fbs$" "_generated.h" generated "${fbs_file}")
     list(APPEND _schema_outputs
-         "${_program_schema__include_dir}/executorch/schema/${generated}"
+         "${_program_schema__output_dir}/${generated}"
     )
   endforeach()
-  
+
   # Generate the headers from the .fbs files.
   add_custom_command(
     OUTPUT ${_schema_outputs}
     COMMAND
       ${FLATC_EXECUTABLE} --cpp --cpp-std c++11 --gen-mutable --scoped-enums -o
-      "${_program_schema__include_dir}/executorch/schema" ${_schema_srcs}
+      "${_program_schema__output_dir}" ${_schema_srcs}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     DEPENDS ${FLATC_EXECUTABLE} ${_schema_srcs}
     COMMENT "Generating ${_schema_name} headers"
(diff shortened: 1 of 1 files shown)


lucylq (commented):
Thanks for making the changes and for the detailed write-up! This looks good to me. 

cc @dbort

schema/CMakeLists.txt
```diff
@@ -26,10 +26,10 @@ function(generate_program_schema _schema_srcs _schema_name)
   foreach(fbs_file ${_schema_srcs})
     string(REGEX REPLACE "[.]fbs$" "_generated.h" generated "${fbs_file}")
     list(APPEND _schema_outputs
-         "${_program_schema__include_dir}/executorch/${generated}"
+         "${_program_schema__include_dir}/executorch/schema/${generated}"
     )
   endforeach()
-
+
```
dbort, line 32: Please remove these stray trailing whitespace characters

schema/CMakeLists.txt
```diff
@@ -26,10 +26,10 @@ function(generate_program_schema _schema_srcs _schema_name)
   foreach(fbs_file ${_schema_srcs})
     string(REGEX REPLACE "[.]fbs$" "_generated.h" generated "${fbs_file}")
     list(APPEND _schema_outputs
-         "${_program_schema__include_dir}/executorch/${generated}"
+         "${_program_schema__include_dir}/executorch/schema/${generated}"
```
dbort, line 29: The underlying problem was that this path didn't match the path in the custom_command. To make it more clear that they need to agree, and to avoid mismatches in the future, consider using a common local variable for both of them, like
```
set(_include_dir "${_program_schema__include_dir}/executorch/schema")
```
and then
```
list(APPEND _schema_outputs "${_include_dir}/${generated}")
```
here and again using `_include_dir` on line 38.

dbort (approved):
Thanks for fixing this! The outputs were definitely out of sync with the custom_command.

schema/CMakeLists.txt
cptspacemanspiff: oops, removed that.

schema/CMakeLists.txt
cptspacemanspiff: Changed to this, named it _program_schema__output_dir to be a little more clear.

dbort:
Thanks for the changes, and for finding/fixing this; looks great!

PR #7520, opened 2025-01-06
Change status to beta in pyproject.toml file

mergennachin:
### Summary
Looks like we never updated this value for 0.4 release. 

Will do this change and cherry-pick onto 0.5 instead.

### Test plan

Inspection


dbort (approved):
Good catch!

dbort:
Worth picking this into the v0.5.0 release

PR #7536, opened 2025-01-07
[DRAFT] Use mask instead of cond for attention conditional logic

jackzhxng:
### Summary
Use masking to avoid resorting to `torch.cond`, which prevents us from mutating inside branches thereby forcing us to clone the kv cache and create lots of unnecessary copies.

Also gets past the current limitation that the partitioners don't automatically recursively partition conditional subgraphs, allowing us to directly partition Llama 3.2 MM with XNNPack.

#### Llama 3.2 MM comparison against XNNPack + KV cache + custom SDPA
| **Metric** | **Before** | **After** |
| --- | --- | --- |
| Activations | 0.95 GB | **0.27 GB** |
| .pte size | 60 GB | **30 GB** |
| Prefill | 20.91 s | **0.86 s** |
| Generation | 0.28 tok/s | **6.26 tok/s** |

### Test plan
Rely on existing regression tests ([test_attention](https://github.com/pytorch/executorch/blob/main/extension/llm/modules/test/test_attention.py) and [test_kv_cache](https://github.com/pytorch/executorch/pull/7279/files#diff-a5fcc5119e8382e6f3fcf84b47014736440253c9c56216a617b40de4b750cc2c) - about to be merged), which have adequate coverage over kv cache and multi-head attention edge cases.

### DIFF
--- a/extension/llm/modules/attention.py
+++ b/extension/llm/modules/attention.py
@@ -246,6 +246,7 @@ class MultiHeadAttention(nn.Module):
         # x has shape [b, s_x, d]
         # y has shape [b, s_y, d]
         b, s_x, _ = x.shape
+        s_y = y.shape[1] if y is not None else 0
 
         # q has shape [b, s_x, num_heads * head_dim]
         q = self.q_proj(x)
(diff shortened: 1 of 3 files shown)


extension/llm/modules/attention.py
```diff
@@ -280,37 +288,12 @@ def calculate_kv(y):
...
-            )
             # Update key-value cache
-            self.kv_cache.k_cache.copy_(k)
-            self.kv_cache.v_cache.copy_(v)
-            self.kv_cache.cache_pos.copy_(cache_pos)
+            if self.kv_cache is not None and self.cache_enabled:
+                k, v = self.kv_cache.update(k, v)
 
-        output = self._sdpa(q, k, v, b, s_x, mask=mask)
+        output = self._sdpa(q, k, v, b, s_x)
```
dvorjackz, line 296: Note to self: the refactor removed the mask arg, make sure to add it back

PR #7540, opened 2025-01-07
Refactor CadenceQuantizer

mcremon-meta:
Summary:
The current class structure is hard to cleanly extend. This diff:
- Makes `CadenceQuantizer` a base class
- Creates a `CadenceDefaultQuantizer` that is exactly the same as the previous `CadenceQuantizer` class
- Removes the qconfig from the `CadenceQuantizer`, since it really belongs to the `CadenceAtenQuantizer` (it is defined per op)
- Makes both the default qconfig and the default quantizer list module level variables

Using this structure will make it much cleaner to add new quantizers in the future.

Differential Revision: D67645196

### DIFF
--- a/backends/cadence/aot/quantizer/quantizer.py
+++ b/backends/cadence/aot/quantizer/quantizer.py
@@ -60,6 +60,13 @@ wgt_qspec = QuantizationSpec(
 
 bias_qspec: Optional[QuantizationSpec] = None
 
+_default_qconfig = QuantizationConfig(
+    act_qspec,
+    act_qspec,
+    wgt_qspec,
+    None,
+)
+
 
 class CadenceAtenQuantizer(Quantizer):
     def __init__(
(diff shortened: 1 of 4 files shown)


backends/cadence/aot/quantizer/quantizer.py
```diff
@@ -60,6 +60,13 @@
 
 bias_qspec: Optional[QuantizationSpec] = None
 
+_default_qconfig = QuantizationConfig(
+    act_qspec,
+    act_qspec,
+    wgt_qspec,
+    None,
```
zonglinpeng, line 67: why do we need None as a default?

backends/cadence/aot/quantizer/quantizer.py
```diff
@@ -140,31 +147,39 @@ def get_supported_operators(cls) -> List[OperatorConfig]:
...
-                CadenceAtenQuantizer(ReluPattern1(), static_qconfig),
-            ]
-        )
+    def __init__(self, quantizers: List[Quantizer]) -> None:
+        super().__init__(quantizers)
+
+
+class CadenceDefaultQuantizer(CadenceQuantizer):
+    """
+    Default quantizer for Cadence backend, defined in fbcode/executorch/backends/cadence/aot/quantizer/quantizer.py
```
zonglinpeng, line 178: Nit: lint to break into two lines

zonglinpeng (approved):

PR #7541, opened 2025-01-07
Arm unittest refactor of Add and Conv2D test cases

AdrianLundell:
This is a showcase of multiple improvements which can be done across all tests:
1. Move pipeline definition from tests to a general (flexible) pipeline class
  - Define a default step of stages using e.g. TosaPipelineBI()
  - Add custom config or debug stages using helper functions s.a. .change_args(), .add_stage(), .dump_artifact() etc.
  - Run the full pipeline using .run()
2. Move towards a pure pytest approach to remove dependencies on unittest and parametrize
3. Separate tests running on FVP from tests not running on FVP rather than configuring this from the command line
  - FVP tests are skipped if not installed
  - To filter out tests one may instead use pytest markers/name filtering
  - This should give a clearer picture of what has been tested
4. Introduces one favored way of marking tests as xfails, in the parameterize decorator


AdrianLundell:
@digantdesai Curious to see what you think about this, in particular this removes unittest support in favour of a simpler pytest-only setup + it introduces one extra layer of complexity by wrapping test stages in the new test_pipeline.

backends/arm/test/common.py
```diff
@@ -185,3 +189,41 @@ def get_target_board(compile_spec: list[CompileSpec]) -> str | None:
             elif "u85" in flags:
                 return "corstone-320"
     return None
+
+
+u55_fvp_mark = pytest.mark.skipif(
```
digantdesai, line 194: Nit: Make it more intuitive
```suggestion
SkipIfNoCS300 = pytest.mark.skipif(
```

backends/arm/test/common.py
```diff
@@ -185,3 +189,41 @@ def get_target_board(compile_spec: list[CompileSpec]) -> str | None:
...
+""" Marks a test as running on Ethos-U85 FVP, e.g. Corstone 320. Skips the test if this is not installed."""
+
+
+def parametrize(
+    arg_name: str, test_data: dict[str, Any], xfails: dict[str, str] = None
+):
+    """
+    Custom version of pytest.mark.parametrize with some syntatic sugar and added xfail functionality
+        - test_data is expected as a dict of (id, test_data) pairs
+        - alllows to specifiy a dict of (id, failure_reason) pairs to mark specific tests as xfail
```
digantdesai, line 211: If these two dicts are meant to be looked up using same keys then, perhaps,
```
test_data: dict[str, Tuple[Any, str]] 
```
where key is a string, value is a tuple(any_test_data, failure_reason)

backends/arm/test/common.py
```diff
@@ -185,3 +189,41 @@ def get_target_board(compile_spec: list[CompileSpec]) -> str | None:
...
+    """
+    if xfails is None:
+        xfails = {}
+
+    def decorator_func(func):
```
digantdesai, line 216: Add a docstring.

backends/arm/test/tester/test_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+from executorch.backends.arm.test.tester.arm_tester import ArmTester
+from executorch.exir.backend.compile_spec_schema import CompileSpec
+
+
+logger = logging.getLogger(__name__)
+T = TypeVar("T")
+""" Generic type used for test data in the pipeline. Depends on which type the operator expects."""
+
+
+class BasePipeline(Generic[T]):
```
digantdesai, line 20: Nit: To help improve readability

```suggestion
class BasePipelineMaker(Generic[T]):
```

if you are not a fan for `BasePipelineFactory` or something along those lines :)

backends/arm/test/tester/test_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+            self.add_stage(-1, self.tester.partition)
+        self.add_stage(-1, self.tester.check_not, self.exir_ops)
+        self.add_stage(
+            -1,
+            self.tester.check_count,
+            {"torch.ops.higher_order.executorch_call_delegate": 1},
+        )
+        self.add_stage(-1, self.tester.to_executorch)
+
+    def add_stage(self, pos: int, func: Callable, *args, **kwargs):
```
digantdesai, line 87: Add logging, this meta debugging can get hard. 🙏

backends/arm/test/tester/test_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+        for stage in self._stages:
+            try:
+                stage()
+            except Exception as e:
+                logger.error(f"\nFailure in stage <{stage.id}>: \n   {str(e)}")
+                raise e
+
+
+class TosaPipelineBI(BasePipeline, Generic[T]):
+    def __init__(
```
digantdesai, line 142: doc string

backends/arm/test/tester/test_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+        )
+        self.add_stage_after(
+            "quantize",
+            self.tester.check,
+            ["torch.ops.quantized_decomposed.quantize_per_tensor.default"],
+        )
+
+
+class TosaPipelineMI(BasePipeline, Generic[T]):
+    def __init__(
```
digantdesai, line 176: docstring for the class

backends/arm/test/common.py
digantdesai: If you want you can take it even more fine grained and allow individual test_data entry failure reason.

backends/arm/test/tester/test_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+                self.kwargs = kwargs
+            else:
+                raise RuntimeError(f"{self.id} args updated after being called.")
+
+    def __init__(
+        self,
+        module: torch.nn.Module,
+        test_data: T,
+        aten_ops: str | List[str],
+        exir_ops: str | List[str],
```
digantdesai, line 56: can this be looked up in a table? OK if you want to pass it in

backends/arm/test/ops/test_add.py
```diff
@@ -1,178 +1,144 @@
...
+exir_op = "executorch_exir_dialects_edge__ops_aten_add_Tensor"
+
+test_data = {
+    "5d_float": (torch.FloatTensor([1, 2, 3, 5, 7]),),
+    "1d_ones": ((3 * torch.ones(8),)),
+    "1d_randn": (10 * torch.randn(8),),
+    "4d_ones_1": (torch.ones(1, 1, 4, 4),),
+    "4d_ones_2": (torch.ones(1, 3, 4, 2),),
+}
+T1 = Tuple[torch.Tensor]
```
digantdesai, line 30: seems more pain than worth IMHO, what are your thoughts on these typing hints?

backends/arm/test/ops/test_conv2d.py
```diff
@@ -228,118 +231,72 @@ def forward(self, x):
...
+
+@common.parametrize("test_module", test_modules)
+def test_conv2d_u85_BI(test_module):
+    pipeline = EthosU85PipelineBI[T1](
+        test_module, test_module.get_inputs(), aten_op, exir_op, run_on_fvp=False
+    )
+    pipeline.run()
+
+
+xfails = {
```
digantdesai, line 281: these should live close to these keys in the modules dict IMO

digantdesai (commented):
I love the new approach. It's easier to add tests, cleaner, and reduces code duplication.

However, I have a concern that this might lead to "blind copy-paste" of new tests without thoroughly thinking about the test cases, especially from new users. Over time, we may end up with solid coverage for things that are already covered in the pipeline generator, but leave gaps elsewhere. For example, consider the conv test from your PR. When adding a module with two_conv, if I were to manually add a test case, in edge IR after to_edge, I would check for two convolution nodes to be present. Although it's possible to add a stage in the pipeline object, someone more likely to just do what you did.

Originally, the tester was designed to address code duplication and redundancy, without taking away too much control. 

That said, at this point, my concerns might not be too high, so I think we can proceed with this approach. We can evaluate when to use this more vigilantly.

backends/arm/test/common.py
AdrianLundell: Done

backends/arm/test/common.py
AdrianLundell: I think I prefer this way, what we often want to do is to reuse the same test data for multiple targets where only one is failing. Here we avoid having to create separate dicts for this, by specifying a unique xfail-dict just for one target.

backends/arm/test/common.py
AdrianLundell: I didn't understand exactly where this docstring should be added, there is one right above. Should it be moved?

backends/arm/test/tester/test_pipeline.py
AdrianLundell: I want to pass it in to allow to use these pipelines for any type of module, including networks and so on. The operators expected in a graph is also a type of test data so it seems reasonable to keep it in the tests.

backends/arm/test/ops/test_add.py
AdrianLundell: I kind of agree but it also seems kind of nice to have the quick feedback if you give wrong test data.

I have restructured the Add test case a bit, with that as a starting point I do not think it would be too annoying to follow this pattern.

backends/arm/test/common.py
digantdesai: I was thinking about adding another about decoror meta fn and what its doing. Just so make it more readable and makes the intent clear.

backends/arm/test/ops/test_add.py
digantdesai: > nice to have the quick feedback if you give wrong test data
Yeah, I can see the value now.

backends/arm/test/tester/test_pipeline.py
digantdesai: I meant given (almost always) 1:1 mapping between these two sets of operators, ignoring dim_order ones, but not a strong preference TBH.

digantdesai:
LGTM, bunch of CI issues, not sure how many are related. Please make it green. Let me pull it in internally, see if something goes wrong with `parametrize`

AdrianLundell:
Great, I'll look into it when CI is stable again

AdrianLundell:
None of these CI-failures look related to me

digantdesai:
rebase please, couldn't import it internally. Thanks.

AdrianLundell:
@digantdesai Are you able to import it now?

digantdesai:
Trying again now.

zingo:
@digantdesai  will you handle the merged or should we do it ? Not really sure about best action when PR is "imported". :)

zingo (approved):

PR #6564, opened 2024-10-29
Use pinned commit PyTorch on CI instead of nightly

huydhn:
I think this is a regression.  CI needs to use PyTorch built from the pinned commit instead of installing nightly (using `install_requirements.sh`).  Otherwise, both PT and ET CI won't be able to catch regression against their latest commits.

Various examples:

* https://github.com/pytorch/pytorch/pull/134319#issuecomment-2443209139 supposes to fail ET OSS test, but it passes because CI is running against PyTorch nightly, not the one built from the PR
* Issue encountered by  @tugsbayasgalan https://fb.workplace.com/groups/222849770514616/permalink/523920143740909 where CI runs against nightly instead of his PR

I attempt to fix the issue by instroducing `--use-pt-pinned-commit` to `install_requirements.sh`, so that it can keep the current behavior of installing pinned PT nightly when running locally while use pinned PT commit on CI.  Note that the Docker image only has the pinned PT commit installed.

### Testing

The correct PyTorch built from the pinned commit (23d590e518688f96e1d1947a08e9ca27df3e67e4) is used https://github.com/pytorch/executorch/actions/runs/11601294150/job/32303693643#step:12:6496

On the other hand, using nightly looks like https://github.com/pytorch/executorch/actions/runs/11600922135/job/32302497353#step:12:6552

PR #4826, opened 2024-08-21
Clean up install scripts

lucylq:
Now that setup-macos.sh and setup-linux.sh are using install_requirements, we can remove installing flatc from source, pytorch and pytorch domains install from them.

PR #7558, opened 2025-01-09
Fix the use of PyTorch pin commit in MacOS

huydhn:
I missed this in https://github.com/pytorch/executorch/pull/6564, we need to apply the same fix for MacOS and add a test to make sure that PyTorch built from the pin commit is used instead of nightly.

### DIFF
--- a/.ci/scripts/utils.sh
+++ b/.ci/scripts/utils.sh
@@ -58,7 +58,10 @@ install_pytorch_and_domains() {
   git checkout "${TORCH_VERSION}"
   git submodule update --init --recursive
 
-  export _GLIBCXX_USE_CXX11_ABI=1
+  export USE_PYTORCH_METAL_EXPORT=1
+  export USE_COREML_DELEGATE=1
+  export USE_DISTRIBUTED=0
+
   # Then build and install PyTorch
   python setup.py bdist_wheel
   pip install "$(echo dist/*.whl)"
--- a/.ci/scripts/utils.sh
+++ b/.ci/scripts/utils.sh
@@ -58,6 +58,7 @@ install_pytorch_and_domains() {
   git checkout "${TORCH_VERSION}"
   git submodule update --init --recursive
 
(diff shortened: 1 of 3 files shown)


swolchok (changes requested):
broken: https://github.com/pytorch/executorch/actions/runs/12681330349/job/35344881397?pr=7558#step:9:435

I don't like repeating the computer, but having green CI is the only point for this change, so I have to request changes. please feel free to ping me when it's fixed for expedited review

.ci/scripts/utils.sh
```diff
@@ -40,6 +40,41 @@ install_pip_dependencies() {
   popd || return
 }
 
+install_domains() {
```
huydhn, line 43: These functions were removed by https://github.com/pytorch/executorch/pull/4826.  This was wrong for CI job and we need to bring them back.

swolchok (approved):
shipit

swolchok:
now we have distributed/triton failures https://github.com/pytorch/executorch/actions/runs/12681512566/job/35345401675?pr=7558#step:9:50147

huydhn:
> now we have distributed/triton import failures https://github.com/pytorch/executorch/actions/runs/12681512566/job/35345401675?pr=7558#step:9:50147 . leaving accept in place because we should be directionally correct at this point and I trust you not to ship a red PR

Yeah, this is new. I will make sure to get to the bottom of this before landing the change.

huydhn:
Darn, a GitHub outage is happening https://www.githubstatus.com/incidents/plgzz71xn6zq, need to wait for this storm to pass while testing this locally

huydhn:
It's working now, it turns out that nightly binaries are built with `USE_DISTRIBUTED=1` for MacOS https://github.com/pytorch/pytorch/blob/main/.ci/wheel/build_wheel.sh#L182.  Last time, I remembered that distributed is not available on MacOS, this has probably changed.

PR #7546, opened 2025-01-07
Integrate torchgen exception boundary with ExecuTorch

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7042
* #7041
* __->__ #7546
* #7040

As of #7746, we build with exceptions by default, so we just need to use them.

Differential Revision: [D67904052](https://our.internmc.facebook.com/intern/diff/D67904052/)

PR #7578, opened 2025-01-09
stop double-installing ExecuTorch in one-off linux jobs

swolchok:
* #7042
* #7041
* #7546
* #7040
* #7579
* __->__ #7578

setup-linux.sh already installs ExecuTorch with XNNPACK (and it passes use-pt-pinned-commit as it should).

Differential Revision: [D67996460](https://our.internmc.facebook.com/intern/diff/D67996460/)


digantdesai (approved):

swolchok:
I don't know why, but this is apparently causing timeouts for affected jobs. Any ideas? @digantdesai @dbort @larryliu0820 @mergennachin

.github/workflows/pull.yml
```diff
@@ -199,9 +199,6 @@ jobs:
 
         PYTHON_EXECUTABLE=python bash .ci/scripts/setup-linux.sh "cmake"
 
-        # install pybind
```
dbort, line 202: I agree that we should stop double-installing, but I'd much rather remove the bespoke CI-only installation stuff in favor of install_requirements. I think that if we remove these calls to install_requirements, then none (or very few) of the CI jobs will actually exercise install_requirements. And removing the CI-only logic would reduce overall maintenance costs.

swolchok:
abandoning in favor of #7922

PR #7579, opened 2025-01-09
use-pt-pinned-commit for test-arm-{backend,reference}-delegation

swolchok:
* __->__ #7579
* #7922

Without this, these builds don't respect the torchgen pinned commit and thus fail with #7546.

Differential Revision: [D67996459](https://our.internmc.facebook.com/intern/diff/D67996459/)


digantdesai (approved):

swolchok:
looks like I missed a spot of include path fixing in #7040? not sure exactly why this would make it show up though. https://github.com/pytorch/executorch/actions/runs/12700684319/job/35403838935#step:14:13497

swolchok:
> not sure exactly why this would make it show up though.

yep, turns out I needed to add ciflow/trunk on #7040. will fix there.

huydhn (approved):

freddan80 (approved):

PR #7589, opened 2025-01-10
Make Tester check shapes of reference and model output

oscarandersson8218:
### Summary
torch.allclose() will broadcast data if neccessary. This means that Tester.run_method_and_compare_outputs() may pass even though a delegate produced an output of wrong shape.

### DIFF
--- a/backends/xnnpack/test/tester/tester.py
+++ b/backends/xnnpack/test/tester/tester.py
@@ -1,4 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright 2025 Arm Limited and/or its affiliates.
 # All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


backends/xnnpack/test/tester/tester.py
```diff
@@ -679,6 +680,9 @@ def _assert_outputs_equal(model_output, ref_output, atol=1e-03, rtol=1e-03):
         for i in range(len(model_output)):
             model = model_output[i]
             ref = ref_output[i]
+            assert (
```
digantdesai, line 683: Good catch.

digantdesai (approved):
I hope this didn't unearth some hidden failures :p

digantdesai:
`TestConv1d.test_qs8_conv1d_batchnorm_seq` looks flaky.

PR #7603, opened 2025-01-10
Reduce size of Method::parse_types

swolchok:
* __->__ #7603
* #7602

It looks like this function was big enough that inlining gave up, since I was seeing non-inlined calls to val_as_Foo() that should have been easily optimized away. Do said optimization manually.

Differential Revision: [D68037113](https://our.internmc.facebook.com/intern/diff/D68037113/)


lucylq (approved):
Thank you!

PR #7604, opened 2025-01-10
Buckify Llama multimodal export

jackzhxng:
Summary: Buckify export path for Llama multimodal

Reviewed By: iseeyuan

Differential Revision: D67949092

### DIFF
--- a/examples/models/llama3_2_vision/vision_encoder/model.py
+++ b/examples/models/llama3_2_vision/vision_encoder/model.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# pyre-ignore-all-errors
+
 from dataclasses import dataclass, field
 from typing import Optional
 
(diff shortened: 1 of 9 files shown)


examples/models/llama3_2_vision/vision_encoder/model.py
```diff
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# pyre-ignore-all-errors
+
```
tarun292, line 8: Generally we prefer not to add this. Can they be fixed?

tarun292 (approved):

examples/models/llama3_2_vision/vision_encoder/model.py
dvorjackz: My internal pyre service is running into proxy issues that I'm trying to sort out, was thinking could just skip since pyre is getting deprecated very soon

tarun292: Do they pass the oss linter? If so then it's ok.

tarun292 (approved):

PR #7606, opened 2025-01-10
Fix duplication of  classes in modules

wesleyer:
Summary: If both the executorch and executorch_llama modules are included, this could cause class duplication errors. This fixes it by reorganizing the modules.

Differential Revision: D68027619


mcr229:
@kirklandsign can you take a look at this?

Gasoonjia (approved):
Accept for unblocking instagram usage

kirklandsign:
Thank you!

PR #7615, opened 2025-01-11
[ET-VK] Request VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT, not RANDOM

swolchok:
### Summary
It looks like we are careful to use only copy_from and copy_to with StagingBuffer on CPU, in which case we only need SEQUENTIAL_WRITE.
This matters on Raspberry Pi 5, where there appears (from debugging) to be only one memory type; that type is HOST_VISIBLE and HOST_COHERENT but not HOST_CACHED.
This is my first time working with Vulkan, so please be critical.

### Test plan
Followed vulkan README.md vk_add example on Raspberry Pi 5; was failing with vmaCreateBuffer returned -8 and now succeeds.


swolchok:
https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/usage_patterns.html recommends doing this for staging data to send to the GPU. It's not immediately clear to me whether we also rely on the "readback" usage pattern mentioned later on.

SS-JIA (approved):

PR #7636, opened 2025-01-13
add qnn test template job

cccclai:
Differential Revision: D68112936

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -405,7 +405,7 @@ jobs:
       docker-image: executorch-ubuntu-22.04-qnn-sdk
       submodules: 'true'
       ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 90
+      timeout: 180
       script: |
         # The generic Linux job chooses to use base env, not the one setup by the image
         CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -395,6 +395,25 @@ jobs:
         # Test llama2
         PYTHON_EXECUTABLE=python bash .ci/scripts/test_llama.sh -model stories110M -build_tool "${BUILD_TOOL}" -mode "${MODE}" -dtype "${DTYPE}" -pt2e_quantize "${PT2E_QUANTIZE}"
 
+  test-qnn-models-linux:
+    name: test-qnn-models-linux
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    strategy:
+      fail-fast: false
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-qnn-sdk
+      submodules: 'true'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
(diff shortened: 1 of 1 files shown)


.github/workflows/pull.yml
```diff
@@ -395,6 +395,25 @@ jobs:
         # Test llama2
         PYTHON_EXECUTABLE=python bash .ci/scripts/test_llama.sh -model stories110M -build_tool "${BUILD_TOOL}" -mode "${MODE}" -dtype "${DTYPE}" -pt2e_quantize "${PT2E_QUANTIZE}"
 
+  test-qnn-models-linux:
+    name: test-mediatek-models-linux
```
haowhsu-quic, line 399: Might be `qnn` I think.

haowhsu-quic:
Thank you for this, we 'll figure out which test cases are mandatory. I think we could just have them run on the emulator, or you have some connected hardware could be used?

cccclai:
> Thank you for this, we 'll figure out which test cases are mandatory. I think we could just have them run on the emulator, or you have some connected hardware could be used?

We can only run emulators on the CI for now. However, @guangy10 is working on the perf benchmark with real device and he has enabled to qnn model running there. I notice that there are some real models in the test_qnn_delegate.py and we can have those running on real device. They will run periodically, but not per pull request, though should be good enough to catch failure and track down the issue.

.github/workflows/pull.yml
```diff
@@ -395,6 +395,25 @@ jobs:
...
+    name: test-qnn-models-linux
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    strategy:
+      fail-fast: false
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-qnn-sdk
+      submodules: 'true'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
```
huydhn, line 408: I increase the timeout value to a large number because of a recent change to wait for the Docker image to be available, which could take up to an hour when it happens

```suggestion
      timeout: 180
```

huydhn (approved):

zingo:
Hi
This might have broken pull and trunk jobs since around this patch got merged they are crashing , this PR also seem to have the same problem:  
https://hud.pytorch.org/pr/pytorch/executorch/7636

zingo:
Might not be this patch, I tried a "revert" PR and get the same issue in it so it might probably be something else causing it.

huydhn:
Ops, that is a real error, let me submit a forward fix

huydhn:
The fix PR https://github.com/pytorch/executorch/pull/7731

guangy10:
@cccclai We already have the qnn models in trunk.yml: https://github.com/pytorch/executorch/blob/main/.github/workflows/trunk.yml#L302-L323, which will be triggered on PR if files under tracked paths got modified. Why do we need to add same to pull.yml?

cccclai:
> @cccclai We already have the qnn models in trunk.yml: https://github.com/pytorch/executorch/blob/main/.github/workflows/trunk.yml#L302-L323, which will be triggered on PR if files under tracked paths got modified. Why do we need to add same to pull.yml?

oh this pr is to cover the perf op test defined in https://github.com/pytorch/executorch/blob/57a09f49027ab05515afef4315d0a797657222f3/backends/qualcomm/tests/test_qnn_delegate.py#L4

PR #7647, opened 2025-01-14
Remove quantize_io from compile_spec

Erik-Lundell:
quantize_io was only used in arm_partitioner
and is not needed there anymore when
running the delegate in the graph (https://github.com/pytorch/executorch/pull/7552).


Erik-Lundell:
Ci failure unrelated

freddan80 (approved):

PR #7649, opened 2025-01-14
Use ArmQuantizer to quantize bias

per:
Remove the 'manual' quantization of bias parameter and let the quantizer handle the quantization instead.

### Test plan
Uses Arm backend unit tests for testing.


per:
- Cancelled macos jobs not related to this change.
- check label job has run successfully.

freddan80 (approved):

PR #7653, opened 2025-01-14
add begin/end to ExecuTorch pytree::arr

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7655
* #7654
* __->__ #7653

Allows use of range-for.

Differential Revision: [D68166302](https://our.internmc.facebook.com/intern/diff/D68166302/)

PR #7655, opened 2025-01-14
Replace pytree_assert with production pytree_check. Remove pytree_unreachable

swolchok:
* __->__ #7655
* #7654
* #7653

When handling untrusted input, it's not appropriate to use debug-only checks; we should be checking in prod as these are not programmer errors. pytree_unreachable was similarly being used for input validation.

Differential Revision: [D68166301](https://our.internmc.facebook.com/intern/diff/D68166301/)


extension/pytree/pybindings.cpp
```diff
@@ -145,7 +145,8 @@ class PyTree {
           } else if (py::isinstance<py::int_>(key)) {
             s.key(i) = py::cast<int32_t>(key);
           } else {
-            pytree_assert(false);
+            throw std::runtime_err(
+                "invalid key in pytree dict; must be int or string");
```
dbort, line 149: It'd be helpful to print the key type here, though the old code didn't do it either so not a blocker. Similar for other new `throw` messages in this PR.

dbort (approved):
Much nicer behavior than before!

PR #7663, opened 2025-01-14
Validate tensor sizes during method load

dbort:
Summary:
If a tensor's `sizes` array contains a negative value, the TensorImpl ctor will panic.

Validate sizes before creating the TensorImpl to treat this as a non-fatal error when loading malformed PTE files.

Differential Revision: D68180029


JacobSzwejbka (approved):

JacobSzwejbka:
Why didnt we just make it a uint in the schema originally 🤔

PR #7664, opened 2025-01-15
Additional QNN version fix

guangy10:
QNN version is bumped in https://github.com/pytorch/executorch/pull/6811. Additional fix in android-perf


cccclai (approved):
Thanks!

cccclai:
Should we consolidate to use .ci/scripts/build-qnn-sdk.sh to avoid issues like this?

guangy10:
> Should we consolidate to use .ci/scripts/build-qnn-sdk.sh to avoid issues like this?

Yeah, there was an issue of extracting those env vars from the GA. I can look into it and follow up in a separate PR.

PR #7669, opened 2025-01-15
Align handling of flaky arm unittests

AdrianLundell:
- Removes uses of torch_manual_seed which previously fixed the random state.
- Adds pytest plugin pytest-rerunfailures to mark flaky tests.
- Refactors flaky tests to use data generators in favor of pregenerated data, which ensures that data is randomized between reruns.
- Updates layer_norm testcase to use same qtol value for TOSA/EthosU targets.

Note that fixing the randomness may lead to that we will see more flakyness in CI, this will have to be adressed with the flaky mark on a case by case basis over time.


AdrianLundell:
Looking into unittest-arm/linux-job errors, others are unrelated:
`ERROR: No matching distribution found for torchtune==0.4.0.dev20241112`

AdrianLundell:
I'll have to wait with this until CI is fixed, do not merge.

zingo (approved):

zingo:
Macos and phi fails seems unrealted.

zingo:
@digantdesai This touch a file outside arm folder is it OK to merge? 
e.g.  pyproject.toml

AdrianLundell:
Now only unrelated CI errors left!

AdrianLundell:
@digantdesai What do you think, is it ok to add the new pytest-rerunfailures dependency?

pyproject.toml
```diff
@@ -61,6 +61,7 @@ dependencies=[
   "parameterized",
   "pytest",
   "pytest-xdist",
+  "pytest-rerunfailures",
```
digantdesai, line 64: Yeah should be OK. 

before we land though, I would like to understand what is the root-cause of flakiness, can it be resolved by increasing ATOL?

AdrianLundell: The root-cause is numerical inaccuracies in our random tests as you suggest. The reason for preferring reruns over increasing atol is to indicate that the op generally passes except for a few flukes, whereas a high atol kind of suggests an overall lack of accuracy. We can also more easily track if flakyness decreases over time, while atol values have to be updated manually.

digantdesai: Kind of agree. It is not so much as lack of accuracy but correct expectation of accuracy given a distribution of fp32 input, data graph, and quantization method used.

PR #7670, opened 2025-01-15
Update torchtune pin

mergennachin:
### Summary 

Current torchtune pin is not in the S3 bucket anymore. https://download.pytorch.org/whl/nightly/torchtune/

Updating to latest

### Test

CI

### DIFF
--- a/examples/models/llama3_2_vision/install_requirements.sh
+++ b/examples/models/llama3_2_vision/install_requirements.sh
@@ -5,7 +5,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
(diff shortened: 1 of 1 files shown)


ezyang (approved):

jackzhxng (commented):
Why is it not in the bucket anymore - do we evict old versions? Ideally we only update the TorchTune pin out of necessity since the test coverage for this model is not completely there yet

mergennachin:
@dvorjackz - yeah i think we only keep like 90 days (or 60 days) in our nightlies https://download.pytorch.org/whl/nightly/torchtune/ 

perhaps to save on storage spaces and for maintenance.

PR #7599, opened 2025-01-10
Upgrade numpy version requirement to 2.0

mergennachin:
Upgrade numpy version requirement to 2.0.0

### Summary:

1.21 is too old now. For 3.10 or above, we should be able to use 2.0.0

We are already on CoreMLTools 8.1, which is https://github.com/apple/coremltools/blob/8.1/reqs/build.pip

We had to upgrade to the latest transformers as well

Fixes: 

https://github.com/pytorch/executorch/issues/6639

### Test Plan (pending)

- [x] CI
- [x] Run setup script (https://pytorch.org/executorch/stable/getting-started-setup.html)
- [x] Run coreml setup script (https://github.com/pytorch/executorch/blob/main/backends/apple/coreml/setup.md)
- [x] Run llama example (https://github.com/pytorch/executorch/blob/main/examples/models/llama/README.md)
- [x] Run llama 3.1 example with coreml backend https://github.com/pytorch/executorch/blob/main/examples/models/llama/non_cpu_backends.md
- [ ] Run llava example (https://github.com/pytorch/executorch/tree/main/examples/models/llava)
- [ ] Run phi-3-mini example (https://github.com/pytorch/executorch/tree/main/examples/models/phi-3-mini)

PR #7675, opened 2025-01-15
Move ai-edge-model-explorer into devtools/install_requirements.sh

mergennachin:
### Summary

From top-level requirements, move to devtools/install_requirement.sh instead.

ai-edge-model-explorer is too new and currently requires numpy<2 (https://github.com/google-ai-edge/model-explorer/blob/main/src/server/package/pyproject.toml#L26) which conflicts with our recent upgrade (https://github.com/pytorch/executorch/pull/7599)

Let's not take core dependency on this tool yet.

### Test Plan

python devtools/visualization/visualization_utils_test.py

make sure error is thrown

run `sh devtools/install_requirement.sh`

make sure the script is successfully run.

### DIFF
--- a/devtools/install_requirements.sh
+++ b/devtools/install_requirements.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Conflict: this requires numpy<2 whereas ExecuTorch core requires numpy>=2
+# Follow https://github.com/google-ai-edge/model-explorer/issues/277 for potential
+# resolution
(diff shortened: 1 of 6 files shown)


devtools/install_requirements.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Conflict: this requires numpy<2 whereas ExecuTorch core requires numpy>=2
```
digantdesai, line 8: How does this will work, once ET moves to numpy>=2?
Does Google have a plan to upgrade to a newer numpy?

digantdesai (approved):

per:
FYI: @Erik-Lundell

PR #7687, opened 2025-01-16
Add uint16 to supported dtypes and regenerate edge.yaml

tarun292:
Summary: Adding `uint16` support and regenerating `edge.yaml` as we now support `uint16` in the ExecuTorch stack.

Differential Revision: D68241997


zonglinpeng (approved):
I can confirm it works for G3 permute thanks!

PR #7690, opened 2025-01-16
[not for commit] Tosa ser fix on 0.5

digantdesai:


digantdesai:
This doesn't have unit-test skip just yet to see, if we get the same errors or not.

digantdesai:
can we try with pulling https://github.com/pytorch/executorch/pull/7670 as well? on release/0.5?

PR #7703, opened 2025-01-16
install_requirements.py: use argparse, minor cleanup

swolchok:
Replace the bespoke argument parser with argparse in preparation for separating requirements installation from building and installing ExecuTorch itself.

Test Plan: (with debug print of args.pybind to save time)
install_requirements.sh --pybind xnnpack --pybind mps --pybind coreml
install_requirements.sh --pybind xnnpack mps coreml
install_requirements.sh --clean

CI should cover use-pt-pinned-commit

### DIFF
--- a/install_requirements.py
+++ b/install_requirements.py
@@ -6,7 +6,9 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import argparse
 import glob
+import itertools
 import os
 import platform
 import re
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

install_requirements.py
```diff
@@ -63,175 +65,196 @@ def python_is_compatible():
...
-        dirs = glob.glob("cmake-out*/") + glob.glob("cmake-android-out/")
-        for d in dirs:
-            print(f"Cleaning {d}...")
-            shutil.rmtree(d, ignore_errors=True)
-        print("Done cleaning build artifacts.")
-        sys.exit(0)
-    elif arg == "--use-pt-pinned-commit":
+            for pybind_arg in args.pybind:
+                if pybind_arg not in ["coreml", "mps", "xnnpack"]:
+                    continue
```
mergennachin, line 120: raise exception here?

install_requirements.py
```diff
@@ -63,175 +65,196 @@ def python_is_compatible():
...
+                EXECUTORCH_BUILD_PYBIND = "ON"
+                CMAKE_ARGS += f" -DEXECUTORCH_BUILD_{pybind_arg.upper()}=ON"
+
+    if args.clean:
```
mergennachin, line 124: do this if statement first?

mergennachin (approved):
See inline

install_requirements.py
swolchok: whoops! I misinterpreted the original code after I'd made some modifications to it. will fix.

PR #7715, opened 2025-01-16
Add install_requirements.{sh,py} that only install requirements

swolchok:
Now that the previous PR renamed install_requirements to install_executorch, we can use those names for scripts that actually just install requirements.


swolchok:
* (to be filled)

mergennachin (approved):

install_requirements.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import sys
+
+import install_executorch
+
+
```
mergennachin, line 13: np: don't you wanna do the other way around? 

move the `install_requirements()` function into this file instead.

swolchok: done

swolchok:
merging on the strength of previous green CI despite a few runs not quite having finished.

PR #7719, opened 2025-01-17
[xnnpack] Serialize Access to delegate::init

digantdesai:
### Summary
init() can be called in parallel, and XNNPACK API is not thread-safe.

### Test plan
CI

cc @mcr229


kimishpatel (approved):

mcr229 (changes requested):
Sorry let's add a test for this first before, so we can verify the exact case that this is fixing

digantdesai:
> Sorry let's add a test for this first before, so we can verify the exact case that this is fixing

We already have an internal test, but it is not in OSS.

mcr229 (approved):
will approve to unblock, but would like a test to catch this.

PR #7721, opened 2025-01-17
[devtools/visualization] Add visualize_graph

Erik-Lundell:
When working with passes, you might have access to a modified graph_module rather than an exported_program. visualize_graph allows visualization of this graph_module by combining  the modified graph_module with an exported_program. Note that the graph_module can't be set directly, a new exported_program needs to be constructed.

Additionally, we disable the operator validation for the newly constructed ExportedProgram. This is ok since it is only used for visualization.

### DIFF
--- a/devtools/visualization/visualization_utils.py
+++ b/devtools/visualization/visualization_utils.py
@@ -6,9 +6,13 @@
 
 import subprocess
 import time
+from typing import Any, Callable, Type
 
 from executorch.exir import EdgeProgramManager, ExecutorchProgramManager
+from executorch.exir.program._program import _update_exported_program_graph_module
+from torch._export.verifier import Verifier
 from torch.export.exported_program import ExportedProgram
+from torch.fx import GraphModule
 
 try:
     from model_explorer import config, consts, visualize_from_config  # type: ignore
@@ -27,7 +31,7 @@ class SingletonModelExplorerServer:
 
     server: None | subprocess.Popen = None
     num_open: int = 0
-    wait_after_start = 2.0
+    wait_after_start = 3.0
 
     def __init__(self, open_in_browser: bool = True, port: int | None = None):
         if SingletonModelExplorerServer.server is None:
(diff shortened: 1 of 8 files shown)


zingo:
Hi @digantdesai this PR touch non Arm folders, OK?

devtools/visualization/visualization_utils.py
```diff
@@ -124,3 +128,29 @@ def visualize(
         no_open_in_browser=no_open_in_browser,
         **kwargs,
     )
+
+
+def visualize_graph(
```
digantdesai, line 133: Curious, why not overload the `visualize` method, and add support for additional graph_module in `_get_exported_program()`?

devtools/visualization/visualization_utils_test.py
```diff
@@ -145,6 +147,17 @@ def test_visualize_to_executorch(server):
...
 
+def test_visualize_graph(server):
+    with server():
+        model = Linear(20, 30)
+        exported_program = torch.export.export(model, model.get_inputs())
+        exported_program = to_edge_transform_and_lower(
+            exported_program
+        ).exported_program()
+        modified_gm = DecomposeLinearPass()(exported_program.graph_module).graph_module
+        visualize_graph(modified_gm, exported_program)
```
digantdesai, line 158: For the future, IMO it is more valuable to show a "diff view" (not sure how it will work) when a graph module is provided along with an exported program? Because most likely that is what I would care about.

devtools/visualization/visualization_utils_test.py
```diff
@@ -158,3 +171,5 @@ def test_visualize_to_executorch(server):
         test_visualize_to_edge(SingletonModelExplorerServer)
         test_visualize_partition(SingletonModelExplorerServer)
         test_visualize_to_executorch(SingletonModelExplorerServer)
+        test_visualize_graph(SingletonModelExplorerServer)
+        time.sleep(3.0)
```
digantdesai, line 175: Comment why?

devtools/visualization/visualization_utils_test.py
```diff
@@ -145,6 +147,17 @@ def test_visualize_to_executorch(server):
...
 
 
+def test_visualize_graph(server):
+    with server():
+        model = Linear(20, 30)
+        exported_program = torch.export.export(model, model.get_inputs())
+        exported_program = to_edge_transform_and_lower(
+            exported_program
+        ).exported_program()
+        modified_gm = DecomposeLinearPass()(exported_program.graph_module).graph_module
```
digantdesai, line 157: why can't we just `_transform` and create a new ep and use `visualize(new_ep)`?

devtools/visualization/visualization_utils.py
Erik-Lundell: I would argue that constructing a new ExportedProgram from an old + GraphModule is different enough from just visualizing that it's good to signal that with another function name.

devtools/visualization/visualization_utils_test.py
Erik-Lundell: One use case I had in mind for this was inside passes. Also, if you use your own PassManager, it returns a PassResult that contains a graph_module. So in this particular test case you could do that but I think this approach is more flexible.

devtools/visualization/visualization_utils_test.py
Erik-Lundell: That's a very interesting idea! Definitely something to think about for the future.

devtools/visualization/visualization_utils_test.py
Erik-Lundell: Ack

digantdesai (approved):

PR #7722, opened 2025-01-17
[Arm backend] Refactor runner_utils, removing RunnerUtil

Erik-Lundell:
As the execution of models on reference_model
or on FVP has simplified, there is no longer need
for a class to handle the execution.
Instead, running a model should should only be a
function call with the model, inputs and some metadata. This makes it easier to reuse runner_utils.

Change-Id: Ib6addbdaa2fa492c356e92ad1b95c2dac2885b9d


zingo (approved):

Erik-Lundell:
Unrelated ci failure.

PR #7727, opened 2025-01-17
Manual cherry-pick to update numpy as well as all dependencies

mergennachin:
- **Install torchao by default (#7470)**
- **Upgrade numpy version requirement to 2.0 (#7599)**
- **Update torchtune pin to release version**
- **[Arm] TOSA reference model update**


mergennachin:
@dvorjackz @SS-JIA 

I just realized that we have to update the torchtune pin to release version, just like how we do for everything else (torch, torchaudio, torchvision etc)

Otherwise, even if we update the nightly pin, when someone uses the release binary in a few months, one of its dependencies, torchtune, will not be working

Initially I tried cherry-picked (https://github.com/pytorch/executorch/pull/7670), but later manually override with release version.

mergennachin:
> would I be correct to assume that the only next steps would be to update the release workflow to say that torchtune must be updated as well?

Yes, that would be correct

SS-JIA (approved):

jackzhxng (commented):
@mergennachin yup should be fine as long as tests past

PR #7742, opened 2025-01-17
Don't clobber existing CMAKE_CXX_FLAGS_RELEASE

swolchok:
We were likely losing -DNDEBUG. Preserve our current intent to clobber the -O flag in particular, though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ endif()
 # they can be properly gc'd. -s: strip symbol. -fno-exceptions -fno-rtti:
 # disables exceptions and runtime type.
 set(CMAKE_CXX_FLAGS_RELEASE
-    "-ffunction-sections -fdata-sections -fno-exceptions -fno-rtti"
+    "-ffunction-sections -fdata-sections -fno-exceptions -fno-rtti ${CMAKE_CXX_FLAGS_RELEASE}"
 )
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

CMakeLists.txt
```diff
@@ -149,19 +149,19 @@ endif()
...
+    "-ffunction-sections -fdata-sections -fno-exceptions -fno-rtti ${CMAKE_CXX_FLAGS_RELEASE}"
 )
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
 endif()
 
 option(OPTIMIZE_SIZE "Build executorch runtime optimizing for binary size" OFF)
 if(OPTIMIZE_SIZE)
-  # -Os: Optimize for size
-  set(CMAKE_CXX_FLAGS_RELEASE "-Os ${CMAKE_CXX_FLAGS_RELEASE}")
```
dbort, line 161: It's pretty typical to put the overridable CMAKE_CXX_FLAGS_RELEASE last so it can override `-Os`/`-O2` if users want to; the new order will always override whatever CMAKE_CXX_FLAGS_RELEASE says. But if that's the intent (like the PR description says), then this seems to do it.

But since this is an unusual pattern, please update the comments on these two cases to clarify that we're intentionally overriding the `-O` flag so that someone doesn't come back and "fix" it.

dbort (approved):

CMakeLists.txt
swolchok: > someone doesn't come back and "fix" it.

I would be happy for someone to come back and change this further; these two `OPTIMIZE_SIZE` changes were to preserve existing behavior.

Before: we clobber *all* of CMAKE_CXX_FLAGS_RELEASE on line 152 above.
After: we only clobber the `-O` setting with `-Os` or `-O2`. 

I think our setup is deeply weird and `OPTIMIZE_SIZE` shouldn't exist; I am just following general good practice by sending an incremental diff that makes a small amount of forward progress rather than attempting to fix everything all at once.

dbort: > After: we only clobber the -O setting with -Os or -O2.

Agreed that it's an improvement. But someone a year from now who's only looking at this line might think "why is -Os last? I'll move it back before the `${CMAKE_CXX_FLAGS_RELEASE}`". A comment would help avoid that.

swolchok: I'm hesitant to enshrine behavior I think is wrong and simply haven't gotten around to fixing yet.

PR #7743, opened 2025-01-17
Don't repeat default for CMAKE_CXX_FLAGS_DEBUG

swolchok:
Debug build type already has defaults, and the flags we set are likely to be those defaults on Unix-like platforms. Why break Windows compatibility?


swolchok:
* (to be filled)

dbort (approved):

PR #7746, opened 2025-01-17
Don't default to -fno-exceptions -fno-rtti

swolchok:
We need exceptions/RTTI for non-core-only use cases (see #7736).
Sending first version of this without patching any specific targets to restore these flags because I want to see if anything actually breaks.


swolchok:
* (to be filled)

swolchok:
phi-3-mini and emformer-transcribe failures are from trunk, so it looks likely that this doesn't break anything.

swolchok:
Specific question for reviewers: which CI builds, if any, do we want/need to disable exceptions/rtti? I could go either way on the size test; for the case where we include non-portable ops we need to leave them enabled, but it's up to us what we want to do for the core-only case.

mergennachin:
https://github.com/pytorch/executorch/actions/runs/12919450651/job/36029994766?pr=7746 -- interesting, it is still under the limit

mergennachin:
@mcremon-meta - are there places in cadence code where we rely on these flags to be present? if so, you might need to change your config explicitly

mergennachin:
@robell , @freddan80 - FYI, see https://github.com/pytorch/executorch/discussions/7736 for background and context

swolchok:
> https://github.com/pytorch/executorch/actions/runs/12919450651/job/36029994766?pr=7746 -- interesting, it is still under the limit

#7742 lowered it from 51768 to 47664, probably because it caused us to -DNDEBUG in release builds. Mildly surprised we are only up to 47672 after this diff; I wonder why these flags were so ineffective at reducing size.

mcremon-meta:
> @mcremon-meta - are there places in cadence code where we rely on these flags to be present? if so, you might need to change your config explicitly

cc @hsharma35 @zonglinpeng in case we need to adjust something

CMakeLists.txt
```diff
@@ -149,7 +149,7 @@ endif()
 # they can be properly gc'd. -s: strip symbol. -fno-exceptions -fno-rtti:
```
malfet, line 149: Looks like this could be edited as part of the same PR

malfet (commented):
I would like to see those flags be preserved by default, if one say compiles Qualcomm delegate. 
I'm not sure of ET build system has separate set of flags for tracer vs real runtime.

malfet (approved):
I'm not an expert in ET build system, and IMO RTTI/exceptions should be enabled  only for desktop/iOS/Android targets, but in order to unblock you..

mergennachin:
Can you adjust the current CI test for size_test and size_test_all_ops to use --no-exceptions, --no-rtti flags.

digantdesai:
>  which CI builds, if any, do we want/need to disable exceptions/rtti? 

I guess arm-delegation job ([test-arm-backend-delegation](https://github.com/pytorch/executorch/actions/runs/12936013289/job/36080643522?pr=7746#logs)) which is somewhat sensitive to this (builds a bare metal bin for a microcontroller) are failing for an unrelated reason. I would get that signal for this PR, and perhaps add a no-except, no-rtti flavor as well.

swolchok:
> perhaps add a no-except, no-rtti flavor as well.

patched the build script in the latest update. 

> Can you adjust the current CI test for size_test and size_test_all_ops to use --no-exceptions, --no-rtti flags.

done! here is the effect. before restoring `-fno-exceptions -fno-rtti`:

```
ExecuTorch with no ops binary size, unstripped:
-rwxr-xr-x  1 swolchok  staff  87192 Jan 27 12:12 cmake-out/test/size_test
ExecuTorch with portable ops binary size, unstripped:
-rwxr-xr-x  1 swolchok  staff  1373168 Jan 27 12:12 cmake-out/test/size_test_all_ops

stripped:
~/src/executorch> ls -al cmake-out/test/size_test
-rwxr-xr-x  1 swolchok  staff  73672 Jan 27 12:39 cmake-out/test/size_test
~/src/executorch> ls -al cmake-out/test/size_test_all_ops
-rwxr-xr-x  1 swolchok  staff  1172688 Jan 27 12:39 cmake-out/test/size_test_all_ops
```

after:
```
ExecuTorch with no ops binary size, unstripped:
-rwxr-xr-x  1 swolchok  staff  68584 Jan 27 12:10 cmake-out/test/size_test
ExecuTorch with portable ops binary size, unstripped:
-rwxr-xr-x  1 swolchok  staff  1370064 Jan 27 12:10 cmake-out/test/size_test_all_ops

stripped:
~/src/executorch> ls -al cmake-out/test/size_test
-rwxr-xr-x  1 swolchok  staff  56552 Jan 27 12:41 cmake-out/test/size_test
~/src/executorch> ls -al cmake-out/test/size_test_all_ops
-rwxr-xr-x  1 swolchok  staff  1172064 Jan 27 12:41 cmake-out/test/size_test_all_ops
```

I am surprised that core is affected so much more than core + portable ops.

mergennachin (approved):

swolchok:
should probably also patch the cadence build scripts to set -fno-exceptions -fno-rtti so as to preserve the example being good

CMakeLists.txt
swolchok: done

digantdesai (approved):

PR #7747, opened 2025-01-17
Guard against missing list values

dbort:
Summary: We checked for missing lists on some types, but not all. Add checks to everything in method.cpp that calls `->items()`.

Differential Revision: D68344960


swolchok (approved):
shipit

tarun292 (approved):

PR #7763, opened 2025-01-18
[executorch][core] NamedDataMap interface

lucylq:
* #7900
* __->__ #7763

Add NamedDataMap interface to runtime.

Differential Revision: [D66834552](https://our.internmc.facebook.com/intern/diff/D66834552/)


runtime/core/named_data_map.h
```diff
@@ -0,0 +1,47 @@
...
+  get_metadata(const char* fqn) const = 0;
+  /**
+   * Get tensor data by fully qualified name (FQN).
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @return Result containing a span of uint8_t representing the tensor data.
+   */
+  ET_NODISCARD virtual Result<Span<const uint8_t>> get_data(
+      const char* fqn) const = 0;
+};
```
JacobSzwejbka, line 44: Can you add the functions to get the list of fqns?

JacobSzwejbka (changes requested):

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,47 @@
...
+   */
+  ET_NODISCARD virtual Result<const executorch::runtime::TensorLayout>
+  get_metadata(const char* fqn) const = 0;
+  /**
+   * Get tensor data by fully qualified name (FQN).
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @return Result containing a span of uint8_t representing the tensor data.
+   */
+  ET_NODISCARD virtual Result<Span<const uint8_t>> get_data(
```
dbort, line 42: This needs to return a FreeableBuffer, otherwise there's no way to manage the data. Span doesn't own its data, so anything returned here would be forced to leak memory.

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,47 @@
...
+#include <executorch/runtime/core/result.h>
+#include <executorch/runtime/core/span.h>
+#include <executorch/runtime/core/tensor_layout.h>
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch {
+namespace runtime {
+
+/**
+ * Interface to access and retrieve data via name from a loaded data file.
```
dbort, line 22: NDM has no opinion about where the data comes from, so "loaded data file" is too specific. Pointing to the example is ok, though.

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,47 @@
...
+ * See executorch/extension/flat_tensor/ for an example.
+ */
+class NamedDataMap {
+ public:
+  virtual ~NamedDataMap() = default;
+  /**
+   * Get tensor metadata by fully qualified name (FQN).
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @return Result containing a pointer to the metadata.
```
dbort, line 32: Doesn't appear to be a pointer

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,47 @@
...
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch {
+namespace runtime {
+
+/**
+ * Interface to access and retrieve data via name from a loaded data file.
+ * See executorch/extension/flat_tensor/ for an example.
+ */
+class NamedDataMap {
```
dbort, line 25: Please annotate this as ET_EXPERIMENTAL so that it doesn't become part of our supported API just yet.

dbort (changes requested):

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,62 @@
...
+   */
+  ET_NODISCARD virtual Result<const executorch::runtime::TensorLayout>
+  get_metadata(const char* fqn) const = 0;
+  /**
+   * Get tensor data by fully qualified name (FQN).
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @return Result containing a FreeableBuffer with the tensor data.
+   */
+  ET_NODISCARD virtual Result<FreeableBuffer> get_data(
```
JacobSzwejbka, line 42: Ok last request from my side is for a get_data_into method. This lets us avoid a useless load into the freeableBuffer and lets us just load directly into the buffer we are targeting for mutation. Primary usage is training, but it would also be used if someone put serialized mutable state like say an initial cache into the .ptd

JacobSzwejbka: https://github.com/pytorch/executorch/blob/main/runtime/executor/program.h#L251
and
https://github.com/pytorch/executorch/blob/main/runtime/core/data_loader.h#L107

for code pointers of where we do this in .pte loading. I think this would be roughly parallel to the program.h code.

JacobSzwejbka (approved):

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+      const char* fqn) const = 0;
+
+  /**
+   * Loads data corresponding to the fqn into the provided buffer.
```
dbort, line 46: Would be good to tell users how to figure out how big a buffer to provide (get metadata and call .nbytes()?)

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @param size The number of bytes to load.
+   * @param buffer The buffer to load the data into. Must point to at least
+   * `size` bytes of memory.
+   * @return An error code on if the load was successful.
```
dbort, line 52: Not sure what this means.

Also, what happens if the size is too small or too large?

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+   */
+  ET_NODISCARD virtual Error
```
dbort, line 54: Consider returning `Result<size_t>`, returning the number of bytes written on success. Then, users can provide a too-big buffer but still know how many bytes are actually valid.

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+  load_data_into(const char* fqn, size_t size, void* buffer) const = 0;
+
+  /**
+   * Get the number of keys in the NamedDataMap.
+   *
+   * @return Result containing the number of keys.
+   */
+  ET_NODISCARD virtual Result<int> get_num_keys() const = 0;
```
dbort, line 62: Please use size_t for this
```suggestion
  ET_NODISCARD virtual Result<size_t> get_num_keys() const = 0;
```

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Get the key at the given index.
+   *
+   * @param index The index of the key to retrieve.
+   * @return Result containing the key at the given index.
+   */
+  ET_NODISCARD virtual Result<const char*> get_key(int index) const = 0;
```
dbort, line 70:
```suggestion
  ET_NODISCARD virtual Result<const char*> get_key(size_t index) const = 0;
```

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @return Result containing the number of keys.
+   */
+  ET_NODISCARD virtual Result<int> get_num_keys() const = 0;
+
+  /**
+   * Get the key at the given index.
+   *
+   * @param index The index of the key to retrieve.
+   * @return Result containing the key at the given index.
```
dbort, line 68: Please mention that the returned pointer is only valid for the lifetime of the NDM.

dbort (changes requested):
Looking good!

runtime/core/named_data_map.h
lucylq: The NamedDataMap implementation should throw an error if the size is too small. Thanks for the suggestion to return Result<size_t>, this can provide information on if the buffer was too large.

runtime/core/named_data_map.h
lucylq: thanks!

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Loads data corresponding to the fqn into the provided buffer.
+   *
+   * @param fqn Fully qualified name of the tensor.
+   * @param size The number of bytes to load. Use `get_metadata` to retrieve the
+   * size of the tensor for a given fqn.
+   * @param buffer The buffer to load the data into. Must point to at least
+   * `size` bytes of memory.
+   * @return Result containing the number of bytes written on success.
```
dbort, line 53: Please clarify if this will fail or truncate if `size` is too small

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,76 @@
...
+   */
+  ET_NODISCARD virtual Result<size_t>
+  load_data_into(const char* fqn, size_t size, void* buffer) const = 0;
```
dbort, line 56: nit: traditionally (like in memcpy) the buffer comes before the size

```suggestion
  load_data_into(const char* fqn, void* buffer, size_t size) const = 0;
```

runtime/core/named_data_map.h
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Get the number of keys in the NamedDataMap.
+   *
+   * @return Result containing the number of keys.
+   */
+  ET_NODISCARD virtual Result<size_t> get_num_keys() const = 0;
```
dbort, line 63: These two methods use "key" to refer to the strings, but the other methods use "fqn". It'd be better to use a single term for the entire interface.

Since NDM is fairly generic, I like "key" better, since we might use an NDM to look up data that is keyed by strings that aren't technically pytorch FQNs. But the most important thing is that the API is consistent.

dbort (approved):
Some final minor things but this looks good!

PR #7783, opened 2025-01-21
Test the IOQuantization pass does not have Q/DQ nodes

gggekov:

### DIFF
--- a/backends/arm/test/passes/test_ioquantization_pass.py
+++ b/backends/arm/test/passes/test_ioquantization_pass.py
@@ -1,0 +1,50 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+
+import torch
+
+from executorch.backends.arm.test import common
+
+from executorch.backends.arm.test.tester.arm_tester import ArmTester
+from executorch.exir.passes.quantize_io_pass import QuantizeInputs, QuantizeOutputs
+
+
+class SimpleModel(torch.nn.Module):
+    def forward(self, x, y):
+        return x + y
+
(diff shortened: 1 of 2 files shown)


backends/arm/test/passes/test_ioquantization_pass.py
```diff
@@ -0,0 +1,50 @@
...
+            .quantize()
+            .export()
+            .to_edge()
+            .partition()
+        )
+        edge = tester.get_artifact()
+        edge.transform(
+            passes=[QuantizeInputs(edge, [0, 1]), QuantizeOutputs(edge, [0])]
+        )
+        tester.check_not(["edge__ops_quantized_decomposed_quantize_per_tensor"])
```
digantdesai, line 49: can you add additional checks to verify,
1 - exactly 3 q, and 3 dq nodes are in the graph before partition
2 - 2 q and 1 dq node in the graph after `partition()` or before the pass
3 - 0 q and 1 dq after the pass

digantdesai (approved):
LGTM, left a comment about improving checks.

backends/arm/test/passes/test_ioquantization_pass.py
gggekov: Thanks @digantdesai . I can check we have exactly 3 Q & 3 DQs before partition, 2 Q & 1 DQ after partitioning, but what do you mean under 1DQ after the pass ? After the pass, the forward method takes int8 and returns int8 so from the embedded application perspective, the output of the NN is in int8. With the pass, in the pte we have baked in methods to get the quantization parameters if needed. Am I missing something about the DQ after the pass ?

digantdesai: sorry it is a typo, 0 q and 0 dq nodes :)

digantdesai (approved):

zingo (approved):

PR #7785, opened 2025-01-21
Add git hook suggestions for Arm development

Erik-Lundell:
Pre-commit runs lintrunner on commited files,
tries applying the changes.
Pre-push does not push if lintrunner or
a simple license check does not pass.

If WIP is included in commit header,
no checks are done.

### DIFF
--- a/backends/arm/scripts/pre-commit
+++ b/backends/arm/scripts/pre-commit
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Check 1: If commit header contains WIP, everything is ok
+git rev-list --format=%s --max-count=1 HEAD | grep -q WIP && exit 0
+
+# Check 2: lintunner on latest patch.
+lintrunner -a --revision 'HEAD^' --skip MYPY
+commit_files=$(git diff-tree --no-commit-id --name-only --diff-filter=M HEAD -r)
+git add $commit_files || true
--- a/backends/arm/scripts/pre-push
+++ b/backends/arm/scripts/pre-push
@@ -1,0 +1,45 @@
+#!/bin/bash
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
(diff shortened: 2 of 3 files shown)


backends/arm/scripts/pre-commit
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Check 1: If commit header contains WIP, everything is ok
+git rev-list --format=%s --max-count=1 HEAD | grep -q WIP && exit 0
+
+# Check 2: lintunner on latest patch.
+lintrunner -a --revision 'HEAD^' --skip MYPY
```
digantdesai, line 11: I was thinking about doing this on Friday :)

digantdesai: `-a` will change stuff without user noticing, is that how we want to do it?

backends/arm/scripts/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
digantdesai, line 1: why does this have to be arm specific?

digantdesai (commented):
Looks good, let's discuss some of the comments I left.

backends/arm/scripts/pre-commit
Erik-Lundell: I guess it doesn't, should I move it to the top-level scripts folder?

backends/arm/scripts/pre-commit
Erik-Lundell: My reasoning was that the user usually don't care about linting fixes anyways. Do you see a case where this could be an issue/ have unintended consequences?

digantdesai: They might care if the linter changes messes up with let's say mypy annotations.

backends/arm/scripts/pre-commit
digantdesai: I would create another PR for ET for `pre-commit`

oscarandersson8218 (approved):

PR #7796, opened 2025-01-21
Fix test to be robust against fp numerics

mergennachin:
Summary:

https://github.com/pytorch/pytorch/actions/runs/12890471648/job/35941113858


ezyang (approved):

mergennachin:
https://github.com/pytorch/pytorch/actions/runs/12890471648/job/35941113858

My guess is that it is for even length case. p50 means taking the average of two middle elements, which may cause numerical differences...

For odd length cases, it should be exact though.

PR #7803, opened 2025-01-21
Print delegation info in export_llama in verbose

iseeyuan:
### Summary
For better dev and debugging experience, add delegation info prints in verbose in export_llama. It only shows up with `-v` option. 

### Test plan
```
python -m examples.models.llama.export_llama \
  --model "llama3_2" \
  --checkpoint "${LLAMA_CHECKPOINT:?}" \
  --params "${LLAMA_PARAMS:?}" \
  -kv \
  --use_sdpa_with_kv_cache \
  -X \
  -d bf16 \
  --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' \
  --output_name="llama3_2.pte"
  -v
```

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -21,6 +21,7 @@ from typing import Callable, List, Optional, Union
 
 import pkg_resources
 import torch
+from executorch.devtools.backend_debug import get_delegation_info
 
 from executorch.devtools.etrecord import generate_etrecord
 from executorch.exir.passes.init_mutable_pass import InitializedMutableBufferPass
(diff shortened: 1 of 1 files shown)


examples/models/llama/export_llama_lib.py
```diff
@@ -816,6 +818,13 @@ def _export_llama(args) -> LLMEdgeManager:  # noqa: C901
 
         builder = builder.to_executorch(passes=additional_passes)
 
+    if args.verbose:
+        graph_module = builder.edge_manager.exported_program().graph_module
```
tarun292, line 822: The `EdgeManager` is modified inplace so after `to_executorch` it inserts `memory.alloc` nodes for out variant ops which are not generated in the final PTE file but used for memory allocation for out-variant ops, so the delegation count given here will be inaccurate. I think we should print this right after `to_backend` above.

iseeyuan: That makes sense. Let me update it. Thanks @tarun292 !

tarun292 (approved):

PR #7804, opened 2025-01-21
Move XNNPACKQuantizer from PyTorch to ExecuTorch

digantdesai:
Summary:
X-link: https://github.com/pytorch/pytorch/pull/144940

This migrates XNNPACKQuantizer from PyTorch to ExecuTorch.

Rationale:
Main motivation is to avoid pytorch pin update in OSS after updating XNNPACKQuantizer, which can be rather frequent.

Other impact and considerations:
- PT2e flow (which lives in PyTorch) relies havily on XNNPACKQuantizer for a "example" implementation for quantizer and more importantly tests. Fow now, we will keep the torch.ao.quantization.xnnpack_quantizer as is but mark is as not BC, and deprecated to discourace future new dependencies on it.

- Other OSS repository using XNNPACKQuantizer from PyTorch now have to take an additional dependency on ExecuTorch.

Differential Revision: D68191752

### DIFF
--- a/backends/cadence/aot/TARGETS
+++ b/backends/cadence/aot/TARGETS
@@ -65,6 +65,7 @@ python_library(
         "//executorch/backends/cadence/aot/quantizer:fusion_pass",
         "//executorch/backends/cadence/runtime:runtime",
         "//executorch/backends/cadence/aot/quantizer:quantizer",
+        "//executorch/backends/xnnpack/quantizer:xnnpack_quantizer",
         "//executorch/backends/transforms:decompose_sdpa",
         "//executorch/backends/transforms:remove_clone_ops",
         "//executorch/exir:lib",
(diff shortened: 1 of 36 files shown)


mcr229 (approved):

mergennachin:
cc @jerryzh168  - FYI

Yeah, I agree ET is the right place for XNNPACKQuantizer since it is backend specific.

> Fow now, we will keep the torch.ao.quantization.xnnpack_quantizer as is but mark is as not BC, and deprecated to discourace future new dependencies on it.

Can you create a github issue to track this. We should mark it as deprecated and warn if people are using (both via lints as well as print error message when people run it). We should follow the 6 months BC breaking window to remove this.

> Other OSS repository using XNNPACKQuantizer from PyTorch now have to take an additional dependency on ExecuTorch.

Are there any dependencies that don't depend on ExecuTorch which uses XNNPACKQuantizer?

Issue #7748, opened 2025-01-17
Unsupported Scalar Type 5? -- Portable/optimized ops don't consistently support half/bfloat16


bluejack:
### 🐛 Describe the bug

After exporting a model to pte form and running it through executor_runner, I get:

```
E 00:00:02.220756 executorch:inputs_portable.cpp:45] Unsupported scalar type 5
```

I believe this is the "Half" type, or float16

Does that simply mean executor_runner does not support float16? Or does the whole framework not support float16?

Noted that when I run some investigation on the file using a python script, I get as far as sending it my float16 tensors, but it still fails to execute with a similar error:

```
[op_native_layer_norm.cpp:169] In function operator()(), assert failed (false): Unhandled dtype Half for native_layer_norm.out
```

I'm including the versions below, but note that this is using executorch built from head, rather than the last release. Should I expect the framework to support float16? And look to my own code for the error?

### Versions

PyTorch version: 2.6.0.dev20250104
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 14.6.1 (arm64)
GCC version: Could not collect
Clang version: 16.0.0 (clang-1600.0.26.4)
CMake version: version 3.31.4
Libc version: N/A

Python version: 3.12.7 | packaged by Anaconda, Inc. | (main, Oct  4 2024, 08:22:19) [Clang 14.0.6 ] (64-bit runtime)
Python platform: macOS-14.6.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Pro

Versions of relevant libraries:
[pip3] executorch==0.6.0a0+cd0e584
[pip3] flake8==7.0.0
[pip3] mypy==1.11.2
[pip3] mypy-extensions==1.0.0
[pip3] numpy==2.0.0
[pip3] numpydoc==1.7.0
[pip3] torch==2.6.0.dev20250104
[pip3] torchao==0.8.0+git2e032c6b
[pip3] torchaudio==2.6.0.dev20250104
[pip3] torchsr==1.0.4
[pip3] torchvision==0.22.0.dev20250104
[...]

PR #7807, opened 2025-01-21
Support BFloat16 in convolution_backward

swolchok:
Partial fix for #7748.


swolchok:
* __->__ #7807

kernels/test/op_convolution_backward_test.cpp
```diff
@@ -62,139 +62,155 @@ class OpConvolutionBackwardOutTest : public OperatorTest {
...
+    auto expected_grad_input = tf.make({2, 6, 7, 5}, expected_grad_input_data);
+    auto expected_grad_weight =
+        tf.make({4, 3, 4, 2}, expected_grad_weight_data);
+    auto expected_grad_bias = tf.make({4}, expected_grad_bias_data);
 
-  auto expected_grad_input = tf.make({2, 6, 7, 5}, expected_grad_input_data);
-  auto expected_grad_weight = tf.make({4, 3, 4, 2}, expected_grad_weight_data);
-  auto expected_grad_bias = tf.make({4}, expected_grad_bias_data);
+    if (DTYPE == ScalarType::Half || DTYPE == ScalarType::BFloat16) {
+      EXPECT_TENSOR_CLOSE_WITH_TOL(grad_input, expected_grad_input, 1e-2, 1e-8);
```
manuelcandales, line 200: Why not use defaults here? `EXPECT_TENSOR_CLOSE_WITH_TOL` should apply the right tolerance given the type

swolchok: because the default rtol is 1e-5; rtol and atol are different

manuelcandales: right, but in the same way that we have `kDefaultHalfAtol` and `kDefaultBFloat16Atol` I think we should have `kDefaultHalfRtol` and `kDefaultBFloat16Rtol` and set it to a proper value.
You seem to be using 1e-2 for most of these tests. Why not introduced `kDefaultHalfRtol` and `kDefaultBFloat16Rtol` with value 1e-2?

swolchok: > Why not introduced kDefaultHalfRtol and kDefaultBFloat16Rtol with value 1e-2?

Because not all operators require the higher rtol.

swolchok: group_norm is one of the ops that automatic mixed precision will autocast to float32: https://intel.github.io/intel-extension-for-pytorch/xpu/1.10.200+gpu/tutorials/features/amp.html

I think the norm ops are just particularly prone to roundoff error, but I'm certainly not a numerical analysis person.

manuelcandales (approved):

PR #7810, opened 2025-01-21
Fix write back to buffers for buffer mutation

cptspacemanspiff:
### Summary

This solves my own issue, #7515, where the write back pass does not add in a copy operation if the buffer mutation src is in the input. 

This change allows one to copy directly from one input to another.


JacobSzwejbka (approved):
Thanks for the fix! When does this come up btw? Why for instance would you not just use x everywhere you would use buffer?

kirklandsign:
@cptspacemanspiff Could you please fix the linter error?

cptspacemanspiff:
@JacobSzwejbka 

So as I was working through how to get shared state between multiple methods working, my debug module to export was a 'set_cache' 'get_cache' operation, which are dummy methods that set and load values from the cache.

A real use case might be to save and load precalculated k-v caches from previously ran prompts and such... The main thing is not having this behavior is very confusing when debugging why your copy operation is missing when you explicitly say copy.

JacobSzwejbka:
Ok the linux test failure is xnnpack being flakey. Im not sure whats up with the phi3 failure but I dont think this is related either so Im just gonna merge and if it gets reverted (unlikely) ill fix it.

PR #7831, opened 2025-01-22
Arm backend: enable dim_order

oscarandersson8218:
### Summary
Removes edge_compile_config.skip_dim_order = True which means that to_copy will be converted into to_dim_order_copy nodes. This commit adds support for to_dim_order_copy.

Change-Id: I46a30cd88125f3e2e26810352a1e0626bd5cce4a

### DIFF
--- a/backends/arm/operator_support/to_copy_support.py
+++ b/backends/arm/operator_support/to_copy_support.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Arm Limited and/or its affiliates.
+# Copyright 2024-2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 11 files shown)


zingo (approved):

backends/arm/operator_support/to_copy_support.py
```diff
@@ -22,7 +22,10 @@
 
 @register_tosa_support_check
 class ToCopySupported(SupportedTOSAOperatorCheck):
-    targets = [exir_ops.edge.aten._to_copy.default]
+    targets = [
```
digantdesai, line 25: I guess we don't lower empty?

digantdesai (approved):

backends/arm/operator_support/to_copy_support.py
oscarandersson8218: Not as of today.

zingo:
Failing test seem unrealted to PR and fails on other PR's also.

PR #7836, opened 2025-01-22
Use default overloads when calling custom ops

Erik-Lundell:
If a node is created without specifying an overload, A OpOverloadPacket is created, rather than an OpOverload. This works in a GraphModule, but the OpOverloadPacket is not a valid operator type in the _EXIREdgeDialectVerifier, which means that Edge ExportedPrograms can't contain a GraphModule with such ops.

In short, specifying using the default overload seems to be the more correct way of calling a custom
operator.

### DIFF
--- a/backends/arm/scripts/pre-commit
+++ b/backends/arm/scripts/pre-commit
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Check 1: If commit header contains WIP, everything is ok
+git rev-list --format=%s --max-count=1 HEAD | grep -q WIP && exit 0
+
+# Check 2: lintunner on latest patch.
+lintrunner -a --revision 'HEAD^' --skip MYPY
+commit_files=$(git diff-tree --no-commit-id --name-only --diff-filter=M HEAD -r)
+git add $commit_files || true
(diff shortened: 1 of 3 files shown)


Erik-Lundell:
CI Failure unrelated.

zingo (approved):

backends/arm/_passes/insert_table_ops.py
```diff
@@ -104,7 +104,11 @@ def call(self, graph_module: GraphModule) -> PassResult:
                     out_quantargs=output_qparams[0],
                 )
                 # Register buffer in self.exported_program.state_dict
-                self.register_buffer(buffer_name=table_node.name, buffer=buffer)
+                # When the graph is retraced, the implementation _table is used and the suffix _default disappears from the node name
+                # Remove it here to make it possible to find in the node_visitor
+                self.register_buffer(
```
digantdesai, line 109: This is OK but not the cleanest I guess.

digantdesai (approved):
IIUC https://github.com/pytorch/executorch/blob/31f35d9bed8f77524fc702cbc9826f20e857c83f/exir/verification/verifier.py#L68 allows OpOverloadPacket for torch ops but not for custom ops? Is that the main problem?

If yes, this is OK.

Erik-Lundell:
> IIUC
> 
> https://github.com/pytorch/executorch/blob/31f35d9bed8f77524fc702cbc9826f20e857c83f/exir/verification/verifier.py#L68
> allows OpOverloadPacket for torch ops but not for custom ops? Is that the main problem?
> 
> If yes, this is OK.

I think https://github.com/pytorch/executorch/blob/31f35d9bed8f77524fc702cbc9826f20e857c83f/exir/verification/verifier.py#L239 is the relevent line but yes, that is the main problem

PR #7839, opened 2025-01-22
Remove cadence submodule from top-level dependencies

### LINKED ISSUE #7081: nnlib-hifi4 is too big during setting up ExecuTorch from source

During slow internet connection, it takes a long time to set up ExecuTorch. This is due to download all gitsubmodules. Here's the root level filesizes (after `git submodule sync; git submodule update --init`)

```
708M	backends/
172K	build/
128K	codegen/
 12K	configurations/
540K	devtools/
 31M	docs/
 59M	examples/
2.3M	exir/
 74M	extension/
 24M	kernels/
 40K	profiler/
1.4M	runtime/
 60K	schema/
 32K	scripts/
176K	shim/
168K	test/
 42M	third-party/
 24K	util/
 ```
 
 Within backends/ folder, here's the breakdown: 

```
1.6M	apple/
4.2M	arm/
485M	cadence/
960K	example/
228K	mediatek/
1.3M	qualcomm/
 96K	transforms/
 23M	vulkan/
192M	xnnpack/
```

The main culprit is this library: https://github.com/pytorch/executorch/tree/main/backends/cadence/hifi/third-party/nnlib

Is it possible move cadence related to installation into its own set up script, like we do for other backends like coreml? For example:  https://github.com/pytorch/executorch/blob/main/backends/apple/coreml/scripts/install_requirements.sh

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

### PR DESCRIPTION
**Pull request by mergennachin**
Summary:

Every time we clone/set up ExecuTorch, we're downloading cadence dependencies.

Let's make this an on-demand step. Otherwise, this is downloading almost 1GB of code.

Fixes: https://github.com/pytorch/executorch/issues/7081

Before and after repo size: https://gist.github.com/mergennachin/f23ea67f9ba950d6f0bf3797d82e139d

### DIFF
--- a/backends/cadence/install_requirements.sh
+++ b/backends/cadence/install_requirements.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR_PATH="$(
+    cd -- "$(dirname "$0")" >/dev/null 2>&1
+    pwd -P
+)"
+
+red=`tput setaf 1`
+green=`tput setaf 2`
+
+EXECUTORCH_ROOT_PATH=$(realpath "$SCRIPT_DIR_PATH/../../")
+CADENCE_DIR_PATH="$EXECUTORCH_ROOT_PATH/backends/cadence"
+HIFI_DIR_PATH="$CADENCE_DIR_PATH/hifi/third-party/nnlib/nnlib-hifi4"
+FUSION_DIR_PATH="$CADENCE_DIR_PATH/fusion_g3/third-party/nnlib/nnlib-FusionG3"
+
+cd "$EXECUTORCH_ROOT_PATH"
+
+## HiFi 
+
+rm -rf "$HIFI_DIR_PATH"
+mkdir -p "$HIFI_DIR_PATH"
+
+echo "${green}ExecuTorch: Cloning hifi nnlib"
+git clone "https://github.com/foss-xtensa/nnlib-hifi4.git" $HIFI_DIR_PATH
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: backends/cadence/install_requirements.sh [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+echo "${green}ExecuTorch: Cloning fusion g3"
+git clone "https://github.com/foss-xtensa/nnlib-FusionG3.git" $FUSION_DIR_PATH
+cd $FUSION_DIR_PATH
+STATUS=$?
+if [ $STATUS -ne 0 ]; then
+    echo "${red}ExecuTorch: Failed to clone fusion g3."
+    exit 1
+fi
+
+git checkout 11230f47b587b074ba0881deb28beb85db566ac2
```
mergennachin, line 49: btw, current submodule approach is pointing to an orphan commit, which is not a good in terms of maintainence

https://github.com/foss-xtensa/nnlib-FusionG3/tree/8ddd1c39d4b20235ebe9dac68d92848da2885ece

instead, i am now pointing the latest one in main branch, feel free to suggest a change.

https://github.com/foss-xtensa/nnlib-FusionG3/commits/main/

tarun292 (approved):

mcremon-meta (approved):
Looks good to me, thanks a lot @mergennachin for taking care of this. Linking to current main commit is fine by me, if we need to change that it should be trivial anyway.
cc @hsharma35 and @zonglinpeng

PR #7841, opened 2025-01-22
add cpp serializer for flat tensor (ptd)

JacobSzwejbka:
Summary:
Leverage the flatbuffer builder apis to generate a .ptd on device. This will be used by training for checkpointing.

No other use cases really exist for generating a .ptd on device right now, so I didnt worry about making this more easily extensible by coming up with a cpp equivalent of a cord or by trying to think ahead of how this might integrate with delegates.

Later if we add support for delegates owning the weights under ET training we can revisit this.

Differential Revision: D67992901

### DIFF
--- a/extension/flat_tensor/serialize/serialize.cpp
+++ b/extension/flat_tensor/serialize/serialize.cpp
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/extension/flat_tensor/serialize/serialize.h>
+
+#include <executorch/extension/flat_tensor/serialize/flat_tensor_header.h>
+#include <executorch/extension/flat_tensor/serialize/scalar_type_generated.h>
+#include <executorch/extension/flat_tensor/serialize/schema_generated.h>
+
+#include <fstream>
+#include <string>
+
+namespace executorch {
+namespace extension {
+namespace flat_tensor {
+
+namespace {
+size_t padding_required(size_t offset, size_t alignment) {
+  // Returns the padding required to align `offset` to `alignment`.
+  size_t remainder = offset % alignment;
+  if (remainder != 0) {
+    return alignment - remainder;
+  }
+  return 0;
+}
+
+size_t aligned_size(size_t input_size, size_t alignment) {
+  // Returns input_size padded up to the next whole multiple of alignment.
+  return input_size + padding_required(input_size, alignment);
+}
+
+void write_nulls(std::ostream& out, size_t num_bytes) {
+  for (size_t i = 0; i < num_bytes; i++) {
+    out.write("\0", 1);
+  }
+}
+} // namespace
+
+runtime::Error save_ptd(
+    const std::string& path,
+    const std::map<std::string, exec_aten::Tensor>& tensor_map,
+    const size_t tensor_alignment) {
+  // Create File
+  std::ofstream file;
+  file.open(path);
+  runtime::Error e = save_ptd(file, tensor_map, tensor_alignment);
+  file.close();
+  return e;
+}
+
+runtime::Error save_ptd(
+    std::ostream& out,
+    const std::map<std::string, exec_aten::Tensor>& tensor_map,
+    const size_t tensor_alignment) {
+  // Assert the system is little endian. Since we are sending the data over
+  // the wire, we need to ensure that the data is always in the same format.
+  // for now we only support little endian.
+  int n = 1;
+  if (*(char*)&n != 1) {
+    ET_LOG(Error, "Cannot save_ptd on big endian system");
+    return runtime::Error::NotSupported;
+  }
+  // Create flatbuffer
+  flatbuffers::FlatBufferBuilder builder;
+
+  std::vector<flatbuffers::Offset<::flat_tensor_flatbuffer::TensorMetadata>>
+      tensors;
+  std::vector<flatbuffers::Offset<::flat_tensor_flatbuffer::DataSegment>>
+      buffers;
+
+  // Write the tensors.
+  size_t total_segment_size = 0;
+  size_t i = tensor_map.size();
+  for (const auto& [name, tensor] : tensor_map) {
+    auto name_offset = builder.CreateString(name);
+    // Write the tensor metadata.
+    auto tensor_metadata = ::flat_tensor_flatbuffer::CreateTensorMetadata(
+        builder,
+        name_offset,
+        static_cast<executorch_flatbuffer::ScalarType>(tensor.scalar_type()),
+        builder.CreateVector(tensor.sizes().data(), tensor.sizes().size()),
+        builder.CreateVector(
+            tensor.dim_order().data(), tensor.dim_order().size()),
+        0, // segment index
+        total_segment_size);
+
+    tensors.push_back(tensor_metadata);
+    // Don't pad last entry.
+    if (i != 1) {
+      // Precalculate the size of the data blob.
+      total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
+    } else {
+      total_segment_size += tensor.nbytes();
+    }
+    i--;
+  }
+  // Only have one segment
+  buffers.push_back(::flat_tensor_flatbuffer::CreateDataSegment(
+      builder, 0, total_segment_size));
+
+  auto flat_tensor = CreateFlatTensor(
+      builder,
+      kSchemaVersion,
+      tensor_alignment,
+      builder.CreateVector(tensors),
+      builder.CreateVector(buffers));
+  builder.Finish(flat_tensor); // Our flatbuffer is created now.
+
+  // Calculate flatbuffer padding.
+  auto padded_flatbufer_size =
+      aligned_size(builder.GetSize(), tensor_alignment);
+  auto padded_header_size =
+      aligned_size(FlatTensorHeader::kHeaderExpectedLength, tensor_alignment);
+
+  // Write header
+  out.write(FlatTensorHeader::kMagic, sizeof(FlatTensorHeader::kMagic));
+  out.write(
+      reinterpret_cast<const char*>(&FlatTensorHeader::kHeaderExpectedLength),
+      sizeof(FlatTensorHeader::kHeaderExpectedLength));
+
+  FlatTensorHeader header = {
+      padded_header_size, // Offset to flatbuffer
+      builder.GetSize(), // flatbuffer size
+      padded_header_size + padded_flatbufer_size, // offset to segments
+      total_segment_size // segment data size
+  };
+
+  out.write(
+      reinterpret_cast<const char*>(&header.flatbuffer_offset),
+      sizeof(header.flatbuffer_offset));
+  out.write(
+      reinterpret_cast<const char*>(&header.flatbuffer_size),
+      sizeof(header.flatbuffer_size));
+  out.write(
+      reinterpret_cast<const char*>(&header.segment_base_offset),
+      sizeof(header.segment_base_offset));
+  out.write(
+      reinterpret_cast<const char*>(&header.segment_data_size),
+      sizeof(header.segment_data_size));
+
+  // Write header padding
+  write_nulls(
+      out,
+      padding_required(
+          FlatTensorHeader::kHeaderExpectedLength, tensor_alignment));
+
+  // Write flatbuffer
+  out.write(
+      reinterpret_cast<const char*>(builder.GetBufferPointer()),
+      builder.GetSize());
+
+  // Write flatbuffer padding
+  write_nulls(out, padding_required(builder.GetSize(), tensor_alignment));
+
+  // Write segment: buffers + tensor padding
+  i = tensor_map.size();
+  for (const auto& [name, tensor] : tensor_map) {
+    out.write(
+        reinterpret_cast<const char*>(tensor.data_ptr()), tensor.nbytes());
+    // Don't pad last entry.
+    if (i != 1) {
+      write_nulls(out, padding_required(tensor.nbytes(), tensor_alignment));
+    }
+    i--;
+  }
+  return runtime::Error::Ok;
+}
+
+} // namespace flat_tensor
+} // namespace extension
+} // namespace executorch
--- a/extension/flat_tensor/serialize/serialize.h
+++ b/extension/flat_tensor/serialize/serialize.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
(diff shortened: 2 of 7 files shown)


extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <string>
+#include <fstream>
+
+namespace executorch {
+namespace extension {
+namespace flat_tensor {
+
+namespace {
+  size_t padding_required(size_t offset, size_t alignment) {
```
dbort, line 21: Code shouldn't be indented inside a namespace

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+    // Returns input_size padded up to the next whole multiple of alignment.
+    return input_size + padding_required(input_size, alignment);
+  }
+
+  void serialize_nulls(std::ostream& out, size_t num_bytes) {
+    for (size_t i = 0; i < num_bytes; i++) {
+      out.write("\0", 1);
+    }
+  }
+}
```
dbort, line 40:
```suggestion
} // namespace
```

extension/flat_tensor/serialize/serialize.h
```diff
@@ -0,0 +1,67 @@
...
+#include <executorch/runtime/core/exec_aten/exec_aten.h>
+
+#include <string>
+#include <map>
+
+namespace executorch {
+namespace extension {
+namespace flat_tensor {
+namespace internal {
+  // cpp mirror of the python definition in serialize.py
```
dbort, line 20: Code shouldn't be indented inside a namespace

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <string>
+#include <fstream>
+
+namespace executorch {
+namespace extension {
+namespace flat_tensor {
+
+namespace {
+  size_t padding_required(size_t offset, size_t alignment) {
+    // Returns the padding required to align `offset` to `alignment`.
```
dbort, line 22: This should be a comment above the function, not inside the function. Same for the next one

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  void serialize_nulls(std::ostream& out, size_t num_bytes) {
+    for (size_t i = 0; i < num_bytes; i++) {
+      out.write("\0", 1);
+    }
+  }
+}
+
+ET_EXPERIMENTAL runtime::Error save_ptd(
```
dbort, line 42: The EXPERIMENTAL annotations need to be in the .h file not the .cpp file, otherwise call sites won't emit warnings for using them

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+  const std::string& path,
+  const std::map<std::string, exec_aten::Tensor>& tensor_map,
+  const size_t tensor_alignment
+) {
+  // Assert the system is little endian
```
dbort, line 47: Not really necessary because the other save_ptd will do the same when this calls it

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+  file.close();
+  return e;
+}
+
+ET_EXPERIMENTAL runtime::Error save_ptd(
+  std::ostream& out,
+  const std::map<std::string, exec_aten::Tensor>& tensor_map,
+  const size_t tensor_alignment
+) {
+  // Assert the system is little endian
```
dbort, line 65: Please explain why you need to make this check.

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+      name_offset,
+      static_cast<executorch_flatbuffer::ScalarType>(tensor.scalar_type()),
+      builder.CreateVector(tensor.sizes().data(), tensor.sizes().size()),
+      builder.CreateVector(tensor.dim_order().data(), tensor.dim_order().size()),
+      0, // segment index
+      total_segment_size);
+    tensors.push_back(tensor_metadata);      // Precalculate the size of the data blob
+    total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
+  }
+  // Only have one segment
```
dbort, line 92: Why is there only one segment? Seems like we'd want to be able to load and free large tensors independently

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+  buffers.push_back(::flat_tensor::CreateDataSegment(builder, 0, total_segment_size));
+  
+  auto flat_tensor = CreateFlatTensor(builder, internal::VERSION, tensor_alignment, builder.CreateVector(tensors), builder.CreateVector(buffers));
```
dbort, line 95: Longer than 80 columns, probably needs a lint reformat

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+  builder.Finish(flat_tensor);  // Our flatbuffer is created now.
+
+
+  // Calculate flatbuffer padding.
+  auto padded_flatbufer_size = aligned_size(builder.GetSize(), tensor_alignment);
+  auto padded_header_size = aligned_size(internal::HEADER_EXPECTED_LENGTH, tensor_alignment);
+
+  // Serialize header
+  out.write(internal::HEADER_MAGIC, sizeof(internal::HEADER_MAGIC));
```
dbort, line 104: This is arranged differently from PTE files. I hid the extended header inside the flatbuffer data so that a PTE file can still be read by standard flatbuffer tools, which has been really useful.

See https://github.com/pytorch/executorch/blob/2411c9913309914ce8dd6db7e17b96ef85907bd6/exir/_serialize/_program.py#L55 for that logic

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,138 @@
...
+    padded_header_size + padded_flatbufer_size, // offset to segments
+    total_segment_size // segment data size
+  };
+
+  out.write(reinterpret_cast<const char *>(&header.flatbuffer_offset), sizeof(header.flatbuffer_offset));
+  out.write(reinterpret_cast<const char *>(&header.flatbuffer_size), sizeof(header.flatbuffer_size));
+  out.write(reinterpret_cast<const char *>(&header.segment_base_offset), sizeof(header.segment_base_offset));
+  out.write(reinterpret_cast<const char *>(&header.segment_data_size), sizeof(header.segment_data_size));
+
+  // serialize header padding
```
dbort, line 119: Super nit: these steps aren't serializing, they're writing. Serializing is the act of arranging the data in a particular format, regardless of where that data lives. These steps are writing out already-serialized data.

dbort (changes requested):

extension/flat_tensor/serialize/serialize.cpp
JacobSzwejbka: Thats the behavior of the python serializer today. I was just mimicking it. Theres not an immediate need to separate today, it will come if we start combining mutable and non mutable weights in the same .ptd for training, or with delegates presumably. Can add that functionality later though with no bc breaks.

extension/flat_tensor/test/test_serialize.cpp
```diff
@@ -0,0 +1,115 @@
...
+      *(uint32_t*)(byte_buffer + 4),
+      executorch::extension::FlatTensorHeader::kHeaderExpectedLength);
+  ASSERT_EQ(
+      *(uint64_t*)(byte_buffer + 8),
+      48); // Header is 40 bytes + 8 bytes of padding today, and then the
+           // flatbuffer starts.
+  ASSERT_EQ(
+      *(uint64_t*)(byte_buffer + 16),
+      224); // This is fragile, and depends on the schema, the builder, and the
+            // padding needed.
```
dbort, line 73: Please mention which fields these are. Flatbuffer offset + flatbuffer length?

extension/flat_tensor/test/test_serialize.cpp
```diff
@@ -0,0 +1,115 @@
...
+  const uint64_t segment_offset = 48 +
+      224; // Segment offset, depends on the padded header and flatbuffer sizes.
+  ASSERT_EQ(*(uint64_t*)(byte_buffer + 24), segment_offset);
+  ASSERT_EQ(
+      *(uint64_t*)(byte_buffer + 32),
+      32); // Segment total size, 8 bytes of data (2 floats), 24 bytes of
```
dbort, line 79: If there's only one segment, there shouldn't be any padding after it: there should only be padding between adjacent segments

extension/flat_tensor/test/test_serialize.cpp
```diff
@@ -0,0 +1,115 @@
...
+           // padding.
+
+  // Check Flatbuffer
+  auto flat_tensor = ::flat_tensor::GetFlatTensor(byte_buffer + 48);
```
dbort, line 83: Once the extended header is hidden properly, you should be able to load the flatbuffer starting at offset zero of byte_buffer.

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,160 @@
...
+        name_offset,
+        static_cast<executorch_flatbuffer::ScalarType>(tensor.scalar_type()),
+        builder.CreateVector(tensor.sizes().data(), tensor.sizes().size()),
+        builder.CreateVector(
+            tensor.dim_order().data(), tensor.dim_order().size()),
+        0, // segment index
+        total_segment_size);
+    tensors.push_back(
+        tensor_metadata); // Precalculate the size of the data blob
+    total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
```
dbort, line 89:
```suggestion
        tensor_metadata);
    // Precalculate the size of the data blob
    total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
```

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,160 @@
...
+      builder.GetSize());
+
+  // Write flatbuffer padding
+  write_nulls(out, padding_required(builder.GetSize(), tensor_alignment));
+
+  // Write segment: buffers + tensor padding
+  for (const auto& [name, tensor] : tensor_map) {
+    out.write(
+        reinterpret_cast<const char*>(tensor.data_ptr()), tensor.nbytes());
+    write_nulls(out, padding_required(tensor.nbytes(), tensor_alignment));
```
dbort, line 153: The final entry doesn't need padding after it

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,160 @@
...
+      padding_required(
+          FlatTensorHeader::kHeaderExpectedLength, tensor_alignment));
+
+  // Write flatbuffer
+  out.write(
+      reinterpret_cast<const char*>(builder.GetBufferPointer()),
+      builder.GetSize());
+
+  // Write flatbuffer padding
+  write_nulls(out, padding_required(builder.GetSize(), tensor_alignment));
```
dbort, line 147: Really, only stuff after this point needs to be aligned to tensor_alignment; the header and flatbuffer data could just be aligned to 8. PTE aligns its flatbuffer data because the flatbuffer actually contains tensor buffers, but we don't do that in this format.

Aligning to a larger value is fine, though. And we could change the earlier alignments at any time without breaking BC

dbort (changes requested):
Thanks for the changes -- the tests look great

extension/flat_tensor/test/test_serialize.cpp
JacobSzwejbka: Yeah I update the test in the following diff.

extension/flat_tensor/serialize/serialize.cpp
JacobSzwejbka: Ah yeah Im matching the python behavior here which just padded everything to tensor alignment.

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,173 @@
...
+            tensor.dim_order().data(), tensor.dim_order().size()),
+        0, // segment index
+        total_segment_size);
+
+    tensors.push_back(tensor_metadata);
+    // Don't pad last entry.
+    if (i != 1) {
+      // Precalculate the size of the data blob.
+      total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
+    }
```
dbort, line 97: Missing the `--i` like in the loop below. Please add a test that would have caught these issues.

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -0,0 +1,173 @@
...
+        builder.CreateVector(
+            tensor.dim_order().data(), tensor.dim_order().size()),
+        0, // segment index
+        total_segment_size);
+
+    tensors.push_back(tensor_metadata);
+    // Don't pad last entry.
+    if (i != 1) {
+      // Precalculate the size of the data blob.
+      total_segment_size += aligned_size(tensor.nbytes(), tensor_alignment);
```
dbort, line 96: If there's only one tensor, total_segment_size will be zero. I think you need to increment the size by the unpadded amount for the final entry.

extension/flat_tensor/serialize/serialize.h
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+namespace executorch {
+namespace extension {
+namespace flat_tensor {
+
+/**
+ *
+ * Schema version of the .ptd format. Should be kept in sync with Python.
+ * https://github.com/pytorch/executorch/blob/main/extension/flat_tensor/serialize/serialize.py#L269
```
dbort, line 23: The problem with pointing to a specific line is that it won't be the right line after serialize.py changes over time. I'd just point to the file in general. And since it's in the same dir you could skip the full URL

```suggestion
 * Schema version of the .ptd format. Should be kept in sync with the same value in serialize.py.
```

extension/flat_tensor/serialize/serialize.py
```diff
@@ -266,7 +266,7 @@ def serialize(
         # Create FlatTensor, which describes of the contents of the file and
         # points to all the data segments. It will be serialized to flatbuffer.
         flat_tensor = FlatTensor(
-            version=0,
+            version=0,  # Keep in sync with c++ version number in serialize.h
```
dbort, line 269: Consider adding a top-level `_SCHEMA_VERSION: int = 0` to raise visibility of this, and to make it map more clearly to the C++ equivalent.

dbort (approved):
Thanks for all of the changes -- looks great apart from the new bugs.

PR #7846, opened 2025-01-22
Support Half/BFloat16 in native_group_norm (needs accuracy fix)

swolchok:


swolchok:
* __->__ #7846

manuelcandales (approved):

kernels/test/op_native_group_norm_test.cpp
```diff
@@ -20,110 +20,140 @@ using exec_aten::ScalarType;
...
+         0.38669431,
+         0.38038814,
+         0.75809801});
+    op_native_group_norm_out(
+        input, weight, bias, 5, 6, 4, 3, eps, out0, out1, out2);
+    if (DTYPE == ScalarType::Half || DTYPE == ScalarType::BFloat16) {
+      EXPECT_TENSOR_CLOSE_WITH_TOL(
+          out0,
+          out0_expected,
+          2e-1,
```
manuelcandales, line 135: That's 20%. This doesn't make me feel comfortable.

manuelcandales (approved):

kernels/test/op_native_group_norm_test.cpp
swolchok: group_norm is one of the ops that automatic mixed precision will autocast to float32: https://intel.github.io/intel-extension-for-pytorch/xpu/1.10.200+gpu/tutorials/features/amp.html

I think the norm ops are just particularly prone to roundoff error, but I'm certainly not a numerical analysis person.

swolchok: Dug in a little further. Here's the PR that originally made PyTorch group_norm support Half: https://github.com/pytorch/pytorch/pull/100234/files#diff-7927db349f568afca2de9b94d74ea5c3b8cb468cb6a433d0cc1e61e65c515a36

It looks like the test is atol=rtol=5e-3. I think it's reasonable to argue that if we can't get the tolerances to be broadly similar then we have a correctness issue and thus don't actually support Half. I'll see what I can do; this one might have to wait for code sharing.

swolchok: test doesn't pass with `atol=rtol=5e-3`. Holding off on group_norm until we have code sharing.

swolchok: > until we have code sharing.

this is a portable op, so code sharing is not applicable. Not sure what we should do about it going forward.

swolchok:
looks like the op supports Half/BFloat16 now.

PR #7856, opened 2025-01-22
Coerce to true_ctype in tensor_factory

swolchok:
This should fix the problem where attempts to test `bool` are often wonky in OSS and fail UBSAN internally; it is undefined behavior to store a value other than 0 or 1 for type `bool`.

### DIFF
--- a/runtime/core/exec_aten/testing_util/tensor_factory.h
+++ b/runtime/core/exec_aten/testing_util/tensor_factory.h
@@ -279,7 +279,10 @@ class TensorFactory {
       t = empty_strided(sizes, strides);
     }
     if (t.nbytes() > 0) {
-      memcpy(t.template data<true_ctype>(), data.data(), t.nbytes());
+      std::transform(
+          data.begin(), data.end(), t.template data<true_ctype>(), [](auto x) {
+            return static_cast<true_ctype>(x);
+          });
     }
     return t;
   }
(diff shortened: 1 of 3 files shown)


swolchok:
* (to be filled)

runtime/core/exec_aten/testing_util/tensor_factory.h
```diff
@@ -1019,7 +1032,14 @@ class TensorFactory {
               data_.data(),
               dim_order_.data(),
               strides_.data(),
-              dynamism) {}
+              dynamism) {
+      // The only valid values for bool are 0 and 1; coerce!
+      if constexpr (std::is_same_v<true_ctype, bool>) {
```
kimishpatel, line 1037: It seems in `ifdef aten` section this transformation is unconditional while here it is applied only if it is bool. I dont fully understand.

swolchok: it doesn't build with bits16 and we don't have bit_cast, otherwise I would make it unconditional here and use bit_cast.

kimishpatel: I am guessing bit_cast not being usable is related to it available only in c++20.

kimishpatel (approved):

runtime/core/exec_aten/testing_util/tensor_factory.h
swolchok: yes; we will have access to c10::bit_cast once I get back to code sharing but currently we don't have it.

PR #7857, opened 2025-01-22
Support Half/BFloat16 in prod operator

swolchok:
Partial fix for #7748.

### DIFF
--- a/kernels/test/op_prod_test.cpp
+++ b/kernels/test/op_prod_test.cpp
@@ -45,6 +45,24 @@ class OpProdOutTest : public ::testing::Test {
     // first.
     torch::executor::runtime_init();
   }
+
+  template <ScalarType DTYPE>
+  void test_dtype() {
+    TensorFactory<DTYPE> tf;
+    TensorFactory<
+        executorch::runtime::isIntegralType(DTYPE, /*includeBool*/ true)
+            ? ScalarType::Long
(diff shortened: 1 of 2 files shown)


swolchok:
* (to be filled)

kernels/test/op_prod_test.cpp
```diff
@@ -54,30 +72,32 @@ class OpProdIntOutTest : public ::testing::Test {
...
-  Tensor self = tfFloat.make({2, 3}, {1, 2, 3, 4, 5, 6});
-  optional<ScalarType> dtype{};
-  Tensor out = tfFloat.zeros({});
-  Tensor out_expected = tfFloat.make({}, {720});
-  op_prod_out(self, dtype, out);
-  EXPECT_TENSOR_CLOSE(out, out_expected);
-}
+    Tensor self = tf.make({2, 3}, {1, 2, 3, 4, 5, 6});
+    int64_t dim = 0;
+    bool keepdim = false;
```
kimishpatel, line 82: has there been a test with keepdim=True?

kimishpatel (approved):

kernels/test/op_prod_test.cpp
swolchok: I don't know, but adding such a test would be out of scope for this change.

PR #7862, opened 2025-01-23
Support Half/BFloat16 in round

swolchok:
Partial fix for #7748.


manuelcandales (approved):

swolchok:
checking CI specifically because the unittest jobs are failing; OK to land when they're green

PR #7868, opened 2025-01-23
[DevX] Skip disabled configs for benchmarking

guangy10:
Some benchmark jobs have been failing for a long time, creating misleading signal about what expected and unexpected by looking at those jobs. For example:
 - On Android: https://hud.pytorch.org/hud/pytorch/executorch/main/2?per_page=50&name_filter=mobilebert%2C%20qnn_q8&mergeLF=true
 - On Apple: https://hud.pytorch.org/hud/pytorch/executorch/main/2?per_page=50&name_filter=mobilebert%2C%20mps&mergeLF=true

The reason is that those configs are either not supported or having bugs to be fixed. There is no way to specify which config for which model should be disabled/bypassed in the past. 

This PR adds the ability to do so. Going forward, the nightly benchmark jobs (`android-perf` and `apple-perf`) should always be green, if not, it clearly signals that some configs are failing/flaky and should be fixed. 

This PR also set a process to track disabled model+config in the GitHub Issue, the test will catch it if trying to disable a config for a model but not providing a link to GitHub Issue. The enforcement is to make sure regressions is tracked properly.

Test:
 - Android: `qnn_q8` config is skipped for `w2l`: https://github.com/pytorch/executorch/actions/runs/12920378484
 - Apple: `mps` config is skipped for `mobilebert`: https://github.com/pytorch/executorch/actions/runs/12920397985

```python
python .ci/scripts/tests/test_gather_benchmark_configs.py
```
```
----------------------------------------------------------------------
Ran 14 tests in 0.139s

OK
```

### DIFF
--- a/.ci/scripts/gather_benchmark_configs.py
+++ b/.ci/scripts/gather_benchmark_configs.py
@@ -10,7 +10,7 @@ import logging
 import os
 import re
 import sys
-from typing import Any, Dict, List
+from typing import Any, Dict, List, NamedTuple
 
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "../..")))
 from examples.models import MODEL_NAME_TO_MODEL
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!

guangy10:
Subscribed @kimishpatel @digantdesai to DevX related work.

PR #7870, opened 2025-01-23
TensorLayout updates

lucylq:
Summary: Addressing comments from D67048723

Differential Revision: D68535453

### DIFF
--- a/runtime/core/tensor_layout.cpp
+++ b/runtime/core/tensor_layout.cpp
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/runtime/core/exec_aten/exec_aten.h>
+#include <executorch/runtime/core/exec_aten/util/scalar_type_util.h>
+#include <executorch/runtime/core/span.h>
+#include <executorch/runtime/core/tensor_layout.h>
+
+namespace executorch {
+namespace runtime {
+
+namespace {
+Result<size_t> calculate_nbytes(
+    const Span<const int32_t>& sizes,
+    const exec_aten::ScalarType& scalar_type) {
+  ssize_t n = 1;
+  for (ssize_t i = 0; i < sizes.size(); i++) {
+    if (sizes[i] < 0) {
+      return Error::InvalidArgument;
+    }
+    n *= sizes[i];
+  }
+  // Use the full namespace to disambiguate from c10::elementSize.
+  return n * executorch::runtime::elementSize(scalar_type);
+}
+} // namespace
+
+Result<TensorLayout> TensorLayout::create(
+    Span<const int32_t> sizes,
+    Span<const uint8_t> dim_order,
+    executorch::aten::ScalarType scalar_type) {
+  auto nbytes = calculate_nbytes(sizes, scalar_type);
+  if (!nbytes.ok()) {
+    return nbytes.error();
+  }
+
+  if (dim_order.size() != sizes.size()) {
+    return Error::InvalidArgument;
+  }
+
+  for (size_t i = 0; i < dim_order.size(); i++) {
+    if (dim_order[i] >= sizes.size()) {
+      return Error::InvalidArgument;
+    }
+  }
+  return TensorLayout(sizes, dim_order, scalar_type, nbytes.get());
+}
+} // namespace runtime
+} // namespace executorch
--- a/runtime/core/tensor_layout.h
+++ b/runtime/core/tensor_layout.h
@@ -10,55 +10,48 @@
 
 #include <executorch/runtime/core/exec_aten/exec_aten.h>
 #include <executorch/runtime/core/exec_aten/util/scalar_type_util.h>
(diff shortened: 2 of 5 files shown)


runtime/core/tensor_layout.cpp
```diff
@@ -0,0 +1,45 @@
...
+  }
+  // Use the full namespace to disambiguate from c10::elementSize.
+  return n * executorch::runtime::elementSize(scalar_type);
+}
+} // namespace
+
+/**
+ * NOTE: the spans (sizes, dim_order) passed here must outlive the
+ * TensorLayout and all copies of it.
+ */
```
dbort, line 34: This is a message for the user creating the TensorLayout, and should go in the header.

runtime/core/tensor_layout.cpp
```diff
@@ -0,0 +1,45 @@
...
+namespace executorch {
+namespace runtime {
+
+namespace {
+size_t calculate_nbytes(
+    const Span<const int32_t>& sizes,
+    const exec_aten::ScalarType& scalar_type) {
+  ssize_t n = 1;
+  for (ssize_t i = 0; i < sizes.size(); i++) {
+    ET_CHECK(sizes[i] >= 0);
```
dbort, line 23: This will crash if `sizes` contains a negative value, which we need to avoid. See my comment on the .h file about using a static factory, which would then let this helper return `Result<size_t>` to handle the error case non-fatally.

Our core tensor type suffers from the same problem, which is why the runtime has to do an [awkward pre-check](https://github.com/pytorch/executorch/blob/9bb9ad0564453bdbc656ea3502656da9eaa38c28/runtime/executor/tensor_parser_portable.cpp#L108) before creating tensors.

dbort (changes requested):
Thanks for responding to my comments on the previous PR!

runtime/core/tensor_layout.h
```diff
@@ -15,38 +15,19 @@
...
  *
- * The NamedDataMap used to create the TensorLayout must outlive the
- TensorLayout.
+ * NOTE: the underlying data structure/s backing the TensorLayout must outlive
+ * the TensorLayout and all copies of it.
  */
-class TensorLayout {
+class ET_EXPERIMENTAL TensorLayout final {
  public:
   TensorLayout(
```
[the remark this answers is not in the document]
lucylq: thanks for explaining this!

runtime/core/tensor_layout.h
```diff
@@ -10,55 +10,47 @@
...
  public:
-  TensorLayout(
-      executorch::aten::ScalarType scalar_type,
-      Span<const int32_t> sizes,
-      Span<const uint8_t> dim_order)
-      : sizes_(sizes),
-        dim_order_(dim_order),
-        scalar_type_(scalar_type),
-        nbytes_(calculate_nbytes(sizes_, scalar_type_)) {}
+  TensorLayout() = delete;
```
dbort, line 24: I don't think you need to do this since you declare a non-default ctor.

https://en.cppreference.com/w/cpp/language/default_constructor
> ### Implicitly-declared default constructor
> If there is no user-declared constructor or constructor template for a class type, the compiler will implicitly declare a default constructor as an inline public member of its class.

runtime/core/tensor_layout.h
```diff
@@ -10,55 +10,47 @@
...
 
-  TensorLayout(const TensorLayout&) = default;
-  TensorLayout(TensorLayout&&) = default;
-  TensorLayout& operator=(const TensorLayout&) = default;
-  TensorLayout& operator=(TensorLayout&& other) = default;
-  ~TensorLayout() = default;
+  /** Creates a TensorLayout from the given parameters.
+   *
```
dbort, line 27: nit
```suggestion
  /**
   * Creates a TensorLayout from the given parameters.
```

runtime/core/tensor_layout.cpp
```diff
@@ -0,0 +1,44 @@
...
+    n *= sizes[i];
+  }
+  // Use the full namespace to disambiguate from c10::elementSize.
+  return n * executorch::runtime::elementSize(scalar_type);
+}
+} // namespace
+
+Result<TensorLayout> TensorLayout::create(
+    Span<const int32_t> sizes,
+    Span<const uint8_t> dim_order,
```
dbort, line 35: It'd also be a good idea to validate the dim_order entries with a loop like https://github.com/pytorch/executorch/blob/1bf20e3289da509c4344dd65d685f139fc7a172b/runtime/core/exec_aten/util/dim_order_util.h#L24-L30

And also ensure than sizes.size() == dim_order.size()

runtime/core/test/tensor_layout_test.cpp
```diff
@@ -23,9 +25,12 @@ TEST(TestTensorLayout, Ctor) {
   Span<const int32_t> sizes_span = {sizes, sizes + 2};
   Span<const uint8_t> dim_order_span = {dim_order, dim_order + 2};
 
-  TensorLayout layout =
-      TensorLayout(ScalarType::Float, sizes_span, dim_order_span);
+  Result<TensorLayout> layout_res =
```
dbort, line 28: Now that this can fail cleanly, please add tests for the various failure cases.

dbort (approved):
Looking good! Just some final requests for new tests, but the public API looks good to me.

PR #7881, opened 2025-01-23
Commands needs to be all strings

per:
Mixing Path with strings makes the RuntimeError string creation fail on line 318 below in case of a failed run.


zingo (approved):

zingo:
Failing test seems unrelated to PR (and fails other PR's also)

PR #7888, opened 2025-01-23
Support Half/BFloat16 in gelu

swolchok:
Partial fix for #7748.

### DIFF
--- a/kernels/portable/cpu/op_gelu.cpp
+++ b/kernels/portable/cpu/op_gelu.cpp
@@ -37,7 +37,7 @@ Tensor& gelu_out(
   ET_KERNEL_CHECK(
       ctx, tensors_have_same_dim_order(in, out), InvalidArgument, out);
 
-  ET_SWITCH_FLOAT_TYPES(in.scalar_type(), ctx, "gelu.out", CTYPE, [&]() {
+  ET_SWITCH_FLOATHBF16_TYPES(in.scalar_type(), ctx, "gelu.out", CTYPE, [&]() {
     if (approximate == "tanh") {
       apply_unary_map_fn(
           [](const CTYPE x) {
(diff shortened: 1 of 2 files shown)


manuelcandales (approved):

swolchok:
manually ran Mac unittests locally , should be good to land.

zingo:
Hi @swolchok there might be a problem in this change or the Arm tools used.

From this patch (on main) we see this fails on the Arm Backend testing. It don't seem to happen on other targets, I wonder if it is a compiler version kind of thing, but would expect it to show up on other backends compiling for Arm in that case, but maybe out embeded Arm gcc version differ. 
```
/pytorch/executorch/kernels/portable/cpu/op_gelu.cpp: In lambda function:
/pytorch/executorch/kernels/portable/cpu/op_gelu.cpp:54:44: error: ambiguous overload for 'operator+' (operand types are 'int' and 'executorch::runtime::etensor::Half')
   54 |             const CTYPE ret = 0.5 * x * (1 + std::tanh(inner));
      |                                          ~ ^ ~~~~~~~~~~~~~~~~
      |                                          |            |
      |                                          int          executorch::runtime::etensor::Half
```
I thing we are using gcc 13.3 when building the Arm code e.g. 
arm-gnu-toolchain-13.3.rel1-aarch64-arm-none-eabi

Add tried to add the label trunk in this patch to verify if it start causing the problem but it seem I can't after it is merged.

zingo:
Using 1.0 in 
            const CTYPE ret = 0.5 * x * (1.0 + std::tanh(inner));
 fixed it 
I do not know if  static_cast<CTYPE>(1.0) would be preferred but as other constants in the code used float notation directly I tried to go for more easy to read :)

zingo:
Possible fix here https://github.com/pytorch/executorch/pull/7937

swolchok:
> tried to add the label trunk

you can check hud.pytorch.org instead. This did break the ARM builds: https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=test-arm&mergeLF=true

PR #7891, opened 2025-01-23
Add more context for buck users on the coreml pte2 quantization errors

derekxu:
Summary:
Add more context for buck users on the coreml pte2 quantization errors
- See D68569845 for details

Differential Revision: D68569845


cccclai (approved):
Thanks - are you linking the wrong diff in the summary?

PR #7900, opened 2025-01-23
[executorch][flat_tensor] DataMap implementation

lucylq:
* __->__ #7900
* #8201

DataMap implementation that
* Loads a flat_tensor file
* Populates a map with {fqn: tensor} and {fqn: TensorLayout}.
* Makes tensor information available via the named_data_map.h interface.

For now, DataMap doesn't store the DataLoader.
- If/when tensors are in their own segments, DataMap should also store a DataLoader.

Differential Revision: [D67064580](https://our.internmc.facebook.com/intern/diff/D67064580/)


extension/flat_tensor/named_data_map/data_map.h
```diff
@@ -0,0 +1,84 @@
...
+  DataMap(const DataMap& rhs) = delete;
+  DataMap& operator=(DataMap&& rhs) noexcept = delete;
+  DataMap& operator=(const DataMap& rhs) = delete;
+
+  // FlatTensor flatbuffer data. Contains the data backing up
+  // TensorLayout information in the _name_to_tensor map; must outlive it.
+  executorch::runtime::FreeableBuffer _flat_tensor_data;
+
+  // Map of name to {segment index, offset, TensorLayout}.
+  std::unordered_map<
```
JacobSzwejbka, line 74: Im not super opinionated on this, but one thing to consider by having this state as part of the class is you force it to keep dangling around after all the method inits are done which is wasteful. I'm not really super sold on this being a huge latency win either rather then just constructing these things as they are needed.

extension/flat_tensor/named_data_map/data_map.cpp
```diff
@@ -0,0 +1,215 @@
...
+  // Only support one segment for now.
+  assert(s_data_segment->size() == 1);
+  // First segment offset should be 0.
+  int segment_offset = s_data_segment->Get(0)->offset();
+  assert(segment_offset == 0);
+  // First segment size should be <= the total segment data size.
+  int segment_size = s_data_segment->Get(0)->size();
+  assert(segment_size <= segment_data_size);
+
+  Result<FreeableBuffer> _data_ro = loader->load(
```
JacobSzwejbka, line 198: Ok I was trying to figure out how I wanted to handle this for mutable state. Here is what Ive come up with.

Lazily load segments, but internally keep a list of all the loaded segments. If the first time you need to load a segment is through "get_data" then put it in the preserve lis. If the first time you need to load a segment is through load_into then instead you just call data_loader load into which doesnt keep it hanging around.

extension/flat_tensor/named_data_map/data_map.h
lucylq: Ok, this sounds good. Esp. for training case where there aren't many tensors and a linear search (or logn if tensors are sorted) isn't time consuming. If there are latency issues later down the track we can create an impl with a map.

dbort (approved):

PR #7936, opened 2025-01-24
modified all the operators to support generic implementation in specific cases

ckmadhira:
### Summary
modified all the operators to support generic implementation when input data types are different or when input data type is different from output datatype


backends/cadence/fusion_g3/operators/op_cat.cpp
```diff
@@ -106,7 +101,16 @@ Tensor& cat_out(
     out_shapes[i] = out_size[i];
   }
 
-  if ((out.scalar_type() == ScalarType::Int) ||
+  bool optimized = 1;
```
hsharma35, line 104: Please use true/false instead of implicit conversion from int (1) to bool true.

hsharma35:
LGTM, will wait for @zonglinpeng to test internally before we merge.

zonglinpeng (approved):
all tests passed internally :) Layer norm might caused by something else yesterday will keep an eye on it.

hsharma35 (approved):

PR #7937, opened 2025-01-24
Fix for ambiguous overload in kernels/portable/cpu/op_gelu.cpp

zingo:
This fix a error you get when using newer (more strict) compiler version


swolchok (approved):
looks good

zingo:
Sorry for removing the label, I think it's a WebGUI problem in github if labels are created in parallel while the "add label window" is open.
I have noticed this will CLA before, if you are fast and open the Label "cogwheel" before other labels are added by other people/bots they are not selected in your popup and get removed when you close it (e.g. when you adding yours).

zingo:
@digantdesai  OK to merge? (Not in Arm folders so Ill asp for permission here :) )

PR #7940, opened 2025-01-24
Make sure size_test uses -Wuninitialized flags

mergennachin:
Regression test for internal failure (https://www.internalfb.com/diff/D68509826)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -21,6 +21,8 @@ project(size_test)
 
 set(CMAKE_CXX_STANDARD 17)
 
+add_compile_options("-Wuninitialized")
+
 set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/..)
 
 include(${EXECUTORCH_ROOT}/build/Utils.cmake)


swolchok (changes requested):
-Werror is probably a bad idea; we're then accountable for every error that gets introduced in all compiler versions people might try to build with. (this also won't work on Windows or other non-GCC-lineage compilers)

swolchok:
This will still break on Windows. I believe the recommended option is to use compiler-specific presets to configure non-essential flags like warnings; looking into it now.

mergennachin:
> This will still break on Windows. I believe the recommended option is to use compiler-specific presets to configure non-essential flags like warnings; looking into it now.

Btw, I am not changing that this is the default compilation option for all of ExecuTorch. I'm only changing the compile option for size_test and size_test_all_ops tests.

We don't run size_test for Windows

swolchok:
FWIW, -Werror is controlled by `CMAKE_COMPILE_WARNING_AS_ERROR`; it's supposed to be set either on the command line or in a preset, not in CMakeLists.txt

PR #7942, opened 2025-01-24
Restore std::array constructors for ArrayRef

swolchok:
Checked with @suo, shouldn't be any reason to keep these gone.


swolchok:
* (to be filled)

kimishpatel:
is this to match array ref functionality in pt core?

swolchok:
> is this to match array ref functionality in pt core?

yes, it is a manual partial revert of D35259739

kimishpatel (approved):

Issue #7902, opened 2025-01-23
We need a good way to log Tensor shapes in portable ops' error messages


swolchok:
### 🚀 The feature, motivation and pitch

Currently, when we complain about tensors' shapes, we don't include them in the log message because it's not obvious or convenient how to do this. I think we can write a shape_to_string function that uses an extremely finite amount of stack space (`kTensorDimensionLimit` * 4 digits per dim + commas, spaces, and parentheses) and use that to do this logging. (@dbort , how big an array can I put on the stack on a prototypical embedded system these days, and/or do you have any other ideas for how to do this?)

### Additional context

Motivating example: https://github.com/pytorch/executorch/issues/7748#issuecomment-2610994326

swolchok:
I just found sizes_to_string, used in one place: https://github.com/search?q=repo%3Apytorch%2Fexecutorch%20sizes_to_string&type=code
Should replace this with tensor_shape_to_string from #7943.

swolchok:
Closing because we now have a good way. Systematically fixing error messages seems covered by #8273

PR #7943, opened 2025-01-24
Add function to stringify tensor shapes

swolchok:
First step toward fixing #7902


swolchok:
* (to be filled)

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -1120,6 +1120,29 @@ bool extract_scalar_tensor(executorch::aten::Tensor tensor, BOOL_T* out_val) {
   return true;
 }
 
+/**
+ * Maximum size of a string returned by tensor_shape_to_c_string, for
+ * stack allocation.
+ */
+constexpr size_t kTensorShapeStringSizeLimit = 1 + /* opening parenthesis */
+    10 * kTensorDimensionLimit + /* maximum digits we will print; update
```
swolchok, line 1128: @dbort I am happy to lower digits per item if necessary. I don't know what amount of stack space I can reasonably expect to be available.

dbort: What does this value work out to? If it's a couple hundred bytes it's probably ok

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <executorch/runtime/platform/assert.h>
+
+namespace executorch::runtime {
+/**
+ * Shared implementation for tensor_util.h, may only contain code that
+ * works whether or not ATen mode is active.
+ */
+void tensor_shape_to_c_string(
+    char out[kTensorShapeStringSizeLimit],
```
dbort, line 21: I'm worried about the safety of this, since a caller could pass in any size despite the sized parameter:
```
% cat main.cpp
void test(char c[15]) {}

int main() {
  char b[7];
  test(b);
  return 0;
}

% clang++ --std=c++17 -Wall main.cpp
(no output; success)
```

A `char*` and `size_t` size or a `Span<char>` would make it more explicit.

Or if you wanted to stick with a fixed size array you could use `std::array<char, kTensorShapeStringSizeLimit>` which is enforced:
```
% cat main.cpp
#include <array>

void test(std::array<int, 15>c) {}

int main() {
  std::array<int, 7> b;
  test(b);
  return 0;
}


% clang++ --std=c++17 -Wall main.cpp
main.cpp:8:3: error: no matching function for call to 'test'
    8 |   test(b);
      |   ^~~~
main.cpp:3:6: note: candidate function not viable: no known conversion from 'array<[...], 7>' to 'array<[...], 15>' for 1st argument
    3 | void test(std::array<int, 15>c) {
      |      ^    ~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

dbort (changes requested):

runtime/core/exec_aten/util/tensor_util.h
swolchok: kTensorDimensionLimit is 16, so it's 1 + 160 + 32 + 1 = 194 bytes

dbort: That seems ok, especially if callers guard the definition of those stack buffers with an ifdef that will compile them out when logging is disabled

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <executorch/runtime/platform/assert.h>
+
+namespace executorch::runtime {
+/**
+ * Shared implementation for tensor_util.h, may only contain code that
+ * works whether or not ATen mode is active.
+ */
+std::array<char, kTensorShapeStringSizeLimit> tensor_shape_to_c_string(
+    executorch::aten::ArrayRef<executorch::aten::SizesType> shape) {
```
dbort, line 21: We're trying to get away from ArrayRef for non-aten-compatible APIs; could you use executorch::runtime::Span for this?

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -1120,6 +1120,28 @@ bool extract_scalar_tensor(executorch::aten::Tensor tensor, BOOL_T* out_val) {
...
+                                 * NUL terminator for last element */
+    1; /* padding for temporary NUL terminator for simplicity of implementation
+        */
+
+constexpr size_t kMaximumPrintableTensorShapeElement =
+    std::is_same_v<executorch::aten::SizesType, int32_t>
+    ? std::numeric_limits<int32_t>::max()
+    : std::numeric_limits<uint32_t>::max();
+
+std::array<char, kTensorShapeStringSizeLimit> tensor_shape_to_c_string(
```
dbort, line 1142: Please add a doxygen comment for this since it's a public API in a header.

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -1120,6 +1120,28 @@ bool extract_scalar_tensor(executorch::aten::Tensor tensor, BOOL_T* out_val) {
...
+    10 * kTensorDimensionLimit + /* maximum digits we will print; update
+                                  * kMaximumPrintableTensorShapeElement
+                                  * if changing */
+    2 * kTensorDimensionLimit + /* comma and space after each item,
+                                 * overwritten with closing paren and
+                                 * NUL terminator for last element */
+    1; /* padding for temporary NUL terminator for simplicity of implementation
+        */
+
+constexpr size_t kMaximumPrintableTensorShapeElement =
```
dbort, line 1137: Does this need to be public? Could it live in `internal::` or in the .cpp file?

runtime/core/exec_aten/util/test/tensor_util_test.cpp
```diff
@@ -605,3 +610,36 @@ TEST_F(TensorUtilTest, SameShapesDifferentDimOrder) {
...
+  std::ostringstream expected;
+  expected << '(' << kMaximumPrintableTensorShapeElement;
+  for (int ii = 0; ii < kTensorDimensionLimit - 1; ++ii) {
+    expected << ", " << kMaximumPrintableTensorShapeElement;
+  }
+  expected << ')';
+  auto expected_str = expected.str();
+
+  EXPECT_EQ(expected_str, str.data());
+}
```
dbort, line 645: Please also add tests for:
- The "ERR" cases (negative and too large)
- A very-high-dimensional tensor that causes the string to be truncated

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,44 @@
...
+  std::array<char, kTensorShapeStringSizeLimit> out;
+  char* p = out.data();
+  *p++ = '(';
+  for (const auto elem : shape) {
+    if (elem < 0 || elem > kMaximumPrintableTensorShapeElement) {
+      strcpy(p, "ERR, ");
+      p += strlen("ERR, ");
```
dbort, line 28: Is there a guarantee that p has at least 6 bytes available? `strncpy` would be safer

dbort (changes requested):
The new function signature looks great

runtime/core/exec_aten/util/tensor_util.cpp
swolchok: yes, there is a guarantee. we have enough space for a number, so we have enough space for 6 bytes. can add a static_assert about the number of digits in kMaximumPrintableTensorShapeElement

dbort: Nothing in this function seems to check that shape.size() < kTensorDimensionLimit.

...though looking around the codebase I see that there's a ton of code that assumes that limit without checking, which is pretty terrifying. That constant was originally a little helper value for logging, and afaik the original users of it checked that the dim was smaller before using it. Looks like the kernels have inherited an assumption that the constant is some kind of guarantee, which it is not: the runtime can legally deserialize tensors with a larger number of dims than that value.

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,54 @@
...
+ */
+std::array<char, kTensorShapeStringSizeLimit> tensor_shape_to_c_string(
+    executorch::runtime::Span<const executorch::aten::SizesType> shape) {
+  std::array<char, kTensorShapeStringSizeLimit> out;
+  char* p = out.data();
+  if ET_UNLIKELY (shape.size() > kTensorDimensionLimit) {
+    static constexpr char kLimitExceededError[] =
+        "(ERR: tensor ndim exceeds limit, can't happen)";
+    static_assert(sizeof(kLimitExceededError) <= kTensorShapeStringSizeLimit);
+    std::memcpy(p, kLimitExceededError, sizeof(kLimitExceededError));
```
dbort, line 28: Could simplify things with snprintf()

And "can't happen" isn't true: I could create a PTE file where this does happen. Right now there's a dangerous assumption across the kernel libraries that dim is always <= kTensorDimensionLimit.

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return out;
+  }
+  *p++ = '(';
+  for (const auto elem : shape) {
+    if (elem < 0 || elem > internal::kMaximumPrintableTensorShapeElement) {
+      static_assert(
+          internal::kMaximumPrintableTensorShapeElement > 99999,
```
dbort, line 35: Took me a minute to understand where `>99999` came from: please add a comment pointing out that it's the same number of digits as "ERR, " with the NUL

runtime/core/exec_aten/util/tensor_util.cpp
```diff
@@ -0,0 +1,54 @@
...
+          "must have room for error string!");
+      strcpy(p, "ERR, ");
+      p += strlen("ERR, ");
+    } else {
+      // snprintf returns characters *except* the NUL terminator, which is what
+      // we want.
+      p += snprintf(
+          p,
+          kTensorShapeStringSizeLimit - (p - out.data()),
```
dbort, line 44: Should there be a -1 in here to account for the NUL byte? Not sure

dbort (approved):
Thanks for all of the changes here. I know I'm being paranoid but we really need to be extra paranoid in the core runtime.

A few final requests but looks good.

swolchok:
had to move the function into kernels/portable/cpu/util to fix an apparent layering violation -- portable_kernels failed to link because it needed to depend on executorch_core for this new function, but executorch_core already depends on portable_kernels

swolchok:
> apparent layering violation -- portable_kernels failed to link because it needed to depend on executorch_core for this new function, but executorch_core already depends on portable_kernels

I'll send a diff to move this back to runtime/core. portable_kernels really should be able to use code from runtime/core and either I was wrong or we should dig into why not.

PR #7944, opened 2025-01-24
Include tensor shapes in get_broadcast_target_size error message

swolchok:
This is the motivating example for #7902.

Test Plan: Injected failure to new broadcast_test and saw shapes in error message.


swolchok:
* (to be filled)

swolchok:
would like feedback on this one before sending more PRs to log tensor shapes broadly

kernels/portable/cpu/util/broadcast_util.cpp
```diff
@@ -213,10 +213,20 @@ ET_NODISCARD Error get_broadcast_target_size(
     Tensor::SizesType* out_sizes,
     const size_t out_sizes_len,
     size_t* out_dim) {
-  ET_CHECK_OR_RETURN_ERROR(
-      tensors_are_broadcastable_between(a_size, b_size),
-      InvalidArgument,
-      "Two input tensors should be broadcastable.\n");
+  if (!tensors_are_broadcastable_between(a_size, b_size)) {
+    const auto a_shape_str = tensor_shape_to_c_string(
```
dbort, line 217: This new API is great, keeping the caller totally ignorant of how big a buffer to allocate.

kernels/portable/cpu/util/broadcast_util.cpp
```diff
@@ -213,10 +213,20 @@ ET_NODISCARD Error get_broadcast_target_size(
     Tensor::SizesType* out_sizes,
     const size_t out_sizes_len,
     size_t* out_dim) {
-  ET_CHECK_OR_RETURN_ERROR(
-      tensors_are_broadcastable_between(a_size, b_size),
-      InvalidArgument,
-      "Two input tensors should be broadcastable.\n");
+  if (!tensors_are_broadcastable_between(a_size, b_size)) {
```
dbort, line 216: Maybe thrown an UNLIKELY on this, but compared to the work done during broadcasting this is probably inconsequential

```suggestion
  if (ET_UNLIKELY(!tensors_are_broadcastable_between(a_size, b_size))) {
```

dbort (approved):

kernels/portable/cpu/util/broadcast_util.cpp
```diff
@@ -213,10 +213,22 @@ ET_NODISCARD Error get_broadcast_target_size(
...
+            a_size.data(), a_size.size()));
+    const auto b_shape_str = tensor_shape_to_c_string(
+        executorch::runtime::Span<const Tensor::SizesType>(
+            b_size.data(), b_size.size()));
+#endif
+    ET_LOG(
+        Error,
+        "Two input tensors should be broadcastable but got shapes %s and %s.",
+        a_shape_str.data(),
+        b_shape_str.data());
```
dbort, line 229: I think it'd be less surprising for readers if the if/endif completely covered the scopes of the locals that are defined in it, so they don't need to do the extra mental jump of "I guess ET_LOG evaluates to nothing when ET_LOG_ENABLED is false"

```suggestion
    ET_LOG(
        Error,
        "Two input tensors should be broadcastable but got shapes %s and %s.",
        a_shape_str.data(),
        b_shape_str.data());
#endif
```

PR #7948, opened 2025-01-24
Enable constant methods in cadence export.

hsharma35:
Summary: Add option to specify constant methods in export.

Differential Revision: D68637824


zonglinpeng (approved):

backends/cadence/aot/compiler.py
```diff
@@ -8,7 +8,7 @@
 
 import logging
 from pathlib import Path
-from typing import Callable, cast, Optional
+from typing import Any, Callable, cast, Optional
```
zonglinpeng, line 11: Use object than Any

PR #7950, opened 2025-01-24
Execute on exec_aten namespace deprecation from #5296

swolchok:
[fastmod](https://github.com/facebookincubator/fastmod)` 'exec_aten::' 'executorch::aten::'`
Manually fix op_sdpa.cpp to use `::executorch::aten`, run lintrunner.

(FWIW, I disagree with our use of nested namespaces for the reasons outlined in https://abseil.io/tips/130, but we can always find/replace back and leaving the migration unfinished seems worse.)


swolchok:
* __->__ #7950

swolchok:
unittest and lintrunner failures appear to be unrelated

dbort (approved):
Thanks for migrating all of these!

PR #7962, opened 2025-01-27
add size and rank of nightly facto

zonglinpeng:
Summary: x2 scale number of testcases

Differential Revision: D68690963

### DIFF
--- a/examples/cadence/operators/facto_util.py
+++ b/examples/cadence/operators/facto_util.py
@@ -18,58 +18,36 @@ random_manager.seed(1729)
 
 def apply_tensor_contraints(op_name: str, tensor_constraints: list[object]) -> None:
     match op_name:
-        case (
-            "sigmoid.default"
-            | "_softmax.default"
-            | "rsqrt.default"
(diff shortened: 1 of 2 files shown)


examples/cadence/operators/facto_util.py
```diff
@@ -18,58 +18,36 @@
 
 def apply_tensor_contraints(op_name: str, tensor_constraints: list[object]) -> None:
     match op_name:
-        case (
-            "sigmoid.default"
-            | "_softmax.default"
-            | "rsqrt.default"
-            | "exp.default"
-            | "mul.Tensor"
-            | "div.Tensor"
```
mcremon-meta, line 27: does that mean the removed ones use default params now?

zonglinpeng: correct. they used to require stricter contraints because of kernel impl. now they should be all fixed @mcremon-meta

hsharma35 (approved):

mcremon-meta (approved):

PR #7964, opened 2025-01-27
Add missing dtypes to gen_selected_op_variants.py

swolchok:
Summary: These were in scalar_type.h but not here. We should be able to understand all dtypes in this script even if they're not all implemented, because a custom op might support them.

Differential Revision: D68714101


lucylq (approved):
Thanks for adding these!

swolchok:
unittest / linux failure looks unrelated, will bypass

PR #7965, opened 2025-01-27
embed extended header inside .ptd flatbuffer section

JacobSzwejbka:
Summary: Embed the header inside the flatbuffer. We do this for .pte and it lets us reuse a lot of flatbuffer tools natively.

Differential Revision: D68578075

### DIFF
--- a/extension/flat_tensor/serialize/serialize.cpp
+++ b/extension/flat_tensor/serialize/serialize.cpp
@@ -109,13 +109,38 @@ runtime::Error save_ptd(
       tensor_alignment,
       builder.CreateVector(tensors),
       builder.CreateVector(buffers));
-  builder.Finish(flat_tensor); // Our flatbuffer is created now.
+  builder.Finish(flat_tensor, ::flat_tensor_flatbuffer::FlatTensorIdentifier());
+  // Our flatbuffer is created now.
 
   // Calculate flatbuffer padding.
   auto padded_flatbufer_size =
       aligned_size(builder.GetSize(), tensor_alignment);
   auto padded_header_size =
       aligned_size(FlatTensorHeader::kHeaderExpectedLength, tensor_alignment);
+
+  // The general structure of the file is:
+  // [flatbuffer offset to root table][flatbuffer file indentifier]
+  //   [FlatTensorHeader][padding][flatbuffer contents][padding]
+  //   [segment data].
+  // This means we first serialize the first 8 bytes of the flatbuffer,
+  // updating the offset to the root table, then the header, then the
+  // flatbuffer. We are embedding the header inside the flatbuffer doing
+  // this which allows us to continue using flatbuffer tools directly on the
+  // .ptd file.
+
+  // Calculate new offset to root table.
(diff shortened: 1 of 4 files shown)


lucylq (approved):
Lgtm, 

nit to change `extended_header` to `flat_tensor_header` for flat_tensor?

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -109,14 +109,38 @@ runtime::Error save_ptd(
...
+  uint32_t new_offset = current_offset + padded_header_size;
+
+  // Write flatbuffer offset to root table
+  out.write(reinterpret_cast<const char*>(&new_offset), sizeof(new_offset));
+
+  // Write flatbuffer magic bytes
+  out.write(
+      reinterpret_cast<const char*>(builder.GetBufferPointer()) +
+          sizeof(new_offset),
+      4); // This is the 'FT01' magic bytes from flat_tensor.fbs.
```
dbort, line 142: So this doesn't become stale if "FT01" becomes "FT02"
```suggestion
      4); // This is the file identifier from flat_tensor.fbs.
```

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -109,14 +109,38 @@ runtime::Error save_ptd(
...
+  // Our flatbuffer is created now.
 
   // Calculate flatbuffer padding.
   auto padded_flatbufer_size =
       aligned_size(builder.GetSize(), tensor_alignment);
   auto padded_header_size =
       aligned_size(FlatTensorHeader::kHeaderExpectedLength, tensor_alignment);
 
+  // The general structure of the file is:
+  // [flatbuffer offset to root table][flat tensor magic bytes][header]
```
dbort, line 122: "flat tensor magic bytes" is ambiguous since it could also refer to the magic in FlatTensorHeader.
"flatbuffer file identifier" would be more specific.

Similar with "header"; could say "FlatTensorHeader". And there may be some padding after the the header before the flatbuffer contents.

extension/flat_tensor/serialize/serialize.cpp
```diff
@@ -149,10 +173,10 @@ runtime::Error save_ptd(
       padding_required(
           FlatTensorHeader::kHeaderExpectedLength, tensor_alignment));
 
-  // Write flatbuffer
+  // Write flatbuffer, offset by 8 bytes since we wrote those before the header.
```
dbort, line 176: Some specifics to make it more clear
```suggestion
  // Write flatbuffer, offset by 8 bytes (4-byte root table offset + 4-byte file
  // identifier) since we wrote those before the FlatTensorHeader.
```

dbort (approved):
Looks good! Thanks for making this change.

A good double-check would be to try parsing a .ptd file using the `flatc` commandline tool, like
```
flatc --json --defaults-json --strict-json -o /tmp extension/flat_tensor/serialize/flat_tensor.fbs -- input.ptd
```
and see if it was able to create /tmp/input.json

PR #7975, opened 2025-01-27
aten.leakyrelu.default in unary_ops

hossein1387:
Differential Revision: D68688186

### DIFF
--- a/backends/vulkan/runtime/graph/ops/impl/UnaryOp.cpp
+++ b/backends/vulkan/runtime/graph/ops/impl/UnaryOp.cpp
@@ -114,6 +114,17 @@ float get_val_or_inf(ComputeGraph& graph, const ValueRef& val, bool max) {
         "hardshrink");                                                   \
   }
 
+#define DEFINE_LEAKY_RELU_FN(op_name)                                    \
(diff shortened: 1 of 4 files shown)


backends/vulkan/runtime/graph/ops/impl/UnaryOp.cpp
```diff
@@ -114,6 +114,17 @@ float get_val_or_inf(ComputeGraph& graph, const ValueRef& val, bool max) {
         "hardshrink");                                                   \
   }
 
+#define DEFINE_LEAKY_RELU_FN(op_name)                                    \
```
swolchok, line 117: why do we need a macro that is only used once?

swolchok (approved):

backends/vulkan/runtime/graph/ops/impl/UnaryOp.cpp
hossein1387: I couldn't use other macros (DEFINE_ACTIVATION_FN for instance) since the func signature is different. Also, I thought maybe other activations (like PReLU) could use the same macro in the future.

PR #7990, opened 2025-01-28
Increase calibration samples and tolerance for flaky quantized op tests

GregoryComer:
### Summary
Several tests, including test_add_qs8 and test_qs8_conv1d, are failing intermittently. This PR adds additional calibration samples to reduce dependence on the single input sample and adjusts test tolerance accordingly.

Note: We likely want to expose calibration_samples as a numeric option in the tester, in addition to the logic I added to pass in samples. If we can re-use generate_random_inputs, that would be ideal, though it requires a little refactoring to make it available from inside of a stage. We can take that as a follow-up.

### Test plan
Re-ran the above failing tests 10k times (100 full runs with 100 iterations of run_method_and_compare_outputs).

### DIFF
--- a/backends/xnnpack/test/ops/test_add.py
+++ b/backends/xnnpack/test/ops/test_add.py
@@ -7,7 +7,7 @@
 import unittest
 
 import torch
-from executorch.backends.xnnpack.test.tester import Tester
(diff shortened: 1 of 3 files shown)


mcr229 (approved):
thanks for this change!

PR #7992, opened 2025-01-28
Support rsqrt in XNNPACK backend

swolchok:
I think I updated everywhere that needs updating?

Test Plan: python -m unittest backends/xnnpack/test/ops/test_rsqrt.py

### DIFF
--- a/backends/xnnpack/test/ops/test_rsqrt.py
+++ b/backends/xnnpack/test/ops/test_rsqrt.py
@@ -1,0 +1,42 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+
+import torch
+from executorch.backends.xnnpack.test.tester import Tester
+
+
+class TestRsqrt(unittest.TestCase):
+    class Rsqrt(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            x = torch.abs(x)
+            z = torch.rsqrt(x)
+            return z
+
+    def _test_rsqrt(self, inputs):
+        (
+            Tester(self.Rsqrt(), inputs)
+            .export()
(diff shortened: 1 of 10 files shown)


swolchok:
* (to be filled)

iseeyuan (approved):
LGTM. Thanks @swolchok for this quick fix! @mcr229 and @digantdesai could you take a look?

iseeyuan:
@swolchok you mentioned rsqrt was used since Llama2. I'm wondering how it would help the overall performance. Probably it's fine because it's not a performance bottleneck there. It's used in RMSNorm, where the sqrt is applied to a scalar of RMS(x), not on each element of x. It may be an issue for element-wise sqrt for some operations we discussed offline.

backends/xnnpack/test/ops/test_rsqrt.py
```diff
@@ -0,0 +1,42 @@
...
+            super().__init__()
+
+        def forward(self, x):
+            x = torch.abs(x)
+            z = torch.rsqrt(x)
+            return z
+
+    def _test_rsqrt(self, inputs):
+        (
+            Tester(self.Rsqrt(), inputs)
```
digantdesai, line 25: Nit: Do you want to improve this test by adding dynamic_shape support like [this](https://github.com/pytorch/executorch/blob/019df698c4840adaef80cf39b7dd7ae42c9be6e5/backends/xnnpack/test/ops/test_softmax.py#L34).

digantdesai (approved):

backends/xnnpack/test/ops/test_rsqrt.py
swolchok: > improve this test by adding dynamic_shape support

it's a pointwise operator, why does it care about the shape? Everything is the same as sqrt including the test

digantdesai: Yeah, just checking boxes, i.e. adding dynamic shape test for every operator.

swolchok: out of scope for this change! :)

PR #7996, opened 2025-01-28
Fix helper method for getting input names

Erik-Lundell:
The previous version missed ConstantInputs which
caused problems in the runtime when the provided
inputs did not match the GraphSignature.

You could argue that having to provide constant inputs that are already encoded in the graph as arguments is unnecessary. However, this solution is more general and does not stop us from adding a pass that prunes unwanted inputs from the graph signature in the future. (if that is possible)

Encountered an issue with too long input names,
had to rename some parameters in testcases to
shorter names.


digantdesai (approved):

zingo (approved):

zingo:
macos job seems unrelated and fail on other PR and on "main"

PR #7997, opened 2025-01-28
Arm backend: manually annotate metadata of inserted transposes

oscarandersson8218:
### Summary
Manually annotate node.meta["val"] of inserted transposes.

### DIFF
--- a/backends/arm/_passes/annotate_channels_last_dim_order_pass.py
+++ b/backends/arm/_passes/annotate_channels_last_dim_order_pass.py
@@ -129,6 +129,7 @@ class AnnotateChannelsLastDimOrder(ExportPass):
             permute_node.meta["tosa_dim_order"] = tuple(
                 range(len(input_node.meta["val"].size()))
             )
+            permute_node.meta["val"] = input_node.meta["val"]
 
     @staticmethod
     def insert_output_transpose(node, graph_module):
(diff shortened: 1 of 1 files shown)


backends/arm/_passes/annotate_channels_last_dim_order_pass.py
```diff
@@ -129,6 +129,7 @@ def insert_input_transpose(node, input_node, graph_module):
             permute_node.meta["tosa_dim_order"] = tuple(
                 range(len(input_node.meta["val"].size()))
             )
+            permute_node.meta["val"] = input_node.meta["val"]
```
digantdesai, line 132: Curious why is this needed, `ExportPass` should trace and update the graph once done running. Is it because you might run both input and output transpose inserts before the we get to tracing? If yes, would it be easier to split this into two passes?

digantdesai (approved):

backends/arm/_passes/annotate_channels_last_dim_order_pass.py
oscarandersson8218: > Is it because you might run both input and output transpose inserts before the we get to tracing?

Yes, we end up with two consecutive transposes, and I agree that this pass should probably be split into two passes. I think we'll update this pass in the near future.

zingo (approved):

zingo:
macos fails unrealted

PR #8003, opened 2025-01-28
Add option to enforce alignment constraint when planning memory

skrtskrtfb:
Summary: Add ability to enforce start alignment in Cadence Memory Planner

Differential Revision: D68762973


backends/cadence/aot/compiler.py
```diff
@@ -257,6 +257,7 @@ def export_to_executorch_gen_etrecord(
     alloc_graph_output: bool = True,
     memory_config: Optional[MemoryConfig] = None,
     dump_graphs: bool = False,
+    mem_alignment: int = 0,
```
zonglinpeng, line 260: is this the default memory block size for alignment?

zonglinpeng (approved):

PR #8011, opened 2025-01-28
fix missing `if constexpr` in optimized gelu (unbreak sync of D68745515/#7987)

swolchok:
The code gated under this `if` does disallowed static_casts if the `if` condition doesn't pass; this is what `if constexpr` is for.


swolchok:
* (to be filled)

GregoryComer (approved):

PR #8016, opened 2025-01-28
Make sure size_test CI uses -Wall -Werror

mergennachin:
I am not changing the default compilation option for all of ExecuTorch. I'm only changing the compile option for size_test and size_test_all_ops tests.

We don't run size_test for Windows, at least for today

Test Plan: 

Run `sh test/build_size_test.sh`
Add op_fill.cpp with this patch at the end of the file.
```
int x;
printf("%d", x); // x is uninitialized
```

Make sure the compilation fails, https://gist.github.com/mergennachin/25b78c63069b9e9b2203e66f94b1a0be

Make sure the compilation succeeds without the "-Wall -Werror" flags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -215,6 +215,14 @@ must work with threading**
 
 ## Testing
 
+### Running Tests Locally
+
+CI is run automatically on all pull requests. However, if you want to run tests locally, here are some example commands (not exhaustive):
+
+- The `sh test/build_size_test.sh` script will compile the C++runtime along with portable kernels.
+- The `test/run_oss_cpp_tests.sh` script will build and run C++ tests locally
+- Running `pytest` from the root directory will run Python tests locally.
+
 ### Writing Tests
 To help keep code quality high, ExecuTorch uses a combination of unit tests and
 end-to-end (e2e) tests. If you add a new feature or fix a bug, please add tests
@@ -228,8 +236,6 @@ that it did actually run.
 If it's not clear how to add a test for your PR, take a look at the blame for
 the code you're modifying and find an author who has more context. Ask them
 for their help in the PR comments.
-
-The `test/run_oss_cpp_tests.sh` script will build and run C++ tests locally.
 
 ### Continuous Integration
 See https://hud.pytorch.org/hud/pytorch/executorch/main for the current state of
--- a/kernels/portable/cpu/op_full.cpp
+++ b/kernels/portable/cpu/op_full.cpp
@@ -38,7 +38,8 @@ Tensor& full_out(
 
   ET_SWITCH_SCALAR_OBJ_TYPES(val_type, ctx, name, CTYPE_VAL, [&] {
     CTYPE_VAL val;
-    utils::extract_scalar(fill_value, &val);
+    ET_KERNEL_CHECK(
+        ctx, utils::extract_scalar(fill_value, &val), InvalidArgument, );
 
     ET_SWITCH_REALHBBF16_TYPES(out_type, ctx, name, CTYPE_OUT, [&] {
       CTYPE_OUT val_casted = static_cast<CTYPE_OUT>(val);
--- a/kernels/portable/cpu/util/broadcast_util.cpp
+++ b/kernels/portable/cpu/util/broadcast_util.cpp
@@ -215,18 +215,16 @@ ET_NODISCARD Error get_broadcast_target_size(
     size_t* out_dim) {
   if ET_UNLIKELY (!tensors_are_broadcastable_between(a_size, b_size)) {
 #ifdef ET_LOG_ENABLED
-    const auto a_shape_str = tensor_shape_to_c_string(
-        executorch::runtime::Span<const Tensor::SizesType>(
-            a_size.data(), a_size.size()));
-    const auto b_shape_str = tensor_shape_to_c_string(
-        executorch::runtime::Span<const Tensor::SizesType>(
-            b_size.data(), b_size.size()));
-#endif
+    executorch::runtime::Span<const Tensor::SizesType> a_size_span(
+        a_size.data(), a_size.size());
+    executorch::runtime::Span<const Tensor::SizesType> b_size_span(
+        b_size.data(), b_size.size());
     ET_LOG(
         Error,
         "Two input tensors should be broadcastable but got shapes %s and %s.",
-        a_shape_str.data(),
-        b_shape_str.data());
+        tensor_shape_to_c_string(a_size_span).data(),
+        tensor_shape_to_c_string(b_size_span).data());
+#endif
     return executorch::runtime::Error::InvalidArgument;
   }
 
--- a/test/build_size_test.sh
+++ b/test/build_size_test.sh
@@ -11,11 +11,15 @@ set -e
 # shellcheck source=/dev/null
 source "$(dirname "${BASH_SOURCE[0]}")/../.ci/scripts/utils.sh"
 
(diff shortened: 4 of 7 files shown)


mergennachin:
Continuing where I left off: https://github.com/pytorch/executorch/pull/7940

Not changing CMakeLists.txt this time.

swolchok (approved):
love it

swolchok:
I don't see size_test in the CI jobs?

mergennachin:
> I don't see size_test in the CI jobs?

Just rebased (which contain CI fixes)

size_test for clang passed, but not for gcc 

https://github.com/pytorch/executorch/actions/runs/13022554836/job/36326025227?pr=8016

runtime/kernel/operator_registry.cpp
```diff
@@ -135,7 +135,7 @@ void make_kernel_key_string(Span<const TensorMeta> key, char* buf) {
     // If no tensor is present in an op, kernel key does not apply
     return;
   }
-  strncpy(buf, "v1/", 3);
+  strcpy(buf, "v1/");
```
mergennachin, line 138:
```
/pytorch/executorch/runtime/kernel/operator_registry.cpp:187:3: note: in expansion of macro 'ET_LOG_TENSOR_META'
  187 |   ET_LOG_TENSOR_META(meta_list);
      |   ^~~~~~~~~~~~~~~~~~
In file included from /usr/include/string.h:535,
                 from /usr/include/c++/9/cstring:42,
                 from /pytorch/executorch/../executorch/runtime/kernel/operator_registry.h:11,
                 from /pytorch/executorch/runtime/kernel/operator_registry.cpp:9:
In function 'char* strncpy(char*, const char*, size_t)',
    inlined from 'void executorch::runtime::internal::make_kernel_key_string(executorch::runtime::Span<const executorch::runtime::TensorMeta>, char*)' at /pytorch/executorch/runtime/kernel/operator_registry.cpp:138:10:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: error: 'char* __builtin_strncpy(char*, const char*, long unsigned int)' output truncated before terminating nul copying 3 bytes from a string of the same length [-Werror=stringop-truncation]
```

mergennachin:
- Ignore 'sign-compare' but created a follow-up issue https://github.com/pytorch/executorch/issues/8149
- Fixed an error that was erroring with -Werror=stringop-truncation
- Modified the CONTRIBUTION.md file to build the size_test information

kernels/portable/cpu/op_full.cpp
```diff
@@ -38,7 +38,8 @@ Tensor& full_out(
 
   ET_SWITCH_SCALAR_OBJ_TYPES(val_type, ctx, name, CTYPE_VAL, [&] {
     CTYPE_VAL val;
-    utils::extract_scalar(fill_value, &val);
+    ET_KERNEL_CHECK(
```
mergennachin, line 41: Had to adjust due to errors like this: 

https://github.com/pytorch/executorch/actions/runs/13123283587/job/36614101657?pr=8016

swolchok: I don't think ET_KERNEL_CHECK will do the right thing in a lambda -- it returns from the lambda, not the operator

runtime/kernel/operator_registry.cpp
swolchok: the warning here was unhelpful IMO

test/build_size_test.sh
```diff
@@ -11,11 +11,14 @@ set -e
 # shellcheck source=/dev/null
 source "$(dirname "${BASH_SOURCE[0]}")/../.ci/scripts/utils.sh"
 
+# Remove -Wno-sign-compare (https://github.com/pytorch/executorch/issues/8149)
```
swolchok, line 14: did you mean to put a TODO here?

runtime/kernel/operator_registry.cpp
dbort: Should this just be `memcpy(buf, "v1/", 3);`? Or are we relying on the NUL byte that strcpy adds?

dbort: I completely rewrote this function in https://github.com/pytorch/executorch/pull/8327, so watch out for merge conflicts (if mine lands first)

kernels/portable/cpu/op_full.cpp
mergennachin: any suggestions on how to fix? 

i think it's fine, since it will return void from the lambda, but sets context to be fail, with error code InvalidArgument.

mergennachin:
> @mergennachin do you need anything else from me on this one?

@swolchok, @dbort 

i don't think we have come to a resolution on errors on boolean errors: 

https://github.com/pytorch/executorch/actions/runs/13123890485/job/36616081625?pr=8016

runtime/kernel/operator_registry.cpp
mergennachin: yeah, thanks for the heads up, will be careful during rebase

swolchok:
&&= does not exist; you'd need &=. IMO such a change would be clear uglification of the code only to suppress a warning

swolchok:
I don't think the code change you are considering would address the issue you just raised in any way.

mergennachin:
Okay, I will suppress that error for now and create a follow-up issue to track it

CONTRIBUTING.md
```diff
@@ -215,6 +215,11 @@ must work with threading**
 
 ## Testing
 
+### Running Tests
```
dbort, line 218:
```suggestion
### Running Tests Locally
```

test/build_size_test.sh
```diff
@@ -11,11 +11,15 @@ set -e
 # shellcheck source=/dev/null
 source "$(dirname "${BASH_SOURCE[0]}")/../.ci/scripts/utils.sh"
 
+# TODO(https://github.com/pytorch/executorch/issues/8149): Remove -Wno-sign-compare
+# TODO(https://github.com/pytorch/executorch/issues/8357): Remove -Wno-int-in-bool-context
```
dbort, line 15: We should have a convention for this. Since GH auto-links string like #8149, I think we could settle on `TODO(#8149): ...` when referring to an issue within the ET repo.

dbort (approved):

kernels/portable/cpu/util/broadcast_util.cpp
```diff
@@ -215,18 +215,16 @@ ET_NODISCARD Error get_broadcast_target_size(
...
+    executorch::runtime::Span<const Tensor::SizesType> b_size_span(
+        b_size.data(), b_size.size());
     ET_LOG(
         Error,
         "Two input tensors should be broadcastable but got shapes %s and %s.",
-        a_shape_str.data(),
-        b_shape_str.data());
+        tensor_shape_to_c_string(a_size_span).data(),
+        tensor_shape_to_c_string(b_size_span).data());
+#endif
```
digantdesai, line 227: Thanks!

swolchok:
@mergennachin it seems suboptimal for us to enforce -Wall -Werror, but only in a CI job that runs only on trunk, as this will result in unnecessarily-delayed detection of broken trunk. could we perhaps have a pull-time CI job that builds with these flags?

swolchok:
e.g. https://github.com/pytorch/executorch/pull/8314 promptly broke this on trunk because of -Werror

mergennachin:
@swolchok test-binary-size is a pull job

perhaps, in this #8314 case your base commit was old and it didn't catch during PR?

PR #8018, opened 2025-01-28
remove the exec_aten namespace

swolchok:
Following up on #7950, it was deprecated and appears to now be unused.


swolchok:
* (to be filled)

dbort:
Totally up for removing this, but there are still a lot of users inside Meta

swolchok:
> Totally up for removing this

wait, aren't we prohibited from removing this by our BC policy?

dbort:
Good question: it was deprecated in 0.4, and we just released 0.5, so it should be safe to remove in 0.6 (i.e., now)

Issue #8030, opened 2025-01-29
Multi Entry points and Shared State implementation Request for comments/feedback.


cptspacemanspiff:
### 🚀 The feature, motivation and pitch

# Overview
Hi, I am not really sure what the process for this is, but I made some changes/a wrapping to executorch to be able to have a shared state between multiple exported methods of a torch.module. 

For now I implemented this in my own repo: [My Repo](https://github.com/[cptspacemanspiff/execu-tools](https://github.com/cptspacemanspiff/execu-tools))

My initial motivation for this is that when exporting Encoder/Decoder models one would really like to use a precalculated cross attention cache, but this is hard to do without conditional branching (via torch.cond)/rewriting or extraneous copies/calculations.

* Aside from this somewhat specific use-case, I think it can be more generally useful, as it allows the export of more of the whole model's generate graph without rewriting as much into C++, which causes more surface area to introduce bugs/etc.
  * for example you can just place the branching in C++, call methods in sequence, and do not have to manually rewrite C++ to pass data between methods, since things can get passed under the hood via shared state.
* Additionally, there may be KV cache management strategies that might benefit from outside access to the cache, from outside of a particular models decode function. 

My high-level thought process is that, for the most part, for a given model class (sequence to sequence, decoder only, etc), the location of forced graph breaks from the torch export process (due to either data-dependent branching or what not) are mostly in the same place. If you export the individual parts in python then glue these together with minimal code on the C++ side, The C++ runtime implementation should be able to generalize to other models of the same class. From what I can tell, this is already the case for decoder only models, but for other architectures that still have state this is less well defined.

Currently, I have 2 examples exported / working:
1. A dummy module that just sets/reads from a cache.
2. A hugging face BART model (OPUS translation),:
   *  It uses Hugging face static cache, logit postprocessing, and stopping conditions from huggingface's generate implementation,
   *  Uses a C++ wrapped version of hugging faces tokenization library, loading the configuration json blob from the model.
   *  However, I am manually using argmax for the `greedy` search, and there were changes to the models attention to support non-fixed batch sizes and encoder sequence lengths that are still not merged into transformers.

## What I am asking for:
I have been working on this on my own, but it seems like it is something that would be more generally useful. I just started really digging into the torch.export process/ executorch in the past month so may have some fundamental misunderstandings.

[...]

PR #8031, opened 2025-01-29
Changes to allow the export of functions with no user input.

cptspacemanspiff:
### Summary
This is a set of changes that I have been using to get executorch working with Multiple Entrypoints with shared state.
This is related to: #8030 

### Issue & Changes
Currently trying to export a function:

```python
def init():
  self.shared_state.fill_(0)
```

Will fail, because in to_edge in exir in the program.py the lifted constants are assumed to be placed before the first user input. This causes weird behavior when there are no user inputs. I changed it to place the lifted constant after the last buffer, if there are no user inputs to place before. However, this relies on my understanding of the implicit layout of the graph inputs.

At the same time later in the after the memory planning phase of to_executorch, it validates that memory planning was correct based on whether the `graph_input_allocated` flag is set, this only applies to user inputs, of which we have none, so it errors out. I added a check to bypass this error if there are no user inputs, but I honestly do not understand enough of the validation check to know if that is appropriate.

### Comments

In the current executorch with no support for shared state, this case does not make sense, but #8030 is my attempt at adding that capability. and having initialization methods that init the buffers from constants is useful, especially since their initail state is undefined.

Currently this is not ready, I have no tests/ and have my random notes in the commit as comments, and other than validating that it worked as I was working on the shared state export have not done anything in depth....

But I kind-of want feedback if my solution seems correct, or if I am missing something. Particularly regarding my understanding of placeholder ordering and signature logic, and whether bypassing the graph_input_allocated validation is appropriate.

Out of all the changes I had to make for shared state, this is the one I am least sure about.

 

cc @JacobSzwejbka @angelayi


JacobSzwejbka:
@ydwu4 @angelayi can you guys take a look at this one

JacobSzwejbka:
Hmm I thought we already supported functions with no inputs, but I guess before we only tested with constant returns. Now theres no user return either. 

cc @tarun292  who I think worked on the no inputs thing before

exir/memory_planning.py
```diff
@@ -236,7 +236,10 @@ def verify_graph_input_output(self) -> None:
                     graph_output_allocated = allocated
                     has_dynamic_unbound_output |= has_dynamic_unbound_tensor
 
-        if "placeholder" in check_list:
+        # only check if inputs are allocated if there are user inputs:
+        user_inputs_exist = len(list(filter(lambda input: input.kind == InputKind.USER_INPUT, self.graph_signature.input_specs))) > 0
+
+        if "placeholder" in check_list and user_inputs_exist:
```
JacobSzwejbka, line 242: this change looks fine. On our side we should probably take a pass over all the memory planning infra. A lot of it was written before export graphs were functionalized so the concept of non user inputs didnt make exist.

exir/program/_program.py
```diff
@@ -306,6 +335,9 @@ def lift_constant_tensor_pass(ep):
             new_input_specs.extend(lifted_constants)
```
angelayi, line 335: And if we just insert all the constants at the beginning, we can just directly prepend to input_specs.

exir/program/_program.py
```diff
@@ -306,6 +335,9 @@ def lift_constant_tensor_pass(ep):
...
             lifted_constants.clear()
         new_input_specs.append(s)
```
angelayi, line 337: Can you add a test case?

cptspacemanspiff: I added a test case, shifted to the simpler implementation of just prepending the signature since the order of the non-user inputs does not matter.

cptspacemanspiff: I have also validated that it works with my export process as well.

exir/program/_program.py
cptspacemanspiff: done, and removed all the logic I had previously, now:

No user inputs -> inserting before `None` -> inserts at the beginning.

exir/tests/test_passes.py
```diff
@@ -1057,6 +1057,35 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            exported_program,
+            compile_config=EdgeCompileConfig(_skip_dim_order=False),
+        )
+        ep = edge.exported_program()
+        # because there is no user input, the lifted constant should be the first input.
+        FileCheck().check("_lifted_tensor_constant1").check(
+            "b_a"  # followed by the buffer input.
+        ).run(ep.graph_module.code)
+        # the graph signature should also be the same:
+        assert ep.graph_signature.input_specs[0].arg.name == "_lifted_tensor_constant1"
```
angelayi, line 1081:
```suggestion
        self.assertEquals(ep.graph_signature.input_specs[0].arg.name, "_lifted_tensor_constant1")
```

exir/tests/test_passes.py
```diff
@@ -1057,6 +1057,35 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        assert ep.graph_signature.input_specs[1].arg.name == "b_a"
+
+        executorch_program = edge.to_executorch()
+        # # the graph signature should also be the same:
+        # executorch_program.graph_signature.input_specs[0].arg.name == "_lifted_tensor_constant1"
+        # executorch_program.graph_signature.input_specs[1].arg.name == "b_a"
```
angelayi, line 1087: why are these commented out?

cptspacemanspiff: I accidentally left those in, when I was initially writing the test/debugging and validating changes I just copy/pasted those lines so I had something to drop a debugger breakpoint on.

I removed those, and left in the to_executorch call, this is there to ensure that the validation steps that were previously failing are correctly bypassed. (Does not throw an exception.)

cptspacemanspiff:
@angelayi

angelayi (approved):
thanks!!

JacobSzwejbka:
Im rerunning CI and then assuming it passes Ill merge it

shoumikhin:
@cptspacemanspiff any updates? Do we still want this change?
cc @JacobSzwejbka

PR #8032, opened 2025-01-29
Enhance load_method to support optional planned memory allocator

cptspacemanspiff:
### Summary
The module api for the executorch runtime is pretty nice, however the load_method function forces the user to use the heirarchical allocator auto-generated by load_method. This is less than ideal if we have to do anything special with our memory allocation.

My specific use-case is having the ability to share memory allocations between multiple methods. Specifically, I have a memory manager does all the bookwork, and the user can request a hierarchical allocator from the manager during the load method call.

This is related to issue #8030 

That being said, this is probably useful in general for cases where a default 'buffer' is not desired, but where one would still like to use the Module extension.

### Test plan
Currently testing has just been that it works, all it does is:
1. add an additional parameter (pointer to hierarchical allocator), by default null.
2. checks if the pointer is null, in which case no allocator was provided and a allocator is created.
3. if non-null skip allocator creation and use the provided allocator. the lifetime of the allocator is handled externally.

cc @JacobSzwejbka @dbort


JacobSzwejbka (commented):
Looks good to me will let @shoumikhin take a look as well

shoumikhin:
@cptspacemanspiff thanks a lot for the idea and implementation, I think that's a great way to provide the customers yet another point of customization. I started leaving suggestings to the code, but then realized it may be more readable as a [demo PR](https://github.com/pytorch/executorch/pull/8046), please feel free to borrow and update yours.
Thanks!

cptspacemanspiff:
@shoumikhin 
Sorry for delay,
I copied your changes and tested to make sure things still worked.
I think this should be good?

cptspacemanspiff:
@shoumikhin

Can you look over this?

shoumikhin (approved):

shoumikhin (approved):

PR #8039, opened 2025-01-29
Add abstract base class for attention mechanisms with unified interface

iseeyuan:
### Summary
Add abstract base class for attention mechanisms with unified interface.
It creates the interface to multiple Attention definitions, like NPU friendly attentions, or other attention types like Multi-Head Latent Attention (MLA) used in Deepseek.
A simple registry is provided to easily add and register a new attention class.

Moved the current attention implementation to attention.py and rename it to AttentionMHA. 

### Test plan
CI

### DIFF
--- a/examples/models/llama/attention.py
+++ b/examples/models/llama/attention.py
@@ -1,0 +1,253 @@
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Optional, Tuple, Type
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from executorch.examples.models.llama.model_args import ModelArgs
+from executorch.examples.models.llama.rope import Rope
+
+
+class Attention(nn.Module, ABC):
+    """Abstract base class for attention mechanisms with unified interface."""
+
+    @abstractmethod
+    def forward(
+        self,
+        x: torch.Tensor,
+        freqs_cos: torch.Tensor,
+        freqs_sin: torch.Tensor,
+        mask: Optional[torch.Tensor] = None,
+        input_pos: Optional[torch.Tensor] = None,
+        in_cache_state: Optional[Any] = None,
+        out_cache_state: Optional[Any] = None,
+    ) -> Tuple[torch.Tensor, Optional[Any]]:
+        """Forward pass for attention mechanism.
+
+        Args:
+            x: Input tensor of shape (batch_size, seq_len, dim)
+            freqs_cos, freqs_sin: Rotary position embedding frequencies
+            mask: Optional attention mask
+            input_pos: Positions for KV cache updates
+            in_cache_state/out_cache_state: Cache states
+
+        Returns:
+            Tuple of (output tensor, updated cache state)
+        """
+        pass
+
+
+ATTENTION_REGISTRY: Dict[str, Type[Attention]] = {}
+
+
+def register_attention(name: str):
+    """Decorator to register attention classes"""
+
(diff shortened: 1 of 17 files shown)


examples/models/llama/attention.py
```diff
@@ -0,0 +1,135 @@
...
+    @abstractmethod
+    def forward(
+        self,
+        x: torch.Tensor,
+        freqs_cos: torch.Tensor,
+        freqs_sin: torch.Tensor,
+        mask: Optional[torch.Tensor] = None,
+        input_pos: Optional[torch.Tensor] = None,
+        in_cache_state: Optional[Any] = None,
+        out_cache_state: Optional[Any] = None,
```
iseeyuan, line 19: replace them with kwargs

iseeyuan:
@CypherpunkSamurai I'm trying to complete the refactor and land it this week. Let me create an issue of adding the MLA to this interface.

examples/models/llama/attention.py
iseeyuan: directly using kwargs may break type-safe. Keep them as is and consider using TypedDict and Unpack for kwarg type checking later.

jackzhxng (commented):
Is this ready for review? Also would probably tag with an actual release note tag since I assume this would be good to highlight in 0.6 release notes

examples/models/llama/attention.py
```diff
@@ -0,0 +1,253 @@
...
+from typing import Any, Dict, Optional, Tuple, Type
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from executorch.examples.models.llama.model_args import ModelArgs
+from executorch.examples.models.llama.rope import Rope
+
+
+class Attention(nn.Module, ABC):
```
sxu, line 11: So far a specialized implementation is only used during lowering and on device, and it needs to be able to accept checkpoint from whatever definition was used during training. What do see as the usage pattern going forward? Is the `AttentionMHA` below the standard definition that specialization of this class needs to support?

iseeyuan: @sxu 
> Is the AttentionMHA below the standard definition that specialization of this class needs to support?
Not necessarily. The attention type is added into the model args. Usually the model arg and checkpoint will be saved in one place. We use model arg to build the model, and load the checkpoint as state_dict. If the checkpoint does not match the model architecture there will be error. We don't break the standard process of PyTorch.

iseeyuan:
> Is this ready for review? Also would probably tag with an actual release note tag since I assume this would be good to highlight in 0.6 release notes

@dvorjackz It's currently in the example model. I'm fine to promote this to extension/llm, rename it to llm_transformer, or simply transformer, and mark it user faced.

examples/models/llama/attention.py
sxu: I see, we don't usual expect the training to be done on a specialized NPU implementation, but I guess we can tweak the state dict loading on a case by case basis.

sxu (commented):
Looks reasonable to me. The cache, mask, and rope frequencies still need to passed down from the top level transformer and updates from each layer returned need to be returned in a follow up PR.

tarun292 (approved):

iseeyuan:
> Looks reasonable to me. The cache, mask, and rope frequencies still need to passed down from the top level transformer and updates from each layer returned need to be returned in a follow up PR.

Thanks @sxu ! I'll land this one when all CI pass, and please feel free to add further PRs to improve it.

PR #8079, opened 2025-01-30
[docs][ez] Fix doc build workflow

SS-JIA:
[Doc build](https://github.com/pytorch/executorch/actions/workflows/doc-build.yml) workflows have been failing for a while due to [redefinition of the `contents:` field](https://github.com/pytorch/executorch/actions/runs/13037800951/workflow) in the workflow file. This is a simple fix for that.

cc @mergennachin @byjlw

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -85,7 +85,6 @@ jobs:
     if: github.repository == 'pytorch/executorch' && github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
     permissions:
       contents: write
-      contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-build.yml
```diff
@@ -85,7 +85,6 @@ jobs:
     if: github.repository == 'pytorch/executorch' && github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
     permissions:
       contents: write
```
svekars, line 87: based on [this](https://github.com/pytorch/executorch/pull/7819/files#diff-1048a3f83dfee0a99ce3c07fede976bbd6a6b65f8bdae55ada11e6a1dda452c7), should this be changed to `id-token: write` @jzhang337

PR #8083, opened 2025-01-31
Fix ATen mode op_upsample_nearest2d_test

swolchok:
* (to be filled)

Was broken, now it's not.

Differential Revision: [D68930165](https://our.internmc.facebook.com/intern/diff/D68930165/)


swolchok:
accidental double-export

PR #8090, opened 2025-01-31
Add support for quantized transposed conv

digantdesai:
Summary: As title. Updates the tests accordingly.

Differential Revision: D68939306

cc @mcr229


mcr229 (approved):
do you need to make in partitioner config changes? Might be useful to tightend some of the check_constraints for transposed convolution.

digantdesai:
There are also some tests failing I need to take a look at those.

PR #8091, opened 2025-01-31
Add hardsigmoid operator to Arm backend

tom-arm:
* Add filter_fn to ops_to_not_decompose, to filter out quantized candidate ops
* Added aten.hardsigmoid as a table op for the BI case

Signed-off-by: Tom Allsop <tom.allsop@arm.com>

### DIFF
--- a/backends/arm/arm_partitioner.py
+++ b/backends/arm/arm_partitioner.py
@@ -113,8 +113,40 @@ class ArmPartitioner(Partitioner):
         self,
         ep: ExportedProgram,
     ) -> Tuple[List[torch._ops.OpOverload], Optional[Callable[[torch.fx.Node], bool]]]:
+        ops_to_not_decompose_if_quant_op = [
+            torch.ops.aten.hardsigmoid.default,
+        ]
+
+        def filter_fn(node: torch.fx.Node) -> bool:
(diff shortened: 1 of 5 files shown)


zingo (approved):

zingo:
macos job fail not related

digantdesai:
> macos job fail not related

This is something we are trying to fix. But running into other issues. Sorry about that.

backends/arm/arm_partitioner.py
```diff
@@ -113,8 +113,40 @@ def ops_to_not_decompose(
         self,
         ep: ExportedProgram,
     ) -> Tuple[List[torch._ops.OpOverload], Optional[Callable[[torch.fx.Node], bool]]]:
+        ops_to_not_decompose_if_quant_op = [
```
digantdesai, line 116: Nit: why another list?

tom-arm: This list is for ops where we specifically only want to not decompose if it's quantized. Also, another operator will be coming in a follow on commit, so a list is needed

digantdesai: I see that you are using the same list in the filter_fn, sounds good, false alarm.

PR #8093, opened 2025-01-31
Adjust tolerance for quantized XNN conv1d tests

GregoryComer:
### Summary
Quantized conv1d tests are flaky due to tolerance being too low. This PR bumps the tolerance slightly. This should be safe, as the underlying kernels are believed to be correct and this issue has been observed for some time (not a recent regression in kernel accuracy).

### Test plan
Ran tests locally for 1000 iterations and did not observe any failures.

### DIFF
--- a/backends/xnnpack/test/ops/test_conv1d.py
+++ b/backends/xnnpack/test/ops/test_conv1d.py
@@ -122,9 +122,13 @@ class TestConv1d(unittest.TestCase):
         # For some tests we want to skip to_executorch because otherwise it will require the
         # quantized operators to be loaded and we don't want to do that in the test.
         if not skip_to_executorch:
-            tester.to_executorch().serialize().run_method_and_compare_outputs(
-                num_runs=10, atol=0.01, rtol=0.01
-            )
+            tester.to_executorch().serialize()
+            if quantized:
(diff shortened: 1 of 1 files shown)


backends/xnnpack/test/ops/test_conv1d.py
```diff
@@ -122,9 +122,13 @@ def _test_conv1d(
         # For some tests we want to skip to_executorch because otherwise it will require the
         # quantized operators to be loaded and we don't want to do that in the test.
         if not skip_to_executorch:
-            tester.to_executorch().serialize().run_method_and_compare_outputs(
-                num_runs=10, atol=0.01, rtol=0.01
-            )
+            tester.to_executorch().serialize()
```
digantdesai, line 125: Might be ok but feels like sliding under the rug 😄 

As an alternative, does it make sense to mark it as flaky (I know this is not the correct use of the flaky tag but..) so the unit-tests won't block CI and create a task to look into it?

mcr229: agreed

JakeStevens:
I have not noticed this flakiness. Closing for now, will monitor

PR #8107, opened 2025-01-31
Refactor and add Llama Python library build

limintang:
Summary: As title. To use static llama export outside QC dir.

Differential Revision: D68937637


cccclai (approved):
Seems reasonable to me. @winskuo-quic @shewu-quic @haowhsu-quic please take a second look

haowhsu-quic (approved):

shewu-quic (approved):

PR #8117, opened 2025-01-31
Remove empty line

kirklandsign:
### Summary
Remove a trailing empty line

@diff-train-skip-merge
In internal repo, it's already removed


GregoryComer (approved):
I think linter might expect a newline at EOF? Not sure, so I'm stamping to unblock.

kirklandsign:
> I think linter might expect a newline at EOF? Not sure, so I'm stamping to unblock.

We should already have EOF. I will wait for the linter.

lucylq (approved):

PR #8120, opened 2025-01-31
Fix doc build workflow

GregoryComer:
### Summary
The doc build job has been broken for a few days since the workflow files were updated. It is currently failing with the following error:
```
The workflow is not valid. .github/workflows/doc-build.yml (Line: 88, Col: 7): 'contents' is already defined
```
I copied the pattern used in the rest of the workflow files updated with the ci linux job v2 change. It looks like this one might have just not been fully updated?

### Test plan
Testing on local PR CI. If it passes here, it should pass on trunk.


kirklandsign (approved):
Stamp to unblock

PR #8124, opened 2025-01-31
Fix pyre

kirklandsign:
Differential Revision: D68986388


GregoryComer (approved):

backends/arm/_passes/fuse_batchnorm2d_pass.py
```diff
@@ -112,6 +112,7 @@ def try_set_param(
             if not try_set_param(conv_bias_node, fused_conv_bias) and try_set_param(
                 bn_bias_node, fused_conv_bias
             ):
+                # pyre-ignore[60]
```
zingo, line 115: Is this still needed? I heard rumors that pyre was disabled on backends/arm when mypy was introduced?
Asking Is we where thinking of removed all # pyre stuff soon but maybe we should wait.

digantdesai: +1

PR #8127, opened 2025-02-01
support input_pos > 0 for prefill model

billmguo:
Summary: test input_pos>0 for prefill, not intention for landing but for sync with qc

Differential Revision: D68847677


billmguo:
let me explain a little on this, Tokens, freq_cos, freq_sin, mask, k, v caches will be passed in both prefill and decode model
The freq_cos /freq_sin/mask . Here we remove  input_pos since prefill model inference will be like AR-pre_fill_len model , the input_pos does not come into play, runtime can fill freq_cos/freq_sin/kv/mask information directly. let me know about your thoughts, in case you have better ideas to support multi-turn, that will be great!

PR #8137, opened 2025-02-03
Bugfix in TosaSpecification representation

per:
Fix up TOSA 1.0 class string representation handling and add test cases for the fixed functionality.

### Test plan
Tests added to unit tests.


zingo (approved):

zingo:
Tests fails unrelated

PR #8142, opened 2025-02-03
don't double-install executorch for one-off linux pull-time jobs

swolchok:
I remember somebody mentioning they wanted install_executorch to be the one solution previously. CI is red for these jobs due to timeouts and we should not block a time improvement just because it is not the ideal long-term solution.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -220,9 +220,6 @@ jobs:
 
         PYTHON_EXECUTABLE=python bash .ci/scripts/setup-linux.sh "cmake"
 
-        # install pybind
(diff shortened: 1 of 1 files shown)


swolchok:
* __->__ #8142

dbort (approved):
I was going off of old information, and was afraid that this change would remove all CI testing of install_executorch.sh.

In the past, the CI setup scripts did their own version of install_executorch, but now it calls down to the script: https://github.com/pytorch/executorch/blob/821a2fee5e730d10dd0ca11714de3ea07d2dcb3a/.ci/scripts/utils.sh#L23

So this looks good! Thanks for removing these.

PR #8143, opened 2025-02-03
[cadence][hifi] update quantized_relu_per_tensor_out signature to match internal flow

zonglinpeng:
Differential Revision: D69015308

### DIFF
--- a/backends/cadence/hifi/operators/op_quantized_relu_out.cpp
+++ b/backends/cadence/hifi/operators/op_quantized_relu_out.cpp
@@ -48,24 +48,27 @@ void quantized_relu_(
 void quantized_relu_per_tensor_out(
     KernelRuntimeContext& ctx,
     const Tensor& input,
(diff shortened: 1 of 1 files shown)


backends/cadence/hifi/operators/op_quantized_relu_out.cpp
```diff
@@ -48,24 +48,27 @@ void quantized_relu_(
 void quantized_relu_per_tensor_out(
```
hsharma35, line 48: Is this not the operator called by ET? If so, please move this inside an anonymous namespace.

hsharma35 (approved):

backends/cadence/hifi/operators/op_quantized_relu_out.cpp
zonglinpeng: realized most of hifi ops are not in anonymous namespace yet. Will do separately

PR #8160, opened 2025-02-04
use _upload_docs job

metascroy:
Following advice here to unblock broken doc job: https://fb.workplace.com/groups/4571909969591489/permalink/9125894517526322/

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -82,41 +82,8 @@ jobs:
 
   upload-gh-pages:
     needs: build
-    if: github.repository == 'pytorch/executorch' && github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
-    permissions:
-      id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-build.yml
```diff
@@ -82,41 +82,10 @@ jobs:
...
-        else
-          TARGET_FOLDER="main"
-        fi
-        echo "Target Folder: ${TARGET_FOLDER}"
-
-        mkdir -p "${TARGET_FOLDER}"
-        # Clean up target folder if exists and copy html output to the
-        # Target folder
-        rm -rf "${TARGET_FOLDER}"/*
-        mv "${RUNNER_ARTIFACT_DIR}"/html/* "${TARGET_FOLDER}"
```
clee2000, line 116: You might want to double check that the files are what you expect them to be.  The upload_docs workflow looks slightly different here

clee2000:
The most recent docs job on main actually succeeded, so it seems the upstream fix did work

Issue #3189, opened 2024-04-20
build of executorch triggers a warning related to too many arguments provided for format string


mikekgfb:
build of executorch triggers a warning related to too many arguments provided for format string.
https://github.com/pytorch/torchchat/actions/runs/8768143958/job/24062159094?pr=327

```
  In file included from /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/../../../executorch/runtime/core/error.h:18,
                   from /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/../../../executorch/runtime/core/result.h:19,
                   from /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/../../../executorch/runtime/core/data_loader.h:14,
                   from /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/../../../executorch/extension/data_loader/mmap_data_loader.h:11,
                   from /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/mmap_data_loader.cpp:9:
  /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/mmap_data_loader.cpp: In function ‘void torch::executor::util::{anonymous}::MunmapSegment(void*, void*, size_t)’:
  /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/mmap_data_loader.cpp:140:9: warning: too many arguments for format [-Wformat-extra-args]
    140 |         "munmap(0x%zx, %zu) failed: %s (ignored)",
        |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/runner/work/torchchat/torchchat/etorch/executorch/extension/data_loader/../../../executorch/runtime/platform/log.h:161:11: note: in definition of macro ‘ET_LOG’
    161 |           _format,                                                            \
        |           ^~~~~~~
  [ 50%] Building CXX object kernels/portable/CMakeFiles/portable_kernels.dir/cpu/op_argmax.cpp.o
```

digantdesai:
@dbort  - is this still relevant? Ok to close?

dbort:
It's actually still a problem; I'll put out a PR to fix it real quick. Thanks for reviving this!

PR #8161, opened 2025-02-04
[data_loader] Fix a log message that didn't use all of its args

dbort:
Fixes issue #3189


swolchok (approved):
personally I would just drop the raw errno value, but if you're sure you want to include it go ahead

dbort:
All of the other logs in this file print both, just going for consistency. Thanks for the quick review!

PR #7922, opened 2025-01-24
--use-pt-pinned-commit in one-off linux jobs

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7579
* __->__ #7922

This is needed to respect the torchgen pinned commit for #7546.

Differential Revision: [D68593517](https://our.internmc.facebook.com/intern/diff/D68593517/)

PR #8173, opened 2025-02-04
Increase CI timeout

metascroy:
Since https://github.com/pytorch/executorch/pull/7922, 2 CI jobs consistently hit timeout.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -212,7 +212,7 @@ jobs:
       docker-image: executorch-ubuntu-22.04-clang12
       submodules: 'true'
       ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 90
+      timeout: 180
       script: |
         # The generic Linux job chooses to use base env, not the one setup by the image
         CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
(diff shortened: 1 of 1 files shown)


swolchok:
Something is definitely wrong. In [the raw logs](https://ossci-raw-job-status.s3.amazonaws.com/log/pytorch/executorch/36506265502) for the first test-llava-runner-linux timeout on main, there are almost 40 minutes of "gaps" in the logs with no timestamps. Specifically:

- 25 minute "gap in logs from 2025-01-31T23:25:21.2243143Z to 2025-01-31T23:42:40.6914293Z , and the second message is just a job timeout.

swolchok:
tracking issue: #8180

metascroy:
> Are people really going to wait this long? I already had no patience with the jobs taking an hour.
> 
> I looked into it briefly and I dont know why they would be taking longer; the pinned PyTorch commit is prebuilt at docker container generation time and I’m fairly sure it is in fact a release build.

It's a good question, but currently this is one of the issues blocking viable/strict from updating and I thought having a short-term "fix" as the investigation continues would be better than nothing.

There are several ways we could fix:

* Increase the timeout (this PR)
* Increase the runner size (your https://github.com/pytorch/executorch/issues/8180); although, it looks like the CI is waiting for the larger runners, so I don't know if it will be faster than increasing the timeout.
* Revert https://github.com/pytorch/executorch/pull/7922
* Disable the tests

swolchok (approved):
unblocking seems good

PR #8181, opened 2025-02-04
try to use 48xlarge runners for llava-runner and eval_llama-mmlu

### LINKED ISSUE #8180: CI timeout (test-llava-runner-linux) since #7922
### 🐛 Describe the bug

### Versions

N/A

swolchok:
These jobs are timing out. I found big gaps in the timestamps of the logs even before the blame PR, which to me suggests swapping. Let's try bigger runners and see what happens on this PR.

Fixes #8180 (if it lands)

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -208,7 +208,7 @@ jobs:
     strategy:
       fail-fast: false
     with:
-      runner: linux.24xlarge
+      runner: linux.48xlarge
       docker-image: executorch-ubuntu-22.04-clang12
       submodules: 'true'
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

swolchok:
1) looks like there may be no such thing as a 48xlarge runner. The runners appear to use c5 AWS instances, and [24xlarge is the largest in that category](https://aws.amazon.com/ec2/instance-types/).
2) c5.24xlarge instances have 192 GB of RAM. If we're swapping on one of those, something is probably wrong.

Issue #6955, opened 2024-11-19
I pulled the latest code, and the model is reporting errors everywhere


yangh0597:
### 🐛 Describe the bug

I pulled the latest code, and the model is reporting errors everywhere. Two days ago it was fine.Looks like the code forgot to commit

command is:
```
 python -m examples.models.llama.export_llama --checkpoint "${MODEL_DIR}/consolidated.00.pth" -p "${MODEL_DIR}/params.json" -kv --disable_dynamic_shape --qnn --pt2e_quantize qnn_16a16w -d fp32 --num_sharding 4 --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' --output_name="Llama-Guard-3.2-1B-qnn_16a16w_s4.pte"
```

error is 

```
Traceback (most recent call last):
  File "/opt/anaconda3/envs/et_qnn/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/anaconda3/envs/et_qnn/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/executorch/examples/models/llama/export_llama.py", line 32, in <module>
    main()  # pragma: no cover
  File "/opt/executorch/examples/models/llama/export_llama.py", line 28, in main
    export_llama(args)
  File "/opt/executorch/examples/models/llama/export_llama_lib.py", line 508, in export_llama
    builder = _export_llama(args)
  File "/opt/executorch/examples/models/llama/export_llama_lib.py", line 783, in _export_llama
    from executorch.backends.qualcomm.utils.utils import canonicalize_program
ImportError: cannot import name 'canonicalize_program' from 'executorch.backends.qualcomm.utils.utils' (/opt/executorch/backends/qualcomm/utils/utils.py)

```
I checked the/opt/executorch/backends/qualcomm/utils/utils. Py files, do not canonicalize_program this method

### Versions

main

cc @cccclai

yangh0597:
> Tagged this with QNN.
> 
> @yangh0597 I noticed that the executorch modules are not calling something from an anaconda environment (/opt/executorch/examples/models/llama/export_llama.py), werhereas runpy is (/opt/anaconda3/envs/et_qnn/lib/python3.10/runpy.py). Did you install ET to your anaconda environment after pulling in the latest code?
What is ET?
[...]

PR #8188, opened 2025-02-04
Fix missing canonicalize_program in qnn

metascroy:
Addresses https://github.com/pytorch/executorch/issues/6955

cc @cccclai


cccclai (approved):
Thank you!

PR #8190, opened 2025-02-04
Add `strict=True` explicit flag to Executorch invocations to torch.export.export

gmagogsfm:
Differential Revision: D69068797


iseeyuan (approved):
LGTM. Would we moved to strict=False as suggested?

PR #8193, opened 2025-02-04
Trial revert of --use-pt-pinned-commit for llava

swolchok:
Now that we've bumped the nightly version, let's make sure that --use-pt-pinned-commit is definitely causing the long running times here (#8180).


swolchok:
* __->__ #8193

Issue #7775, opened 2025-01-20
Export llama3.1 Runtime error: Missing out variants: {'quantized_decomposed::dequantize_per_token'.......


BodhiHu:
### 🐛 Describe the bug

Hello, we got and runtime error when trying to export the llama3.1 8B model, here's the reproduce script:

```bash
#!/bin/bash

checkpoint=path/to/meta-llama-3.1-8b-instruct/original/consolidated.00.pth
params=path/to/meta-llama-3.1-8b-instruct/original/params.json
output_name="llama3_kv_sdpa_xnn_qe_4_32.pte"

python -m examples.models.llama.export_llama \
  --checkpoint "$checkpoint" \
  -p "$params" \
  -kv \
  --use_sdpa_with_kv_cache \
  -X \
  -qmode 8da4w \
  --group_size 128 \
  -d bf16 \
  --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' \
  --embedding-quantize 4,32 \
  --output_name="$output_name"
```

Then we got error:

```
INFO:root:Failed converting '<EdgeOpOverload: quantized_decomposed.quantize_per_token.default>: schema = quantized_decomposed::quantize_per_token(Tensor input, Tensor scales, Tensor zero_points, int quant_min, int quant_max, ScalarType dtype) -> Tensor' to its out variant with error: 'SchemaKind.out variant of operator quantized_decomposed::quantize_per_token can't be found. We've found the schemas of all the overloads: ['quantized_decomposed::quantize_per_token(Tensor input, Tensor scales, Tensor zero_points, int quant_min, int quant_max, ScalarType dtype) -> Tensor']'
INFO:root:Failed converting '<EdgeOpOverload: quantized_decomposed.dequantize_per_token.default>: schema = quantized_decomposed::dequantize_per_token(Tensor input, Tensor scales, Tensor zero_points, int quant_min, int quant_max, ScalarType dtype, ScalarType output_dtype) -> Tensor' to its out variant with error: 'SchemaKind.out variant of operator quantized_decomposed::dequantize_per_token can't be found. We've found the schemas of all the overloads: ['quantized_decomposed::dequantize_per_token(Tensor input, Tensor scales, Tensor zero_points, int quant_min, int quant_max, ScalarType dtype, ScalarType output_dtype) -> Tensor']'
[...]

PR #8195, opened 2025-02-04
Warn if wrong type is given for Llama export for XNNPACK

mergennachin:
Followup to https://github.com/pytorch/executorch/issues/7775

Test Plan:

```
(executorch) mnachin@mnachin-mbp executorch % python -m examples.models.llama.export_llama \
    --checkpoint /Users/mnachin/models/Llama-3.2-1B/original/consolidated.00.pth \
        -p /Users/mnachin/models/Llama-3.2-1B/original/params.json \
        -kv \
        --use_sdpa_with_kv_cache \
        -X \
        -qmode 8da4w \
        --group_size 128 \
        -d bf16 \
        --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' \
        --embedding-quantize 4,32 \
        --output_name="Llama-8B.pte"
```

```
import error: No module named 'triton'
  File "/Users/mnachin/miniconda/envs/executorch/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/mnachin/miniconda/envs/executorch/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/mnachin/executorch/examples/models/llama/export_llama.py", line 32, in <module>
    main()  # pragma: no cover
  File "/Users/mnachin/executorch/examples/models/llama/export_llama.py", line 28, in main
    export_llama(args)
  File "/Users/mnachin/executorch/examples/models/llama/export_llama_lib.py", line 542, in export_llama
    builder = _export_llama(args)
  File "/Users/mnachin/executorch/examples/models/llama/export_llama_lib.py", line 685, in _export_llama
    _validate_args(args)
  File "/Users/mnachin/executorch/examples/models/llama/export_llama_lib.py", line 679, in _validate_args
    raise ValueError(
ValueError: XNNPACK supports either fp32 or fp16 dtypes only for now. Given bf16.
```

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -674,6 +674,12 @@ def _validate_args(args):
                 "If you need this feature, please file an issue."
             )
 
+    if args.xnnpack:
+        if args.dtype_override not in ["fp32", "fp16"]:
+            raise ValueError(
+                f"XNNPACK supports either fp32 or fp16 dtypes only for now. Given {args.dtype_override}."
+            )
+
 
 def _export_llama(args) -> LLMEdgeManager:  # noqa: C901
     _validate_args(args)


mcr229 (approved):

mergennachin:
@mcr229 @digantdesai 

wait a sec, how does Llama3.2 1B/3B bf16, unquantized format work for XNNPACK?

https://github.com/pytorch/executorch/blob/main/examples/models/llama/README.md#option-a-download-and-export-llama32-1b3b-model

mcr229:
@mergennachin I don't think that should work...

swolchok:
@mergennachin it doesn't work with XNNPACK. unquantized bf16 does not use XNNPACK.

mergennachin:
@swolchok 

Actually I just tried reproducing it. It's the other way around. For unquantized bf16 1B/3B bf16 model:

With -X and with -d bf16 flags during export, it is successfully able to generate '.pte' file and the llama runner is able to execute. Here's even an internal CI test: https://fburl.com/code/4nvjis6p

Without -X and with -d bf16 flag during export, it is also successfully able to generate '.pte' file but the llama runner is error-ing out with 
   ```
   E 00:00:01.963468 executorch:tensor_util.h:482] Check failed (t.scalar_type() == dtype): Expected to find Half type, but tensor has type BFloat16
   ```

swolchok:
> @swolchok
> 
> Actually I just tried reproducing it. It's the other way around. For unquantized bf16 1B/3B bf16 model:
> 
> With -X and with -d bf16 flags during export, it is successfully able to generate '.pte' file and the llama runner is able to execute. Here's even an internal CI test: https://fburl.com/code/4nvjis6p
> 
> Without -X and with -d bf16 flag during export, it is also successfully able to generate '.pte' file but the llama runner is error-ing out with
> 
> ```
> E 00:00:01.963468 executorch:tensor_util.h:482] Check failed (t.scalar_type() == dtype): Expected to find Half type, but tensor has type BFloat16
> ```

hmm, this is strange. XNNPACK did not support bfloat16 last I checked, and I elected to support it with portable/optimized ops instead of adding that support to XNNPACK. something is wrong.

swolchok:
I would guess that probably the -X flag just doesn't do anything, but then why does the exported model work with it but not without? needs investigation

swolchok:
> just tried reproducing it

@mergennachin are you following the direction from that readme you linked? the cmake command to build the runner is failing to find executorchConfig.cmake, probably because it doesn't set CMAKE_PREFIX_PATH

swolchok:
> Actually I just tried reproducing it. It's the other way around. For unquantized bf16 1B/3B bf16 model:

I tried as well; works both with and without the -X flag for me on the executorch main branch, commit hash 9020fd2f9b762068f285a2e0c4efd9885d6af1fa

mcr229:
@swolchok yea this makes sense, if unquantized bf16, applying -X would do nothing as our partitioner would recognize none of the ops as lowerable, so this should do nothing. There would only be a problem with quantized bf16, because again XNNPACK would be unable to lower anything, and the resulting quantization ops would throw an error at the to_executorch stage.

digantdesai (approved):

PR #8198, opened 2025-02-04
method_meta add uses_backend function

cmt0:
Summary: For a runtime that supports models of a couple different types it helps to be able to check ahead of time whether certain backend types are present. In our case, there are processors that can use this information to make scheduling decisions without having the actual required backend implementations present.

Differential Revision: D69143825

### DIFF
--- a/runtime/executor/method_meta.cpp
+++ b/runtime/executor/method_meta.cpp
@@ -228,6 +228,17 @@ Result<int64_t> MethodMeta::memory_planned_buffer_size(size_t index) const {
   return s_plan_->non_const_buffer_sizes()->Get(index + 1);
 }
 
+bool MethodMeta::uses_backend(const char* backend_name) const {
+  const auto delegates = s_plan_->delegates();
+  for (size_t i = 0; i < delegates->size(); i++) {
+    auto delegate = delegates->Get(i);
+    if (strcmp(delegate->id()->c_str(), backend_name) == 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
 size_t MethodMeta::num_instructions() const {
   const auto chains = s_plan_->chains();
   if (chains == nullptr) {
--- a/runtime/executor/method_meta.h
+++ b/runtime/executor/method_meta.h
@@ -186,6 +186,14 @@ class MethodMeta final {
   Result<int64_t> memory_planned_buffer_size(size_t index) const;
 
   /**
+   * Check to see if a backend is used in this method.
+   *
+   * @param[in] backend_name The name of the backend to search for.
+   * @returns true if a backend is used in this method, otherwise false.
+   */
+  bool uses_backend(const char* backend_name) const;
+
+  /**
    * Get the number of instructions in this method.
    *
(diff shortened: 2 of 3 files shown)


runtime/executor/method_meta.h
```diff
@@ -192,6 +192,13 @@ class MethodMeta final {
    */
   ET_EXPERIMENTAL size_t num_instructions() const;
 
+  /**
+   * Check to see if a backend is used in this method.
+   *
+   * @returns Whether a backend is used in this method.
+   */
+  ET_EXPERIMENTAL bool contains_backend(const char* backend_name) const;
```
dbort, line 200: Please call it `uses_backend`: a method doesn't "contain" a backend per se.

runtime/executor/method_meta.h
```diff
@@ -192,6 +192,13 @@ class MethodMeta final {
    */
   ET_EXPERIMENTAL size_t num_instructions() const;
 
+  /**
+   * Check to see if a backend is used in this method.
+   *
+   * @returns Whether a backend is used in this method.
+   */
```
dbort, line 199: Please add a `@param`, and clarify true-vs-false rather than "whether".

```suggestion
  /**
   * Check to see if a backend is used by this method.
   *
   * @param backend_name The name of the backend to search for.
   * @returns true if this method uses this backend; false otherwise.
   */
```

runtime/executor/method_meta.cpp
```diff
@@ -248,5 +248,16 @@ size_t MethodMeta::num_instructions() const {
   return num_instructions;
 }
 
+bool MethodMeta::contains_backend(const char* backend_name) const {
```
dbort, line 251: Please add a test for this new method.

You could add a couple lines to https://github.com/cmt0/executorch/blob/09938978bf530202032943c89c5e7f482d1032a2/runtime/executor/test/backend_integration_test.cpp#L343, checking that it's true for a backend named `kName`, and that it's false for some other string like "INVALID_BACKEND_NAME".

dbort (changes requested):
Thanks for adding this! It's a reasonable feature.

runtime/executor/method_meta.h
cmt0: Added clarification

runtime/executor/method_meta.cpp
cmt0: Good idea, added a test which does this

cmt0:
@dbort Does it need an ET_EXPERIMENTAL tag, or can I define it normall to avoid deprecated declarations warning?

runtime/executor/method_meta.h
dbort: Please add the @param line too

dbort:
> @dbort Does it need an ET_EXPERIMENTAL tag, or can I define it normall to avoid deprecated declarations warning?

I think the new method is general enough to remove the ET_EXPERIMENTAL annotation. The behavior is well-defined compared to e.g. the instruction count.

dbort (approved):

PR #8209, opened 2025-02-05
Arm backend: Add is_node_supported checks for 4 ops

Erik-Lundell:
For convolution, maxpool2d, avgpool2d, and sum.
The checks mostly target hardware constraints on
Ethos-U55, though convolution also checks for some
unsupported behavior.


Erik-Lundell:
CI failure is unrelated, the PR only limits what ops we partition and should not affect accuracy

backends/arm/test/ops/test_max_pool.py
```diff
@@ -246,3 +246,39 @@ def test_maxpool2d_tosa_u85_BI_mult_batches(
...
             tester.run_method_and_compare_outputs(qtol=1, inputs=(test_data,))
+
+    reject_data_suite = [
+        (MaxPool2d(1, 1, 0), torch.rand(2, 5, 5, 5)),
+        (MaxPool2d(1, 4, 0), torch.rand(1, 10, 10, 10)),
+        (MaxPool2d((1, 257), 1, 0), torch.rand(1, 16, 5, 300)),
+        (MaxPool2d((800, 90), 1, 0), torch.rand(1, 16, 850, 100)),
+    ]
+
+    @parameterized.expand(reject_data_suite)
```
zingo, line 257: Do we need to move multi test here also?
Strangly its marked as @conftest.expectedFailureOnFVP so it should have worked anyway but I see this error 

FAILED backends/arm/test/ops/test_max_pool.py::TestMaxPool2d::test_maxpool2d_tosa_u55_BI_mult_batches_0_randn - RuntimeError: Expected to find "torch.ops.higher_order.executorch_call_delegate" but did not find it
Searched string:

Erik-Lundell: Ah you are correct, sorry about that. I'll simply delete the test since there is already a reject test for N>1

zingo (approved):

PR #8233, opened 2025-02-05
Fix allocation-size-too-big crash in prepare_input_tensors

dbort:
Summary:
(Adapted from an LLM-suggested fix for a fuzzer-discovered crash)

The crash is an allocation-size-too-big error that occurs when the `prepare_input_tensors` function attempts to allocate an excessively large amount of memory for the `inputs` array. This crash is caused by the function's inability to handle large numbers of inputs, resulting in an attempt to allocate a huge amount of memory that exceeds the system's limits.

The root cause of the crash is the lack of bounds checking on the `num_inputs` variable, which allows the function to attempt to allocate an arbitrarily large amount of memory. This is exacerbated by the fact that the function allocates memory for each input tensor separately, without checking the total size of all tensors before allocating memory for the `inputs` array.

The patch fixes the crash by adding bounds checking on the `num_inputs` variable and calculating the total size of all tensors before allocating memory for the `inputs` array.

Differential Revision: D68876117

### DIFF
--- a/extension/runner_util/inputs.cpp
+++ b/extension/runner_util/inputs.cpp
@@ -22,15 +22,36 @@ using executorch::runtime::TensorInfo;
 namespace executorch {
 namespace extension {
 
-Result<BufferCleanup> prepare_input_tensors(Method& method) {
+Result<BufferCleanup> prepare_input_tensors(
+    Method& method,
+    PrepareInputTensorsOptions options) {
   MethodMeta method_meta = method.method_meta();
   size_t num_inputs = method_meta.num_inputs();
-  size_t num_allocated = 0;
-  void** inputs = (void**)malloc(num_inputs * sizeof(void*));
 
+  // A large number of small allocations could exhaust the heap even if the
+  // total size is smaller than the limit.
+  ET_CHECK_OR_RETURN_ERROR(
+      num_inputs <= options.max_inputs,
+      InvalidProgram,
+      "Too many inputs: %zu > %zu",
+      num_inputs,
+      options.max_inputs);
+
+  // Allocate memory for the inputs array
+  void** inputs = (void**)malloc(num_inputs * sizeof(void*));
+  ET_CHECK_OR_RETURN_ERROR(
+      inputs != nullptr,
+      MemoryAllocationFailed,
+      "malloc(%zd) failed",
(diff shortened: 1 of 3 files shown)


extension/runner_util/inputs.cpp
```diff
@@ -40,10 +62,29 @@ Result<BufferCleanup> prepare_input_tensors(Method& method) {
     }
     Result<TensorInfo> tensor_meta = method_meta.input_tensor_meta(i);
     if (!tensor_meta.ok()) {
+      BufferCleanup cleanup({inputs, num_allocated});
```
dbort, line 65: This PR also fixes this buffer leak

extension/runner_util/inputs.cpp
```diff
@@ -22,15 +22,37 @@ using executorch::runtime::TensorInfo;
...
-  size_t num_allocated = 0;
+
+  // A large number of small allocations could exhaust the heap even if the
+  // total size is smaller than the limit.
+  ET_CHECK_OR_RETURN_ERROR(
+      num_inputs <= max_inputs,
+      InvalidProgram,
+      "Too many inputs: %zu > %zu",
+      num_inputs,
+      max_inputs);
```
larryliu0820, line 39: Can we add a test for this?

extension/runner_util/inputs.cpp
```diff
@@ -40,10 +62,29 @@ Result<BufferCleanup> prepare_input_tensors(Method& method) {
...
+      ET_LOG(
+          Error,
+          "Allocating %zu bytes for input %zu would exceed "
+          "max_total_allocation_size %zu",
+          tensor_size,
+          i,
+          max_total_allocation_size);
+      BufferCleanup cleanup({inputs, num_allocated});
+      return Error::InvalidProgram;
+    }
```
larryliu0820, line 81: We can test this as well

dbort:
I was thinking that testing would require adding PTE files with a large number of inputs and large allocations, but I could use an existing PTE and just reduce the limits. I'll look into it.

dbort:
Added tests, thanks for pushing me to do so :) So many of these fuzzer fixes aren't unit-testable without corrupt PTE files, but this fix is.

larryliu0820 (approved):

PR #8248, opened 2025-02-06
Install mkl for PyTorch and libjpeg/libpng for TorchVision

### LINKED ISSUE #8180: CI timeout (test-llava-runner-linux) since #7922

huydhn:
Fixes https://github.com/pytorch/executorch/issues/8180

I should have caught this earlier that we didn't have mkl here, so PyTorch performance on x86 CPU suffered.


swolchok (approved):
looks like you've got conda library problems but this is clearly a PR I should accept, contingent on you seeing llava-runner job improve

huydhn:
> looks like you've got conda library problems but this is clearly an accept contingent on you seeing llava-runner job improve

Let me look more into that conda issue.  This is rather unexpected, I have done a rebase, so the failure is legit.

huydhn:
> looks like you've got conda library problems but this is clearly a PR I should accept, contingent on you seeing llava-runner job improve

Just FYI, the linking error is from this hack https://github.com/pytorch/executorch/blob/main/.ci/docker/common/install_conda.sh#L46-L58

huydhn:
The run time is around 1h20m now https://github.com/pytorch/executorch/actions/runs/13229090225/job/36927181503 v.s 2h30m https://github.com/pytorch/executorch/actions/runs/13213708877/job/36890417988

swolchok:
eval_llama-mmlu seems to have gone from 1.9h to under 9 minutes: https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=mmlu&mergeLF=true

PR #8277, opened 2025-02-06
Implement _fft_r2c core ATen op

swolchok:
Mostly following suit on PyTorch impl here. (For your reference: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/mkl/SpectralOps.cpp#L281)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,3 +67,6 @@
 [submodule "backends/cadence/utils/FACTO"]
 	path = backends/cadence/utils/FACTO
 	url = https://github.com/pytorch-labs/FACTO.git
+[submodule "third-party/pocketfft"]
+	path = third-party/pocketfft
+	url = https://github.com/mreineck/pocketfft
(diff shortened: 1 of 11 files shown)


swolchok:
* (to be filled)

swolchok:
update coming soon to fix internal build issues

iseeyuan (approved):
LGTM. Thanks @swolchok !

kimishpatel:
no code sharing for this one?

swolchok:
> no code sharing for this one?

What's to share? The interesting bits are pocketfft, and we use that, same as PyTorch core.

kimishpatel:
> > no code sharing for this one?
> 
> What's to share? The interesting bits are pocketfft, and we use that, same as PyTorch core.

I havent looked in depth but shouldnt this file kernels/optimized/cpu/op_fft_r2c.cpp be subset of https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/mkl/SpectralOps.cpp#L281

PR #8278, opened 2025-02-06
Fix unittests failures introduced from D69212362

zhxchen17:
Summary: There'll be some minor test failures caused by D69212362, fixing these before the source diff get landed.

Differential Revision: D69251115


larryliu0820 (approved):
Thanks for the pre-emptive fix

PR #8296, opened 2025-02-07
Move tensor_shape_to_c_string back to runtime/core

swolchok:
I previously moved it to the wrong place because of build issues. Had to couple this with replacing `sizes_to_string` so that the buck2->cmake system would place it in executorch_core instead of portable_kernels and not break the build.


swolchok:
* (to be filled)

swolchok:
this doesn't build because the new file is getting placed in `portable_kernels__srcs` by the buck2->cmake stuff, but kernels/portable/CMakeLists.txt doesn't respect `portable_kernels__srcs` and instead globs it itself! long-term plan is to stop using buck2 to drive CMake builds, so a short-term plan of "couple this with replacing dim_order_util's sizes_to_string" seems OK.

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -893,16 +894,6 @@ inline bool tensor_is_scalar(executorch::aten::Tensor t) {
...
 
-/**
- * The expected output size may not be the existing size of any inputs and
- * outputs if the operator supports both broadcast and dynamic shape.
- * Therefore such operators needs extra space to store the calculated expected
- * output size. such dynamic allocation is troublesome in executorch so we can
- * just hard code a static value of a relatively small value because users
- * don't create high dimensional tensors.
- */
-constexpr size_t kTensorDimensionLimit = 16;
```
dbort, line 904: Removing this is technically a BC-breaking change to the public API. To follow our policy we'd need to mark this as ET_DEPRECATED and keep it around before removing it in release 0.8.

dbort: Oh ok I see it moved to a different header, and this header includes that one.

dbort (approved):

PR #8298, opened 2025-02-07
Add a pass to convert rank-0 tensor to rank-1 tensor

cccclai:
Differential Revision: D69281867

Test with following code, and no long see the error/warning to complain rank-0 tensor
```

class Model(torch.nn.Module):
    def forward(self, x, y):
        return x + y

model = Model()
model.eval()

example_inputs = (torch.tensor(1.0), torch.tensor(2.0))
exported_program_manager_aten = torch.export.export(model, example_inputs)
exported_program_manager_edge = executorch.exir.to_edge(
    exported_program_manager_aten
).transform([Rank0ToRank1Pass()])

delegated_module = to_backend(
    CoreMLBackend.__name__, exported_program_manager_edge.exported_program(), []
)
```

### DIFF
--- a/backends/transforms/rank_0_to_rank_1.py
+++ b/backends/transforms/rank_0_to_rank_1.py
@@ -1,0 +1,18 @@
+import torch
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class Rank0ToRank1Pass(ExportPass):
+    """
+    Replace Rank-0 Tensor to Rank-1 Tensor for all the inputs.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
+        for node in graph_module.graph.nodes:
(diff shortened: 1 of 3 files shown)


shoumikhin (approved):

shoumikhin:
@cccclai @YifanShenSZ  I wonder if Core ML is not gonna support scalar tensors, does it make sense to embed such workaround somewhere deeper to happen automatically under the hood when using the Core ML backend?

metascroy:
> @cccclai @YifanShenSZ I wonder if Core ML is not gonna support scalar tensors, does it make sense to embed such workaround somewhere deeper to happen automatically under the hood when using the Core ML backend?

Specifically, adding new cases here to cast scalars to tensors: https://github.com/pytorch/executorch/blob/main/backends/apple/coreml/runtime/delegate/coreml_backend_delegate.mm#L72-L95

It looks like ct.convert already does the cast, we just don't have it in CoreML's ET runtime.

manuelcandales (approved):

PR #8304, opened 2025-02-07
oss qnn bug fix

billmguo:
Summary:
1. the graph actually include the torch.ops.aten.silu.default not the torch.nn.functional.silu
2. the aten.mul.tensor is arg is scalar, this could apply to other ops

Differential Revision: D69286388


haowhsu-quic (approved):
Thank you.

limintang (approved):

PR #8307, opened 2025-02-07
[ArmBackend] Fix in setup.sh for MacOS

freddan80:
Make sure setup continues if FVP's are already installed

cc @digantdesai @per @zingo @oscarandersson8218 @mergennachin @byjlw


zingo (approved):

zingo:
macos fail unrelated

PR #8313, opened 2025-02-07
Add lift scalar to constant tensor pass

cccclai:
Differential Revision: D69318149

Sometime the op in the graph will receive a scalar input, and quantization can only deal with tensor. In this case, the recommended workflow is to lift the scalar constant to be a tensor, and annotate them, then these constants will have corresponding q/dq nodes.

This PR is mostly a demonstration regarding where to insert the pass and how the pass should look like


haowhsu-quic (commented):
This is super helpful for us instead of tackling constant inside op builder. Will try make it more general and do some clean up jobs in codebase.
Thank you for this PR.

cccclai:
https://github.com/pytorch/executorch/pull/8642 is merged instead

PR #8314, opened 2025-02-07
Use tensor_shape_to_c_string for error in check_mask_indices

swolchok:
Rolling out for #7902


swolchok:
* __->__ #8314

kernels/portable/cpu/util/advanced_index_util.cpp
```diff
@@ -49,9 +50,22 @@ bool check_mask_indices(const Tensor& in, TensorOptList indices) {
...
+                    index.sizes().data(), index.sizes().size()));
+            auto input_shape = executorch::runtime::tensor_shape_to_c_string(
+                executorch::runtime::Span<const Tensor::SizesType>(
+                    in.sizes().data() + in_i, index.sizes().size()));
+#endif // ET_LOG_ENABLED
+            ET_LOG(
+                Error,
+                "The shape of mask index %s must match the sizes of the corresponding input dimensions %s.",
+                mask_shape.data(),
+                input_shape.data());
```
dbort, line 66: Same here, I think it'd be less surprising if the ifdef covered the entire scope of the vars that are defined inside it.

```suggestion
#ifdef ET_LOG_ENABLED
            auto mask_shape = executorch::runtime::tensor_shape_to_c_string(
                executorch::runtime::Span<const Tensor::SizesType>(
                    index.sizes().data(), index.sizes().size()));
            auto input_shape = executorch::runtime::tensor_shape_to_c_string(
                executorch::runtime::Span<const Tensor::SizesType>(
                    in.sizes().data() + in_i, index.sizes().size()));
            ET_LOG(
                Error,
                "The shape of mask index %s must match the sizes of the corresponding input dimensions %s.",
                mask_shape.data(),
                input_shape.data());
#endif // ET_LOG_ENABLED
```

dbort (approved):

manuelcandales (approved):

cccclai:
It seems like this PR breaks multiple tests

swolchok:
> It seems like this PR breaks multiple tests

ugh, why do we have tests that only run on trunk, grumble grumble

swolchok:
suspiciously it is mostly qnn stuff + test-binary-size-linux. investigating

swolchok:
#8403 should fix test-binary-size-linux

swolchok:
#8404 might fix QNN, running ciflow/trunk on the stack

PR #8317, opened 2025-02-07
s/IF_FALSE/UNLESS/ in ET_LOG macro names

swolchok:
"do something unless X" reads much better than "do something if false X".


swolchok:
* (to be filled)

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -338,14 +338,19 @@
...
+#define ET_LOG_AND_RETURN_UNLESS(cond)             \
   do {                                             \
     if (!(cond)) {                                 \
       ET_LOG(Error, "Check failed (%s): ", #cond); \
       return false;                                \
     }                                              \
   } while (false)
 
+/**
+ * Backward compatibility shim for the old name of ET_LOG_AND_RETURN_UNLESS.
```
dbort, line 350:
```suggestion
 * DEPRECATED: Please use ET_LOG_AND_RETURN_UNLESS instead.
```

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -354,14 +359,19 @@
...
+#define ET_LOG_MSG_AND_RETURN_UNLESS(cond, message, ...)                  \
   do {                                                                    \
     if (!(cond)) {                                                        \
       ET_LOG(Error, "Check failed (%s): " message, #cond, ##__VA_ARGS__); \
       return false;                                                       \
     }                                                                     \
   } while (false)
 
+/**
+ * Backward compatibility shim for the old name of ET_LOG_MSG_AND_RETURN_UNLESS
```
dbort, line 371:
```suggestion
 * DEPRECATED: Please use ET_LOG_MSG_AND_RETURN_UNLESS instead.
```

runtime/core/exec_aten/util/tensor_util.h
```diff
@@ -338,14 +338,19 @@
  *
  * @param[in] cond the condition to check
  */
-#define ET_LOG_AND_RETURN_IF_FALSE(cond)           \
+#define ET_LOG_AND_RETURN_UNLESS(cond)             \
```
dbort, line 341: These feel like they should live near https://github.com/pytorch/executorch/blob/d99970ba72d44ab80087fc95bfdf68f3d363f87d/runtime/core/error.h#L121 and be called ET_CHECK_OR_RETURN_FALSE to be consistent with the core helpers. But that's a separate issue since this PR doesn't change where they live.

dbort (approved):
Explicitly marking these as DEPRECATED will let us remove them in a couple of releases

swolchok:
will abandon and send PR to move these directly to their endpoint

PR #8323, opened 2025-02-07
Expand dict API to allow dtype strings

lucylq:
Summary:
Allow users to specify dtypes using strings, instead of kernel keys.

Note: this generates a 'fake' kernel key with no dim order data. I think this is OK, as I don't think ExecuTorch uses the kernel key for anything besides dtype.

https://www.internalfb.com/code/fbsource/[c02ea3c66de17a60b2f0e65462a03ccf5cba9804]/fbcode/caffe2/torchgen/selective_build/selector.py?lines=49-52

Differential Revision: D69335674


larryliu0820 (approved):
Thank you for adding the test

PR #8325, opened 2025-02-08
Change llm getting started tutorial to use to_edge_transform_and_lower

tarun292:
Change the LLM getting started tutorial to use to_edge_transform_and_lower instead of to_edge + to_backend to be consistent with the other XNNPack tutorials.

cc @mergennachin @byjlw


GregoryComer (approved):
Looks good to me. Thanks for updating this.

PR #8327, opened 2025-02-08
Add safety checks when rendering kernel key strings

dbort:
Summary:
The old code assumed that it was handed a MAX_SIZE buffer, and that the list of TensorMeta values would never generate a string longer than that size.

This PR adds explicit size tracking and an error code to the API, and now returns an error if the buffer is too small for the provided values.

While I'm here, move MAX_SIZE out of the public API, since it's not an intrinsic aspect of kernel keys. This is technically a BC-breaking change, but I don't expect that any users are actually depending on it.

Also:
 - Fix some unused var warnings, which uncovered a couple places where we ignored an Error. Along similar lines, replaced EXPECT with ASSERT when validating kernel registration so we don't try to run the rest of the test after a registration error.
- Silence the Meta-internal linter by moving to std::array in several code chunks I needed to touch anyway because of the new `make_kernel_key` parameter.

Add unit tests for all modified code.

Differential Revision: D69324821

### DIFF
--- a/runtime/kernel/operator_registry.cpp
+++ b/runtime/kernel/operator_registry.cpp
@@ -114,44 +114,106 @@ Error register_kernels(const Span<const Kernel> kernels) {
 }
 
 namespace {
-int copy_char_as_number_to_buf(char num, char* buf) {
-  if ((char)num < 10) {
+/**
+ * Writes `num` as a decimal string to `buf` and returns the number of bytes
+ * written. Returns -1 if `buf` is too small or if `num` is not supported.
+ */
+int copy_char_as_number_to_buf(int num, char* buf, size_t buf_size) {
+  if (num < 0) {
+    return -1;
+  }
+  if (num < 10) {
+    if (buf_size < 1) {
+      return -1;
+    }
     *buf = '0' + (char)num;
-    buf += 1;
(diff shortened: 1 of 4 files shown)


runtime/kernel/operator_registry.cpp
```diff
@@ -114,44 +114,106 @@ Error register_kernels(const Span<const Kernel> kernels) {
...
+    if (buf_size > 0) {
+      buf[0] = '\0';
+    }
+    return Error::Ok;
   }
-  strncpy(buf, "v1/", 3);
+
+  // Reserve one byte for null terminator.
+  if (buf_size < 1) {
+    return Error::InvalidArgument;
```
larryliu0820, line 159: Do you think it's useful to have a new error type like BufferTooSmall?

larryliu0820 (approved):
Thank you for adding the check! Just left one comment

runtime/kernel/operator_registry.cpp
dbort: I think that's too specific, and not very actionable by users who get it. Even InvalidArgument isn't very useful once it's bubbled up to higher levels tbh. The logs are what will provide the most information.

PR #8330, opened 2025-02-09
Add XNN U8 op support via quantization

GregoryComer:
### Summary
Support U8 ops in the XNNPACK delegate by treating the input and output tensors as u8 asymmetric-quantized tensors with scale=1 and zero_point=0. This PR adds U8 support for upsample_bilinear2d, cat, slice, and _to_copy (when used to convert u8 to f32). More ops are possible with this method.

Conversion from u8 to f32 is done via transformation into a dequantize op. This is implemented in a new pass - ReplaceU8ConvertWithDqPass. The general U8 to quantized U8 transformation is done in `define_tensor` in `node_visitor.py`. U8 inputs are created as quantized tensors with the appropriate qparams (scale=1, zp=0).

### Test plan
I've added op-level u8 tests to each of the new ops, as well as tests for the ReplaceU8ConvertWithDqPass. I've also added an end-to-end test for MobileNetV3 with a wrapper to take U8 inputs, resize and crop, and then convert to f32 and run the model.

### DIFF
--- a/backends/xnnpack/_passes/__init__.py
+++ b/backends/xnnpack/_passes/__init__.py
@@ -24,6 +24,9 @@ from executorch.backends.xnnpack._passes.fuse_batch_norm_with_conv import (
 )
 from executorch.backends.xnnpack._passes.prelu_reshape_pass import PReLUReshapePass
 from executorch.backends.xnnpack._passes.remove_getitem_op import RemoveGetItemPass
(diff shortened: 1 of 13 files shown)


mcr229 (approved):
This generally looks right to me.

digantdesai:
Ah reviewed internally first, please look at the comments there, next time I will remember.

PR #8336, opened 2025-02-10
Support multi-module exports in Inspector and ETRecord

cptspacemanspiff:
So, this is a prototype of changes required to get performance data out of a multi-method export with ETRecords and ETDump.

I do not have a good understanding of if this is correct/the best way, and my understanding of the performance analysis stuff is limited.

## What this does:

1. In _etrecord.py, it allows for having a blank edge_dialect_program, and just exporting the ETRecord with export_modules:
```
export_modules [Optional]: **Should be ignored by OSS users**. A dictionary of graph modules with the key being the user provided name and the
            value being the corresponding exported module. The exported graph modules can be either the
            output of `torch.export()` or `exir.to_edge()`.
```

2. In the inspector tool, adds the ability to specify the module/method to pull the data from the ETRecord (since we are not using the global edge_program_method.) 

FWIW I was able to get the statistics table with node labeling with these changes.

From the sounds of this this might be used internally and there is probably a better way to handle this?

related to #8030

### DIFF
--- a/devtools/etrecord/_etrecord.py
+++ b/devtools/etrecord/_etrecord.py
@@ -232,9 +232,10 @@ def generate_etrecord(
             edge_dialect_program.exported_program,
         )
     else:
-        raise RuntimeError(
-            f"Unsupported type of edge_dialect_program passed in {type(edge_dialect_program)}."
-        )
+        if export_modules is None:
+            raise RuntimeError(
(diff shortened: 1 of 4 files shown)


cptspacemanspiff:
@JacobSzwejbka

shoumikhin:
@JacobSzwejbka, @Tarun292, @Gasoonjia can you take a look please?

PR #8337, opened 2025-02-10
Minor improvements in Arm testing.

Erik-Lundell:
Add markers to flaky tests
Check that reference model was run in TosaReferenceModelDispatch
If the if statment looking for executorch_call_delegates fails,
the delegate can run the model in eager mode and still produce correct results
without the user noticing. We need to make sure we actually run the reference model.


zingo (approved):

zingo:
MacOs fails unrelated

PR #8338, opened 2025-02-10
Make stage ids unique in the Arm TestPipeline

AdrianLundell:
Some stages in the TestPipeline may be added multiple times, such as .check(). To be able to target these by id, give them an unique suffix -> 'id.suffix' Refering to stages in terms of id instead of an index is more self documenting and future proof.

This change modifies the add/pop_stage interface:
  - pos arg in add_stage is now an optional kwarg, appending to the pipline as default
  - suffix is added to add_stage as an optional_kwarg. If a suffix is not given to a non unique stage, a number is added instead.
  - pop_stage now allows to use ids for referring to stages.

Additionally adds .visualize(stage) for quickly adding visualizing stages to the pipeline.


zingo (approved):

zingo:
MacOs fails seems unrealted

PR #8339, opened 2025-02-10
Lower timeout for llava-runner and eval_llama-mmlu jobs again

### LINKED ISSUE #8180: CI timeout (test-llava-runner-linux) since #7922

swolchok:
#8173 raised these timeouts. Now that #8248 has landed to fix #8180, we should be able to lower them again. (I'm sending this early so I don't forget; double-check llava-runner running time)


swolchok:
* (to be filled)

huydhn (approved):

metascroy (approved):

huydhn:
Let me try to get some stats on how much time had been wasted because of missing mkl

PR #8345, opened 2025-02-10
Add extension/llm/custom_ops/spinquant/test to CMake

swolchok:
Need to run it to make sure build changes in this directory aren't breaking things.

Test Plan: added EXPECT_TRUE(false) to fast_hadamard_transform_test.cpp, saw failure in test/run_oss_cpp_tests.sh


swolchok:
* (to be filled)

swolchok:
this isn't correct yet; test doesn't run when I run test/run_oss_cpp_tests.sh and I'm not sure why yet.

swolchok:
temporarily adding `--trace` to the problem cmake command in build-llm-demo since I can't repro locally

swolchok:
> temporarily adding `--trace` to the problem cmake command in build-llm-demo since I can't repro locally

the correct cmake command was in extension/android_test/setup.sh, but I still can't figure out what's going on so now i'm just tracing all 4 cmake invocations

swolchok:
the inexplicable build of tests seems to be because including abseil-cpp's CMakeLists.txt sets BUILD_TESTING to ON by default because it does `include(CTest)`. trying out find_package(GTest) in et_cxx_test to see if it fixes the issue

swolchok:
this looks good, please review

swolchok:
ping reviewers @kirklandsign @kimishpatel @larryliu0820 @jackzhxng @iseeyuan @GregoryComer @dbort

GregoryComer (approved):

PR #8354, opened 2025-02-11
Python hugging face tokenizer

jackzhxng:
### Summary
Add python tokenizer for hugging face tokenizer format (-`.json` files)

### Test plan
- Existing CI runner tests for regression on TikToken / SentencePiece
- Generated coherent output on Qwen2.5 with HuggingFace tokenizer.json


examples/models/llama/runner/eager.py
```diff
@@ -74,6 +75,12 @@ def build_args_parser() -> argparse.ArgumentParser:
         help="Have multi-turn chat with the model",
     )
 
+    parser.add_argument(
+        "--tokenizer_config_path",
+        type=str,
+        default=None,
+    )
```
larryliu0820, line 82: Please add `help` to clarify how to use this argument

larryliu0820: Also this should be mutually exclusive with tokenizer_path? We should make sure only one is passed in.

extension/llm/tokenizer/utils.py
```diff
@@ -4,16 +4,23 @@
...
 )
 
 
-def get_tokenizer(tokenizer_path):
-    try:
-        tokenizer = SentencePieceTokenizer(model_path=str(tokenizer_path))
-    except Exception:
-        print("Using Tiktokenizer")
-        tokenizer = Tiktoken(model_path=str(tokenizer_path))
+def get_tokenizer(tokenizer_path: str, tokenizer_config_path: Optional[str] = None):
```
larryliu0820, line 15: I don't like how this is written lol. If we are taking both arguments it's ambiguous which one is the source of truth.

examples/models/llama/runner/eager.py
jackzhxng: Should be just metadata like [this](https://huggingface.co/Qwen/Qwen2.5-1.5B/blob/main/tokenizer_config.json), e.g. it tells you what the eos_token is, which isn't info that is in the `tokenizer.json` itself

extension/llm/tokenizer/utils.py
larryliu0820: A better way is to ask the user to pass in a directory. We then try our best to determine if the directory contains a `tokenizer.[model|json]` and try to parse it. If no such file or the format is wrong, we error out.

jackzhxng: Discussion continued in above comment ^

larryliu0820 (approved):

Issue #6813, opened 2024-11-13
How to convert tokenizer of SmolLM model as accepted by executorch


Arpit2601:
Hi,
I am trying to convert [SmolLm-135M-Instruct](https://huggingface.co/HuggingFaceTB/SmolLM-135M-Instruct) model to .pte format and then run on an android device.
I have been successful in converting the model but executorch requires the tokenizer in either .bin format or .model format which can then be converted into .bin format. But on huggingface tokenizer.model or tokenizer.bin files are not present. 

How would I go about converting the tokenizer.json file into the appropriate format.

larryliu0820:
@guangy10 do you know the answer to this?

guangy10:
I tried it a while ago. `tokenizer.save_pretrained` will save the json format, even with `legacy=True`it doesn't save to the format that can be accepted by the llama_runner. I was trying to use the `tokenizers.save` for the convention as shown below, which is WIP and I haven't got a chance to back on it
```
import os
import argparse
from tokenizers import Tokenizer
from tokenizers.models import BPE
from tokenizers.pre_tokenizers import Whitespace
def create_tokenizer_model(input_dir, output_file):
    vocab_file = os.path.join(input_dir, "vocab.json")
    merges_file = os.path.join(input_dir, "merges.txt")
    # Create BPE model from files
    bpe = BPE.from_file(vocab_file, merges_file)
    
    # Create tokenizer
    tokenizer = Tokenizer(bpe)
    
    # Set pre-tokenizer
    tokenizer.pre_tokenizer = Whitespace()
    
    # Save the tokenizer model
    tokenizer.save(output_file)
    print(f"Tokenizer model saved to {output_file}")
    # Verify the tokenizer
    loaded_tokenizer = Tokenizer.from_file(output_file)
    test_text = "Hello, world! This is a test."
    encoded = loaded_tokenizer.encode(test_text)
    decoded = loaded_tokenizer.decode(encoded.ids)
    print(f"Test encoding/decoding: '{test_text}' -> '{decoded}'")
if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="Create a tokenizer")
    parser.add_argument("--input_dir", help="Directory containing vocab.json and merges.txt")
    parser.add_argument("--output", default="tokenizer.model", help="Output file name (default: tokenizer.model)")
    args = parser.parse_args()
    create_tokenizer_model(args.input_dir, args.output)
```

Arpit2601:
[...]

PR #8355, opened 2025-02-11
Add qwen 2.5

jackzhxng:
### Summary
Add qwen model. Instead of loading the model and directly exporting, we load it into a format for `export_llama` to consume via `llama_transformer.py`.

Was a relatively painless process, used TorchTune checkpointing utils but didn't need to, was only for convenience.

### Test plan
Convert weights:
```
python examples/models/qwen2_5/convert_weights.py <path-to-checkpoint-dir> <output-path>
```

Exporting (without quantization and at `fp32` because ao quantizer doesn't support bias in 8da4w and xnnpack doesn't support bf16):
```
python -m examples.models.llama.export_llama   \
--model qwen2_5 --params examples/models/qwen2_5/1_5b_config.json  \
--checkpoint qwen2_5-1_5b.pth -kv --use_sdpa_with_kv_cache  \
-X -d fp32 --metadata '{"get_bos_id":151643, "get_eos_ids":[151643]}'  \
--output_name="qwen2_5-1_5b.pte" --verbose
```

Generate:
```
python -m examples.models.llama.runner.native \
--model qwen2_5 --pte qwen2_5-1_5b.pte \
--tokenizer ~/.cache/huggingface/hub/models--Qwen--Qwen2.5-1.5B/snapshots/8faed761d45a263340a0528343f099c05c9a4323/tokenizer.json \
--prompt "Who is the founder of Meta?" --params examples/models/qwen2_5/1_5b_config.json \
--max_len 64 --temperature 0 -kv
```

cc @mergennachin @iseeyuan @lucylq @helunwencser @tarun292 @kimishpatel


examples/models/llama/attention.py
```diff
@@ -175,9 +175,10 @@ def __init__(self, args: ModelArgs, layer_id: int, rope: Rope):
         self.max_batch_size = args.max_batch_size
         self.max_context_len = args.max_context_len
         self.dim = args.dim
-        self.wq = nn.Linear(self.dim, self.n_heads * self.head_dim, bias=False)
-        self.wk = nn.Linear(self.dim, self.n_kv_heads * self.head_dim, bias=False)
-        self.wv = nn.Linear(self.dim, self.n_kv_heads * self.head_dim, bias=False)
+        # TODO: parametrize bias for attention and feedforward.
```
jackzhxng, line 178: TODO: flag this

examples/models/llama/model.py
```diff
@@ -170,7 +171,7 @@ def __init__(self, **kwargs):
 
         # Within the device="meta" context, tensors that are created do not carry data.
         # They possess all other metadata a tensor carries such as size, stride, requires_grad.
-        with torch.device("meta"):
+        with torch.device("cpu"):
```
jackzhxng, line 174: TODO: find out why meta doesn't work

examples/models/llama/rope.py
```diff
@@ -209,18 +210,67 @@ def hf_apply_rotary_emb_to_k(k, cos, sin, position_ids=None, unsqueeze_dim=1):
     return k_embed
 
 
+# ======================= Qwen2 Implementation ========================
```
iseeyuan, line 213: If we put existing llama_transformer implementation, and Qwen2 implementation, side by side, what are the differences?

examples/models/llama/model.py
iseeyuan: Using cpu may introduce extra memory usage.

examples/models/llama/attention.py
```diff
@@ -175,9 +175,10 @@ def __init__(self, args: ModelArgs, layer_id: int, rope: Rope):
...
+        self.wq = nn.Linear(self.dim, self.n_heads * self.head_dim, bias=True)
```
iseeyuan, line 179: Can we add this in ModelArgs?

examples/models/llama/rope.py
jackzhxng: Ah I meant to remove this.

examples/models/llama/model.py
lucylq: Hmn, seems like using 'meta' has been problematic for many open source models. What error did you run into when exporting with meta?

jackzhxng: It works again. Not sure what I was doing the first time around to run into the error

examples/models/llama/attention.py
jackzhxng: Should be okay to do for now

larryliu0820:
Can we add a test? Basically extend `EagerModelBase` and add it to `examples/models/__init__.py`

examples/models/llama/model.py
jackzhxng: I found the issue - https://github.com/pytorch/executorch/pull/8487. It was happening before I set the output token embedding equal to the input token embedding, which left it uninitialized as a meta tensor

jackzhxng:
@larryliu0820 had to make some model loading changes to make the `EagerModelBase` test work, separated into a separate PR: https://github.com/pytorch/executorch/pull/8552

mergennachin:
Can you also a README.md page?

guangy10:
@mergennachin Thank you for tagging me on this PR. 

BTW, this model has been enabled in HuggingFace [optimum-executorch](https://github.com/huggingface/optimum-executorch) a while ago using HF's modeling code, with the simplest XNNPACK recipe (`fp32`) as a proof-of-experience. Everybody can quickly run it by just following the Quick Start and replace the model_id with `"Qwen/Qwen2.5-0.5B"`.

IMO, the real gap preventing users from deploying this model on-device is converting the Hugging Face tokenizer into a binary format recognizable by llama_runner. Enabling this feature is tracked in this task: [pytorch/executorch#6813](https://github.com/pytorch/executorch/issues/6813). This issue arises after leaving the Hugging Face ecosystem and attempting deployment on the actual target device. I explored this a while ago but haven’t had the time to revisit it yet. We should prioritize to fill this gap if we think it's high priority.

I can see that PR adds an additional path for users to convert HF weights using the ET-friendly modeling code (`llama_transformer.py`), which is good. However, it also adds additional examples, which I’m somewhat concerned about—depending on how far we want to go down this path. Do we plan to keep adding tens of thousands of examples to our repo? If no, where do we draw the line?

jackzhxng:
@guangy10 I don't think that this is apples-to-apples with whatever is on HuggingFace. What's here isn't just for "proof-of-experience", this is able to freely leverage all of our previous work in optimizing the `llama_transformer` for our Llama models. This involves source transformation for custom flash attention kernels (`--use_sdpa_with_kv_cache`), quantization (embedding, kv_cache, and linear), etc. I haven't done any benchmarks yet since I still have a blocker on the ao side for quantization, but I expect this to be very performant out of the box.

As for the tokenizer, we just added a json tokenizer for Python to use with pybindings [here](https://github.com/pytorch/executorch/pull/8354), and @larryliu0820 is working on integrating [one for C++](https://github.com/pytorch-labs/tokenizers/blob/main/src/hf_tokenizer.cpp) into ExecuTorch.

Also adding this specific example wasn't very difficult at all - it's much easier than writing a model. I had a few hiccups along the way to deal with setting up things such as the json tokenizer and (currently) bias support for quantization, but overall it wasn't too involved. The goal is that future transformer decoder models can quickly be integrated this way to leverage all of our pre-built performance features.

guangy10:
> This involves source transformation for custom flash attention kernels (--use_sdpa_with_kv_cache), quantization (embedding, kv_cache, and linear), etc.

What prevent using these optimizations as building blocks outside ExecuTorch repo?

larryliu0820:
@guangy10 I think it's a very good user experience to leverage optimum-executorch whenever it's possible. I can see some limitations though:
1. Do we have a good story for the models not supported by optimum-executorch?
2. What should we tell a user to do, if they want to lower the model to a backend other than xnnpack? What about quantization?

Moving forward I suppose optimum-executorch will gradually increase coverage but we should have a coherent solution for people wanting to use ExecuTorch for the scenarios mentioned above.

guangy10:
> @guangy10 I think it's a very good user experience to leverage optimum-executorch whenever it's possible. I can see some limitations though:
> 
> 1. Do we have a good story for the models not supported by optimum-executorch?

Trying to get more clarity on this. Do you mean if the model is not published to Hugging Face? For example, some private modeling or weights from an end user?

> 2. What should we tell a user to do, if they want to lower the model to a backend other than xnnpack? What about quantization?
> 
> Moving forward I suppose optimum-executorch will gradually increase coverage but we should have a coherent solution for people wanting to use ExecuTorch for the scenarios mentioned above.

Initially I think we should host all model code rewriting and recipes (to other backends and with quantization) in `optimum` as it's how other accelerators and backends work on HF. After chatting with @tarun292, I'm open to host canonical recipes in our repo and use it elsewhere. This is possible by leveraging @tarun292's work. I shared the doc with you 😄 (both mine and Tarun's proposal is WIP so won't share it publicly here), and we can discuss it in tomorrow's meeting.

jackzhxng:
@guangy10 Definitely possible - I'm just highlighting that optimum-executorch at the moment is pretty basic and only supports a simple XNNPack export, so it will likely take a good amount of effort match our performance llama_transformer.

examples/models/llama/attention.py
```diff
@@ -175,9 +175,16 @@ def __init__(self, args: ModelArgs, layer_id: int, rope: Rope):
         self.max_batch_size = args.max_batch_size
         self.max_context_len = args.max_context_len
         self.dim = args.dim
-        self.wq = nn.Linear(self.dim, self.n_heads * self.head_dim, bias=False)
-        self.wk = nn.Linear(self.dim, self.n_kv_heads * self.head_dim, bias=False)
-        self.wv = nn.Linear(self.dim, self.n_kv_heads * self.head_dim, bias=False)
+        self.attention_qkv_bias = args.attention_qkv_bias
```
iseeyuan, line 178: Should we add the same to static_attention, so that Qwen would work with accelerators?

larryliu0820:
> Do you mean if the model is not published to Hugging Face? For example, some private modeling or weights from an end user?

No like "Falconsai/text_summarization", although it's hosted by HF, this model can't leverage optimum-executorch. I haven't looked into details but from a user's perspective I would be curious to learn what model is supported by optimum-executorch.

iseeyuan (approved):
LGTM as a good example on how export_llama can be extended to other models, with optimizations for all backends!

examples/models/qwen2_5/convert_weights.py
```diff
@@ -0,0 +1,90 @@
...
+        output_dir=".",
+        model_type="QWEN2",
+    )
+
+    print("Loading checkpoint...")
+    sd = checkpointer.load_checkpoint()
+
+    print("Converting checkpoint...")
+    sd = qwen_2_tune_to_meta(sd["model"])
+    # torch.save(sd, "/home/jackzhxng/models/qwen2_5-1_5b.pth")
```
swolchok, line 83: delete?

PR #8367, opened 2025-02-11
Add additional checks for operator support.

Erik-Lundell:
This can be used to avoid partitioning parts
of a model when debugging.

Though any OperatorSupportBase can be used, we add three OperatorSupport as utilities:

- DontPartition: Don't partition based on node target
- DontPartitionName: Don't partition based on node name
- DontPartitionModule: Don't partition based on which module the op comes from.

All these checks can match parts of the target name, and save a list of the nodes they reject for debugging.


zingo (approved):

zingo:
MacOs fails unrelated

PR #8370, opened 2025-02-11
Arm backend: Add Deeplab3 test

Erik-Lundell:
The model causes problems that we see in other models such as incorrect quantization and int64s showing up.
I plan to use the model as a case study for improving what errors we get.


zingo:
MacOS and llama fails are unrelated

zingo (approved):

PR #8384, opened 2025-02-11
Fix logging.info not appearing

jackzhxng:
### Summary
Logs at the info level were not appearing for `export_llama`, omitting useful logs for intermediate export states with `--verbose`. This also fixes for when a debugger is being used.

### Test plan
Ran `export_llama` instructions from README


lucylq (approved):
lgtm

as a note, seems like this is overriding the warning log in xnnpack partitioner:
https://github.com/pytorch/executorch/blob/main/backends/xnnpack/partition/xnnpack_partitioner.py#L26

could potentially move that to info?

jackzhxng:
Yeah it's overriding the earliest logging levels which is set somewhere in the export_llama dependencies. Maybe one of the recursive module imports recently had a logging level initialized to WARNING or lower.

Since logging level [can only be configured once](https://stackoverflow.com/a/12158233), should we remove the logging.basicConfig calls in modules such as https://github.com/pytorch/executorch/blob/main/extension/llm/export/builder.py#L46? If they want to specify a level they should probably be using named loggers cc @dbort

PR #8388, opened 2025-02-11
Unbreak build-benchmark-app

swolchok:
We delete the headers after symlinking them into frameworks. whoops!

### DIFF
--- a/build/build_apple_frameworks.sh
+++ b/build/build_apple_frameworks.sh
@@ -211,7 +211,7 @@ sed -i '' '1i\
 sed -i '' '1i\
 #define C10_USING_CUSTOM_GENERATED_MACROS
 ' $HEADERS_PATH/executorch/runtime/core/portable_type/c10/macros/Export.h
-ln -s $HEADERS_PATH/executorch/runtime/core/portable_type/c10 "$HEADERS_PATH/"
+cp -r $HEADERS_PATH/executorch/runtime/core/portable_type/c10 "$HEADERS_PATH/"
 
 
 cp "$SOURCE_ROOT_DIR/extension/apple/ExecuTorch/Exported/"*.h "$HEADERS_PATH/executorch"


swolchok:
* __->__ #8388

shoumikhin (approved):

build/build_apple_frameworks.sh
```diff
@@ -211,7 +211,7 @@ sed -i '' '1i\
 sed -i '' '1i\
 #define C10_USING_CUSTOM_GENERATED_MACROS
 ' $HEADERS_PATH/executorch/runtime/core/portable_type/c10/macros/Export.h
-ln -s $HEADERS_PATH/executorch/runtime/core/portable_type/c10 "$HEADERS_PATH/"
```
shoumikhin, line 214: iirc, here's what worked for me:
`ln -s executorch/runtime/core/portable_type/c10 "$HEADERS_PATH/c10"`

swolchok: I would expect that we are likely to accidentally produce a framework containing a symbolic link. I would like to retreat to safety. I don't expect saving disk I/O for a couple dozen text files to be a significant win in 2025.

shoumikhin:
Happy to unblock, but curious what the root cause is?
Symlinks merely point to some location, and we do have all the headers inside .xcframework, so I'd guess it's all about composing the symlink to point to the eventual headers location?

swolchok:
> the root cause 

- we make a symlink (the patched line)
- [we delete the linked-to-files](https://github.com/pytorch/executorch/blob/main/build/build_apple_frameworks.sh#L253)
- [we attempt to copy the linked-to-files later, but they are gone](https://github.com/pytorch/executorch/blob/main/.github/workflows/apple.yml#L342)

I agree that we could get it working with a symlink, I just don't think it is an impactful thing to spend time on.

PR #8389, opened 2025-02-11
Remove duplicated portable test targets

kirklandsign:
### Summary
`_common_op_test` already generates portable test targets.

### Test plan
CI


swolchok (approved):
shipit

PR #8403, opened 2025-02-12
s/ifdef ET_LOG_ENABLED/if ET_LOG_ENABLED/

swolchok:
Looks like we define it to 0 rather than leaving it undefined to disable.


swolchok:
* (to be filled)

cccclai (approved):
This PR seems reasonable but not sure how it fixes the CI..

swolchok:
> how it fixes the CI.

when ET_LOG_DISABLED was defined to 0 (disabling logging), we were creating the strings to log (gated on #ifdef) but ET_LOG itself was disabled (gated on #if), so the strings were unused, triggering -Wunused-variable.

PR #8404, opened 2025-02-12
executorch-the-CMake-target should re-export executorch_core, shouldn't it?

swolchok:
This seems to fix QNN build failures, and it doesn't make sense to me that every executorch (the CMake target) consumer should also have to explicitly link executorch_core

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -596,7 +596,7 @@ endif()
 # any backends.
 #
 add_library(executorch ${_executorch__srcs})
-target_link_libraries(executorch PRIVATE executorch_core)
+target_link_libraries(executorch PUBLIC executorch_core)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -596,7 +596,7 @@ endif()
 # any backends.
 #
 add_library(executorch ${_executorch__srcs})
-target_link_libraries(executorch PRIVATE executorch_core)
+target_link_libraries(executorch PUBLIC executorch_core)
```
cccclai, line 599: I'm not quite sure how this change fixes the error, but looks like CI is green

cccclai (approved):
Accept to unblock

swolchok:
landing despite two job cancellations for timeout because it seems unlikely this would be related and it's an unbreak

PR #8413, opened 2025-02-12
Move backend docs to new locations

GregoryComer:
### Summary
This PR moves backend docs into their new location under Backends/. This doesn't make any content changes yet, but it done in preparation for doing so.

### Test plan
Built docs locally and validated.


mergennachin:
Have you done `git mv` command? It seems like you did copy and delete? 

That way, we can preserve the git history for these files.

mergennachin (changes requested):
Sorry, it's a bit annoying to do this -- but it would be really great to preserve git history with `git mv` command. This is especially important as these docs are written by our partners.

GregoryComer:
> Sorry, it's a bit annoying to do this -- but it would be really great to preserve git history with `git mv` command. This is especially important as these docs are written by our partners.

It's a good point. I'll re-do this to make sure history is explicitly maintained.

PR #8415, opened 2025-02-12
Add pass for replacing dq-q patterns with rescale

Erik-Lundell:
When an int8 op meets an int32 op, the int8 is first dequantized to a float which is then quantized to the desired int8 dtype.

This produces a (int8 dq -> q int32) pattern that
we can replace with a TOSA.RESCALE since they are approximately mathematically equivalent, differing only in how the rounding is done.

This requires a few changes:
- Introduce custom rescale op
- Create pass to replace the dq-q pattern with the rescale op
- Implement node_visitor for rescale op

The change makes it possible to mix int8 and int32 quantization, as showcased in the new test_add_i32_tosa_BI test.

### DIFF
--- a/backends/arm/test/ops/test_add.py
+++ b/backends/arm/test/ops/test_add.py
@@ -9,6 +9,8 @@
 from typing import Tuple
 
 import torch
+from executorch.backends.arm.arm_backend import get_tosa_version
+from executorch.backends.arm.quantizer import arm_quantizer
 from executorch.backends.arm.test import common
 from executorch.backends.arm.test.tester.test_pipeline import (
     EthosU55PipelineBI,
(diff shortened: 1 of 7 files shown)


zingo (approved):

backends/arm/test/ops/test_add.py
```diff
@@ -9,13 +9,19 @@
 from typing import Tuple
 
 import torch
+from executorch.backends.arm.arm_backend import get_tosa_version
```
dbort, line 12: This seems to have broken the unittest-arm job:
https://github.com/pytorch/executorch/actions/runs/13311971065/job/37176555490#step:15:11475
```
_______________________ ERROR collecting ops/test_add.py _______________________
ImportError while importing test module '/pytorch/executorch/backends/arm/test/ops/test_add.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/conda/envs/py_3.10/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
backends/arm/test/ops/test_add.py:12: in <module>
    from executorch.backends.arm.arm_backend import get_tosa_version
E   ImportError: cannot import name 'get_tosa_version' from 'executorch.backends.arm.arm_backend' (/opt/conda/envs/py_3.10/lib/python3.10/site-packages/executorch/backends/arm/arm_backend.py)
```

zingo: Yes i also think it did, sorry for the messup. I conform with a revert fixes it. If so lets merge it and fix/retry this PR again later.

PR #8423, opened 2025-02-12
Fix HiFi relu for int8

mcremon-meta:
Summary: As titled.

Differential Revision: D69541729

### DIFF
--- a/backends/cadence/hifi/operators/op_quantized_relu_out.cpp
+++ b/backends/cadence/hifi/operators/op_quantized_relu_out.cpp
@@ -26,11 +26,11 @@ void quantized_relu_per_tensor_out(
     const int64_t out_multiplier,
     const int64_t out_shift,
     Tensor& output) {
-  const uint8_t _in_zero_point = static_cast<uint8_t>(in_zero_point);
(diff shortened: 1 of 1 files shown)


backends/cadence/hifi/operators/op_quantized_relu_out.cpp
```diff
@@ -26,11 +26,11 @@ void quantized_relu_per_tensor_out(
     const int64_t out_multiplier,
     const int64_t out_shift,
     Tensor& output) {
-  const uint8_t _in_zero_point = static_cast<uint8_t>(in_zero_point);
-  const uint8_t _out_zero_point = static_cast<uint8_t>(out_zero_point);
-  const int32_t _out_multiplier = static_cast<int32_t>(out_multiplier);
```
zonglinpeng, line 31: do these two need to move?

zonglinpeng (approved):

backends/cadence/hifi/operators/op_quantized_relu_out.cpp
mcremon-meta: they don't you're right

PR #8401, opened 2025-02-12
Enable pybind with mps

cccclai:
Summary:
install executorch with mps
```
./install_executorch.sh --pybind mps
```
test the mps .pte file can run with
```
executorch_module = _load_for_executorch(filename)
out = executorch_module.forward(list(inputs))

print("output: ", out)
```

Differential Revision: D69497059

PR #8433, opened 2025-02-12
Fix lint error due to main function too complex

cccclai:
https://github.com/pytorch/executorch/pull/8401 causes lintrunner error and it complains main function too complex. This PR refactors the main function a bit to address the error


swolchok (approved):
thanks!

PR #8436, opened 2025-02-13
ANE-friendly static llama

metascroy:
This directory contains ANE-friendly Llama models.

Export model with:
```
python export.py -n /path/to/output/model.pte -p /path/to/params.json -c /path/to/model.pth --seq_length 64 --max_seq_length 1024 --coreml-quantize c4w
```

The runner is written in python and is only intended to serve as an example for how the model inputs should be processed; it is not performant.

Run model with:
```
python run.py -m /path/to/model.pte -p /path/to/params.json -t /path/to/tokenizer.model --seq_length 64 --max_seq_length 1024 --prompt "Once upon a time," --n_steps 512
```

The model here is based on a "sliding" cache, where old tokens are evicted from the cache.  By default, the cache size is max_seq_length - seq_length, but you can explicitly pass in a smaller cache size (e.g., --cache_size 512).  This can speed up computation and reduce memory.  Keep in mind that once cache_size is reached, older tokens get evicted from the cache and do not participate in attention.

cc @kimishpatel @YifanShenSZ @cymbalrush


examples/apple/coreml/llama/llama_transformer.py
```diff
@@ -0,0 +1,565 @@
...
+                k_caches[i] if self.use_cache_list else k_caches[i, :, :, :, :],
+                v_caches[i] if self.use_cache_list else v_caches[i, :, :, :, :],
+                attn_mask,
+            )
+            k_out.append(new_k)
+            v_out.append(new_v)
+
+        if not self.generate_full_logits:
+            # Only the last logit is used for the new generated token
+            h = h[:, input_length - 1, :]
```
YIWENX14, line 404: Add `.squeeze(1)` to make h of 2d shape?

examples/apple/coreml/llama/llama_transformer.py
```diff
@@ -0,0 +1,564 @@
...
+                ] = new_v_caches[i][:, :, start : (start + length), :]
+        else:
+            assert new_k_caches.shape == self.get_cache_shape(self.seq_length)
+            assert new_v_caches.shape == self.get_cache_shape(self.seq_length)
+            self.k_caches[:, :, :, (self.cache_pos) : (self.cache_pos + length), :] = (
+                new_k_caches[:, :, :, start : (start + length), :]
+            )
+            self.v_caches[:, :, :, (self.cache_pos) : (self.cache_pos + length), :] = (
+                new_v_caches[:, :, :, start : (start + length), :]
+            )
```
YifanShenSZ, line 507: This still looks like an index put to me? Does it successfully run on ANE?

examples/apple/coreml/llama/export.py
```diff
@@ -0,0 +1,265 @@
...
+        super(SplitLinearModule, self).__init__()
+        self.num_splits = max(out_features // target_size, 1)
+        self.common_size = out_features // self.num_splits
+        self.remainder = out_features % self.num_splits
+        self.splits = torch.nn.ModuleList(
+            [
+                torch.nn.Linear(in_features, self.common_size)
+                for _ in range(self.num_splits)
+            ]
+        )
```
YifanShenSZ, line 40: Awesome! So split linear is found to be more performant on ANE? Empirically `1024` is found to be the best?

PS: On our end we found split softmax would be more performant https://github.com/apple/coremltools/pull/2418

examples/apple/coreml/llama/extract_and_combine.py
```diff
@@ -0,0 +1,77 @@
...
+        src_function_name="main",
+        target_function_name="model1",
+    )
+    desc.add_function(
+        f"{output_dir}/model2.mlpackage",
+        src_function_name="main",
+        target_function_name="model2",
+    )
+    desc.default_function_name = "model1"
+    ct.utils.save_multifunction(desc, f"{output_dir}/combined.mlpackage")
```
YifanShenSZ, line 77: Core ML multifunction is already runable now via ExecuTorch?

YifanShenSZ (approved):
Generally LGTM! The major changes I found are:
1. The dedicated KV cache
2. The split linear

Anything I missed?

examples/apple/coreml/llama/export.py
metascroy: We haven't done extensive testing yet on different hardware to find the right splitting values yet.  I just noticed that 1024 works better on my M1 Pro recently.

On the SDPA pass: https://github.com/apple/coremltools/pull/2418:

We observed something similar.  We can get better Llama performance by processing tokens in smaller seq_length chunks (e.g., 256) (this not only chunks the SDPA, but all ops).  This is easy enough to do, but it only chunks the Q seq_length (target_seq_length) in SDPA.  It doesn't chunks the source_seq_length (which is more realistically the bigger value from the K/V caches, e.g., max_context_length).  I suspect the chunking will help here too.  But unlike chunking the target_seq_length, chunking the source_seq_length will require decomposing the SDPA op.  Do you have plans to add support for this?

examples/apple/coreml/llama/extract_and_combine.py
metascroy: No, I need to remove this script from the PR.  I'll do an update.

examples/apple/coreml/llama/llama_transformer.py
metascroy: InputManager and its methods (_update_cache, etc) are not part of the model.  We will have a C++ implementation of it that runs on CPU.  This python implementation is intended to serve as a reference for the C++ one.

cccclai (approved):
LGTM! Let's build on top of it. Also there are lots of errors in CI

metascroy:
> LGTM! Let's build on top of it. Also there are lots of errors in CI

I'll take a look at the CI failures.

examples/apple/coreml/llama/llama_transformer.py
```diff
@@ -0,0 +1,570 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = x.view(-1, self.dim)
+        # T = num_tokens, E = num_experts, D = hidden dim, A = activated experts
+        # x: [T, D]
+        scores = self.gate(x)  # [T, E]
+        expert_weights, expert_indices = torch.topk(scores, 2, dim=-1)  # [T, A], [T, A]
+        expert_weights = expert_weights.softmax(dim=-1)  # [T, A]
+        expert_outs = self.cond_ffn(x, expert_indices)
+        return torch.einsum("tai,ta -> ti", expert_outs, expert_weights)
```
kimishpatel, line 245: not relevant

examples/apple/coreml/llama/export.py
kimishpatel: How did you find that splitting linear actually improves performance?

Issue #8393, opened 2025-02-12
FlatTensor: support >1 segment and freeing buffers


lucylq:
1. FlatTensor only supports 1 segment at the moment.
2. The NamedDataMap implementation persists the segment; this means it cannot be freed. This change makes method hold the tensors after receiving them from the NamedDataMap.

PR #8437, opened 2025-02-13
[flat_tensor] Persist FreeableBuffers of external constants in method

lucylq:
* __->__ #8437

## Problem
Currently, the FlatTensorDataMap persists tensors, and returns a FreeableBuffer with an empty free function.

The NamedDataMap should not persist data, as most cases (eg. delegate) will want it to be freed.

Ownership should be on the caller; `get_data` returns a FreeableBuffer that 'owns' the data. The FreeableBuffer in turn is owned by the caller.

NOTE: this doesn't support the case where we want to share plain tensors between methods/pte files at runtime. A custom NDM could support that use-case.

## This diff:
1. Introduces a 'NamedData' struct to method.h. This holds a key and a FreeeableBuffer.
2. Iterate over all the flatbuffer tensors to count the constants tagged with EXTERNAL.
NOTE: this will increase load time for all users. Potentially allocate chunks of 16 and use a linked list to store external constants, or store this number in PTE file.
3. Allocate space for num_external_constants using the method allocator.
4. Iterate over all flatbuffer tensors and use the named_data_map to resolve EXTERNAL tensors into the array of NamedData.
5. Pass the resolved external constants to tensor_parser, along with NDM (used for mutable external tensors).
6. Resolved external tensors are stored inside method. They are freed when the method is destructed.

Some notes:
https://docs.google.com/document/d/1_PBi4JgODuClUPD4PCUWrKNjyUH54zOUHGUJ3QHDNes/edit?tab=t.0#heading=h.blsvwraxss7g

Differential Revision: [D69477027](https://our.internmc.facebook.com/intern/diff/D69477027/)


lucylq:
For: https://github.com/pytorch/executorch/issues/8393

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -146,76 +185,43 @@ ET_NODISCARD Result<void*> getTensorDataPtr(
...
-          i);
-      ET_CHECK_OR_RETURN_ERROR(
-          s_tensor->dim_order()->Get(i) == tensor_layout.dim_order()[i],
-          InvalidExternalData,
-          "Dim order mismatch. Expected %d, got %d for dim at index %d.",
-          s_tensor->dim_order()->Get(i),
-          tensor_layout.dim_order()[i],
-          i);
-    }
+    const char* key =
```
dbort, line 188: Since this has a fairly long scope, `fqn` or `fully_qualified_name` would make it more self-documenting. It happens to be used as a key, but an FQN is what it _is_.

runtime/executor/method.cpp
```diff
@@ -299,23 +411,33 @@ Error Method::parse_values(const NamedDataMap* named_data_map) {
...
-    // Ensure that the `val_as_X()` calls will return non-null pointers.
-    ET_CHECK_OR_RETURN_ERROR(
-        serialization_value != nullptr &&
-            (serialization_value->val_type() ==
-                 executorch_flatbuffer::KernelTypes::Null ||
-             serialization_value->val() != nullptr),
-        InvalidProgram,
-        "Null value at index %" ET_PRIsize_t,
-        i);
-
```
dbort, line 318: Is this handled somewhere else now? This was a fix for crashes when loading corrupt PTEs

dbort (changes requested):
The overall shape of this looks good. The main issues right now:
- ~Method can look at uninitialized NamedData entries
- Some new types/functions shouldn't be part of the public API

Also, are there tests for this code path already, or are you planning to add some later?

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -11,6 +11,7 @@
 #include <executorch/runtime/core/evalue.h>
 #include <executorch/runtime/core/exec_aten/exec_aten.h>
 #include <executorch/runtime/executor/memory_manager.h>
+#include <executorch/runtime/executor/method.h>
```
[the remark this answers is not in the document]
lucylq: thanks for the clarification here!

runtime/executor/method.cpp
```diff
@@ -289,6 +289,118 @@ Result<bool> parse_cond_value(const EValue& cond_value) {
...
+        ET_CHECK_OR_RETURN_ERROR(
+            buffer.ok(),
+            InvalidExternalData,
+            "Buffer retrieved from get_data is not valid");
+        external_constants_[index].buffer =
+            memory_manager_->method_allocator()
+                ->allocateInstance<FreeableBuffer>();
+        if (external_constants_[index].buffer == nullptr) {
+          return Error::MemoryAllocationFailed;
+        }
```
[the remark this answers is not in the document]
lucylq: Thanks!

runtime/executor/method.h
```diff
@@ -43,6 +43,17 @@ using OpFunction = void (*)(KernelRuntimeContext&, EValue**);
 /// argument list for a single instruction
 using InstructionArgs = Span<EValue*>;
 
+/// Data structure to hold key and data buffer for external data used
+/// in a method.
+struct NamedData {
+  const char* key;
+  FreeableBuffer* buffer;
+};
```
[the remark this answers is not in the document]
lucylq: thanks for the suggestions! Updated by forward-declaring it inside method.h and keeping the definition in tensor_parser.h

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -146,76 +185,43 @@ ET_NODISCARD Result<void*> getTensorDataPtr(
...
-          s_tensor->extra_tensor_info()->fully_qualified_name()->c_str());
-      if (!data_res.ok()) {
-        return data_res.error();
+      for (int i = 0; i < external_constants.size(); i++) {
+        if (strcmp(external_constants[i].key, key) == 0) {
+          // The const_cast is 'ok' here because program and runtime should
+          // guarantee that this data is never modified.
+          return const_cast<void*>(external_constants[i].buffer->data());
+        }
       }
```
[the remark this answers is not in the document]
lucylq: this makes sense, thanks!

runtime/executor/method.cpp
lucylq: I added it to the first iteration through the flatbuffer values (in get_num_external_constants)

runtime/executor/method.cpp
```diff
@@ -289,6 +290,109 @@ Result<bool> parse_cond_value(const EValue& cond_value) {
...
 } // namespace
 
+Result<size_t> Method::get_num_external_constants() {
+  auto flatbuffer_values = serialization_plan_->values();
+  size_t n_value = flatbuffer_values->size();
+
+  size_t n_external_constants = 0;
+  for (size_t i = 0; i < n_value; ++i) {
+    auto serialization_value = flatbuffer_values->Get(i);
+    // Ensure that the `val_as_X()` calls will return non-null pointers.
```
lucylq, line 300: @dbort I moved the check here, as this is now the first iteration over the flatbuffer values.

JacobSzwejbka (approved):

runtime/executor/tensor_parser.h
```diff
@@ -21,18 +21,34 @@ namespace executorch {
...
 namespace deserialization {
 
+/// Data structure to hold key and data buffer for external data used
+/// in a method.
+struct NamedData {
+  const char* key;
+  FreeableBuffer buffer;
+};
+
+NamedData* get_data_by_key(const char* key, Span<NamedData> external_constants);
```
dbort, line 31: This function searches a generic array of NamedData, and could be used in situations other than looking for external constants: a generic param name like `entries` would make that more clear.

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -111,12 +111,62 @@ ET_NODISCARD Result<BoxedEvalueList<executorch::aten::Tensor>> parseTensorList(
       evalp_list, tensor_list, tensor_indices->size());
 }
 
+ET_NODISCARD Error validateExternalTensor(
```
dbort, line 114: Nothing about this is specific to external tensors per se; `validateTensorLayout` (and renaming the param to something generic like `expected_layout`) would make that more clear.

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -111,12 +111,62 @@ ET_NODISCARD Result<BoxedEvalueList<executorch::aten::Tensor>> parseTensorList(
...
+    const executorch_flatbuffer::Tensor* s_tensor,
+    const TensorLayout& external_tensor) {
+  // Compatibility checking.
```
dbort, line 117: This is more of a comment on the entire function

runtime/executor/tensor_parser_exec_aten.cpp
```diff
@@ -146,76 +196,41 @@ ET_NODISCARD Result<void*> getTensorDataPtr(
...
-          s_tensor->extra_tensor_info()->fully_qualified_name()->c_str());
-      if (!data_res.ok()) {
-        return data_res.error();
+      for (int i = 0; i < external_constants.size(); i++) {
+        if (strcmp(external_constants[i].key, fqn) == 0) {
+          // The const_cast is 'ok' here because program and runtime should
+          // guarantee that this data is never modified.
+          return const_cast<void*>(external_constants[i].buffer.data());
+        }
+      }
```
dbort, line 210: Could use `get_data_by_key(fqn)->buffer.data()` here (though with a check that get_data_by_key returns non-null)

dbort (changes requested):
Thanks for the changes, looking good. The biggest remaining issue is the buffer leak when parse_values fails.

runtime/executor/method.cpp
```diff
@@ -299,23 +403,34 @@ Error Method::parse_values(const NamedDataMap* named_data_map) {
...
+    external_constants_ =
+        memory_manager_->method_allocator()->allocateList<NamedData>(
+            max_external_constants.get());
+    if (external_constants_ == nullptr) {
+      return Error::MemoryAllocationFailed;
+    }
+    Result<size_t> n_resolved_constants =
+        parse_external_constants(named_data_map);
+    if (!n_resolved_constants.ok()) {
+      return n_resolved_constants.error();
```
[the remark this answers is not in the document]
lucylq: Thanks for pointing this out! Sorry, I should have checked the value-loading loop more carefully.

runtime/executor/tensor_parser_exec_aten.cpp
lucylq: thanks!

runtime/executor/method.cpp
```diff
@@ -289,6 +290,113 @@ Result<bool> parse_cond_value(const EValue& cond_value) {
...
+    // Save the buffer.
+    Result<FreeableBuffer> buffer = named_data_map->get_data(key);
+    ET_CHECK_OR_RETURN_ERROR(
+        buffer.ok(),
+        InvalidExternalData,
+        "Buffer retrieved from get_data is not valid");
+    new (&external_constants_[n_external_constants_].buffer)
+        FreeableBuffer(std::move(buffer.get()));
+
+    n_external_constants_ = i + 1;
```
dbort, line 395: Setting this to `i + 1` will cause it to skip ahead after encountering non-Tensor entries, in most cases causing an overrun of the external_constants_ array. It should only advance when adding an entry.

```suggestion
    n_external_constants_ += 1;
```

runtime/executor/tensor_parser.h
```diff
@@ -21,18 +21,34 @@ namespace executorch {
...
+
+NamedData* get_data_by_key(const char* key, Span<NamedData> entries);
+
 ET_NODISCARD Result<executorch::aten::Tensor> parseTensor(
     const Program* program,
     MemoryManager* memory_manager,
     const executorch_flatbuffer::Tensor* s_tensor,
-    const NamedDataMap* named_data_map = nullptr);
+    const NamedDataMap* named_data_map = nullptr,
+    Span<NamedData> entries = {});
```
dbort, line 38: In this case `external_constants` (or whatever) was appropriate, because that's what they are. It's only get_data_by_key that is fully generic.

dbort (changes requested):
One last bug, but looks good overall. ASAN would have caught this if we had tests that exercised the new paths.

runtime/executor/method.cpp
lucylq: ah, oops. Thank you!

dbort (approved):

PR #8438, opened 2025-02-13
Guard against null flatbuffer containers when getting constant data

dbort:
Summary:
Missing flatbuffer containers will be `null`, so it's not safe to blindly call `->size()` or similar on them.

Check for missing lists when looking up constant buffer data.

Differential Revision: D69562793


lucylq (approved):
Thanks for adding these checks!

PR #8441, opened 2025-02-13
Fix dtype casting in Llama model

jackzhxng:
### Summary
Dtype casting should be done outside of model.py, which should be limited to instantiating and load the weights of the Llama model.

This solves precision loss in situations such as the following (just one example):
- Dtype override is fp32
- Model instantiated with non weight attributes (such as rope cache) in fp32
- Checkpoint in bf16 loaded
- model.py casts model to bf16, causing the rope cache to lose precision
- export_llama casts model back to dtype override of fp32

### Test plan
Rely on existing CI


examples/models/llama/export_llama_lib.py
```diff
@@ -562,12 +562,10 @@ def _prepare_for_llama_export(args) -> LLMEdgeManager:
     weight_type = WeightType.FAIRSEQ2 if args.fairseq2 else WeightType.LLAMA
 
     # dtype override
-    if args.dtype_override is not None:
```
jackzhxng, line 565: Removing since args.dtype_override has a default of fp32

examples/models/llama/export_llama_lib.py
```diff
@@ -971,32 +969,16 @@ def _load_llama_model(
...
     )
-    if dtype_override:
-        assert isinstance(
-            dtype_override, DType
-        ), "Override dtype needs to be of type <DType>"
-        torch_dtype = dtype_override.to_torch_dtype()
-        logging.info(f"model.to {torch_dtype}")
-        model = model.to(dtype=torch_dtype)
-        dtype = dtype_override
-    else:
```
jackzhxng, line 982: Also removing since dtype_override is never None

PR #8443, opened 2025-02-13
Fix TrainingModule Parameter Bug

JacobSzwejbka:
Summary: The params were being cloned and then cached so in python we observed them changing but the actual cpp weights being run against were not.

Differential Revision: D69568035

### DIFF
--- a/extension/training/pybindings/_training_module.py
+++ b/extension/training/pybindings/_training_module.py
@@ -41,20 +41,20 @@ class TrainingModule:
             self.parameters_method_prefix + method_name, ()
         )[0]
 
-        full_outputs = self.model.run_method(method_name, inputs)
-
-        user_outs = full_outputs[:grad_start_idx]
-        grads = full_outputs[grad_start_idx:params_start_idx]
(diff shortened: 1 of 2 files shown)


extension/training/pybindings/_training_module.py
```diff
@@ -41,20 +41,20 @@ def forward_backward(self, method_name: str, inputs: Sequence[Any]) -> List[Any]
...
-        # Important that the outputs are not cloned because we need the optimizer to
-        # be able to mutate the actual weights and not clones of them.
-        fqn = self.model.run_method(
-            self.fqn_method_prefix + method_name, (), clone_outputs=False
-        )
+        fqn = self.model.run_method(self.fqn_method_prefix + method_name, ())
 
         self.named_grads = dict(zip(fqn, grads))
         if self.named_params is None:
+            params = full_outputs[params_start_idx:]
```
jackzhxng, line 57: nit: would be more clear if you added the comment about not cloning here instead imo

jackzhxng (approved):

PR #8448, opened 2025-02-13
[BE] Cmake clean up

lucylq:
### Summary
1. CMAKE_BINARY_DIR --> CMAKE_CURRENT_BINARY_DIR in `runtime/executor/test/CMakeLists.txt`. Test artifacts should be placed in the test directory (instead of the root)
2. Generate ModuleAdd artifact for extension_runner_util test (currently relying on the artifact generated by runtime/executor/test/CMakeLists.txt`
3. run `cmake-format -i CMakeLists.txt` on the above CMake files and the flat_tensor CMake files.

### Test plan
ci


larryliu0820 (approved):
Thank you this looks a lot cleaner

PR #8462, opened 2025-02-13
Do not offload transposed convolutions with non-zero output padding

pssrawat:
Summary: As title. Otherwise it currently gives runtime error saying XNNPack does not support output padding.

Differential Revision: D69578428


mcr229 (approved):
Thank you for your contribution!

PR #8472, opened 2025-02-13
Remove FlatTensor from OSSTestConfig.json

lucylq:
### Summary
OSSTestConfig.json is only used to generate cmake files. I generated the flat tensor cmake file using it, and then edited it to create the test pte files. Removing flat tensor from the config so that future cmake generation runs do not overwrite the file. 

### Test plan
ci


kirklandsign (approved):

swolchok (approved):
shipit (I miss macros)

PR #8474, opened 2025-02-13
Automatically put torch in CMAKE_PREFIX_PATH so users/scripts don't have to

swolchok:
Removed all explicit setting of CMAKE_PREFIX_PATH to the torch directory in favor of automatically doing it when we find_package(Torch). Hat tip to @huydhn for showing the way to only find torch once.


swolchok:
* (to be filled)

swolchok:
this one should wait for fully green CI before landing since I'm messing with CI scripts

huydhn (approved):

larryliu0820 (approved):
Thank you for doing this

PR #8482, opened 2025-02-13
[Benchmark] fail test if model artifact does not exist

### LINKED ISSUE #8125: DevX: Improve error reporting for benchmark jobs.
### 🐛 Describe the bug

Image

As shown in an example run, there are two issues that make it difficult to understand the expected behavior and locate the exact error:
  1. Unexpected job success status: Two configurations failed to export, yet the `benchmark-on-device` jobs were still marked as successful (highlighted in the red box). The expected behavior is that `benchmark-on-device` jobs should either not be scheduled or be marked as skipped/canceled if a dependent job, such as export, fails.
  2. In the `export-models` job, the error in "Upload artifacts to S3" step is misleading. It should be marked as skipped or canceled, as the "Run script in the container" step failed to generate the .pte file (highlighted in the yellow box).

cc: @huydhn @cbilgin @digantdesai @kimishpatel 

### Versions

trunk

cc @huydhn @kirklandsign @shoumikhin @mergennachin @byjlw

yangw-dev:
### Summary
Fixes https://github.com/pytorch/executorch/issues/8125

Fails the benchmark-on-device job the job at verification step if artifacts don't exist.
- a export-model job failed 
   -  warning shows at artifact upload step
-  prepare-test-spec job failed to check the existence of the model.zip
-  benchmark-on-device job  failed at `Verily Test Spect` to download the test-spec

### Details
move prepare-test-specs under export-models, add dependency on export-models job.
The prepare-test-specs will fail to generate test-spec if no model exists in s3, this can cause the benchmark-on-device job fail at step `verily test-spec`

### others
the failure in this pr seems like not related to this pr
see test pr: https://github.com/pytorch/executorch/pull/8484
just added echo, it still fails

### DIFF
--- a/.github/workflows/android-perf.yml
+++ b/.github/workflows/android-perf.yml
@@ -301,11 +301,8 @@ jobs:
           # The model will be exported in the next step to this S3 path
           MODEL_PATH="https://gha-artifacts.s3.amazonaws.com/${{ github.repository }}/${{ github.run_id }}/artifacts/${{ matrix.model }}_${{ matrix.config }}/model.zip"
 
-          # Check if the model artifact is successfully uploaded from previous step, failed to upload the test-spec if it does not exist.
-          if ! curl -s --head -f ${MODEL_PATH}; then
-            echo "failed to retrieve ${MODEL_PATH}"
-            exit 1;
-          fi
+          # Check if the model artifact exists, fail this step skip generating test-spec.
+          curl -s --head -f ${MODEL_PATH}
 
           # We could write a script to properly use jinja here, but there is only one variable,
           # so let's just sed it
--- a/.github/workflows/android-perf.yml
+++ b/.github/workflows/android-perf.yml
@@ -96,63 +96,6 @@ jobs:
 
           PYTHONPATH="${PWD}" python .ci/scripts/gather_benchmark_configs.py $ARGS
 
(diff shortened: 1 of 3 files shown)


.github/workflows/android-perf.yml
```diff
@@ -335,6 +278,72 @@ jobs:
...
+          BENCHMARK_CONFIG: ${{ toJSON(matrix) }}
+        working-directory: extension/benchmark/android/benchmark
+        run: |
+          set -eux
+
+          # The model will be exported in the next step to this S3 path
+          MODEL_PATH="https://gha-artifacts.s3.amazonaws.com/${{ github.repository }}/${{ github.run_id }}/artifacts/${{ matrix.model }}_${{ matrix.config }}/model.zip"
+
+          # Check if the model artifact is successfully uploaded from previous step, failed to upload the test-spec if it does not exist.
+          if ! curl -s --head -f ${MODEL_PATH}; then
```
yangw-dev, line 305: [optional]add if statement to provide more info, could just ` curl -s -I ` since we `set -eux `

huydhn (approved):
LGTM! As the build is broken atm with a fix https://github.com/pytorch/executorch/pull/8474#pullrequestreview-2616476490 coming.  So, let's wait for fix to land and do a rebase

PR #8447, opened 2025-02-13
[executorch][flat_tensor] implement load into and dont hold onto the segment

lucylq:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #8447
* #8437

1. Implement load_into in FlatTensorDataMap
2. Do not persist 'data_ro' in the FlatTensorDataMap. From `get_data`, return the FreeableBuffer given by the data loader.

TODO: add test for load_into.

Differential Revision: [D69148652](https://our.internmc.facebook.com/intern/diff/D69148652/)

PR #8483, opened 2025-02-13
[executorch][flat tensor] Store number of external tensors in flatbuffer

lucylq:
* __->__ #8483
* #8447
* #8437

Store the number of external tensors in each ExecutionPlan.

This saves time during loading, as we can directly allocate memory for external tensors without doing a pass to count how many external tensors require allocation.

Differential Revision: [D69618283](https://our.internmc.facebook.com/intern/diff/D69618283/)

### DIFF
--- a/exir/emit/_emit_program.py
+++ b/exir/emit/_emit_program.py
@@ -163,6 +163,7 @@ def emit_program(
             operator_cache={},
             delegate_cache={},
             emit_stacktrace=emit_stacktrace,
+            num_external_constants=0,
         )
 
         gm = _remove_non_user_outputs(exported_program)
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -143,6 +143,7 @@ class _EmitterState:
     """
 
     values: List[EValue]
+    num_external_constants: int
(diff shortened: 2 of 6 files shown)


JacobSzwejbka:
So this is an FC breaking change but thats fine since no one is using external constants yet.

Otherwise adding things to the schema is a pretty high bar typically. I think this change overall is fine since the overhead is small, but I'd still prefer we icebox this change until its clear the overhead of the scan requires this. Since we only ever have to perform the scan if a .ptd was provided anyway.

larryliu0820 (approved):

larryliu0820:
> So this is an FC breaking change but thats fine since no one is using external constants yet.
> 
> Otherwise adding things to the schema is a pretty high bar typically. I think this change overall is fine since the overhead is small, but I'd still prefer we icebox this change until its clear the overhead of the scan requires this. Since we only ever have to perform the scan if a .ptd was provided anyway.

Oh just saw this - I would love to make things right before people start to use it. 

> until its clear the overhead of the scan requires this

Even if the scan is trivial, it is something not necessary at runtime and we probably want the runtime logic to be as simple as possible.

Issue #7585, opened 2025-01-10
Export error in models from huggingface, related to device("meta")


DongGeun123:
### 🐛 Describe the bug

Here’s the English version of your content:

I am currently trying to export Hugging Face models in a way similar to what was done in executorch. Following the example for LLaMA, in executorch/examples/models/llama/model.py, the model is loaded on a meta device as follows:

```
with torch.device("meta"):
    self.model_ = Transformer(model_args)
```

When I replace the model with another Hugging Face model, I encounter an error during torch.export.export or export_for_training:

```
dtype = torch.get_autocast_dtype(device_type)
torch._dynamo.exc.InternalTorchDynamoError: RuntimeError: unsupported scalarType
```

I suspect this error occurs when exporting a model initialized on a meta device due to an issue with torch.amp.autocast_mode in the model’s forward method. However, if I skip the meta device initialization and directly export a Hugging Face model, the export works fine. For example:
```

import torch
from transformers import AutoModelForCausalLM, AutoTokenizer

llama = AutoModelForCausalLM.from_pretrained(
    "meta-llama/Llama-3.2-1B-Instruct", low_cpu_mem_usage=True
)
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-1B-Instruct")

llama.eval()
inputs = tokenizer("How do you", return_tensors="pt")

exported_llama = torch.export.export(llama, args=(inputs["input_ids"], inputs["attention_mask"]))
```

My question is: why does the executorch example define the model on a meta device before exporting it?

### Versions

[pip3] executorch==0.5.0a0+e89e320
[pip3] mypy-extensions==1.0.0
[pip3] numpy==1.21.3
[pip3] torch==2.6.0.dev20241112+cpu
[pip3] torchao==0.7.0+git75d06933
[pip3] torchaudio==2.5.0.dev20241112+cpu
[pip3] torchsr==1.0.4
[pip3] torchvision==0.20.0.dev20241112+cpu
[conda] executorch                0.5.0a0+e89e320          pypi_0    pypi
[conda] numpy                     1.21.3                   pypi_0    pypi
[conda] torch                     2.6.0.dev20241112+cpu          pypi_0    pypi
[conda] torchao                   0.7.0+git75d06933          pypi_0    pypi
[conda] torchaudio                2.5.0.dev20241112+cpu          pypi_0    pypi
[conda] torchsr                   1.0.4                    pypi_0    pypi
[...]

PR #8487, opened 2025-02-13
Eliminate confusion about meta weights during quantization

jackzhxng:
### Summary
Meta device errors (https://github.com/pytorch/executorch/issues/7585) during quantization occur when weights aren't loaded on meta tensors, since quantization requires real values.

Why is this error not present in CI? Because we are using Llama models with properly formatted checkpoints, each weight in the module's state dict is populated with the loaded checkpoint.

Quantizing non-Llama OSS models recently have run into this issue since their checkpoints are formatted differently, and when we load the checkpoint into the model, some or all of the checkpoint keys will not match the state dict fqn's, causing the state dict to remain partially/fully uninitialized with meta tensors.

Here we check that none of the missing items returned form load_state_dict are weights (some missing items could be non-weight buffers such as kv cache, which is ok). We don't use strict=False because it should be okay for the checkpoint to contain unnecessary items as well.

### Test plan
Export and quantize an OSS model that was previously complaining about meta device errors, now it intentionally raises an informative error letting the user know to format the weights properly.


examples/models/llama/model.py
```diff
@@ -244,13 +244,16 @@ def __init__(self, **kwargs):
...
-            print("============= unexpected keys ================")
-            print(unexpected)
-            print("============= /unexpected ================")
+
+        if missing:
+            missing_weights = [fqn for fqn in missing if fqn.endswith(".weight")]
+            if missing_weights:
+                raise ValueError(
+                    f"The provided checkpoint is missing the following weights that are expected by the model: {missing_weights}. Please fix the fqn's in your checkpoint to match."
+                )
```
lucylq, line 253: Keep the print for missing fqns that aren't weights?

lucylq (approved):
Thanks for debugging this and adding the clear error message!

examples/models/llama/model.py
jackzhxng: I think no point to list out the buffers as missing weights, some people (like me) might get confused what it means at first

PR #8494, opened 2025-02-14
patch qnn silu lowering

cccclai:
Summary: There are some edge lowering case missing. Change the logic to look for the op first because it's not decomposed

Differential Revision: D69636086

### DIFF
--- a/backends/qualcomm/_passes/decompose_silu.py
+++ b/backends/qualcomm/_passes/decompose_silu.py
@@ -7,7 +7,6 @@ from typing import Dict
 
 import torch
 from executorch.exir.pass_base import ExportPass, PassResult
-from torch.fx.passes.utils.source_matcher_utils import get_source_partitions
 
 
 class DecomposeSilu(ExportPass):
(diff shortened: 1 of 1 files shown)


backends/qualcomm/_passes/decompose_silu.py
```diff
@@ -22,24 +22,22 @@ def _copy_meta(self, meta: Dict):
...
-            graph, [torch.nn.functional.silu, torch.ops.aten.silu.default]
-        )
-        for _, src_partitions in partitions.items():
-            for src_partition in src_partitions:
-
-                inputs = src_partition.input_nodes
-                silu_node = src_partition.nodes[0]
-                with graph_module.graph.inserting_after(inputs[0]):
+        for node in graph.nodes:
+            if node.op == "call_function" and node.target in [
```
haowhsu-quic, line 26: Maybe we can use set here `{`.

haowhsu-quic (approved):
Thank you!

backends/qualcomm/_passes/decompose_silu.py
cccclai: oh actually it's just one op and we can just compare...

winskuo-quic (approved):
LGTM. Thanks.
Seems like there are lint issues.

kirklandsign (approved):

PR #8496, opened 2025-02-14
[devtool] introduce datasink class to etdump

Gasoonjia:
* #8604
* #8514
* #8499
* __->__ #8496

this diff introduce datasink class, the class for managing the customized debug data storage pipeline.

Detials can be found in https://docs.google.com/document/d/1y_m32mKdj-OgLcLUz9TKhBW3PC3bBDYSBbeAH544EfM/edit?tab=t.0

Differential Revision: [D69583422](https://our.internmc.facebook.com/intern/diff/D69583422/)


devtools/etdump/data_sink.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <executorch/devtools/etdump/data_sink_base.h>
+#include <executorch/runtime/core/exec_aten/exec_aten.h>
+#include <executorch/runtime/core/span.h>
+
+namespace executorch {
+namespace etdump {
+
+class DataSink : public DataSinkBase {
```
dbort, line 12: Please add doc comments for this class.

dbort, line 12: Since there will be multiple subclasses of DataSinkBase, it's important to pick names that differentiate them: "DataSink" is too generic. This could be "BufferDataSink" since it writes into a buffer.

devtools/etdump/data_sink_base.h
```diff
@@ -0,0 +1,49 @@
...
+   *         debug storage, which will be recorded in corresponding tensor
+   * metadata of ETDump.
+   */
+  virtual size_t write_tensor(const executorch::aten::Tensor& tensor) = 0;
+  /**
+   * Get the maximum capacity of the debug storage in bytes.
+   *
+   * @return The total size of the debug storage.
+   */
+  virtual size_t get_storage_size() const = 0;
```
dbort, line 39: Please expand the comment: what should subclasses return when the sink has infinite capacity? E.g. when writing to a file, pipe, UART, network socket etc?

devtools/etdump/data_sink_base.h
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Write tensor data into the debug storage. This method should be implemented
+   * by derived classes to handle the specifics of data storage.
+   *
+   * @param[in] tensor The tensor data to be written into the storage.
+   * @return The offset of the starting location of the tensor data within the
+   *         debug storage, which will be recorded in corresponding tensor
+   * metadata of ETDump.
+   */
+  virtual size_t write_tensor(const executorch::aten::Tensor& tensor) = 0;
```
dbort, line 33: The class comment talks about writing debug data, but the interface only allows writing bare tensors. Are you planning to add methods to write other types of data?

And for this method, it seems like there should be some kind of metadata attached to the tensor so that consumers of this data know how to interpret it. E.g. if I create a DataSink that writes a bunch of tensors as JSON to a file, how will my JSON-parsing tools know what do do with tensor[1] vs tensor[15]?

... reading this more, I see that this returns an offset that's referred to by another set of data. In that case, this class isn't a generic "data sink" (which I would expect to have a way to write general records), it's specifically Tensor-related, like a "TensorRecord" or "TensorDataSink". And the class comment should make it clear that the data written here needs to be paired with a separate data stream that records the locations of tensors provided to this class.

But this also raises questions for me about the format of the data that's written, and how the data is parsed later. The current API requires that each subclass implements its own serialization method for the Tensor. When reading back data that's written by a subclass, a parser would need to know which subclass wrote the data so it knew how to parse it.

If this class is meant to purely represent a data destination abstraction, then it seems like some higher layer should handle the (standard) serialization, and then pass a simple `Span<uint8_t>` to the subclass for writing. Then we have a guarantee that all DataSink-written files/buffers contain the same data format, which can be parsed by a common tool.

... reading this even more :) it looks like this specifically only writes a blob of tensor data, which isn't clear from the API, and could be misinterpreted by subclasses. Given that, please change this method to take a `Span<uint8_t>`, and have the caller say
```
ds->write_tensor({tensor.const_data_ptr(), tensor.nbytes()});
```
Then, this class doesn't need to depend on Tensor, and becomes a pure data sink. At that point, this method isn't even `write_tensor()`, it's just `write()`. And since the caller knows how many bytes it wrote, you could ignore my earlier comment about returning the number of bytes written.

devtools/etdump/data_sink_base.h
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Virtual destructor to ensure proper cleanup of derived classes.
+   */
+  virtual ~DataSinkBase() = default;
+  /**
+   * Write tensor data into the debug storage. This method should be implemented
+   * by derived classes to handle the specifics of data storage.
+   *
+   * @param[in] tensor The tensor data to be written into the storage.
+   * @return The offset of the starting location of the tensor data within the
```
dbort, line 29: Returning only the offset isn't enough to be safe: it's be better to also return the length, so that readers of this data know how much they're allowed to read at that offset.

devtools/etdump/targets.bzl
```diff
@@ -103,9 +149,11 @@ def define_common_targets():
             ],
             deps = [
                 "//executorch/runtime/platform:platform",
+                ":utils",
             ],
             exported_deps = [
                 ":etdump_schema_flatcc",
+                ":utils",
```
dbort, line 156: Doesn't need to be in both deps and exported_deps; since it's internal-only, it'd be better to stay in non-exported deps if possible

devtools/etdump/data_sink.cpp
```diff
@@ -0,0 +1,32 @@
...
+size_t DataSink::write_tensor(const executorch::aten::Tensor& tensor) {
+  if (tensor.nbytes() == 0) {
+    return static_cast<size_t>(-1);
+  }
+  uint8_t* offset_ptr =
+      internal::alignPointer(debug_buffer_.data() + offset_, 64);
+  offset_ = (offset_ptr - debug_buffer_.data()) + tensor.nbytes();
+  ET_CHECK_MSG(
+      offset_ <= debug_buffer_.size(),
+      "Ran out of space to store tensor data.");
```
dbort, line 18: Please do not use ET_CHECK/ET_CHECK_MSG, which kills the entire process when it fails. This method should return `Result<size_t>` so that it can fail non-fatally when there's a problem.

devtools/etdump/data_sink.cpp
```diff
@@ -0,0 +1,32 @@
...
+namespace executorch {
+namespace etdump {
+
+size_t DataSink::write_tensor(const executorch::aten::Tensor& tensor) {
+  if (tensor.nbytes() == 0) {
+    return static_cast<size_t>(-1);
+  }
+  uint8_t* offset_ptr =
+      internal::alignPointer(debug_buffer_.data() + offset_, 64);
+  offset_ = (offset_ptr - debug_buffer_.data()) + tensor.nbytes();
```
dbort, line 15: When bumping the offset forward, I strongly suggest zeroing out the memory in the padding. If you don't do that, and this buffer is written to a file, then that file may contain random data from the process.

a) this makes the file contents non-deterministic, and b) it could inadvertently write user data or secrets into the file if they had been stored in the heap and then freed before being used for this buffer.

devtools/etdump/data_sink.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <executorch/devtools/etdump/data_sink.h>
+#include <executorch/devtools/etdump/utils.h>
+
+namespace executorch {
+namespace etdump {
+
+size_t DataSink::write_tensor(const executorch::aten::Tensor& tensor) {
+  if (tensor.nbytes() == 0) {
+    return static_cast<size_t>(-1);
```
dbort, line 11: size_t is unsigned, so `-1` will look like a very large positive number: 18 exabytes on a 64-bit machine. If this is part of the contract, please document it in the base class. But I'd recommend just returning `offset_` in this case since it's a zero-sized buffer and the offset doesn't matter.

devtools/etdump/data_sink.h
```diff
@@ -0,0 +1,28 @@
...
+ public:
+  explicit DataSink(::executorch::runtime::Span<uint8_t> buffer)
+      : debug_buffer_(buffer), offset_(0) {}
+
+  size_t write_tensor(const executorch::aten::Tensor& tensor) override;
+  size_t get_storage_size() const override;
+  size_t get_used_bytes() const override;
+
+ private:
```
dbort, line 21: Please make sure that this class isn't copyable: right now it might be because its fields are trivially-copyable. But if someone creates two copies and writes to them, their offsets will be out of sync, violating the expected behavior by potentially corrupting the underlying data.

dbort (changes requested):
I think that the API should change to take a simple `Span<uint8_t>` instead of a Tensor; see my comments for details.

devtools/etdump/data_sink_base.h
Gasoonjia: Thanks for your detailed suggestion! Impl a `write` function instead of current `write_tensor` would be a better choice.

devtools/etdump/data_sink.cpp
Gasoonjia: let me create another diff for solving that; this function copy&paste from copy_tensor_to_debug_buffer function in ETDumpGen; updating the returning data here may introduce BP issue

devtools/etdump/targets.bzl
Gasoonjia: Removed `:utils` in deps; there're will be some internal issue in `arvr` if only exist in non-exported deps

devtools/etdump/data_sink_base.h
Gasoonjia: i think that should be fine cuz user knows how many data they want to be written in the buffer.

If the write succeeds, the number of bytes should be allowed to read at that offset is just equal to the number of bytes want to write;
If not, the function will return an Error.

tarun292:
Looks great to me overall and i think most of @dbort's comments are addressed now. Will let him take a final look and stamp.

devtools/etdump/buffer_data_sink.cpp
```diff
@@ -0,0 +1,49 @@
...
+  // Zero out the padding between data blobs.
+  size_t n_zero_pad = offset_ptr - debug_buffer_.data() - offset_;
+  memset(debug_buffer_.data() + offset_, 0, n_zero_pad);
+
+  offset_ = (offset_ptr - debug_buffer_.data()) + length;
+
+  // Raise access error if offset_ is out of range.
+  if (offset_ > debug_buffer_.size()) {
+    return Error::OutOfResources;
+  }
```
dbort, line 34: This check should happen before mutating the class in any way (zeroing padding, updating offset_). I.e., a failure case should leave the state untouched.

E.g., someone might say `write(ptr, 1MB)` which fails, then `write(ptr 1kB)` which succeeds. The offset shouldn't move, and the buffer shouldn't be modified, when the first call fails.

devtools/etdump/buffer_data_sink.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return static_cast<size_t>(-1);
+  }
+  uint8_t* offset_ptr =
+      internal::align_pointer(debug_buffer_.data() + offset_, 64);
+
+  // Zero out the padding between data blobs.
+  size_t n_zero_pad = offset_ptr - debug_buffer_.data() - offset_;
+  memset(debug_buffer_.data() + offset_, 0, n_zero_pad);
+
+  offset_ = (offset_ptr - debug_buffer_.data()) + length;
```
dbort, line 29: This seems correct but it's tough to follow. I suggest using temp pointers to simplify a lot of the arithmetic, like
```
uint8_t* base = debug_buffer_.data() + offset_;
uint8_t* new_base = internal::align_pointer(base, 64);
size_t padding = new_base - base;
memset(base, 0, padding);
offset_ += padding;
```

devtools/etdump/utils.h
```diff
@@ -0,0 +1,24 @@
...
+#include <cstddef>
+#include <cstdint>
+
+#pragma once
+
+namespace executorch {
+namespace etdump {
+namespace internal {
+
+inline uint8_t* align_pointer(void* ptr, size_t alignment) {
```
dbort, line 12: Please add a doc comment, and point out that this assumes that `alignment` is a power of 2.

devtools/etdump/data_sink_base.h
```diff
@@ -0,0 +1,70 @@
...
+   *           metadata of ETDump, or
+   *         - An error code indicating the failure reason, if any issue
+   *           occurs during the write process.
+   */
+  virtual ::executorch::runtime::Result<size_t> write(
+      const void* ptr,
+      size_t length) = 0;
+  /**
+   * Get the maximum capacity of the debug storage in bytes.
+   * Should return Error::NotSupported if the capacity is not available
```
dbort, line 52: If this might return NotSupported, then users of the base class can't depend on it. Ideally, base classes should only contain methods that all subclasses will be able to implement.

I recommend removing this method.

dbort (changes requested):
Thanks for renaming the subclass

devtools/etdump/buffer_data_sink.cpp
```diff
@@ -0,0 +1,49 @@
...
+using ::executorch::runtime::Error;
+using ::executorch::runtime::Result;
+
+namespace executorch {
+namespace etdump {
+
+Result<size_t> BufferDataSink::write(const void* ptr, size_t length) {
+  if (length == 0) {
+    return static_cast<size_t>(-1);
+  }
```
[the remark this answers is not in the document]
dbort: But this is a new function that no-one uses yet. If `ETDumpGen::copy_tensor_to_debug_buffer()` calls this, then you can isolate the old behavior there, like
```
size_t ETDumpGen::copy_tensor_to_debug_buffer(executorch::aten::Tensor tensor) {
  // (comment explaining this special case)
  if (tensor.nbytes() == 0) {
    return static_cast<size_t>(-1);
  }
  return data_sink_->write(tensor.const_data_ptr(), tensor.nbytes());
}
```
This way, the new method can behave properly from the start, and other users/subclasses of DataSinkBase don't need to duplicate the old behavior.

devtools/etdump/buffer_data_sink.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  // The beginning of the next data blob must be aligned to the alignment
+  uint8_t* cur_data_begin = internal::align_pointer(last_data_end, alighment_);
+  uint8_t* cur_data_end = cur_data_begin + length;
+
+  // Raise OutOfResources error if the end of current data blob is out of range.
+  if (cur_data_end > debug_buffer_.data() + debug_buffer_.size()) {
+    return Error::OutOfResources;
+  }
+
```
dbort, line 33: Please bring back the memset to zero out the padding, and add a test that checks for zero-value pad bytes

devtools/etdump/buffer_data_sink.h
```diff
@@ -0,0 +1,88 @@
...
+class BufferDataSink : public DataSinkBase {
+ public:
+  /**
+   * Constructs a BufferDataSink with a given buffer.
+   *
+   * @param[in] buffer A Span object representing the buffer where data will be
+   * stored.
+   */
+  explicit BufferDataSink(::executorch::runtime::Span<uint8_t> buffer)
+      : debug_buffer_(buffer), offset_(0), alighment_(64) {}
```
dbort, line 34: Typo
```suggestion
      : debug_buffer_(buffer), offset_(0), alignment_(64) {}
```

dbort, line 34: May as well add an alignment param. And please document that it must be a power of 2
```suggestion
  explicit BufferDataSink(::executorch::runtime::Span<uint8_t> buffer, size_t alignment = 64)
      : debug_buffer_(buffer), offset_(0), alignment_(alignment) {}
```

devtools/etdump/data_sink_base.h
dbort: I only see one use of get_debug_buffer_size today, at https://fburl.com/code/3od7ch5t; and that is a bad use of the API. We should talk to them and see if they can do it in a different way, removing the only usage of this method.

> we need to have a way for retreiving the size of every datasink

This just isn't possible for all datasinks. If the old ETDump interface needs to support it for now, then we can find a way to isolate that knowledge in the ETDump code, instead of requiring all DataSinks to follow the old pattern.

devtools/etdump/utils.h
```diff
@@ -0,0 +1,34 @@
...
+ * Aligns a pointer to the next multiple of `alignment`. `alignment` must be a
+ * power of 2.
+ *
+ * @param[in] ptr Pointer to align.
+ * @param[in] alignment Alignment to align to. Must be a power of 2.
+ *
+ * @returns A pointer aligned to `alignment`.
+ */
+
+inline uint8_t* align_pointer(void* ptr, size_t alignment) {
```
dbort, line 22: - No need to mention the power of 2 detail in the top message since it's on the param line
- Doxygen comments should not have blank lines after them

```suggestion
/**
 * Aligns a pointer to the next multiple of `alignment`.
 *
 * @param[in] ptr Pointer to align.
 * @param[in] alignment Alignment to align to. Must be a power of 2.
 *
 * @returns A pointer aligned to `alignment`.
 */
inline uint8_t* align_pointer(void* ptr, size_t alignment) {
```

dbort (changes requested):

devtools/etdump/buffer_data_sink.cpp
```diff
@@ -0,0 +1,49 @@
...
+using ::executorch::runtime::Error;
+using ::executorch::runtime::Result;
+
+namespace executorch {
+namespace etdump {
+
+Result<size_t> BufferDataSink::write(const void* ptr, size_t length) {
+  if (length == 0) {
+    return (size_t)offset_;
+  }
```
dbort, line 21: No cast required

```suggestion
  if (length == 0) {
    return offset_;
  }
```

devtools/etdump/buffer_data_sink.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+class BufferDataSink : public DataSinkBase {
+ public:
+  /**
+   * Constructs a BufferDataSink with a given buffer.
+   *
+   * @param[in] buffer A Span object representing the buffer where data will be
+   * stored.
+   * @param[in] alignment The alignment requirement for the buffer. Default is
+   * 64.
```
dbort, line 33: "Must be a power of two"

devtools/etdump/buffer_data_sink.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace executorch {
+namespace etdump {
+
+/**
+ * BufferDataSink is a concrete implementation of the DataSinkBase class,
+ * designed to store debug data in a pre-allocated, user-owned buffer. This
+ * class provides methods to write raw data and tensor data into the buffer,
+ * ensuring proper alignment and managing padding as needed. It is the standard
+ * DataSink used by ETDumpGen.
```
dbort, line 23: If this fact changes, the comment will be incorrect. It's generally better to avoid comments that say who uses a particular type/function/etc.

```suggestion
 * ensuring proper alignment and managing padding as needed.
```

dbort (approved):
Thank you for all of these changes! A couple minor requests, but this looks great.

PR #8499, opened 2025-02-14
[devtool] make ETDumpGen use datasink

Gasoonjia:
* __->__ #8499

This diff enables customized debug data pipeline by making ETDumpGen leverage user-provided datasink.

Details can be found in https://docs.google.com/document/d/1y_m32mKdj-OgLcLUz9TKhBW3PC3bBDYSBbeAH544EfM/edit?tab=t.0#heading=h.jlkcrurw482r

Differential Revision: [D69647096](https://our.internmc.facebook.com/intern/diff/D69647096/)


devtools/etdump/buffer_data_sink.h
```diff
@@ -25,14 +25,24 @@ namespace etdump {
...
-   * Constructs a BufferDataSink with a given buffer.
+   * Constructs a BufferDataSink with a given span buffer.
    *
    * @param[in] buffer A Span object representing the buffer where data will be
    * stored.
    */
   explicit BufferDataSink(::executorch::runtime::Span<uint8_t> buffer)
       : debug_buffer_(buffer), offset_(0) {}
 
+  /**
```
tarun292, line 36: Are there internal users who need this?

devtools/etdump/etdump_flatcc.cpp
```diff
@@ -497,27 +499,15 @@ ETDumpResult ETDumpGen::get_etdump_data() {
 }
 
 void ETDumpGen::set_debug_buffer(Span<uint8_t> buffer) {
```
tarun292, line 501: I think we should remove the `set_debug_buffer` api. Is there any reason to keep it around? Makes it a little confusing for users i think to have these 2 around.

devtools/etdump/etdump_flatcc.cpp
```diff
@@ -497,27 +499,15 @@ ETDumpResult ETDumpGen::get_etdump_data() {
...
-  debug_buffer_ = buffer;
+  data_sink_ = std::make_shared<BufferDataSink>(buffer);
```
tarun292, line 502: Curious why `make_shared` here. This is only owned internally by this class right?

devtools/etdump/etdump_flatcc.cpp
Gasoonjia: just for BC. Currently all our users use `set_debug_buffer` to set the memory for debug data, and their work will be impacted if we remove that

devtools/etdump/buffer_data_sink.h
Gasoonjia: not right now but introducing this constructor may make our friends life easier and I'd like to make it as a primary entrance in the future

If our user contructs `BufferDataSink` from Span:

```
void* ptr = malloc(2048)
Span<uint8_t> span= Span(ptr, 2048)
BufferDataSink bds(span)
```
However if from the new constructor:


```
void* ptr = malloc(2048)
BufferDataSink bds(ptr, 2048)
```

More straightforward and sturcure, user doesn't need to take care about span or anything else.

devtools/etdump/etdump_flatcc.cpp
Gasoonjia: just the underlying data_sink_ is a shared_ptr;

Gasoonjia: Our user may still want to use their`data_sink` after `ETDumpGen` lifecycle, so that I make `data_sink_` as `shared_ptr`.

devtools/etdump/etdump_flatcc.cpp
tarun292: Actually we should keep this API around i think. Only users who want to do their custom thing should know about data sinks, otherwise most users should just be able to call `set_debug_buffer`.

devtools/etdump/etdump_flatcc.cpp
```diff
@@ -636,7 +626,28 @@ bool ETDumpGen::is_static_etdump() {
...
+  Result<size_t> ret = data_sink_->get_storage_size();
+  ET_CHECK_MSG(
+      ret.ok(),
+      "Failed to get storage size with error 0x%" PRIx32,
+      static_cast<uint32_t>(ret.error()));
+  return ret.get();
+}
+
+long ETDumpGen::write_tensor_or_raise_error(Tensor tensor) {
+  ET_CHECK_MSG(data_sink_, "Must set data sink before writing data\n");
```
tarun292, line 643: nit : don't need `\n`.

tarun292 (approved):

devtools/etdump/etdump_flatcc.cpp
Gasoonjia: User only need standard data dumping pipeline can use both datasink and `set_debug_buffer`, since we have provided `BufferDataSink` which is exactly the same as debug buffer pipeline. But from API design prospective, I would like to deprecate `set_debug_buffer` api and only focus on data sink for better structure.

devtools/etdump/buffer_data_sink.h
dbort: I think that the Span ctor would work with syntax like
```
void* ptr = malloc(2048);
BufferDataSink bds({ptr, 2048});
```

Gasoonjia: Program syntactically yes, but user need to figure out what's span is, how to construct it, and how to make the logic shorter, which introduces lots of ramp up cost.
Put into user's shoe, they don't need to knwo what span is. A buffer data sink should be directly constructed on the user-provided memory space, and all type transforming should be happened under-the-neath.

PR #8501, opened 2025-02-14
Switch to new ao quant api for 8da4w

### LINKED ISSUE #8422: Migrate export_llama to new ao quantize API

`Int8DynActInt4WeightQuantizer` for `-qmode 8da4w` is no longer being developed by ao and doesn't support bias. Migrate to the new `quantize_` api which can take in `int8_dynamic_activation_int4_weight`.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @mergennachin @iseeyuan @lucylq @helunwencser @tarun292 @kimishpatel @cccclai

jackzhxng:
### Summary
Closes https://github.com/pytorch/executorch/issues/8422

### Test plan


jerryzh168 (approved):
thanks, please make sure the lowering still works as well

jackzhxng:
@mcr229 taking a stab at that here https://github.com/pytorch/executorch/pull/8624

jackzhxng:
^Above has been closed and xnnpack is now using `to_edge_transform_and_lower`, giving CI another crack at this

jackzhxng:
^Above has been closed and xnnpack is now using `to_edge_transform_and_lower`, giving CI another crack at this

mcr229:
looks like this is running into issues with loading the quantized checkpoints because aten._copy is unimplemented? seems sus, let me know if you need help from me for any part of enabling this.

jackzhxng:
Yeah this is weird, our one test for quantized embeddings is erroring, presumably because the embedding quantization happens after the linear quantization which is now using the API

jerryzh168:
please check out https://pytorch.org/ao/stable/serialization.html to see how serialization works in torchao, copy_ a torchao quantized tensor to a normal tensor is indeed not supported, we typically use `load_state_dict(...., assign=True)`

jackzhxng:
Update - should be resolved. This error was happening since the quantize embedding code was loading the model's state dict which contains ao-quantized tensors from the linear transform before which now uses `quantize_`. Need the `assign=True` to do the deserialization process properly

jackzhxng:
Actually I should have imported and run internal tests before merging, if diff train gets blocked I'll forward fix

examples/models/llama/source_transformation/quantize.py
```diff
@@ -119,11 +119,10 @@ def quantize(  # noqa C901
...
         if group_size is None:
             raise Exception("For 8da4w quantization, group size must be specified.")
-        from torchao.quantization.quant_api import Int8DynActInt4WeightQuantizer
 
-        model = Int8DynActInt4WeightQuantizer(
-            precision=torch_dtype, groupsize=group_size
-        ).quantize(model)
+        from torchao.quantization import int8_dynamic_activation_int4_weight, quantize_
+
+        quantize_(model, int8_dynamic_activation_int4_weight(group_size=group_size))
```
jerryzh168, line 125: is torch_dtype not applied here? should it be applied to `model`?

jackzhxng: Yeah I think it would good to ensure the dtype here by applying it to the model in general the model is already in fp32 for when the test passes before this PR and when it fails after this PR

Issue #8489, opened 2025-02-14
Can't run install on Ubuntu


byjlw:
### 🐛 Describe the bug

```
git clone https://github.com/pytorch/executorch.git
cd executorch
python3.11 -m venv .venv
source .venv/bin/activate

git submodule sync
git submodule update --init

./install_executorch.sh

```

Output
```
Using pip 24.0 from /home/warden/source/executorch/.venv/lib/python3.11/site-packages/pip (python 3.11)
Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/nightly/cpu
Processing /home/warden/source/executorch
  Running command Preparing metadata (pyproject.toml)
  /home/warden/source/executorch/.venv/lib/python3.11/site-packages/setuptools/config/pyprojecttoml.py:108: _BetaConfiguration: Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.
    warnings.warn(msg, _BetaConfiguration)
  running dist_info
  creating /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info
  writing /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/PKG-INFO
  writing dependency_links to /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/dependency_links.txt
  writing entry points to /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/entry_points.txt
  writing requirements to /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/requires.txt
  writing top-level names to /tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/top_level.txt
  writing manifest file '/tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/SOURCES.txt'
  reading manifest file '/tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/SOURCES.txt'
  adding license file 'LICENSE'
  writing manifest file '/tmp/pip-modern-metadata-twj7o8yl/executorch.egg-info/SOURCES.txt'
  creating '/tmp/pip-modern-metadata-twj7o8yl/executorch-0.6.0a0+319c8d7.dist-info'
  Preparing metadata (pyproject.toml) ... done
Collecting expecttest (from executorch==0.6.0a0+319c8d7)
[...]

PR #8503, opened 2025-02-14
Add python-dev install to FAQ

larryliu0820:
Summary:
#8489
Adding instructions for users to install python-dev library as a prerequisite of installing ExecuTorch

Differential Revision: D69680232

cc @mergennachin @byjlw


mergennachin (approved):

GregoryComer (approved):
Thanks for adding this. Having an install section of the FAQ is a good idea.

PR #8505, opened 2025-02-14
Add CoreML support for to_edge_transform_and_lower

metascroy:
CoreML supports many ATen ops that decomposed during to_edge (e.g., SDPA, pixel_unshuffle, etc.)

To prevent that users can instead use to_edge_transform_and_lower, but CoreML partitioner did not support that flow.  This PR adds support.

Test plan:
New coreml partitioner test.


shoumikhin (approved):

cccclai (approved):
This is awesome, thank you! cc: @YifanShenSZ

PR #8506, opened 2025-02-14
Add small repro test for unsigned -> signed et loss error

dulinriley:
Summary:
There was a difference in behavior from `quantized_decomposed.quantize_per_tensor`
and `cadence.quantize_per_tensor`, specifically how rounding half values worked.
The former rounds towards even (based on `torch.round` which does that).
The latter rounds away from zero.

Make sure the python implementation matches the Executorch implementation in this
regard.

Differential Revision: D69668881

### DIFF
--- a/backends/cadence/aot/fuse_ops.py
+++ b/backends/cadence/aot/fuse_ops.py
@@ -16,6 +16,9 @@ from collections import deque
 from numbers import Number
 from typing import cast, Sequence
 
+# Import these for the cadence function signatures.
+import executorch.backends.cadence.aot.ops_registrations # noqa: F401
+
 import torch
 import torch.fx
 from executorch.backends.cadence.aot.compiler_utils import (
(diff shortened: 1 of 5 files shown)


sabarishsnk (approved):

sabarishsnk (approved):

skrtskrtfb (approved):

backends/cadence/aot/fuse_ops.py
```diff
@@ -16,6 +16,9 @@
 from numbers import Number
 from typing import cast, Sequence
 
+# Import these for the cadence function signatures.
+import executorch.backends.cadence.aot.ops_registrations # noqa: F401
```
dbort, line 20: This line broke the linter job: https://github.com/pytorch/executorch/actions/runs/13398623818/job/37423762070#step:15:7593

I think it needs two spaces before the `#`. Please run `lintrunner -a` to fix this, and make sure that the job is passing before merging PRs.

dbort: Sent https://github.com/pytorch/executorch/pull/8575 to fix this

PR #8507, opened 2025-02-14
Enable dtype selective build by default

GregoryComer:
Summary: Now that dtype selective build is relatively mature and well-supported, let's try enabling it by default. There have been a few cases where operator libs in production were built without dtype selective build and ended up much larger than they need to be.

Differential Revision: D69686466


swolchok (approved):
sure why not

PR #8511, opened 2025-02-17
Move iOS docs to top, add Android placeholders

GregoryComer:
### Summary
Move Apple platform docs up to the top level. Add Android doc placeholders.

### Test plan
Built docs locally.

### DIFF
--- a/docs/source/using-executorch-ios.md
+++ b/docs/source/using-executorch-ios.md
@@ -1,3 +1,0 @@
-# Using ExecuTorch on iOS
-
-Placeholder for top-level iOS documentation
--- a/docs/source/using-executorch-ios.md
+++ b/docs/source/using-executorch-ios.md
@@ -1,6 +1,8 @@
-# Integrating and Running ExecuTorch on Apple Platforms
+# Using ExecuTorch on iOS
 
-**Author:** [Anthony Shoumikhin](https://github.com/shoumikhin)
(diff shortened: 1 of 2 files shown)


docs/source/using-executorch-ios.md
```diff
@@ -1,3 +1,324 @@
 # Using ExecuTorch on iOS
 
-Placeholder for top-level iOS documentation
\ No newline at end of file
+ExecuTorch supports both iOS and macOS via Objective-C, Swift, and C++. ExecuTorch also provides backends to leverage Core ML and Metal Performance Shaders (MPS) for hardware-accelerated execution on Apple platforms.
```
mergennachin, line 3: Is this basically a move? 

Then do `git mv` to preserve the git history. And it's easier to review the small deltas (if any).

mergennachin (approved):
Can you add the hyperlinks to these pages in Getting Started page? 

https://docs-preview.pytorch.org/pytorch/executorch/8511/getting-started.html

Currently, they're just in text format.

docs/source/using-executorch-ios.md
GregoryComer: Definitely. I did actually git mv in here. Looking at the blame on the branch, it looks like the history is preserved. I think it doesn't show up as a direct move in the PR since there are a few changes in the subsequent commit and the PR shows the aggregate deltas.

PR #8512, opened 2025-02-17
Qualcomm AI Engine Direct - CI For Llama

winskuo-quic:
### Summary

- Isolate LLM to an individual class for test_qnn_delegate.py
- This PR should not affect Executorch's CI. This is mainly for internal CI  that checks pte size, accuracy, and inference speed. Runs stories110m and Llama 3.2 1B

cc @cccclai @shewu-quic @cbilgin @mergennachin @byjlw

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
+++ b/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
@@ -18,6 +18,7 @@
 #include <executorch/runtime/core/exec_aten/util/scalar_type_util.h>
 #include <executorch/runtime/platform/log.h>
 #include <ctime>
+#include <fstream>
 #include <sstream>
 
 using executorch::aten::Tensor;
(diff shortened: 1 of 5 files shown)


winskuo-quic:
Hi @cccclai,
This PR is mainly for our internal CI to perform inference speed test on Stories Llama and Llama 3.2 1B.
ExecuTorch's CI will still work as usual, which tests pte size and accuracy using Stories Llama.
Please have a look.
Thanks.

examples/qualcomm/oss_scripts/llama/runner/runner.cpp
```diff
@@ -518,6 +519,19 @@ void printReport(const Runner::Stats& stats) {
       stats.num_generated_tokens,
       (double)stats.aggregate_sampling_time_ms /
           stats.SCALING_FACTOR_UNITS_PER_SECOND);
+
+  // For now, we just print the total inference time for CI, can save more info
+  // in future if needed.
+  std::ofstream outfile("outputs/inference_speed.txt");
```
cccclai, line 525: It seems default to write to this path. If users don't have this path then I assume it will fail? - can we make this runner more generic so it can be reused directly

cccclai (approved):
Looks good!

examples/qualcomm/oss_scripts/llama/runner/runner.cpp
winskuo-quic: Yes, I think by running the executable directly without using llama.py, it will fail. I will make a separate PR making this more flexible to users not using python script.

PR #8518, opened 2025-02-17
Arm: Support bitwise and, xor, and or ops

Erik-Lundell:
Ops are very similar and thus clumped together.
No quantization since that doesn't make sense for
bitwise ops.

Add a factory for creating simple two input NodeVisitors. This can be extended for future such ops.

### DIFF
--- a/backends/arm/operators/ops_binary.py
+++ b/backends/arm/operators/ops_binary.py
@@ -1,0 +1,51 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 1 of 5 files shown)


zingo:
MacOS fails unrelated

zingo (approved):

backends/arm/operators/ops_binary.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-unsafe
```
digantdesai, line 6: drop this in the future?

PR #8519, opened 2025-02-17
Move ReplaceScalarWithTensorArgPass to transforms

Erik-Lundell:
The pass is general and can be used by multiple backends. Use it in Arm backend and make small adjustments to make it work.


zingo (approved):

zingo:
MacOs tests unrelated

PR #8520, opened 2025-02-17
Allow getting all backend names

cccclai:
Summary: Allow getting all backends name in both python and c++

Differential Revision: D69691354

### DIFF
--- a/extension/pybindings/pybindings.cpp
+++ b/extension/pybindings/pybindings.cpp
@@ -23,6 +23,7 @@
 #include <executorch/extension/data_loader/buffer_data_loader.h>
 #include <executorch/extension/data_loader/mmap_data_loader.h>
 #include <executorch/extension/memory_allocator/malloc_memory_allocator.h>
+#include <executorch/runtime/backend/interface.h>
 #include <executorch/runtime/core/data_loader.h>
 #include <executorch/runtime/core/exec_aten/util/scalar_type_util.h>
 #include <executorch/runtime/executor/method.h>
@@ -91,6 +92,8 @@ using ::executorch::runtime::DataLoader;
 using ::executorch::runtime::Error;
 using ::executorch::runtime::EValue;
 using ::executorch::runtime::EventTracerDebugLogLevel;
+using ::executorch::runtime::get_backend_name;
+using ::executorch::runtime::get_num_registered_backends;
 using ::executorch::runtime::get_registered_kernels;
 using ::executorch::runtime::HierarchicalAllocator;
 using ::executorch::runtime::Kernel;
@@ -975,6 +978,18 @@ py::list get_operator_names() {
   return res;
 }
 
+py::list get_registered_backend_names() {
+  size_t n_of_registered_backends = get_num_registered_backends();
+  py::list res;
+  for (size_t i = 0; i < n_of_registered_backends; i++) {
+    auto backend_name_res = get_backend_name(i);
+    THROW_IF_ERROR(backend_name_res.error(), "Failed to get backend name");
+    auto backend_name = backend_name_res.get();
+    res.append(backend_name);
+  }
+  return res;
+}
+
 } // namespace
 
 PYBIND11_MODULE(EXECUTORCH_PYTHON_MODULE_NAME, m) {
@@ -1027,6 +1042,10 @@ PYBIND11_MODULE(EXECUTORCH_PYTHON_MODULE_NAME, m) {
             reinterpret_cast<const char*>(prof_result.prof_data),
             prof_result.num_bytes);
       },
+      call_guard);
+  m.def(
+      "_get_registered_backend_names",
+      &get_registered_backend_names,
       call_guard);
   m.def("_get_operator_names", &get_operator_names);
   m.def("_create_profile_block", &create_profile_block, call_guard);
--- a/extension/pybindings/test/test_backend_pybinding.py
+++ b/extension/pybindings/test/test_backend_pybinding.py
@@ -1,0 +1,14 @@
+import unittest
+
+from executorch.runtime import Runtime
+
+
+class TestBackendsPybinding(unittest.TestCase):
+    def test_backend_name_list(
+        self,
+    ) -> None:
+
+        runtime = Runtime.get()
+        registered_backend_names = runtime.backend_registry.registered_backend_names
+        self.assertGreaterEqual(len(registered_backend_names), 1)
+        self.assertIn("XnnpackBackend", registered_backend_names)
--- a/runtime/__init__.py
+++ b/runtime/__init__.py
@@ -42,7 +42,7 @@ Example output:
 import functools
 from pathlib import Path
 from types import ModuleType
-from typing import Any, BinaryIO, Dict, Optional, Sequence, Set, Union
+from typing import Any, BinaryIO, Dict, List, Optional, Sequence, Set, Union
 
 try:
     from executorch.extension.pybindings.portable_lib import (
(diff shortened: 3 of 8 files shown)


tarun292 (approved):

larryliu0820:
We are moving to `[runtime/__init__.py](https://github.com/pytorch/executorch/blob/main/runtime/__init__.py#L136)` for our runtime python APIs. Can you add this new API there as well?

larryliu0820 (changes requested):
Put it back to your queue

runtime/backend/interface.cpp
```diff
@@ -55,5 +55,9 @@ Error register_backend(const Backend& backend) {
   return Error::Ok;
 }
 
+Span<const Backend> get_registered_backends() {
```
dbort, line 58: If we ever allow for runtime-registered (or deregistered) backends, this API will be very hard to support: the function must return a single, contiguous buffer containing all backends, and the caller could maintain a pointer to that buffer for the lifetime of the process.

It would be safer and more maintainable to have an API like
```
size_t get_num_registered_backends();
Result<const char*> get_backend_name(size_t index);
```

Also I think we should hide/remove the current `Backend` struct, which is really an internal detail of how the table is stored. It's used to register backends, but that actually complicates things for callers since they need to create that struct. It'd be easier for callers if the API were `register_backend(const char* name, BackendInterface* backend);`

dbort (changes requested):

cccclai:
@larryliu0820 @dbort updated and addressed the comments. Please take another look.

larryliu0820 (approved):
Thanks for addressing comments

GregoryComer:
This could be done as a follow-up as I wouldn't want to block this diff, but it would be nice to be able to get only available backends - i.e. backends where is_available() returns true. This is useful for determining if the actual hardware is present. For example, the Qualcomm backend might be linked but not running on a device that actually has a supported NPU. The older API that returned the Backend directly would have allowed for calling is_available(), but the current approach does not. That's fine, as we can add another API to do this - just a consideration.

extension/pybindings/pybindings.cpp
```diff
@@ -975,6 +979,19 @@ py::list get_operator_names() {
...
 }
 
+py::list get_backend_names() {
+  size_t n_of_registered_backends = get_num_registered_backends();
+  py::list res;
+  for (size_t i = 0; i < n_of_registered_backends; i++) {
+    auto backend_name_res = get_backend_name(i);
+    THROW_IF_ERROR(backend_name_res.error(), "Failed to get backend name");
+    auto backend_name = backend_name_res.get();
+    printf("backend name: %s\n", backend_name);
```
GregoryComer, line 989: Maybe leftover debug print?

cccclai:
> This could be done as a follow-up as I wouldn't want to block this diff, but it would be nice to be able to get only available backends - i.e. backends where is_available() returns true. This is useful for determining if the actual hardware is present. For example, the Qualcomm backend might be linked but not running on a device that actually has a supported NPU. The older API that returned the Backend directly would have allowed for calling is_available(), but the current approach does not. That's fine, as we can add another API to do this - just a consideration.

oh I actually already have this diff as part of the follow up. Just haven't submitted yet.

runtime/core/error.h
```diff
@@ -93,6 +93,9 @@ enum class Error : error_code_t {
   /// Execute stage: The handle is invalid.
   DelegateInvalidHandle = 0x32,
 
+  /// Backend registration: trying to access an invalid index in the backend
+  /// registration table.
+  BackendTableAccessFail = 0x33,
```
mergennachin, line 98: I wouldn't add a new error message, just for this. InvalidArgument = 0x12 is sufficient for this case.

runtime/backend/interface.cpp
```diff
@@ -55,5 +55,16 @@ Error register_backend(const Backend& backend) {
   return Error::Ok;
 }
 
+size_t get_num_registered_backends() {
+  return num_registered_backends;
+}
+
+Result<const char*> get_backend_name(size_t index) {
+  if (index >= num_registered_backends) {
+    return Error::BackendTableAccessFail;
```
mergennachin, line 64: InvalidArgument = 0x12

runtime/__init__.py
```diff
@@ -125,6 +125,21 @@ def load_method(self, name: str) -> Optional[Method]:
...
 
+class BackendRegistry:
+    """The registry of backends that are available to the runtime."""
+
+    def __init__(self, legacy_module: ModuleType) -> None:
+        # TODO: Expose the kernel callables to Python.
+        self._legacy_module = legacy_module
+
+    @property
+    def _get_backend_names(self) -> Set[str]:
```
mergennachin, line 136: why start with _?

runtime/__init__.py
```diff
@@ -125,6 +125,21 @@ def load_method(self, name: str) -> Optional[Method]:
...
+        """
+        Returns the names of all registered backends as a list of strings.
```
mergennachin, line 138: as a set?

extension/pybindings/test/test_backend_pybinding.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+
+from executorch.extension.pybindings.portable_lib import _get_backend_names
+
+
+class TestBackendsPybinding(unittest.TestCase):
+    def test_backend_name_list(
+        self,
+    ) -> None:
+        all_backend_name = _get_backend_names()
```
mergennachin, line 10: In the unit test, use the new pybindings API

```
from executorch.runtime import Runtime

runtime = Runtime.get()

runtime.backend_registry.get_backend_names()
```

runtime/__init__.py
mergennachin, line 136: I'd rename to `get_registered_backends()`

extension/pybindings/pybindings.cpp
```diff
@@ -975,6 +979,18 @@ py::list get_operator_names() {
   return res;
 }
 
+py::list get_backend_names() {
```
mergennachin, line 982: get_registered_backends

extension/pybindings/pybindings.cpp
```diff
@@ -975,6 +979,18 @@ py::list get_operator_names() {
...
+  size_t n_of_registered_backends = get_num_registered_backends();
+  py::list res;
```
mergennachin, line 984: is the result list or set? in the typehint, it is Set[str] but you're returning a list. Which one is it?

mergennachin (changes requested):

runtime/core/error.h
cccclai: Sure

runtime/__init__.py
cccclai: updated

runtime/__init__.py
cccclai: list is better. Borrowed some code from `get_operators_names` but now I don't think we will have duplication

extension/pybindings/test/test_backend_pybinding.py
cccclai: updated

extension/pybindings/pybindings.cpp
cccclai: updated to `get_registered_backend_names`

extension/pybindings/pybindings.cpp
cccclai: It should be list

mergennachin (approved):

extension/pybindings/pybindings.cpp
```diff
@@ -87,10 +88,13 @@ using ::executorch::extension::BufferDataLoader;
 using ::executorch::extension::MallocMemoryAllocator;
 using ::executorch::extension::MmapDataLoader;
 using ::executorch::runtime::ArrayRef;
+using ::executorch::runtime::Backend;
```
mergennachin, line 91: do you need this?

cccclai: updated

extension/pybindings/pybindings.cpp
```diff
@@ -1028,6 +1043,10 @@ PYBIND11_MODULE(EXECUTORCH_PYTHON_MODULE_NAME, m) {
             prof_result.num_bytes);
       },
       call_guard);
+  m.def(
+      "_get_registered_backend_names",
```
dbort, line 1047: Please add this new function to https://github.com/pytorch/executorch/blob/main/extension/pybindings/pybindings.pyi

dbort (approved):
Thanks for changing the API to be index-based. Looks good apart from my final request to update the .pyi file.

extension/pybindings/pybindings.cpp
cccclai: updated

PR #8521, opened 2025-02-17
Allow querying all backends registered with the executorch runtime

### LINKED ISSUE #8267: Provide API to query backend availability at runtime

Internally, backends implement an `is_available` method to determine if the backend is actually present at runtime (detecting where a supported NPU is present, for example). To my knowledge, this isn't actually exposed through the public runtime API surface. 

This is an important usability feature for users who want to support multiple delegates, as users need a way to determine which hardware is actually present at runtime. It is more or less required for people to be able to use NPU delegates, unless some other solution exists to solve this.

This could be done by simply adding a method to the core runtime to enumerate available backends. Alternatively, there was an ongoing discussion and design proposal for generalized backend capabilities that may also provide this functionality (I believe @guangy10 was driving this a while back).

Edit: More clearly enumerating the problem and use case, imagine you are a user that wants to target one or more NPUs on Android, falling back to CPU if none are available. You generate a PTE for each backend. How do you know which one to actually load at runtime? Ideally, executorch should be able to just say qualcomm and xnnpack backends are available, for example, or maybe just xnnpack on a  device without a supported qualcomm NPU. The user code can then choose a .pte based on this information.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @JacobSzwejbka @dbort @mergennachin @byjlw

omerjerk:
Summary: Partially resolves: https://github.com/pytorch/executorch/issues/8267

Differential Revision: D69754074

### DIFF
--- a/runtime/backend/interface.cpp
+++ b/runtime/backend/interface.cpp
@@ -41,6 +41,14 @@ BackendInterface* get_backend_class(const char* name) {
   return nullptr;
 }
 
+std::vector<BackendInterface*> get_all_backend_classes() {
+  std::vector<BackendInterface*> backends(num_registered_backends);
+  for (size_t i = 0; i < num_registered_backends; i++) {
+    backends[i] = registered_backends[i].backend;
+  }
+  return backends;
+}
+
 Error register_backend(const Backend& backend) {
   if (num_registered_backends >= kMaxRegisteredBackends) {
     return Error::Internal;
--- a/runtime/backend/interface.h
+++ b/runtime/backend/interface.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <cstring>
+#include <vector>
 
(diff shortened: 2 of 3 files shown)


swolchok:
Adding reviewers in the top-right corner of this PR page will help get your PR looked at.

std::vector cannot be used in runtime/backend; it is part of ExecuTorch core. See https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#cc-standard-library-usage . (We should update that document to clarify what is in ExecuTorch core and what is not...)

swolchok (changes requested):
for std::vector. see top-level comment on PR. not sure offhand how to get this done without heap allocation but I imagine there are other examples of "enumeration functions" as mentioned in the issue. I would recommend just checking with issue owner @GregoryComer

PR #8541, opened 2025-02-18
Initialize the ET PAL in broadcast_test

swolchok:
Test is broken internally because this is missing. (Is there a different way I'm supposed to fix this?)


mergennachin:
> Test is broken internally because this is missing. 

Any regression tests we can add in OSS to catch similar issues in the future?

swolchok:
@mergennachin i was hoping somebody on the review would know why it has to be initialized internally but not externally. Since you don't, I can look

GregoryComer:
Yeah, there's not really a great solution for PAL init in tests at this point. I've run into this in the past and added a common fixture for operator-level tests to solve this there (https://github.com/pytorch/executorch/blob/main/kernels/test/TestUtil.h#L97). It might be better to use GTest global setup to do this - I haven't looked into it extensively.

GregoryComer (approved):

PR #8550, opened 2025-02-18
Run unittests in debug mode at pull time and release mode on trunk builds

swolchok:
Per discussion in discord with @mergennachin. Motivation was to have debug coverage to catch debug-only test failures.


swolchok:
* (to be filled)

swolchok:
linux unittest jobs are green and mac test is broken on trunk, think this is ready to go

.ci/scripts/utils.sh
```diff
@@ -124,7 +124,7 @@ build_executorch_runner() {
   if [[ $1 == "buck2" ]]; then
     build_executorch_runner_buck2
```
mergennachin, line 125: for buck we are ignoring and using whatever's default?

mergennachin (commented):
Shouldn't we also plumb all the way to 
.ci/scripts/unittest-linux-cmake.sh and
https://github.com/pytorch/executorch/blob/main/test/run_oss_cpp_tests.sh?

.ci/scripts/utils.sh
swolchok: I don't think it's necessary to do both a debug and a release build with buck2 specifically; whether the code works with optimizing compiler flags should be orthogonal to choice of build system.

swolchok:
> Shouldn't we also plumb all the way to
.ci/scripts/unittest-linux-cmake.sh and
https://github.com/pytorch/executorch/blob/main/test/run_oss_cpp_tests.sh?

As you can see in [the linux release build logs](https://github.com/pytorch/executorch/actions/runs/13401575977/job/37433420335?pr=8550) and [the linux debug build logs](https://github.com/pytorch/executorch/actions/runs/13401575554/job/37433455048?pr=8550), CMAKE_BUILD_TYPE is correctly set as-is. CMake caches settings, and 1) unittest-linux-cmake.sh does not itself invoke cmake 2) the cmake invocation in run_oss_cpp_tests.sh doesn't set CMAKE_BUILD_TYPE explicitly.

.ci/scripts/utils.sh
swolchok: yes. I think that whether the code works with optimized/debug build flags is orthogonal to the choice of build system. (The ExecuTorch .buckconfig is configured to produce debug builds by default.)

mergennachin (approved):

.ci/scripts/unittest-linux.sh
```diff
@@ -0,0 +1,48 @@
...
+.ci/scripts/setup-linux.sh "$BUILD_TOOL" "$BUILD_MODE"
+
+# Install llama3_2_vision dependencies.
+PYTHON_EXECUTABLE=python ./examples/models/llama3_2_vision/install_requirements.sh
+
+if [[ "$BUILD_TOOL" == "cmake" ]]; then
+    .ci/scripts/unittest-linux-cmake.sh
+elif [[ "$BUILD_TOOL" == "buck2" ]]; then
+    .ci/scripts/unittest-buck2.sh
+else
```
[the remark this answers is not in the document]
swolchok: good catch. fix is to remove it in unittest-macos.sh

mergennachin (approved):

swolchok:
double-checking CMAKE_BUILD_TYPE in the unittest cmake jobs on this diff: all 3 mac jobs are good, all 3 linux jobs are good. found unittest-arm while looking; should talk about whether we need to give that one the same treatment. merging.

PR #8442, opened 2025-02-13
Fix shadowing of C <complex.h> by c10 include directory

swolchok:
New test dont_shadow_complex_test.c illustrates the previous problem; new README.md explains the fix.

Test Plan:  dont_shadow_complex_test.c used to fail to build, now it succeeds.

PR #8551, opened 2025-02-18
[cadence][hifi] Fixed all hifi ops and cmake

zonglinpeng:
### Summary
- fixed renaming ops in https://github.com/pytorch/executorch/pull/8491
- fixed c10 updates from https://github.com/pytorch/executorch/pull/8320 and https://github.com/pytorch/executorch/pull/8442
- updated cpp returns which had warning

### Test plan
backends/cadence/build_cadence_hifi4.sh

### DIFF
--- a/backends/cadence/CMakeLists.txt
+++ b/backends/cadence/CMakeLists.txt
@@ -23,7 +23,7 @@ include(${EXECUTORCH_ROOT}/build/Utils.cmake)
 
 # Let files say "include <executorch/path/to/header.h>".
 set(_common_include_directories ${EXECUTORCH_ROOT}/..
-  ${EXECUTORCH_ROOT}/runtime/core/portable_type)
+  ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10)
 
 add_compile_definitions(C10_USING_CUSTOM_GENERATED_MACROS)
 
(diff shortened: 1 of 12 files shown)


backends/cadence/CMakeLists.txt
```diff
@@ -22,7 +22,10 @@ endif()
 include(${EXECUTORCH_ROOT}/build/Utils.cmake)
 
 # Let files say "include <executorch/path/to/header.h>".
-set(_common_include_directories ${EXECUTORCH_ROOT}/..)
+set(_common_include_directories ${EXECUTORCH_ROOT}/..
+  ${EXECUTORCH_ROOT}/runtime/core/portable_type)
```
swolchok, line 26: I believe this needs to be ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10; if your build is working without it then maybe you don't need it at all?

swolchok (commented):
leaving accept for cadence folks

backends/cadence/CMakeLists.txt
zonglinpeng: good catch. Surprisingly this path is required for build but still worked without c10

skrtskrtfb (approved):

PR #8552, opened 2025-02-18
Add tests for qwen + allow uninitialized weights in Llama model

jackzhxng:
### Summary
Add basic ci test for [Qwen](https://github.com/pytorch/executorch/pull/8355) model. Requires some changes to `llama/model.py` to allow uninitialized (random) weights.

### Test plan
- Tested and passed locally
- See if CI passes for the new test.


examples/models/llama/export_llama_lib.py
```diff
@@ -84,13 +84,15 @@
...
 
+# All models that leverage the transformer architecture defined in llama_transformer.py.
 EXECUTORCH_DEFINED_MODELS = [
     "stories110m",
     "llama2",
     "llama3",
     "llama3_1",
     "llama3_2",
     "static_llama",
+    "qwen2_5",
```
larryliu0820, line 95: keep sorted pls

examples/models/llama/model.py
```diff
@@ -236,14 +236,23 @@ def __init__(self, **kwargs):
...
+        try:
+            # assign=True: load params/buffers by assignment instead of performing an in-place copy.
+            # Because we are using device="meta", tensors do not have memory associated with them
+            # and an in-place copy is a no-op. Use assign=True in load_state_dict for this scenario.
+            missing, unexpected = self.model_.load_state_dict(
+                checkpoint,
+                strict=False,
+                assign=True,
+            )  # self.model_ = Transformer(gptconf)
+        except RuntimeError as e:
```
larryliu0820, line 249: Why is this needed?

larryliu0820 (approved):
I'm ok with the changes but I'm really concerned on llama/model.py and I think we should clean it up. I'll create a separate issue.

Issue #8418, opened 2025-02-12
Make sure all license headers are BSD license (or compatible)


mergennachin:
- Make sure all license headers are BSD license.
- Make sure lint rules are in place for future such cases.

mergennachin:
Hi @mcremon-meta 

Majority of these licenses are in cadence folder. https://gist.github.com/mergennachin/83bef4ab9bb5d721284ddb50168289d4

executorch repo is BSD license altogether, unless it is a third-party library. 

A standard license header for individual files are something like this:

"# Copyright (c) Meta Platforms, Inc. and affiliates.
# All rights reserved.
#
# This source code is licensed under the BSD-style license found in the
# LICENSE file in the root directory of this source tree."

Can we codemod these files?

mergennachin:
@mcremon-meta 

It's currently coming from our partners. In particular, during the IP review, their attorneys are raising this as potential legal risks when contributing to our repo. So, I'd say somewhat urgent.

In parallel, I will create OSS lintrunner rule to make sure it doesn't happen in the future.

zonglinpeng:
@mergennachin Thanks for catching this - Will fix them now and target to ship by EOD if that works? Updating BSD license header was long on my TODO but good to know it's urgent :) 
cc. @mcremon-meta

mcremon-meta:
https://github.com/pytorch/executorch/pull/8420 should fix our side, let me know if I missed any!

mergennachin:
@digantdesai 

remaining here: https://github.com/pytorch/executorch/pull/8428

PR #8554, opened 2025-02-19
Add c10::irange to ExecuTorch

swolchok:
* __->__ #8554

irange is a header-only utility that is both more readable than the usual way to write for loops and also solves -Wsign-compare issues. I've previously vetted that it generates the same assembly as a regular for loop.

Differential Revision: [D69817195](https://our.internmc.facebook.com/intern/diff/D69817195/)

### DIFF
--- a/runtime/core/portable_type/c10/c10/targets.bzl
+++ b/runtime/core/portable_type/c10/c10/targets.bzl
@@ -26,6 +26,7 @@ def define_common_targets():
             "util/TypeSafeSignMath.h",
             "util/bit_cast.h",
             "util/floating_point_utils.h",
+            "util/irange.h",
         ],
         exported_preprocessor_flags = [
             # NOTE: If we define C10_EMBEDDED to prevent Half and
--- a/runtime/core/portable_type/c10/c10/util/irange.h
+++ b/runtime/core/portable_type/c10/c10/util/irange.h
@@ -1,0 +1,123 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
+
+#include <c10/util/TypeSafeSignMath.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <type_traits>
+
+namespace c10 {
+
+namespace detail {
+
+template <
+    typename I,
+    bool one_sided = false,
+    std::enable_if_t<std::is_integral_v<I>, int> = 0>
+struct integer_iterator {
+  using iterator_category = std::input_iterator_tag;
+  using value_type = I;
+  using difference_type = std::ptrdiff_t;
+  using pointer = I*;
+  using reference = I&;
+
+  explicit integer_iterator(I value) : value(value) {}
+
+  I operator*() const {
+    return value;
+  }
+
+  I const* operator->() const {
+    return &value;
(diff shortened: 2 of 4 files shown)


runtime/core/portable_type/c10/c10/targets.bzl
```diff
@@ -26,6 +26,7 @@ def define_common_targets():
             "util/TypeSafeSignMath.h",
             "util/bit_cast.h",
             "util/floating_point_utils.h",
+            "util/irange.h",
```
kimishpatel, line 29: is irange.h not part of headers from pytorch? or we decided that is not worth sharing?

runtime/core/portable_type/c10/c10/util/irange.h
```diff
@@ -0,0 +1,123 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#pragma once
```
kimishpatel, line 3: If this file is copied over, can you add a comment here to the effect?

runtime/core/portable_type/c10/c10/util/irange.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <c10/util/TypeSafeSignMath.h>
+
+#include <algorithm>
```
mergennachin, line 7: does it need `<algorithm>`?

runtime/core/portable_type/tensor_impl.cpp
```diff
@@ -11,6 +11,8 @@
 #include <algorithm>
```
mergennachin, line 11: not related to this diff per se, but makes me curious why we need <algorithm> in this file too.

runtime/core/portable_type/c10/c10/util/irange.h
```diff
@@ -0,0 +1,123 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
```
mergennachin, line 1: +1

Add BSD license header

Also, i have a task to add a lint rule: https://github.com/pytorch/executorch/issues/8418

mergennachin (approved):

runtime/core/portable_type/c10/c10/util/irange.h
swolchok: > can you add a comment

no. the file should be identical to the version in pytorch core.

runtime/core/portable_type/c10/c10/targets.bzl
swolchok: it is a pytorch header, but it is used in executorch core. therefore it must be copied over just like everything else in runtime/core/portable_type/c10/c10 to keep the build hermetic for embedded users

runtime/core/portable_type/c10/c10/util/irange.h
swolchok: not relevant; this is a direct copy from pytorch core.

runtime/core/portable_type/tensor_impl.cpp
swolchok: [std::equal](https://en.cppreference.com/w/cpp/algorithm/equal)

PR #8566, opened 2025-02-19
Arm: Fix various small issues

Erik-Lundell:
- Softmax incorrectly set to can_delegate=False in aot_arm_compiler
- Incorrect % with rank instead of dim length. Fix test for this.
- Remove .dump_artifact():s


zingo (approved):

zingo:
fails unrelated

PR #8568, opened 2025-02-19
Use portable lib for finetuning demo

dpalmasan:
Summary: Update the example of LLM fine-tuning on ET using portable_lib and XNNPack backend

Reviewed By: jackzhxng

Differential Revision: D69058862

### DIFF
--- a/examples/llm_pte_finetuning/runner.py
+++ b/examples/llm_pte_finetuning/runner.py
@@ -15,7 +15,7 @@ from executorch.examples.llm_pte_finetuning.training_lib import (
     update_function,
 )
 
-from executorch.extension.pybindings.aten_lib import (  # @manual
+from executorch.extension.pybindings.portable_lib import (  # @manual
     _load_for_executorch_from_buffer,
 )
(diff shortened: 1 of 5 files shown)


examples/llm_pte_finetuning/runner.py
```diff
@@ -30,6 +30,18 @@
 )
 parser.add_argument("--cfg", type=str, help="Path to the config file.")
 parser.add_argument("--model_file", type=str, help="Path to the ET model file.")
+parser.add_argument(
+    "--num_training_steps",
+    type=int,
+    help="Number of training steps, assuming 1 epoch.",
+    default=1,
```
JacobSzwejbka, line 37: should we do a different default here? Like the avg number we've noticed to see some convergence?

JacobSzwejbka (approved):

examples/llm_pte_finetuning/runner.py
dpalmasan: I updated it to 5 eval steps and 100 steps as we did in our manual tests

PR #8577, opened 2025-02-19
Log when a Result is created from Error::ok

JacobSzwejbka:
Summary: If this ctor is firing something has probably gone wrong. I found it a little hard to debug so adding logging to hopefully save someone else time in the future.

Differential Revision: D69870978


dbort (approved):
Seems reasonable.

Issue #8576, opened 2025-02-19
Resolve the discrepancy of latency report between LLMs and non-LLMs


guangy10:
### 🐛 Describe the bug

![Image](https://github.com/user-attachments/assets/d3ef85bf-c057-490a-ab1d-96f5d691817a)

As shown on the dashboard, the `avg_inference_latency (ms)` is skipped for LLM, and report only `generate_time (ms)` instead.

Upon checking the iOS run for example, a LLM job will run three tests on-device to report different metrics:
  1. `test_load_llama_3_2_1b_llama3_fb16_pte_iOS_17_2_1_iPhone15_4`
  2. `test_forward_llama_3_2_1b_llama3_fb16_pte_iOS_17_2_1_iPhone15_4`
  3. `test_generate_llama_3_2_1b_llama3_fb16_pte_tokenizer_model_iOS_17_2_1_iPhone15_4`
While a non-LLM job will only run the first two tests (test_load_ and test_forward_ ) instead.

See detailed jobs here: 
  - LLM: https://github.com/pytorch/executorch/actions/runs/13403521306/job/37441009799
  - non-LLM: https://github.com/pytorch/executorch/actions/runs/13403521306/job/37441008720

Three things to get clarification:
  1. Because `test_forward_*` is reported to both LLM and non-LLM, why isn't reported to the dash?
  2. Let's annotate each metrics in the DB so users will know what exactly is measured by each.
  3. Confirm if Android is measuring and reporting exact same metrics

### Versions

trunk

guangy10:
Android benchmark app doesn’t save inference latency for LLM https://github.com/pytorch/executorch/blob/main/extension/benchmark/android/benchmark/app/src/main/java/org/pytorch/minibench/LlmBenchmarkActivity.java#L95-L112 while non-LLM does

@kirklandsign Can we add raw inference latency for LLM? It will be useful in the case to detect whether the slowness is from core runtime or the tokenizer itself.

shoumikhin:
forward tests run forward and measure latency on any model.

generate tests measure tokens per second specifically, leveraging the llama runner to predict the next token several times consecutively, and the runner eventually calls forward under the hood each time.

@huydhn @guangy10 let me know if you need any further details.

guangy10:
> forward tests run forward and measure latency on any model.
> 
> generate tests measure tokens per second specifically, leveraging the llama runner to predict the next token several times consecutively, and the runner eventually calls forward under the hood each time.
> 
[...]

PR #8580, opened 2025-02-19
Measure generate_time on iOS benchmark

huydhn:
Addresses the iOS part of https://github.com/pytorch/executorch/issues/8576#issuecomment-2669634098

Let's see if this fix the issue

### Testing

~~https://github.com/pytorch/executorch/actions/runs/13423283360~~ Another test https://github.com/pytorch/executorch/actions/runs/13442356702

cc @guangy10 @kirklandsign @shoumikhin

### DIFF
--- a/.github/scripts/extract_benchmark_results.py
+++ b/.github/scripts/extract_benchmark_results.py
@@ -229,11 +229,7 @@ def extract_ios_metric(
 
     elif method == "forward":
         if metric_name == "Clock Monotonic Time, s":
-            benchmark_result["metric"] = (
-                "generate_time(ms)"
-                if "llama" in test_name
-                else "avg_inference_latency(ms)"
-            )
+            benchmark_result["metric"] = "avg_inference_latency(ms)"
             benchmark_result["actualValue"] = metric_value * 1000
 
         elif metric_name == "Memory Peak Physical, kB":
@@ -241,9 +237,14 @@ def extract_ios_metric(
             benchmark_result["metric"] = "peak_inference_mem_usage(mb)"
             benchmark_result["actualValue"] = metric_value / 1024
 
-    elif method == "generate" and metric_name == "Tokens Per Second, t/s":
-        benchmark_result["metric"] = "token_per_sec"
-        benchmark_result["actualValue"] = metric_value
+    elif method == "generate":
+        if metric_name == "Clock Monotonic Time, s":
+            benchmark_result["metric"] = "generate_time(ms)"
+            benchmark_result["actualValue"] = metric_value * 1000
+
+        elif metric_name == "Tokens Per Second, t/s":
+            benchmark_result["metric"] = "token_per_sec"
+            benchmark_result["actualValue"] = metric_value
 
     return benchmark_result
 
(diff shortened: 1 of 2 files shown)


shoumikhin (approved):

.github/scripts/extract_benchmark_results.py
```diff
@@ -229,21 +229,22 @@ def extract_ios_metric(
...
-        benchmark_result["metric"] = "token_per_sec"
-        benchmark_result["actualValue"] = metric_value
+    elif method == "generate":
+        if metric_name == "Clock Monotonic Time, s":
+            benchmark_result["metric"] = "generate_time(ms)"
+            benchmark_result["actualValue"] = metric_value * 1000
+
+        elif metric_name == "Tokens Per Second, t/s":
+            benchmark_result["metric"] = "token_per_sec"
+            benchmark_result["actualValue"] = metric_value
```
guangy10, line 247: Seems like they are measuring same thing, we won't need both if that's the case. Left a comment in https://github.com/pytorch/executorch/issues/8576#issuecomment-2672384424 for @shoumikhin to clarify.

.github/scripts/extract_benchmark_results.py
```diff
@@ -229,21 +229,22 @@ def extract_ios_metric(
...
             benchmark_result["metric"] = "peak_inference_mem_usage(mb)"
             benchmark_result["actualValue"] = metric_value / 1024
 
-    elif method == "generate" and metric_name == "Tokens Per Second, t/s":
-        benchmark_result["metric"] = "token_per_sec"
-        benchmark_result["actualValue"] = metric_value
+    elif method == "generate":
+        if metric_name == "Clock Monotonic Time, s":
+            benchmark_result["metric"] = "generate_time(ms)"
+            benchmark_result["actualValue"] = metric_value * 1000
```
guangy10, line 243: I think in code, `TPS` and `generate_time` are calculated independently. That's why I see some fields report regression only on `TPS` but not on `generate_time`. See the screenshot below. @shoumikhin Should we consolidate the measurement and only report one metric instead I'd prefer TPS as it's more human readable.

Screenshot 2025-02-20 at 10 54 28 AM

guangy10 (approved):
Left a few comment inline. Idk Anthony would have time to take a look as he is OOO this week. @huydhn if you agree with my comment, we can go ahead fix the confusion now, and open to make further adjustment on feedback

.github/scripts/extract_benchmark_results.py
huydhn: Oh, I think this is the bug is about, the current `generate_time` is actually `avg_inference_latency` from `forward`. After this change fixes this issue, we can consider keeping just TPS I guess

guangy10:
@kirklandsign we should make the same fix on Android side as well

huydhn:
> The title should be measure "avg_inference_time" right?

Nah, it measure `generate_time`.  What is called `generate_time` atm is actually `avg_inference_time`

.github/scripts/extract_benchmark_results.py
guangy10: @huydhn Do you want to remove/hide `generate_time` from the dashboard in a follow-up PR? The Android side needs to be corrected first I guess before it. cc: @kirklandsign

huydhn: Yup, the PR for that is ready here https://github.com/pytorch/test-infra/pull/6314

PR #8593, opened 2025-02-20
Arm backend: Add additional tosa_supported_op checks for BI

oscarandersson8218:
### Summary
If a TosaSpecification without floating point support is used, additional checks will be made during paritioning to make sure that we don't partition operators that:
- are not quantized properly, i.e. does not have a dq-q pair surrounding them.
- should have been decomposed prior to qunatization, e.g. div should be decomposed to a mul and recip before quantization.


zingo (approved):

zingo:
Fails are unrelated

PR #8611, opened 2025-02-21
Use faster method to find the torch Python module's __path__

swolchok:
I noticed that the previous method took over 1 second on my M1 Mac Pro, so I found this method, which takes under 100 ms. I believe the root cause is that actually importing torch takes a little time.

cc @larryliu0820 @lucylq


swolchok:
* (to be filled)

mergennachin (approved):

PR #8613, opened 2025-02-21
Use portable lib for finetuning demo

dpalmasan:
Summary: Update the example of LLM fine-tuning on ET using portable_lib and XNNPack backend

Differential Revision: D69954671

### DIFF
--- a/examples/llm_pte_finetuning/model_loading_lib.py
+++ b/examples/llm_pte_finetuning/model_loading_lib.py
@@ -9,8 +9,9 @@
 from typing import Any, Dict, Tuple
 
 import torch
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
 from executorch.examples.llm_pte_finetuning.training_lib import TrainingModule
-from executorch.exir import to_edge
+from executorch.exir import EdgeCompileConfig, to_edge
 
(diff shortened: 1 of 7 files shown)


examples/llm_pte_finetuning/model_loading_lib.py
```diff
@@ -72,16 +73,70 @@ def export_model_lora_training(
...
+        ep = joint_graph
+
+        # Currently there is no implementation of empty_permuted for edge dialect.
+        # We manually make a pass to rewrite the empty_permuted to empty and permute.
+        for node in ep.graph.nodes:
+            if (
+                node.op == "call_function"
+                and node.target == torch.ops.aten.empty_permuted.out
+            ):
+                print("find empty_permute: ", node)
```
swolchok, line 85: nit: I think the past tense "found" is more typical for log output than the present tense "find"

JacobSzwejbka (approved):
Looks good to me. cc @mergennachin who had been thinking about using a different repo for examples.

examples/llm_pte_finetuning/model_loading_lib.py
dpalmasan: Makes sense. Will update it, thanks!

Issue #7634, opened 2025-01-13
QNN: wav2letter doesn't export


guangy10:
### 🐛 Describe the bug

```
python -m examples.qualcomm.scripts.wav2letter -b cmake-out -m SM8450 --compile_only -p ./states_fused.pth
```
fails to export

where states_fused.pth is downloaded from https://github.com/nipponjo/wav2letter-ctc-pytorch/tree/main?tab=readme-ov-file#wav2letter-ctc-pytorch

cc: @cccclai @cbilgin 

### Versions

latest main

cccclai:
Is there any error message?

winskuo-quic:
Hi @guangy10,
I am trying to reproduce this error, but everything is working fine on my side.
I am currently using `torchaudio 2.6.0.dev20250131+cpu`.

guangy10:
@winskuo-quic If you think the model is working fine, should we just enable it in the CI, the setup there can be the source of truth for future reference. Here is the QNN models we are currently running on CI: https://github.com/pytorch/executorch/blob/main/.github/workflows/trunk.yml#L305-L329

Can you add wav2letter to it? The CI only need to test it in `--compile_only` mode, to ensure there is no complication issue

winskuo-quic:
@guangy10 Could you please share where I can place my model's weight file? I can't use wget to retrieve the weights, so I think I might need a location to place the weights where CI can also have access to.

guangy10:
> [@guangy10](https://github.com/guangy10) Could you please share where I can place my model's weight file? I can't use `wget` to retrieve the weights, so I think I might need a location to place the weights where CI can also have access to.

Can we use random weights instead, for compile_only, since the model is not calibrated nor evaluating the accuracy. In the CI the goal is to just ensure it's functioning end-to-end with low latency. Users can always rerun these models locally by disabling the compile_only so that they will get the high quality model easily. wdyt @winskuo-quic

cccclai:
> random weights

Random weights possibly have a different latency number than the actual weights. If it will be used on dashboard, it's probably better to use real weights

guangy10:
> > random weights
> 
> Random weights possibly have a different latency number than the actual weights. If it will be used on dashboard, it's probably better to use real weights

If the weights are public, one solution is to cache the weights in the docker image for qnn. It can avoid redownloading the weights again and again. And we already have that docker.

cccclai:
[...]

PR #8616, opened 2025-02-21
Qualcomm AI Engine Direct - Meta CI for Mobilebert , W2L, and Llama

winskuo-quic:
### Summary

- Some fix from Llama
- Mobilebert CI enablement
- Wav2Letter CI enablement

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -164,6 +164,7 @@ test_model_with_qnn() {
   export LD_LIBRARY_PATH=$QNN_SDK_ROOT/lib/x86_64-linux-clang/
   export PYTHONPATH=$EXECUTORCH_ROOT/..
 
+  EXTRA_FLAGS=""
   if [[ "${MODEL_NAME}" == "dl3" ]]; then
     EXPORT_SCRIPT=deeplab_v3
   elif [[ "${MODEL_NAME}" == "mv3" ]]; then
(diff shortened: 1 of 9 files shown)


winskuo-quic:
Hi @guangy10, @cccclai,
This PR should be able to fix some of the CI issues: including Mobilebert , W2L, and Llama.
For now, I am using 1 epoch for Mobilebert to expedite training process.
For W2L, I am using empty weights since we are just compiling the model. However, if weights can cause regression during inference, we should probably be using real weights in future.
Also, I notice that most of the example files are using only 1 random input when setting `args.compile_only`. This is probably not straight forward to community users, and I think we should either consider creating a new flag for CI or use actual data set when compiling the model.
Please let me know your thoughts. Thanks

winskuo-quic:
> There is an CI error, can you fix it?

Thanks for the catch.
Another thing we noticed is that we are using nightly environment to develop, while it seems like CI is not using nightly libraries. Think this can possibly cause CI to fail sometimes while our environment is working fine. For example: https://github.com/pytorch/executorch/issues/7634#issuecomment-2673191641 could possibly be caused by using different libraries version.

guangy10 (approved):

winskuo-quic:
> > > There is an CI error, can you fix it?
> > 
> > 
> > Thanks for the catch. Another thing we noticed is that we are using nightly environment to develop, while it seems like CI is not using nightly libraries. Think this can possibly cause CI to fail sometimes while our environment is working fine. For example: [#7634 (comment)](https://github.com/pytorch/executorch/issues/7634#issuecomment-2673191641) could possibly be caused by using different libraries version for `torchaudio`.
> 
> Hmm I think CI is using this version
> 
> https://github.com/pytorch/executorch/blob/65432b1a27e14f3d84d26f401ee94b5d426d913d/install_requirements.py#L70
> 
> maybe we need to bump the version more regularly.

@cccclai ,
Unsure if I traced this correctly, but seems like the CI starts off from
`setup-linux.sh`, which calls `install_executorch "use-pt-pinned-commit"`
and inside `install_executorch.py`, I saw the following code block that sets `use_pytorch_nightly = False`  if `args.use_pt_pinned_commit`, as shown below:
image
Therefore, it will not be using the nightly version but the released version.

cccclai:
@guangy10 @kirklandsign  Do you know which version we're running?

cccclai:
This PR needs to rebase...

cccclai:
Still failing..maybe it's related to https://github.com/pytorch/executorch/pull/8642. Can you rebase again?

cccclai:
Hmm looks like CI is failing..

PR #8624, opened 2025-02-21
Use to_edge_lower_and_transform for XNNPack

### LINKED ISSUE #8621: Move export_llama to to_edge_transform_and_lower API

Trying to export_llama to to_edge_transform_and_lower API. A few blockers from code that requires the intermediary exported to_edge model, curious if there are any workarounds:
- [QNN transforms](https://github.com/pytorch/executorch/blob/main/examples/models/llama/export_llama_lib.py#L762-L795) @cccclai 
- [Vulkan transforms](https://github.com/pytorch/executorch/blob/main/examples/models/llama/export_llama_lib.py#L733) @SS-JIA 
- [ET record generation](https://github.com/pytorch/executorch/blob/main/examples/models/llama/export_llama_lib.py#L812) @tarun292 

cc @mergennachin @iseeyuan @lucylq @helunwencser @tarun292 @kimishpatel

jackzhxng:
### Summary
Use `to_edge_transform_and_lower` in `export_llama` for XNNPack. As part of these changes, this also means that you cannot specify multiple backends in `export_llama` in the args, although I'm not sure if that is happening anywhere at the moment.

Closes https://github.com/pytorch/executorch/issues/8621

Performance regression benchmarking for xnnpack (on android) vs. past 3 days:
Screenshot 2025-02-24 at 11 39 52 AM

These benchmark numbers also normally fluctuate a bit across runs and these differences are within the usual fluctuation ranges.

### Test plan
See if CI passes

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -676,47 +676,62 @@ def _validate_args(args):
             )
 
 
-def _export_llama(args) -> LLMEdgeManager:  # noqa: C901
-    _validate_args(args)
-
-    pt2e_quant_params, quantizers, quant_dtype = get_quantizer_and_quant_params(args)
-
-    # export_to_edge
-    builder_exported = _prepare_for_llama_export(args).export()
-
-    builder_exported.run_canonical_optimizations()
-
-    if args.export_only:
-        exit()
-
-    builder_exported_to_edge = builder_exported.pt2e_quantize(
-        quantizers
-    ).export_to_edge()
-
-    modelname = builder_exported_to_edge.modelname
-
-    # to_backend
+def _to_edge_and_lower_llama_xnnpack(
+    builder_exported,
+    modelname,
+    additional_passes,
+    pt2e_quant_params,
+    quantizers,
+    quant_dtype,
+    args,
+) -> LLMEdgeManager:  # noqa: C901
     partitioners = []
 
     # Order matters here, dynamic quantization should be applied first when both xnnpack and xnnpack_extended_ops are enabled
-    if (
-        pt2e_quant_params is not None and pt2e_quant_params.quantize_linear is not None
-    ) or (args.xnnpack):
-        partitioners.append(
-            get_xnnpack_partitioner(dynamic_quant_only_partitioner=True)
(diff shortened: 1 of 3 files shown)


iseeyuan:
Have you tested the performance of a model (like llama 3B), before and after? Asking because export_llama is used by different users to prepare for the .pte files in cpu. Please make sure there's no perf regress.

examples/models/llama/export_llama_lib.py
```diff
@@ -674,47 +674,91 @@ def _validate_args(args):
...
+    )
+    if args.verbose:
+        print_delegation_info(builder.edge_manager.exported_program().graph_module)
+    if args.num_sharding > 0 and args.qnn:
+        from executorch.backends.qualcomm.utils.utils import canonicalize_program
+
+        # pyre-fixme[16]: Module `backends` has no attribute `qualcomm`.
+        canonicalize_program(builder.edge_manager.exported_program())
+
+    return builder.to_executorch(passes=additional_passes)
```
tarun292, line 720: Hmm there's a return in the middle of the function?

examples/models/llama/export_llama_lib.py
```diff
@@ -674,47 +674,91 @@ def _validate_args(args):
...
+        raise NotImplementedError(
+            f="export_llama does not support XNNPack and generating ETRecord at the moment. "
+        )
+
+    builder = builder_exported.pt2e_quantize(quantizers).to_edge_transform_and_lower(
+        partitioners
+    )
+    if args.verbose:
+        print_delegation_info(builder.edge_manager.exported_program().graph_module)
+    if args.num_sharding > 0 and args.qnn:
```
tarun292, line 714: This code shouldn't be here right?

jackzhxng: Oh yeah technically should remove since qnn. Will remove after benchmarking finishes

jackzhxng:
@iseeyuan please see the performance benchmark graph I posted in the pr description

mergennachin:
Can you run internal CI tests before merging? Otherwise looks good.

mergennachin (approved):

PR #8641, opened 2025-02-23
Add dtype, fix RMS norm for FP16

metascroy:
Llama1B quality in CoreML is bad due to FP16 arithmetic.  Here is a sample of generated text:

> "Once upon a time,we had our way,as we navigated,through the vast expanse of our understanding,as we journeyed,through the treacherous terrain of our experiences,as we faced our fears,as we stood tall,as we confronted our demons,as we emerged victorious,as we transcend our limits,as we ascend,as we unite,as we become,as we lose ourselves,as we search,as we remember,as we come back,as we return,as we find myself,as I am,as I am not,as I stand tall,as I hear my voice,as I remember,as I forgive,as I am, as I am, as I become, as I lose myself, as I find myself, as I remember, as I come back, as I return, as I find myself, as I am, as I am not, as I stand tall, as I Hear My Voice, as I Remember, as I Find Myself, as I am, as I Become, as I Lose Myself, as I Find Myself, asных, as I Become, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose Myself, as I Lose My"

The corresponding FP16 eager mode model has much better generated text:

> "Once upon a time, in a small village nestled between two great mountains, there lived a young girl named Akira. She was a curious and adventurous soul, with a heart full of wonder and a mind full of questions. Akira lived with her grandmother, a wise and kind woman named Kana, who taught her the ways of the world and the secrets of the universe.

> One day, while exploring the village, Akira stumbled upon a mysterious shop tucked away in a quiet alley. The sign above the door read "Moonlit Curios and Antiques," and the windows were filled with a dazzling array of strange and exotic objects. Akira felt an inexplicable pull towards the shop, as if the very fabric of the universe was calling to her."

The discrepancy is that the eager mode model actually computes the RMSNorm in FP32 due to a cast operation (which CoreML appears to ignore):
```
self._norm(x.float()).type_as(x)
```

Moreover, the norm computation appears unstable in FP16 and gives bad results.  We can improve the numeric quality of the norm in FP16 by first dividing x by its maximum absolute value.  Here is the generated text from CoreML in FP16 after this change:

> "Once upon a time, in a small village nestled in the rolling hills of Provence, there lived a young girl named Colette. Colette was a curious and adventurous soul, with a heart full of wonder and a mind full of questions. She spent most of her days exploring the village, visiting the local market, and listening to the tales of the old villagers.

> One day, while wandering through the village, Colette stumbled upon a small, mysterious shop tucked away on a quiet street. The sign above the door read "Curios and Wonders," and the windows were filled with a dazzling array of strange and exotic objects. Colette's curiosity was piqued, and she pushed open the door to reveal a dimly lit interior filled with the scent of old books and dust."  

Note, for 4-bit channelwise quantization, the results do not look good even after this change.  The ideal solution is to do QAT for llama1B with 4-bit channelwise quantization + FP16 arithmetic.

### DIFF
--- a/examples/apple/coreml/llama/llama_transformer.py
+++ b/examples/apple/coreml/llama/llama_transformer.py
@@ -23,29 +23,6 @@ from executorch.examples.models.llama.rope import (
 from torch import nn
 
 
-# These are just to prevent to_edge from decomposing SDPA
-# A better method is to use the to_edge_transform_and_lower API for CoreML
-# and not decompose SDPA
-@torch.library.custom_op("coreml::sdpa", mutates_args=())
-def sdpa(
-    q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, attn_mask: torch.Tensor
-) -> torch.Tensor:
-    """Same as F.scaled_dot_product_attention, but with custom op to avoid lowering during dialect conversion."""
-    return torch.ops.aten.scaled_dot_product_attention.default(
-        q, k, v, attn_mask=attn_mask
-    )
-
-
-@torch.library.register_fake("coreml::sdpa")
-def _(
-    q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, attn_mask: torch.Tensor
-) -> torch.Tensor:
-    """Fake implementation with the right output shape, which is required for torch.compile/export/fx tracing."""
-    expected_shape = list(q.shape)
-    expected_shape[-1] = v.shape[-1]
-    return q.new_empty(expected_shape)
-
-
 def find_multiple(n: int, k: int) -> int:
     if n % k == 0:
         return n
@@ -149,10 +126,15 @@ class RMSNorm(torch.nn.Module):
             torch.Tensor: The normalized tensor.
 
         """
-        x_max, _ = torch.abs(x).max(-1, keepdim=True)
-        x = x / x_max  # This makes the op more stable in FP16
-        eps = self.eps / (x_max * x_max)
-        return x * torch.rsqrt((x * x).mean(-1, keepdim=True) + eps)
+        # CoreML ignores casts to FP32, so existing implementation of RMSNorm was not stable
+        # We instead use (x * sqrt(n)) / norm(x, dim=-1)
+        # Using torch.norm and preserving this op in CoreML improves stability
+        # Note, we ignore eps, but could add it by using torch.norm(torch.concat(x, sqrt(n*eps))) in the denominator
+        # In future, we want to add CoreML support for the functional RMSNorm op
+        rms_norm_eps0 = (
+            x * torch.sqrt(torch.tensor(self.dim, dtype=x.dtype))
+        ) / torch.linalg.vector_norm(x, dim=-1, keepdim=True)
+        return rms_norm_eps0
 
     def forward(self, x):
         """
@@ -352,7 +334,9 @@ class Attention(nn.Module):
             k = k.repeat_interleave(self.n_rep, dim=1)
             v = v.repeat_interleave(self.n_rep, dim=1)
 
-        output = torch.ops.coreml.sdpa(q, k, v, attn_mask)
+        output = torch.ops.aten.scaled_dot_product_attention.default(
+            q, k, v, attn_mask=attn_mask
+        )
         output = output.transpose(1, 2).contiguous().view(bsz, seqlen, -1)
         output = self.wo(output)
         return output, new_k, new_v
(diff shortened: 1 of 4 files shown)


metascroy:
@YifanShenSZ It is probably a bug for CoreML to ignore the cast.  It was presumably added because FP16 arithmetic was not sufficient enough.

examples/apple/coreml/llama/llama_transformer.py
```diff
@@ -121,6 +119,56 @@ def __post_init__(self):
...
+
+        Args:
+            x (torch.Tensor): The input tensor.
+
+        Returns:
+            torch.Tensor: The normalized tensor.
+
+        """
+        x_max, _ = torch.abs(x).max(-1, keepdim=True)
+        x = x / x_max  # This makes the op more stable in FP16
```
swolchok, line 153: i'll leave review for somebody who is better at math, but I'll just note that it is not at all obvious to me that this does not change the result of the operation

examples/apple/coreml/llama/llama_transformer.py
```diff
@@ -121,6 +119,56 @@ def __post_init__(self):
             self.head_dim = self.dim // self.n_heads
 
 
+class RMSNorm(torch.nn.Module):
```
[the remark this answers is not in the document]
metascroy: I don't see it here: https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/frontend/torch/ops.py

examples/apple/coreml/llama/llama_transformer.py
metascroy: It shouldn't because both numerator/denominator are divided by same thing (x_max).  Because denominator is under square root, we divide by x_max**2 there.

YIWENX14: I think the function produces equal results before and after. But would it be a concern that if we get very small values of `x_max` and the result of `eps = self.eps / (x_max * x_max)` could overflow? Should we dynamically use `torch.finfo(x.dtype).eps` for different dtypes?

examples/apple/coreml/llama/llama_transformer.py
metascroy, line 122: In sync with CoreML team, we might try using https://pytorch.org/docs/stable/generated/torch.nn.functional.rms_norm.html and then right an CoreML op definition for it here: https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/frontend/torch/ops.py

@YifanShenSZ mentioned they have a fused norm op that could be used.

YifanShenSZ: Yes, there are 2 possibilities
1. (Simpler) As mentioned above, we have [torch.norm](https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/frontend/torch/ops.py#L625-L630) translation using Core ML fused [reduced_l2_norm](https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/frontend/torch/ops.py#L625-L630), so we may have a RMS norm torch source like `x / torch.norm(x)`
2. (Better) I think we can further fused to Core ML [l2_norm](https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/frontend/torch/ops.py#L625-L630), which may directly correspond to RMS norm? (with some restriction in Core ML, though, e.g. `x` must have `rank >= 3`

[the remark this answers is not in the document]
metascroy: I think this is a slightly different op when eps > 0, although I'm not sure how much it matters in practice.

RMSNorm would actually be something like x / torch.norm([x, sqrt(n * eps)])

examples/apple/coreml/llama/llama_transformer.py
kimishpatel: I dont fully follow why this rewrite should hold better for fp16? If you are normalizing by max value then I am presuming that rsqrt is suffering from precision loss of fp16? It is not at all clear

examples/apple/coreml/llama/llama_transformer.py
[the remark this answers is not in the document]
metascroy: I'll update to use norm, and then maybe we can work on a longer term solution of support rmsnorm in CoreML @YifanShenSZ?

cccclai (approved):
Looks good - maybe let's add CI to prevent regression

metascroy:
> Looks good - maybe let's add CI to prevent regression

There is currently no CI for this model.  Let me create a task for that.

YifanShenSZ:
Looks good! Maybe also test by setting temperature to 0?

PR #8646, opened 2025-02-24
Revert D68939461 'fix memory planning API to use run()'

abauserman:
Summary:
This diff reverts D68939461, "fix memory planning API to use run()" which was a fix for:
https://github.com/pytorch/executorch/pull/8622

Planning to redo this change but need to revert for now, as it breaks some of our virtual platform testing - multisected to 8 test fails in T216023551, and also larger set of failures tracked in T216134349.

Differential Revision: D70103236


mergennachin (approved):

abauserman:
ended up not having to revert this change after all (i.e. the original diff corresponding to this PR was abandoned)

PR #8653, opened 2025-02-24
etLLM: add options to apply embedding or output.

iseeyuan:
### Summary
Add options to apply embedding or output in llama_transformer. 

Currently the transformer's forward is applying embedding based on inputs (`tokens is not None and h is None`). However, if embedding is not applied, the embedding matrix should not be initialized to take extra memory. So add this option to control it. 

It's useful for models without embeddings, or the embeddings need to be done outside of transformer, or the embedding need to be done differently than the one inside transformer. 

Similar option is added to the output layer.

### Test plan


madhu-fb (approved):
LGTM!

larryliu0820 (approved):

PR #8661, opened 2025-02-25
Don't define aten-mode targets in OSS

swolchok:
Define get_aten_mode_options in runtime_wrapper and use it.

Test Plan: buck2 build //runtime/... has many fewer errors now (aten-mode targets were failing)


swolchok:
* __->__ #8661

larryliu0820 (approved):
Nice

PR #8672, opened 2025-02-25
Clean up a few more doc sections and links

GregoryComer:
### Summary
Clean up a few remaining placeholders, orphaned pages, and broken links.

### Test plan
Built docs locally to verify changes.

### DIFF
--- a/docs/source/backends-xnnpack.md
+++ b/docs/source/backends-xnnpack.md
@@ -51,9 +51,57 @@ The XNNPACK partitioner API allows for configuration of the model delegation to 
 
 ### Quantization
 
-Placeholder - document available quantization flows (PT2E + ao), schemes, and operators.
+The XNNPACK delegate can also be used as a backend to execute symmetrically quantized models. To quantize a PyTorch model for the XNNPACK backend, use the `XNNPACKQuantizer`. `Quantizers` are backend specific, which means the `XNNPACKQuantizer` is configured to quantize models to leverage the quantized operators offered by the XNNPACK Library. 
+
+### Configuring the XNNPACKQuantizer
+
+```python
+from executorch.backends.xnnpack.quantizer.xnnpack_quantizer import (
+  XNNPACKQuantizer,
+  get_symmetric_quantization_config,
+)
+quantizer = XNNPACKQuantizer()
+quantizer.set_global(get_symmetric_quantization_config())
+```
+Here, the `XNNPACKQuantizer` is configured for symmetric quantization, indicating that the quantized zero point is set to zero with `qmin = -127` and `qmax = 127`. `get_symmetric_quantization_config()` can be configured with the following arguments:
+* `is_per_channel`
+    * Weights are quantized across channels
+* `is_qat`
+    * Quantize aware training
+* `is_dynamic`
+    * Dynamic quantization
+
+```python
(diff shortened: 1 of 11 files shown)


docs/source/backends-xnnpack.md
```diff
@@ -51,19 +51,61 @@ The XNNPACK partitioner API allows for configuration of the model delegation to
...
+After `convert_pt2e`, the model can be exported and lowered using the normal ExecuTorch XNNPACK flow. For more information on PyTorch 2 quantization [here](https://pytorch.org/tutorials/prototype/pt2e_quant_ptq.html).
+
+### Testing the Model
+
+After generating the XNNPACK-delegated .pte, the model can be tested from Python using the ExecuTorch runtime python bindings. This can be used to sanity check the model and evaluate numerical accuracy. See [Testing the Model](using-executorch-export.md#testing-the-model) for more information.
 
 ## Runtime Integration
 
+To run the model on-device, use the standard ExecuTorch runtime APIs. See [Running on Device](getting-started.md#running-on-device) for more information.
+
```
mergennachin, line 105: In this section, as a follow-up, please remember to add an example CMake file and c++ example runner code that uses XNNPACK (our current xnnpack_runner is too tied our repo, so not that one)

On top of that, we can also add to pytorch-labs/executorch-examples to showcase how it is used from out-of-tree repo.

mergennachin (approved):

mergennachin:
Are the links fixed in https://github.com/pytorch/executorch/tree/executorch-just-works/examples/xnnpack?

I think bunch of links that has "stable/..." will stop working on the next release

I think it makes sense to from "stable/..." to "main/...", at least for the links you changed and modify the link name accordingly. There are a few 

https://github.com/search?q=repo%3Apytorch%2Fexecutorch%20stable%2F&type=code

GregoryComer:
> Are the links fixed in https://github.com/pytorch/executorch/tree/executorch-just-works/examples/xnnpack?
> 
> I think bunch of links that has "stable/..." will stop working on the next release
> 
> I think it makes sense to move from hard-coded "stable/..." prefix to relative path, at least for the links you changed and modify the link name accordingly. There are a few
> 
> https://github.com/search?q=repo%3Apytorch%2Fexecutorch%20stable%2F&type=code

Good point on the stable docs. Those don't show up as warnings in the doc build, which I've been relying on to catch broken links. I'll see if I can replace all of the stable links with relative links.

GregoryComer:
@mergennachin Thinking more about the stable links, this is a little tricky. Relative paths don't work as well, since the readmes are hosted in the repo, not in the docs. I did fix the remaining couple of stable links in the docs, though. If I change the readme links now, it will break the nightly docs. I'm kind of wondering if we either (a) fix them immediately prior to 0.6 release, or (b) set up redirects for the moved pages. I'm leaning towards (b).

I'm tracking here: https://github.com/pytorch/executorch/issues/8728, and I added to the 0.6 milestone. I'll figure out a solution before 0.6 release.

docs/source/backends-xnnpack.md
GregoryComer: I added the CMake example here: https://github.com/pytorch/executorch/pull/8730. I'll take the examples as the runner examples as a follow-up task.

PR #8679, opened 2025-02-25
Arm backend: Add initial Llama model test case

mansnils:
Adds Llama model test case for TOSA-0.80+MI.
Handles Add and Mul where inputs have different ranks. New unit test parameters --llama_inputs added, without it test will be skipped. Tested with smaller stories, see examples/models/llama/UTILS.md. Adds get_llama_model() to export_llama_lib used in test case.


mansnils:
topic: not user facing

mansnils:
topic: "not user facing"

backends/arm/test/models/test_llama_arm.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: I will take a look but just a nit
`s/test_llama_arm.py/test_llama.py/`

zingo:
MacOs fails are unrelated

backends/arm/test/models/test_llama_arm.py
zingo: > I will take a look ...
Merge time? :)

.ci/scripts/setup-arm-baremetal-tools.sh
```diff
@@ -9,3 +9,5 @@
 git config --global user.email "github_executorch@arm.com"
 git config --global user.name "Github Executorch"
 bash examples/arm/setup.sh --i-agree-to-the-contained-eula
+
+./examples/models/llama3_2_vision/install_requirements.sh
```
digantdesai, line 13: is this the best place to put this?

backends/arm/operators/op_add.py
```diff
@@ -45,6 +45,12 @@ def define_node(
         # Handle int8 (quantized) and int32
         assert inputs[0].dtype in [ts.DType.INT8, ts.DType.INT32]
 
+        dim_order = (
```
digantdesai, line 48: I guess we already assumed that the type of dim-order (and also rank) is same earlier.
By broadcasting to a larger dim-order helps with different rank. But for completeness, do you want to add an assert to make sure the dim-orders are of same type?

backends/arm/operators/op_slice.py
```diff
@@ -32,9 +32,11 @@ def define_node(
         output: TosaArg,
     ) -> None:
 
+        # See slice_copy_support.py
+        assert len(inputs) == 4 or (len(inputs) == 5 and inputs[4].number == 1)
```
digantdesai, line 36: Nit: add an error message

backends/arm/test/conftest.py
```diff
@@ -43,7 +43,7 @@ def pytest_configure(config):
             )
         # Only enable if we also have the TOSA reference model available.
         pytest._test_options["corstone_fvp"] = True  # type: ignore[attr-defined]
-
+    pytest._test_options["llama_inputs"] = config.option.llama_inputs  # type: ignore[attr-defined]
```
digantdesai, line 46: Not sure if this is general enough to go through the pytest test_options. Any other options?

backends/arm/test/models/test_llama.py
```diff
@@ -0,0 +1,128 @@
...
+class TestLlama(unittest.TestCase):
+    """
+    Test class of Llama models. Type of Llama model depends on command line parameters:
+    --llama_inputs <path to .pt file> <path to json file>
+    Example: --llama_inputs stories110M/stories110M.pt stories110M/params.json
+    """
+
+    _edge_compile_config: EdgeCompileConfig = EdgeCompileConfig(
+        _check_ir_validity=False,
+        _skip_dim_order=True,  # TODO(T182928844): Delegate dim order op to backend.
```
digantdesai, line 43: why is this needed?

backends/arm/test/models/test_llama.py
```diff
@@ -0,0 +1,128 @@
...
+                    constant_methods=llama_meta,
+                )
+                .export()
+                .to_edge_transform_and_lower(
+                    edge_compile_config=self._edge_compile_config
+                )
+                .check_count({"torch.ops.higher_order.executorch_call_delegate": 14})
+                .to_executorch()
+                .run_method_and_compare_outputs(
+                    inputs=llama_inputs, atol=1.8, rtol=0.01
```
digantdesai, line 126: how did we come up with atol? Also this is running on MI + FVP w/o quantization, right?

.ci/scripts/setup-arm-baremetal-tools.sh
zingo: The idea was to have it in place before the testing but thinking about it now I think you are right 
Having it as part of all the steps that use pytest in the file
`backends/arm/test/test_arm_baremetal.sh`
(or create a prepare test step)
Might be better and easier for anyone to just run the test and they will install needed dependencies them self.

mansnils: Done

backends/arm/test/conftest.py
mansnils: I don't know of an alternative at the moment, do you have any suggestions?

backends/arm/test/models/test_llama.py
mansnils: It should not be needed. Removed.

mansnils:
ping @digantdesai

shoumikhin:
@digantdesai how do we want to proceed with this?

backends/arm/test/conftest.py
mansnils: Is --model-inputs better perhaps?

kimishpatel:
> @digantdesai how do we want to proceed with this?

Digant is on pat leave right now, so expect delays

digantdesai:
LGTM.

digantdesai (approved):

zingo:
MacOS fails unrelated

zingo:
just det den, jag vet faktist inte

kirklandsign:
Sorry @zingo we probably need to revert this and import and run internal tests before landing.

executorch/backends/arm/test:cast_int64_pass - main
```
executorch/backends/arm/test/conftest.py:47: in pytest_configure
    pytest._test_options["llama_inputs"] = config.option.llama_inputs  # type: ignore[attr-defined]
E   AttributeError: 'Namespace' object has no attribute 'llama_inputs'
```

This line:
```
pytest._test_options["llama_inputs"] = config.option.llama_inputs  # type: ignore[attr-defined]
```
probably need to use getattr to check?

larryliu0820:
@kirklandsign @digantdesai  is it possible to add the tests into `unittest-buck` CI job?

kirklandsign:
> @kirklandsign @digantdesai is it possible to add the tests into `unittest-buck` CI job?

Oh are these tests triggered on GH right now? Seems like some config missing issue?

larryliu0820:
> > @kirklandsign @digantdesai is it possible to add the tests into `unittest-buck` CI job?
> 
> Oh are these tests triggered on GH right now? Seems like some config missing issue?

Created #9476 to track this.

PR #8680, opened 2025-02-25
Arm backend: Add support for DevTools BundleIO

zingo:
Add support for input and output testing using BundleIO in devtools.

Can be tested with:
run.sh --model_name=add --target=ethos-u85-128 --bundleio

aot_arm_compiler.py changes:
--bundleio      - Will generate a bundle .bpte file with input and reference output data
--model_name    - now also support .pt/.pth files e.g. a torch.save():ed model
--model_input   - If supplied will overried the model input data supply a torch.save():ed data
--intermediates - Will now save model.pth, input_xx.pt and output_ref_xx.pt files so testing can be reproduced

arm_executor_runner.cpp - can be build with ET_BUNDLE_IO to enable a .bpte and a input/output_ref testing flow

test_model.py script uses bundleio to verify input/output

test_arm_baremetal.sh script now tests models with BundleIO so output is compared to ref output


zingo:
Failed macos test is unrelated

oscarandersson8218 (approved):

PR #8687, opened 2025-02-25
Fix pyre error for logits

iseeyuan:
Summary: Fix the pyre error when logits is not initialized

Reviewed By: swolchok

Differential Revision: D70183946


jackzhxng (approved):
Trusting @swolchok's internal review

PR #8689, opened 2025-02-25
Update CadenceMemoryPlanning to support per-memory alignment constraint

sabarishsnk:
Summary: Currently `CadenceMemoryPlanning` only supports a single `mem_alignment` parameter that it applies to all memories in the hierarchy. This diff moves the alignment constraint to the `MemoryConfig` definition and makes it per-memory.

Reviewed By: dulinriley

Differential Revision: D69881079


dulinriley (approved):

zonglinpeng (approved):
could you run

```
pip install lintrunner==0.12.7
pip install lintrunner-adapters==0.12.4
lintrunner init
lintrunner -a
```
I can help doing it if pip is not working

Issue #8684, opened 2025-02-25
Failed to build installable wheels for some pyproject.toml based projects (executorch)


ManojShanmugamTruminds:
I have followed https://github.com/pytorch/executorch/blob/main/examples/demo-apps/apple_ios/LLaMA/docs/delegates/mps_README.md link to config executorch but getting error after passing below command in my MacBook Pro.
OS version 15.3.1 
Apple M1 chip
Command : 
./install_executorch.sh 

Error while generating /Users/manojshanmugam/executorch/pip-out/temp.macosx-11.1-arm64-cpython-310/cmake-out/executorch_srcs.cmake. Exit code: 1
  Output:

  Error:
  Traceback (most recent call last):
    File "/Users/manojshanmugam/executorch/build/extract_sources.py", line 255, in <module>
      main()
    File "/Users/manojshanmugam/executorch/build/extract_sources.py", line 240, in main
      target_to_srcs[name] = sorted(target.get_sources(graph, runner, buck_args))
    File "/Users/manojshanmugam/executorch/build/extract_sources.py", line 132, in get_sources
      sources: set[str] = set(runner.run(["cquery", query] + buck_args))
    File "/Users/manojshanmugam/executorch/build/buck_util.py", line 26, in run
      cp: subprocess.CompletedProcess = subprocess.run(
    File "/opt/anaconda3/envs/et_xnnpack/lib/python3.10/subprocess.py", line 501, in run
      with Popen(*popenargs, **kwargs) as process:
    File "/opt/anaconda3/envs/et_xnnpack/lib/python3.10/subprocess.py", line 966, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/opt/anaconda3/envs/et_xnnpack/lib/python3.10/subprocess.py", line 1842, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'buck2'

  CMake Error at build/Utils.cmake:236 (message):
    executorch: source list generation failed
  Call Stack (most recent call first):
    CMakeLists.txt:403 (extract_sources)

  error: subprocess-exited-with-error
  
  × Building wheel for executorch (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
[...]

PR #8692, opened 2025-02-25
Check in collect_env.py to improve UX of GitHub Issue reporting

guangy10:
### Summary
During my oncall shift, I noticed that not just one user forget to include the dev env when report GitHub issues to us ([issue1](https://github.com/pytorch/executorch/issues/8684), [issue2](https://github.com/pytorch/executorch/issues/8665)). And they are not aware of that the script is not included in the repo by default and they have to download it separately (https://github.com/pytorch/executorch/issues/8684#issuecomment-2682971982).

I think a better UX, to make it easier for both us and users, we should make the process easier by including the script in our repo. So in this PR, we check in the `collect_env.py` to `util/` and update the guide in issue template.

### Test plan
`python util/collect_env.py` locally

### DIFF
--- a/util/collect_env.py
+++ b/util/collect_env.py
@@ -1,8 +1,15 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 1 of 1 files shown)


guangy10:
Fixed linter

mergennachin (approved):

PR #8694, opened 2025-02-25
increase timeout for test-models-linux to debug timeout issue

wdvr:
### Summary
Debugging timeout issue - increasing for now to see impact


swolchok:
don't think we need this now that the job is gone; taking the liberty of closing it for your convenience

PR #8703, opened 2025-02-25
Remove tombstone messages proactively

bigfootjon:
Summary:
Originally we created a bunch of empty `TARGETS` files to allow us to enable `BUCK` files in fbcode by hiding the existing BUCK file. These files were subsequently merged together using `non_fbcode_target` so these tombstones are no longer necessary.

This diff fixes all files that WOULD have had the useless tombstone merged into them. To create this diff, I just ran the merger script that Codemod Service is using and then deleted the "merged from" and tombstone lines with `sed`, `arc f` and reverted any lines that didn't make sense

Differential Revision: D69994481


examples/demo-apps/android/ExecuTorchDemo/app/src/main/BUCK
```diff
@@ -1,9 +1,10 @@
+load("@fbcode_macros//build_defs:build_file_migration.bzl", "fbcode_target", "non_fbcode_target")
```
swolchok, line 1: I don't love that this file is not available in the open-source shim. however, they are not building in OSS today, so I won't block.

swolchok (approved):

PR #8709, opened 2025-02-25
introduce sanity check when creating bufferdatasink

Gasoonjia:
Summary: This diff introduces sanity check on data alignment when creating buffer_data_sink. Also, for better control the raised error, we changed to use static function instead of constructor for constuction.

Differential Revision: D70190912

### DIFF
--- a/devtools/etdump/tests/buffer_data_sink_test.cpp
+++ b/devtools/etdump/tests/buffer_data_sink_test.cpp
@@ -16,6 +16,7 @@
 using namespace ::testing;
 using ::executorch::aten::ScalarType;
 using ::executorch::aten::Tensor;
+using ::executorch::etdump::BufferDataSink;
 using ::executorch::runtime::Error;
 using ::executorch::runtime::Result;
 using ::executorch::runtime::Span;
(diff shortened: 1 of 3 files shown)


swolchok (approved):
seems reasonable

devtools/etdump/tests/buffer_data_sink_test.cpp
```diff
@@ -39,11 +39,13 @@ class BufferDataSinkTest : public ::testing::Test {
   size_t buffer_size_;
   void* buffer_ptr_;
   Span<uint8_t> buffer_;
-  std::unique_ptr<executorch::etdump::BufferDataSink> data_sink_;
 };
 
 TEST_F(BufferDataSinkTest, StorageSizeCheck) {
-  Result<size_t> ret = data_sink_->get_storage_size();
+  Result<BufferDataSink> buffer_data_sink = BufferDataSink::create(buffer_);
+  ASSERT_EQ(buffer_data_sink.error(), Error::Ok);
```
dbort, line 46: It'd be good to put this same ASSERT after all `create` calls in this file

dbort (approved):
This is a great change; a static factory is a much better way to handle creation in our exception-free system.

devtools/etdump/tests/buffer_data_sink_test.cpp
Gasoonjia: Thanks for the suggestion! Making `buffer_data_sink` as a private variable in `BufferDataSinkTest` class, and putting ASSERT after creations, for better structure.

PR #8715, opened 2025-02-26
Add support for splitting in_features in linear layers

metascroy:
This PR:
* Moves SplitLinearModule to a separate utils.py and extends it to support splitting in_features.  I did not find splitting in_features worked very well, at least in my initial experiments.
* Updates readme with latest setting recommendations for Llama1B


metascroy:
@cccclai can I get a review on this?

examples/apple/coreml/llama/utils.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+
+class SplitLinearModule(torch.nn.Module):
```
cccclai, line 10: I might miss some context - I thought you said splitting linear speeds up the perf in ANE?

cccclai: Oh, you said in features, in the description. What did you split to get better perf?

YifanShenSZ (approved):

examples/apple/coreml/llama/utils.py
metascroy: Yes, so currently landed we have ability to split up out_features in linear modules and this speeds up perf by out 30%.

This diff adds the ability to split in_features in linear modules.  It did not speed up perf in the few parameters I tried, but it's nice to have this ability it landed so people can experiment further.

cccclai (approved):
Looks good, thank you!

PR #8720, opened 2025-02-26
Remove unnecessary field in optional

barakugav:
### Summary

Remove an unnecessary field in optional.
Before this PR, `optional<uint64 *>` was `24` bytes, after this PR its `16`.

cc @JacobSzwejbka


dbort:
Great catch! Thank you for finding and fixing this.

PR #8702, opened 2025-02-25
remove llava from test-models-linux

swolchok:
See comment for explanation -- already have test-llava-runner-linux

PR #8721, opened 2025-02-26
Debug llava build timeout

huydhn:
Just a debug PR, no need to review


huydhn:
Fixed by https://github.com/pytorch/executorch/pull/8702

PR #8738, opened 2025-02-26
add backend is available

cccclai:
Summary: Add a pybind API to see if the backend is available

Differential Revision: D69810445

### DIFF
--- a/extension/pybindings/pybindings.cpp
+++ b/extension/pybindings/pybindings.cpp
@@ -88,10 +88,12 @@ using ::executorch::extension::BufferDataLoader;
 using ::executorch::extension::MallocMemoryAllocator;
 using ::executorch::extension::MmapDataLoader;
 using ::executorch::runtime::ArrayRef;
+using ::executorch::runtime::BackendInterface;
 using ::executorch::runtime::DataLoader;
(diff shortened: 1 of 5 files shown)


extension/pybindings/pybindings.cpp
```diff
@@ -990,6 +992,20 @@ py::list get_registered_backend_names() {
...
+  size_t n_of_registered_backends = get_num_registered_backends();
+  for (size_t i = 0; i < n_of_registered_backends; i++) {
+    auto backend_name_res = get_backend_name(i);
+    THROW_IF_ERROR(backend_name_res.error(), "Failed to get backend name");
+    auto backend_name = backend_name_res.get();
+    if(strcmp(backend_name, target_backend_name.c_str()) == 0) {
+      BackendInterface* backend = get_backend_class(backend_name);
+      return backend->is_available();
+    }
+  }
```
mergennachin, line 1005: No need to iterate here, just call `get_backend_class(target_backend_end)`, which does the iteration.

cccclai: Ah yeah, updated.

mergennachin (approved):

Issue #8261, opened 2025-02-06
Standalone AAR build script


GregoryComer:
### 🚀 The feature, motivation and pitch

We currently provide pre-built Android AARs for download. It would be desirable to have a standalone build script (maybe build_android_libraries.sh?) that can reproduce the binaries with one line. The use case is for users to be able to easily switch to building from source if they need to change the configuration relative to the published binaries (adding backends/extensions, changing compiler flags, etc.). The current AAR build logic is under build/build_android_library.sh. We can likely split it out into a dedicate script.

cc @kirklandsign @mergennachin @byjlw

GregoryComer:
[#8747](https://github.com/pytorch/executorch/pull/8747) renames the existing script to build_android_library.sh. We can document the usage under using-executorch-building-from-source.md and using-executorch-android.md.

mergennachin:
@kirklandsign don't we need to update this? 

https://pytorch.org/executorch/main/using-executorch-building-from-source.html#android

kirklandsign:
> [@kirklandsign](https://github.com/kirklandsign) don't we need to update this?
> 
> https://pytorch.org/executorch/main/using-executorch-building-from-source.html#android

Yes. I need to update this as well.

PR #8747, opened 2025-02-26
[Android] Rename build script

kirklandsign:
### Summary
build_android_llm_demo.sh -> build_android_library.sh
It's not only for LLM demo, but for general use case as well

### Test plan
CI


GregoryComer:
Thanks! We'll also want to document usage under using-executorch-android.md and using-executorch-building-from-source as a follow-up.

GregoryComer (approved):

kirklandsign:
> Thanks! We'll also probably want to document usage under using-executorch-android.md and using-executorch-building-from-source as a follow-up.

Thank you. Mind sharing the PR or issue number so we can remind ourself later?

GregoryComer:
> > Thanks! We'll also probably want to document usage under using-executorch-android.md and using-executorch-building-from-source as a follow-up.
> 
> Thank you. Mind sharing the PR or issue number so we can remind ourself later?

Sure. We can track in https://github.com/pytorch/executorch/issues/8261.

PR #8755, opened 2025-02-27
portable arg{max,min}: optimize update check

swolchok:
We don't need a second isnan; see code comment. (This is a small optimization.)


swolchok:
* (to be filled)

swolchok:
messed up my stack, resubmitting

PR #8757, opened 2025-02-27
Enable quantized cat

mcremon-meta:
Summary: As titled. Use it only in the WakeWord quantizer for now, because it has implications on the numerics in general.

Reviewed By: zonglinpeng

Differential Revision: D69499329

### DIFF
--- a/backends/cadence/aot/quantizer/fusion_pass.py
+++ b/backends/cadence/aot/quantizer/fusion_pass.py
@@ -13,6 +13,7 @@ from executorch.backends.cadence.aot.quantizer.patterns import (
     AddmmPattern,
     AddPattern,
     BmmPattern,
+    CatPattern,
     Conv1dPattern,
     Conv2dPattern,
     LayerNormPattern,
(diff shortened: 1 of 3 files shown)


zonglinpeng (approved):

tarun292 (approved):

mcremon-meta:
@dbort @mergennachin @tarun292  have you seen linter errors like this one? It doesn't seem suggest a change, and linting on my own clone of ET (native mac) is clean. Not sure how to proceed.

mergennachin:
@mcremon-meta when you run locally, do you have the same lintrunner version as the CI? 

https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

mcremon-meta:
> @mcremon-meta when you run locally, do you have the same lintrunner version as the CI?
> 
> https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

yes I did those exact steps this morning, so should be the same

mcremon-meta:
might have been trailing spaces, let's see if new version is happy. If it's that it'd be nice if the linter message could say that, since the spaces don't show on the diff view (even when I highlight)

PR #8760, opened 2025-02-27
Neuron buffer allocator decouple from ExecuTorch framework

### LINKED ISSUE #6074: Unsafe prebuilt boundary in MediaTek backend
### 🐛 Describe the bug

While making some changes to backends/mediatek, I noticed:
- It [defines a BufferAllocator class](https://github.com/pytorch/executorch/blob/03e451647e20ddcb5a5557435a9715618d890045/backends/mediatek/runtime/include/NeuronBufferAllocator.h#L105) that inherits from ET's [MemoryAllocator](https://github.com/pytorch/executorch/blob/main/runtime/core/memory_allocator.h) class
- The .cpp code for that subclass does not seem to appear in the repo, so I think it may be part of the prebuilt binaries supplied by MediaTek

ET does not make any ABI compatibility promises. If the MT libraries really do contain a prebuilt version of a subclass of an ET type, ET could legally make changes to MemoryAllocator that would break that prebuilt code, possibly in subtle or dangerous ways.

For example, if we made a change that reordered the vtable, a call like this could end up calling a different method, which wouldn't expect the same parameters or register values:
```
mediatek::BufferAllocator ba;  // points to a vtable with an old layout
executorch::runtime::MemoryAllocator *ma = &ba;
ma->allocate(...);  // expects a new vtable layout
```
In the best case this would immediately crash, but in the worst case it would silently corrupt memory or execute random code.

Things would also go bad if we added a new field to the base class, because it could alias to an offset used by a subclass field.

At a higher level, it is unsafe to have prebuilt code that refers to any ExecuTorch type, unless it's guaranteed that the prebuilt code will link against a very specific (i.e., a single git hash) version of the ET runtime.

## Solution

The prebuilt code should not refer to any ExecuTorch types.

## Mitigation/detection

There's no way to reliably check these sorts of things; it's just plain unsafe.

But as a first-order check, the backend could compare `sizeof(MemoryAllocator)` across the prebuilt and non-prebuilt boundaries. E.g., it could have a function like `size_t get_prebuilt_memory_allocator_size()` whose implementation is inside the prebuilt, and returns `sizeof(MemoryAllocator)` (which will be determined at prebuild time). Then some code in a non-prebuilt .cpp file could `assert sizeof(MemoryAllocator) == get_prebuilt_memory_allocator_size()` before constructing a `BufferAllocator`, and return an error if the sizes don't match.

But this check could be defeated if the base class's fields were reordered in a way that did not affect the size of the type. It also wouldn't catch the re-ordering, addition, or removal of a virtual method, since the vtable isn't part of the size of the type.

### Versions

Issue is present in main as of at least https://github.com/pytorch/executorch/commit/afb26644c1523c14a92844a7f6c874458f35a7c4 on 2024-10-09

### LINKED ISSUE #7549: mediatek backend.so build failed
### 🐛 Describe the bug

I have followed the steps in the backend mediatek and demo app example mediatek, installed the relevant dependency libraries and environment variables, and used ndk 26.3.11579264. However, when running   ./mtk_build.sh   or   ./mtk_build_examples.sh   to build the mtk backend, I encountered the problem shown in the figure.
build error
mtk_ndk_build_failed

### Versions

0.5.0 and 0.4.0

cc @cccclai @neuropilot-captain @cbilgin

### LINKED ISSUE #10389: Undefined symbol while compiling runtime for MediaTek
Hello, during compilation for MediaTek I'm receiving following issue:

```
ld.lld: error: undefined symbol: torch::executor::neuron::BufferAllocator::Find(void*)
>>> referenced by NeuronBackend.cpp:132 (/Users/nono/dev/executorch/backends/mediatek/runtime/NeuronBackend.cpp:132)
>>>               CMakeFiles/neuron_backend.dir/runtime/NeuronBackend.cpp.o:(executorch::backends::neuron::NeuronExecuTorchDelegate::execute(executorch::runtime::BackendExecutionContext&, executorch::runtime::EValue**) const)
>>> referenced by NeuronBackend.cpp:150 (/Users/nono/dev/executorch/backends/mediatek/runtime/NeuronBackend.cpp:150)
>>>               CMakeFiles/neuron_backend.dir/runtime/NeuronBackend.cpp.o:(executorch::backends::neuron::NeuronExecuTorchDelegate::execute(executorch::runtime::BackendExecutionContext&, executorch::runtime::EValue**) const)
>>> referenced by NeuronBackend.cpp:175 (/Users/nono/dev/executorch/backends/mediatek/runtime/NeuronBackend.cpp:175)
>>>               CMakeFiles/neuron_backend.dir/runtime/NeuronBackend.cpp.o:(executorch::backends::neuron::NeuronExecuTorchDelegate::HintNeuronBackend(executorch::runtime::EValue**) const)
>>> referenced 1 more times

ld.lld: error: undefined symbol: torch::executor::neuron::BufferAllocator::GetInstance()
>>> referenced by NeuronBackend.cpp:168 (/Users/nono/dev/executorch/backends/mediatek/runtime/NeuronBackend.cpp:168)
```

I did all steps as described in the README. Is there any problem I miss? Thank you for help!

cc @cccclai @neuropilot-captain @cbilgin

neuropilot-captain:
### Summary
1. Decouple out Neuron buffer allocator as a standalone shared library
2. Refactor and update the build scripts
3. Pending the release of the new buffer allocator lib

Fixes #6074 
Fixes #7549 
Fixes #10389 

cc @cccclai @cbilgin


neuropilot-captain:
Hi @cccclai, this PR is ready for review! Thanks!

cccclai:
Hey can you rebase? There seems to be CI failure

cccclai (approved):
Thank you!

cccclai:
Hey seems like https://github.com/pytorch/executorch/pull/11125 is merged and causing merge conflict, can you resolve the conflict so we can merge it?

PR #8765, opened 2025-02-27
Arm backend: Remove N=1 constraint for MaxPool2d on U55

Erik-Lundell:
This is not needed anymore after Vela bump.

### DIFF
--- a/backends/arm/operator_support/pool_2d_support.py
+++ b/backends/arm/operator_support/pool_2d_support.py
@@ -26,8 +26,8 @@ def stride_check(strides: tuple[int, int]) -> bool:
 
 
 def dim_check(shape=torch.Size) -> bool:
-    check = shape[0] == 1
-    for dim in shape:
+    check = True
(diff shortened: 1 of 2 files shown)


zingo:
Failed MacOS tests are unrelated

zingo (approved):

backends/arm/operator_support/pool_2d_support.py
```diff
@@ -26,8 +26,8 @@ def stride_check(strides: tuple[int, int]) -> bool:
 
 
 def dim_check(shape=torch.Size) -> bool:
-    check = shape[0] == 1
-    for dim in shape:
+    check = True
+    for dim in shape[1:]:
         check &= 1 <= dim <= 65536
     return check
```
swolchok, line 32: JFYI this function could be written more idiomatically as `return all([1 <= dim <= 65536 for dim in shape[1:]])`

PR #8775, opened 2025-02-27
fix a missed get_aten_mode_options spot

swolchok:
This was the only spot that didn't use a tuple, so my find/replace missed it.


swolchok:
* (to be filled)

mergennachin (approved):

PR #8776, opened 2025-02-27
Fixes to_edge_transform_and_lower when unsupported ops are asked for preservation

metascroy:
If a partitioner requests to_edge_transform_and_lower keep mutable / aliasing ops (e.g., transpose, view, permute, etc), lowering with ExecuTorch fails because those ops cannot be functionalized when wrapped in the EDGE_DO_NOT_DECOMP namespace as custom ops.

This PR filters out unsupported ops that backends request for preservation.


exir/program/_program.py
```diff
@@ -965,6 +969,21 @@ def _sanity_check_graph_for_non_decomp_ops(
...
+    def keep(op):
+        schema = op._schema
+        native_schema = _pybind_schema_to_native_schema(schema)
+        if native_schema.is_mutable:
+            return False
+        if native_schema.aliased_return_names() != [None]:
+            return False
+        return True
+
+    return list(filter(keep, ops_to_not_decompose))
```
larryliu0820, line 984: Hmm this is a bit too sneaky in my mind. If the user wants to debug why an op is decomposed vs not decomposed, it might take them a while to find this filter.

Do you think it's reasonable to raise error at the partitioner level when they try to keep the aliasing op instead of silently filtering them out?

metascroy: > Do you think it's reasonable to raise error at the partitioner level

The user does not specify ops that should be decomposed or not.  The partitioner automatically returns the list based on what the backend supports.  If we did this filtering in the partitioner, each partitioner would just need to repeat this same filtering logic because it's actually a limitation with ExecuTorch.  In fact, I originally did add this logic to the CoreML partitioner, but then I thought it applies to all partitioners, so why not add it to _program.py instead.

We could add logging for this filter though?

I think the ideal solution for to_edge_transform_and_lower is:

1) First functionalize the graph and remove aliased ops
2) Ask partitioners for ops to preserve
3) Run decompositions

Today, step 1 happens after step 2 and that is the issue.  Note that doing the functionalization and removing aliased ops before asking partitioners what ops to preserve has the same effect as the filter above.

tarun292: @metascroy i remember vaguely that there was a way to functionalize the graph and remove aliased ops without calling `run_decompositions`. @tugsbayasgalan @angelayi is this possible?

larryliu0820: > I think the ideal solution for to_edge_transform_and_lower is:
> 
> 1. First functionalize the graph and remove aliased ops
> 2. Ask partitioners for ops to preserve
> 3. Run decompositions
> 
> Today, step 1 happens after step 2 and that is the issue. Note that doing the functionalization and removing aliased ops before asking partitioners what ops to preserve has the same effect as the filter above.

100% agree and this should be the long term fix. Can we create an issue to track that? We can go with the current approach to unblock ourselves.

exir/program/_program.py
```diff
@@ -965,6 +969,46 @@ def _sanity_check_graph_for_non_decomp_ops(
...
+                f"Op {op} was requested for preservation by partitioner.  This request is ignored because it aliases output."
+            )
+            return False
+
+        # Explicit block list of ops that don't work if asked for
+        # preservation
+        if op in [
+            # Hits infinte recursion error when op is in
+            # EDGE_DO_NOT_DECOMP namespace
+            torch.ops.aten._to_copy.default,
```
mcr229, line 995: perhaps I didn't understand the full context, but why can't a partitioner just not specify these ops for preservation? I don't think they'll be decomposed anyways?

metascroy: Because if a partitioner requests these for preservation, they are no longer aten ops (they become ops in EDGE_DO_NOT_DECOMP), and that runs into issues in during export.

exir/program/_program.py
metascroy: Created tracking issue here: https://github.com/pytorch/executorch/issues/8781

metascroy: It looks like I need to expand the blocklist to include contiguous as well, otherwise I can see complaints of view on improperly strided tensor (x.transpose(0,1).contigous().view(15, -1), the contiguous op gets put in DO_NOT_DECOMPOSE namespace and doesn't compose well with PyTorch).

I'm also wondering about ops like torch._check as well, which will be put into the DO_NOT_DECOMPOSE namespace.

We can continue expanding the blocklist over time, but this whole approach to to_edge_transform_and_lower of turning aten ops into custom ops in a DO_NOT_DECOMPOSE namespace is looking very brittle and does not compose well with PyTorch.

Any other solution ideas here @mcr229  @larryliu0820 @tarun292 ?

metascroy: @larryliu0820 are there any more concerns on this short term fix?

mcr229 (approved):
seems ok for now, gonna accept, but do you mind linking the issue # to the _remove_invalid_ops_for_not_decompose function?

metascroy:
> seems ok for now, gonna accept, but do you mind linking the issue # to the _remove_invalid_ops_for_not_decompose function?

Added

exir/program/_program.py
larryliu0820: No thanks for creating the issue

PR #8787, opened 2025-02-27
[ExecuTorch][XNNPACK] Don't partition per_tensor weights with qd8

digantdesai:
* #8844
* __->__ #8787

This is not supported, so we shouldn't partition it.

Add an expectedFailure test to indicate that this is not supported.

Differential Revision: [D70343584](https://our.internmc.facebook.com/intern/diff/D70343584/)


mcr229 (approved):
nice change, i didn't add these types of checks when i first added this because it was going to be too tedious along with all the other refactoring i had to do, so I kind of left this as a todo, when users encounter it.

digantdesai:
> nice change, i didn't add these types of checks when i first added this because it was going to be too tedious along with all the other refactoring i had to do, so I kind of left this as a todo, when users encounter it.

No worries. I started to feel gemm_config needs a refactor. It is too complex now. 😃

mcr229:
@digantdesai perhaps we can add a util function that goes through all the quant/dequant nodes in the partition, and validates those. And we can call that for every quantized op

PR #8820, opened 2025-02-28
Add qk norm optionally before attention calculation

madhu-fb:
Differential Revision: D70355802


iseeyuan (approved):
LGTM. Thank you @madhu-fb !

PR #8821, opened 2025-02-28
Arm backend: Remove output buffering when running Corstone FVP with tee

zingo:
This improves the log output feedback when running the Corstone FVP so you don't have to wait for long taking models without any output.


digantdesai (approved):

zingo:
MacOs job fail unrelated

PR #8825, opened 2025-02-28
Buck changes for ObjC/Swift bindings.

shoumikhin:
Summary: Add resources for the tests.

Differential Revision: D70377502


mergennachin (approved):

kirklandsign:
From my understanding, we want to export a model every time we run the test, unless we are doing BC/FC check.

shoumikhin:
@kirklandsign ptal at https://github.com/pytorch/executorch/tree/main/extension/module/test/resources, we kinda replicate the same setup here

Issue #8360, opened 2025-02-11
[RFC] Objective-C/Swift APIs for ExecuTorch


shoumikhin:
### 🚀 The feature, motivation and pitch

**Problem**

Currently, ExecuTorch provides a [C++ API](https://pytorch.org/executorch/0.5/extension-module.html) for running exported models on Apple platforms.  While functional, this presents a significant hurdle for developers working primarily in Objective-C and Swift.  The [documentation](https://pytorch.org/executorch/main/apple-runtime.html) suggests bridging the C++ APIs manually, which is complex, error-prone, and introduces a significant maintenance burden.  It also requires deep knowledge of both C++ and the intricacies of ExecuTorch's internals.  This negatively impacts developer productivity and increases the barrier to entry for using ExecuTorch within the Apple ecosystem. The only currently supported API in Objective-C/Swift are the [logging](https://pytorch.org/executorch/0.5/apple-runtime.html#logging) functions.

**Motivation**

We aim to make ExecuTorch easily accessible to the vast majority of iOS, macOS, and other Apple platform developers who work primarily in Objective-C and Swift.  A native API will significantly improve the developer experience, making integration simpler, safer, and more efficient.

**Pitch**

We propose creating a set of native Objective-C wrappers around the core ExecuTorch C++ APIs. These wrappers will be designed for seamless bridging into Swift, providing a natural and idiomatic experience for Swift developers.  This will drastically simplify the process of loading and executing ExecuTorch models within Apple platform apps. The user experience will be similar to the current one, but offering the native API: users will still integrate ExecuTorch using Swift Package Manager, import the `ExecuTorch` module, but will now have access to classes like `Module`, `Tensor`, and `Value` directly in Objective-C and Swift.

**Example**

Objective-C:
```objc
#import <ExecuTorch/ExecuTorch.h>

NSError *error;
ExecuTorchModule *module = [[ExecuTorchModule alloc] initWithFilePath:modelPath];
ExecuTorchTensor *tensor = [[ExecuTorchTensor alloc] initWithScalars:@[@1.0, @2.0, @3.0]];
NSArray<ExecuTorchValue *> *outputs = [module forwardWithTensors:@[tensor] error:&error];

if (outputs) {
  NSLog(@"Output: %@", outputs.firstObject.tensorValue);
} else {
  NSLog(@"Error: %@", error);
}
```

Swift:
```swift
import ExecuTorch

do {
  let module = Module(filePath: "path/to/model.pte")
[...]

PR #8826, opened 2025-02-28
Lay out ObjC/Swift bindings skeleton

shoumikhin:
Landing changes for this RFC
https://github.com/pytorch/executorch/issues/8360

### DIFF
--- a/extension/apple/ExecuTorch/Exported/ExecuTorchError.h
+++ b/extension/apple/ExecuTorch/Exported/ExecuTorchError.h
@@ -6,8 +6,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-#import "ExecuTorchError.h"
-#import "ExecuTorchLog.h"
-#import "ExecuTorchModule.h"
-#import "ExecuTorchTensor.h"
-#import "ExecuTorchValue.h"
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+FOUNDATION_EXPORT NSErrorDomain const ExecuTorchErrorDomain NS_SWIFT_NAME(ErrorDomain);
+
+NS_ASSUME_NONNULL_END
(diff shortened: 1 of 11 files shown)


mergennachin:
1. Added "release notes: api" label instead of "topic: not user facing" since it looks like user facing API change.

2. While we're still finalizing the API, do you wanna tag them as "experimental" as per the lifecycle? https://pytorch.org/executorch/main/api-life-cycle.html#how-to-mark-api-state In 0.6, we make the swift/objC bindings as experimental, and in 0.7 we can promote to stable APIs.. That way, you have more flexibility in terms of changing APIs until 0.7 without breaking our API policy.

shoumikhin:
@mergennachin thanks for suggestions!
This PR (likewise #8825 and a bunch landed yesterday) basically creates a frame for @bsoyluoglu to fill with some gems. So that the actual implementation can immediately become a part of OSS nightly builds, as well as available internally via Buck.

mergennachin:
> This PR (likewise https://github.com/pytorch/executorch/pull/8825 and a bunch landed yesterday) basically creates a frame for @bsoyluoglu to fill with some gems. So that the actual implementation can immediately become a part of OSS nightly builds, as well as available internally via Buck.

@shoumikhin Yeap, I understand that the APIs can be immediately part of the OSS builds, which is great. And you are building the scaffolding so that @bsoyluoglu fill in directly the implementations.

I am just making sure that if the users start immediately using the APIs in the nightly, they can expect a certain level of changes, including breaking changes in the API until ET v0.7 release. Otherwise, if you don't mark as experimental, it will be harder to make API breaking changes after 0.6 release (March/April), because by default it will be considered a stable API.

On top of the files, you can just mark them as experimental in the docblock.

shoumikhin:
@mergennachin those must be experimental for sure, until we switch internal clients to use them and prove everything works well. I've already marked them as such per your suggestion.

mergennachin (approved):

PR #8829, opened 2025-02-28
Arm backend: Add support for amax/max/amin/min

AdrianLundell:
Max/min can be decomposed as for example
  max(x) = (amax(x), argmax(x))

For MI, an operator support check is added to support max-ops for which the argmax is not used. For BI, the int64 dtype returned by argmax is currently not supported by the arm_quantizer, and the program will crash. This is the same behaviour as before the patch, but with an improved error message.

- Adds op_amax/op_amin node visitors.
- Renames op_max/min->op_maximum/minimum to clearly separate the two ops.
- Adds convert_minmax_pass to make min/max/amin/amax TOSA-compatible.
- Adds unittests

Util updates:
- Updates analyze_output_utils to support rank 0.
- Adds quantization to OpNotSupportedPipeline


zingo (approved):

zingo:
MacOS fails unrelated

PR #8833, opened 2025-02-28
Arm backend: Remove unused github ci git setup

zingo:
Setup.sh has changed how things are installed and a git user is not needed anymore.

### DIFF
--- a/.ci/scripts/setup-arm-baremetal-tools.sh
+++ b/.ci/scripts/setup-arm-baremetal-tools.sh
@@ -1,11 +1,10 @@
 #!/bin/bash
-# Copyright 2024 Arm Limited and/or its affiliates.
+# Copyright 2024-2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
 # NB: This function could be used to install Arm dependencies
 # Setup arm example environment (including TOSA tools)
-git config --global user.email "github_executorch@arm.com"
-git config --global user.name "Github Executorch"
+
 bash examples/arm/setup.sh --i-agree-to-the-contained-eula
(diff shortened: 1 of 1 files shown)


mergennachin (approved):

digantdesai (approved):

zingo:
Interesting, the FVP timeout when running should not be related to this change as this would have caused havoc during initial setup. But there is no hurry for this change, so Ill create a patch to bump the timeout and we can rerun this jobs to make sure. I assume the github machines can be a bit slow during heavy workloads so it would make sense to have a bigger value anyway.

digantdesai (approved):
I guess it should timeout without this charge often?

zingo:
I would expect it to, but don't see much, but when looking more closely there are not many jobs running with trunk flag.
If we need this urgently it should be unrelated to  to this patch. I have patch on internal review that bumps the timeout from 4min to 10min so my plan is to merge it first and reuse this job to check "green" status.

zingo:
I assume it could fail more/less depending on loads of the github machines as the timeout is in "normal time" but the FVP is simulating and running in whatever time it got to run on on the current virtual machine.

.ci/scripts/setup-arm-baremetal-tools.sh
```diff
@@ -1,11 +1,10 @@
...
+# Copyright 2024-2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
 # NB: This function could be used to install Arm dependencies
 # Setup arm example environment (including TOSA tools)
-git config --global user.email "github_executorch@arm.com"
-git config --global user.name "Github Executorch"
+
```
digantdesai, line 9: Nice, thanks!

digantdesai: cc @swolchok

PR #8850, opened 2025-02-28
add merge function for NamedDataStore

lucylq:
Summary:
Allow us to change the NamedDataStore model. Usage will now be:
```
named_data_store
for backend in backends:
    bytes, named_data_store = backend.preprocess()
    named_data_store.merge(named_data_store)
```

Note:
- named_data_store is held by edge program manager
- preprocess returns the named data store as part of PreprocessResult

Differential Revision: D70409078

### DIFF
--- a/exir/_serialize/test/test_named_data_store.py
+++ b/exir/_serialize/test/test_named_data_store.py
@@ -83,3 +83,62 @@ class TestNamedDataStore(unittest.TestCase):
         self.assertEqual(len(output.pte_data), 1)
         self.assertEqual(output.pte_data["key"], 0)
         self.assertEqual(len(output.external_data), 0)
+
+    def test_merge(self) -> None:
+        store1 = NamedDataStore()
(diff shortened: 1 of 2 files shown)


exir/_serialize/test/test_named_data_store.py
```diff
@@ -83,3 +83,41 @@ def test_add_duplicate_key_fail(self) -> None:
         self.assertEqual(len(output.pte_data), 1)
         self.assertEqual(output.pte_data["key"], 0)
         self.assertEqual(len(output.external_data), 0)
+
+    def test_merge_named_data_store(self) -> None:
```
mcr229, line 87: can we just add one negative test case where two NamedDataStore have the same key but different data, and expect the error thrown?

lucylq: thanks, updated!

mcr229 (approved):

PR #8857, opened 2025-03-01
add pass to remove cat from slice pass

zonglinpeng:
Summary:
only keep the cat before slice iff the slice range overlaps with *both* tensors in cat.
TODO: trace to 1+ level for more cat in a china
TODO: support >2 tensors in a cat

Differential Revision: D70425971


skrtskrtfb (approved):

swolchok:
this added a lintrunner failure https://github.com/pytorch/executorch/actions/runs/13730950536/job/38407785027

PR #8862, opened 2025-03-01
Basic performance logging for Llama python runner

jackzhxng:


lucylq (approved):
Lgtm! What are you trying to log times for btw?

jackzhxng:
@lucylq models like Qwen and Phi which we don't have json tokenizer cpp support for yet :'|

PR #8864, opened 2025-03-01
add BroadcastIndexesRange

swolchok:
See class comment. In brief, this adds an iterable range to make broadcasting ops convenient and efficient to implement.


swolchok:
* (to be filled)

kernels/portable/cpu/util/broadcast_indexes_range.h
```diff
@@ -0,0 +1,207 @@
...
+  pointer operator->() const {
+    return &current_indexes_;
+  }
+
+  BroadcastIndexesIterator& operator++() {
+    output_index()++;
+    // TODO: add optimization for particular input tensors not being
+    // broadcasted?
+    for (auto ii = output_dim_ - 1; ii >= 0; --ii) {
+      // You might wonder what happens if output_shape_[ii] == 0. In that case,
```
manuelcandales, line 78: shouldn't we check for this before starting to iterate at all?

kernels/portable/cpu/util/broadcast_indexes_range.h
```diff
@@ -0,0 +1,207 @@
...
+        "input to broadcasting op should have dim at most output dim, but %d > %d!",
+        (int)t.dim(),
+        (int)output.dim());
+
+    const auto num_leading_ones = output.dim() - t.dim();
+    for (const auto idx : c10::irange(num_leading_ones)) {
+      result[idx] = 0;
+    }
+    const auto t_sizes = t.sizes();
+    const auto t_strides = t.strides();
```
manuelcandales, line 136: does this take dim order into account?

kernels/portable/cpu/util/broadcast_indexes_range.h
```diff
@@ -0,0 +1,207 @@
...
+  ssize_t output_dim_;
+  ArrayRef<exec_aten::SizesType> output_shape_;
+  // The linear index for a broadcast tensor is
+  // sum(delinearized_output_index_[i] * input_stride_[i] if
+  // padded_input_shape_[i] != 1 else 0), where padded_input_shape is
+  // input.sizes() with leading 1s added to make its size equal to
+  // output_dim. This is straightforwardly implementable with an
+  // adjusted stride array that contains 0s where the padded input
+  // shape would contain 1s.
+  std::array<ShapeType, kNumInputs> effective_input_broadcast_strides_ = {
```
manuelcandales, line 161: I love this kNumInputs generalization. This is great!

kernels/portable/cpu/util/test/broadcast_indexes_range_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+      {4, 0},
+  };
+  EXPECT_EQ(expected, actual);
+}
+
+// [1] -> [H, W]
+// [1, W] -> [H, W]
+// [H, 1] -> [H, W]
+// [H, W] -> [H, W]
+// Cover all these at the same time to also exercise multiple input tensors.
```
manuelcandales, line 75: you are also covering [1, 1] -> [H, W] and [W] -> [H, W] here. would be good to mention as well.

kernels/portable/cpu/util/test/broadcast_indexes_range_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+      {8, 0, 0, 0, 0, 2, 8},
+      {9, 0, 0, 0, 1, 2, 9},
+      {10, 0, 0, 0, 2, 2, 10},
+      {11, 0, 0, 0, 3, 2, 11},
+  };
+  EXPECT_EQ(expected, actual);
+}
+
+// Here we assume that the previous tests established that padding
+// with leading 1s is working, and test:
```
manuelcandales, line 114: You are testing 5 out of 8 possibilities here. You might as well add the remaining 3:
[1, 1, 1] -> [C, H, W]
[1, 1, W] -> [C, H, W] (this one in particular would be good to have, i.e. multiple leading ones)
[1, H, W] -> [C, H, W]

kernels/portable/cpu/util/test/broadcast_indexes_range_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+      {9, 2, 1, 0, 2, 9},   {10, 2, 2, 0, 2, 10}, {11, 2, 3, 0, 2, 11},
+      {12, 3, 4, 1, 0, 12}, {13, 3, 5, 1, 0, 13}, {14, 3, 6, 1, 0, 14},
+      {15, 3, 7, 1, 0, 15}, {16, 4, 4, 1, 1, 16}, {17, 4, 5, 1, 1, 17},
+      {18, 4, 6, 1, 1, 18}, {19, 4, 7, 1, 1, 19}, {20, 5, 4, 1, 2, 20},
+      {21, 5, 5, 1, 2, 21}, {22, 5, 6, 1, 2, 22}, {23, 5, 7, 1, 2, 23},
+  };
+  EXPECT_EQ(expected, actual);
+}
+
+// 4-D should generalize, but we will go ahead and test:
```
manuelcandales, line 148: this is great!

manuelcandales (approved):

kernels/portable/cpu/util/broadcast_indexes_range.h
swolchok: This comment is meant to be explaining why every caller does check for that already -- in that case begin() == end() and any loop that uses this thing won't be entered. I'll see if I can make that a bit clearer.

kernels/portable/cpu/util/broadcast_indexes_range.h
swolchok: I don't recall how dim_order affects strides and sizes. if the tests pass, either it works or we have no tests for dim_order support (which would mean it didn't work before this diff).

kernels/portable/cpu/util/test/broadcast_indexes_range_test.cpp
swolchok: good catch, I'll rename it to OneAndTwoDExhaustive

kernels/portable/cpu/util/broadcast_indexes_range.h
kimishpatel: at the minimal you should add checks for the dim order assumptions that are being made here. That is is assumes whatever the default dim order is, nothing fancy. If in future when the tests with dim order are added, at least this will be caught more gracefully rather than having to go down the debug rabbit hole

kernels/portable/cpu/util/test/broadcast_indexes_range_test.cpp
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+}
+
+// 4-D should generalize, but we will go ahead and test:
+// [N, 1, H, 1] -> [N, C, H, W]
+// [1, C, 1, W] -> [N, C, H, W]
+TEST(BroadcastIndexesRangeTest, FourDBroadcasting) {
+  TensorFactory<ScalarType::Int> tf;
+  Tensor out = tf.zeros({2, 3, 4, 5});
```
kimishpatel, line 171: What about out = {2, 3, 1, 5} and a = {1, 3, 1, 5} and b = {2, 1, 1, 5}

Mainly highlighting that it there is size 1 dim in the output, is it taken care of? From cursory looks I presume the answer is yes, but not sure

swolchok: >  it there is size 1 dim in the output, is it taken care of? From cursory looks I presume the answer is yes, but not sure

why would H == 1 be special? I'll add an exhaustive test for that for 1- and 2-D just in case in a follow-up.

swolchok: #8964

PR #8866, opened 2025-03-01
Add optimized op_where

swolchok:
It materializes separate kernels for the cases where the two input data tensors have the same dtype and the third one has dtype bool.


swolchok:
* (to be filled)

kernels/optimized/cpu/op_where.cpp
```diff
@@ -0,0 +1,97 @@
...
+          (a_is_broadcasted || b_is_broadcasted || cond_is_broadcasted);
+      const CTYPE_COMPUTE* const data_a = a.const_data_ptr<CTYPE_COMPUTE>();
+      const CTYPE_COMPUTE* const data_b = b.const_data_ptr<CTYPE_COMPUTE>();
+      const bool* const data_cond = cond.const_data_ptr<bool>();
+      CTYPE_COMPUTE* const data_out = out.data_ptr<CTYPE_COMPUTE>();
+      if (any_is_broadcasted) {
+        for (const auto [out_index, a_index, b_index, cond_index] :
+             BroadcastIndexesRange<3>(out, a, b, cond)) {
+          data_out[out_index] =
+              data_cond[cond_index] ? data_a[a_index] : data_b[b_index];
```
digantdesai, line 64: Love how clean it reads :)

Couple of high level comments,
- Not too familiar with how people use this op i.e. any common case we see often in the wild, but are there any short-circuits we can do to avoid coming here?
- Similarly, depending on the condition, if we know we are biased towards A vs. B, we can first copy A to the result and then go through data_cond and pick B.
- Lastly, I am assuming we want to do SIMD later, if we do we can use predicates, that should make it less "branchy". It may not help much with load/stores though.

digantdesai (approved):
Looks good to me.

kernels/optimized/cpu/op_where.cpp
swolchok: >  if we know we are biased towards A vs. B

we can't possibly know this in general.

> SIMD later

not currently on my agenda, but I may have to come back.

manuelcandales (approved):

swolchok:
> I think this will cause -Wunused -Werror builds to fail though

Checked, works fine, I misread my own code.

PR #8879, opened 2025-03-03
Fix docstring of to_backend in backend_api.py

YufengShi-dudu:
The second method overloading of to_backend() has the wrong signature in docstring

Change-Id: I9e1c5eccea00003b5e2821e666a9e0f77902d794


tarun292 (approved):

tarun292:
@YufengShi-dudu good catch. Thanks for fixing this!

PR #8880, opened 2025-03-03
Arm backend: Enable test_w2l_u85_BI

martinlsm:
The TOSA compiler previously had a bug that caused a segmentation fault when loading the Wav2Letter model on Ethos-U85. This issue has now been fixed.

Enable the test that previously failed due to this bug.


zingo (approved):

zingo:
MacOS fails are unrelated

PR #8882, opened 2025-03-03
Add editable mode in Building from source file

mergennachin:


larryliu0820 (approved):

docs/source/using-executorch-building-from-source.md
```diff
@@ -80,6 +80,14 @@ portability details.
    ./install_executorch.sh --pybind off
    ```
 
+   For development, install the package in `--editable` mode, which allows to modify Python source code and see changes reflected immediately.
```
GregoryComer, line 83: Thanks for adding this to the docs. In the long-term, I wonder if we should split out contributing vs user build from source info, but we don't really have a great place for it for now, so this seems good.

PR #8883, opened 2025-03-03
Delete old Getting Started Setup.md

mergennachin:
Let's not have two copies

Current copy is disintegrated into: 

- https://pytorch.org/executorch/main/getting-started.html
- https://pytorch.org/executorch/main/using-executorch-building-from-source.html


mergennachin:
Oh looks like I can't delete this right away, still lots of references

https://github.com/search?q=repo%3Apytorch%2Fexecutorch+getting-started-setup&type=code&p=2

mergennachin:
Add a redirection instead

GregoryComer (approved):
Looks good. Thanks for cleaning this up - I take a look and see if I can remove any more old pages.

PR #8887, opened 2025-03-03
[executorch][runtime] Introduce PteDataMap for weight sharing

lucylq:
* __->__ #8887

PteDataMap is the NamedDataMap that will live in the runtime. It is used to give delegates access to opaque named data stored in the PTE file. Open to alternative naming suggestions, maybe 'PTEDataMap' or 'ProgramDataMap'?

**Usage**
The PteDataMap is owned by the program, and instantiated at program load time if named_data exists in the PTE file. We introduce usage of 'std::optional' here. I think we can also use executorch::aten::optional to avoid adding standard lib ?

When initializing delegates, the PteDataMap is given to delegate_init. Delegates can retrieve opaque delegate data by key using 'get_data'. This gives them a FreeableBuffer that they can free later.

**Testing**
This test uses the C++ flatbuffer API to build a fake program containing named data. We also creates a temp file with sample data that the data loader can wrap around.

TODO: e2e test once delegate aot is ready and we can generate a file with named data.

**Note**
As the PteDataMap wraps around flatbuffer constructs, the Program must outlive the PteDataMap.

PteDataMap does not implement
- get_metadata; currently, all data stored is opaque. Later, we can implement get_metadata if a backend stores plain tensor data.
- load_into; this is mostly used for the training case, and isn't used by delegates, at least not at the moment

Differential Revision: [D70213646](https://our.internmc.facebook.com/intern/diff/D70213646/)


runtime/executor/program.h
```diff
@@ -13,6 +13,7 @@
 
 #include <cinttypes>
 #include <cstdint>
+#include <optional>
```
mcr229, line 16: is this right? Can we add this dependency to program?

lucylq: updated

larryliu0820 (approved):

PR #8895, opened 2025-03-03
[ExecuTorch] Arm Ethos:Make get_compile_spec() configurable

digantdesai:
* __->__ #8895
* #8894
* #8893

Allow users to generate different compile specs based on the Args.
Preserve the behavior.
No change in the default spec returned.

Differential Revision: [D70351486](https://our.internmc.facebook.com/intern/diff/D70351486/)

### DIFF
--- a/backends/arm/test/common.py
+++ b/backends/arm/test/common.py
@@ -11,7 +11,7 @@ import tempfile
 from datetime import datetime
 
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
 
 import pytest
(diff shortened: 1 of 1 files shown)


AdrianLundell (approved):
Approved with a side note that the intended use of this API is to configure the unbuilt spec, so you could achieve the same behaviour with something like:
```
spec = get_u55_compile_spec_unbuilt().ethosu_compile_spec(<config args>).build()
```
I can see that this is more convenient however.

PR #8902, opened 2025-03-03
Add cpu_thread setting logic to xnn_executor_runner

swolchok:


swolchok:
* (to be filled)

iseeyuan (approved):
LGTM. Thanks!

mcr229 (approved):
this seems fine, but honestly coming back to this after year+ since our first release, do we need to have an xnn_executor_runner which is separate from the executor_runner? Should our CMake just be if BUILD_XNNPACK=ON, we link XNNPACK to executor_runner, if not then we dont?

swolchok:
> this seems fine, but honestly coming back to this after year+ since our first release, do we need to have an xnn_executor_runner which is separate from the executor_runner? Should our CMake just be if BUILD_XNNPACK=ON, we link XNNPACK to executor_runner, if not then we dont?

sounds like a great idea to me! this change is still an incremental improvement though so I will land it.


On another note, CI is green, rebasing to make sure extra changes from previous commit go away and then merging without waiting for CI to run again

PR #8904, opened 2025-03-03
Fix ANE llama export

metascroy:
Fixes ANE llama export and adds unit test

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -508,22 +508,25 @@ jobs:
         bash .ci/scripts/setup-conda.sh
         eval "$(conda shell.bash hook)"
 
-        # Setup MacOS dependencies as there is no Docker support on MacOS atm
-        PYTHON_EXECUTABLE=python \
-        EXECUTORCH_BUILD_PYBIND=ON \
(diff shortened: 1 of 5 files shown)


YIWENX14 (approved):
LGTM, thanks for fixing the export issue

.github/workflows/pull.yml
```diff
@@ -495,6 +495,38 @@ jobs:
         # Test static llama weight sharing and accuracy
         PYTHON_EXECUTABLE=python bash .ci/scripts/test_qnn_static_llama.sh
 
+  test-static-llama-ane:
+    name: test-static-llama-ane
+    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
```
cccclai, line 500: I thought all the mac jobs are in trunk because of the cost/reliability issue?

cccclai: cc: @kirklandsign @huydhn

cccclai: See https://github.com/pytorch/executorch/pull/610, not sure if it's still true

huydhn: Yeah, macos_job ideally should be in trunk.yml and can be run on PR using `ciflow/trunk`.  The job could be here is it's important enough to justify running in every pull requests.  I assume that this is not the case here.

metascroy: Ok, I moved the job to trunk.yml

PR #8910, opened 2025-03-04
Cleanup XNN delegate post removal of upsample decomposition

GregoryComer:
Summary: Following the removal of the decompositions for upsample_bilinear2d in PyTorch, we don't need to pattern match and recompose upsample_bilinear2d. This PR cleans up the convert_to_upsample_bilinear2d pass and associated logic.

Differential Revision: D68374585


digantdesai (approved):

mcr229:
nice dead code!

mcr229 (approved):

GregoryComer:
Thanks. With https://github.com/pytorch/pytorch/pull/147153, we should never see a decomposed upsample unless it somehow gets manually added to the decomp table. It's core and is now never decomposed by default.

PR #7217, opened 2024-12-06
Support build on native Windows x64

chinazhangchao:
### Summary
This PR is the first one of a series of PRs split from https://github.com/pytorch/executorch/pull/6979 to support for out-of-the-box builds on Windows x64.
1. **(this one) make `./install_requirements.bat` build and install `executorch` python package successfully, without xnnpack**
2. make `./install_requirements.bat --pybind xnnpack` work
3. Support `examples/models/llama/`

### Test plan
PR tested by running minimal executor_runner example (https://pytorch.org/executorch/stable/getting-started-setup.html#run-your-program) in the following environment:

- Windows 11 x64
- Visual Studio Enterprise 2022 with these components installed:
  - C++ Clang tools for Windows
  - MSBuild support for LLVM (clang-cl) toolset
  - C++ CMake tools for Windows

Steps:
```
# run the following in command prompt
git clone -b win_build_basic https://github.com/vortex-captain/executorch.git
cd executorch/  # contains code of this PR
git submodule sync
git submodule update --init

./install_requirements.bat
python .\export_add.py  # export_add.py from https://pytorch.org/executorch/stable/getting-started-setup.html#export-a-program
msbuild pip-out\temp.win-amd64-cpython-310\cmake-out\executorch.sln /t:Rebuild
pip-out\temp.win-amd64-cpython-310\cmake-out\Debug\executor_runner.exe --model-path add.pte
```

### References
https://github.com/pytorch/executorch/pull/6979
https://github.com/pytorch/executorch/pull/4681
https://github.com/pytorch/executorch/pull/5252
https://github.com/pytorch/executorch/pull/4899
https://github.com/pytorch/executorch/pull/5164
https://github.com/pytorch/executorch/pull/4993

PR #8913, opened 2025-03-04
[Windows] [file_data_loader.cpp] Add compat_unistd.h

SamGondelman:
### Summary
This is the second PR in my attempt to get Windows support working.  I've pulled the changes to file_data_loader.cpp from #7217 into their own PR and removed the other changes for now.

### Test plan
No functional change since we aren't actually building Windows yet, just verify everything builds.

### DIFF
--- a/backends/xnnpack/CMakeLists.txt
+++ b/backends/xnnpack/CMakeLists.txt
@@ -74,17 +74,33 @@ foreach(fbs_file ${_xnnpack_schema__srcs})
 endforeach()
 
 # Generate the headers from the .fbs files.
-add_custom_command(
-  OUTPUT ${_xnnpack_schema__outputs}
-  COMMAND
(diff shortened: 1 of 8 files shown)


SS-JIA (approved):

SS-JIA:
re: testing, I think CI is sufficient because there should be no change to existing build

SS-JIA:
@SamGondelman all tests pass which is good. However, there is a lint error; seems you need to remove whitespace at Line 21 of `extension/data_loader/file_data_loader.cpp`.

If you want to make managing lints easier, you can set up `lintrunner` by just installing it through `pip`

```
pip install lintrunner_adapters
pip install lintrunner
```

Then from repo source

```
lintrunner init
lintrunner lint --apply-patches --verbose
```

PR #8914, opened 2025-03-04
[Windows] [memory_allocator.h] remove ET_TRY macros

SamGondelman:
### Summary
This is the third PR in my attempt to get Windows support working. I hit these static asserts because these macros weren't implemented.  Let's remove the macros and just directly allocate the memory everywhere.

### Test plan
Verify everything builds + unit tests.


jackzhxng (commented):
The goal is to eventually remove all usages of `ET_TRY_ALLOCATE_OR, ET_TRY_ALLOCATE_INSTANCE_OR, and ET_TRY_ALLOCATE_LIST_OR`, these exist to retain compatibility with the non mvc compilers while enabling mvc compilation. Since you are hitting these static assertions, I'm guessing that you are using one of these three macros somewhere in your code - that's where you should replace the macro with the recommended `memory_allocator` method

jackzhxng (approved):

PR #8915, opened 2025-03-04
[Windows] don't use invalid flags on Windows

SamGondelman:
### Summary
This is the fourth PR in my attempt to get Windows builds working.  Since "-Wno-missing-prototypes", `#pragma GCC diagnostic push`, etc. aren't valid with non-gcc compilers, let's wrap them so we are only doing them on non-Windows platforms.

### Test plan
Since we aren't actually building on Windows yet, this should have no effect on existing platforms, so just confirm that everything builds.

### DIFF
--- a/backends/vulkan/targets.bzl
+++ b/backends/vulkan/targets.bzl
@@ -5,7 +5,13 @@ load("@fbsource//tools/build_defs:platform_defs.bzl", "ANDROID", "CXX", "FBCODE"
 
 
 def get_vulkan_compiler_flags():
-    return ["-Wno-missing-prototypes", "-Wno-global-constructors"]
+    return select({
+        "DEFAULT": [
+            "-Wno-global-constructors",
+            "-Wno-missing-prototypes",
+        ],
+        "ovr_config//os:windows": [],
+    })
 
 def get_labels(no_volk):
     if no_volk:
--- a/runtime/core/named_data_map.h
+++ b/runtime/core/named_data_map.h
@@ -7,9 +7,12 @@
  */
 
 #pragma once
+
+#ifdef __GNUC__
 // Disable -Wdeprecated-declarations, as some builds use 'Werror'.
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
 
 #include <executorch/runtime/core/exec_aten/exec_aten.h>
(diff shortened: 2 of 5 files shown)


backends/vulkan/targets.bzl
```diff
@@ -5,7 +5,14 @@ load("@fbsource//tools/build_defs:platform_defs.bzl", "ANDROID", "CXX", "FBCODE"
...
 
 def get_vulkan_compiler_flags():
-    return ["-Wno-missing-prototypes", "-Wno-global-constructors"]
+    return select({
+        "DEFAULT": [
+            "-Wno-global-constructors",
+            "-Wno-missing-prototypes",
+            "-Wno-sign-compare",
+        ],
+        "ovr_config//os:windows": [],
```
SS-JIA, line 14: I _think_ we can leave the existing `#pragma` stuff and just add a `select` call with a windows option like you did here so that unsupported options are not used when compiling with Windows

runtime/core/named_data_map.h
```diff
@@ -7,9 +7,6 @@
  */
 
 #pragma once
-// Disable -Wdeprecated-declarations, as some builds use 'Werror'.
-#pragma GCC diagnostic push
```
SS-JIA, line 11: wrap these with

```
#ifdef __GNUC__
```

and I think there will not be any need to remove these lines.

SS-JIA:
Overall, I think we only need to change the `.bzl` files to make sure that we aren't using unsupported compile flags for windows. I would try to avoid modifying the `.h`/`.cpp`/ files if possible, except to add `#ifdef _GNUC_` guards wherever necessary.

SS-JIA (changes requested):

backends/vulkan/targets.bzl
SamGondelman: for sure, I can do that!

SamGondelman:
@SS-JIA thank you!  I've made those changes and it's much better.

If I grep for `compile_flags = ` I'm noticing now that there are several other places that might need the select for Windows...since we aren't actually building on Windows yet, I'm hesitant to go in and change all of them yet for fear of breaking something.  what do you think?

SS-JIA (approved):

SS-JIA:
@SamGondelman LGTM! Yes, we can update compile flags incrementally.

PR #8919, opened 2025-03-04
[minibench] Drop outliers from benchmark result

kirklandsign:
### Summary
Currently the result has large variance from outliers, so only use 80% samples in the middle (trimmean 0.2)

### Test plan
CI

### DIFF
--- a/extension/benchmark/android/benchmark/app/src/main/java/org/pytorch/minibench/BenchmarkActivity.java
+++ b/extension/benchmark/android/benchmark/app/src/main/java/org/pytorch/minibench/BenchmarkActivity.java
@@ -85,8 +85,8 @@ public class BenchmarkActivity extends Activity {
         // 80% samples in the middle (trimmean 0.2)
         Collections.sort(stats.latency);
         int resultSize = stats.latency.size();
-        List<Double> usedLatencyResults = stats.latency.subList(
-            resultSize / 10, resultSize * 9 / 10);
+        List<Double> usedLatencyResults =
+            stats.latency.subList(resultSize / 10, resultSize * 9 / 10);
 
         results.add(
             new BenchmarkMetric(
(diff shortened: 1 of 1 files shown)


kirklandsign:
@guangy10 @huydhn 

The result is not bad. Now the difference between different runs is reduced to about 2%.

For load time, unfortunately it cannot be addressed yet, because we only have 1 load time measurement during a run.

@huydhn now I use "avg_inference_latency" field, but actually it's trimmean. Please let me know if you are unhappy with using the existing field. Honestly I think it's ok 😜 

ic4
xnnpack_q8
Samsung Galaxy S24 (Android 14)
24.74 → 25.03
51.46 → 64.21
ic4
xnnpack_q8
Samsung Galaxy S24 Ultra (Android 14)
23.4 → 23.72
51.05 → 56.44
ic4
xnnpack_q8
Samsung Galaxy S24+ (Android 14)
26.17 → 26.48
64.59 → 56.06

kirklandsign:
Again, very promising result after I tried one more time.

ic4
xnnpack_q8
Samsung Galaxy S24 (Android 14)
23.53 → 24.74
76.1 → 51.46
ic4
xnnpack_q8
Samsung Galaxy S24 Ultra (Android 14)
23.69 → 23.4
55.37 → 51.05
ic4
xnnpack_q8
Samsung Galaxy S24+ (Android 14)
27.44 → 26.17
55.61 → 64.59

huydhn (approved):
LGTM!

huydhn:
> @huydhn now I use "avg_inference_latency" field, but actually it's trimmean. Please let me know if you are unhappy with using the existing field. Honestly I think it's ok 😜

I'm ok with this too although I don't know enough statistics to decide if 0.2 is a reasonable value to use, i.e. why not 0.1.  For the load time, I think we could consider running minibench via abd multiple times but it feels overkill, maybe it's ok to have a load time with high variance and just use a higher alert threshold for that metric on the dashboard.

kirklandsign:
> I'm ok with this too although I don't know enough statistics to decide if 0.2 is a reasonable value to use, i.e. why not 0.1.

Unfortunately I tried 0.1 but it is not so good. It's quite left skewed.

PR #8933, opened 2025-03-04
Tiny fixes to make gcc pedantic build happy

digantdesai:
### Summary
Tiny fixes to make gcc 13 baremetal, pedantic build happy.

### Test plan
CI


kirklandsign (approved):
linter

kirklandsign:
Also please update the title/commit message a bit. It has more than Fixup CMake summary message

kernels/portable/cpu/op_constant_pad_nd.cpp
```diff
@@ -189,7 +189,7 @@ Tensor& constant_pad_nd_out(
     CTYPE value_v;
     ET_SWITCH_SCALAR_OBJ_TYPES(
         value_type, ctx, "constant_pad_nd.out", CTYPE_VALUE, [&]() {
-          CTYPE_VALUE val;
+          CTYPE_VALUE val{};
```
swolchok, line 192: what is the error message motivating this? it seems silly and reminiscent of misguided linter advice I've seen previously, but perhaps I don't know something. Is it -Wuninitialized? If so, why empty braces instead of `= 0;`?

runtime/executor/method_meta.cpp
```diff
@@ -233,7 +233,7 @@ bool MethodMeta::uses_backend(const char* backend_name) const {
   const auto delegates = s_plan_->delegates();
   for (size_t i = 0; i < delegates->size(); i++) {
     auto delegate = delegates->Get(i);
-    if (std::strcmp(delegate->id()->c_str(), backend_name) == 0) {
+    if (strncmp(delegate->id()->c_str(), backend_name, delegate->id()->size()) == 0) {
```
swolchok, line 236: 1) nit: why drop std::?
2) this is incorrect -- it passes if the delegate ID is a prefix of the backend_name

swolchok (changes requested):
changes requested for strncmp bug

kernels/portable/cpu/op_constant_pad_nd.cpp
digantdesai: yeah it was that. `COMPLEX_TYPES` is what I was thinking about when not using `=0`. But I guess I can use it when there is no complex.

runtime/executor/method_meta.cpp
digantdesai: `std::` is an oversight. 

Prefix is a real issue, let me also check the length of the incoming in addition to strncmp.

kernels/portable/cpu/op_constant_pad_nd.cpp
digantdesai: Done.

runtime/executor/method_meta.cpp
digantdesai: Thanks. Done.

swolchok (approved):
unblocking, but I'd like to hear about why we can't simply use strcmp

digantdesai:
> unblocking, but I'd like to hear about why we can't simply use strcmp

Thanks. Basically `-Werror=stringop-overread` was triggering here. I tried some tweaks but couldn't get it to stop on `strcmp` so just did `strncmp`.

swolchok:
> -Werror=stringop-overread was triggering here.

IIRC this one being buggy in some random gcc version has come up previously, possibly on @mergennachin's size_test warning stuff? In general I don't think we should be held responsible for every warning in every quasi-reasonable version of every compiler because of stuff like this.

digantdesai:
yeah this was arm's bare metal gcc 13, I generally agree with you. That said since this was the only one I got tempted 😅

PR #8934, opened 2025-03-04
[ExecuTorch][XNNPACK] Rename linear weight partitioning flag for clarity

kirklandsign:


kirklandsign:
This has different content from D70372220!

mcr229 (approved):
i'm guessing there is some diff train issues which is why this is coming up again. Accepting to unblock

Issue #8932, opened 2025-03-04
Parallelize portable ops if threadpool is available, with fallback to parallel_for-as-for-loop


swolchok:
### 🚀 The feature, motivation and pitch

It seems suboptimal to me that we have to create separate optimized ops just to get basic stuff like parallelization (and vectorization, but let's start with parallelization). Here's what I'd like to do: (The timeline here is "ASAP", but I'm opening an issue because this got too long for chat and so that I can point to this issue on the PRs.)

1) Set up a proper CMake build for extension/parallel; right now it's free-riding on buck and getting automatically duplicated into 3 different targets per the generated executorch_srcs.cmake. (done; #8938)
2) Make extension_threadpool itself export [the  `-DET_USE_THREADPOOL` macro we already use and define somewhat ad-hoc](https://github.com/search?q=repo%3Apytorch%2Fexecutorch%20ET_USE_THREADPOOL&type=code). (done; #8947)
3) move [extension/parallel/thread_parallel.h](https://github.com/pytorch/executorch/blob/main/extension/parallel/thread_parallel.h) to core. (@larryliu0820 suggests `runtime/kernel/thread_parallel.h`) (Yes I will leave a stub header behind for backward compatibility.) Move thread_parallel.cpp to threadpool, since there will be no reason not to provide it when threads are available. Provide a default implementation of parallel_for if threadpool is not built (gated behind `ET_USE_THREADPOOL`) that is just an inlinable `for` loop. (#8983)
4) use `parallel_for` in at least one portable op, either directly or via the workhorse "util" functions. (#8986)
5) Verify that, because the optimized library is built with threadpool, it gets parallelization. Adjust build configuration for optimized ops lib if necessary. (#8987)
5) Roll out `parallel_for` across portable ops and workhorse "util" functions.

Thoughts? Blockers?

### Alternatives

status quo -- slow portable ops

cc @larryliu0820 @manuelcandales

kimishpatel:
One of the goals on portable op was readability as well. SO it would be good to at the least touch upon that.

> move [extension/parallel/thread_parallel.h](https://github.com/pytorch/executorch/blob/main/extension/parallel/thread_parallel.h) to core

probably be better find a different location because it will start pulling in pthreadpool and other dependencies when you actually want threadpool. Keeping in extension would mean portable ops will depend on something outside of the core, but within the core will complicate the build system?

swolchok:
[...]

PR #8938, opened 2025-03-04
Add proper CMake build for extension_parallel

swolchok:
Previously it was copied in several places per executorch_srcs.cmake.

Needed for #8932

Test Plan: Compare cmake-out/executorch_srcs.cmake before/after for my usual testing cmake config with "all the CPU stuff" on; found that thread_parallel.cpp is now duplicated only in one place instead of multiple (it's in llama_runner, which needs a general fixup because it's duplicating several extensions).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -749,9 +749,9 @@ endif()
 
 if(EXECUTORCH_BUILD_PTHREADPOOL
    AND EXECUTORCH_BUILD_CPUINFO
-   AND CMAKE_CXX_STANDARD GREATER_EQUAL 14
 )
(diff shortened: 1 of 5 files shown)


swolchok:
* __->__ #8938

larryliu0820 (approved):
What is the difference between `extension_threadpool` and `extension_parallel`?

swolchok:
> What is the difference between `extension_threadpool` and `extension_parallel`?

extension_threadpool lives in extension/threadpool, and extension_parallel lives in extension/parallel. they're different code.

kirklandsign (approved):

PR #8941, opened 2025-03-04
Arm backend: Fix Timing Adapter settings depending on the memory mode…

gggekov:
… & placement in the linker script


zingo:
For some unknown reason the trunk setup job failed so I rebased it to re-trigger the run of the trunk tests.

zingo (approved):

PR #8947, opened 2025-03-04
Properly export ET_USE_THREADPOOL from the threadpool extension

swolchok:
We had a bunch of targets that would define this macro ad-hoc. It's supposed to indicate that the threadpool extension is available, so just make sure that we have it as a PUBLIC `target_compile_definition` in CMake and an `exported_preprocessor_flags` entry in Buck for extension_threadpool.

Test Plan: CI on following NOCOMMIT PR, which fails builds if ET_USE_THREADPOOL is not defined in affected places.

Needed for #8932.


swolchok:
* (to be filled)

kirklandsign (approved):

PR #8948, opened 2025-03-04
NOCOMMIT: fail CI if ET_USE_THREADPOOL isn't defined where we think it should be

swolchok:


swolchok:
* (to be filled)

swolchok:
noting that CI is 100% green in case I rebase

PR #8954, opened 2025-03-05
Use CMake EXPORT

swolchok:
The CMake EXPORT feature makes target_compile_options/target_link_libraries etc. work nicely for both in-project builds and also projects that consume your project by installing it and finding it with find_package. We are currently working around our lack of EXPORT by (incompletely and therefore incorrectly!) repeating our configuration in tools/cmake/executorch-config.cmake.

I was previously stalled here by XNNPACK's lack of use of EXPORT, but I discovered that I simply wasn't being thorough enough with installing/exporting the entire target graph.

Note that this diff raises the size test thresholds. This is because previously the size test wasn't including prim op registration, and this diff effectively *fixes* builds using find_package to always include that.


swolchok:
* #12792
* __->__ #8954

swolchok:
have an update cooking, need to make sure check whether warnings I'm seeing about macOS version mismatch during install_executorch.sh are new and if so problematic

swolchok:
Clear TODO list from test failures:
- [X] Stack yet another PR to convert FXDiv to ExternalProject
- [X] mpsdelegate needs a path update (manually test build_apple_frameworks and the 4 apple presets)
- [X] extension_module needs a path update (manually test build_android_library)
- [X] pthreadpool, cpuinfo, XNNPACK, xnnpack-microkernels-prod need to be put back in executorch-config.cmake (manually test llama-runner)
- [X] The examples/models/llama and extension/llm/tokenizers builds need a bunch of fixing (again, manually test llama-runner)
- [X] neuron_backend has a bad path in INTERFACE_SOURCES (test by CI)

swolchok:
I've updated extension/llm/tokenizers to point at https://github.com/pytorch-labs/tokenizers/pull/103 . We'll presumably need to fix the submodule reference after that gets merged.

swolchok:
test-binary-size-linux regression is real. text size was 35810 before and is now 43358.

swolchok:
I think the size regression is the same old "now we are counting register_prim_ops" fake regression from https://github.com/pytorch/executorch/pull/12320, and I think we should just accept it and move the goalpost, since it is caused by building the size test the same way we build internally. If this really isn't what we want, then we should stop using force_load/whole_archive on the top-level ExecuTorch target, but I strongly suspect that *that* is not what we want. CC @mergennachin

.github/workflows/pull.yml
```diff
@@ -371,7 +371,7 @@ jobs:
         size=${arr[4]}
         # threshold=48120 on devserver with gcc11.4
         # todo(lfq): update once binary size is below 50kb.
-        threshold="55584"
+        threshold="63776"
```
digantdesai, line 374: This is a bigger jump, do you know what is getting pulled in the build with these changes esp just with gcc and not with clang?

tools/cmake/executorch-config.cmake
```diff
@@ -97,105 +85,27 @@ set(lib_list
...
-                                    "cpuinfo;pthreadpool"
-  )
-endif()
-
-set(shared_lib_list
-    # executorch -- size tests fail due to regression if we include this and I'm
-    # not sure it's needed.
-    optimized_native_cpu_ops_lib portable_ops_lib quantized_ops_lib
-    xnnpack_backend vulkan_backend quantized_ops_aot_lib
+# The ARM baremetal size test doesn't use a CMAKE_TOOLCHAIN_FILE, so our
```
digantdesai: It should - https://github.com/pytorch/executorch/blob/d1c87e4995b46db332361503a15c3bf01b1c0b55/.github/workflows/trunk.yml#L270

.github/workflows/pull.yml
digantdesai: nvm saw this - https://github.com/pytorch/executorch/pull/8954#issuecomment-3092596498

digantdesai: why just for gcc though?

swolchok: it's not just for gcc. they both had a significant size increase; I can't speak to the difference between them or the difference in overage in the thresholds previously.

swolchok:
looks like we might finally have green CI (given the upstream repo issue) except that the arm baremetal size tests need a threshold bump!

CMakeLists.txt
```diff
@@ -447,53 +470,54 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/configurations)
...
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/executorch/extension/threadpool
   FILES_MATCHING
   PATTERN "*.h"
 )
 install(
   TARGETS executorch executorch_core
+  EXPORT ExecuTorchTargets
   INCLUDES
-  DESTINATION ${_common_include_directories}
+  DESTINATION include
```
swolchok: should be ${CMAKE_INSTALL_INCLUDEDIR}

backends/apple/coreml/CMakeLists.txt
```diff
@@ -105,18 +105,23 @@ endif()
 
 add_library(coreml_util ${UTIL_SOURCES})
 target_include_directories(
-  coreml_util PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/util
+  coreml_util
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/runtime/util>
+    $<INSTALL_INTERFACE:include/executorch/backends/apple/coreml/runtime/util>
```
swolchok: s/include/${CMAKE_INSTALL_INCLUDEDIR}/

backends/apple/coreml/CMakeLists.txt
```diff
@@ -105,18 +105,23 @@ endif()
...
 
+install(TARGETS coreml_util DESTINATION lib)
+
 install(
-  TARGETS coreml_util
-  DESTINATION lib
-  INCLUDES
-  DESTINATION ${_common_include_directories}
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/runtime/util
+  DESTINATION include/executorch/backends/apple/coreml/runtime/util
```
swolchok: ditto includedir

backends/apple/coreml/CMakeLists.txt
```diff
@@ -134,7 +139,10 @@ endif()
 
 add_library(coreml_inmemoryfs ${INMEMORYFS_SOURCES})
 target_include_directories(
-  coreml_inmemoryfs PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/inmemoryfs
+  coreml_inmemoryfs
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/runtime/inmemoryfs>
+    $<INSTALL_INTERFACE:include/executorch/backends/apple/coreml/runtime/inmemoryfs>
```
swolchok: ditto includedir

backends/apple/coreml/CMakeLists.txt
```diff
@@ -150,6 +158,13 @@ install(
   DESTINATION ${_common_include_directories}
 )
 
+install(
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/runtime/inmemoryfs
+  DESTINATION include/executorch/backends/apple/coreml/runtime/inmemoryfs
```
swolchok: ditto includedir

backends/apple/coreml/CMakeLists.txt
```diff
@@ -232,9 +247,10 @@ if(APPLE)
   endif()
 
   install(
-    TARGETS coremldelegate
+    TARGETS coremldelegate coreml_util coreml_inmemoryfs
+    EXPORT ExecuTorchTargets
     DESTINATION lib
     INCLUDES
-    DESTINATION ${_common_include_directories}
+    DESTINATION include
```
swolchok: ditto includedir

backends/xnnpack/CMakeLists.txt
```diff
@@ -102,28 +103,63 @@ target_link_libraries(
...
 )
 target_compile_options(xnnpack_backend PUBLIC ${_common_compile_options})
 executorch_target_link_options_shared_lib(xnnpack_backend)
 
+move_interface_include_directories_to_build_time_only(XNNPACK)
 install(
-  TARGETS xnnpack_backend
+  TARGETS xnnpack_backend xnnpack_schema
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok: should be ${CMAKE_INSTALL_LIBDIR}

backends/xnnpack/CMakeLists.txt
```diff
@@ -102,28 +103,63 @@ target_link_libraries(
...
+          xnnpack-operator-utils
+          xnnpack-pack-lh
+          xnnpack-packing
+          xnnpack-sanitizers
+          xnnpack-subgraph
+          xnnpack-datatype
+          xnnpack-reference-ukernels
+          xnnpack-logging
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok: ditto LIBDIR

extension/llm/custom_ops/CMakeLists.txt
```diff
@@ -35,9 +35,12 @@ extract_sources(${EXECUTORCH_SRCS_FILE})
...
 include(${EXECUTORCH_SRCS_FILE})
 
-# Let files say "include <executorch/path/to/header.h>".
-set(_common_include_directories ${EXECUTORCH_ROOT}/..)
-list(APPEND _common_include_directories ${EXECUTORCH_ROOT}/third-party/ao)
+set(_common_include_directories
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..,${CMAKE_CURRENT_SOURCE_DIR}/runtime/core/portable_type/c10>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR},${CMAKE_INSTALL_INCLUDEDIR}/executorch/runtime/core/portable_type/c10>
+)
+# REVIEW: are we installing AO? will this mess up EXPORT?
```
swolchok: remove REVIEW comment

extension/llm/runner/CMakeLists.txt
```diff
@@ -37,33 +37,42 @@ include(${EXECUTORCH_SRCS_FILE})
...
 target_include_directories(
   extension_llm_runner
   INTERFACE ${_common_include_directories}
-            ${EXECUTORCH_ROOT}/extension/llm/tokenizers/include
 )
 
+install(
+  TARGETS extension_llm_runner
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok: CMAKE_INSTALL_LIBDIR

extension/llm/runner/CMakeLists.txt
```diff
@@ -37,33 +37,42 @@ include(${EXECUTORCH_SRCS_FILE})
...
+  INCLUDES
+  DESTINATION ${_common_include_directories}
+)
+install(
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
+  DESTINATION include/executorch/extension/llm/runner
```
swolchok: CMAKE_INSTALL_INCLUDEDIR

extension/llm/runner/CMakeLists.txt
```diff
@@ -37,33 +37,42 @@ include(${EXECUTORCH_SRCS_FILE})
...
+  FILES_MATCHING
+  PATTERN "*.h"
+)
+# TODO: remove this once we create a proper CMake setup for sampler.
+install(
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/../sampler/
+  DESTINATION include/executorch/extension/llm/sampler
```
swolchok: ditto

kernels/portable/cpu/util/CMakeLists.txt
```diff
@@ -32,3 +32,9 @@ target_compile_definitions(
   kernels_util_all_deps PUBLIC C10_USING_CUSTOM_GENERATED_MACROS
 )
 target_compile_options(kernels_util_all_deps PUBLIC ${_common_compile_options})
+
+install(
+  TARGETS kernels_util_all_deps
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok: CMAKE_INSTALL_LIBDIR

schema/CMakeLists.txt
```diff
@@ -63,3 +65,11 @@ generate_program_schema("${common_schema_srcs}" "common_schema")
 set(program_schema_srcs program.fbs)
 generate_program_schema("${program_schema_srcs}" "program_schema")
 add_dependencies(program_schema common_schema)
+
+install(
+  TARGETS common_schema program_schema
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok: CMAKE_INSTALL_LIBDIR

tools/cmake/Utils.cmake
```diff
@@ -278,3 +278,36 @@ macro(find_package_torch)
     find_package(Torch CONFIG REQUIRED)
   endif()
 endmacro()
+
+# Modify ${targetName}'s INTERFACE_INCLUDE_DIRECTORIES by wrapping each entry in
+# $<BUILD_INTERFACE:...> so that they work with CMake EXPORT.
+function(move_interface_include_directories_to_build_time_only targetName)
```
swolchok: namespace these with `executorch_`

tools/cmake/executorch-config.cmake
swolchok: not during the ExecuTorch build it doesn't

swolchok: oh, hm, you're right. I guess it's just that the toolchain file doesn't prevent find_library from finding the system's libdl; will update the comment

swolchok:
This is getting very close to green CI if it's not there already, so marking ready for review.

GregoryComer (approved):

swolchok:
looks like ARM size tests need their bump increased and otherwise this looks very promising

swolchok:
finally 100% green CI!

PR #8958, opened 2025-03-05
[XNNPACK] resolve ambiguity around 2d affine quantized tensors

mcr229:
### Summary
There is some ambiguity around deriving per_token and per_channel_group semantics from quantize affine. Specifically when rank is two. Take for example the following input shape with the following group size:
- input_shape (5, 2048)
- group_size (1, 2048)

This tensor has 5 scales, and can be seen has having one scale per batch dimension. Or one scale per token. However, this can also be seen as a single weight in which each group is the size of a channel, effectively giving per_channel semantics. This ambiguity does not play well within the XNNPACK Backend as we are must parse these differing quantization types. For now we rely on the fact that per_token quantization happens dynamically. Meaning that the scales and zero points are dynamically choosen. As a result, we check that the scales come from getitem and is dynamically chosen. We further ensure that per_channel_group checks are not per_token. 

### Test plan
```
python -m unittest backends.xnnpack.test.ops.test_linear.TestLinear.test_linear_qd8_f32_per_token_weight_per_channel_group_int4
```

### DIFF
--- a/backends/xnnpack/operators/op_dynamic_dequantize_ops.py
+++ b/backends/xnnpack/operators/op_dynamic_dequantize_ops.py
@@ -13,6 +13,7 @@ from executorch.backends.xnnpack.operators.node_visitor import (
 )
 from executorch.backends.xnnpack.serialization.xnnpack_graph_schema import XNNGraph
 from executorch.backends.xnnpack.utils.quant_utils import (
+    is_dynamic_qdq,
     is_per_channel_group,
     is_per_token,
 )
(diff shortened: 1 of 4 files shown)


jackzhxng (approved):
This is needed to resolve numerical inaccuracies introduced by switching to the `quantize_` API here: https://github.com/pytorch/executorch/pull/8772. Will leave to @digantdesai / @kimishpatel for final stamp

GregoryComer (approved):

PR #8963, opened 2025-03-05
Arm backend: Add WhyNoPartitionReporter and report rejected nodes

Erik-Lundell:
With more complex checks for if nodes are supported or not, it can probably be complicated for a user to understand why a ceratin node was rejected from partitioning.
The WhyNoPartitionReporter attempts to mitigate this by producing a table with all rejected nodes with a reason for why they were rejected.

The reasoning behind using a class rather than log statements is
- The table keeps all reject information in the same place.
- Uniform formatting.
- Only log the first reject report of a node.
- Easier to change how and when the output is logged/dumped to file.


zingo (approved):

exir/backend/utils.py
```diff
@@ -569,3 +572,81 @@ def __call__(self, node: torch.fx.Node, reason: str) -> None:
 
     def __str__(self) -> str:
         return f"WhyNoPartition: Node {self.node} was not partitioned because {self.reason}."
+
+
+class WhyNoPartitionReporter:
```
mergennachin, line 577: cc @mcr229, @cccclai  - can you take a look at thi?

cccclai: @Erik-Lundell can you start a separate PR with this change so it's easier for discussion.

exir/backend/utils.py
```diff
@@ -569,3 +572,81 @@ def __call__(self, node: torch.fx.Node, reason: str) -> None:
...
+
+        # In your backend partitioner file(s)
+        reporter = WhyNoPartitionReporter()
+
+        # hypothetical function that checks if a node can be lowered
+        if not can_be_lowered(node):
+            reporter.report_reject(node, "This node was not lowered because ...")
+
+        # Back in partitioner
+        logger.info(reporter.get_table_report())
```
mcr229, line 592: do you have an example of what the table report looks like?

mcr229, line 592: do you have an example of what this looks like

exir/backend/utils.py
```diff
@@ -569,3 +572,81 @@ def __call__(self, node: torch.fx.Node, reason: str) -> None:
...
+            )
+        if len(reject_report) > 0:
+            return (
+                tabulate(
+                    reject_report,
+                    ["Name", "Target", "Torch func", "Reason"],
+                    tablefmt="fancy_grid",
+                    maxcolwidths=35,
+                )
+                + "\n(Placeholders and outputs are not included in this list)"
```
mcr229, line 623: why omit placeholders?

exir/backend/utils.py
```diff
@@ -569,3 +572,81 @@ def __call__(self, node: torch.fx.Node, reason: str) -> None:
...
+            )
+        else:
+            return "Partitioned all operators."
```
mcr229, line 626: Perhaps better to say "No Operators Rejected" Since partitioner doesn't necessarily partition all operators in this case.

mcr229 (commented):
In general this looks good to me, and is what i'm hoping our partitioner logging infra can look like moving forward. 

Would love to see placeholder tensors reported in the future, but that can be added later i suppose.

exir/backend/utils.py
Erik-Lundell: Added a docstring to `.get_table_report()` with an example. The format was inspired by `devtools/backend_debug/delegation_info.py::print_delegation_info()` for consistency.

exir/backend/utils.py
digantdesai: can we make this as a util which can be shared by all the delegates? I wanted to do that when we first introduced it in xnnpack.

zingo:
MacOs fails unrelatred

PR #8712, opened 2025-02-26
Use runtime.python_library in our BUCK/TARGETS files

swolchok:
IIUC, we should've been doing this all along for compatibility with OSS Buck. Was blocking `buck2 build //runtime/...` (which is still failing, but differently).

Differential Revision: [D70647459](https://our.internmc.facebook.com/intern/diff/D70647459)

PR #8969, opened 2025-03-05
Default ExecuTorch targets to ExecuTorch-wide Buck visibility

swolchok:
IMO, Buck visibility is just inverse deps. We should trust that people have a good reason to add deps rather than attempt to police them and require double entry in both `deps` and `visibility`, especially since we seem to be committed to APIs by default in OSS anyway.

Specific motivation is that #8712 would otherwise have to ad-hoc slap ExecuTorch-wide visibility on a lot of targets, but I've held this view for a long time.

Differential Revision: [D70647462](https://our.internmc.facebook.com/intern/diff/D70647462)

### DIFF
--- a/shim_et/xplat/executorch/build/runtime_wrapper.bzl
+++ b/shim_et/xplat/executorch/build/runtime_wrapper.bzl
@@ -187,6 +187,9 @@ def _patch_kwargs_common(kwargs):
     for dep_type in ("deps", "exported_deps"):
         env.patch_deps(kwargs, dep_type)
 
+    if "visibility" not in kwargs:
+        kwargs["visibility"] = ["//executorch/..."]
+
     # Patch up references to "//executorch/..." in lists of build targets,
     # if necessary.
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

mergennachin:
Can we make this consistent both internally in fbsource and shim layer? 

Otherwise, a package could be potentially private in fbsource but visible to ET in OSS, which could lead to diff train issues

mergennachin:
Oh wait, is this file `shim_et/xplat/executorch/build/runtime_wrapper.bzl` actually the same file used both internally and externally?

mergennachin (approved):

swolchok:
@swolchok has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D70647462).

swolchok:
> Oh wait, is this file `shim_et/xplat/executorch/build/runtime_wrapper.bzl` actually the same file used both internally and externally?

yes.

PR #4993, opened 2024-08-29
[Build] Set correct flatcc path for Windows

python3kgae:
The binary directory for Windows will have build type like Debug/Release at the end.

For #4661

PR #8970, opened 2025-03-05
Fix flatcc path for Windows

SS-JIA:
### Summary
This PR is based on https://github.com/pytorch/executorch/pull/4993 but includes an additional fix to use the correct equivalent of the `rm -rf` command for Windows.

### Test plan

Built ExecuTorch for Windows using the following command:

```
del -Recurse -Force cmake-out; `
  cmake . `
  -DCMAKE_INSTALL_PREFIX=cmake-out `
  -DPYTHON_EXECUTABLE=C:\\Users\\ssjia\\AppData\\Local\\miniconda3\\python.exe `
  -DCMAKE_PREFIX_PATH=C:\\Users\\ssjia\\AppData\\Local\\miniconda3\\Lib\\site-packages `
  -DCMAKE_BUILD_TYPE=Release `
  -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON `
  -T ClangCL `
  -Bcmake-out; `
  cmake --build cmake-out -j64 --target install
```

### DIFF
--- a/devtools/CMakeLists.txt
+++ b/devtools/CMakeLists.txt
@@ -20,7 +20,11 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(NOT FLATCC_EXECUTABLE)
-  set(FLATCC_EXECUTABLE ${_flatcc_source_dir}/bin/flatcc)
+  if(WIN32)
+    set(FLATCC_EXECUTABLE ${_flatcc_source_dir}/bin/${CMAKE_BUILD_TYPE}/flatcc)
+  else()
+    set(FLATCC_EXECUTABLE ${_flatcc_source_dir}/bin/flatcc)
+  endif()
 endif()
 
 # Source root directory for executorch.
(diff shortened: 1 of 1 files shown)


devtools/CMakeLists.txt
```diff
@@ -159,6 +163,15 @@ file(MAKE_DIRECTORY
      ${_program_schema__include_dir}/executorch/devtools/bundled_program
 )
 
+message("Hello")
+message("${CMAKE_GENERATOR}")
```
tarun292, line 167: Remove these?

devtools/CMakeLists.txt
```diff
@@ -159,6 +163,15 @@ file(MAKE_DIRECTORY
...
+
+if(WIN32)
+  set(RM_COMMAND rmdir /s /q)
+else()
+  set(RM_COMMAND rm -rf)
```
tarun292, line 172: rm -rf in a cmakefile makes me a little uneasy, why do we need this?

devtools/CMakeLists.txt
SS-JIA: oh, good catch :p

devtools/CMakeLists.txt
tarun292: ah ok looks like it was there before, so hopefully should be fine i guess

SS-JIA:
* Removed debug logging
* Removed force flags from `rm` commands

cc: @tarun292

tarun292 (approved):

SS-JIA:
test failure is most probably unrelated.

PR #8971, opened 2025-03-05
Fix phi4mini test model

jackzhxng:


cccclai:
Seems like there are still failing CI, not quite sure if they're all relavant

cccclai (approved):
Thank you for the fix!

PR #8974, opened 2025-03-05
[Fix bug] always run prepare-test-spec

yangw-dev:
Follow up with https://github.com/pytorch/executorch/issues/8125

Example of failed benchmark run: https://github.com/pytorch/executorch/actions/runs/13666476394

We made prepare-test-spec have dependency on export-model, it seems like when there is any failure happens in the export model step, it will skip the whole prepare-test-spec. We want to always run prepare-test-spec.

in step prepare-test-spec, we will check wether the export-model successfully build the model, if not, no test-spec will be uploaded. This makes the final step mobile-on-device step fails early at verify-test-spec part

### DIFF
--- a/.github/workflows/android-perf.yml
+++ b/.github/workflows/android-perf.yml
@@ -279,6 +279,7 @@ jobs:
         echo "::endgroup::"
 
   prepare-test-specs:
+    if: always()
     runs-on: linux.2xlarge
     needs:
       - set-parameters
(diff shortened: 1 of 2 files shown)


huydhn (changes requested):
Oops, I think we should revert https://github.com/pytorch/executorch/pull/8482 instead because having `always()` here will negate https://github.com/pytorch/executorch/pull/8482.  In hindsight, we shouldn't make `prepare-test-spec` depend on `export-models`.  I forgot to tell you that the issue https://github.com/pytorch/executorch/issues/8125 has been fixed by https://github.com/pytorch/executorch/pull/8786 (I did it last minute before signing off on Thursday, my bad)

yangw-dev:
got it!

PR #8982, opened 2025-03-06
[minibench] Add RSS stats

kirklandsign:
### Summary
RSS can be used as an indicator for RAM usage. Let's collect it, but it might not be stable right now.
### Test plan
CI


kirklandsign:
@claude review

PR #8983, opened 2025-03-06
Split & remove extension_parallel

swolchok:
As per plan in #8932, we want to be able to include thread_parallel.h to build
libraries that are *capable* of parallelization, but don't *require* it. So, we
move the header to ExecuTorch core and add a fallback implementation (with
tests!) of `parallel_for` that just does a regular `for` loop. Then, targets
that link `extension_threadpool` will get parallelization automagically.

This PR doesn't add any optionally-parallelized code; that will be in the next
PR.


swolchok:
* (to be filled)

extension/parallel/thread_parallel.h
```diff
@@ -8,46 +8,9 @@
...
-using ::executorch::extension::get_thread_num;
-using ::executorch::extension::parallel_for;
-using ::executorch::extension::set_thread_num;
-} // namespace executor
-} // namespace torch
+// This header is a stub left behind after the move to
+// executorch/runtime/kernel. Depend on this target and include this
+// header if you have a hard requirement for threading; if you want to
+// cleanly use parallelization if available, then depend on and use
+// the below header instead.
```
swolchok, line 15: will shorten this comment to clarify that users should just include the below header instead in an upcoming revision.

extension/threadpool/CMakeLists.txt
```diff
@@ -21,7 +21,8 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 add_library(
-  extension_threadpool threadpool.cpp threadpool_guard.cpp cpuinfo_utils.cpp
+  extension_threadpool threadpool.cpp threadpool_guard.cpp thread_parallel.cpp
```
kimishpatel, line 24: nit: ok with this but thread_parallel provides interface to leverage underlying threadpool.

swolchok: I don't understand your point here. I can say that I originally left extension_parallel around, but it struck me as silly to make code depend on thread_parallel separately rather than getting it for free as a way to use the thread pool.

kimishpatel: > make code depend on thread_parallel separately rather than getting it for free as a way to use the thread pool

That also makes sense. I was making a bit of a pedantic point but no value in arguing over it.

extension/threadpool/test/thread_parallel_test.cpp
```diff
@@ -11,13 +11,16 @@
...
-#include <executorch/extension/parallel/thread_parallel.h>
+#include <executorch/runtime/kernel/thread_parallel_interface.h>
 #include <executorch/runtime/platform/platform.h>
 
 using namespace ::testing;
 using ::executorch::extension::parallel_for;
 
-class ParallelTest : public ::testing::Test {
+#ifndef ET_USE_THREADPOOL
+#endif
```
swolchok, line 21: whoops. will update to remove after I finish going through the rest of the stack to avoid excess churn on the later diffs

swolchok:
ping reviewers @larryliu0820 @kirklandsign @lucylq @jackzhxng @iseeyuan @GregoryComer @kimishpatel

JacobSzwejbka (approved):

PR #8987, opened 2025-03-06
Build optimized_portable_kernels if threadpool is enabled

swolchok:
This is step (5) of #8932.

At this exact moment, this rebuild is inefficient because it rebuilds the whole portable op library, but ops don't support optional parallelization just yet. This will become less true when we roll out parallel_for support across portable ops immediately following this PR.


swolchok:
* (to be filled)

swolchok:
Tried to verify whether this helps performance by running an internal model using 100 iterations per trial, 3 trials. mean time  before was 32.17s, after 30.91s. Speedup of 1.26s, or 3.9%. Noise was low compared to the improvement.

mcr229 (approved):

PR #8916, opened 2025-03-04
[Windows] [mmap_data_loader.cpp] mmap equivalent for Windows

SamGondelman:
### Summary
#5164 with the code suggestions applied.

### Test plan
I am not familiar with how to test this area of the code.  On non-Windows systems, it should be functionally the same.
  #5164 suggests a command to run some relevant tests?

PR #8989, opened 2025-03-06
fix mman header issues

SamGondelman:
### Summary
#8916 broke tests.  mman.h is needed by mmap_data_loader_test.cpp, and we need to use select

### Test plan
Unit tests - not sure why this didn't show on the first PR?


SS-JIA (approved):

SS-JIA:
Before merging, let's import to `fbsource` to verify that this fix makes internal CI green

PR #8991, opened 2025-03-06
Revert "[Windows] [mmap_data_loader.cpp] mmap equivalent for Windows"

SamGondelman:
Reverts pytorch/executorch#8916

this caused some internal test failures, reverting and then will apply a proper fix


cccclai (approved):
Thanks

PR #8997, opened 2025-03-06
Arm backend: Add FuseViewCopyTransform and FuseConstantsPass in arm_pass_manager

AdrianLundell:
These passes both removes redundant ops from the graph:
- FuseViewCopyTransform pass is added from backends/transforms to merge sequential view ops.
- FuseConstantOpsPass is created to compute ops with constant inputs AOT.
  - This is not done in cases where the result is a larger tensor, to avoid increasing the constant memory size.
  - For BI, ops are quantized with the q/dq-ops as to not change the behaviour of the graph.
  - Pass order is important: the pass must be placed after all passes which may add constant ops, but before the
    InsertTableOpsPass, since it doesn't handle TOSA _table-ops.

### DIFF
--- a/backends/arm/_passes/arm_pass_manager.py
+++ b/backends/arm/_passes/arm_pass_manager.py
@@ -51,6 +51,7 @@ from executorch.backends.arm._passes.fold_qdq_with_annotated_qparams_pass import
     RetraceFoldedDtypesPass,
 )
 from executorch.backends.arm._passes.fuse_batchnorm2d_pass import FuseBatchnorm2DPass
+from executorch.backends.arm._passes.fuse_constant_ops_pass import FuseConstantOpsPass
 from executorch.backends.arm._passes.fuse_quantized_activation_pass import (  # type: ignore[import-not-found]
     FuseQuantizedActivationPass,
 )
@@ -78,6 +79,7 @@ from executorch.backends.arm._passes.unsqueeze_scalar_placeholders_pass import (
     UnsqueezeScalarPlaceholdersPass,
 )
 from executorch.backends.arm.tosa_specification import TosaSpecification
+from executorch.backends.transforms.fuse_view_copy import FuseViewCopyTransform
 
 from executorch.backends.transforms.replace_scalar_with_tensor import (
     ReplaceScalarWithTensorArgPass,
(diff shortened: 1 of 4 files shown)


zingo (approved):

kirklandsign:
Sorry @zingo would you mind if we revert and land this again from internal? This is breaking our internal tests.

cc @zonglinpeng do you think a forward fix is possible?

zonglinpeng:
> [zonglinpeng](/zonglinpeng) pushed a commit that referenced this pull request [ Mar 6, 2025 ](#ref-commit-ca25c7f)
> [ ](/AdrianLundell)

I'm not familiar with arm passes. But based on the error message `ModuleNotFoundError: No module named 'executorch.backends.transforms.fuse_view_copy`, fuse_view_copy needs to added or change reference as the fix.

zingo:
Of course, if there is a problem let's revert so it works again, and we can figure out the problem much calmer. Unfortunately I'm not at my desk right now so I cannot do it right now. Feel free to go ahead if you can. 🙏

zingo:
Created https://github.com/pytorch/executorch/pull/9070 if still needed
I cant self review it so please feel free to review and merge if this is still an issue.

kirklandsign:
Hi @zingo 

We have the following internal error log:

```
        debug_mode = "ALL" if logger.level <= logging.DEBUG else None
>       outputs_np, status = tosa_reference_model.run(
            graph,
            inputs_np,
            verbosity=_tosa_refmodel_loglevel(logger.level),
            tosa_profile=tosa_profile,
            initialize_variable_tensor_from_numpy=1,  # True
            debug_mode=debug_mode,
        )
E       AttributeError: 'NoneType' object has no attribute 'run'

```

Do you have idea?

kirklandsign:
Hi @zingo 

Seems that all 3 models in this test https://github.com/pytorch/executorch/blob/main/backends/arm/test/passes/test_fuse_constant_ops_pass.py#L110-L115 failed

AdrianLundell:
Hi  @kirklandsign , can you give more context here:
1. The first error indicates that the tosa_reference_model is none, but it's impossible to say why without knowing  how it was created.
2. Where is the unittest failing, as they passed when running on the PR?
Thanks!

kirklandsign:
Hi @AdrianLundell 

Sorry I don't have the full context about how the CI is run on our internal CI. Maybe @digantdesai will give more context. 

I'm not sure why tosa_reference_model is none here.

PR #8998, opened 2025-03-06
Arm backend: Move ReplaceScalarTensorWithFullPass to transforms

mansnils:
The pass is general and can be used by multiple backends. The aten.scalar_tensor is replaced by a aten.full which is already supported by Arm backend.
Adds new method to Arm tester for getting output as the nn.module in the unit test does not take any input. The output is then manually compared within the unit test.


shoumikhin:
@DigantDesai, @KimishPatel can you take a look please?

kimishpatel (approved):
Looks like just moving a pass around. Didnt fully follow but accepting it. cc @mcremon-meta for cadence related code movement. Bunch of tests are failing though. @tarun292 does it need to be imported internally to validate? Please wait for @tarun292 's answer before landing

kimishpatel:
> looks fine to share the code from our end! We would just ask that (like in this PR) the passes are not modified at least in the PR that moves them so that we can keep track of changes :)

Looks like you are ok with landing these then.

backends/transforms/replace_scalar_tensor_with_full.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: Thank you this is the right thing to do.

byjlw:
@mansnils this branch needs to be rebased so and conflicts need to be resolved so we can merge

mansnils:
This has been replaced with [another PR, so closing this](https://github.com/pytorch/executorch/pull/9711)

PR #9044, opened 2025-03-07
Make serial parallel_for "polyfill" iterate backwards in debug builds

swolchok:
Catches users who assume it goes forwards in order (like my attempt to use parallel_for in reductions, whoops).


swolchok:
* (to be filled)

JacobSzwejbka (approved):
The debug loop looks fine to me, though do we have much other NDEBUG ifdefs in the codebase? Is that a mode people often use with ET?

swolchok:
>  Is that a mode people often use with ET?

Release builds define NDEBUG. it's the same macro that gates `assert`.

PR #9053, opened 2025-03-07
[cadence][g3] enable fp32 permute

zonglinpeng:
Summary: titled

Differential Revision: D70796738


backends/cadence/fusion_g3/operators/op_permute_copy.cpp
```diff
@@ -156,4 +157,4 @@ Tensor& permute_copy_out(
 } // namespace native
 } // namespace G3
 } // namespace impl
-} // namespace cadence
\ No newline at end of file
+} // namespace cadence
```
mcremon-meta, line 160: if linter is ok, probably doesn't matter :)

mcremon-meta (approved):

PR #9056, opened 2025-03-07
Split delinearize_index out of broadcast_util

swolchok:
Left an #include behind for backward compatibility. Just needed so that broadcast_indexes_range.h can use delinearize_index


swolchok:
* (to be filled)

kernels/portable/cpu/util/delinearize_index.h
```diff
@@ -0,0 +1,43 @@
...
+ * @param[in] linear_index The flattened index
+ * @param[in] shape The tensor shape
+ * @param[out] out_indexes The per-dimension indexes
+ * @param[in] out_indexes_len The maximum size of the out_indexes array
+ * @returns void
+ */
+void delinearize_index(
+    size_t linear_index,
+    executorch::aten::ArrayRef<Tensor::SizesType> shape,
+    size_t* out_indexes,
```
JacobSzwejbka, line 26: we have Span fwiw.

extension/threadpool/test/CMakeLists.txt
```diff
@@ -12,28 +14,14 @@
 #
 
 cmake_minimum_required(VERSION 3.19)
-project(extension_parallel_test)
-
-# Use C++17 for test.
-set(CMAKE_CXX_STANDARD 17)
```
JacobSzwejbka, line 18: Why'd the standard line go away?

extension/parallel/thread_parallel.h
```diff
@@ -8,46 +8,7 @@
...
-namespace executor {
-// TODO(T197294990): Remove these deprecated aliases once all users have moved
-// to the new `::executorch` namespaces.
-using ::executorch::extension::get_thread_num;
-using ::executorch::extension::parallel_for;
-using ::executorch::extension::set_thread_num;
-} // namespace executor
-} // namespace torch
+// This header is a stub left behind after the move to
+// executorch/runtime/kernel. As such, it is deprecated; include and
```
JacobSzwejbka, line 12: were any of our tutorials using thread_parallel.h?

JacobSzwejbka (approved):

extension/threadpool/test/CMakeLists.txt
swolchok: we don't need it

extension/parallel/thread_parallel.h
swolchok: fbgs says no

kernels/portable/cpu/util/delinearize_index.h
swolchok: Not particularly interested in changing API here because 1) out of scope for this diff 2) might break backward compatibility 3) don't see a reason to migrate between the two equivalent reference-to-array classes.

PR #9058, opened 2025-03-07
Use template instead of duplicating code in elementwise_util.h

swolchok:
Now all the apply functions share a common implementation, which means further changes (e.g., parallel_for, generating specialized dtypes for the case where all inputs have the same type) don't need to be repeated 3 times.

(Interestingly, this seems to increase the effectiveness of the following parallelization change. Not entirely sure why, but I checked the generated code for optimized op_where and it seems to have improved, which is surprising.)


swolchok:
* (to be filled)

JacobSzwejbka (approved):

PR #9069, opened 2025-03-08
Include custom ops in xnn_executor_runner if they're built

swolchok:
(We should probably merge xnn_executor_runner and executor_runner if there's no particular use for a portable-only runner, but hey this is an incremental improvement.)


swolchok:
* (to be filled)

backends/xnnpack/CMakeLists.txt
```diff
@@ -124,6 +124,10 @@ else()
   list(APPEND xnn_executor_runner_libs portable_ops_lib)
 endif()
 
+if(EXECUTORCH_BUILD_KERNELS_CUSTOM)
+  list(APPEND xnn_executor_runner_libs custom_ops)
```
iseeyuan, line 128:
```suggestion
  list(APPEND xnn_executor_runner_libs $<LINK_LIBRARY:WHOLE_ARCHIVE,custom_ops>)
```

iseeyuan: The above diff works for me.

swolchok: fixed

iseeyuan (approved):
Let's land it!

PR #9076, opened 2025-03-10
fix lintrunner

cccclai:
Fix the lint runner error from https://github.com/pytorch/executorch/pull/8857

### DIFF
--- a/backends/cadence/aot/remove_ops.py
+++ b/backends/cadence/aot/remove_ops.py
@@ -808,7 +808,9 @@ class RemoveBranchedQuantDequant(ExportPass):
 
 
 class RemoveCatFromSliceCopyPass(ExportPass):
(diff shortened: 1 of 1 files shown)


kirklandsign (approved):

zonglinpeng (approved):
Ah good to know the trick! Thought "advice" wouldnt break global lint. Thank you for the fix

backends/cadence/aot/remove_ops.py
```diff
@@ -808,7 +808,9 @@ def remove_branched(
 
 
 class RemoveCatFromSliceCopyPass(ExportPass):
-    def _remove_unused_cat(self, graph_module: torch.fx.GraphModule) -> None:
+    def _remove_unused_cat(  # noqa: C901
+        self, graph_module: torch.fx.GraphModule
+    ) -> None:
```
larryliu0820, line 813: Would also be good if we can actually reduce the complexity of this function

PR #9077, opened 2025-03-10
Build flatc for the host instead of the target platform

### LINKED ISSUE #7260: Flatc should be built for the host rather than the target when cross-compiling
### 🐛 Describe the bug

https://github.com/pytorch/executorch/blob/a6841d565f707fc5953d61a008ab17fd75023266/CMakeLists.txt#L478

It could be fixed with something like this.
``` diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 46fec4cf0..62dfa6e62 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -475,12 +475,24 @@ if(EXECUTORCH_BUILD_FLATC)
       OFF
       CACHE BOOL ""
   )
-  add_subdirectory(third-party/flatbuffers)
 
-  # exir lets users set the alignment of tensor data embedded in the flatbuffer,
-  # and some users need an alignment larger than the default, which is typically
-  # 32.
-  target_compile_definitions(flatc PRIVATE FLATBUFFERS_MAX_ALIGNMENT=1024)
+  if(CMAKE_CROSSCOMPILING)
+    set(FLATC_EXECUTABLE  ${CMAKE_BINARY_DIR}/_host_build_flatbuffers/flatc)
+    execute_process(
+      COMMAND
+        ${CMAKE_COMMAND} ${CMAKE_CURRENT_SOURCE_DIR}/third-party/flatbuffers -DCMAKE_C_FLAGS=-DFLATBUFFERS_MAX_ALIGNMENT=1024 -B${CMAKE_BINARY_DIR}/_host_build_flatbuffers
+    )
+    execute_process(
+      COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}/_host_build_flatbuffers -j
+    )
+  else()
+    add_subdirectory(third-party/flatbuffers)
+
+    # exir lets users set the alignment of tensor data embedded in the flatbuffer,
+    # and some users need an alignment larger than the default, which is typically
+    # 32.
+    target_compile_definitions(flatc PRIVATE FLATBUFFERS_MAX_ALIGNMENT=1024)
+  endif()
 endif()
 if(NOT FLATC_EXECUTABLE)
   message(
```

### Versions

``` yaml
PyTorch version: 2.6.0.dev20241112+cpu
Is debug build: False
CUDA used to build PyTorch: Could not collect

OS: Ubuntu 24.04.1 LTS (x86_64)
GCC version: (Ubuntu 13.2.0-23ubuntu4) 13.2.0
Clang version: Could not collect
CMake version: version 3.31.1
Libc version: glibc-2.39

Python version: 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] (64-bit runtime)
Python platform: Linux-6.8.0-48-generic-x86_64-with-glibc2.39
Is CUDA available: False
CUDA runtime version: 11.5.119
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.5.0
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        43 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               12
On-line CPU(s) list:                  0-11
Vendor ID:                            AuthenticAMD
Model name:                           AMD Ryzen 5 1600 Six-Core Processor
CPU family:                           23
Model:                                1
Thread(s) per core:                   2
Core(s) per socket:                   6
Socket(s):                            1
Stepping:                             1
Frequency boost:                      enabled
CPU(s) scaling MHz:                   51%
CPU max MHz:                          3200.0000
CPU min MHz:                          1550.0000
BogoMIPS:                             6388.26
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw skinit wdt tce topoext perfctr_core perfctr_nb bpext perfctr_llc mwaitx cpb hw_pstate ssbd vmmcall fsgsbase bmi1 avx2 smep bmi2 rdseed adx smap clflushopt sha_ni xsaveopt xsavec xgetbv1 clzero irperf xsaveerptr arat npt lbrv svm_lock nrip_save tsc_scale vmcb_clean flushbyasid decodeassists pausefilter pfthreshold avic v_vmsave_vmload vgif overflow_recov succor smca sev
L1d cache:                            192 KiB (6 instances)
L1i cache:                            384 KiB (6 instances)
L2 cache:                             3 MiB (6 instances)
L3 cache:                             16 MiB (2 instances)
NUMA node(s):                         1
NUMA node0 CPU(s):                    0-11
Vulnerability Gather data sampling:   Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Not affected
Vulnerability Reg file data sampling: Not affected
Vulnerability Retbleed:               Mitigation; untrained return thunk; SMT vulnerable
Vulnerability Spec rstack overflow:   Vulnerable: Safe RET, no microcode
Vulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Retpolines; STIBP disabled; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Not affected

[pip3] numpy==2.2.0
[pip3] torch==2.6.0.dev20241112+cpu
[pip3] torchaudio==2.5.0.dev20241112+cpu
[pip3] torchsr==1.0.4
[pip3] torchvision==0.20.0.dev20241112+cpu
[conda] numpy                     1.21.6                   pypi_0    pypi
[conda] torch                     1.12.1                   pypi_0    pypi
[conda] torchvision               0.9.0                    pypi_0    pypi
```

cc @JacobSzwejbka @dbort @larryliu0820

jathu:
### Summary

* Fixes https://github.com/pytorch/executorch/issues/7260
    * We use flatc as part of the build step to generate some files from FlatBuffer definitions. This implies the tool runs on the host. However, we currently build flatc as part of the main build — which propagates the target CXX flags to flatc. Thus, flatc gets built targeting the target platform. To ensure flatc is built for the host, we can use include flatc as an [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html). This does not propagate the CXX flags
* Some targets implicitly depended on flatc, we now make that requirement explicit
* We currently spread the `FLATC_EXECUTABLE` defaulting across the project. Let's just centralize this at the root

### Test plan

```bash
$ ./install_executorch.sh

# Previously flatc was built against the Android target, now they target the host
$ ./build/build_android_library.sh
$ file /Users/jathu/executorch/cmake-out-android-arm64-v8a/third-party/flatbuffers/flatc
/Users/jathu/executorch/cmake-out-android-arm64-v8a/third-party/flatbuffers/flatc: Mach-O 64-bit executable arm64

# Apple builds work as usual, but they use flatc from pip
$ ./build/build_apple_frameworks.sh
```

cc @larryliu0820 @lucylq @swolchok @dbort


CMakeLists.txt
```diff
@@ -474,6 +474,17 @@ cmake_dependent_option(
   "NOT FLATC_EXECUTABLE;EXECUTORCH_BUILD_HOST_TARGETS" OFF
 )
 
+
+set(FLATBUFFERS_BUILD_FLATC OFF CACHE BOOL "")
```
larryliu0820, line 478: Do you think it makes sense to have a separate file `flatc.cmake` to include flatc only logic? This file can live in `build/` for example. Then we can do `include build/flatc.cmake` in the root `CMakeLists.txt`

jathu: I want to do a broader CMake clean up, so let's do this in a follow up?

CMakeLists.txt
```diff
@@ -482,41 +493,40 @@ if(EXECUTORCH_BUILD_FLATC)
...
+    flatc
+    PREFIX ${CMAKE_CURRENT_BINARY_DIR}/third-party/flatbuffers
+    BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/third-party/flatbuffers
+    SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/third-party/flatbuffers
+    CMAKE_ARGS -DFLATBUFFERS_BUILD_FLATC=ON
+               -DFLATBUFFERS_BUILD_FLATHASH=${FLATBUFFERS_BUILD_FLATHASH}
+               -DFLATBUFFERS_BUILD_FLATLIB=${FLATBUFFERS_BUILD_FLATLIB}
+               -DFLATBUFFERS_BUILD_TESTS=${FLATBUFFERS_BUILD_TESTS}
+               -DFLATBUFFERS_INSTALL=${FLATBUFFERS_INSTALL}
+               -DCMAKE_CXX_FLAGS="-DFLATBUFFERS_MAX_ALIGNMENT=${FLATBUFFERS_MAX_ALIGNMENT}"
```
swolchok, line 509: shouldn't we -DCMAKE_BUILD_TYPE=Release?

swolchok (approved):
looks great, just the one minor comment

CMakeLists.txt
```diff
@@ -482,41 +493,40 @@ if(EXECUTORCH_BUILD_FLATC)
...
-      OFF
-      CACHE BOOL ""
-  )
-  add_subdirectory(third-party/flatbuffers)
 
-  # exir lets users set the alignment of tensor data embedded in the flatbuffer,
-  # and some users need an alignment larger than the default, which is typically
-  # 32.
-  target_compile_definitions(flatc PRIVATE FLATBUFFERS_MAX_ALIGNMENT=1024)
+  # Build flatc for the *host* to generate files as part of the build step.
```
mergennachin, line 497: Don't you wanna cache so that it doesn't build all the time? 

```
find_program(FLATC_EXECUTABLE flatc)
if(FLATC_EXECUTABLE)
  message(STATUS "Found flatc: ${FLATC_EXECUTABLE}")
else()
  include(ExternalProject)
  ExternalProject_Add(...)
endif()
```

backends/apple/mps/CMakeLists.txt
```diff
@@ -50,6 +46,7 @@ add_custom_command(
     "${_mps_schema__include_dir}/executorch/backends/apple/mps"
     ${_mps_schema__srcs}
   WORKING_DIRECTORY ${EXECUTORCH_ROOT}
+  DEPENDS flatc
```
mergennachin, line 49: Claude Sonnet 3.7 tells me this:

"
- flatc is the name of the ExternalProject target, not the actual executable file. Using DEPENDS flatc only tells CMake that your command depends on the ExternalProject's build process completing, but doesn't establish the dependency on the actual executable file. 

- In parallel builds, CMake might try to run your command before the executable is actually available, even if the ExternalProject build process has technically "completed."
"

So, I would add 

```
  add_executable(flatc_target IMPORTED GLOBAL)
  set_property(TARGET flatc_target PROPERTY IMPORTED_LOCATION ${FLATC_EXECUTABLE})
  add_dependencies(flatc_target flatc)
```

in root level CMakeLists.txt

and instead of

`DEPENDS flatc` do `DEPENDS flatc_target`

jathu: Nice find! Yes, this is currently just relying on the project instead of the executable directly. I was able to use your code snippet to directly rely on the binary instead now

CMakeLists.txt
swolchok: @mergennachin 1) will that work? 2) I had hoped ExternalProject would do something reasonable in this regard; shouldn't cmake notice the CMAKE_ARGS haven't changed and not rebuild? have you tried it?

jathu: I also assumed the underlying build system (Make, Ninja) would already cache the targets after building it once

swolchok: hmmm, no change in this regard

jathu: @swolchok how are you testing this? repro steps?

swolchok: > cleanup BINARY_DIR or INSTALL_DIR in between installs there won't be any caching behavio

Sure, but why would you do that?

swolchok: > how are you testing this?

```
$ rm -rf cmake-out
$ time cmake . \
                                                                 -DCMAKE_INSTALL_PREFIX=cmake-out \
                                                                 -DEXECUTORCH_BUILD_KERNELS_CUSTOM=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_CUSTOM_AOT=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_OPTIMIZED=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_QUANTIZED=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
                                                             -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
                                                                 -DEXECUTORCH_BUILD_DEVTOOLS=ON \
                                                                 -DEXECUTORCH_BUILD_XNNPACK=ON \
                                                                 -DEXECUTORCH_BUILD_TESTS=ON \
                                                             -DCMAKE_BUILD_TYPE=Release \
                                                             -DEXECUTORCH_ENABLE_LOGGING=ON \
                                                                 -Bcmake-out -GNinja
```
(this is just my normal cmake invocation for executorch development)

then run the cmake command a second time

jathu: Ah I see you're timing the cmake generation. Tbh I don't think there is much value trying to cache optimize this step. For the actual build, Make/Ninja will cache the targets — so I think we can punt on this

mergennachin (approved):

mergennachin:
@jathu I added 'ciflow/trunk' label, which triggers all trunk jobs

Looks like there are a bunch of CI failures

Please fix them.

jathu:
These failing tests are unrelated to this diff:

* [build-benchmark-app / macos-job](https://github.com/pytorch/executorch/actions/runs/13839932185/job/38724605861?pr=9077#logs) — ` fatal error: 'omp.h' file not found` is occurring in main
* [test-arm-reference-delegation / linux-job](https://github.com/pytorch/executorch/actions/runs/13839933547/job/38724561738?pr=9077#logs) — occurring in main
* [test-models-macos (cmake, linear, xnnpack-quantization-delegation, macos-m1-stable, 90) / macos-job](https://github.com/pytorch/executorch/actions/runs/13839933547/job/38724616452?pr=9077#logs) — similar `omp.h` problem as above
* [test-static-llama-ane / macos-job](https://github.com/pytorch/executorch/actions/runs/13839933547/job/38724564319?pr=9077#logs) — incorrect python version issue also occurring in main

swolchok:
Now seeing this:
```
$ cmake . \
                                                                 -DCMAKE_INSTALL_PREFIX=cmake-out \
                                                                 -DEXECUTORCH_BUILD_KERNELS_CUSTOM=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_CUSTOM_AOT=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_OPTIMIZED=ON \
                                                                 -DEXECUTORCH_BUILD_KERNELS_QUANTIZED=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL=ON \
                                                                 -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
                                                             -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
                                                                 -DEXECUTORCH_BUILD_DEVTOOLS=ON \
                                                                 -DEXECUTORCH_BUILD_XNNPACK=ON \
                                                                 -DEXECUTORCH_BUILD_TESTS=ON \
                                                             -DCMAKE_BUILD_TYPE=Release \
                                                             -DEXECUTORCH_ENABLE_LOGGING=ON \
                                                                 -Bcmake-out -GNinja
$  cmake --build cmake-out
ninja: error: 'third-party/flatbuffers/flatc', needed by 'schema/include/executorch/schema/program_generated.h', missing and no known rule to make it
```

Repros on this commit and not the previous one.

swolchok:
Ninja is extra-picky about dependencies. #9246 fixes

PR #9083, opened 2025-03-10
Arm backend: Make build script source the generated setup_path.sh

zingo:
This makes it more easy to run the scripts from various tools like your editor.


zingo:
Lintrunner and MacOS fails are unrelated

freddan80 (approved):

PR #9093, opened 2025-03-10
Bump memory for phi-4-mini test_model

jackzhxng:
Attempt to fix test on trunk where mac is passing but linux is failing, e.g. https://github.com/pytorch/executorch/actions/runs/13770685177/job/38508355264


SS-JIA (approved):
Pre-emptive stamp in case failing test passes.

PR #9095, opened 2025-03-10
Support MPS Backend also on iOS < 16 (#9089)

f-meloni:
Summary: Pull Request resolved: https://github.com/pytorch/executorch/pull/9089

Differential Revision: D70795041

The MPS backend currently does not compile for iOS < 16.
I have added some conditional code to make it usable anyway

### DIFF
--- a/backends/apple/mps/runtime/MPSDevice.mm
+++ b/backends/apple/mps/runtime/MPSDevice.mm
@@ -22,11 +22,11 @@ static inline MTLLanguageVersion getMetalLanguageVersion(const id<MTLDevice>& de
   // MPS Advanced Indexing needs at least Metal 2.0 (support for Argument Buffers and function constants)
   // host_name attribute needs at least Metal 2.2 and ulong needs Metal 2.3 (supported on MacOS 11+)
   MTLLanguageVersion languageVersion = MTLLanguageVersion2_3;
-#if defined(__MAC_13_0)
-  if (macOS13Plus) {
-    languageVersion = MTLLanguageVersion3_0;
+  if (@available(iOS 16, macOS 13, *)) {
+    if (macOS13Plus) {
+      languageVersion = MTLLanguageVersion3_0;
+    }
   }
-#endif
 
   ET_CHECK_MSG([device supportsFamily:MTLGPUFamilyMac2], "Missing Metal support for MTLGPUFamilyMac2");
   return languageVersion;
--- a/backends/apple/mps/runtime/operations/IndexingOps.mm
+++ b/backends/apple/mps/runtime/operations/IndexingOps.mm
@@ -206,25 +206,32 @@ MPSGraphBuilder::mpsIndexPutOp(NodePtr nodePtr) {
 
 Error
 MPSGraphBuilder::mpsScatterOp(NodePtr nodePtr) {
-  auto graphNode = nodePtr->mpsnode_union_as_MPSScatter();
-  ET_LOG(
-    Debug, "%s %d: %d",
-    __FUNCTION__, graphNode->input1_id(), graphNode->output_id()
-  );
+  if (@available(iOS 15.4, macOS 12.3, *)) {
+    auto graphNode = nodePtr->mpsnode_union_as_MPSScatter();
+    ET_LOG(
+      Debug, "%s %d: %d",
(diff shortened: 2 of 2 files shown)


f-meloni:
cc @DenisVieriu97

shoumikhin:
@DenisVieriu97 please take a look!

shoumikhin:
@DenisVieriu97 shall we just merge it?

backends/apple/mps/runtime/operations/IndexingOps.mm
```diff
@@ -206,24 +206,27 @@
 
 Error
 MPSGraphBuilder::mpsScatterOp(NodePtr nodePtr) {
-  auto graphNode = nodePtr->mpsnode_union_as_MPSScatter();
-  ET_LOG(
-    Debug, "%s %d: %d",
-    __FUNCTION__, graphNode->input1_id(), graphNode->output_id()
-  );
+  if (@available(iOS 16, macOS 13, *)) {
```
cccclai, line 209: Should we return an error for the else statement

f-meloni: If the if is not true, then we do not do `_idToMPSGraphTensor[graphNode->output_id()] = ...` which I think would then not set the tensor value and that would fail the prediction, but only if this is used.
Do you think we should return `Error::notSupported` or something like that?

cccclai: Yeah, just we can handle it in a non-fatal way.

f-meloni: @cccclai Is the latest version OK?
Thank you for the review :)

backends/apple/mps/runtime/operations/IndexingOps.mm
```diff
@@ -206,25 +206,30 @@
...
+      [_mpsGraph scatterAlongAxis:dim
+                   withDataTensor:inputTensor
+                    updatesTensor:updatesTensor
+                    indicesTensor:indicesTensor
+                             mode:MPSGraphScatterModeSet
+                             name:nil];
+
+    return Error::Ok;
+  } else {
+    return Error::NotSupported;
```
cccclai, line 231: Thanks, one minor thing - can you add an error log here like
```
ET_LOG(Error, "...")
```

cccclai (approved):

backends/apple/mps/runtime/operations/IndexingOps.mm
f-meloni: Added, and I also changed the requirements to iOS 15.4 and MacOS 12.3

backends/apple/mps/runtime/MPSDevice.mm
```diff
@@ -22,11 +22,11 @@ static inline MTLLanguageVersion getMetalLanguageVersion(const id<MTLDevice>& de
   // MPS Advanced Indexing needs at least Metal 2.0 (support for Argument Buffers and function constants)
   // host_name attribute needs at least Metal 2.2 and ulong needs Metal 2.3 (supported on MacOS 11+)
   MTLLanguageVersion languageVersion = MTLLanguageVersion2_3;
-#if defined(__MAC_13_0)
-  if (macOS13Plus) {
-    languageVersion = MTLLanguageVersion3_0;
+  if (@available(iOS 16, macOS 13, *)) {
```
DenisVieriu97, line 25: Thank you for the change

backends/apple/mps/runtime/operations/IndexingOps.mm
```diff
@@ -206,25 +206,32 @@
 
 Error
 MPSGraphBuilder::mpsScatterOp(NodePtr nodePtr) {
-  auto graphNode = nodePtr->mpsnode_union_as_MPSScatter();
-  ET_LOG(
-    Debug, "%s %d: %d",
-    __FUNCTION__, graphNode->input1_id(), graphNode->output_id()
-  );
+  if (@available(iOS 15.4, macOS 12.3, *)) {
```
DenisVieriu97, line 209: good catch

DenisVieriu97 (approved):
Looks good - thanks @f-meloni for the changes

PR #9100, opened 2025-03-10
[Windows] Add macro for weak linkage for Windows platforms

SS-JIA:
### Summary
Currently, when trying to build pybindings on Windows, there is a duplicate symbol error for `et_pal_emit_log_message`. This is because of weak symbols are currently not supported on Windows; the symbol is implemented once in `posix.cpp` and once again in `pybindings.cpp`.

On Windows, a function **implementation** can be marked weak via

```
#pragma weak my_func
void my_func() {
  ...
}
```

However note that this is different from GCC, which specifies the function **definition** as weak via `__attribute__((weak))`.

To support weak linkage on Windows, introduce the following block to function implementations in `posix.cpp`

```
#ifdef _MSC_VER
#pragma weak func_name
#endif // _MSC_VER
```

### Test plan

Build ExecuTorch with pybindings:

```powershell
del -Recurse -Force cmake-out; `
  cmake . `
  -DCMAKE_INSTALL_PREFIX=cmake-out `
  -DPYTHON_EXECUTABLE=C:\\Users\\ssjia\\AppData\\Local\\miniconda3\\python.exe `
  -DCMAKE_PREFIX_PATH=C:\\Users\\ssjia\\AppData\\Local\\miniconda3\\Lib\\site-packages `
  -DCMAKE_BUILD_TYPE=Release `
  -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON `
  -DEXECUTORCH_BUILD_PYBIND=ON `
  -T ClangCL `
  -Bcmake-out; `
  cmake --build cmake-out -j64 --target install
```

### DIFF
--- a/runtime/platform/default/posix.cpp
+++ b/runtime/platform/default/posix.cpp
@@ -75,6 +75,9 @@ static bool initialized = false;
  * This function should be called before any other function provided by the PAL
  * to initialize any global state. Typically overridden by PAL implementer.
  */
+#ifdef _MSC_VER
+#pragma weak et_pal_init
+#endif // _MSC_VER
 void et_pal_init(void) {
   if (initialized) {
     return;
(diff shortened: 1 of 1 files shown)


SS-JIA:
So current macro doesn't work unfortunately, I think GCC doesn't allow macros with arguments in certain areas? We will need to declare the pragma only for windows.

tarun292 (approved):

PR #9105, opened 2025-03-10
[Executorch][SDPA] Fix bug in sdpa

kimishpatel:
* #9350
* __->__ #9105

This diff fixes two bugs
1. When doing flash attention, the partical q @ k block may contain some entries that needs to be masked out. This logic had a bug. Maybe this bug also exist in PT core. I will look into that to add test and see if I can prove it.
2. Due to special handling via start_pos in SDPA it also exposed the bug in 1 when doing really long sequence prefill in chunked manner.

It is probably better to just use mask though.

Code has detail comments on the issue and fix.

Differential Revision: [D70922039](https://our.internmc.facebook.com/intern/diff/D70922039/)


larryliu0820 (approved):
Looks good, thanks for adding the test

extension/llm/custom_ops/op_sdpa.cpp
```diff
@@ -382,7 +382,7 @@ void cpu_flash_attention(
       /* qk_sum */ qSplitSize +
       /* dst    */ qSplitSize * headSize;
 
-  int64_t size_bytes = size_per_thread * num_thread * query.element_size();
+  int64_t size_bytes = size_per_thread * num_thread * query.element_size() * 4;
```
iseeyuan, line 385: Is *4 for fp32?

iseeyuan (approved):

extension/llm/custom_ops/op_sdpa.cpp
kimishpatel: oh really good catch. Thats was left over from local debug. need to remove

mergennachin: please finalize so that @spalatinate has the correct version

kimishpatel: here is the fix pr https://github.com/pytorch/executorch/pull/9492

PR #9107, opened 2025-03-10
Refactor apply_over_dim_list to plan-then-execute

swolchok:
First step toward plan-then-execute for reductions, which will cause preparatory work to be done once rather than once per output element.

### DIFF
--- a/kernels/portable/cpu/util/reduce_util.h
+++ b/kernels/portable/cpu/util/reduce_util.h
@@ -45,7 +45,7 @@ template <typename Fn>
 void apply_on_flat_ix_with_dim_mask_and_base(
     const Fn& fn,
     const Tensor& in,
-    bool* dim_mask,
+    const bool* dim_mask,
     const size_t base,
     const size_t start,
     const size_t end) {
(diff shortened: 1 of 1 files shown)


swolchok:
* (to be filled)

kernels/portable/cpu/util/reduce_util.h
```diff
@@ -295,6 +295,92 @@ void apply_over_dim(
...
+    for (const auto& d : dim_list.value()) {
+      const size_t non_neg_d = d < 0 ? d + in.dim() : d;
+      is_in_dim_list_[non_neg_d] = true;
+    }
+
+    mode_ = ExecutionMode::NormalDimMask;
+  }
+
+  template <typename Fn>
+  void execute(const Fn& fn, const size_t out_ix) const {
```
JacobSzwejbka, line 340: Not really in scope of this diff, but it would be better to return Error up to the root kernel which could then properly set the error state and safely terminate model execution rather then just aborting.

kernels/portable/cpu/util/reduce_util.h
```diff
@@ -295,6 +295,92 @@ void apply_over_dim(
...
+    ET_CHECK_MSG(out_ix < out_numel_, "Out index %zd is out of bounds", out_ix);
+
+    switch (mode_) {
+      case ExecutionMode::NothingToDo:
```
JacobSzwejbka, line 344: Why do we end up calling the plan at all if theres nothing to do since the mode is set once. Is it just a safety thing so that each kernel doesnt have to bail out before calling it?

JacobSzwejbka (approved):

kernels/portable/cpu/util/reduce_util.h
swolchok: 1) encapsulation -- the kernel code doesn't know anything about the details of the plan
2) the call can be inlined, so why not design it that way?

PR #9108, opened 2025-03-10
Refactor: MapReduceOverDimListPlan

swolchok:
Another step toward plan-then-execute for reductions.


swolchok:
* (to be filled)

kernels/portable/cpu/util/reduce_util.h
```diff
@@ -309,7 +309,7 @@ class ApplyOverDimListPlan {
           dim_list,
       const int64_t start = 0,
       const int64_t end = -1)
```
JacobSzwejbka, line 311: does -1 just mean go to end?

JacobSzwejbka (approved):

kernels/portable/cpu/util/reduce_util.h
swolchok: I think so. don't think I changed that behavior

PR #9109, opened 2025-03-10
Refactor: ReduceOverDimListPlan

swolchok:
Now we can plan-then-execute for all reductions that use dim lists.


swolchok:
* (to be filled)

kernels/portable/cpu/util/reduce_util.h
```diff
@@ -606,6 +606,28 @@ std::tuple<CTYPE, long> reduce_over_dim(
...
+  ReduceOverDimListPlan(
+      const executorch::aten::Tensor& in,
+      const executorch::aten::optional<executorch::aten::ArrayRef<int64_t>>&
+          dim_list)
+      : plan_(in, dim_list) {}
+
+  template <typename CTYPE, typename ReduceOp>
+  CTYPE execute(const ReduceOp& reduce_fun, const size_t out_ix) {
+    return plan_.execute<CTYPE, CTYPE>(
+        [](CTYPE v) { return v; }, reduce_fun, out_ix);
```
JacobSzwejbka, line 624: does the identity macro carry any meaningful overhead?

JacobSzwejbka (approved):

kernels/portable/cpu/util/reduce_util.h
swolchok: I would expect inlining to eliminate it; the one thing it could possibly do is implicitly convert the argument to `CTYPE`

PR #9111, opened 2025-03-10
special case dim-list ops for dim list having size 1

swolchok:
Applying over a single dim is much faster; the general apply-over-dim-list logic has to do extra work (tracking the current delinearized index).


swolchok:
* (to be filled)

iseeyuan (approved):
Looks good to me to specialize 1 dim.

PR #9134, opened 2025-03-11
Arm backend: Don't handle const placeholders in match_arg_ranks_pass

martinlsm:
This pass did previously handle constant placeholders directly by modifying the data in the state_dict, whereas call_functions are handled by inserting view ops. Since the introduction of the FuseConsantOpsPass however, ops acting on constant placeholders are computed AOT and fused into one single node.

We can therefore now handle all nodes the same way here by inserting view nodes, removing the need for a special case in this pass.


shoumikhin:
@DigantDesai, @KimishPatel can you please take a look?

zingo (approved):

zingo:
macos fails unrelated

PR #9135, opened 2025-03-11
Do auto redirect for old getting started page

mergennachin:
The new page name (getting-started.html) is strictly better than the old one (getting-started-setup.html)

Let's make a redirect so that we don't force users to make an unnecessary click.

Test Plan: 

https://docs-preview.pytorch.org/pytorch/executorch/9135/getting-started-setup.html

### DIFF
--- a/docs/source/getting-started-setup.rst
+++ b/docs/source/getting-started-setup.rst
@@ -1,0 +1,13 @@
+Setting Up ExecuTorch
+=====================
+
+This page is re-organized into the following pages:
+
+* `Getting Started with ExecuTorch <getting-started.html>`_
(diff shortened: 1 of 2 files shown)


docs/source/getting-started-setup.rst
```diff
@@ -0,0 +1,13 @@
+Setting Up ExecuTorch
```
mergennachin, line 1: Did `git mv` and modified

byjlw (approved):

GregoryComer (approved):
Looks good to me. I do wonder if a server-side redirect is better for SEO, though this isn't my area of expertise.

PR #9151, opened 2025-03-11
[ExecuTorch][Weight Sharing] Track Named Data Store in EdgeProgramManager

mcr229:
* #9155
* #9154
* #9153
* #9152
* __->__ #9151

We enable Backends to return Named Data by adding NamedDataStoreOutput to the preprocess result. This is a completely BC change, as no backends with an implemented preprocess will see any change if nothing is explicitly implemented.

For backend developers to leverage the new NamedDataStore, they can initialize a new NamedDataStore() within preprocess, add_named_data to the data store, and return the NamedDataStore.get_named_data_store_output() in the preprocess result like such:

```
def preprocess(ExportedProgram, List[CompileSpecs]) -> PreprocessResult:
    named_data_store = NamedDataStore()

    for node in exported_program.graph.nodes:
        named_data_store.add_named_data("name", bytes)

    return PreprocessResult(
        processed_bytes=bytes,
        debug_handle_map={},
        data_store_output= named_data_store.get_named_data_store_output()
    )
```

Under the hood, the data store output is embedded in the loweredbackendmodule, (serializing loweredbackendmodule by itself with the a named_data_store_output is still a todo). But via the EdgeProgramManager path, we add the named_data_store_outputs to the edge_program_manger's named data store to keep track of all the named data returned by backends.

Differential Revision: [D70451660](https://our.internmc.facebook.com/intern/diff/D70451660/)


cccclai (approved):

JacobSzwejbka:
What is XNNPack serializing here? Is it just the unpacked tensors and their names? In that case could we just put them in the namedTensorStore instead of the named data store so that its plug and play with training .ptds?

PR #9153, opened 2025-03-11
[ExecuTorch][Weight Sharing][XNNPACK] Serialize constant tensors into named data map

mcr229:
* #9155
* #9154
* __->__ #9153
* #9152
* #9151

We serialize tensors into the named data map, and return the output in preprocess result. Allowing for XNNPACK to share tensors with the same name (instead of duplicating).

A key change here is with fused tensors. For BN and Convolution Fusion, we fuse the conv weights and bias with the BN parameters creating new tensors. We then create get_attr nodes for these new parameters. Due to the graph.fx interpreter in export pass base, the new names we create for these new tensors are lost each time. As a result, at the end we introduce a new pass to preserve the names we created. This seems a little hacky for now, but is the only way to preserve the new fused names.

Differential Revision: [D70315207](https://our.internmc.facebook.com/intern/diff/D70315207/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D70315207/)!


kirklandsign (approved):

SS-JIA:
For context, I messed up the merging order of the bot generated diffs, I accidentally started merging from this PR and then later noticed that the one below it hadn't been merged yet. After I merged the one below this, I thought that the changes from this PR onward didn't make it into `main` so I was trying to figure out a remediation... however it seems that somehow all the changes got rolled up into the merge commit of the second PR. Therefore, even though https://github.com/pytorch/executorch/commit/23fe285c1b830e582e44ea85abdb96270883788a has the message of the second PR it actually contains changes for the top 4 PRs of this stack. So no further action is needed (I think).

cc: @kirklandsign @mcr229

PR #9158, opened 2025-03-11
Add export_llama performance regression test using expected ops

jackzhxng:
### Summary
Add a proxy for an `export_llama` performance regression test by comparing the ops in the graph before and after the PR. The export happens without loading a checkpoint or params file, which means that all of the base `ModelArgs` values for `llama_transformer` will be used.

### Test plan
N/A


examples/models/llama/tests/test_export_llama_lib.py
```diff
@@ -0,0 +1,81 @@
...
+import torch
+
+from executorch.devtools.backend_debug import get_delegation_info
+from executorch.examples.models.llama.export_llama_lib import _export_llama, build_args_parser
+
+
+# Ops expected to be found in the default exported llama_transformer. Obtained through
+# print_delegation_info from the backend_debug module, which is displayed with
+# export_llama under --verbose.
+BASE_EXPECTED_OPS = {
```
tarun292, line 18: I think this might be a little risky to test. If the IR changes which is a common possibility this count will change and we'll have to keep fixing this.

examples/models/llama/tests/test_export_llama_lib.py
```diff
@@ -0,0 +1,81 @@
...
+    "aten_squeeze_copy_dims": 20,
+    "aten_sub_tensor": 10,
+    "aten_unsqueeze_copy_default": 20,
+    "aten_cat_default": 10,
+    "update_cache": 10,
+    "llama_custom_sdpa_default": 5,
+    "aten_sigmoid_default": 5,
+}
+
+UNWANTED_OPS = [
```
tarun292, line 41: Just test for this maybe?

examples/models/llama/tests/test_export_llama_lib.py
```diff
@@ -0,0 +1,50 @@
...
+
+        If this test breaks, please ensure that the difference in ops
+        is intentional before updating the expected ops.
+        """
+        # Since we aren't loading a checkpoint, it doesn't
+        # matter what model we specify. Note that
+        # we cannot test quantization args in this way
+        # since quantization requires promoting meta tensors
+        # to the cpu device, which requires real weights.
+        export_args_str = """
```
tarun292, line 35: Also why do this? Why not just generate the args directly and use them.

tarun292 (approved):

examples/models/llama/tests/test_export_llama_lib.py
jackzhxng: Oh I feel like it's more clear to read?

PR #9176, opened 2025-03-12
cover //kernels/prim_ops/... in unittest-buck

swolchok:
Everything but the Python test (which depends on //caffe2:torch) is fine.


swolchok:
* (to be filled)

mergennachin (approved):

Issue #9019, opened 2025-03-06
Build pip wheel and validate AOT workflow on different backends & host environments in CI


larryliu0820:
### 🚀 The feature, motivation and pitch

We need to guard our AOT workflow in CI, to make sure our pip wheel can give users out-of-the-box user experience.

In CI we should cover the flow of:
* Build pip wheel on PR or trunk, for MacOS or Linux
* Install pip wheel
* Export a simple model (MV3) with common backend delegation (XNNPACK, Coreml, QNN) and retrieve .pte files
* Load the .pte files in python and run sample input (optional, depend on feasibility)

|              | MacOS | Linux |
| :---------------- | :------: | ----: |
| XNNPACK        |   <ul><li>- [ ] </li></ul> | <ul><li>- [ ] </li></ul> |
| CoreML           |   <ul><li>- [ ] </li></ul>   | <ul><li>- [ ] No need to run</li></ul>|
| QNN    |  <ul><li>- [ ] </li></ul>   | <ul><li>- [ ] </li></ul> |

larryliu0820:
@jathu We actually build nightly wheels. I was hoping that we can extend the nightly wheel build job to include the test we want to add. [Here's the nightly flow: ](https://github.com/pytorch/executorch/blob/main/.github/workflows/nightly.yml)

jathu:
@larryliu0820 it would be great if we can do it in every PR, but nightly is good enough too. So we currently run a [simple model test](https://github.com/pytorch/executorch/blob/9dcee22c1fe8c1b69ff6b42aa967ad2eafd29653/build/packaging/smoke_test.py) in the wheel workflow. We will be updating the workflow to build MV3, and also test the respective outputs on iOS and Android?

PR #9192, opened 2025-03-12
Test MV3 model in CI wheel builds

jathu:
### Summary

As part of https://github.com/pytorch/executorch/issues/9019, we want to ensure we can export models on P0 hosts targeting P0 backends. So, let's test this in our wheel creation. For now, we introduce macOS/Linux XNNPACK export.

Given the wheel building GitHub workflow doesn't support using jobs as a test step, we must go use a python script. This runs a similar test to what we do in PRs:

https://github.com/pytorch/executorch/blob/1011fdc98bcd61063b845670fb1779635aa30e3a/.github/workflows/pull.yml#L85-L136

### Test plan

CI → Build {Linux, macOS} Wheels

cc @larryliu0820 @lucylq


larryliu0820 (approved):
Thanks! This is way better than before and thanks for adding the test after building the wheel.

.ci/scripts/wheel/test_base.py
```diff
@@ -0,0 +1,66 @@
...
+    #   - We stop running the Docker image as root: https://github.com/pytorch/test-infra/issues/5091
+    envvars = os.environ.copy()
+    envvars.pop("HOME")
+
+    for model_test in model_tests:
+        subprocess.run(
+            [
+                os.path.join(_repository_root_dir(), ".ci/scripts/test_model.sh"),
+                str(model_test.model),
+                # What to build `executor_runner` with for testing.
```
mergennachin, line 59: why do we need so many redirections? 

Can we just call this command directly in the .yml file?

mergennachin (commented):
Instead of having this

```
smoke-test-script: ${{ matrix.smoke-test-script }}
```

can we just run the script directly, similar to how we do in pull.yml and trunk.yml

jathu:
> Instead of having this
> 
> ```
> smoke-test-script: ${{ matrix.smoke-test-script }}
> ```
> 
> can we just run the script directly, similar to how we do in pull.yml and trunk.yml

I don't think this is possible with the current job template, it only support running python scripts:

https://github.com/pytorch/test-infra/blob/dd355a4aa131ba56a81bee1ff408ff14b1bcbf70/.github/workflows/build_wheels_macos.yml#L220-L251

mergennachin (approved):

PR #9193, opened 2025-03-12
Attempt to unbreak build_apple_frameworks

swolchok:
Not sure why the threadpool extension isn't mentioned or built in here; we should follow up on that. Also attempting to speed up the build from the hour it currently takes.


swolchok:
* __->__ #9193

build/build_apple_frameworks.sh
```diff
@@ -189,6 +188,7 @@ cmake_build() {
         ${platform_target:+-DDEPLOYMENT_TARGET=$platform_target} \
         --log-level=VERBOSE
     cmake --build . \
+        -j 6 \
```
shoumikhin, line 191: We're building with xcodebuild here (see `-G Xcode`), so it'll scale the build for us and this options wouldn't have any effect

shoumikhin (approved):

build/build_apple_frameworks.sh
swolchok: OK, I'll revert this as long as the CI job still takes an hour to finish

swolchok: ah, I checked the logs closer and the PyTorch build is taking 40 minutes so it's unlikely that this is going to help

swolchok:
build-frameworks-ios is green. removing the -j stuff and landing

PR #9195, opened 2025-03-12
Update XNNPACK and backend template docs

GregoryComer:
### Summary
Update XNNPACK backend doc page to cover quantization schemes and generally clean up the format. Update backend template doc with additional detail.

### Test plan
Built docs locally and verified contents.

cc @mergennachin @byjlw


byjlw (approved):
I didn't actually test the commands but the copy looks good.

PR #9196, opened 2025-03-12
Add _unsafe_reset_threadpool to pybindings

GregoryComer:
Summary: Expose thread count setting in pybindings. Currently, pybindings can be very slow on some server machines (observed 300x slower - went from 12s for 100 iterations to under 40ms for 100 iterations). As a stopgap measure, this diff exposes _unsafe_reset_threadpool from python, which can be used to reduce the thread count. This should be done prior to loading a model.

Differential Revision: D71023514


mcr229 (approved):
looks fine to me

PR #9199, opened 2025-03-12
Create Arm64 Graviton instance docker image

mergennachin:
Summary:

- Let's see if we can set this up first.

In follow up PRs, here's the plan:

- We will add a few CI jobs to run on this docker image and on Arm64 Graviton instance.
- Will migrate some jobs from macOS runners to this.


mergennachin:
Will monitor this: 

https://github.com/pytorch/executorch/actions/runs/13820336018/job/38663878658

kirklandsign (approved):

PR #9200, opened 2025-03-12
Update macros in scalar_type_util.h to handle unsigned int types

shoumikhin:


swolchok (approved):
looks good. don't forget you'll need to unbreak CI failures; they're related

shoumikhin:
This became more complex than expected, since it'd affect type promotion logic, thus taking another route in #9219

PR #9201, opened 2025-03-12
fix ci

metascroy:
CI is assigning Python 3.9 machines, but ET requires 3.10.

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -234,7 +234,7 @@ jobs:
     uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
     with:
       runner: macos-m1-stable
-      python-version: '3.11'
+      python-version: "3.11"
       submodules: 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/trunk.yml
```diff
@@ -234,7 +234,7 @@ jobs:
     uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
     with:
       runner: macos-m1-stable
-      python-version: '3.11'
+      python-version: "3.11"
```
manuelcandales, line 237: if this is a problem, then why not change in the other places as well? like line 218

manuelcandales (approved):

.github/workflows/trunk.yml
metascroy: Well I don't know if this is the problem.  I might have to follow-up with another PR using a more specific version like you mentioned in chat thread.

The other tests aren't as flaky, but they might overcome the issue with CONDA_RUN.

PR #9207, opened 2025-03-12
Add tests on graviton instances

mergennachin:
- **Create Arm64 Graviton instance docker image** https://github.com/pytorch/executorch/pull/9199/
- **Test models on linux aarch64 graviton instances** 

In the next follow-up PR, I will trim macos jobs


swolchok (approved):
I guess for this to make sense you have to think that aarch64 coverage is enough of a substitute for macos coverage. it's a reversible decision if we turn out to have mac-only problems, so sure why not.

swolchok:
(also your tests are timing out, probably want to fix that)

PR #9210, opened 2025-03-12
[Android benchmarking] Cool down the device before a run

kirklandsign:
### Summary
This could reduce the variance.
Example: 25.44 → 24.74 which is good variance.


huydhn (approved):
TIL about how to cool down the device :)

shoumikhin (approved):

kimishpatel:
Is mean closer to median now? We do this in aibench so good thing to force. Although 10s might too small?

PR #9218, opened 2025-03-13
Include libcpublas.a into optimized kernels framework

shoumikhin:


huydhn (approved):

huydhn:
Hmm, the job is still failing I think

shoumikhin:
@huydhn exactly, that's why I'm wondering if that approach works.
It builds and runs flawlessly for my locally when I git clone recursively.

huydhn:
I look at the logs and see that submodules parameter was set to true https://github.com/pytorch/executorch/actions/runs/13827845229/job/38685932365?pr=9218#step:5:7, so I think it will work in main where the parameter has been set to recursive.  I thought the PR should do the same, but I think I'm missing something here.

huydhn:
Anyway, the workflow used in this PR https://github.com/pytorch/executorch/actions/runs/13827845229/workflow#L282 still had submodules set to true, that's why the build job fails.  The fix for the PR would be to do a rebase to latest main branch.

PR #9220, opened 2025-03-13
Arm backend: Add XfailIfNoCorstone300/320.

AdrianLundell:
This decorator will replace SkipIfNoCorstone300/320. The reasoning for this is that the majority of the test still can run and we can avoid double testing with/without FVP. The test will fail as usual for any other than FileNotFoundError, meaning that other errors will be caught.

Proof of concept added in test_add, will not do this for all test cases here for avoiding merge conflicts to get a quicker merge.

Additionally renames add2 -> add_2 to parse op from test name more easily.


AdrianLundell:
Failures not related

zingo (approved):

PR #9225, opened 2025-03-13
Make oss Coreml Llama supports both list and tensor KV cache inputs

derekxu:
Summary:
Make oss Coreml Llama supports both list and tensor KV cache inputs

- reorder input args sequence to match with other code
- Make optional static_seq_len for the input tokens/static_seq_len

Differential Revision: D71081340


metascroy (approved):
Waiting for OSS CI to complete.  I just added the "trunk" label, which is required for the macOS jobs to run.

PR #9227, opened 2025-03-13
Refactor gather_test_model to be hard coded directly

mergennachin:
The CI runner logic should directly be in the corresponding yml files.

Gather_test_models.py contains overly complicated python logic on what models, backends, runners to do, with many switch statements.
I think we should just hard code directly in job descriptions to simplify things and know exactly what jobs are running directly in the yml file.

This is a no-op PR

### DIFF
--- a/.ci/scripts/gather_test_models.py
+++ b/.ci/scripts/gather_test_models.py
@@ -5,6 +5,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# WARNING: The CI runner logic should directly be in the corresponding yml files
+# This file will be deleted once the reference in periodic.yml is deleted.
+
 import itertools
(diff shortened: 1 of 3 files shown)


malfet (approved):

jackzhxng (commented):
Are we able to specify specific runner per model using the matrix? We might need something like this to move the linux trunk test off gather_test_models.py e.g.
```
include:
          - model: efficient_sam
            backend: portable
            runner: linux.4x.large
```

mergennachin:
@jackzhxng 

Yeah, you can use include and exclude logic directly in the matrix

The example you provided works.

mergennachin:
@jackzhxng 

One thing that I realized I had a bug. The include statements should define all the variables. 

https://github.com/pytorch/executorch/pull/9235

PR #9234, opened 2025-03-13
Pull submodules recursively

shoumikhin:


huydhn:
Do you need to update https://github.com/pytorch/executorch/blob/main/.github/workflows/apple.yml#L52 and https://github.com/pytorch/executorch/blob/main/.github/workflows/apple.yml#L139 too?

huydhn (approved):
Stamped, but please check the other build jobs `build-demo-ios` and `build-frameworks-ios` to see if they need submodules set to recursive too before landing (I don't see they fail in trunk)

shoumikhin:
@huydhn thanks! I've addressed those too, just in case.

PR #9236, opened 2025-03-13
Run Apple specific tests on CI

shoumikhin:


.ci/scripts/unittest-macos.sh
```diff
@@ -35,6 +35,7 @@ if [[ "$BUILD_TOOL" == "cmake" ]]; then
     .ci/scripts/unittest-macos-cmake.sh
 elif [[ "$BUILD_TOOL" == "buck2" ]]; then
     .ci/scripts/unittest-buck2.sh
+    .ci/scripts/unittest-macos-buck2.sh
```
huydhn, line 38: Any reason why it couldn't go into https://github.com/pytorch/executorch/blob/main/.ci/scripts/unittest-buck2.sh.  You can use `uname` to run `buck2 test //extension/apple:ExecuTorch` only on MacOS if needed

shoumikhin: Do you prefer that approach? These tests are always gonna be Apple-specific, I thought I'd be nice to not pollute the cross-platform scripts with them, wdyt?

huydhn: I don't feel too strongly about either conventions here.  As there are already 2 scripts for macOS:

* https://github.com/pytorch/executorch/blob/main/.ci/scripts/unittest-macos.sh
* and https://github.com/pytorch/executorch/blob/main/.ci/scripts/unittest-macos-cmake.sh

So having another `.ci/scripts/unittest-macos-buck2.sh` follows that convention.  On the other hand, https://github.com/pytorch/executorch/blob/main/.ci/scripts/unittest-buck2.sh seems to be a cross platform script that is on both Linux and MacOS.

huydhn (approved):

PR #9238, opened 2025-03-13
[Benchmark][BE]Deprecate v2

yangw-dev:
Issue: https://github.com/pytorch/test-infra/issues/6294
Remove benchmark v2  schema logics, still keep the way to store v3 with v3 folder, since we might have higher version of schema in the future

next step is introduce the failure handling for benchmark record


huydhn (approved):
LGTM!

PR #9243, opened 2025-03-13
Use parallel_for in elementwise_util

swolchok:
More straightforward rollout.

(Parallelizing over BroadcastIndexesRange is ugly, but so far we have exactly two instances of it. If it keeps cropping up, we can add a utility to make it nicer.)


swolchok:
hmm, not good that this is breaking tests.

manuelcandales (approved):

swolchok:
this was for #8932

PR #9245, opened 2025-03-13
Add (and fix) OSS check that et_operator_library dtype_selective_build builds

swolchok:
This wasn't checked, and it's easy to forget to update the all_deps targets if we don't check. Now we check.


kernels/portable/cpu/pattern/targets.bzl
```diff
@@ -11,7 +11,7 @@ def define_common_targets():
     # build, where the portable ops are built from source and linked with :all_deps
     runtime.cxx_library(
         name = "all_deps",
-        deps = [
+        exported_deps = [
```
swolchok, line 14: it's odd that this wasn't required internally

manuelcandales (approved):

swolchok:
test-apple-model failure looks non-flaky but unrelated; HUD confirms it's broken on trunk
test-llama-runner-mac failure is output mismatch; HUD confirms it's flaking on trunk

swolchok:
test-arm-backend doesn't use buck, should be a flake as well. rerunning failed jobs and importing to run internal CI just in case

larryliu0820 (approved):

lucylq:
thanks for adding this check :0

PR #9247, opened 2025-03-13
[Benchmark] Generate benchmark record for job failure

yangw-dev:
# Description
compose failure benchmark record

Issue: https://github.com/pytorch/test-infra/issues/6294

# Related Query File 
https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql

# Details
when a job fails in git_job level, or a device fails in the benchmark test,
we return a benchmark record to indicate the failure, so that we can properly render the HUD UI ti distinguish metris not run and metris run with failures

This pr may introduce `Unknown`  to fields  in the HUD execubench table temporarily,  will fix it in HUD UI to handle special benchmark value.

for both level of failure, the metric name will be "FAILURE_REPORT".
In HUD, We mainly use the special metric name to identify failure, if more information needed, we get which level the job fails in benchmark.extra_info.

## Step Failure
When a failure detected:
we try to extract model info from git_job_name, the step will fail if the model info cannot be extracted.

# Example of benchmark record for failure benchmark:
## When a job failed at device-job-level, 
- device_name:  get name from job_report.name . for instance `iPhone 15`
- device_os: job_report.os with prefix  "Android" or "iOS". this should match both android and ios setting
- model.name: extract from git job name
- model.backend: extract from git job name
- metric.name: "FAILURE_REPORT"

```
{
  "benchmark": {
    "name": "ExecuTorch",
    "mode": "inference",
    "extra_info": {
      "app_type": "IOS_APP",
      "job_conclusion": **"FAILED"**,
      "failure_type": **"DEVICE_JOB"**,
      "job_report": "..."
  },
  "model": {
    "name": **"ic4",**
    "type": "OSS model",
    "backend": **"mps"**
  },
  "metric": {
    "name": **"FAILURE_REPORT"**,
    "benchmark_values": 0,
    "target_value": 0,
    "extra_info": {
      "method": ""
    }
  },
  "runners": [
    {
      "name": **"iPhone 15"**,
      "type": **"iOS 18.0"**,
    }
  ]
}
```

## when a job failed at git-job-level (there is no job_reports)
this happens when a job fails before it runs the benchmark job

- device_name:  device_pool_name from git job bane #exmaple:  sumsung_galaxy_22
- device_os: "Android" or "iOS"
- model.name: extract from git job name
- model.backend: extract from git job name
- metric.name: "FAILURE_REPORT"

the failure benchmark record looks like:
```
{
  "benchmark": {
    "name": "ExecuTorch",
    "mode": "inference",
    "extra_info": {
      "app_type": "IOS_APP",
      "job_conclusion": **"FAILURE"**,
      "failure_type": **"GIT_JOB"**,
      "job_report": "{}"
    }
  },
  "model": {
    "name": "ic4",
    "type": "OSS model",
    "backend": "mps"
  },
  "metric": {
    "name": "FAILURE_REPORT",
    ...
  },
  "runners": [
    {
      "name": "samsung_galaxy_s22",
      "type": "Android",
     ...
    }
  ]
}
```


.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,271 @@ def transform(
...
-            if not app_type or app_type not in ["ANDROID_APP", "IOS_APP"]:
-                info(
-                    f"App type {app_type} is not recognized in artifact {json.dumps(artifact)}"
-                )
-                continue
 
-            job_name = artifact["job_name"]
-            artifact_type = artifact["type"]
-            artifact_s3_url = artifact["s3_url"]
+def transform_failure_record(
```
yangw-dev, line 422: @huydhn  main code to generate the failure record

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+    Transform the benchmark results into the format writable into the benchmark database for job failures
+    """
+    # From https://github.com/pytorch/pytorch/wiki/How-to-integrate-with-PyTorch-OSS-benchmark-database
+    return {
+        "benchmark": {
+            "name": "ExecuTorch",
+            "mode": "inference",
+            "extra_info": {
+                "app_type": app_type,
+                "job_conclusion": result,
```
yangw-dev, line 443: I leave the job_conclusion and failure_level in benchmark, 
wonder if leave it in metric would be better or not, since we put FAILURE_REPORT there

@huydhn

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+    )
+
+
+def process_benchmark_results(content: Any, app: str, benchmark_configs: str):
+    """
+    main code to run to extract benchmark results from artifacts.
+    Job can be failed at two levels: GIT_JOB and DEVICE_JOB. If any job fails, generate failure benchmark record.
+    """
+    artifacts = content.get("artifacts")
+    git_job_name = content["git_job_name"]
```
huydhn, line 624: A word of caution when trying to extract the information about the run from the job name.  The job name comes from this line https://github.com/pytorch/executorch/blob/main/.ci/scripts/gather_benchmark_configs.py#L335.  So, I think:

* Make the error when failing to parse the job name in `extract_model_info` clearer by referring to `gather_benchmark_configs` script.  Mostly likely, it has been updated without updating `extract_benchmark_results`
* Add a comment on both script that they need to be in sync

.github/scripts/extract_benchmark_results.py
huydhn: Keeping them in benchmark makes more sense to me because the benchmark itself is failing, maybe in the future there might be failure at the metric level, then we can keep information about the failed metric in `metric.extra_info`

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+    """
+    # From https://github.com/pytorch/pytorch/wiki/How-to-integrate-with-PyTorch-OSS-benchmark-database
+    return {
+        "benchmark": {
+            "name": "ExecuTorch",
+            "mode": "inference",
+            "extra_info": {
+                "app_type": app_type,
+                "job_conclusion": result,
+                "failure_level": level,
```
huydhn, line 444: Nit: `failure_type` sounds easier to understand to me

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+        "model": {
+            "name": model_name,
+            "type": "OSS model",
+            "backend": model_backend,
+        },
+        "metric": {
+            "name": "FAILURE_REPORT",
+            "benchmark_values": 0,
+            "target_value": 0,
+            "extra_info": {
```
huydhn, line 457: Nit: if you don't need it, you could just leave out `extra_info`, `avail_mem_in_gb`, and `total_mem_in_gb`.  They are all optional fields

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+def getDeviceOsType(type: str):
+    match type:
+        case "ios":
+            return "iOS"
+        case "android":
+            return "Android"
+    return "UNKNOWN"
+
+
+def generateGitJobLevelFailureRecord(git_job_name: str, app: str) -> Any:
```
huydhn, line 557: Nit: this script use `_` instead of camel case, so maybe changing these functions to follow the same convention? i.e. `generate_git_job_level_failure_record`

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+    """
+    generates benchmark record for GIT_JOB level failure, this is mainly used as placeholder in UI to indicate job failures.
+    """
+    level = "GIT_JOB"
+    app_type = getAppType(app)
+    device_prefix = getDeviceOsType(app)
+
+    model_infos = extract_model_info(git_job_name)
+    model_name = "UNKNOWN"
```
huydhn, line 566: I think it's cleaner to fail the step when `extract_model_info` fails instead of using an `UNKNOWN` here:

* If someone changes the job name and causes `extract_model_info` to fail, they will see the failure on CI.  If we are using `UNKNOWN` as a fallback here, they will need to check the dashboard instead, which is easy to miss
* It will also be confusing if `UNKNOWN` starts to show up on the dashboard

.github/scripts/extract_benchmark_results.py
```diff
@@ -385,54 +389,293 @@ def transform(
...
+
+def generateDeviceLevelFailureRecord(
+    git_job_name: str, job_report: Any, app: str
+) -> Any:
+    """
+    generates benchmark record for DEVICE_JOB level failure, this is mainly used as placeholder in UI to indicate job failures.
+    """
+    level = "DEVICE_JOB"
+    model_infos = extract_model_info(git_job_name)
+    model_name = "UNKNOWN"
```
huydhn, line 593: Same comment, I think it's cleaner to fail the step when `extract_model_info` fails instead of using an `UNKNOWN` here.

.github/scripts/extract_benchmark_results.py
yangw-dev: sounds good

.github/scripts/extract_benchmark_results.py
yangw-dev: my bad. will change this

.github/scripts/extract_benchmark_results.py
yangw-dev: sounds good!

.github/scripts/extract_benchmark_results.py
yangw-dev: sounds good!

yangw-dev: I also raise exception for get_app_type and get_device_os_type

yangw-dev: maybe we can add a unit test to check if this works as expected. i also added comment in perf.yml for job step change

 if this script uses out of those yml files, we can make the regex prefix more flexible to let user pass the step name for chekcing

huydhn:
While checking the results from your branch on the [dashboard](https://hud.pytorch.org/benchmark/llms?startTime=Sat%2C%2008%20Mar%202025%2003%3A19%3A35%20GMT&stopTime=Sat%2C%2015%20Mar%202025%2002%3A19%3A35%20GMT&granularity=day&lBranch=addFake&lCommit=69c42fae7b0d9d58eaff9b9763246903ad775379&rBranch=addFake&rCommit=69c42fae7b0d9d58eaff9b9763246903ad775379&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=All%20Models&backendName=All%20Backends&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms), I notice a curious issue where older commits from your branch have a newer timestamps:

Screenshot 2025-03-14 at 19 33 42

It seems like an issue for later.

huydhn (approved):
LGTM!  Let's take an action item to notice the next failure from these workflows after this lands to double check the results on the dashboard before resolving the issue

yangw-dev:
> While checking the results from your branch on the [dashboard](https://hud.pytorch.org/benchmark/llms?startTime=Sat%2C%2008%20Mar%202025%2003%3A19%3A35%20GMT&stopTime=Sat%2C%2015%20Mar%202025%2002%3A19%3A35%20GMT&granularity=day&lBranch=addFake&lCommit=69c42fae7b0d9d58eaff9b9763246903ad775379&rBranch=addFake&rCommit=69c42fae7b0d9d58eaff9b9763246903ad775379&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=All%20Models&backendName=All%20Backends&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms), I notice a curious issue where older commits from your branch have a newer timestamps:
> 
> Screenshot 2025-03-14 at 19 33 42
> 
> It seems like an issue for later.

interesting, created a issue here :https://github.com/pytorch/test-infra/issues/6427

yangw-dev:
> LGTM! Let's take an action item to notice the next failure from these workflows after this lands to double check the results on the dashboard before resolving the issue

sounds good, I also need to update UI

PR #9248, opened 2025-03-13
Include custom_ops and XNNPACK in executor_runner if built

swolchok:
Somebody (@larryliu0820?) pointed out that there's no particular
reason to require a separate xnn_executor_runner target, and indeed we
already put several other optimized things in here if they're built.


swolchok:
* (to be filled)

swolchok:
hm, apparently the linux runners have an old CMake, but I thought install_requirements.sh installed latest cmake via pip...

swolchok:
I have a local patch for llama-runner-qnn-linux trunk job. to avoid excessive CI usage, I will wait for the other trunk jobs to come back and if they are green, I will just add the patch and merge right away.

swolchok:
>  merge right away.

whoops, thought this was reviewed but it's not. should be good to go

kirklandsign (approved):

swolchok:
hmm, 3 test-models-macos jobs timed out, but they're all succeeding quickly on master. rerunning.

swolchok:
noting that CI is green. rebasing and merging

backends/xnnpack/CMakeLists.txt
```diff
@@ -118,7 +118,7 @@ include(cmake/Dependencies.cmake)
 list(TRANSFORM _xnnpack_backend__srcs PREPEND "${EXECUTORCH_ROOT}/")
 add_library(xnnpack_backend STATIC ${_xnnpack_backend__srcs})
 target_link_libraries(
-  xnnpack_backend PUBLIC ${xnnpack_third_party} executorch_core xnnpack_schema
+  xnnpack_backend PUBLIC ${xnnpack_third_party} executorch_core xnnpack_schema extension_threadpool
```
shoumikhin, line 121: @swolchok just checking, does it have any implications on [build_apple_frameworks.sh](https://github.com/pytorch/executorch/blob/main/build/build_apple_frameworks.sh#L56)'s `backend_xnnpack` framework? Specifically, the fact `kernels_custom` already links against `extension_threadpool`. Can we run into duplicated symbols when both are linked with an app? All iOS CI tests that could have flagged were red by the time this change merged due to some previous changes.

swolchok: I don't think we should be requiring everyone who makes a change to the CMake dependency graph to think about build_apple_frameworks.sh. We should look into using CMake's built-in support for building Apple frameworks instead if possible.

swolchok: > Can we run into duplicated symbols when both are linked with an app? 

I don't know, but I would expect to see red CI jobs in that case.

shoumikhin: Each framework is sorta a higher level abstraction and consists of several CMake artifacts built according to the rules in `build_apple_frameworks.sh`. Apps can link them selectively, which may reveal missing or duplicate symbols. To catch such issues, we rely on demo and benchmark apps running tests as part of CI. However, both are currently failing due to unrelated prior changes.

swolchok: if I'm not mistaken, the build_apple_frameworks job for this diff successfully compiled and linked the demo app and it's failing in tests.

shoumikhin: That’s correct - it means the frameworks were linked as instructed. Now, the app tests will determine whether the frameworks contain all the required symbols, are missing any, or have duplicates. These tests don’t run on every PR but only on specific path changes and in CI (last I heard, Mac runners were costly to run every time). Since the CI is already failing for them, I’m proactively checking if there are any other potentially relevant changes we should investigate once those issues are resolved.

PR #9249, opened 2025-03-13
Split android instrumentation from build script

kirklandsign:
### Summary
We have https://github.com/pytorch/executorch/blob/27bacff5371aa36833069a2f230568c8442d2b73/extension/android_test/setup.sh#L1 which is E2E for building the test. We don't need it when we build the AAR. This reduces the time for user to build AAR.

For benchmark, before, it took 30 minutes to build; now it takes 20 minutes.

### Test plan
CI


shoumikhin (approved):

cccclai (approved):
Looks like big improvement on the build time

PR #9254, opened 2025-03-14
Move MacOS jobs (phi-4-mini, qwen2_5) etc to Arm64

mergennachin:


jackzhxng (approved):
If no ooming on ciflow/trunk then 👍🏻

PR #9264, opened 2025-03-14
Use non-fatal input checks for view operator

### LINKED ISSUE #9130: View op asserts should use non-fatal input checks
### 🐛 Describe the bug

The current implementation of the view operator ([et_view.cpp](https://github.com/pytorch/executorch/blob/main/kernels/prim_ops/et_view.cpp#L79)) uses ET_CHECK macros to check input args. This has the effect of unrecoverably crashing the process when input checks fail. We have observed this on an internal app, where a check failure causes a full app crash.

The kernel should use non-fatal ET_KERNEL_CHECK assertions (logging and returning an error code, rather than calling abort), similar to other portable ops. This allows the caller to gracefully handle model execution failures. See https://github.com/pytorch/executorch/pull/2115/files for an example of how other kernels were updated to use ET_KERNEL_CHECK instead of ET_CHECK macros.

### Versions

N/A

zhenyan-zhang-meta:
### Summary
Fixes #9130

Use non-fatal input checks for view operator, just like https://github.com/pytorch/executorch/pull/2115/files, since there are existing cases that an internal app crashed due to a check failure.

Test is updated as well for `RegisterPrimOpsTest`.

### Test plan
Pass updated test:
```
mkdir cmake-out && cd cmake-out && cmake -DEXECUTORCH_BUILD_TESTS=ON .. && cmake --build . --target kernels_prim_ops_test -j9 && ctest -R kernels_prim_ops_test
```

Pass existing CI/CD.

cc @GregoryComer

### DIFF
--- a/kernels/prim_ops/et_view.cpp
+++ b/kernels/prim_ops/et_view.cpp
@@ -102,7 +102,7 @@ void et_view(KernelRuntimeContext& context, EValue** stack) {
           out,
           /*buffer=*/self.mutable_data_ptr(),
           /*buffer_size=*/out.nbytes()) == Error::Ok,
-      InvalidArgument,
+      Internal,
       ,
       "Failed to set data_ptr for out to self.");
 }
--- a/kernels/prim_ops/et_view.cpp
+++ b/kernels/prim_ops/et_view.cpp
@@ -88,7 +88,7 @@ void et_view(KernelRuntimeContext& context, EValue** stack) {
       resize_tensor(
           out, {expected_output_size, static_cast<size_t>(out.dim())}) ==
           Error::Ok,
-      InvalidArgument,
+      Internal,
       ,
       "Failed to resize output tensor.");
 
--- a/kernels/prim_ops/et_view.cpp
+++ b/kernels/prim_ops/et_view.cpp
@@ -72,28 +72,38 @@ void et_view(KernelRuntimeContext& context, EValue** stack) {
   auto size = (*stack[1]).toIntList();
   auto out = (*stack[2]).toTensor();
 
-  ET_CHECK(tensors_have_same_dtype(self, out));
+  ET_KERNEL_CHECK(
+      context, tensors_have_same_dtype(self, out), InvalidArgument, );
 
   // Compute output size
   SizesType expected_output_size[kTensorDimensionLimit];
-  ET_CHECK(get_view_target_size(self, size, out.dim(), expected_output_size));
+  ET_KERNEL_CHECK(
+      context,
+      get_view_target_size(self, size, out.dim(), expected_output_size),
+      InvalidArgument, );
 
   // Resize for dynamic shape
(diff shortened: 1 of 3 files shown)


kernels/prim_ops/et_view.cpp
```diff
@@ -72,28 +72,38 @@ void et_view(KernelRuntimeContext& context, EValue** stack) {
   auto size = (*stack[1]).toIntList();
   auto out = (*stack[2]).toTensor();
 
-  ET_CHECK(tensors_have_same_dtype(self, out));
+  ET_KERNEL_CHECK(
+      context, tensors_have_same_dtype(self, out), InvalidArgument, );
```
Gasoonjia, line 76: plz remove the extra `,`; same as below

kernels/prim_ops/et_view.cpp
```diff
@@ -72,28 +72,38 @@ void et_view(KernelRuntimeContext& context, EValue** stack) {
...
 
   // Update data ptr
-  ET_CHECK_MSG(
+  ET_KERNEL_CHECK_MSG(
+      context,
       internal::set_tensor_data(
           out,
           /*buffer=*/self.mutable_data_ptr(),
           /*buffer_size=*/out.nbytes()) == Error::Ok,
+      InvalidArgument,
```
Gasoonjia, line 105: @manuelcandales know better, but I'm thinking we should use `Internal` here?

kernels/prim_ops/et_view.cpp
zhenyan-zhang-meta: The commas has to be there since it needs an extra, though empty, param passed in for `retval` (return value).

Similar commas can be seen in main branch: https://github.com/pytorch/executorch/blob/bb3b623fe00c1d962f8012df5097b290fa9a108e/kernels/optimized/cpu/op_add.cpp#L47-L50

Gasoonjia: oh yeah you're right i forgot that. Thanks for noticing!

kernels/prim_ops/et_view.cpp
zhenyan-zhang-meta: Updated as suggested.

swolchok (approved):
shipit (when tests are green)

mergennachin:
Hi, @zhenyan-zhang-meta 
Can you add a corresponding test to exercise the code?

zhenyan-zhang-meta:
> Hi, @zhenyan-zhang-meta Can you add a corresponding test to exercise the code?

Sure, actually I am exploring how to run the relevant tests locally and some of them needs to be updated based on the failed signals.

zhenyan-zhang-meta:
I think the current 3 failing checks left are not relevant to the changes I made.

GregoryComer (approved):
Looks good to me. Thanks for taking this!

PR #9265, opened 2025-03-14
Add null runtime check for XNNExecutor::prepare_args

### LINKED ISSUE #9131: XNNExecutor::prepare_args should check for null runtime
### 🐛 Describe the bug

In the event that an XNNPACK-delegated model is freed right when inference happens, XNNExecutor::prepare_args can crash accessing the runtime. This is technically the user's fault, but we can prevent a crash by checking for a null runtime and returning an error code. We've hit this case with one internal app during shutdown scenarios.

We make this check already in forward https://github.com/pytorch/executorch/blob/ddf0d9eeebf79563d14a65c29dcbe17b7db6c93d/backends/xnnpack/runtime/XNNExecutor.cpp#L137-L140. We should add a similar check at the top of prepare_args here: https://github.com/pytorch/executorch/blob/ddf0d9eeebf79563d14a65c29dcbe17b7db6c93d/backends/xnnpack/runtime/XNNExecutor.cpp#L70. This should be a tiny change - just checking for null runtime and returning an error.

### Versions

N/A

cc @digantdesai @mcr229 @cbilgin

zhenyan-zhang-meta:
### Summary
Fixes https://github.com/pytorch/executorch/issues/9131

In the event that an XNNPACK-delegated model is freed right when inference happens, `XNNExecutor::prepare_args` can crash accessing the runtime. Hence we need to check for null runtime for it, similar to what we are already doing for `XNNExecutor::forward`.

### Test plan
Pass all CI/CD.

### DIFF
--- a/backends/xnnpack/runtime/XNNExecutor.cpp
+++ b/backends/xnnpack/runtime/XNNExecutor.cpp
@@ -68,6 +68,11 @@ ET_NODISCARD Error XNNExecutor::initialize(
  * delegate->execute()
  */
 ET_NODISCARD Error XNNExecutor::prepare_args(EValue** args) {
+  ET_CHECK_OR_RETURN_ERROR(
+      runtime_ != nullptr,
+      Internal,
+      "XNNPACK Delegate did not compile correctly");
+
   // Create xnn_externals_value from evalue args
   xnn_status status;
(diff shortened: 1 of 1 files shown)


mergennachin:
Can you write a test to exercise this code path?

GregoryComer (approved):

PR #9268, opened 2025-03-14
Arm backend: use dbg_fail when node visitors raise exceptions

oscarandersson8218:
Adds a try-expect around the node_visitor call to be able to call dbg_fail() when an error/exception is raised.


zingo (approved):

larryliu0820:
This broke our internal tests and we need to revert it first. Could you please re-submit the PR, and wait for us to import and run internal CI, and paste the error message?

oscarandersson8218:
@larryliu0820 I've re-submitted the PR in https://github.com/pytorch/executorch/pull/9391

PR #9284, opened 2025-03-14
Add et_dump related API to module

cccclai:
Summary: It is for reusing the exention/module in the module definition in pybindings https://github.com/pytorch/executorch/blob/1a918c779e16c0ee903a08b30c1c666d1efb2c57/extension/pybindings/pybindings.cpp#L172

Differential Revision: D71135352

### DIFF
--- a/extension/module/module.h
+++ b/extension/module/module.h
@@ -478,6 +478,11 @@ class Module {
     return event_tracer_.get();
   }
 
+  ET_NODISCARD
+  runtime::Span<uint8_t> debug_buffer() {
+    return runtime::Span<uint8_t>(debug_buffer_.data(), debug_buffer_.size());
+  }
+
  private:
   struct MethodHolder {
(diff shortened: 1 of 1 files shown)


extension/module/module.h
```diff
@@ -489,6 +499,8 @@ class Module {
   std::unique_ptr<runtime::EventTracer> event_tracer_;
   std::unique_ptr<runtime::DataLoader> data_map_loader_;
   std::unique_ptr<NamedDataMap> data_map_;
+  std::unique_ptr<uint8_t[]> debug_buffer_;
```
shoumikhin: Is this essentially a std::vector?

shoumikhin:
Does the PR name match the changes?

cccclai:
> Does the PR name match the changes?

Ah no it doesn't. I rename the diff, but the PR's name doesn't change...

shoumikhin (approved):

PR #9298, opened 2025-03-14
Make BroadcastIndexesRange efficient if there is no broadcasting

swolchok:
It seems to be OK to just check for broadcasting when creating the
iterator and then leave a highly-predictable branch inside the
loop. Avoids code near-duplication to handle both broadcast and
non-broadcast cases.


swolchok:
* (to be filled)

manuelcandales (approved):

swolchok:
noting CI is green before rebasing due to spurious conflicts before land

PR #9302, opened 2025-03-14
Add Phi-4-mini README.md

jackzhxng:


examples/models/phi-4-mini/README.md
```diff
@@ -0,0 +1,63 @@
...
+  -X \
+  --metadata '{"get_bos_id":151643, "get_eos_ids":[151643]}' \
+  --output_name="phi-4-mini.pte"
+  --verbose
+```
+
+Run using the executor runner:
+```
+# Currently a work in progress, just need to enable HuggingFace json tokenizer in C++.
+# In the meantime, can run with an example Python runner with pybindings:
```
tarun292, line 51: Add a note here to suggest to users to make sure they have installed ET with pybindings? Although is that the default now already?

tarun292 (approved):

examples/models/phi-4-mini/README.md
jackzhxng: Yup, by default now

PR #9325, opened 2025-03-17
Fix optional graph_signature access in memory planning

jathu:
### Summary

CI is broken due to https://github.com/pytorch/executorch/pull/8031. `graph_signature` is optional, and is None.

### Test plan

CI

cc @JacobSzwejbka @angelayi

### DIFF
--- a/exir/memory_planning.py
+++ b/exir/memory_planning.py
@@ -248,16 +248,7 @@ class Verifier:
                     has_dynamic_unbound_output |= has_dynamic_unbound_tensor
 
         # only check if inputs are allocated if there are user inputs:
-        user_inputs_exist = (
-            len(
-                list(
-                    filter(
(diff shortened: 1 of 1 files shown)


exir/memory_planning.py
```diff
@@ -339,6 +332,22 @@ def _is_mutable_buffer(
     return False
 
 
+def _does_user_inputs_exist(graph_signature: Optional[ExportGraphSignature]) -> bool:
```
swolchok, line 335: nit: subject-verb agreement, should be either `_does_user_input_exist` or `_do_user_inputs_exist`

swolchok (approved):

exir/memory_planning.py
jathu: touché 😅

PR #9326, opened 2025-03-17
set -x in build_apple_frameworks.sh

swolchok:
This causes commands to be echoed as they are executed, making it easier to track progress in CI.


jathu (approved):

swolchok:
has previous green CI run and is just a script change, merging

PR #6545, opened 2024-10-29
Remove 0 bytes lines; merge getitem with batch_norm nodes

Olivia-liu:
Differential Revision: D65121374

PR #9338, opened 2025-03-17
Remove ops with 0 bytes memory allocation in memory profiler. (#6545)

Gasoonjia:
Summary:

During graph conversion we may introduce extra operators whose lifetime overlapped with existing operators and not have any memory allocation  (e.g. several getitem op before and after batch_norm nodes). Such operators may have a thin line in our memory profiler and overlapped with the which makes user hard to understand the profiling result.

This update gets rid of such situation by merging the overlapped operators and removed the op with no memory allocation.

Differential Revision: D65121374

### DIFF
--- a/util/activation_memory_profiler.py
+++ b/util/activation_memory_profiler.py
@@ -29,6 +29,7 @@ class Allocation:
     size_bytes: int
     fqn: str
     file_and_line_num: str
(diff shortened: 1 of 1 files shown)


Gasoonjia:
"topic: not user facing"

PR #9355, opened 2025-03-18
Qualcomm AI Engine Direct - Add submodule quant config setting

chunit-quic:
- Add API to qnn quantizer for setting submodule quant config
- Refine QnnQuantizer setting functions


chunit-quic:
[@pytorchbot](https://github.com/pytorchbot) label "release notes: qualcomm"

cccclai:
@sxu @billmguo can you help reviewing this pr and see if it meets the need?

backends/qualcomm/quantizer/quantizer.py
```diff
@@ -133,37 +177,55 @@ def _annotate(self, gm: GraphModule) -> None:
...
+            'L__self___add': ('add', 'executorch.backends.qualcomm.tests.models.Add')
+        }
+        """
+
+        nn_module_stack = node.meta.get("nn_module_stack")
+        if nn_module_stack:
+            module_source_str, module_str = list(nn_module_stack.values())[-1][
+                -1
+            ].rsplit(".", 1)
+            module_source = importlib.import_module(module_source_str)
```
sxu, line 202: The module might not be available for import. For example when the output of `torch.export.export` is done ahead of time. If I'm not mistaken loading the exported program doesn't require source modules to be available.

backends/qualcomm/quantizer/quantizer.py
```diff
@@ -112,18 +114,60 @@ class QuantDtype(IntEnum):
...
         super().__init__()
         self.quant_ops: Set[OpOverload] = self.SUPPORTED_OPS.copy()
 
-        self.is_qat = False
-        self.quant_dtype = QuantDtype.use_8a8w
-        self.quant_config: QuantizationConfig = get_8a8w_qnn_ptq_config()
-        self.per_channel_quant_config = get_ptq_per_channel_quant_config()
-        self.use_per_channel_weight_quant_ops: Set[OpOverload] = set()
+        self.default_quant_config = ModuleQConfig()
+        self.module_qconfig_dict: Dict[torch.nn.Module, ModuleQConfig] = {}
```
sxu, line 170: This works for setting different qconfigs by submodule type, but another important use case is by subodule name. For example both `self.backbone` and `self.head` contain linear layers, but we want 8-bit activation for backbone and 16-bit for head.

How about making this a `List[Tuple[Callable[[fx.Node], bool], ModelQConfig]]`? First element of the tuple is a predicate determining if the qconfig should be applied to a given node. Each predicate in the list is evaluate sequentially, the earlier ones will have priority over later ones. Then we can provide some utility for creating common predicates, for example:
```
def get_submodule_type_predicate(module_type):
    def predicate(node):
        if nn_module_stack := node.meta.get("nn_module_stack"):
            return module_type in (x[1] for x in nn_module_stack.values())
        return False
```
or
```
def get_submodule_name_predicate(module_name):
    def predicate(node):
        if nn_module_stack := node.meta.get("nn_module_stack"):
            return module_name in (x[1] for x in nn_module_stack.keys())
        return False
```
which can be used like this:
```
predicate_qconfigs = [
    (get_submodule_name_predicate("self.head"), my_16a8w_qconfig),
    (get_submodule_type_predicate("torch.nn.Linear"), my_8a8w_qconfig),
    (lambda n: True, default_qconfig),
]
```

backends/qualcomm/quantizer/quantizer.py
```diff
@@ -133,37 +177,55 @@ def _annotate(self, gm: GraphModule) -> None:
...
+        An example of nn_module_stack
+        {
+            'L__self__': ('', 'executorch.backends.qualcomm.tests.models.SubModules'),
+            'L__self___add': ('add', 'executorch.backends.qualcomm.tests.models.Add')
+        }
+        """
+
+        nn_module_stack = node.meta.get("nn_module_stack")
+        if nn_module_stack:
+            module_source_str, module_str = list(nn_module_stack.values())[-1][
```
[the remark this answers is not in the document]
chunit-quic: Hi @sxu,

Thank you for reviewing!
Yes, based on [the description nn_module_stack](https://pytorch.org/docs/stable/export.ir_spec.html#metadata) from pytorch.org, the order is determined by the stack trace. We can find the same [the -1 logic](https://github.com/pytorch/executorch/blob/1ea101e2d2451e3651fd5d8349c71c8e48a8834f/util/activation_memory_profiler.py#L46) also applied in the other file within the repo.

backends/qualcomm/quantizer/quantizer.py
chunit-quic: Could you please provide a failed example? To my knowledge, you must have a Torch neural network to invoke torch.export.export. To have a Torch neural network, you must import the corresponding modules. Therefore, it seems to me that we are guaranteed to have the module available for import here.

backends/qualcomm/quantizer/quantizer.py
chunit-quic: Thank you very much for providing the use cases and pseudo codes! 
They clarify some of the requirements we were unsure about. I have a few questions to ensure we fully understand the requirements: 
1. So, basically, all submodules belonging to the given name or submodule type should share the same qconfig, correct? 
2. Is there any data structure more robust than a string? We find it challenging to find one after export. As in your example, it seems we can only rely on the string from nn_module_stack, which leads us to use importlib for additional assurance. 
3. Similar to point 2, the name of self.head is defined by users in nn.Module. Therefore, an arbitrary name might appear in any module, which could cause us accidentally set qconfig to an unintended target. Ask more information (like what moudle type user want) can work around this. Yet I guess we might need a more reliable data structure to search for them in the future.

backends/qualcomm/quantizer/quantizer.py
chunit-quic: Quick update. Thanks to your comments, we found that in some cases, certain modules failed to import. Switching to string-based comparison will bypass this issue.
One more thing, could you please share more thoughts on [this comment](https://github.com/pytorch/executorch/pull/9355#discussion_r2015380498)? Thank you :)

backends/qualcomm/quantizer/quantizer.py
chunit-quic: Just a quick update. We change the mapping way based on your comments. Please feel free to let us know any problem. Thank you. :D

sxu (approved):
Sorry I missed this for the last couple of days, but thanks for the follow-up! LGTM.

cccclai:
There is a linterror, can you fix it?

chunit-quic:
> There is a linterror, can you fix it?

Thanks for pointing that out! Fixed.

chunit-quic:
Hi @cccclai, if this PR looks good to you, could you please help merge it? Thank you. :)

cccclai:
Yes, sorry for being late, merging.

PR #9358, opened 2025-03-18
Arm backend: improve common.parameterize decorator

Erik-Lundell:
- Set xfails to strict to detect if failing tests start working
- Add possibility to pass expected Error type


Erik-Lundell:
CI failures unrelated from what I can tell

zingo (approved):

PR #9359, opened 2025-03-18
Arm backend: support int16 and int32 output tables.

Erik-Lundell:
Change insert_table_ops to support int16 table generation. Insert rescale in insert_table_ops if needed.
Remove guard for int16 on insert_rescales.
The new test shows how the quantizer can be configured to change the quantization of Sigmoid.

### DIFF
--- a/backends/arm/_passes/insert_table_ops.py
+++ b/backends/arm/_passes/insert_table_ops.py
@@ -1,5 +1,4 @@
 # Copyright 2024-2025 Arm Limited and/or its affiliates.
-# All rights reserved.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -18,6 +17,7 @@ from executorch.exir.dialects.edge._ops import EdgeOpOverload
 
 from executorch.exir.pass_base import ExportPass, PassResult
 from torch.fx import GraphModule
+
 from torch.library import impl, Library
 
 lib = Library("tosa", "DEF")
@@ -26,7 +26,10 @@ lib.define("_table(Tensor self) -> Tensor")
 
 @impl(lib, "_table")
 def _table_impl(*args, **kwargs):  # pyre-ignore
-    return args[0]
+    in_dtype = args[0].dtype
+    if in_dtype == torch.int8:
+        return args[0]
+    return args[0].to(dtype=torch.int32)
 
 
 class InsertTableOpsPass(ExportPass):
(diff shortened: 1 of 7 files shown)


backends/arm/_passes/insert_table_ops.py
```diff
@@ -59,29 +62,89 @@ def register_buffer(self, buffer_name: str, buffer: torch.Tensor) -> None:
...
+                    steps=256,
+                    # use torch.int64 to avoid overflow when dequantizing (subtracting zp).
+                    # e.g. torch.tensor(-50, dtype=torch.int8) - 100 == torch.tensor(106, dtype=torch.int8)
+                    dtype=torch.int64,
+                )
+            ).to(dtype=torch.int8),
+            0,
+        )
+
+    def generate_16_bit_table_values(
```
digantdesai, line 94: This logic is tricky to follow. Please add more comments, and links to the spec? https://www.mlplatform.org/tosa/tosa_spec.html#_table 🙏

backends/arm/_passes/insert_table_ops.py
```diff
@@ -59,29 +62,89 @@ def register_buffer(self, buffer_name: str, buffer: torch.Tensor) -> None:
...
                 # use torch.int64 to avoid overflow when dequantizing (subtracting zp).
                 # e.g. torch.tensor(-50, dtype=torch.int8) - 100 == torch.tensor(106, dtype=torch.int8)
                 dtype=torch.int64,
             )
-        ).to(dtype=input_dtype)
+        )
+        # Calculate how much we need to shift table values to fit in 16 bits
+        # ceil(log2(max absolute table value)) + 1 bit for signedness - 16
+        # Note: for out_quantargs.dtype=torch.int16, rshift == 0.
+        rshift = int(torch.ceil(torch.log2(lut_values.abs().max()))) + 1 - 16
```
digantdesai, line 125: Not sure why do we need both ceil and + 1?

backends/arm/operators/op_rescale.py
```diff
@@ -48,7 +47,10 @@ def define_node(
                 f"If output dtype is not int8, output_zp must be 0. Got {output_dtype=}, {output_zp=}"
             )
 
-        scale_width = 32 if output_dtype == torch.int32 else 16
+        # scale32 gives higher accuracy but for a higher HW cost.
+        # For now, always go for scale32.
+        scale_32 = True
```
digantdesai, line 52: - how much more to use wider scale?
- why do we always want to do mul_t = i32_t? can we check `multiplier`, `shift` and input value dtype to decide better?
- do we care to specify rounding mode if scale_32 = True?

backends/arm/test/ops/test_sigmoid_16bit.py
```diff
@@ -0,0 +1,212 @@
...
+    return qconfig
+
+
+def get_16bit_sigmoid_quantizer(tosa_str: str):
+    tosa_spec = common.TosaSpecification.create_from_string(tosa_str)
+    quantizer = TOSAQuantizer(tosa_spec)
+    quantizer.set_global(get_symmetric_quantization_config())
+    quantizer.set_io(_get_16_bit_quant_config())
+    quantizer.set_module_type(
+        "torch.nn.modules.activation.Sigmoid", _get_16_bit_quant_config()
```
digantdesai, line 44: Nice

backends/arm/test/ops/test_sigmoid_16bit.py
```diff
@@ -0,0 +1,212 @@
...
+        bias=None,
+    )
+    return qconfig
+
+
+def get_16bit_sigmoid_quantizer(tosa_str: str):
+    tosa_spec = common.TosaSpecification.create_from_string(tosa_str)
+    quantizer = TOSAQuantizer(tosa_spec)
+    quantizer.set_global(get_symmetric_quantization_config())
+    quantizer.set_io(_get_16_bit_quant_config())
```
digantdesai, line 42: is this just for testing?

digantdesai (approved):
LGTM. Left some nit comments. rsqrt is next?

backends/arm/_passes/insert_table_ops.py
Erik-Lundell: Say that the max value in the table is 70 000. We want to fit it in 16 signed bits.
70 000=0b10001000101110000 (17 digits) has ceil(log2(70 000)) = ceil(16.095) = 17 bits.
If we shift it 17-16=1 bit, we do get 16 bits (1000100010111000), but due to signedness this is a negative number! So we need to shift it one more bit.

Erik-Lundell:
Thanks for the detailed review Digant! :)

Erik-Lundell:
> LGTM. Left some nit comments. rsqrt is next?

Rsqrt is in the pipeline, but we want to implement it as a special case of pow so we need to get pow done first

backends/arm/operators/op_rescale.py
digantdesai: > We do have an internal ticket for investigating this.
Thanks, just curious. No need to spend more efforts tbh.

> but maybe not in this PR.
sounds OK.

> But maybe I am wrong?
let's keep double=False unless we need to.

backends/arm/_passes/insert_table_ops.py
digantdesai: Thank you!

PR #9367, opened 2025-03-18
Add config system to export_llama

iseeyuan:
### Summary
The configuration system in export_llama is limited
- only via cli command. There's no python API. We cannot export an arbitrary nn.Module
- the cli args are complicated. It's hard for users to get what can be configured

Add yaml file option for export_llama configuration. 
- Move all default values from command line args to config/default.yaml
- Add `--config` option to load user defined configuration yaml file
- To keep it backward compatible, allow cli args override the configs in the yaml file

### Test plan
Reference run:
```
python3 -m examples.models.llama.export_llama -c stories110M.pt -p params.json -d fp32 -n tinyllama_xnnpack+custom_fp32_main.pte -kv -X --xnnpack-extended-ops -qmode 8da4w -G 128 --use_sdpa_with_kv_cache --output-dir tmp
```
With config:
```
python3 -m examples.models.llama.export_llama --config examples/models/llama/config/examples/stories.yaml
```
Those two commands should provide the same .pte file and used_config.yaml

### DIFF
--- a/.ci/scripts/tests/test_llama_config.py
+++ b/.ci/scripts/tests/test_llama_config.py
@@ -1,0 +1,197 @@
+import os
+import pytest
+from pathlib import Path
+from argparse import Namespace
+from examples.models.llama.config.loader import create_config, merge_configs
+from examples.models.llama.config.validation import ConfigValidationError
+
+def test_basic_config_validation():
+    """Test basic configuration validation."""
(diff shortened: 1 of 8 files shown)


iseeyuan:
Creating new PRs starting from https://github.com/pytorch/executorch/pull/9450

PR #9368, opened 2025-03-18
Don't use an iOS toolchain when building for macOS

jathu:
### Summary
I'm not sure if it's correct to use an iOS toolchain to build for macOS — couldn't find historical reasons. So, let's default to the host's default toolchain when building for macOS?

### Test plan

CI

```bash
$ rm -rf cmake-out /Users/jathu/executorch/examples/apple/coreml/executor_runner/libraries /Users/jathu/executorch/backends/apple/coreml/executorch-cmake-out && CMAKE_BUILD_DIR=/Users/jathu/executorch/backends/apple/coreml/executorch-cmake-out ./examples/apple/coreml/scripts/build_executor_runner.sh
$ rm -rf cmake-out && ./build/build_apple_frameworks.sh --Release
$ rm -rf cmake-out && ./build/build_apple_frameworks.sh --Debug
$ rm -rf cmake-out && ./build/build_apple_frameworks.sh --Release --toolchain="/tmp/fake"
```

cc @larryliu0820 @lucylq @shoumikhin @cbilgin


shoumikhin:
@jathu `ios.toolchain.cmake` name is misleading, it in fact contains everything needed to build for other Apple platforms beyond iOS, thus we use it. Do you have any concerns using those definitions, or see some benefits in not using them?

PR #9371, opened 2025-03-18
Add basic ambiguity check in the tests

digantdesai:
### Summary
This is also an example of how to use the new dim_order APIs to detect
ambiguity when coming from PyTorch Tensor to ExecuTorch.

### Test plan
```
python -m unittest exir.tests.test_memory_format_ops_pass.TestMemoryFormatOpsPass
```


exir/tests/test_memory_format_ops_pass.py
```diff
@@ -124,8 +125,33 @@ def test_op_dim_order_propagation(self) -> None:
...
+                            dtype=torch.float32,
+                            memory_format=torch.contiguous_format,
+                        ),
+                    ),
+                    target_memory_format=torch.channels_last,
+                    _load_for_executorch_from_buffer=_load_for_executorch_from_buffer,
+                ),
+                check_unambiguous_dim_order=True,
+            )
+        except AmbiguousDimOrderError as e:
```
Gasoonjia, line 152: the test right now can be passed when
a. AmbiguousDimOrderError raised, or
b. no error occurs.

b is not what we expected. 
I would like to add an assertion in the bottom of try section,
```suggestion
        assert False, "This test should be failed"
        except AmbiguousDimOrderError as e:
```

exir/tests/test_memory_format_ops_pass_utils.py
```diff
@@ -99,10 +104,57 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    # This is an example of how one can detect ambiguous dim_order anywhere in the graph.
+    # You can be surgical and only detect it in the nodes you are interested in or something else.
+    def detect_ambiguity(gm):
+        """
+        Check every node's output tensor dim_order and raise if it is ambiguous for a list of formats.
+        """
+        for node in gm.graph.nodes:
+            if node.op == "call_function":
+                tensor = node.meta["val"]
+                if not isinstance(tensor, torch.Tensor):
```
Gasoonjia, line 134: Maybe here we need also take care of tensor list?

exir/tests/test_memory_format_ops_pass_utils.py
```diff
@@ -99,10 +104,57 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return t1 * t2
 
 
+class AmbiguousDimOrderError(RuntimeError):
```
Gasoonjia, line 107: I have a broader question is, why should be create this check?
This check went through every output tensor in the model and check whether its dim order is ambiguous. However, we should only care about the operator that involving dim order, otherwise
 a. ambiguous or not may not impact the behavior and
 b. may introduce extra false neagtive (e.g. slice op may introduce lots of ambiguous dim order tensor)

exir/tests/test_memory_format_ops_pass_utils.py
digantdesai: Fixed in [865c7f8](https://github.com/pytorch/executorch/pull/9371/commits/865c7f8b1267ba6d5ad5cbee5ced24c909a33514)

Gasoonjia (approved):
LGTM, plz solve the lint issue before landing it!

PR #9348, opened 2025-03-18
Use parallel_for in functional_util's apply_unary_map_fun

swolchok:
The other ones are reductions.

More #8932 rollout.

PR #9373, opened 2025-03-18
Unbreak BroadcastIndexesRange::operator+= when there is no broadcasting

swolchok:
operator+= had a loop over 0 elements in this case, resulting in the indices array being full of zeros. Added a += test to our test case that covers this.


swolchok:
* #9348
* #9243
* __->__ #9373

PR #9381, opened 2025-03-18
Create the foundation for testing CoreML export for wheels

jathu:
### Summary
As part of https://github.com/pytorch/executorch/issues/9019, this will allow us to test CoreML support for wheels.

#### Why are we only selectively running the executor test? 

Our `macos-*-stable` runners do not have xcode, which is required to build and run these tests. Only GitHub's `macos-*-xlarge` runners have xcode. We currently run these CoreML tests on [trunk](https://github.com/pytorch/executorch/blob/7c6d2f332861907391450974a481b6d8e04732ed/.github/workflows/trunk.yml#L444-L477), which would be way too expensive to use GitHub's runner given the frequency. So let's limit the runner tests to just the wheel jobs. I will work with Test Infra to get Xcode into our runners.

### Test plan

* Wheel tests shouldn't change
* Test locally

```bash
$ GITHUB_WORKSPACE=$(dirname $(pwd)) REPOSITORY=$(basename $(pwd)) python .ci/scripts/wheel/test_macos.py
```

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -209,9 +209,14 @@ test_model_with_qnn() {
   EXPORTED_MODEL=$(find "./${EXPORT_SCRIPT}" -type f -name "${MODEL_NAME}*.pte" -print -quit)
 }
 
+# Run CoreML tests.
+#
+# @param should_test If true, build and test the model using the coreml_executor_runner.
 test_model_with_coreml() {
-  if [[ "${BUILD_TOOL}" == "buck2" ]]; then
-    echo "coreml doesn't support buck2."
+  local should_test="$1"
+
+  if [[ "${BUILD_TOOL}" != "cmake" ]]; then
+    echo "coreml only supports cmake."
     exit 1
   fi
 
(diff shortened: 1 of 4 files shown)


examples/models/__init__.py
```diff
@@ -44,6 +44,7 @@ def __str__(self) -> str:
 
 class Backend(str, Enum):
```
cccclai, line 45: What does this class mean?

jathu: This is just a enum so we have a consistent way of creating tests — using raw strings is error prone. It also makes it easier to refactor, etc.

cccclai: I'd call it recipe as they are more than backends.

jathu: The test itself calls it a backend https://github.com/pytorch/executorch/blob/7c6d2f332861907391450974a481b6d8e04732ed/.ci/scripts/test_model.sh#L25-L29

.ci/scripts/test_model.sh
```diff
@@ -247,7 +260,11 @@ elif [[ "${BACKEND}" == *"qnn"* ]]; then
   fi
 elif [[ "${BACKEND}" == *"coreml"* ]]; then
   echo "Testing ${MODEL_NAME} with coreml..."
-  test_model_with_coreml
+  should_test_coreml=false
+  if [[ "${BACKEND}" == *"verbose"* ]]; then
+    should_test_coreml=true
+  fi
+  test_model_with_coreml "${should_test_coreml}"
```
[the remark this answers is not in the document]
larryliu0820: Sorry just saw the description. I think having a condition is good for now

larryliu0820 (approved):

PR #9386, opened 2025-03-19
Implement unary_ufunc functions using elementwise_util

swolchok:
One less set of independent implementations to worry about going forward (e.g., we don't have to vectorize these separately from elementwise_util and they get all benefits of elementwise_util).


swolchok:
* __->__ #9386
* #11206
* #9584

kernels/portable/cpu/pattern/pattern.h
```diff
@@ -46,48 +46,109 @@ question is a bit more specific, then add a descriptive sufix. */
...
     const Tensor& in,
-    Tensor& out);
+    Tensor& out) {
+  ET_KERNEL_CHECK(ctx, tensor_is_floating_type(out), InvalidArgument, out);
+
+  if (!check_and_resize_inputs(ctx, in, out)) {
+    return out;
+  }
+
+  ET_SWITCH_REALHBBF16_TYPES(in.scalar_type(), ctx, op_name, CTYPE_IN, [&] {
```
manuelcandales, line 140: this is a problem, right? the compute dtype here would be the input's scalar type,

manuelcandales: you need to switch over FLOATHBF16, right?

kernels/portable/cpu/pattern/pattern.h
```diff
@@ -46,48 +46,122 @@ question is a bit more specific, then add a descriptive sufix. */
...
+  ET_KERNEL_CHECK(ctx, tensor_is_floating_type(out), InvalidArgument, out);
+
+  if (!check_and_resize_inputs(ctx, in, out)) {
+    return out;
+  }
+
+  ScalarType compute_type = in.scalar_type() == ScalarType::Double
+      ? ScalarType::Double
+      : ScalarType::Float;
+  ET_SWITCH_FLOAT_TYPES(compute_type, ctx, op_name, CTYPE_COMPUTE, [&] {
```
swolchok, line 151: this is the only ET_SWITCH change remaining -- we are switching over the possible compute types, not the input/output types.

manuelcandales (approved):

swolchok:
tests looking good; few stragglers but it's only llava and then one each of qnn/arm which shouldn't be affected

Issue #9241, opened 2025-03-13
Vectorize optimized_portable_ops versions of portable ops?


swolchok:
### 🚀 The feature, motivation and pitch

Similarly to #8932, we should be able to conditionally compile portable ops to do some vectorization. I imagine that this would look like either passing a second lambda to our util functions, or perhaps passing template lambdas that we then could use for both some scalar T and also `Vectorized<T>`. The second option would require us to get an std-workalike interface to Vectorized operations so that things like `exp` would work seemlessly, which probably would have a similar solution to https://github.com/pytorch/pytorch/issues/144495 .

### Additional context

I've assigned this to myself so that it doesn't go unowned, but I'm not planning to prioritize it immediately.

### RFC (Optional)

As a concrete example, op_add currently calls a util workhorse function with a lambda:
```
    utils::apply_bitensor_elementwise_fn<CTYPE_COMPUTE, op_name>(
        [val_alpha](const CTYPE_COMPUTE val_a, const CTYPE_COMPUTE val_b) {
          return val_a + val_alpha * val_b;
        },
```

We could imagine instead making the call look like this, with a template lambda, so that we could seamlessly use the lambda with Vectorized:
```
    utils::apply_bitensor_elementwise_fn<CTYPE_COMPUTE, op_name>(
        [val_alpha](const auto val_a, const auto val_b) {
          return val_a + val_alpha * val_b;
        },
```

A second, harder example is `op_exp`:

```
Tensor& exp_out(KernelRuntimeContext& ctx, const Tensor& in, Tensor& out) {
  return internal::unary_ufunc_realhbbf16_to_floathbf16(std::exp, ctx, in, out);
}
```

I think ideally we would find a solution to the above-mentioned PyTorch issue and then write this as 

```
Tensor& exp_out(KernelRuntimeContext& ctx, const Tensor& in, Tensor& out) {
  return internal::unary_ufunc_realhbbf16_to_floathbf16_v2([](auto x) { return c10::math::exp(x); }, ctx, in, out);
}
```

using a template lambda that could be instantiated with either a scalar or Vectorized, as outlined above.

swolchok:
> vectorization for elementwise ops

This is trickier than I remembered currently. We outline our loads in elementwise_util.h in the name of build time and code size; getting to a point where vectorization would be the next item on the list would take some time. Accordingly, I am going to skip elementwise_util.h for now and move on to unary_ufunc; will come back to elementwise afterwards.

[...]

PR #9388, opened 2025-03-19
RFC: Specialize for non-mixed-dtype in elementwise_util

swolchok:
Mixed dtype should be uncommon. Here is how we can specialize for the common case. Prepares us to tackle #9241 .

Test Plan: automated tests on this PR verify we didn't break the now-deprecated runtime_out_dtypes mode; tests on the next PR will verify that everything works after migration. Also included migration for exactly one operator, op_mul, to verify that the new code compiles.

To check performance, I edited examples/models/toy_model/model.py so that MulModule used inputs of size 3000, 2000 instead of 3, 2. I exported it with `python3 -m examples.portable.scripts.export --model_name mul` and saved the resulting `mul.pte`. Then I built in release mode with optimized kernels on, but with mul.out removed from kernels/optimized/optimized.yaml, so that we would use the optimized_portable_kernels build of kernels/portable/op_mul.cpp. Finally, I ran 3 trials on my M1 Macbook Pro using `cmake-out/executor_runner  --model_path mul3kby2k.pte --num_executions 1000 --cpu_threads 2`. Resulting times for 1000 iterations in ms:

Previous diff: 8295, 8187, 8139
This diff: 2953, 2806, 2861

(For comparison, the actual optimized mul kernel took around 1000 ms to run 1000 iterations, and #9432 later in the stack arrived at similar numbers.)


swolchok:
* (to be filled)

swolchok:
This is not ready for full review, but I'm leaving it as non-draft because I'd still like directional comments (hence RFC). I have some more PRs I will be adding to the stack once everything builds again.

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -290,6 +290,25 @@ bool check_tensor_dtype(
     SupportedTensorDtypes dtypes,
     const ScalarType compute_type);
 
+/// Return the one output type we are willing to emit specialized code
+/// to handle, given a compute type of CTYPE_COMMON and supported
+/// output types of out_dtypes.
+template <typename CTYPE_COMMON>
```
manuelcandales, line 296: this should be `CTYPE_COMPUTE`, right?

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -290,6 +290,25 @@ bool check_tensor_dtype(
...
+inline constexpr ScalarType specialized_output_scalar_type(
+    SupportedTensorDtypes out_dtypes) {
+  switch (out_dtypes) {
+    case SupportedTensorDtypes::BOOL_OR_BYTE:
+      return ScalarType::Bool;
+    case SupportedTensorDtypes::REALHBBF16:
+    case SupportedTensorDtypes::REALHBF16:
+    case SupportedTensorDtypes::FLOATHBF16:
```
manuelcandales, line 304: this is the place where you would need to handle something like `sin(int8)` having float output.
you would be returning int8 in this case, which is not what you want.

manuelcandales (approved):

PR #9389, opened 2025-03-19
Remove proprietary license

### LINKED ISSUE #9279: Fix license headers for MTK backend
There shouldn't be anything proprietery. According to Contribution License Agreement, the code contributed should be open sourced under BSD license

These files contains contain the word "proprietary"

- https://github.com/pytorch/executorch/blob/main/backends/mediatek/runtime/include/api/NeuronAdapter.h
- https://github.com/pytorch/executorch/blob/main/backends/mediatek/runtime/include/api/NeuronAdapterShim.h
- https://github.com/pytorch/executorch/blob/main/examples/mediatek/executor_runner/mtk_llama_runner.cpp
- https://github.com/pytorch/executorch/blob/main/examples/mediatek/executor_runner/mtk_executor_runner.cpp

You have two options: 

- Remove these files
- Make the files compatible with BSD license. here is an example PR: https://github.com/pytorch/executorch/pull/9271

cc @kimishpatel @iseeyuan @cccclai @neuropilot-captain @cbilgin

neuropilot-captain:
### Summary
- Removed MTK proprietary header: NeauronAdapter.h from repo. Need to be downloaded from NeuroPilot Express SDK.\
- Modify license declaration in multiple header.

Fix #9279


backends/mediatek/runtime/include/api/NeuronAdapter.h
```diff
@@ -1,2385 +0,0 @@
-/* Copyright Statement:
```
cccclai, line 1: Is it intended to delete this file?

cccclai (commented):
@mergennachin can you take a look?

backends/mediatek/runtime/include/api/NeuronAdapter.h
neuropilot-captain: Yes, we have to move this file to NP Express SDK because it is a MTK proprietary header

neuropilot-captain:
Hi @mergennachin, thanks for reviewing this PR.
1. We have removed the license in the latest patch
2. 3. The users will have to copy the header to `api` directory from NeuroPilot Express SDK. We have updated the README for this. The header will be added to the SDK by this month

mergennachin:
Okay, I will merge this PR, as this unblocks the license issue.

But consider a few options moving, as this hinders user adoption experience:

- Automatic script that does necessary stepping the SDK, including the copying .h file.
- Provide a very thin integration header .h file that has more permissible license, and contains only the public API definitions, without any proprietary stuff that can be in any open source repo. And still have a proprietary .h file in the SDK, that defines proprietary stuff.

mergennachin (approved):

PR #9392, opened 2025-03-19
Arm backend: Add tosa_spec info to .tosa files

oscarandersson8218:
Arm backend output files will now be named as `output_tag<N>_<tosa_spec>.tosa` instead of `output_tag.tosa`.


zingo:
Hi @larryliu0820 

> This broke our internal tests and we need to revert it first. Could you please re-submit the PR, and wait for us to import and run internal CI, and paste the error message?

Hi could you help us point out what error you got from this, to help us knowing how to avoid this better in the future or even better could the internal tests be ported to be runned in github also so we can get proper testing while merging?

larryliu0820:
> Hi @larryliu0820
> 
> > This broke our internal tests and we need to revert it first. Could you please re-submit the PR, and wait for us to import and run internal CI, and paste the error message?
> 
> Hi could you help us point out what error you got from this, to help us knowing how to avoid this better in the future or even better could the internal tests be ported to be runned in github also so we can get proper testing while merging?

The original PR was failing all `executorch/backends/arm/test` tests internally. Haven't looked into the details of the tests and also not sure why it's not showing up in OSS CI. I reimported this retry PR and see if it passes this time.

zingo:
This patch renames the tosa files do you have tests that assume some filenames?

larryliu0820:
This seems fine in terms of internal test

larryliu0820 (approved):

zingo:
Hi @larryliu0820 does this mean we should go ahead and merge this again now? Or will you merge it to do the extended testing?

backends/arm/tosa_backend.py
```diff
@@ -122,10 +123,12 @@ def preprocess(  # noqa: C901
 
         if artifact_path:
             tag = _get_first_delegation_tag(graph_module)
+            et_version = version("executorch")
```
digantdesai, line 126: This is causing issues for internal setup, I need to take a look.

zingo: Thanks, This was merged and reverted during your days off and we were asked to reupload the PR. We don't know what caused the problem so the PR was unchanged. One suspicion we have is if you assume and tosa file names in some way in your flow that would break as they get renamed by this patch.

oscarandersson8218: Yes, that's fine for us.

oscarandersson8218:
I messed up the rebase a bit, hence all the added reviewers in the log. Sorry about that. PR should be in a good state again though.

digantdesai (approved):

digantdesai:
LGTM, stamped. Let me pull internally just to be safe (re-revert would be bad :p) - should take a couple of hours. I can merge it if you don't mind.

zingo:
Thanks, please go ahead 🙏

oscarandersson8218:
> Ci failures related? Internal CI is happy.

@digantdesai Don't think they are related. `trunk/test-arm-reference-delegation/linux-job `should be fixed by https://github.com/pytorch/executorch/pull/9641

PR #9394, opened 2025-03-19
Deprioritize top level -O2 in CMAKE_CXX_FLAGS_RELEASE

AdrianLundell:
By prepending rather than appending to CMAKE_CXX_FLAGS_RELEASE, it allows to specify another optimization level earlier in the build process and still have that take precedence over the -O2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
-# Copyright 2024-2025 Arm Limited and/or its affiliates.
 # All rights reserved.
+# Copyright 2024-2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -161,7 +161,7 @@ if(OPTIMIZE_SIZE)
   set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Os")
 else()
   # -O2: Moderate opt.
-  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O2")
+  set(CMAKE_CXX_FLAGS_RELEASE "-O2 ${CMAKE_CXX_FLAGS_RELEASE}")
 endif()
 
 option(EXECUTORCH_BUILD_ANDROID_JNI "Build Android JNI" OFF)


zingo:
@digantdesai this changes a global cmake file you might want to test/review this a bit more maybe?

digantdesai:
I think this is OK. Can we check if "left to right" precedence is generally true beyond gcc, clang, also for msvc for example?

CMakeLists.txt
```diff
@@ -161,7 +161,7 @@ if(OPTIMIZE_SIZE)
   set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Os")
 else()
   # -O2: Moderate opt.
-  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O2")
+  set(CMAKE_CXX_FLAGS_RELEASE "-O2 ${CMAKE_CXX_FLAGS_RELEASE}")
```
digantdesai, line 164: I am not sure if the original intention here is to `force -O2` or `suggest -O2`. I am OK with the latter TBH, as long as we don't rely on this assumption elsewhere already.

zingo: Is there some rationale for not using -O3? If so it would be nice to add it to the comment. -Ofast might not be float complient so that one makes sense to not use.

digantdesai: -O3 can be a code bloat, and does takes longer to compile.

AdrianLundell:
> I think this is OK. Can we check if "left to right" precedence is generally true beyond gcc, clang, also for msvc for example?

This link suggests that it is true for MSVC: https://learn.microsoft.com/en-us/cpp/build/reference/order-of-cl-options?view=msvc-170. Any other important compilers?

A compromise could also be to compile kernels with -O3 and other code with -O2, what do you think about that?

CMakeLists.txt
zingo: @digantdesai ping :)

swolchok: I would recommend landing this PR and sending a separate PR if you want to make another change like dropping -O2. (I would be in favor of that PR but I would want to let other folks, particularly @mergennachin, weigh in.)

AdrianLundell: Sure, I am happy with landing the PR as is if I get an OK.

zingo: Ill rebase to see if we can get the arm testing green

zingo (approved):

zingo:
Got OK to merge from @digantdesai and @swolchok in text form in the commets

PR #9406, opened 2025-03-19
Alternative approach to handling memory offset shift

JakeStevens:
Summary:
To support embedded system builds which threw an error on the warning for left shift by 32 on a 32 bit dtype, the code was modified to:

```
    memory_offset |= static_cast<size_t>(memory_offset_high)
        << (sizeof(size_t) - sizeof(uint32_t));
```

This fails for build of OSS qwen example however.

Instead, we modify to add a check for

```
sizeof(size_t) > sizeof(uint32_t)
```

in the conditional instead of changing the computation.

In our builds of interest, this compiles away the if branch

Reviewed By: digantdesai, dpalmasan

Differential Revision: D71488571


swolchok:
this looks fine, but I'll be surprised if it continues to get rid of the warning

swolchok (approved):

PR #9409, opened 2025-03-19
[trunk.yml] Increase RAM for Arm runners

digantdesai:
Try running arm jobs with linux.2xlarge.memory with 64G RAM.


GregoryComer (approved):

digantdesai:
Unrelated MacOS failure.

PR #9411, opened 2025-03-19
[Cadence] Add scalar cases for binary ops (add, mul, sub, div) on HiFi

mcremon-meta:
Summary:
As titled. Currently those cases will go to the unoptimized broadcast call, which is extremely inefficient. A simple loop will do much better, and can be further optimized later if needed.

Differential Revision: D71495734

### DIFF
--- a/backends/cadence/hifi/operators/op_add.cpp
+++ b/backends/cadence/hifi/operators/op_add.cpp
@@ -138,8 +138,21 @@ Tensor& add_out(
   if ((out_type != ScalarType::Float) || (alpha_val != 1.0))
     optimized = 0;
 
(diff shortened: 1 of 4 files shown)


mcremon-meta:
cc @cad-audio @dijopaul I'll merge this to unblock a couple internal models that showed pretty bad regressions, but I guess these "scalar" cases can be further optimized so I would leave it to you to assess that! No particular rush, this is such a simple op that compiler vectorization is doing pretty well apparently (e.g. we've seen 40M to 123k cycles on one model using mul)

zonglinpeng (approved):

PR #9432, opened 2025-03-20
Add vectorization in elementwise_util

swolchok:
This is a first cut at #9241 . In this PR I've vectorized a small initial set of ops: atan2, clamp, fmod_Scalar, maximum, minimum, mul, pow, and sigmoid. In addition, the following ops should have gotten vectorized automatically because they already used generic lambdas: add, div, rsub, sub.  I've left covering ops that use the `unary_ufunc_*` utilities in  [pattern.h](https://github.com/pytorch/executorch/blob/main/kernels/portable/cpu/pattern/pattern.h) for a follow-up push, because pattern.h and elementwise_util need some work before we can migrate pattern.h's utilities to be backed by elementwise_util.

This PR adds an interesting testing problem: in theory, *all* operators might need test cases long enough to tickle vectorization, because we might accidentally vectorize ops unexpectedly and break their lambdas due to anticipated differences in semantics. I address this issue by using Vectorized for the scalar prologue/epilogue in debug mode (we run tests in both debug and release) so that we can detect broken lambdas. I additionally intentionally introduced a bug in the vectorized path in elementwise_util and manually verified that we saw test failures for each vectorized op called out above.


swolchok:
* #9386
* #11206
* #9584
* #9585
* __->__ #9432

swolchok:
This is draft status because
- need to see PR that migrates ops to vectorization
- need to see size impact of that PR and that it's acceptable

I have confirmed from local measurements with op_mul that this does in fact cause vectorization and it is a significant perf win (matches optimized op_mul for the treat-as-1d case, not too surprisingly).

kimishpatel:
I cant find the pr where we were discussing using vecotrized class for scalar compute, but if we fix the issue in vec_base.h that you highlighted, that is, https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cpu/vec/vec_base.h#L157, size is > 1?, would that make a better solution? Then we can just use Vectorized class everywhere.

kimishpatel:
> We can't use Vectorized everywhere unless we want to commit to keeping the ExecuTorch copy/paste of at::vec::Vectorized in sync and up to date. I would very much like to delete it instead.

Not suggesting to copy/paste it everywhere, but I am not sure what you mean,

> It is also probably going to end up being worse code to use it everywhere

Worse from readability perspective?

kimishpatel:
> from being able to cleanly control when vectorization is used

where do you expect we would have to say not want to vectorize something?

swolchok:
> where do you expect we would have to say not want to vectorize something?

we should not bother generating vectorized loops at least when `at::vec::Vectorized` doesn't have accelerated support for the architecture we are targeting. (I need to update this PR). It is also something we may want to let the user disable to squeeze on size, though I think this is a weaker reason.

swolchok:
This PR is not yet complete (I need to go through and make sure no ops are needlessly held back from vectorization, and I need to overload unary - on at::vec::Vectorized so that op_sigmoid.cpp can vectorize nicely), but it builds and should vectorize a large chunk of elementwise portable ops. Hopefully this makes the direction/vision more clear @kimishpatel

kernels/portable/cpu/op_where.cpp
```diff
@@ -47,7 +47,7 @@ Tensor& where_out(
         CTYPE_COMPUTE,
         op_name,
         utils::SupportedTensorDtypes::SAME_AS_COMMON>(
-        [](const auto val_a, const auto val_b, const auto val_c) {
+        [](const CTYPE_COMPUTE val_a, const CTYPE_COMPUTE val_b, const CTYPE_COMPUTE val_c) {
           return val_c ? val_a : val_b;
         },
```
swolchok, line 52: (ATen, and therefore our own optimized op_where, doesn't vectorize this)

kernels/portable/cpu/util/vectorized_math.h
```diff
@@ -0,0 +1,111 @@
...
+ * Internal-usage macro for making a vectorized variant of a unary
+ * function available in the executorch::math namespace.
+ */
+#define ET_INTERNAL_VECTORIZED_FLOAT_UNARY_FUNC(func_name)               \
+  namespace executorch {                                                 \
+  inline namespace math {                                                \
+  template <typename T>                                                  \
+  auto func_name(at::vec::Vectorized<T> vec) {                           \
+    if constexpr (!::executorch::runtime::is_floating_point<T>::value) { \
+      return at::vec::convert<float>(vec).func_name();                   \
```
kimishpatel, line 35: is this a valid thing to do? that is convert say an instance of `at::vec::Vectorized<int8_t>` to float and apply the `func_name`? Maybe i am misunderstanding how this works

kernels/portable/cpu/util/vectorized_math.h
```diff
@@ -0,0 +1,111 @@
...
+  } // namespace executorch
+
+#ifdef ET_USE_PYTORCH_HEADERS
+/**
+ * Internal-usage macro for making a vectorized variant of a unary
+ * function available in the executorch::math namespace.
+ */
+#define ET_INTERNAL_VECTORIZED_FLOAT_UNARY_FUNC(func_name)               \
+  namespace executorch {                                                 \
+  inline namespace math {                                                \
```
kimishpatel, line 31: what does inline to namespace do?

kernels/portable/cpu/util/vectorized_math.h
```diff
@@ -0,0 +1,111 @@
...
+  }                                                                          \
+  }                                                                          \
+  }
+
+/**
+ * Internal-usage macro for making a C++ standard library
+ * floating-point function and a vectorized variant of it available in
+ * the c10::math namespace. Should be used with functions where the
+ * corresponding operator is a "float op" in TensorIterator parlance
+ * (i.e., uses something like build_borrowing_binary_float_op()),
```
kimishpatel, line 63: I dont know if anyone reading this code would understand what this means?

kimishpatel: But this provides answer to my earlier question

kernels/portable/cpu/op_where.cpp
```diff
@@ -47,7 +47,7 @@ Tensor& where_out(
         CTYPE_COMPUTE,
         op_name,
         utils::SupportedTensorDtypes::SAME_AS_COMMON>(
-        [](const auto val_a, const auto val_b, const auto val_c) {
+        [](const CTYPE_COMPUTE val_a, const CTYPE_COMPUTE val_b, const CTYPE_COMPUTE val_c) {
```
kimishpatel, line 50: is this unrelated change?

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -51,6 +56,34 @@ inline int64_t scalar_to<int64_t>(const Scalar& s) {
 }
 
 namespace internal {
+template <typename Ignore, typename T>
+using ignore_first_yield_second = T;
```
kimishpatel, line 60: I like these names

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -51,6 +56,34 @@ inline int64_t scalar_to<int64_t>(const Scalar& s) {
...
+// at::vec::Vectorized<CTYPE_COMPUTE>?
+//
+// See [NOTE: Generic lambdas] below for requirements on Op.
+template <typename CTYPE_COMPUTE, typename Op, typename... Args>
+constexpr bool can_use_vectorized() {
+  using Vec = at::vec::Vectorized<CTYPE_COMPUTE>;
+  if constexpr (std::is_invocable_v<
+                    Op,
+                    ignore_first_yield_second<Args, Vec>...>) {
+    // For bool, we will get a false positive if we rely on only the
```
kimishpatel, line 73: you mean bool return type?

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -61,8 +94,71 @@ inline void dtype_specialized_elementwise_fn_impl(
...
+  static_assert(
+      (std::is_same_v<Args, std::pair<const Tensor*, SupportedTensorDtypes>> &&
+       ...));
   constexpr auto kNumInputs = sizeof...(inputs);
-  ET_DCHECK(((inputs.first->element_size() == sizeof(CTYPE_COMPUTE)) && ...));
+  // All inputs must be of type CTYPE_COMPUTE.
+  ET_DCHECK(
+      ((inputs.first->scalar_type() ==
+        CppTypeToScalarType<CTYPE_COMPUTE>::value) &&
+       ...));
```
kimishpatel, line 105: Man, this is a good reminder of all the template meta programming magic

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -61,8 +94,71 @@ inline void dtype_specialized_elementwise_fn_impl(
...
+
+#ifdef ET_USE_PYTORCH_HEADERS
+  if constexpr (can_use_vectorized<CTYPE_COMPUTE, Op, Args...>()) {
+    const bool any_is_broadcasted =
+        !(torch::executor::internal::sizes_match_ignoring_leading_1s(
+              inputs.first->sizes(), out.sizes()) &&
+          ...);
+    if (!any_is_broadcasted) {
+      using Vec = at::vec::Vectorized<CTYPE_COMPUTE>;
+      ::executorch::extension::parallel_for(
```
kimishpatel, line 115: I think doing this blindly for each op is a bit risky in that, no all multithreading is always better. some ops benefit from smaller grain size while others with larger

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -61,8 +94,71 @@ inline void dtype_specialized_elementwise_fn_impl(
...
+          out.numel(),
+          ::executorch::extension::internal::GRAIN_SIZE,
+          [&](const auto begin, const auto end) {
+            std::array<const CTYPE_COMPUTE*, kNumInputs> inputs_data_ptrs = {
+                inputs.first->template const_data_ptr<CTYPE_COMPUTE>()...};
+
+            CTYPE_OUT* const data_out = out.mutable_data_ptr<CTYPE_OUT>();
+
+            const auto vectorized_begin =
+                begin + (Vec::size() - begin % Vec::size()) % Vec::size();
```
kimishpatel, line 126: Feels like something that has chances of bug. Hope we test this enough. I would doubt if our test cases will exercise this code path.

kimishpatel: although I do see you treat scalar left overs of both head and tails separately

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -61,8 +94,71 @@ inline void dtype_specialized_elementwise_fn_impl(
...
+       ...));
+
+#ifdef ET_USE_PYTORCH_HEADERS
+  if constexpr (can_use_vectorized<CTYPE_COMPUTE, Op, Args...>()) {
+    const bool any_is_broadcasted =
+        !(torch::executor::internal::sizes_match_ignoring_leading_1s(
+              inputs.first->sizes(), out.sizes()) &&
+          ...);
+    if (!any_is_broadcasted) {
+      using Vec = at::vec::Vectorized<CTYPE_COMPUTE>;
```
kimishpatel, line 114: So the one point of contention for me is that why do we need vectorized_math.h which largely is doing trampoline to underlying vectorized<T> methods. Mainly you dont even need to use `can_use_vectorized`, because on non accelerated platforms Vectorized falls back to scalar impl even if Vec::size() != `. Maybe you said that the generated code would be worse if forced Vectorized, but I am not sure why. Rest makes sense.

However, place where I can potentially see that being useful is for dtype that Vectorized doesnt support but for float I am not sure. So maybe if you can clarify that it would help.

kimishpatel (commented):
Overall this makes sense but I left some questions around why we need `can_use_vectorized` based approach

kernels/portable/cpu/op_where.cpp
swolchok: no, we can't vectorize this op

kernels/portable/cpu/util/elementwise_util.h
swolchok: bool dtype

kernels/portable/cpu/util/elementwise_util.h
swolchok: IIRC xnnpack blindly parallelizes absolutely everything; we're doing strictly better here

kernels/portable/cpu/util/elementwise_util.h
swolchok: > why do we need vectorized_math.h which largely is doing trampoline to underlying vectorized methods

without it, you can't take the same lambda you already wrote for scalars and reuse it for Vectorized (the change isn't zero because you have to point at executorch::math, but crucially it doesn't require separate code)

kernels/portable/cpu/util/elementwise_util.h
kimishpatel: yeah I am not comparing with xnnpack. In fact one bit part of the reason why we ended up leveraging optimized op lib for some of the llama stuff for exactly that reason. That it blindly parallelized everything and that actually hurt perf

kimishpatel:
Ok we synced offline.

So the major value prop as I understood: You can write your lambdas without having to explicitly use Vectorized. Why explicitly using Vectorized might not be good? Because maybe Vectorized does not have everything you need to implement your lambda. So as op author you dont have to worry about Vectorized when writing your lambda (although do have to use executorch::math ops). And later if Vectorized added support for the ::math.. ops you get vectorization for free without having to go back and rewrite your lambda with Vectorized. So this is nice.

kernels/portable/cpu/util/vectorized_math.h
swolchok: this should be fixed in the latest update.

kernels/portable/cpu/util/elementwise_util.h
swolchok: > Hope we test this enough.

Good point; adding tests for each affected op with lengths sufficient to hit the vectorized path.

swolchok:
> This is still in draft state. Are you planning to push more changes?

I was waiting for CI to be green before bothering reviewers. Will publish shortly

manuelcandales (approved):

swolchok:
starting stacked land, noting that CI is green

Issue #9117, opened 2025-03-11
Rename top-level build folder to simplify wheel building


jathu:
Based on the [python docs](https://packaging.python.org/en/latest/tutorials/packaging-projects/), and the [PyTorch wheel building workflow](https://github.com/pytorch/test-infra/blob/e072651fd15d522eea3ea5d363044f5c0c7ad8eb/.github/workflows/build_wheels_linux.yml#L233-L254), building a wheel should just take the following steps:

```bash
# Installs deps referenced in pyproject.toml → build-system.requires
$ python3 -m pip install --upgrade build

# Builds the wheel
$ python3 -m build --wheel
```

Unfortunately, since we have a top-level folder called `build`, I think python is attempting to use that as a module — which is invalid. If we rename the folder from `build`, we should be able to simplify our CI setup to just those steps. 

Although, the effort to rename the folder seems large and daunting. Is it worth the simplified workflow?

cc @larryliu0820 @lucylq

GregoryComer:
I'm in favor of this. When I first started working on ET, I ended up doing cmake builds in the build directory multiple times and making a mess.

mergennachin:
> The effort to rename the folder seems large and daunting. Is it worth the simplified workflow?

Why is this large and daunting? Isn't this a one-time codemod in fbsource?

jathu:
> Why is this large and daunting? Isn't this a one-time codemod in fbsource?

We have to capture all explicit usage of `build/`, but there might be also some implicit references. Not impossible, but would require a back and forth on updating, then letting the CI provide some feedback. 

However, the bigger impact would be to update our docs that reference build scripts under that folder. This will also make any external tutorials stale since those scripts will not exist. Not a huge problem, but is it worth the effort to reduce wheel building by 1-2 steps?

larryliu0820:
So this only happens when you run `python3 -m build --wheel` inside `executorch/`, right?

larryliu0820:
Ah ok I just ran these commands locally and now I understand the issue. Yeah it seems there's no way around but rename `build/`.

mergennachin:
> However, the bigger impact would be to update our docs that reference build scripts under that folder. This will also make any external tutorials stale since those scripts will not exist. Not a huge problem, but is it worth the effort to reduce wheel building by 1-2 steps?

That's a general problem. But we have a way to solve today by version. We have two versions of the doc. 

https://pytorch.org/executorch/main/ - that is mirror of the main branch
[...]

PR #9434, opened 2025-03-20
[build Folder Migration] Move core build files into scripts

jathu:
### Summary
The final diff as part of https://github.com/pytorch/executorch/issues/9117. This is the big one that affects users — we finally move the core build scripts into `scripts/`

### Test plan

CI

cc @larryliu0820 @lucylq


swolchok (approved):
yep, this is a move. I would recommend importing just to make sure it works

mergennachin (approved):

GregoryComer (approved):

PR #9435, opened 2025-03-20
[backend_api] Delete partitioner tags after lowering

mcr229:
* #9378
* __->__ #9435

There was a slight bug i found when lowering with:

XnnpackDynamicallyQuantizedPartitioner --> XnnpackPartitioner. The issue arrises because the second time we partition, delegation tags from the previous partitioner still exist. Specifically they exist on the getitem node because the metadata was propagated.

See:
https://github.com/pytorch/pytorch/blob/main/torch/fx/passes/utils/fuser_utils.py#L235

Since the getitem nodes created in by the XnnpackDynamicallyQuantizedPartitioner still have the delegation tags, they sneak into partitions from XnnpackPartitioner. This shouldn't happen because delegation tags from previous Partitioner should not exist after the previous lowering.

We update the code in to_backend here to erase all the "delegation_tags" from node.meta after we've performed all `_partition_and_lower`

Differential Revision: [D71518362](https://our.internmc.facebook.com/intern/diff/D71518362/)


tarun292 (approved):

exir/backend/backend_api.py
```diff
@@ -401,6 +401,12 @@ def to_backend(
         tagged_exported_program,
     )
 
+    # Partitioner added delegation tags to the graph module nodes,
+    # we make sure to remove them after we finished partition_and_lower
+    for node in tagged_graph_module.graph.nodes:
+        if "delegation_tag" in node.meta:
+            del node.meta["delegation_tag"]
```
cccclai, line 408:
```suggestion
        node.meta.pop("delegation_tag", None)
```

cccclai (approved):
Thanks, just a minor comment

PR #9440, opened 2025-03-20
MethodMeta API to query of backends number and names

shoumikhin:
Add a way to learn which backends are used for a particular Method

### DIFF
--- a/runtime/executor/method_meta.cpp
+++ b/runtime/executor/method_meta.cpp
@@ -239,6 +239,19 @@ bool MethodMeta::uses_backend(const char* backend_name) const {
   return false;
 }
 
+size_t MethodMeta::num_backends() const {
+  const auto delegates = s_plan_->delegates();
+  return delegates ? delegates->size() : 0;
+}
+
+const char* MethodMeta::get_backend_name(size_t index) const {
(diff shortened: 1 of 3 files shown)


kirklandsign (approved):

mergennachin:
Hi @shoumikhin,

There's already very similar API added in https://github.com/pytorch/executorch/pull/8520 and https://github.com/pytorch/executorch/pull/8738

Do we need all these similar APIs - please add more context in the PR summary.

We should consolidate them and disambiguiate which ones to use. 

Also the "topic: not user facing" is not appropriate -- since this is a user facing API addition.

Cc @cccclai @tarun292 , @larryliu0820 , @JacobSzwejbka

JacobSzwejbka:
I think the query on a single model is a reasonable addition, I agree it should have public api release notes. 

I can imagine this being used once we have some support for device or for backend options

tarun292:
@shoumikhin yep i think when you add the api release notes for this also please add when a user should use which API. Although it's reasonably clear to us, for new users i don't think it's very clear.

cccclai:
This one seems helpful. The APIs I added is for the registered backend, but not the backend required from the models

PR #9441, opened 2025-03-20
Add Module API to query for the number of methods

shoumikhin:


kirklandsign (approved):
Maybe explain what's the use case in description?

PR #9442, opened 2025-03-20
[Android docs] Add maven part

kirklandsign:
### Summary
Now user can also add executorch android lib as a maven dependency.

### Test plan
CI


GregoryComer (approved):

GregoryComer:
Thanks for adding this. Out of curiosity, is there a way to add a maven dependency on the latest stable version of the package, so that we don't have to update the link on each release?

PR #9443, opened 2025-03-20
Link quantized ops lib in xnn_executor_runner

madhu-fb:
Summary: Running exported models with quantized embedding ops and such requires these kernels to be linked at runtime.

Differential Revision: D71527836


iseeyuan (approved):
Thanks for the fix!

PR #9444, opened 2025-03-20
Arm backend: Limit number of build jobs

mansnils:
In certain build environments, without limiting the number of build jobs, build may abort because memory is running low.


zingo (approved):

digantdesai:
Try reducing the runner ram again?

zingo:
We could try and see if it helps and be prepared to bump it back again.

Could some of the strange MacOS fails also be OOM related?

PR #9448, opened 2025-03-20
Arm Backend: Add New DecomposeSilu pass to arm_pass_manager

ArmRyan:
* Adds DecomposeSilu pass
 * Adds Tests for DecomposeSilu

Change-Id: Ib9f15d04c4c06d92d38cc9e6297145980052e673

### DIFF
--- a/backends/arm/test/ops/test_silu.py
+++ b/backends/arm/test/ops/test_silu.py
@@ -1,0 +1,113 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from typing import Optional, Tuple
+
+import torch
+from executorch.backends.arm.test import common
+from executorch.backends.arm.test.tester.test_pipeline import (
+    EthosU55PipelineBI,
+    EthosU85PipelineBI,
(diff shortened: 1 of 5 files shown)


ArmRyan:
cc: @zingo @oscarandersson8218 @freddan80 @digantdesai

digantdesai:
Updated permissions, so you should be able to run ci and merge.

backends/arm/test/ops/test_silu.py
```diff
@@ -0,0 +1,149 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: no tests for SDPA?

digantdesai (commented):
Silu part LGTM. Add SDPA tests?

backends/arm/test/ops/test_silu.py
ArmRyan: There were no tests added because it was not a new function essentially, just a new call with the idea is that it is tested elsewhere. We have tested it functionally just not with a unit test

digantdesai: adding test_sdpa.py would ensure we can lower and run it. Where else it is tested? Couldn't find it in the test/ops dir.

ArmRyan: Sorry for the delay, had some issues and went on holidays, came back to more issues so I have removed SDPA for now

zingo (approved):

Issue #9449, opened 2025-03-20
[etLLM] New config system to export_llama


iseeyuan:
## Current status
The configuration system in export_llama is limited:
- Only via cli command. There's no python API. We cannot export an arbitrary nn.Module
- The cli args are complicated. It's hard for users to get what can be configured
- It's a bad practice to have flat configs and pass it deeply into various functions

## Proposal
- Similar to the torchtune config system, use Hydra and omegaconf, so that the configs can be organized in yaml file, and overrides by cli.
- Use structured config, like config.model, config.quantization, etc. Only pass sub configs to each component.
- Keep config and sub configs at high level and directly pass field values  
- An example of config definition in https://gist.github.com/iseeyuan/df9cd283b46dc574d1d506324258f59b 

## Steps
Since export_llama is used in many places, we do incremental changes. 
1. convert args to DictConfig with structure. 
2. Refactor functions for composability and unblock internal use
3. Add new entry point using Hydra
4. Deprecate the legacy argparser systme.

cc @larryliu0820 @mergennachin @cccclai @helunwencser @jackzhxng

kimishpatel:
I +1 to moving to config based thing rather than "add yet another arg to cli"

larryliu0820:
Yes please! I can add some food for thoughts:

* Testability - this config system should be very easy to test
* Composability - allow users to fine tune the yaml based config easily
* Separation of concern - ideally we should have model configs, export configs and optrace, etrecord configs, different configs for different backends should be nicely grouped together (coreml, qnn etc)

PR #9450, opened 2025-03-20
[etLLM][Config, Part1] Convert Args to DictConfig

iseeyuan:
### Summary
1. Add structure to the complicated args
2. Convert args to DictConfig, to decouple the cli args
3. Pass needed sub configs to functions (instead of args)

### Test plan
```
python3 -m examples.models.llama.export_llama -c stories110M.pt -p params.json -d fp32 -n tinyllama_xnnpack+custom_fp32_main.pte -kv -X --xnnpack-extended-ops -qmode 8da4w -G 128 --use_sdpa_with_kv_cache --output-dir tmp -E 8,64
```

### DIFF
--- a/examples/models/llama/eval_llama_lib.py
+++ b/examples/models/llama/eval_llama_lib.py
@@ -4,34 +4,26 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-
 import argparse
-
 from typing import Optional, Union
 
 import torch
-
-from datasets import load_dataset
 from executorch.examples.models.llama.export_llama_lib import (
+    _convert_args_to_config,
+    _prepare_for_llama_export,
+    build_args_parser as _build_args_parser,
     get_quantizer_and_quant_params,
 )
 from executorch.examples.models.llama.tokenizer.tiktoken import Tokenizer as Tiktoken
-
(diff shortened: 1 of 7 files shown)


examples/models/llama/eval_llama_lib.py
```diff
@@ -230,18 +222,94 @@ def gen_eval_wrapper(
...
+    config = _convert_args_to_config(args)
+
+    # Add evaluation-specific settings
+    config.eval = OmegaConf.create()
+    config.eval.tasks = args.tasks
+    config.eval.limit = args.limit
+    config.eval.num_fewshot = args.num_fewshot
+    config.eval.pte = args.pte
+    config.eval.tokenizer_bin = args.tokenizer_bin
+    config.eval.output_eager_checkpoint_file = args.output_eager_checkpoint_file
```
larryliu0820, line 305: Sorry is there a definition of the config?

larryliu0820 (approved):

examples/models/llama/eval_llama_lib.py
iseeyuan: This PR is just the first step: to convert args to config to unblock internal work, where the configs can be used standalone, without cli args and yaml file. More context in https://github.com/pytorch/executorch/issues/9449

PR #9451, opened 2025-03-20
Arm backend: Tosa tools update

per:
### Summary
Move the tosa tools v0.80 to a separate namespace to be able to install the upcoming 1.0 release
in it's place.

### Test plan
Tested through the Arm backend unit test flow.


digantdesai (commented):
I think I understand what is happening but can't say I understand the rationale? Are we planning to have two versions co-exist?

Also this will cause chaos internally if we merge it now :)

per:
@digantdesai Yes, we plan to have the two versions coexist.
The rationale is that we want to start test the 1.0 TOSA flow (including github CI) and also be able to test backend compilers transition to 1.0 (on internal CI).
When everything is green on 1.0 we'll pull out the 0.80 version.

digantdesai:
> When everything is green on 1.0 we'll pull out the 0.80 version.

Though OK for GH OSS, it will be a headache internally. Curious, can we test the 1.0 on a branch on GH and once we are ready we can just drop 0.80? Assuming 1.0 is BC with 0.80 of tosa tools. This will save some throwaway efforts on my side internally.

per:
> 1.0 on a branch on GH and once we are ready we can just drop 0.80? Assuming 1.0 is BC with 0.80 of tosa tools

The 1.0 tooling is _not_ backward compatible with 0.80 unfortunately. Getting this in will let us develop the 1.0 support for reference in parallel as the compiler gets it's support.

per:
@digantdesai Good to go?

per:
Unrelated failure: [trunk / test-models-linux-aarch64 (phi_4_mini, portable) / linux-job (push)](https://github.com/pytorch/executorch/actions/runs/14092879138/job/39473701205?pr=9451)

backends/arm/third-party/reference_model/patches/v0.80/reference_model/0001-Move-tosa-tools-to-be-namespaced-into-tosa-tools.v0_.patch
```diff
@@ -0,0 +1,154 @@
+From 20c2059723d5c6952cecfb7fcde92601639ef825 Mon Sep 17 00:00:00 2001
```
digantdesai, line 1: we need to pull this in

backends/arm/third-party/reference_model/patches/v0.80/serialization_lib/0001-Make-TOSA-serializer-lib-to-be-self-contained.patch
```diff
@@ -0,0 +1,283 @@
+From b3c8c3f779a7e051826f317598fb831fa9cfe923 Mon Sep 17 00:00:00 2001
```
digantdesai, line 1: And this

digantdesai (commented):
I pulled this in, let's see how many different ways it fails, and what I need to do to fix it. Let's hold on to merging this until then? I should be able to do this early next week.

digantdesai:
Sigh something didn't work right, running the internal CI again

zingo (approved):

zingo:
@digantdesai  
> Sigh something didn't work right, running the internal CI again

Is the problems due to the change or other random things?
Is there some error you can share?

digantdesai:
ok so the pull worked this time, as expected broke a bunch of things internally because this module doesn't exist. I am on PTO today, I will take a stab at this on Mon. Sorry for blocking this :(

digantdesai:
Ok got this to work internally. I can merge it with internal stack or you can merge it and I can forward fix it.

If you are doing it, please rebase, fix where and erf ops, and I can pull in again, while it merges with internal, and get ready to apply the fix.

Else if you want me to merge it then perhaps, I have to commandeer this one and merge it with few other commits.

digantdesai (approved):

digantdesai:
@per - https://github.com/pytorch/executorch/blob/main/backends/arm/operators/op_where.py#L8

per:
> @per - https://github.com/pytorch/executorch/blob/main/backends/arm/operators/op_where.py#L8

@digantdesai Fixed here https://github.com/pytorch/executorch/commit/e3a511193350b72156c3927d3766f029914dd3c8#diff-bac7978e44db415ee063435c3f4a2eebe951296f01ade26b18afd4c9d30faa09R8 , right?
Or am I missing something?

per:
Failures are caused by #9938.

SS-JIA:
Will revert the PR as it is causing some internal failures as well.

SS-JIA:
Decided to try and forward fix with https://github.com/pytorch/executorch/pull/10043 since the PR has been synced to our internal repo already. PTAL!

PR #9452, opened 2025-03-20
Arm backend: Fix flaky test LSTM by bumping allowed atol/rtol slightly

zingo:
This fix some rare fails on the LSTM test as we seem to have been too close with the values.


digantdesai (approved):
Thanks. We can debug ATOL/RTOL issues in parallel and keep relaxed version on CI if we want, as opposed to unmarked flaky tests.

PR #9453, opened 2025-03-20
Make nightly versions consistent

mergennachin:
Summary:

- Make the git hash from 3/11 nightly consistent with https://hud.pytorch.org/pytorch/pytorch/commit/d962699548c6740bf5d0f3ccda933de3b6fca643
- Make torchtune nightly consistent


zingo (approved):

install_requirements.py
```diff
@@ -67,6 +67,9 @@ def python_is_compatible():
 # NOTE: If a newly-fetched version of the executorch repo changes the value of
 # NIGHTLY_VERSION, you should re-run this script to install the necessary
 # package versions.
+#
+# NOTE: If you're changing, make the corresponding change in .ci/docker/ci_commit_pins/pytorch.txt
+# by picking the hash from the same date in https://hud.pytorch.org/hud/pytorch/pytorch/nightly/
```
jathu, line 72: There might be an opportunity to force compliance.

1) We should create `./tools/setup/pytorch_nightly_version.txt`, which hosts his version.
2) In both files, we should read directly from that file

```python
NIGHTLY_VERSION = open("${ROOT}/tools/setup/pytorch_nightly_version.txt", "r").read().strip()
```

```bash
NIGHTLY_VERSION=$(cat "${ROOT}/tools/setup/pytorch_nightly_version.txt")
```

jackzhxng: ^+1

mergennachin: Sorry, how do we find the mapping between git commit hash and nightly version?

PR #9454, opened 2025-03-20
Arm backend: Add info about setup_path.sh and arm-none-eabi-gcc

zingo:
Describe that you need to run setup_path.sh when starting a new shell and that you might want to make sure the right arm-none-eabi-gcc is used.


digantdesai (approved):
Thanks!

zingo:
MacOS fail are unrelated

PR #9456, opened 2025-03-20
Make export llama checkpoint and param optional

jackzhxng:
Summary: Refactor to make checkpoint path and param path optional.

Differential Revision: D71404805

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -123,26 +123,19 @@ def verbose_export():
 
 
 def build_model(
-    modelname: str = "llama3",
-    extra_opts: str = "",
-    *,
-    par_local_output: bool = False,
-    resource_pkg_name: str = __name__,
+    model: str,
+    checkpoint: str,
+    params: str,
+    output_dir: Optional[str] = ".",
+    extra_opts: Optional[str] = "",
 ) -> str:
-    if False:  # par_local_output:
-        output_dir_path = "par:."
-    else:
-        output_dir_path = "."
-
-    argString = f"--model {modelname} --checkpoint par:model_ckpt.pt --params par:model_params.json {extra_opts} --output-dir {output_dir_path}"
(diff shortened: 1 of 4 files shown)


examples/models/llama/model_args.py
```diff
@@ -8,7 +8,7 @@ class ModelArgs:
     n_layers: int = 32
     n_heads: int = 32
     n_kv_heads: Optional[int] = None
-    vocab_size: int = -1  # defined later by tokenizer
+    vocab_size: int = 512  # Arbitrary value, should be defined later by tokenizer.
```
larryliu0820, line 11: Why do we set it to arbitrary value?

jackzhxng: We need an actual value to fall back on when a params file isn't provided so that the embedding layer can be instantiated

examples/models/llama/export_llama_lib.py
```diff
@@ -191,7 +184,6 @@ def build_args_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "-c",
         "--checkpoint",
-        default=f"{ckpt_dir}/params/demo_rand_params.pth",
```
larryliu0820, line 194: add `required = False`?

examples/models/llama/export_llama_lib.py
```diff
@@ -273,7 +265,6 @@ def build_args_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "-p",
         "--params",
-        default=f"{ckpt_dir}/params/demo_config.json",
```
larryliu0820, line 276: Same here.

examples/models/llama/export_llama_lib.py
```diff
@@ -273,7 +265,6 @@ def build_args_parser() -> argparse.ArgumentParser:
...
         help="config.json",
```
larryliu0820, line 268: Talk about the behavior when this option is not provided

larryliu0820 (approved):

PR #9462, opened 2025-03-20
Use xplat/caffe2:ovrsource_aten_Config.h for arvr build modes

tarun292:
Summary: For arvr modes we should use `//xplat/caffe2:ovrsource_aten_Config.h` to get the ATen config headers as the `ovrsource` variants are what's used in arvr repos and using ``"//xplat/caffe2:ovrsource_aten_Config.h"` causes the ovrsource ones to get overwritten and hence results in build failures.

Differential Revision: D71506180

### DIFF
--- a/runtime/core/portable_type/c10/c10/targets.bzl
+++ b/runtime/core/portable_type/c10/c10/targets.bzl
@@ -1,4 +1,4 @@
-load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime")
+load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime", "is_arvr_mode")
 
 def get_sleef_preprocessor_flags():
     if runtime.is_oss:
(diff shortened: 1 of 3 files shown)


runtime/core/portable_type/c10/c10/targets.bzl
```diff
@@ -78,9 +78,8 @@ def define_common_targets():
         ] + get_sleef_preprocessor_flags(),
         xplat_exported_deps = [
             "//xplat/caffe2:aten_header",
-            "//xplat/caffe2:generated_aten_config_header",
             "//xplat/caffe2/c10:c10_headers",
-        ],
+        ] + ["//xplat/caffe2:ovrsource_aten_Config.h" if is_arvr_mode() else "//xplat/caffe2:generated_aten_config_header",],
```
swolchok, line 82: do we also need to cover line 63 above?

tarun292: shouldn't be needed as arvr only uses xplat_exported_deps and not fbcode_exported_deps

swolchok (approved):

PR #9472, opened 2025-03-20
Applied quantization for linear with bias=True in pre_quantization

YIWENX14:
Summary: Applied quantization for linear with bias=True in pre_quantization and checkpoint conversion. Verified small checkpoint of speech encoder eager model.

Differential Revision: D71573144


larryliu0820 (approved):

Gasoonjia:
@YIWENX14 looks like this pr breaks oss ci, can you take a look?

larryliu0820:
@Gasoonjia which one?

Gasoonjia:
image
https://github.com/pytorch/executorch/actions/runs/13980352986/job/39144137101#step:9:2878
this one

Gasoonjia:
@larryliu0820 oh great thanks for confirm that. Just from the ci board looks like this pr introducing the issue.

PR #9473, opened 2025-03-20
Revert "Arm backend: Add initial Llama model test case"

larryliu0820:
Reverts pytorch/executorch#8679


kirklandsign (approved):

digantdesai:
No need to revert, forward fixed it -[D71577157](https://www.internalfb.com/diff/D71577157). I don't think I could export it before it landed. @larryliu0820 let's sync?

larryliu0820:
Forward fix landed

PR #9481, opened 2025-03-21
Separate CoreML util and inmemoryfs targets to be reusable

jathu:
### Summary
In this diff, we separate the targets for `coreml/runtime/util` and `coreml/runtime/inmemoryfs`.

#### Why?

As part of https://github.com/pytorch/executorch/issues/9019, we want to support CoreML export out-of-the-box. Unfortunately, part of the workflow now includes installing an executorch+coreml binding as a pip wheel:
 https://github.com/pytorch/executorch/blob/5fdfa511966208c7e237a9e920a7c63f513b4fb7/backends/apple/coreml/scripts/install_inmemoryfs.sh#L20

This then gets used throughout the library. i.e.

https://github.com/pytorch/executorch/blob/5fdfa511966208c7e237a9e920a7c63f513b4fb7/backends/apple/coreml/compiler/coreml_preprocess.py#L19

So we want to now bake this pybinding into the main wheel. Ultimately this `executorchcoreml` module, gets built using inmemoryfs:

https://github.com/pytorch/executorch/blob/5fdfa511966208c7e237a9e920a7c63f513b4fb7/backends/apple/coreml/runtime/inmemoryfs/setup.py#L17-L38

So, to enable creating a pybinding target, we must first decouple the `util` and `inmemoryfs` libraries so they can be used in `coremldelegate` and the future pybinding.

### Test plan
This should really be a no-op w.r.t. builds. So, CI +
```
$ ./install_executorch.sh --pybind coreml
```

cc @larryliu0820 @lucylq

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,6 +67,3 @@
 [submodule "shim"]
 	path = shim
 	url = https://github.com/facebook/buck2-shims-meta
-[submodule "backends/apple/coreml/third-party/json"]
-	path = backends/apple/coreml/third-party/json
-	url = https://github.com/nlohmann/json.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,3 +67,6 @@
 [submodule "shim"]
 	path = shim
 	url = https://github.com/facebook/buck2-shims-meta
+[submodule "backends/apple/coreml/third-party/json"]
+	path = backends/apple/coreml/third-party/json
+	url = https://github.com/nlohmann/json.git
--- a/backends/apple/coreml/.gitignore
+++ b/backends/apple/coreml/.gitignore
@@ -20,8 +20,7 @@ runtime/inmemoryfs/build/
 runtime/test/models/
 runtime/sdk/format/
 runtime/include/executorch/
-third-party/coremltools/
-third-party/nlohmann_json/
+third-party/
 
 xcode-build/
 
--- a/backends/apple/coreml/.gitignore
+++ b/backends/apple/coreml/.gitignore
@@ -20,7 +20,8 @@ runtime/inmemoryfs/build/
 runtime/test/models/
 runtime/sdk/format/
 runtime/include/executorch/
-third-party/
+third-party/coremltools/
+third-party/nlohmann_json/
 
 xcode-build/
 
--- a/backends/apple/coreml/CMakeLists.txt
+++ b/backends/apple/coreml/CMakeLists.txt
@@ -116,20 +116,31 @@ set(PROTOBUF_SOURCES
     runtime/sdk/format/WordTagger.pb.cc
 )
 
+find_library(FOUNDATION_FRAMEWORK Foundation)
+
 # CoreML util
-add_library(coreml_util)
-target_sources(coreml_util PRIVATE ${UTIL_SOURCES})
-target_include_directories(coreml_util
-  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/util
+add_library(coreml_util ${UTIL_SOURCES})
+target_include_directories(coreml_util PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/util)
+target_link_libraries(coreml_util PRIVATE ${FOUNDATION_FRAMEWORK})
+
+install(
+  TARGETS coreml_util
+  DESTINATION lib
+  INCLUDES
+  DESTINATION ${_common_include_directories}
 )
 
 # CoreML inmemoryfs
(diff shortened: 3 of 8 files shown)


larryliu0820 (approved):

backends/apple/coreml/CMakeLists.txt
```diff
@@ -111,11 +99,35 @@ set(PROTOBUF_SOURCES
     runtime/sdk/format/WordTagger.pb.cc
 )
```
mergennachin, line 101: Add Meta copyright on top of the file, in addition to existing Apple one.

backends/apple/coreml/CMakeLists.txt
```diff
@@ -111,11 +99,35 @@ set(PROTOBUF_SOURCES
...
 
+# CoreML util
+add_library(coreml_util)
+target_sources(coreml_util
+  PRIVATE runtime/util/json_util.cpp
```
mergennachin, line 105: just use PRIVATE ${UTIL_SOURCES}

backends/apple/coreml/CMakeLists.txt
```diff
@@ -111,11 +99,35 @@ set(PROTOBUF_SOURCES
...
+          runtime/util/objc_json_serde.mm
+)
+target_include_directories(coreml_util
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/util
+)
+
+# CoreML inmemoryfs
+add_library(coreml_inmemoryfs)
+target_sources(coreml_inmemoryfs
+  PRIVATE runtime/inmemoryfs/inmemory_filesystem.cpp
```
mergennachin, line 115: Just use PRIVATE ${INMEMORYFS_SOURCES}

backends/apple/coreml/CMakeLists.txt
```diff
@@ -111,11 +99,35 @@ set(PROTOBUF_SOURCES
...
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/runtime/inmemoryfs
+)
+target_link_libraries(coreml_inmemoryfs PRIVATE coreml_util)
+
 # Define the delegate library
 add_library(coremldelegate)
-target_sources(
-  coremldelegate PRIVATE ${INMEMORYFS_SOURCES} ${KVSTORE_SOURCES}
-                         ${DELEGATE_SOURCES} ${UTIL_SOURCES}
+target_sources(coremldelegate
```
mergennachin, line 128: can this be just one line?

mergennachin (approved):
See inline comments

mergennachin (approved):

backends/apple/coreml/.gitignore
```diff
@@ -20,7 +20,8 @@ runtime/inmemoryfs/build/
 runtime/test/models/
 runtime/sdk/format/
 runtime/include/executorch/
-third-party/
+third-party/coremltools/
+third-party/nlohmann_json/
```
jathu, line 24: This will eventually be removed once we start including the pybindings

.gitmodules
```diff
@@ -67,3 +67,6 @@
 [submodule "shim"]
 	path = shim
 	url = https://github.com/facebook/buck2-shims-meta
+[submodule "backends/apple/coreml/third-party/json"]
+	path = backends/apple/coreml/third-party/json
+	url = https://github.com/nlohmann/json.git
```
mergennachin, line 72: don't make this change in this PR

it should be a separate PR

nlohmann/json is not used in AOT, but for, runtime only

so technically we shouldn't be touching this at all for pip wheel for coreml.

backends/apple/coreml/CMakeLists.txt
```diff
@@ -111,32 +118,50 @@ set(PROTOBUF_SOURCES
     runtime/sdk/format/WordTagger.pb.cc
 )
 
+find_library(FOUNDATION_FRAMEWORK Foundation)
+
+# CoreML util
+add_library(coreml_util)
+target_sources(coreml_util PRIVATE ${UTIL_SOURCES})
```
swolchok, line 125: nit: why not put the sources in add_library directly?

swolchok (commented):
seems fine, leaving accept for @shoumikhin

backends/apple/coreml/CMakeLists.txt
jathu: I didn't know that was possible!

cymbalrush (approved):

shoumikhin (approved):

PR #9483, opened 2025-03-21
Enable CoreML by default on macOS wheel builds

jathu:
### Summary
Context: https://github.com/pytorch/executorch/pull/9481

* Include the `executorchcoreml` pybinding in the builds
* Remove separate installation option
* Turn on CoreML by default for macOS builds
* Add a dependency on coremltools for macOS

### Test plan

CI

```
$ rm -rf cmake-out pip-out dist && ./install_executorch.sh
$ ./examples/models/llama/install_requirements.sh
$ .ci/scripts/test_llama.sh -model stories110M -build_tool cmake -dtype fp32 -mode coreml
$ .ci/scripts/test_llama.sh -model stories110M -build_tool cmake -dtype fp32 -mode xnnpack+custom+quantize_kv
```

cc @larryliu0820 @lucylq


mergennachin:
coreml has its own install_requirements.sh script that downloads a coremltools repo. What happens to that?

jathu:
> coreml has its own install_requirements.sh script that downloads a coremltools repo. What happens to that?

In the next diff, I will remove that and install it through pip for only macOS

jathu:
I will update the docs on usage of the install script in another PR.

mergennachin:
This is awesome! 

cc @YifanShenSZ  @cymbalrush

pyproject.toml
```diff
@@ -67,6 +67,8 @@ dependencies=[
   "sympy",
   "tabulate",
   "typing-extensions",
+  # Keep this version in sync with: ./backends/apple/coreml/scripts/install_requirements.sh
+  "coremltools==8.1; platform_system == 'Darwin'",
```
mergennachin, line 71: Niiice!

backends/apple/coreml/scripts/install_requirements.sh
```diff
@@ -43,16 +47,7 @@ fi
...
-
-echo "${green}ExecuTorch: Installing coremltools."
-pip install "$COREMLTOOLS_DIR_PATH"
-
-STATUS=$?
-if [ $STATUS -ne 0 ]; then
-    echo "${red}ExecuTorch: Failed to install coremltools."
-    exit 1
-fi
+cmake --build "$COREMLTOOLS_DIR_PATH/build" --parallel --target mlmodel
```
mergennachin, line 50: @YifanShenSZ @cymbalrush -- any reason why we still need to build mlmodel target from source, instead of using directly from pip install coremltools?

cymbalrush (approved):
Thank you!

backends/apple/coreml/scripts/install_requirements.sh
mergennachin: @jathu 

Okay, let's not block it, but create an GH issue to track this...

mergennachin (approved):

backends/apple/coreml/scripts/install_requirements.sh
jathu: Done https://github.com/pytorch/executorch/issues/9651

PR #9485, opened 2025-03-21
Fix exir.load/save to handle named data store map

tarun292:
Summary: After named data store map was added recently to edge dialect the support for serializing/deserializing named data store map was missing in exir.load/save. This diff adds that and also adds a test to check for this.

Differential Revision: D71601647


pssrawat (approved):
LGTM

PR #9489, opened 2025-03-21
Arm backend: Add all ops not supported on Ethos-U55 to support-check

Erik-Lundell:
Note that this includes ops that are not currently implemented. These ops are added for completeness.

### DIFF
--- a/backends/arm/operator_support/tosa_supported_operators.py
+++ b/backends/arm/operator_support/tosa_supported_operators.py
@@ -208,26 +208,41 @@ class BaseTOSASupportList(OperatorSupportBase):
 class EthosU55NotSupported(OperatorSupportBase):
     """
     Certain operators are not supported on U55. These are listed in `unsupported_ops`.
+    The comment mentions the unsupported TOSA operator that the aten operator maps to where it is not obvious.
+    For unimplemented operators, this is the anticipated mapping, and it might be incorrect.
     """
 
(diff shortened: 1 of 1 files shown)


backends/arm/operator_support/tosa_supported_operators.py
```diff
@@ -208,26 +208,41 @@ def is_node_supported(
 class EthosU55NotSupported(OperatorSupportBase):
     """
     Certain operators are not supported on U55. These are listed in `unsupported_ops`.
+    The comment mentions the unsupported TOSA operator that the aten operator maps to where it is not obvious.
+    For unimplemented operators, this is the anticipated mapping, and it might be incorrect.
     """
 
     unsupported_ops = [
```
digantdesai, line 215: Curious why does this have to be a white-list? As opposed to `all - supported`?

Erik-Lundell: Hm. The current logic is that we have a list of (TOSA-)supported ops, a.k.a. ops that we have implemented. The support on Ethos-U55 is that list minus this list (and some additional constraints, these are just the completely unsupported ops) We don't have an explicit list of ops supported on Ethos-U55. Maybe I am missing your point :grimacing:

zingo:
Lintrunner and MacOs fails are unrelated

zingo (approved):

PR #9494, opened 2025-03-21
Define CMake args and environment variables simultaneously in install script

jathu:
### Summary
This script currently configures CMAKE_ARGS, that is eventually propagated to the underlying cmake build (via pip install). However, notice how for some args, we also define an equivalent envvar. This is primarily for the pip setup script: 

https://github.com/pytorch/executorch/blob/78ee0e67fb7a6b93a461edeb18ec5ff07d4aca50/setup.py#L72-L98

So there is an implicit expectation that if you want to access these definitions in the setup script, you need to make sure to also define the envvar along with the cmake args. Well, it took me more than an hour to figure out why my cmake args wasn't being recognized in the setup script 🤦 . So, to make sure consistency — let's force setting args and envvars simultaneously.  

### Test plan

build + check CMakeCache.txt to ensure flags are set

```bash
# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=ON EXECUTORCH_BUILD_COREML=OFF
$ rm -rf pip-out dist && ./install_executorch.sh

# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=OFF EXECUTORCH_BUILD_COREML=ON
$ rm -rf pip-out dist && ./install_executorch.sh --pybind coreml

# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=ON EXECUTORCH_BUILD_COREML=ON
$ rm -rf pip-out dist && ./install_executorch.sh --pybind xnnpack coreml
```

cc @larryliu0820 @lucylq

### DIFF
--- a/install_executorch.py
+++ b/install_executorch.py
@@ -14,6 +14,7 @@ import os
 import shutil
 import subprocess
 import sys
+from typing import Dict, List, ItemsView
 from contextlib import contextmanager
 
 from install_requirements import (
@@ -167,14 +168,31 @@ def build_args_parser() -> argparse.ArgumentParser:
     )
     return parser
 
+# Helper to keep track of cmake define flags.
+class CmakeDefineFlags:
+    def __init__(self):
+        self._values: Dict[str, str] = {}
+    
+    def add(self, name, value):
+        self._values[name] = value
 
-def handle_pybind(args, cmake_args, executorch_build_pybind):
+    def get(self, name) -> str:
+        return self._values[name]
+
+    def cmake_args(self) -> List[str]:
+        return [f"-D{name}={value}" for name, value in self._values.items()]
+
+    def items(self) -> ItemsView[str, str]:
+        return self._values.items()
+
+
+def handle_pybind(args, cmake_define_flags):
     # Flatten list of lists.
     args.pybind = list(itertools.chain(*args.pybind))
     if "off" in args.pybind:
         if len(args.pybind) != 1:
             raise Exception(f"Cannot combine `off` with other pybinds: {args.pybind}")
-        executorch_build_pybind = "OFF"
+        cmake_define_flags.add("EXECUTORCH_BUILD_PYBIND", "OFF")
     else:
         for pybind_arg in args.pybind:
             if pybind_arg not in VALID_PYBINDS:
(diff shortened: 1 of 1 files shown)


swolchok (changes requested):
This doesn't seem right to me. CMake pays attention to both environment variables and defined flags. We shouldn't have to set both. Can you give a specific motivating example?

jathu:
> This doesn't seem right to me. CMake pays attention to both environment variables and defined flags. We shouldn't have to set both. Can you give a specific motivating example?

The environment variable is not for CMake, it is used by the `setup.py` file we created — I included where we check for it in the description

swolchok:
> > This doesn't seem right to me. CMake pays attention to both environment variables and defined flags. We shouldn't have to set both. Can you give a specific motivating example?
> 
> The environment variable is not for CMake, it is used by the `setup.py` file we created — I included where we check for it in the description

OK, then I think the redundant part is putting the flags in CMAKE_ARGS? setup.py is already responsible for that, right?

jathu:
> > > This doesn't seem right to me. CMake pays attention to both environment variables and defined flags. We shouldn't have to set both. Can you give a specific motivating example?
> > 
> > 
> > The environment variable is not for CMake, it is used by the `setup.py` file we created — I included where we check for it in the description
> 
> OK, then I think the redundant part is putting the flags in CMAKE_ARGS? setup.py is already responsible for that, right?

Unfortunately, setup.py doesn't fully commit to that. For example, the reason it expects some *some* of these flags to be defined as an env var, is so it can easily infer if it is turned on then add additional logic like [putting together pybinging](https://github.com/pytorch/executorch/blob/0c1c36289ebc7200365e23ec0db98a24c53135a1/setup.py#L663-L668).  In this case, it does set the cmake flag *again*. It wouldn't know what environment variable should be defined as a cmake arg — we would need to make that explicit mapping. Which can just be difficult to maintain. So we should continue to let CMAKE_ARGS get set.

I guess we can remove the duplicate definition by splitting apart the `CMAKE_ARGS` in `setup.py`, and getting rid of the environment variable logic entirely. The only problem is this will now expect people to switch their workflow from:

```bash
# Before
$ EXECUTORCH_BUILD_PYBIND=ON python setup.py bdist_wheel

# After
$ CMAKE_ARGS="-DEXECUTORCH_BUILD_PYBIND=ON" python setup.py bdist_wheel
```

I was hesitant to change this since it might break people's current usage, and might require wider CI changes. Note that technically this diff should be neutral change because it maintains the same API — is just forces consistency of the two places of definition instead of doing it manually as it is being done now.

Thinking about it more, if we are okay with breaking people's current workflow of building wheels — I think I'd prefer to get rid of defining environment variables and just relying on CMAKE_ARGS. Thoughts?

PR #9495, opened 2025-03-21
[Android] Use new Llm package API

kirklandsign:


larryliu0820 (approved):
Thank you for doing this!

PR #9505, opened 2025-03-21
Run setup-macos.sh in unittest-macos even in Buck mode

swolchok:
We've been seeing flakes due to missing torchgen that have become more common. After some investigation, it appears that #8688 was probably overzealous: installing pytorch was probably also installing torchgen, so let's ~~install pytorch~~just run the macos setup script to avoid proliferating configurations.

Test Plan: unittest-buck / macos on this PR, monitor to see if failures go away.


swolchok:
* (to be filled)

swolchok:
feel free to merge when accepting this PR

GregoryComer (approved):

PR #9517, opened 2025-03-21
Bug fix in Ethos runtime backend

limintang:
Summary: Cortex-M is 32-bit, so pointer type in the runtime backend is also 32-bit. If the address >= 0x80000000, direct cast to uint64_t treates it as signed integer and sign extends it to 0xFFFFFFFFXXXXXXXX, which causes address validity check in Ethos driver to fail. First cast it to unsigned type uintptr_t then cast to uint64_t avoids this issue.

Differential Revision: D71658384

### DIFF
--- a/backends/arm/runtime/EthosUBackend.cpp
+++ b/backends/arm/runtime/EthosUBackend.cpp
@@ -10,6 +10,7 @@
  * ethos-u-core-driver for hardware interaction.
  */
 
+#include <cstdint>
 #include <cstring>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


zingo (approved):

digantdesai (approved):
Thank you!

zingo (changes requested):
Thanks for the fix, good find. The fix is really good but there is just a small lintrunner error that needs to be fixed to keep HEAD clean if its OK🙏
If not I can also create a new PR tomorrow with fix for both this and the lint problem, no problem. Thanks for the help.

zingo (approved):

PR #9520, opened 2025-03-22
Add portable ELU implementation + test

swolchok:


swolchok:
* (to be filled)

kimishpatel:
did we actually decide to make it a portable op @larryliu0820? If we just want to control its decomp behavior we have other ways to do it as well, right?

swolchok:
> did we actually decide to make it a portable op @larryliu0820? If we just want to control its decomp behavior we have other ways to do it as well, right?

Do we not need portable op implementations for all core ATen ops?

kimishpatel:
> > did we actually decide to make it a portable op @larryliu0820? If we just want to control its decomp behavior we have other ways to do it as well, right?
> 
> Do we not need portable op implementations for all core ATen ops?

Sorry I should have been more clear. I was actually asking "did we actually decide to it a core aten op?". If so yes

swolchok:
> core aten op

@kimishpatel https://github.com/pytorch/pytorch/pull/149684

swolchok:
ping reviewers @manuelcandales @kimishpatel

manuelcandales (approved):

PR #9523, opened 2025-03-22
Fix scalar (single element tensor) binary ops on HiFi

mcremon-meta:
Summary:
As titled, when introduced I used `dim` but I should have used `numel`. Fixed here.
Also need to update the quantizer for the add case, will do in a separate diff.

Differential Revision: D71667998


zonglinpeng (approved):
Ah good catch

skrtskrtfb (approved):

PR #9524, opened 2025-03-22
Android aar update and enable emulator test

kirklandsign:
Fix the maven deps. Should be fbjni.


swolchok (approved):
seems fine if CI passes?

shoumikhin (approved):

PR #9526, opened 2025-03-23
Support channels_last format in portable upsample kernels

GregoryComer:
Summary:
Support channels_last input format in portable CPU upsample_bilinear2d and upsample_nearest2d kernels. This is useful for resize-in-model patterns when the user wants to pass inputs in channels_last format. It also (theoretically) allows for more effective auto-vectorization when vectorizing along the channels dim when there are a larger number of channels.

I considered generalizing the kernel to handle arbitrary dim order, but having a specialized channels last version allows for traversing the output in contiguous order. I could add a separate, arbitrarily-strided variant, but we can take that as a follow-up if needed.

To accomplish this, this PR makes the following changes:
- Update `check_upsample_2d_common_args` to relax the dim order restriction. It now allows for both default and channels_last dim order and verifies that the output dim order matches the input.
- In the upsample kernels (bilinear and nearest), split out NCHW and NHWC variants. The NHWC variant interchanges the loop order as to maintain contiguous output accesses.
- Add test coverage to ensure ATen numerical parity.

Differential Revision: D71690379

### DIFF
--- a/kernels/portable/cpu/op_upsample_bilinear2d.cpp
+++ b/kernels/portable/cpu/op_upsample_bilinear2d.cpp
@@ -20,7 +20,7 @@ using executorch::aten::SizesType;
 
 namespace {
 template <typename CTYPE>
-void upsample_bilinear2d_kernel_impl(
+void upsample_bilinear2d_kernel_impl_nchw(
     const Tensor& in,
     bool align_corners,
     const float scale_h,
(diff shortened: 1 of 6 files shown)


swolchok:
never worked with channels_last so probably not the right person to review this. ping me if I need to just learn it for this review

mcr229:
does this have an affect on binary size?

GregoryComer:
> does this have an affect on binary size?

Good question. It should be pretty small, but I can run a BSB check internally to verify.

kernels/portable/cpu/op_upsample_nearest2d.cpp
```diff
@@ -79,11 +104,25 @@ Tensor& upsample_nearest2d_vec_out(
   const auto kernel_scale_w = area_pixel_compute_scale<double>(
       in.sizes()[3], out.sizes()[3], false, scale_w);
 
+  if (tensor_is_default_dim_order(in)) {
```
manuelcandales, line 107: I don't think there is a need to add this if/else here.
Rename the existing `upsample_nearest2d_kernel_impl` as `upsample_nearest2d_kernel_default` or something like that, and then move this if/else login inside a new `upsample_nearest2d_kernel_impl`.

kernels/portable/cpu/op_upsample_bilinear2d.cpp
```diff
@@ -124,11 +193,24 @@ Tensor& upsample_bilinear2d_vec_out(
   const auto kernel_scale_w = area_pixel_compute_scale<double>(
       in.sizes()[3], out.sizes()[3], align_corners, scale_w);
 
-  ET_SWITCH_REALHBF16_TYPES(
-      in.scalar_type(), ctx, "upsample_bilinear2d.out", CTYPE, [&]() {
-        upsample_bilinear2d_kernel_impl<CTYPE>(
-            in, align_corners, kernel_scale_h, kernel_scale_w, out);
-      });
+  if (executorch::runtime::tensor_is_default_dim_order(in)) {
```
manuelcandales, line 196: see my comment above about not doing the if/else here. It is duplicating the dispatch code.

manuelcandales (approved):

kernels/portable/cpu/op_upsample_nearest2d.cpp
manuelcandales: Not as far as I know. I don't think we have that machinery developed. As a matter of fact, I believe this is the first portable op branching out on dim order.
The few ops that currently support channel last in the portable library, are not dispatching specific code for that. In those cases the code is written in a dim order agnostic way, maybe only requiring that all inputs have the same dim order

PR #9530, opened 2025-03-23
Add test_mimi to ci

iseeyuan:
### Summary
Add test_mimi to ci

### Test plan
ci

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -290,6 +290,36 @@ jobs:
         # run e2e (export, tokenizer and runner)
         PYTHON_EXECUTABLE=python bash .ci/scripts/test_llava.sh
 
+  test-moshi-linux:
+    name: test-moshi-linux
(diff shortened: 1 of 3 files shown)


.github/workflows/pull.yml
```diff
@@ -290,6 +290,33 @@ jobs:
         # run e2e (export, tokenizer and runner)
         PYTHON_EXECUTABLE=python bash .ci/scripts/test_llava.sh
 
+  test-moshi-linux:
+    name: test-llava-runner-linux
```
jackzhxng, line 294: Still says Llava

jackzhxng (commented):
Even though Llava is already in pull, I would prefer not having these large tests in pull, trunk would be better.

.github/workflows/pull.yml
iseeyuan: Good catch!

billmguo (approved):

PR #9536, opened 2025-03-24
Qualcomm AI Engine Direct - Flags for CI

winskuo-quic:
### Summary
Initially, `--compile_only` will use 1 random input for CI purpose. Introducing `--ci` flag, so when users are using `--compile_only`, they won't be getting models with poor accuracy.


winskuo-quic:
Hi @cccclai, @guangy10,
This PR is to add a new flag for external CI, so that `compile_only` flag can still generate a pte with good accuracy.
I also removed some unused code under QNN.
Please have a look.
Thanks

examples/qualcomm/utils.py
```diff
@@ -585,6 +585,13 @@ def setup_common_args_and_variables():
         action="store_true",
     )
 
+    parser.add_argument(
+        "--ci",
+        help="This flag is for Continuous Integration(CI) purpose and is NOT recommended to turn on for typical use cases.",
```
cccclai, line 590:
```suggestion
        help="This flag is for Continuous Integration(CI) purpose and is NOT recommended to turn on for typical use cases. It will use random inputs instead of real inputs.",
```

cccclai (approved):
Makes sense, thanks!

examples/qualcomm/utils.py
winskuo-quic: Updated. Thanks for reviewing the PR!

PR #9537, opened 2025-03-24
Arm backend: Fix flaky sigmoid 16/32 bit tests

Erik-Lundell:
- Change constant test data to random generators.
- Tests on Ethos-U55 are meant to xfail as int 16 tables are currently not supported.
- For other tests, add flaky marker. Remove increased qtol, since the inaccuracies only show up sporadically.


zingo (approved):

zingo:
Fails are unrelated

PR #9542, opened 2025-03-24
Refactor LLMEdgeManager's to_dtype

iseeyuan:
### Summary
Small refactor to aggregate dtype override in to_dtype method.

### Test plan
CI

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -618,25 +618,9 @@ def _prepare_for_llama_export(args) -> LLMEdgeManager:
     )
 
     # At this point, the model is loaded in the default fp32.
-
-    # Checkpoint dtype should be lower or equal precision to the dtype override.
+    # override dtype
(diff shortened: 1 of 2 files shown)


examples/models/llama/export_llama_lib.py
```diff
@@ -588,25 +588,9 @@ def _prepare_for_llama_export(args) -> LLMEdgeManager:
     )
 
     # At this point, the model is loaded in the default fp32.
-
-    # Checkpoint dtype should be lower or equal precision to the dtype override.
+    # override dtype
     checkpoint_dtype = edge_manager.model.checkpoint_dtype
```
iseeyuan, line 592: Not sure why we need to keep this original checkpoint_dtype, since it would be overriden in the line below. Just to keep the behavior of the existing code. @jackzhxng and @kimishpatel , what's the reason behind this, for weight quantization?

larryliu0820 (approved):

billmguo (approved):

jackzhxng (approved):
Should be okay

PR #9543, opened 2025-03-24
Arm backend: Add ethos_u_minimal_example jupyter notebook

AdrianLundell:
This example provides a good base understanding of all steps needed to lower and run a Pytorch module through the EthousUBackend. It also works as a starting point for creating scripts for lowering a custom network.

### DIFF
--- a/examples/arm/ethos_u_minimal_example.ipynb
+++ b/examples/arm/ethos_u_minimal_example.ipynb
@@ -1,0 +1,284 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Copyright 2025 Arm Limited and/or its affiliates.\n",
+    "#\n",
+    "# This source code is licensed under the BSD-style license found in the\n",
+    "# LICENSE file in the root directory of this source tree."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ethos-U delegate flow example\n",
+    "\n",
+    "This guide demonstrates the full flow for running a module on Arm Ethos-U using ExecuTorch. \n",
+    "Tested on Linux x86_64 and macOS aarch64. If something is not working for you, please raise a GitHub issue and tag Arm.\n",
+    "\n",
+    "Before you begin:\n",
+    "1. (In a clean virtual environment with a compatible Python version) Install executorch using `./install_executorch.sh`\n",
+    "2. Install Arm cross-compilation toolchain and simulators using `examples/arm/setup.sh --i-agree-to-the-contained-eula`\n",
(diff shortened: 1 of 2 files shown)


examples/arm/ethos_u_minimal_example.ipynb
```diff
@@ -0,0 +1,284 @@
+{
```
digantdesai, line 1: couple of high level comments,
- without CI this will be broken
- why not google colab and keep it live?
- how is it going to stay in sync with docs given there is a large-ish overlap

AdrianLundell: Good points, here are my thoughts:

- We are going to test this in our internal CI, could definitely be worth it adding it to upstream CI as well.
- What's the added benefit of using google colab? I am not that familiar with it but I haven't had any problems with using a regular notebook. 
- I don't think we have a settled "strategy" on documentation, my personal view is that it makes sense to have two types of slightly overlapping documentation, the detailed main docs, and user guides such as this one. Any other docs, README:s etc, should refer to these rather than repeating info. There will be some work needed to keep these two in sync, but an interactive notebook user guide should be easier to keep in sync compared to a doc since it can be tested in CI, in addition to it being a better user experience.

AdrianLundell: @digantdesai What do you think about this answer?

digantdesai: > What's the added benefit of using google colab? I am not that familiar with it but I haven't had any problems with using a regular notebook.

It can be live on the docs page. Checking in a notebook is also fine, it is just an extra step to actually run it.

> to have two types of slightly overlapping documentation

It would be nice if they can originate from the same source, preferably a code in CI.

AdrianLundell: I see two advantages in leaving it as it is, first it is nice to not be reliant on an external service, and second it works as a sanity check to see that you have set up your environment correctly. I am open to having it on google collab as well but we would probably need to have an internal discussion about it as well. 

I see now what you mean with overlapping documentation with Eriks pull request #9712 since that is also a type of user guide, I was thinking about the general docs about executorch concepts s.a quantization/dialects or similarly as the main docs. Since that PR mentions following an official backend-template I assume that that is something we will have to conform to. 

Running code in CI and then generating a .md and a jupyter notebook seems like a bit of work to get correct, but it might be worth it if it is something used more broadly? Or if we can run the notebook in ci and generate only the .md using some tool (I just now found https://github.com/jupyter/nbconvert by googling for example), although that will probably not follow the template exactly. Yet another solution could be to run the code from the .md and the notebook in CI which would be a little easier but require double updating. Finally we can of course completely skip the notebook idea, I do think that it adds value compared to only .md however in that it is interactive.

digantdesai: I think you have better idea about pros and cons than me on this now. My job here is done. :)

I am Ok with this landing, this is not an easy problem, so hopefully we will converge to a better state.

digantdesai (approved):

zingo (approved):

PR #9548, opened 2025-03-24
Updates torchao pin to enable shared embedding quantization

metascroy:
Updates torchao pin to enable shared embedding quantization.


examples/models/llama/README.md
```diff
@@ -412,6 +412,10 @@ python -m examples.models.llama.export_llama \
   -d fp32
 ```
 
+A few notes:
+- If your model shares embedding/unembedding weights (like Llama1B and Llama3B do), you can add `--use_shared_embedding` to take advantage of this and reduce memory.  When this option is enabled, you can specify whether embeddings are quantized with weight zeros or not by specifying a third argument.  For example, `-E "torchao:4,32,true"` means that the embedding is quantized to 4-bits with group_size=32 and uses weight zeros (this is the default behavior if you simply use `-E "torchao:4,32"`), whereas `-E "torchao:4,32,false"` means that the embedding is quantized to 4-bits with group_size=32, but is quantized with scales-only.  If `--use_shared_embedding` is specified, the unembedding (i.e., the final linear layer) is quantized in the same way, but also uses 8-bit dynamically quantized activations.
```
Jack-Khuu, line 416: Not for this PR, but what's the plan for updating our arg selection scheme for quant?

`-E "torchao:4,32,true` isn't user friendly

examples/models/llama/export_llama_lib.py
```diff
@@ -684,6 +689,15 @@ def _validate_args(args):
     if args.num_sharding > 0 and not args.qnn:
         raise ValueError("Model shard is only supported with qnn backend now.")
 
+    if args.use_shared_embedding:
+        if not (
+            args.embedding_quantize is not None
+            and args.embedding_quantize.startswith("torchao:")
```
Jack-Khuu, line 695:
```suggestion
    if args.use_shared_embedding:
        and (
            args.embedding_quantize is None
            or not args.embedding_quantize.startswith("torchao:")
```

nit: nested conditionals into an error

Jack-Khuu (approved):

examples/models/llama/README.md
metascroy: You'd never need to do that.  true is the default (and existing behavior), so you could continue to use -E"torchao:4,32".

examples/models/llama/export_llama_lib.py
```diff
@@ -1122,6 +1136,21 @@ def _get_source_transforms(  # noqa
 
             transforms.append(inject_fast_hadamard_transform_native_for_spin_quant)
 
+    if args.embedding_quantize:
```
jackzhxng, line 1139: Why did we change the order of the source transform?

examples/models/llama/README.md
jackzhxng: I'd make this a bit more clear that shared is only for torchao kernels, or `torchao:`

examples/models/llama/export_llama_lib.py
metascroy: shared_embedding must be applied before linear.  So I changed order to embedding first, and linear second.  I put a code comment to this effect as well.

PR #9553, opened 2025-03-24
STASH: randn_out

swolchok:
Just the operator implementation for randn_out in case we need it in the future. fails torchgen assertion during building currently and needs a cursory test.


swolchok:
* __->__ #9553

PR #9559, opened 2025-03-24
refactor _Dim into Dim

pianpwk:
Summary: forward fix T218515233

Differential Revision: D71769231

### DIFF
--- a/runtime/executor/tensor_parser_exec_aten.cpp
+++ b/runtime/executor/tensor_parser_exec_aten.cpp
@@ -56,7 +56,7 @@ ET_NODISCARD Result<void*> getMemPlannedPtr(
   const uint32_t memory_offset_high = allocation_info->memory_offset_high();
 
   size_t memory_offset = memory_offset_low;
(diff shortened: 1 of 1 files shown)


backends/xnnpack/test/tester/tester.py
```diff
@@ -496,14 +496,14 @@ def generate_random_inputs(self):
                         # the function to compute the derived dim.
                         dim_spec = dim_spec.root
                         fn = dim_spec.fn
-                    elif isinstance(dim_spec, torch.export.dynamic_shapes._Dim):
+                    elif isinstance(dim_spec, torch.export.dynamic_shapes.Dim):
```
digantdesai, line 499: did this change in pytorch?

jingsh (approved):

backends/xnnpack/test/tester/tester.py
jingsh: looks like in https://github.com/pytorch/pytorch/pull/149891/

PR #9569, opened 2025-03-25
Arm backend: Size Test for Cortex-M

digantdesai:
### Summary
Add size test for cortex-M and add it in the CI


.github/scripts/run_nm.py
```diff
@@ -0,0 +1,167 @@
...
+        except FileNotFoundError:
+            print(f"Error: 'nm' command not found. Please ensure it's installed.")
+            sys.exit(1)
+        except subprocess.CalledProcessError as e:
+            print(f"Error running nm on {elf_file_path}: {e}")
+            print(f"stderr: {e.stderr}")
+            sys.exit(1)
+
+    def _get_nm_output(self) -> Dict[str, Symbol]:
+        args = ["--print-size", "--size-sort", "--reverse-sort", "--demangle"]
```
swolchok, line 50: you might want to specify the format. it can differ between systems.

.github/scripts/run_nm.py
```diff
@@ -0,0 +1,167 @@
...
+        lines = output.splitlines()
+        symbols = []
+
+        def parse_line(line: str) -> Optional[Symbol]:
+            import re
+
+            symbol_pattern = re.compile(
+                r"(?P<addr>[0-9a-fA-F]+)\s+(?P<size>[0-9a-fA-F]+)\s+(?P<type>\w)\s+(?P<name>.+)"
+            )
+            match = symbol_pattern.match(line)
```
swolchok, line 65: as written, compiling this regex doesn't help. you'd need to store it somewhere that outlives a single call to this function.

.github/scripts/run_nm.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
swolchok, line 1: perhaps we could use bloaty?

digantdesai: yeah I did think about it. This is just `nm` TBH so not sure if we need it to be more complicated than that. Wrote a python wrapper to print sorted table, and use cross-tools. But `bloaty` is nice though a bit "bloated" with dependency etc.

.github/scripts/run_nm.py
digantdesai: fixed by moving it up out of the function. Thanks.

.github/scripts/run_nm.py
digantdesai: Added!

.github/scripts/run_nm.py
```diff
@@ -0,0 +1,172 @@
...
+            print(f"> Total bytes: {size_total}")
+            print(
+                f"Counted: {len(symbols_to_print)}/{len(self.symbols)}, {symbol_percent:0.2f}% (filter: '{filter_name}')"
+            )
+            print("=====\n")
+
+        # Print tables with different filters
+        if (
+            self.filter is None
+            or self.filter not in ["all", "executorch", "executorch_text"]
```
swolchok, line 128: nit: isn't the None check redundant?

swolchok: you could make this easier to maintain and extend by creating a FILTER_NAME_TO_FILTER_AND_LABEL table:

```
def is_executorch_symbol(s):
    return "executorch" in s.name or s.name.startswith("et")

FILTER_NAME_TO_FILTER_AND_LABEL = {
    "all": (None, "All"),
    "executorch": (is_executorch_symbol, "ExecuTorch"),
    "executorch_text": (lambda s: is_executorch_symbol(s) and s.symbol_type.lower() == "t", "ExecuTorch .text")
}
```

then access it as
```
filter, label = FILTER_NAME_TO_FILTER_AND_LABEL.get(self.filter, FILTER_NAME_TO_FILTER_AND_LABEL["all"])
print_table(filter, label)
```

and use FILTER_NAME_TO_FILTER_AND_LABEL.keys() in the argument definition below

swolchok (approved):
sure why not

.github/scripts/run_nm.py
```diff
@@ -0,0 +1,171 @@
...
+    parser.add_argument(
+        "-e", "--elf-file-path", required=True, help="Path to the ELF file"
+    )
+    parser.add_argument(
+        "-f",
+        "--filter",
+        required=False,
+        default="all",
+        help="Filter symbols by pre-defined filters",
+        choices=["all", "executorch", "executorch_text"],
```
swolchok, line 159: this should be FILTER_NAME_TO_FILTER_AND_LABEL.keys() (wrapped in list() if we must)

PR #9583, opened 2025-03-25
Remove usage of environment variables in wheel building

jathu:
### Summary

We seem to be using a combination of CMAKE_ARGS and environment variables when creating wheels. Ultimately, CMake only uses the cmake args, however we redefine some of these flags as env vars to help `setup.py` determine if a certain feature is turned on. Specifically, it looks for pybinding vars to bundle pybindings. 

Let's remove this redundancy and just use the CMAKE_ARGS as the single source of truth. For more details and other considerations, see https://github.com/pytorch/executorch/pull/9494 (abandoned).

Note that even in the wheel building jobs, we use cmake args instead of environment variables to control features:

https://github.com/pytorch/executorch/blob/644b7ddf14180d97e348faa627f576e13d367d69/.ci/scripts/wheel/envvar_base.sh#L20

https://github.com/pytorch/executorch/blob/644b7ddf14180d97e348faa627f576e13d367d69/.ci/scripts/wheel/envvar_macos.sh#L14-L15

### Test plan

build + check CMakeCache.txt to ensure flags are set

```bash
# Expected: EXECUTORCH_BUILD_PYBIND=OFF EXECUTORCH_BUILD_XNNPACK=OFF EXECUTORCH_BUILD_COREML=OFF
$ rm -rf pip-out dist && ./install_executorch.sh --pybind off

# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=ON EXECUTORCH_BUILD_COREML=OFF
$ rm -rf pip-out dist && ./install_executorch.sh

# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=OFF EXECUTORCH_BUILD_COREML=ON
$ rm -rf pip-out dist && ./install_executorch.sh --pybind coreml

# Expected: EXECUTORCH_BUILD_PYBIND=ON EXECUTORCH_BUILD_XNNPACK=ON EXECUTORCH_BUILD_COREML=ON
$ rm -rf pip-out dist && ./install_executorch.sh --pybind xnnpack coreml

# Throws an error
$ rm -rf pip-out dist && ./install_executorch.sh --pybind coreml off
```

cc @larryliu0820 @lucylq


larryliu0820 (approved):

install_executorch.py
```diff
@@ -221,15 +219,15 @@ def main(args):
     # If --pybind is not set explicitly for backends (e.g., --pybind xnnpack)
     # or is not turned off explicitly (--pybind off)
     # then install XNNPACK by default.
-    if EXECUTORCH_BUILD_PYBIND == "":
-        EXECUTORCH_BUILD_PYBIND = "ON"
-        CMAKE_ARGS += " -DEXECUTORCH_BUILD_XNNPACK=ON"
+    if not has_pybindings:
+        has_pybindings = True
+        cmake_args.append("-DEXECUTORCH_BUILD_XNNPACK=ON")
```
larryliu0820, line 224: This logic.. I don't know what's going on..

larryliu0820: out of the scope of this PR though.

setup.py
```diff
@@ -60,42 +60,63 @@
...
-            return default
-        if val in ("OFF", "0", ""):
+    def _is_truthy(value: Optional[str]) -> bool:
+        if (value is None) or (value.lower() in ("off", "0", "")):
             return False
         return True
 
+    @staticmethod
+    def _is_cmake_arg_enabled(var: str, default: bool) -> bool:
+        value = _cmake_args_defines().get(var, None)
```
swolchok, line 98: I think @byjlw would say that we should *at least* detect if there is an environment variable named `var` enabled and warn people that they need to change their workflow.

install_executorch.py
mergennachin: before this change, there are three paths:

1. --pybindings off # turns off pybind altogether
2. --pybindings coreml xnnpack # turns on explicitly coreml and xnnpack
3. no pybindings options # turns on xnnpack by default

How do you deal with three situations with only has_pybindings=True | False?

setup.py
jathu: @byjlw always thinking of the customer. Good call

mergennachin (changes requested):
See inline comment

install_executorch.py
jathu: @mergennachin yes I didn't account for just off, made the changes now and included that in the test case in the description

mergennachin (approved):

PR #9586, opened 2025-03-25
NOCOMMIT: PoC: vectorize just expm1

swolchok:
Here is how we can vectorize `expm1` given the prior PRs in this stack. It is NOCOMMIT because we actually should implement vectorization for all the `unary_ufunc_*` ops.

### DIFF
--- a/kernels/portable/cpu/op_expm1.cpp
+++ b/kernels/portable/cpu/op_expm1.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <executorch/kernels/portable/cpu/pattern/pattern.h>
+#include <executorch/kernels/portable/cpu/util/elementwise_util.h>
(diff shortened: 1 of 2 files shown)


swolchok:
* (to be filled)

kernels/portable/cpu/op_expm1.cpp
```diff
@@ -7,16 +7,66 @@
...
 #include <executorch/kernels/portable/cpu/pattern/pattern.h>
+#include <executorch/kernels/portable/cpu/util/elementwise_util.h>
 #include <executorch/runtime/kernel/kernel_includes.h>
 #include <cmath>
 
 namespace torch {
 namespace executor {
 namespace native {
 
+// REVIEW: I'm not entirely sure what the best way to implement this
```
swolchok, line 18: either the original vectorization PR or a preliminary PR will address this

PR #9587, opened 2025-03-25
Use sccache to accelerate android build

kirklandsign:
Android didn't use sccache because it has its own NDK clang. Now use env var to trigger clang from sccache.

### DIFF
--- a/.ci/docker/common/install_cache.sh
+++ b/.ci/docker/common/install_cache.sh
@@ -89,10 +89,26 @@ init_sccache() {
   as_ci_user sccache --zero-stats || true
 }
 
+function write_android_sccache_stub() {
+  BINARY=$1
+  mv "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$1" "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/.$1"
+  cat >"/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$1" <<EOF
+#!/bin/sh
+if [ \$(env -u LD_PRELOAD ps -p \$PPID -o comm=) != sccache ]; then
+  exec sccache /opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/.$1 "\$@"
+else
+  exec /opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/.$1 "\$@"
+fi
+EOF
+  chmod a+x "/opt/toolchains/llvm/prebuilt/linux-x86_64/bin/${BINARY}"
+}
+
 write_sccache_stub cc
 write_sccache_stub c++
 write_sccache_stub gcc
 write_sccache_stub g++
 write_sccache_stub clang
 write_sccache_stub clang++
+write_android_sccache_stub clang
+write_android_sccache_stub clang++
 init_sccache
--- a/.ci/docker/common/install_cache.sh
+++ b/.ci/docker/common/install_cache.sh
@@ -89,28 +89,10 @@ init_sccache() {
   as_ci_user sccache --zero-stats || true
 }
 
-function write_android_sccache_stub() {
-  BINARY=$1
-  mv "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$1" "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/.$1"
-  cat >"/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$1" <<EOF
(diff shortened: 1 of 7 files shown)


.github/workflows/_android.yml
```diff
@@ -22,6 +22,10 @@ jobs:
       script: |
         set -eux
 
+        # Use sccache for NDK compiler as well
+        export CMAKE_CXX_COMPILER_LAUNCHER=sccache
+        export CMAKE_C_COMPILER_LAUNCHER=sccache
```
jathu, line 27: What are your thoughts on exporting this in [.ci/scripts/setup-linux.sh](https://github.com/pytorch/executorch/blob/main/.ci/scripts/setup-linux.sh) so we get it for all CI builds?

jathu: Generally, I'm hoping we can find one place to change and export these variables instead of multiple places. If someone adds another android job, they do not get caching if they forget these lines

kirklandsign: Sorry not sure if that's the situation here. Currently for android build, clang path is /opt/ndk/toolchains/llvm/prebuilt/...bin and we should not use /opt/cache/bin/clang (stub which points to the system clang)?

Basically for android, we need /opt/ndk/.../clang to point to a stub as well?

huydhn: Oh ok, I thought that NDK can just use system clang.  If that the case, you're right.  We need a stub for NDK clang too.  What do you think about 1. Install NDK, 2. Install sccache, 3. When installing sccache, checkif NDK exists and install a stub for it too?

kirklandsign: That should work. I am also wondering, how about just use `export CMAKE_CXX_COMPILER_LAUNCHER= sccache ` so we don't need to write to stub?

huydhn: Using a stub works transparently as part of the Docker image.  Setting `CMAKE_CXX_COMPILER_LAUNCHER` needs to be done later, i.e. people need to remember to set that explicitly.

kirklandsign: Honestly I feel like it's hard to do stub... In NDK we have so many variants of clang
```
armv7a-linux-androideabi21-clang    i686-linux-android29-clang++  x86_64-linux-android21-clang
armv7a-linux-androideabi21-clang++  i686-linux-android30-clang    x86_64-linux-android21-clang++
armv7a-linux-androideabi22-clang    i686-linux-android30-clang++  x86_64-linux-android22-clang
armv7a-linux-androideabi22-clang++  i686-linux-android31-clang    x86_64-linux-android22-clang++
armv7a-linux-androideabi23-clang    i686-linux-android31-clang++  x86_64-linux-android23-clang
armv7a-linux-androideabi23-clang++  i686-linux-android32-clang    x86_64-linux-android23-clang++
```

jathu: @kirklandsign I think you can use [$ANDROID_NDK/ndk-which clang](https://android.googlesource.com/platform/ndk/+/HEAD/ndk-which) to get the specific location

kirklandsign: @jathu I can have a try - just use `clang` instead of these variants (which are basically for `clang --target=x86_64-linux-android21` stuff)

I was just worried if somehow android NDK triggered `clang` via `x86_64-linux-android21-clang` not `clang`

.ci/docker/common/install_cache.sh
```diff
@@ -89,10 +89,26 @@ init_sccache() {
   as_ci_user sccache --zero-stats || true
 }
 
+function write_android_sccache_stub() {
+  BINARY=$1
+  mv "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$1" "/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/.$1"
```
kirklandsign, line 94: /opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/clang
/opt/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/clang++

.github/workflows/_android.yml
kirklandsign: @huydhn @jathu I feel like for android it's still hard to get the stub working. See error like https://github.com/pytorch/executorch/actions/runs/14115338695/job/39544105035?pr=9587 for my version https://github.com/pytorch/executorch/pull/9587/commits/f2e12cb0d54b22734436229e50316cffcca68f9c#diff-0a862d15a4e6dff44196719a594b983aece596c7dc38eba416d83e12a95206e6R92.

huydhn (approved):

PR #9612, opened 2025-03-25
Add mixed dtype check for XNNPACK partitioner

### LINKED ISSUE #9023: Partitioner to not delegate mixed dtype operators
## Problem
Currently there is no check that the XnnpackPartitioner does not partition mixed dtype ops for example a multiply that takes in a bf16 and fp32 input (these are not supported by xnnpack). 

## Context
In the XnnpackPartitioner we have configs for operators that describe how operators should be partitioned, and what constraints they must pass. The below link should show constraints that must be passed by all operators

https://github.com/pytorch/executorch/blob/1a9a59b94491f03620dce622fee922c1d2bc1fa7/backends/xnnpack/partition/config/xnnpack_config.py#L116

We should add a check in this constraint to fail partitions that take in mixed dtypes as inputs. The refactoring done in this issue might be a good way to help enforce this more easily for binary inputs. https://github.com/pytorch/executorch/issues/9024

## Verification/Testing
Add a test with a pytorch module for operations that take in different dtypes (fp32, bf16, fp16, int8). Check that these are not partitioned: 

Inverse of this:
https://github.com/pytorch/executorch/blob/1a9a59b94491f03620dce622fee922c1d2bc1fa7/backends/arm/test/ops/test_cat.py#L69-L70
in that we want to check that there are no call_delegate nodes and the original node is still in the graph

Test for the following representative ops:
- cat
- mul
- add
- div
- sub

## Resources
https://discord.gg/a9r5KZDNfZ

cc @digantdesai @cbilgin

keyprocedure:
### Summary
Fixes #9023
Prevents the partitioner from handling ops with mixed dtypes.

### Test plan
Unable to directly test due to auto-casting of dtypes and existing dtype checks in verifier.py.


mcr229:
do you mind running python -m unittest backends/xnnpack/test/ops/*.py 

just to make sure there aren't any unintended consequences

backends/xnnpack/partition/config/xnnpack_config.py
```diff
@@ -174,11 +175,19 @@ def _check_inputs_are_valid_dtypes(self, node, valid_dtypes):
...
+            if reference_dtype is None:
+                reference_dtype = arg_val.dtype
+            elif arg_val.dtype != reference_dtype:
+                return False
+
         return True
 
     def _check_outputs_are_valid_dtypes(self, node, valid_dtypes):
-        # Check outputs are valid dtype
+        # Check outputs are valid and have the same dtypes
```
mcr229, line 187: i'm hesistant to add the output check, because for ops like max_pooling they return float values and int64 indicies.

backends/xnnpack/partition/config/xnnpack_config.py
```diff
@@ -144,9 +144,10 @@ def check_common_constraints(
         return True
 
     def _check_inputs_are_valid_dtypes(self, node, valid_dtypes):
-        # Check inputs are valid dtypes
+        # Check inputs are valid and have the same dtypes
```
mcr229, line 147: let's just make sure that this is accidentally blocking ops that should work.

mcr229:
hmm seems like:
[pull / unittest-buck / macos / macos-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/14071830170/job/39468303196?pr=9612)

is passing which means all those tests are passing. Let's just try and fix the linter issues.

backends/xnnpack/partition/config/xnnpack_config.py
```diff
@@ -174,11 +175,19 @@ def _check_inputs_are_valid_dtypes(self, node, valid_dtypes):
             if arg_val.dtype not in valid_dtypes:
                 return False
 
+            # Check for mixed dtypes
```
mcr229, line 178: let's add a why log for when we fail. Let's try to add what node fails, and what the dtypes are of the inputs:

https://github.com/pytorch/executorch/blob/74e430812810840cc935d25fd64b558dd106afb6/backends/xnnpack/partition/config/generic_node_configs.py#L156-L159

backends/xnnpack/partition/config/xnnpack_config.py
keyprocedure: Should we only enforce the mixed dtype checks for specific ops like add, sub, mul, div, and cat? For example, I can search for "aten.add" in node.target and only enforce the check when there's a match

mcr229: i think this is fine for now, realistically I think we would use a block list rather than an allow list for this case.

keyprocedure: Sorry, are you saying it's fine as is, or that I should implement the check for specific ops?

mcr229: this is fine as is.

mcr229:
will stamp and merge after all ci is run

mcr229 (approved):

mcr229:
Thank you for your contribution!

PR #9613, opened 2025-03-25
elementwise_util: s/common/compute/ almost everywhere and deprecate SAME_AS_COMPUTE

swolchok:
As the title says, this is mostly a few related find-replaces, plus marking SupportedTensorDtypes::SAME_AS_COMPUTE deprecated.

### DIFF
--- a/kernels/portable/cpu/op_convolution.cpp
+++ b/kernels/portable/cpu/op_convolution.cpp
@@ -414,7 +414,7 @@ Tensor& convolution_out(
 
   ET_SWITCH_REALH_TYPES(in.scalar_type(), ctx, name, CTYPE, [&]() {
     const auto load_bias = bias.has_value()
-        ? utils::internal::get_load_to_common_fn<CTYPE, name>(
+        ? utils::internal::get_load_to_compute_fn<CTYPE, name>(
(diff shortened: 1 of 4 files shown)


swolchok:
* (to be filled)

swolchok:
I'm not sure of the best way to handle SAME_AS_COMMON/SAME_AS_COMPUTE. The problems are as follows:
- `SAME_AS_COMPUTE` is likely not what anybody meant; the output type is a function of the input types not a function of whatever precision we happen to compute in. (hence this diff deprecating it)
- We labelled the incoming type `CTYPE_COMMON`, but it's really the compute dtype (hence this diff changing the label).
- We don't actually know the common dtype. We can't add it to the elementwise_util interface without a breaking change because elementwise_util isn't marked `internal`. I'm not sure how to solve this part (but I'm sending a PR right now to deprecate the non-internal elementwise_util interface on grounds that it's demonstrably not stable yet).

swolchok:
we should land this if and only if @manuelcandales agrees it is improving the code, since the diffs I was trying to support with it have been abandoned.

manuelcandales (approved):

PR #9621, opened 2025-03-25
Deprecate non-internal elementwise_util APIs

swolchok:
As the code comment says, these APIs are undergoing development (see e.g. https://github.com/pytorch/executorch/pull/9613) and it's pretty inconvenient that they're incidentally committed-to externally. Mark them deprecated so we have the option to drop that commitment in (IIUC) 0.7.


swolchok:
* (to be filled)

manuelcandales (approved):

PR #9633, opened 2025-03-25
Strip size_test binaries and report their sizes in build_size_test.sh

swolchok:


swolchok:
* (to be filled)

swolchok:
this doesn't support the workflow I want after all, abandoning.

PR #9635, opened 2025-03-25
RFC: Compute only in int32/long/float/double for portable ops to save size

swolchok:
Concern: what if we're running on some sort of 16-bit microcontroller where this is a pessimization?

### DIFF
--- a/kernels/portable/cpu/util/elementwise_util.h
+++ b/kernels/portable/cpu/util/elementwise_util.h
@@ -199,11 +199,16 @@ inline void apply_tritensor_elementwise_fn(
 }
 
 inline ScalarType get_compute_type(ScalarType& common_type) {
-  ScalarType compute_type = common_type;
-  if (common_type == ScalarType::Half || common_type == ScalarType::BFloat16) {
-    compute_type = ScalarType::Float;
+  if (common_type == ScalarType::Long) {
+    return common_type;
   }
-  return compute_type;
+  if (isIntegralType(common_type, /*includeBool=*/true)) {
+    return ScalarType::Int;
+  }
+  if (common_type == ScalarType::Half || common_type == ScalarType::BFloat16) {
+    return ScalarType::Float;
+  }
+  return common_type;
 }
 } // namespace internal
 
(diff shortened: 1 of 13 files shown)


swolchok:
* (to be filled)

swolchok:
> some sort of 16-bit microcontroller

wasn't sure if I was making things up, so: https://developer.arm.com/Processors/Ethos-U55 is a real example from the present day

swolchok:
size impact: on my mac, test/build_size_test.sh reports that size_test_all_ops has size 1205136 before this PR and 1105856 after, a decrease of around 8%

swolchok:
This is known to break tests, I think because it breaks SupportedTensorDtypes::SAME_AS_COMMON for reasons outlined in #9613, hence it is RFC status.

swolchok:
per discussion with @manuelcandales, if we do this then we we need to cast through the "actual" compute type before casting to the output type so that we match ATen. example:

```
torch.ops.aten.mul(torch.tensor([100], dtype=torch.int8), torch.tensor([100], dtype=torch.int8), out=torch.zeros([1], dtype=torch.long))
tensor([16])
```

computing in int32 or int16 would cause this to yield 10000, not 16.

swolchok:
This is a bad idea because smaller compute dtypes benefit from additional SIMD lanes.

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -199,11 +199,29 @@ inline void apply_tritensor_elementwise_fn(
...
-  ScalarType compute_type = common_type;
+  // Code size optimization: on typical 32-bit or 64-bit CPUs, the ALU should be
+  // just as good at 32-bit arithmetic as it is at 16-bit or 8-bit
+  // arithmetic, so don't go out of our way to generate 8-bit or
+  // 16-bit code.
+
+  // Gate above optimization off if we appear to be on some kind of 8-bit or
+  // 16-bit CPU, which would invalidate our assumption about 32-bit
+  // math being just as fast.
+  constexpr bool cpu_appears_to_be_at_least_32_bit = sizeof(void*) >= 4 && sizeof(int) >= 4;
```
digantdesai, line 210: 👍

PR #9660, opened 2025-03-26
Remove exception fall back on checkpoint loading

jackzhxng:
### Summary
Remove the crutch of initializing with 0 weights when the checkpoint loading goes wrong (e.g. in most cases when the checkpoint keys don't match the parameters of the model).

### Test plan
See if CI passes


examples/models/llama/model.py
```diff
@@ -244,33 +244,24 @@ def __init__(self, **kwargs):
...
-            )
-            # Need to provide concrete (empty) values for meta-initialized tensors for quantization.
+        # assign=True: load params/buffers by assignment instead of performing an in-place copy.
+        # Because we are using device="meta", tensors do not have memory associated with them
+        # and an in-place copy is a no-op. Use assign=True in load_state_dict for this scenario.
+
+        # Also, the checkpoint is loaded and dtype promoted to the transformer's dtype, which is
+        # by default initialized to fp32. This is fine because every other supported type
+        # losslessly converts to fp32, so we don't lose precision here.
+        if checkpoint:
```
mcr229, line 254: yea i like this. If the user gives a checkpoint that doesn't work, we should fail, and not hide it under the rug and give them uninitialized model

mcr229 (approved):

PR #9665, opened 2025-03-26
Fix build gating on optimized_portable_kernels

swolchok:
If we don't need optimized kernels, we don't need optimized portable kernels.


swolchok:
* (to be filled)

manuelcandales (approved):

swolchok:
replaced by #9838 . not sure what's going on with ghstack :\

PR #9707, opened 2025-03-27
Add a convenient constructor

larryliu0820:
Summary: As titled. Use default values for `topp` and `rng`.

Differential Revision: D71956172

### DIFF
--- a/extension/llm/runner/stats.h
+++ b/extension/llm/runner/stats.h
@@ -82,8 +82,6 @@ struct ET_EXPERIMENTAL Stats {
   long aggregate_sampling_timer_start_timestamp = 0;
 };
 
-static constexpr auto kTopp = 0.9f;
-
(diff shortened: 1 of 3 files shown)


jackzhxng (approved):

extension/llm/runner/stats.h
```diff
@@ -82,8 +82,6 @@ struct ET_EXPERIMENTAL Stats {
   long aggregate_sampling_timer_start_timestamp = 0;
 };
 
-static constexpr auto kTopp = 0.9f;
```
swolchok, line 85: isn't this bc-breaking because this header doesn't include sampler.h?

larryliu0820: This constant belong to `sampler.h` not `stats.h`. Anyone using this constant should already imported `sampler.h` so it's not breaking anything. They don't have to include `stats.h` right now if they just need to use this constant.

PR #9710, opened 2025-03-27
Arm backend: Add missing __init__.py to passes

oscarandersson8218:
backends/arm/_passes missed an init file which caused pyre to not find imports from _passes.


zingo:
This replaces PR https://github.com/pytorch/executorch/pull/8592 
Its the same patch but rebased on latest, the other PR got some Meta internal version that missmatch we we could not merged it but got OK to merge by @digantdesai in the #8592 PR.

zingo (approved):

PR #9713, opened 2025-03-27
Check for null method in module execute

GregoryComer:
Summary: This is an attempt to partially mitigate use after free crashes when a module is unloaded during inference. This is not a supported use pattern, but it doesn't to sanity check here to avoid a crash.

Differential Revision: D71968421

### DIFF
--- a/extension/module/module.cpp
+++ b/extension/module/module.cpp
@@ -236,6 +236,8 @@ runtime::Result<std::vector<runtime::EValue>> Module::execute(
   ET_CHECK_OK_OR_RETURN_ERROR(load_method(method_name));
   auto& method = methods_.at(method_name).method;
   auto& inputs = methods_.at(method_name).inputs;
+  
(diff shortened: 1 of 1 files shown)


extension/module/module.cpp
```diff
@@ -236,6 +236,8 @@ runtime::Result<std::vector<runtime::EValue>> Module::execute(
   ET_CHECK_OK_OR_RETURN_ERROR(load_method(method_name));
   auto& method = methods_.at(method_name).method;
   auto& inputs = methods_.at(method_name).inputs;
+  
+  ET_CHECK_OR_RETURN_ERROR(method != nullptr, InvalidArgument, "Method is null");
```
shoumikhin, line 240: Trying to wrap my head how it could be possible at all that a unique_ptr held in an unordered_map would ever become nullptr after being once initialized in the past?
It's guaranteed the pointer isn't null by the time this code runs due to load_method() calls, then even if the Module used after deallocation, no code (like destructors) ever nullifies its members, so if the method is found in the dict, it wouldn't be nullptr.

PR #9309, opened 2025-03-17
Add support for torch.pow in the Arm backend

martinlsm:
cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #9715, opened 2025-03-27
Arm backend: Add support to eq.Scalar

fumchin:
Implement eq.Scalar by converting to eq.Tensor using replace_scalar_with_tensor_pass and match_arg_ranks_pass.

* Convert eq.Tensor to eq.Scalar
* Expand test_eq to test both eq.Tensor and eq.Scalar

### DIFF
--- a/backends/arm/_passes/match_arg_ranks_pass.py
+++ b/backends/arm/_passes/match_arg_ranks_pass.py
@@ -47,6 +47,7 @@ class MatchArgRanksPass(ExportPass):
         exir_ops.edge.aten.div.Tensor,
         exir_ops.edge.aten.bitwise_right_shift.Tensor,
         exir_ops.edge.aten.bitwise_left_shift.Tensor,
+        exir_ops.edge.aten.eq.Tensor,
     ]
 
     def _match_op_rank(self, graph_module, node, arg, max_rank):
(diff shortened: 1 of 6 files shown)


zingo (approved):

zingo (approved):
Good work!

kirklandsign:
Hi @zingo seems that it broke some of our internal tests.

cc @zonglinpeng for more contexts

zonglinpeng:
Hi @zingo This diff breaks CI T219571006, introducing large loss in the model output. I verified the new eq scalar-to-tensor replacement caused the issue. Could you make make this replacement optional?

zingo:
Thanks for notifying us, just for my understanding is this test on Arm delegate or other delegates?

We have a follow up PR where we internally discussing if we should move this change to only run on Arm delegate (its our Pow PR) and Im wondering if that will solve the problems you see or if we should investigate more?

As we cant see the test you run we can only guess. Thanks a lot for your help and initial triage.

kirklandsign:
Hi @zingo, the broken test is on other delegates. Seems that this has impacts on all delegates? In this case, if we apply it to arm only, it would be ideal.

zingo:
Thanks for the response, that good news and a fix is almost already prepared by @martinlsm that we can start with and figure out the Arm delegate later.

zonglinpeng:
thanks @zingo I will try if PR https://github.com/pytorch/executorch/pull/9309 fixes on my delegate, thank you @kirklandsign for clarifying!

kirklandsign:
@zingo @zonglinpeng I tried applying this patch but seems that the test was not fixed

martinlsm:
@kirklandsign Which delegate are you using? Is it possible for us at Arm to reproduce?

kirklandsign:
Hi @martinlsm the test is cadence backend. I am not sure how to repro, could @zonglinpeng provide context?

BTW seems that #9309 is not a clean revert of #9715 for that file. Is that related?

zonglinpeng:
Thanks for checking in. I’m still running but some tests seems green on ci
so should be good

PR #9723, opened 2025-03-27
Update install_android.sh

kirklandsign:


kirklandsign:
Won't write to docker, but each time use sccache

PR #9729, opened 2025-03-27
[Bug Fix] Fix padding when running in NHWC

mcr229:
### Summary
There is a bug when there is a constant_pad between two convolutions. In order to minimize permutes associated with memory format changes, we sometimes compute ops in NHWC. This is the case for ConstantPad when it is between two convs:

```
    a = conv(a)
    a = constant_pad(a, paddings=[1, 2, 3, 4])
    a = conv(a)
```

in this case we need to make sure the paddings given to constant_pad are also permuted to nhwc. 

### Test plan
python install_executorch.py --editable

python -m unittest backends.xnnpack.test.ops.test_static_constant_pad.TestStaticConstantPad.test_fp32_static_constant_pad_nhwc

### DIFF
--- a/backends/xnnpack/operators/op_static_constant_pad.py
+++ b/backends/xnnpack/operators/op_static_constant_pad.py
@@ -7,6 +7,7 @@
 from typing import cast, Dict, List
 
 import torch
+
 from executorch.backends.xnnpack.operators.node_visitor import (
     get_tensor_value,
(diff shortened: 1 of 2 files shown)


GregoryComer (approved):
Nice catch.

guangy10 (approved):

PR #9731, opened 2025-03-28
Fix Fake bench data type

yangw-dev:
Fix bug to upload error benchmark report: benchmakr_values:[0]
Unable to see the failure report in clickhouse, wonder if bc this


clee2000 (approved):

huydhn (approved):
LGTM!  I have made the same mistake before using a value instead of a list for `benchmark_values`.  Let's add a check in the upload script to throw an error so that this will fail on CI instead.

PR #9740, opened 2025-03-28
Add build_optimized_size_test.sh

swolchok:
Adding a developer-facing size test script for a non-embedded-centric use case (optimized ops, exceptions on, etc.).


swolchok:
* (to be filled)

PR #9741, opened 2025-03-28
Migrate elementwise_util callers to the variants with out_dtypes in template arguments

swolchok:
This is necessary to take advantage of the previous PR #9388, which creates dtype-specialized implementations for the non-mixed dtype case.

The size cost of this approach looking at `__TEXT` of size_test_all_optimized_ops built with bash test/build_optimized_size_test.sh is 104212 bytes, a 2.2% increase.


swolchok:
* (to be filled)

manuelcandales (approved):

swolchok:
looks like a dupe PR? we landed and/or reapplied a version of this

PR #9742, opened 2025-03-28
Save some size in dtype_util when dtype selective build is not in use

swolchok:
We duplicate a lot of functions depending on the operator name so that dtype selective build will work. We can just detect if dtype selective build is in use and, if not, stop duplicating.

Test Plan: Saves 28288 bytes of text in size_test_all_optimized_ops compared to previous PR on my Mac.


swolchok:
* (to be filled)

manuelcandales (approved):

PR #9752, opened 2025-03-28
etdumpfilter impl

Gasoonjia:
Differential Revision: D72025517

### DIFF
--- a/devtools/etdump/etdump_filter.cpp
+++ b/devtools/etdump/etdump_filter.cpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/devtools/etdump/etdump_filter.h>
+
+#include <executorch/runtime/core/error.h>
+
+using ::executorch::runtime::DelegateDebugIntId;
+using ::executorch::runtime::Error;
+using ::executorch::runtime::kUnsetDelegateDebugIntId;
+
+namespace executorch {
+namespace etdump {
+
+ETDumpFilter::ETDumpFilter() = default;
+
+Result<bool> ETDumpFilter::add_regex(string_view pattern) {
+  auto regex = std::make_unique<re2::RE2>(pattern.data());
+  if (!regex->ok()) {
+    return Error::InvalidArgument; // Error during regex compilation
+  }
+  regex_patterns_.emplace_back(std::move(regex));
+  return true;
+}
+
+Result<bool> ETDumpFilter::set_debug_handle_range(size_t start, size_t end) {
+  if (start >= end) {
+    return Error::InvalidArgument; // Start is greater than end
+  }
+  if (start < 0 || end < 0) {
+    return Error::InvalidArgument; // Start or end is negative
+  }
+  range_start_ = start;
+  range_end_ = end;
+  return true;
+}
+
+Result<bool> ETDumpFilter::filter_name_(const char* name) {
+  if (name == nullptr) {
+    return Error::InvalidArgument;
+  }
+  if (regex_patterns_.empty()) {
+    return true;
+  }
+  for (const auto& regex : regex_patterns_) {
+    if (RE2::FullMatch(name, *regex)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+Result<bool> ETDumpFilter::filter_delegate_debug_index_(
+    DelegateDebugIntId debug_handle) {
+  if (debug_handle == kUnsetDelegateDebugIntId) {
+    return Error::InvalidArgument; // Delegate debug index is unset
+  }
+
+  if (range_start_ == 0 && range_end_ == 0) {
+    return true;
+  }
+
+  if (debug_handle < range_start_ || debug_handle >= range_end_) {
+    return false;
+  }
+
+  return true;
+}
+
+Result<bool> ETDumpFilter::filter(
+    const char* name,
+    DelegateDebugIntId delegate_debug_index) {
+  if ((name == nullptr) == (delegate_debug_index == kUnsetDelegateDebugIntId)) {
+    return Error::InvalidArgument; // Name and delegate debug index should be
+                                   // both set or unset
+  }
+
+  if (name) {
+    return filter_name_(name);
+  } else {
+    return filter_delegate_debug_index_(delegate_debug_index);
+  }
+}
+
+size_t ETDumpFilter::get_n_regex() const {
+  return regex_patterns_.size();
+}
+
+} // namespace etdump
+} // namespace executorch
--- a/devtools/etdump/etdump_filter.h
+++ b/devtools/etdump/etdump_filter.h
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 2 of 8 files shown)


devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <executorch/runtime/core/error.h>
+
+
+using ::executorch::runtime::DelegateDebugIntId;
+using ::executorch::runtime::Error;
+using ::executorch::runtime::kUnsetDelegateDebugIntId;
+
+namespace executorch {
+namespace etdump {
```
swolchok, line 19: we have C++17 -- `namespace executorch::etdump {`

devtools/etdump/etdump_filter.h
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Returns the number of regex patterns in the filter.
+   */
+  size_t get_n_regex() const;
+
+ private:
+  char regex_patterns_[MAX_REGEX_PATTERNS][MAX_PATTERN_LENGTH]{};
+  size_t regex_count_;
+  size_t range_start_;
+  size_t range_end_;
```
swolchok, line 100: nit: you don't need the explicit default constructor if you assign 0 to these here; you can then change the constructor implementation in the .cpp file to `ETDumpFilter::ETDumpFilter() = default;`

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+Result<bool> ETDumpFilter::add_regex(const char* pattern) {
+  if (regex_count_ >= MAX_REGEX_PATTERNS) {
+    return Error::OutOfResources; // Error code for exceeding max patterns
+  }
+  size_t len = strlen(pattern);
+  if (len >= MAX_PATTERN_LENGTH) {
+    return Error::InvalidArgument; // Pattern too long
+  }
+  strcpy(regex_patterns_[regex_count_], pattern);
```
swolchok, line 32: nit: you already have the length, so use it -- `std::memcpy(regex_patterns_[regex_count_], pattern, len)`

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,101 @@
...
+  regex_patterns_[regex_count_][len] = '\0';
```
swolchok, line 33: FYI strcpy already does this. however, if you take my memcpy advice above you'll need to leave it

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,101 @@
...
+  regex_count_++;
+  return true;
+}
+
+Result<bool> ETDumpFilter::set_debug_handle_range(size_t start, size_t end) {
```
swolchok, line 38: what does a successful result of `false` mean? (does `Result<void>` work?)

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,101 @@
...
+    return Error::InvalidArgument; // Start or end is negative
+  }
+  range_start_ = start;
+  range_end_ = end;
+  return true;
+}
+
+Result<bool> ETDumpFilter::filter_name_(const char* name) {
+  if (name == nullptr) {
+    return Error::InvalidArgument; // Name is null
```
swolchok, line 52: comment repeats the code, delete

devtools/etdump/etdump_filter.h
```diff
@@ -0,0 +1,107 @@
...
+      const char* name,
+      ::executorch::runtime::DelegateDebugIntId delegate_debug_index) override;
+
+  /**
+   * Returns the number of regex patterns in the filter.
+   */
+  size_t get_n_regex() const;
+
+ private:
+  char regex_patterns_[MAX_REGEX_PATTERNS][MAX_PATTERN_LENGTH]{};
```
swolchok, line 97: why the hardcoded limit on length instead of a `std::vector<std::string>`? I don't see any particular guarantees from RE2 itself that it doesn't malloc, so isn't studiously avoiding malloc in our wrapper a lost cause?

devtools/etdump/etdump_filter.h
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @return A Result<bool> indicating the success or failure of adding the
+   * regex pattern.
+   *         - True if the pattern is successfully added.
+   *         - False if the pattern could not be added or if the maximum number
+   * of patterns is exceeded.
+   *         - An error code if number of pattern has reached to cap, or any
+   * error occurs during regex compilation.
+   */
+  Result<bool> add_regex(const char* pattern);
```
swolchok, line 51: std::string_view is in C++17. why not use it instead? (if the user ever passes a literal, we won't need to check its length at runtime)

swolchok (commented):
just some C++ comments; leaving for reviewers with more context on the actual goals here

devtools/etdump/etdump_filter.cpp
Gasoonjia: false will not be use. 
We can not use `Result<void>` cuz Result class need to use reference for the T type and void does not have Reference.

devtools/etdump/etdump_filter.h
Gasoonjia: I don't want to use any memory allocation operation in etdump_filter; our user usually have their own memory allocation way and does not support anything like vector or malloc.

swolchok: > does not support anything like vector or malloc

but that's my point -- RE2 mallocs! (for example https://github.com/google/re2/blob/main/re2/compile.cc#L1254)

devtools/etdump/etdump_filter.h
Gasoonjia: thx didn't know we have that; migrating to that.

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,100 @@
...
+using ::executorch::runtime::Error;
+using ::executorch::runtime::kUnsetDelegateDebugIntId;
+
+namespace executorch {
+namespace etdump {
+
+ETDumpFilter::ETDumpFilter()
+    : regex_count_(0), range_start_(0), range_end_(0) {}
+
+Result<bool> ETDumpFilter::add_regex(const char* pattern) {
```
tarun292, line 23: I think let's call this add_filter, we don't necessarily want to tie this to regex i think. On smaller platforms where they pass their own filter implementation they can just do simple string matching.

devtools/etdump/etdump_filter.cpp
```diff
@@ -0,0 +1,100 @@
+/*
```
tarun292, line 1: Overall high level comment, i think it's reasonable that we can assume the etdump filter we provide based on regex will need malloc to be supported. If users want to use this malloc support will be required on their platform. Based on this i think it would be good to remove the limitations we have in this class such as max regex, max char len etc.

tarun292: If users do need a filter for their platform that isn't based on regex they can write their own one. (which will be only very few embedded users).

devtools/etdump/etdump_filter.h
Gasoonjia: ahhh fair point didn't pay attention to RE internal impl

devtools/etdump/etdump_filter.cpp
Gasoonjia: > which will be only very few embedded users
ok  in this case i agree we can make our etdump_filter support regex + malloc, just as what [swolchok](https://github.com/swolchok) mentioned RE use malloc internally. Let me make an update

tarun292 (approved):

PR #9760, opened 2025-03-30
Fix breaks from googletest upgrade

hershi:
Summary: Add missing include for standard C++ library

Reviewed By: 8Keep


hershi:
Had to recreate as https://github.com/pytorch/executorch/pull/9868

PR #9761, opened 2025-03-31
Error handling log intermediate output delegate

### LINKED ISSUE #9549: log_intermediate_output_delegate cannot handle error as expected
This PR https://github.com/pytorch/executorch/pull/9493/files updates the return type of `log_intermediate_output_delegate` function from void to bool, which means:

```
   * @return A Result<bool> indicating the status of the logging operation.
   *         - True if the bool type output was successfully logged.
   *         - False if the bool type output was filtered out and not logged.
   *         - An error code if an error occurs during logging.
``` 

However currently all `log_intermediate_output_delegate` didn't handle the error correctly; the system will be crash on `ET_CHECK_MSG`, instead of returning the error code as we expected: `https://github.com/pytorch/executorch/blob/main/devtools/etdump/etdump_flatcc.cpp#L355

We would love to update the error raising mechanism, as well as the related tests.

Here's an example to return error code instead of crashing system: https://github.com/pytorch/executorch/blob/main/devtools/etdump/data_sinks/buffer_data_sink.cpp#L24

Please kindly leave your comments under the issue, or reach out to us via discord https://github.com/pytorch/executorch?tab=readme-ov-file#contributing, if you have any quesion.

dillondesilva:
### Summary

This PR contains a fix for all `log_intermediate_output_delegate` methods which can now propagate any error caught inside of `log_intermediate_output_delegate_helper`. Fixes https://github.com/pytorch/executorch/issues/9549

release notes: `log_intermediate_output_delegate` no longer aborts runtime if parameters delegate debug index or name are invalid. Instead, an error is returned.

### Test plan

Tests for this PR can be verified by building `sdk_etdump_tests` and using `ctest`


dillondesilva:
Just added another commit to pass lintrunner!

dillondesilva:
@Gasoonjia I'm not too sure why `
pull / test-static-llama-qnn-linux / linux-job (pull_request)` is failing - any thoughts?

Gasoonjia:
i don't think your update crashing the system. Should be infra issue. Let me rebase and rerun the tests.

dillondesilva:
Thanks so much!

Gasoonjia (approved):

Gasoonjia:
Merged! Thanks for your help! @dillondesilva

PR #9762, opened 2025-03-31
Add try-before-set tests for DataSink in DebugEvent and LogDelegateIntermediateOutput

### LINKED ISSUE #9551: Missing try-before-set tests for datasink in etdumpgen tests
### 🚀 The feature, motivation, and pitch

try-before-set tests are important for etdumpgen which avoids unexpected phenomenon when user using debug_buffer or datasink before setting it. 

However right now we only have it in debug_buffer, but not in datasink https://github.com/pytorch/executorch/blob/31d35455801c25f80eea711d309e43818722ea35/devtools/etdump/tests/etdump_test.cpp#L514

The reason is, debug_buffer and datasinks are using a same etdumpgen in every inner for loop; after the first round even though we haven't set the datasink, it already has debug buffer that can handle the debug data dumping, so that error will not be raised.

### possible solution

One solution is for every for-loop round, instead of sharing the same etdumpgen, we can use a different one, to bypass the issue.

We are open to any other solutions!

Please kindly leave your comments under the issue, or reach out to us via discord https://github.com/pytorch/executorch?tab=readme-ov-file#contributing, if you have any quesion.

keyprocedure:
### Summary
Fixes #9551 

Reset ETDumpGen instances in each test iteration to avoid shared state to correctly trigger ET_EXPECT_DEATH when no DataSink is set.

### Test plan
buck2 run devtools/etdump/tests:etdump_test --

### DIFF
--- a/devtools/etdump/tests/etdump_test.cpp
+++ b/devtools/etdump/tests/etdump_test.cpp
@@ -186,11 +186,16 @@ TEST_F(ProfilerETDumpTest, AllocationEvents) {
 }
 
 TEST_F(ProfilerETDumpTest, DebugEvent) {
+  const size_t debug_buf_size = 2048;
+  const size_t etdump_buf_size = 512 * 1024;
+  ASSERT_NE(this->buf, nullptr);
+  Span<uint8_t> span_buf = Span<uint8_t>(this->buf, etdump_buf_size);
+
   for (size_t i = 0; i < 2; i++) {
     for (size_t j = 0; j < 3; j++) {
       etdump_gen[i]->create_event_block("test_block");
 
-      void* ptr = malloc(2048);
+      void* ptr = malloc(debug_buf_size);
 
       EValue evalue_int((int64_t)5);
       etdump_gen[i]->log_evalue(evalue_int);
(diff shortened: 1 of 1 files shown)


keyprocedure:
Hi @Gasoonjia, would you mind taking a look at my changes to confirm they match what you had in mind?

I noticed that the DebugEvent test (starting at line 188) follows the same pattern as LogDelegateIntermediateOutput before I added the additional ET_EXPECT_DEATH checks. Do you think similar checks should be added there as well?

Gasoonjia:
hi @keyprocedure thanks for your great work! 
Yes please update all of them:
https://github.com/pytorch/executorch/blob/main/devtools/etdump/tests/etdump_test.cpp#L214
https://github.com/pytorch/executorch/blob/main/devtools/etdump/tests/etdump_test.cpp#L513
Thanks!

devtools/etdump/tests/etdump_test.cpp
```diff
@@ -497,36 +515,57 @@ TEST_F(ProfilerETDumpTest, VerifyData) {
...
-      void* ptr = malloc(2048);
-      Span<uint8_t> buffer((uint8_t*)ptr, 2048);
+      void* ptr = malloc(debug_buf_size);
+      Span<uint8_t> buffer((uint8_t*)ptr, debug_buf_size);
 
-      auto buffer_data_sink = BufferDataSink::create(ptr, 2048);
+      auto buffer_data_sink = BufferDataSink::create(ptr, debug_buf_size);
       auto file_data_sink = FileDataSink::create(dump_file_path.c_str());
 
       etdump_gen[i]->create_event_block("test_block");
```
Gasoonjia, line 544: perhapes we can move this line under if (j == 0) to make more structural.

devtools/etdump/tests/etdump_test.cpp
```diff
@@ -497,36 +515,57 @@ TEST_F(ProfilerETDumpTest, VerifyData) {
   }
 }
 
+// Triggers ET_EXPECT_DEATH if log_intermediate_output_delegate has no data sink
+static void expect_log_intermediate_delegate_death(
```
Gasoonjia, line 519: Love this helper function! Please make it as a protected member function of `ProfilerETDumpTest ` to prohibit misuse. See https://github.com/pytorch/executorch/blob/main/kernels/test/op_split_copy_test.cpp#L94 for an example.

devtools/etdump/tests/etdump_test.cpp
Gasoonjia: same as above!

Gasoonjia (commented):
Overall LGTM! Thanks @keyprocedure for your contribution! Will stamp it after ci pass.

devtools/etdump/tests/etdump_test.cpp
keyprocedure: Good catch! I moved line 509 into the j==0 block.

Does the second comment, "same as above", refer to line 209? Can that line be moved into the j==0 block? Isn't it needed outside of the block?

devtools/etdump/tests/etdump_test.cpp
keyprocedure: Moved the helper function, I appreciate the example!

Gasoonjia (approved):

Gasoonjia:
Thanks for your great work! Merged!

PR #9764, opened 2025-03-31
Add cadence.where_Scalar op

mcremon-meta:
Summary: The op replaces the regular where op when it uses two `aten.full` ops for the tensors. Those cases do no need broadcast (but would call it in the current state) and can sometimes be constant folded, if the `condition` is a constant tensor. Since `aten.full` is _not_ currently constant folded, it would stay in the graph.

Reviewed By: skrtskrtfb

Differential Revision: D70539497


zonglinpeng (approved):
lgtm

PR #9751, opened 2025-03-28
Improve ET_CHECK_OR_RETURN_FALSE error messages in kernels/portable

swolchok:
We have a lot of `ET_CHECK_OR_RETURN_FALSE` that log a condition, but
not the values of the variables in that condition. This is an attempt
to improve debuggability of these errors.

cc @larryliu0820 @manuelcandales

PR #9768, opened 2025-03-31
Add ET_PRI_* macros for Tensor methods that differ with ATen

swolchok:
#9751 added a bunch of logging to error messages. We discovered after commit that it caused -Wformat warnings (turned into errors via -Werror) in ATen mode, because a bunch of Tensor methods return different types between ExecuTorch and ATen. This PR adds formatting macros to be used with this methods and uses them.


swolchok:
* (to be filled)

kirklandsign (approved):

PR #9769, opened 2025-03-31
Move replacement of conv to linear/fc to opt level 2

mcremon-meta:
Summary: As titled. Now that we have optimized convolutions on HiFi, we can move them up in opt level.

Differential Revision: D72176194


zonglinpeng (approved):
We should make those numbers macros and comment on their meanings

skrtskrtfb (approved):
Shouldn't have any regressions on model performance but could be good to check JIC these passes are used often. LGTM!

PR #9773, opened 2025-03-31
Turn on XNNPACK by default for wheels

jathu:
### Summary

We pretty much always include xnnpack when building wheels, so why not just turn it on by default? If users want it off, they can opt-out.

image

### Test plan

CI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,7 @@ class ShouldBuild:
 
     @classmethod
     def xnnpack(cls) -> bool:
-        return cls._is_cmake_arg_enabled("EXECUTORCH_BUILD_XNNPACK", default=False)
+        return cls._is_cmake_arg_enabled("EXECUTORCH_BUILD_XNNPACK", default=True)
 
     @classmethod
     def training(cls) -> bool:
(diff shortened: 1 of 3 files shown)


larryliu0820 (approved):

setup.py
```diff
@@ -730,18 +730,17 @@ def run(self):
                 "-DEXECUTORCH_BUILD_KERNELS_QUANTIZED_AOT=ON",
             ]
 
+            if ShouldBuild.xnnpack():
+                cmake_args += ["-DEXECUTORCH_BUILD_XNNPACK=ON"]
+                build_args += ["--target", "portable_lib"]
```
mergennachin, line 735: if user explicitly turns of pybind (via `pybind --off`), don't we still want portable_lib?

jathu: We currently only add portable_lib if pybindings is turned on: https://github.com/pytorch/executorch/blame/97bca0547a224e2df15f823b1d7966e8b7546b70/setup.py#L726-L740

I guess I assumed this was just for XNNPACK, but is portable_lib for *any* pybindings?

jathu: Based on the build steps, it seems like it's needed for all pybindings: https://github.com/pytorch/executorch/blob/97bca0547a224e2df15f823b1d7966e8b7546b70/CMakeLists.txt#L858

I'm assuming CI isn't catching this because we never use a single non-xnnpack binding.

mergennachin (approved):

PR #9787, opened 2025-03-31
[TEST ONLY]

kirklandsign:


kirklandsign:
Hi @huydhn I am just wondering, seems that the secret is empty string here. Do you have any ideas?

.github/workflows/android-release-artifacts.yml
```diff
@@ -52,6 +55,16 @@ jobs:
         PYTHON_EXECUTABLE=python bash .ci/scripts/setup-linux.sh --build-tool buck2
         export ARTIFACTS_DIR_NAME=artifacts-to-be-uploaded
 
+        mkdir -p ~/.gradle
```
huydhn, line 58: Just curious if these secrets can be used after building the aar archive, or do they need to be set in `gradle.properties` before the build?

kirklandsign: After the build, before publish in here: https://github.com/pytorch/executorch/pull/9787/files#diff-bb64d4a0ad63db311f02484469ec3e9c1fc264c8fe6c3e190ec962348af963b3R129

This PR is WIP, and will fix afterwards.

PR #9795, opened 2025-04-01
Arm backend: Change _is_ok_for_quantization to support output check

Sebastian-Larsson:
_is_ok_for_quantization now checks the node itself as well to verify that the node can be quantized. Previously it was only checked by looking at the inputs to the node.

This led to TestSplit failing, which is fixed with the change to `is_non_float_tensor` in `arm_quantizer_utils`, which now handles when node.meta["val"] is a `list` of `FakeTensor`. It traverses the list and checks if any of the elements are **not** a `FakeTensor`. If one element is not a `FakeTensor` the function will return `True`.

### DIFF
--- a/backends/arm/quantizer/arm_quantizer_utils.py
+++ b/backends/arm/quantizer/arm_quantizer_utils.py
@@ -11,7 +11,7 @@
 # Utility functions for TOSAQuantizer
 #
 
-from typing import cast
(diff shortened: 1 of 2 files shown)


Sebastian-Larsson:
From what I can tell the CI failure is unrelated.

zingo (approved):

PR #9818, opened 2025-04-01
A short term fix for editable mode install failed to import root level module such as exir

### LINKED ISSUE #9558: `from executorch.exir import CaptureConfig` fails with ExecuTorch pip installed in editable mode
### 🐛 Describe the bug

`python -c 'from executorch.exir import CaptureConfig'` works with ExecuTorch installed via ./install_executorch.sh, but fails with ExecuTorch installed by `pip install -e .`:

```
  File "<string>", line 1, in <module>
    from executorch.exir import CaptureConfig
ImportError: cannot import name 'CaptureConfig' from 'executorch.exir' (unknown location)
```

### Versions

main

larryliu0820:
Summary:

Fixes #9558.

The `src/executorch/exir` file exists primarily due to the limitations of `pip install` in editable mode. Specifically, `pip install -e .` does not recognize `<executorch root>/exir` (or any root level directory with a `__init__.py`) as a valid package module because of the presence of `<executorch root>/exir/__init__.py`. See the following GitHub issue for details: [Issue #9558](https://github.com/pytorch/executorch/issues/9558).

To work around this limitation, a symlink is used. With this symlink and this package entry in `pyproject.toml`:

```toml
[tool.setuptools.package-dir]
# ...
"executorch" = "src/executorch"
```
We are telling `pip install -e .` to treat `src/executorch` as the root of the `executorch` package and hence mapping `executorch.exir` to `src/executorch/exir`. This effectively gets `exir` out from the root level package.

This allows us to perform `pip install -e .` successfully and enables the execution of the following command:

```bash
python -c "from executorch.exir import CaptureConfig"
```

Test Plan:

```bash
./install_executorch.sh --pybind --editable
python -c "from executorch.exir import CaptureConfig"
```

Reviewers:

Subscribers:

Tasks:

Tags:


mergennachin:
Why only exir? 

What if we have a symlink from `root/src/executorch` to `root/` as opposed to only focusing on `root/src/executorch/exir`?

larryliu0820:
> Why only exir?
> 
> What if we have a symlink from `root/src/executorch` to `root/` as opposed to only focusing on `root/src/executorch/exir`?

Gee, I think that's better. I don't think we should symlink the whole root directory, instead we probably want to symlink all directories with python file.

swolchok (approved):
Wow, is exir really the only top-level directory where this is a problem? I'm surprised, but it's not impossible, so here's an accept.

larryliu0820:
> Wow, is exir really the only top-level directory where this is a problem? I'm surprised, but it's not impossible, so here's an accept.

Actually tested 
```
python -c "from executorch.examples.models import MODEL_NAME_TO_MODEL"
```

Same problem so in this update I symlink'd all the modules

swolchok:
@mergennachin you probably want to take a look at this

PR #9823, opened 2025-04-01
[Executorch][to_backend] Introduce preprocess_multimethod

mcr229:
### summary

We introduce a new preprocess_multimethod. This takes the place of preprocess all in the following discussion: https://github.com/pytorch/executorch/discussions/8230

This is a class method so that we can allow the default implementation to leverage the static preprocess method. The default implementation just applies preprocess on all the exported programs given.

### Testing
Testing done in follow up PR 
https://github.com/pytorch/executorch/pull/9824


mcr229:
* (to be filled)

exir/backend/backend_details.py
```diff
@@ -70,6 +61,69 @@ def preprocess(
...
+        """
         # Users should return a compiled blob - a binary that can run the desired
         # program in the backend.
         pass
+
+    @classmethod
+    def preprocess_multimethod(
+        cls,
+        edge_programs: Dict[str, List[ExportedProgram]],
+        compile_specs: Dict[str, List[List[CompileSpec]]],
```
cccclai, line 87: This is the finest granularity which I believe cover all the use cases. For majority of the use case, we probably can support `List[CompileSpec]`, which will be applied to all edge programs for all methods.

cccclai (approved):
Looks good to me. Let's add the unit test. We can have two backends, one with default `preprocess_multimethod` and one with custom `preprocess_multimethod`, then calling  `MyBackend.preprocess_multimethod()` will invoke `preprocess_multimethod`

mcr229:
@cccclai yea let me add a test for the default preprocess_multimethod case

PR #9824, opened 2025-04-01
[ExecuTorch][to_backend]  Enable to_backend API  to leverage preprocess_multimethod

mcr229:
### Summary

We add a new to_backend api which for multi method models. Specifically, we pass in a dictionary mapping MethodName to ExportedProgram, as well as a dictionary mapping MethodName to Partitioner. We then return a dictionary mapping MethodName to the partitioned and lowered exported program.

In addition, we also provide a new preprocess API for backends to implement. This API is preprocess_multimethod. The signature of the new method is as follows:
```
    def preprocess_multimethod(
        cls,
        edge_programs: Dict[str, List[ExportedProgram]],
        compile_specs: Dict[str, List[List[CompileSpec]]],
    ) -> Dict[str, list[PreprocessResult]]:
        """
        Runs preprocess on all partitioned Edge Programs across multiple methods. This allows
        backends to share information across partitioned graphs. Backend can serialize shared
        data by putting the shared data into the data_store_output of the preprocess results.
        This will record the shared data used by that specific partition.

        Default implementation is running the existing preprocess implementation on all

        Args:
            edge_programs: Dictionary mapping the method name to a list of all the partitioned
                           edge_programs from that method to be lowered.
            compile_specs: Dictionary mapping the method name to a list of compile_specs. The
                           list of compile specs maps directly to the list of edge_programs for the
                           same given method name i.e. edge_program[method_name][i] --> compile_specs[method_name][i]

        Returns:
            Dictionary mapping the method name to a list of PreprocessResults. The list of
            PreprocessResults maps directly to the list of edge_programs for the same given
            method name. i.e. edge_program[method_name][i] --> result[method_name][i]
        """
```

This new API enableds backends to preprocess all partitions/methods at once. This way, while processing blobs, they can identify shared components between preprocessed blobs. Shared components can be serialized within the NamedDataStore.

The key change in backend infra, is that when partitioning, we now have to identify all the partitioned graphs to be lowered at once, and pass them to preprocess_multimethod at once. Previously, as we found lowerable partitions, we preprocessed and embedded them into the graph.

### Testing
python -m unittest exir.backend.test.test_to_backend_multi_method


mcr229:
* (to be filled)

exir/backend/test/test_to_backend_multi_method.py
```diff
@@ -0,0 +1,406 @@
...
+        class AddModule(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+
+            def forward(self, x):
+                return x + x
+
+        sin_edgeir_m = to_edge(torch.export.export(SinModule(), (torch.ones(1),)))
+        add_edgeir_m = to_edge(torch.export.export(AddModule(), (torch.ones(1),)))
+        sin_partitioner = AllNodePartitioner(
```
cccclai, line 148: Why do we need `AllNodePartitioner`?

exir/backend/test/test_to_backend_multi_method.py
```diff
@@ -0,0 +1,406 @@
...
+from executorch.exir.lowered_backend_module import (
+    get_lowered_submodules,
+    LoweredBackendModule,
+)
+from torch.export.exported_program import ExportedProgram
+
+from torch.testing import FileCheck
+
+
+class TestToBackendMultiMethod(unittest.TestCase):
```
cccclai, line 36: It looks like the test cases here are mostly AoT part, can we use ExecuTorchBackend to do end to end test?

exir/backend/backend_api.py
```diff
@@ -417,3 +437,330 @@ def to_backend(
...
+    # check the number of submodule_names and submodule_nodes are equal
+    for backend_id in created_submodule_nodes.keys():
+        assert len(created_submodule_nodes[backend_id]) == len(
+            backend_id_to_submodule_name[backend_id]
+        )
+
+    return created_submodule_nodes
+
+
+def _create_partitions(
```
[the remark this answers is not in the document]
mcr229: they're slightly different because this one doesn't actually create the call_delegate with the loweredbackendmodule. This just replaces all the partitions with call_submodule node. Since the goal is to lower all modules at once, we collect all the submodule nodes first and then we run the preprocess_multimethod after. 

There could be some opportunity for refactoring all the logic around to get things working, but I'm ok with leaving that for later as it would require a lot more testing to make sure I don't mess up the existing paths.

exir/backend/test/test_to_backend_multi_method.py
mcr229: this is just for testing.

exir/backend/test/test_to_backend_multi_method.py
```diff
@@ -0,0 +1,517 @@
...
+                add_partitioner,
+                [],
+            ),
+        }
+        with self.assertRaisesRegex(
+            NotImplementedError, "Backend Invalid was not found."
+        ):
+            self._test(test_set)
+
+    def test_multi_method_end_to_end(self):
```
mcr229, line 436: @cccclai this should test the things you were asking about.

We go through to_backend_multimethod path, we lower to BackendWithCompilerDemo which doesn't have preprocess_multimethod implemented (which means it should take the default implementation path, the class method from the previous PR). 

We check that only one delegated blob is created when we are using processed bytes are the same, and we check that it runs end-to-end

cccclai: Yeah this looks nice, thank you!

exir/backend/test/test_to_backend_multi_method.py
```diff
@@ -0,0 +1,517 @@
...
+                method_partitioner,
+            )
+        )
+
+        new_edge_manager = EdgeProgramManager(lowered_ep_dict)
+
+        exec_prog = new_edge_manager.to_executorch()
+
+        program = exec_prog.executorch_program
+        self.assertEqual(len(program.backend_delegate_data), 1)
```
cccclai, line 474: Maybe add a comment to explain why there is only 1 delegate blob

exir/backend/test/test_to_backend_multi_method.py
cccclai: Maybe add this comment to the test case, so other people know the expectation for this test.

mcr229:
```
 pytest devtools/inspector/tests/inspector_utils_test.py::TestInspectorUtils::test_compare_results_uint8 
============================================================================================================================================================ test session starts ============================================================================================================================================================
platform darwin -- Python 3.11.4, pytest-8.3.5, pluggy-1.5.0
rootdir: /Users/maxren/Desktop/oss/executorch
configfile: pytest.ini
plugins: cov-4.1.0, xdist-3.3.1, rerunfailures-15.0, hypothesis-6.84.2
collected 1 item                                                                                                                                                                                                                                                                                                                            

devtools/inspector/tests/inspector_utils_test.py .                                                                                                                                                                                                                                                                                    [100%]

============================================================================================================================================================ 1 passed in 15.11s =============================================================================================================================================================
```

for the failing test, let me rebase and resubmit

cccclai (approved):
Looks good. Thank you for being patient and all the effort!

PR #9830, opened 2025-04-02
DO NOT COMMIT test for u55 + mv2

digantdesai:
### Summary

### Test plan

### DIFF
--- a/backends/arm/arm_vela.py
+++ b/backends/arm/arm_vela.py
@@ -66,7 +66,7 @@ def vela_compile(
         output_dir = os.path.join(tmpdir, "output")
         args.append(f"--output-dir={output_dir}")
         args.append(tosa_path)
-        if verbose:
+        if verbose or True:
             args.append("--verbose-all")
         vela.main(" ".join(args).split(" "))
 
--- a/backends/arm/test/test_arm_baremetal.sh
+++ b/backends/arm/test/test_arm_baremetal.sh
@@ -120,23 +120,25 @@ test_models_ethosu_fvp() { # End to End model tests using model_test.py
     python3 backends/arm/test/test_model.py --test_output=arm_test/test_model --build_libs
 
     # TOSA quantized
-    echo "${TEST_SUITE_NAME}: Test ethos-u target TOSA"
-    python3 backends/arm/test/test_model.py --test_output=arm_test/test_model --target=TOSA --model=mv2
-    python3 backends/arm/test/test_model.py --test_output=arm_test/test_model --target=TOSA --model=mv3
-    python3 backends/arm/test/test_model.py --test_output=arm_test/test_model --target=TOSA --model=lstm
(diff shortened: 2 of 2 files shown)


digantdesai:
@zingo - If I run MV2 locally I get a different number for ethos PMU, any clue on how to start debugging? I believe George also confirmed that something odd is going on after comparing two runs.

zingo:
> @zingo - If I run MV2 locally I get a different number for ethos PMU, any clue on how to start debugging? I believe George also confirmed that something odd is going on after comparing two runs.

When you run it locally do you use run.sh or 
`python3 backends/arm/test/test_model.py --test_output=arm_test/test_model --target=ethos-u55-128 --model=mv2 --extra_flags="-DET_ATOL=2.00 -DET_RTOL=2.00" --build_libs`

zingo:
Found the bug and a PR is on it's way :) With it I now see 5.5M NPU cycles on mv2 when using Sram_Only with test_model.py also so it should match run.sh testing. Thanks for testing and reporting this :)

zingo:
I hope this PR will fix the problem you got. 
https://github.com/pytorch/executorch/pull/9902

gggekov:
Thanks @zingo @digantdesai ! 
For completeness, the performance on mobilenet_v2 on U55-128 on the Corstone-300 is: 

- **Shared Sram** memory mode, **Ethos_U55_Deep_Embedded** system config: 8.3M NPU active cycles
- **Shared Sram** memory mode, **Ethos_U55_High_End_Embedded** system config: 6.7M NPU active cycles
- **Sram Only** memory mode,  **Ethos_U55_Deep_Embedded** and **Ethos_U55_High_End_Embedded** - 5.5M NPU active cycles

For Shared_Sram, we see a difference in the performance between the Ethos_U55_Deep_Embedded and the Ethos_U55_High_End_Embedded system configs and that is expected. This is because the memory bandwidth and latency are different between the two system configurations. With the deeply embedded timing adapter settings, the NPU perceives bus bandwidth of 4bits/clock cycle(=250MB/s if you are @ 500MHz) on the external memory and for the High_End_Embedded, the bus bandwidth is 8bits/cc(or 500MB/s if the NPU is clocked at 500MHz) for the external memory. Note that even though the performance is different for Shared_Sram(8.3M vs 6.7M cycles), the PMU counters for amount of beats read on the AXI0/AXI1 interfaces are the same. In other words, the NPU reads the same amount of data between the deeply embedded & high end embedded system configs. This is again expected because we run the same model, just in the deeply embedded case the NPU has to wait longer for the data to arrive due to the lower bandwidth on the Flash. 

As for Sram_Only, we place everything in the SRAM and the SRAM bandwidth is the same between Deeply Embedded & High End systems, therefore we see identical performance.

zingo:
@digantdesai  OK to close this?

zingo:
Time to close this

PR #9834, opened 2025-04-02
Arm Backend: Fixes related to pytorch updates

SaoirseARM:
### Summary
- Update to error message for unsupported types
- Check file extension in test_debug_feats
- Adding expected failures for some unsupported operators in tests

### DIFF
--- a/backends/arm/test/models/test_nn_functional.py
+++ b/backends/arm/test/models/test_nn_functional.py
@@ -104,7 +104,15 @@ def test_nn_functional_MI(test_data):
             raise e
 
 
-@parametrize("test_data", module_tests)
+x_fails = {
+    "normalize": "MLETORCH-852: Support aten.index_put.default",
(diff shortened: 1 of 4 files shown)


zingo (approved):

backends/arm/test/models/test_nn_functional.py
```diff
@@ -104,7 +104,15 @@ def test_nn_functional_MI(test_data):
             raise e
 
 
-@parametrize("test_data", module_tests)
+x_fails = {
+    "normalize": "MLETORCH-852: Support aten.index_put.default",
```
digantdesai, line 108: MLETORCH-* - are these issues tracking ET work? if yes, would it be possible to put them under ET GH? Just for visibility. If I can access these from outside Arm that would be OK too.

SaoirseARM: Hi, 

Thanks for the review. 
We had some discussions yesterday on the best way to do this and we will start looking into this as soon as possible. 

Thanks, 
Saoirse

PR #9838, opened 2025-04-02
Fix build gating on optimized_portable_kernels

swolchok:
If we don't need optimized kernels, we don't need optimized portable kernels.
Pull-Request-resolved: https://github.com/pytorch/executorch/pull/9665


swolchok:
* (to be filled)

swolchok:
@manuelcandales apologies for the duplicate PR. having trouble with ghstack. you accepted this once already as #9665.

manuelcandales (approved):

PR #9840, opened 2025-04-02
Add build_optimized_size_test.sh

swolchok:
Adding a developer-facing size test script for a non-embedded-centric use case (optimized ops, exceptions on, etc.).


swolchok:
* (to be filled)

test/build_optimized_size_test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
kirklandsign, line 1: How do we trigger this? Add it to CI? Also is it possible to re-use https://github.com/pytorch/executorch/blob/main/test/build_size_test.sh?

swolchok: > How do we trigger this? Add it to CI?

As the description says, it is meant for development use.

> is it possible to re-use

No, that script wants to disable exceptions, which is appropriate in the embedded systems context but not the "everything else" context. Optimized kernels are not for the embedded systems context AIUI.

kirklandsign (approved):

PR #9845, opened 2025-04-02
Fix scalar type logging from #9751

swolchok:
I forgot that we have a nice system for logging ScalarTypes as strings already. Let's use it.


swolchok:
* (to be filled)

manuelcandales (approved):

PR #9868, opened 2025-04-03
Fix breaks from googletest upgrade (vulkan_compute_api_test) (#9760)

hershi:
Summary:

Missing include in vulkan_compute_api_test.cpp - added

Reviewed By: SS-JIA

Differential Revision: D72139603


SS-JIA:
This PR is no longer needed.

PR #9873, opened 2025-04-03
Arm backend: Improve pre-push hook

Sebastian-Larsson:
Now checks:
* License header
* Lintrunner
* Commit message

This is done for all commits that are not on the remote branch the current branch is tracking.

Commit message checks are:
* Subject limit is 72 characters
* Subject should start with "Arm backend: " + an imperative verb
* Body limit is 72 characters
* Must contain Signed-off-by

Lines can be shorter than 72, but unless it's a new paragraph each line should get to as close to 72 characters as possible. This is not checked by the hook.

### DIFF
--- a/backends/arm/scripts/pre-push
+++ b/backends/arm/scripts/pre-push
@@ -4,42 +4,178 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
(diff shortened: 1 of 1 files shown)


Sebastian-Larsson:
CI failure seems unrelated. Started happening after this was merged: https://github.com/pytorch/executorch/pull/9643

zingo (approved):

PR #9884, opened 2025-04-03
Create release.yml

metascroy:
Create release.yml for automatic GH release note generation


byjlw (approved):
it would be cool to have this broken in sub category so backends are under a backend top level category

PR #9895, opened 2025-04-04
Arm backend: Add info message to assertions in fold_qdq_with_annotated_qparams_pass

Sebastian-Larsson:
Add explanation for why the asserts would trigger.


zingo:
qnn fails are unrelated

zingo (approved):

PR #9896, opened 2025-04-04
Arm backend: Convert assert to throw ValueError in op_tanh

Sebastian-Larsson:
Asserts are converted to proper raises to ensure graph integrity.

It should not be possible for tanh to have more than 1 input for a correctly formatted graph, but in the node visitor we cannot know for sure that the graph is formatted correctly.

torch.tanh supports more data types than fp32, which is why it should be checked.


zingo (approved):

zingo:
fails are unrelated

PR #9897, opened 2025-04-04
Arm backend: Convert assert to throw TypeError in op_add

Sebastian-Larsson:
Asserts are converted to proper raises to ensure graph integrity.


zingo:
qnn fails are unrelated

zingo (approved):

PR #9899, opened 2025-04-04
Arm backend: Add pre-push checks for op tests

AdrianLundell:
- Check that @SkipIfNoCorstone is not used
- Check that @expectedFailureOnFVP is not used
- Check that unittest.TestCase is not used
- Check that on_fvp suffix is not used
- Check that op and target is parsed from the test name


digantdesai (approved):
Nice!

PR #9900, opened 2025-04-04
Arm backend: Align MobileNetV2 with MobileNetV3 unittest

oscarandersson8218:
- Port MV2 to use new test pipelines
- Use rand instead of randn data
- Adjust atol


zingo:
fails are unrelated

zingo (approved):

PR #9901, opened 2025-04-04
Arm backend: Remove use of TOSA_DBG_VERBOSE

oscarandersson8218:
Environment variable TOSA_DBG_VERBOSE was used to determine and setting the logging levels inside ArmBackend. This patch removes all uses of that variable.


zingo:
fails are unrelated

zingo (approved):

PR #9903, opened 2025-04-04
Arm backend: Add support for Leaky ReLU

gggekov:
Decompose LeakyReLU as operators supported by the Arm backend


oscarandersson8218 (approved):
Unrelated CI failures

Ninja91:
@gggekov Why do we need to use torch.full tensor creation instead of just using scalar multiplication? Why not do something like:
image

gggekov:
Hi @Ninja91 , 
The `%op4 = mul(%op3,%op2)` relies on the multiplication node visitor in `executorch/backends/arm/operators/op_mul.py` and that code assumes we multiply two tensors(we only support `aten.mul.Tensor`). We don't support multiplication of tensor by a scalar for now.

PR #9906, opened 2025-04-04
Arm backend: Skip instead of returning in Llama test

martinlsm:
If the user does not provide input files with the --llama_inputs argument, the test is not run. Change the return statement into a pytest.skip so that the test output show a skip rather than a pass.


digantdesai (approved):

zingo:
qnn fails is unrelated

zingo (approved):

PR #9923, opened 2025-04-05
Support slice ops with default start

pssrawat:
Summary:
Since a recent change we see None arg in the slice op

This is causing failure during XNNPack delegation, since XNNPack slice pass is trying to compare start_idx 'None' to 0.
This diff fixes that.

Differential Revision: D72503552


mcr229 (approved):
Thanks for the contribution :)

PR #9928, opened 2025-04-07
Arm backend: Add support for sqrt

fumchin:
The sqrt operation is implemented via the pow operator, sqrt(x) = pow(x, 0.5).


backends/arm/_passes/arm_pass_manager.py
```diff
@@ -103,6 +104,7 @@ def _tosa_080_BI_pipeline(self, exported_program: ExportedProgram) -> GraphModul
         self.add_pass(KeepDimsFalseToSqueezePass())
         self.add_pass(Conv1dUnsqueezePass(exported_program))
         self.add_pass(DecomposeSelectPass())
+        self.add_pass(DecomposeSqrtPass())
```
oscarandersson8218, line 107: I don't think we should run this pass in the BI case. If the sqrt isn't decomposed by `transform_for_annotation` (or if some other op is decomposed to a sqrt) it will not be quantized and we should not partition it. I propose you remove the pass from this pipeline and add sqrt here to guard us from partitioning it in such cases: https://github.com/pytorch/executorch/blob/302cb06209934fd97b866610f70b82016af64e01/backends/arm/operator_support/tosa_supported_operators.py#L229

oscarandersson8218 (changes requested):

oscarandersson8218 (approved):
Looks good to me! CI failures should be unrelated. Just waiting for https://github.com/pytorch/executorch/pull/9451 to be merged before merging this.

zingo (approved):

PR #9934, opened 2025-04-07
per_channel_group can't be dynamic

mcr229:
Summary: There are some dynamism issues that arise when checking the semantics of quantize_affine nodes. We avoid them by accounting for free_symbols.

Differential Revision: D72488540

### DIFF
--- a/backends/xnnpack/utils/quant_utils.py
+++ b/backends/xnnpack/utils/quant_utils.py
@@ -12,6 +12,7 @@ import torch
 from executorch.exir.backend.canonical_partitioners.config_partitioner import (
     format_target_name,
 )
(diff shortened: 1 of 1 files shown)


digantdesai (approved):
Looks good, we need to be more precise in the code and in the comments/docs about the word `dynamic`. We have dynamic shape, dynamic quant, dynamic weights (RHS in bmm in xnnpack), dynamic config. Just easy to assume the wrong one.

mcr229:
@digantdesai yea lol, it can definitely be confusing for folks. I wonder if there are more precise terms we can use to describe the different natures of dynamism. I guess we can add qualifiers like dynamic quantization, dynamic shape, dynamic weights, etc..

PR #9935, opened 2025-04-07
Allow emitting mutable buffer names in schema

JacobSzwejbka:
Summary: part 1 of letting people query tensors by names in runtime.

Differential Revision: D72579501

cc @angelayi


lucylq (approved):
Looks good! What is this intended to be used for, btw?

tarun292 (approved):

Issue #9940, opened 2025-04-07
torchao not installed by default for ExecuTorch 0.6 RC


mergennachin:
### 🐛 Describe the bug

After I install 

```
pip install --extra-index-url https://download.pytorch.org/whl/test/cpu executorch==0.6.0
```

The following commmand doesn't work: 

```
import torchao
```

Let's either 

- pin to particular version of ao in pyproject.toml directly similar to https://github.com/pytorch/executorch/pull/8038/files
- Make pip wheel build ao from third-party/ao submodule

If we go with the first route, one thing to make sure is that third-party/ao submodule hash is the same as pinned torchao release

### Versions

```
(executorch_06_qa_1) mnachin@mnachin-mbp executorch % pip freeze
attrs==25.3.0
cattrs==24.1.3
coremltools==8.1
exceptiongroup==1.2.2
execnet==2.1.1
executorch==0.6.0
expecttest==0.3.0
flatbuffers==25.2.10
hypothesis==6.130.10
iniconfig==2.1.0
mpmath==1.3.0
numpy==2.2.4
packaging==24.2
pandas==2.2.3
parameterized==0.9.0
pluggy==1.5.0
protobuf==6.30.2
pyaml==25.1.0
pytest==8.3.5
pytest-rerunfailures==15.0
pytest-xdist==3.6.1
python-dateutil==2.9.0.post0
pytz==2025.2
PyYAML==6.0.2
ruamel.yaml==0.18.10
ruamel.yaml.clib==0.2.12
six==1.17.0
sortedcontainers==2.4.0
sympy==1.13.3
tabulate==0.9.0
tomli==2.2.1
tqdm==4.67.1
typing_extensions==4.13.1
tzdata==2025.2
```

PR #9947, opened 2025-04-07
[Release 0.6] update torchao pin

metascroy:
This PR is the first step to addressing https://github.com/pytorch/executorch/issues/9940 and making torchao installable on ET/release/0.6.

1. This PR bumps torchao pin to torchao/release/0.10.0 branch.

2. We will follow-up with a PR in release/0.6 to bump the pin and add torchao==0.10.0 to the pyproject.toml file.

### DIFF
--- a/.github/workflows/build-wheels-macos.yml
+++ b/.github/workflows/build-wheels-macos.yml
@@ -61,7 +61,7 @@ jobs:
       post-script: ${{ matrix.post-script }}
       package-name: ${{ matrix.package-name }}
       # Meta's macOS runners do not have Xcode, so use GitHub's runners.
-      runner-type: macos-15-xlarge
+      runner-type: macos-latest-xlarge
       setup-miniconda: true
       smoke-test-script: ${{ matrix.smoke-test-script }}
       trigger-event: ${{ github.event_name }}
--- a/.github/workflows/build-wheels-macos.yml
+++ b/.github/workflows/build-wheels-macos.yml
@@ -61,7 +61,7 @@ jobs:
       post-script: ${{ matrix.post-script }}
       package-name: ${{ matrix.package-name }}
       # Meta's macOS runners do not have Xcode, so use GitHub's runners.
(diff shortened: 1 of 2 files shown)


.github/workflows/build-wheels-macos.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       post-script: ${{ matrix.post-script }}
       package-name: ${{ matrix.package-name }}
       # Meta's macOS runners do not have Xcode, so use GitHub's runners.
-      runner-type: macos-latest-xlarge
+      runner-type: macos-15-xlarge
```
metascroy, line 64: torchao does not build on < macos15 due to "std::aligned_alloc" not being implemented until macos15.

mergennachin (approved):
Okay great, let's cherry-pick onto ET 0.6 branch.

pyproject.toml
```diff
@@ -66,6 +66,8 @@ dependencies=[
   "ruamel.yaml",
   "sympy",
   "tabulate",
+  # Keep this version in sync with: third-party/ao pin
+  "torchao==0.10.0",
```
mergennachin, line 70: In main, how would you depend on latest ao changes? I'm worried that if you bump the hash in third-party/ao (that may not have the a dedicated release) and it becomes inconsistent and confusing.

Here's what I suggest

In main: 

- Only update third-party/ao to a hash that is consistent with 0.10.0. But not update pyproject.toml

This is not 100% necessary but it's good because (1) we're testing the CI in main (2) making sure main's hash is not older than release branch's

In release branch branch:

- update third-party/ao to a hash that is consistent with 0.10.0. 
- update pyproject.toml to be 0.10.0
- update the release checklist to reflect this new process about pinning to torchao.

mergennachin:
Also in the summary of the PR, please add context and corresponding issue (https://github.com/pytorch/executorch/issues/9940)

PR #9953, opened 2025-04-08
[ET][Testing] Build test_backend_compiler_lib when testing is on

mcr229:
Adding test_backend_compiler_lib to pybinds when tests are on. This allows us to run some more backend tests in the OSS build. There were a couple of other test files that weren't able to be enabled: 
```
    --ignore=exir/backend/test/test_backends.py
    --ignore=exir/backend/test/test_backends_lifted.py
    --ignore=exir/backend/test/test_partitioner.py
```

This was because they were doing some torch.ops.load_library("buck_path") or there was another backend ExecutorBackend that was required. I'll limit the changes of this PR to just be enablement of the BackendWithCompilerLibDemo backend.


mcr229:
* (to be filled)

larryliu0820 (approved):

exir/backend/test/test_lowered_backend_module.py
```diff
@@ -109,11 +107,7 @@ def forward(self, x):
             torch.allclose(model_outputs[0], expected_res, atol=1e-03, rtol=1e-03)
         )
 
-    @given(
-        unlift=st.booleans(),  # verify both lifted and unlifted graph
```
cccclai, line 113: Any specific reason we don't test lifted vs unlifted?

mcr229: these variables aren't actually being used, I'm not sure why it's not being tested, but it wasn't being tested to begin with, and pytest complains about that, so i'm just cleaning up here.

cccclai: Oh hmm, how did you find out they are not tested?

mcr229: unlift variable isn't used in the test body. pytest complains about that and says given is useless if that's the case.

cccclai: Ah looks like it was deleted in  https://github.com/pytorch/executorch/pull/3169

cccclai: Ah I understand now, it was used when exported program decided to lift all the params. It's not needed now params are all lifted anyway

cccclai (approved):
Thanks!

Gasoonjia (approved):

PR #9980, opened 2025-04-08
Add maven version in main in Getting Started page

mergennachin:


docs/source/getting-started.md
```diff
@@ -111,6 +111,8 @@ To add the library to your app, add the following dependency to gradle build rul
 dependencies {
   implementation("org.pytorch:executorch-android:0.5.1")
 }
+
+# See available versions in https://repo.maven.apache.org/maven2/org/pytorch/executorch-android
```
kirklandsign, line 115: I would recommend https://mvnrepository.com/artifact/org.pytorch/executorch-android which is better formatted. This page takes about 1 day to update after we upload.

kirklandsign (approved):

PR #9989, opened 2025-04-08
Improve android instrumentation test experience and add docs

kirklandsign:
Add README for Android

Make instrumentation test easier for users on their local development.


GregoryComer (approved):
Thanks for working on this. Looks great. Can we document the test setup process under a README or similar?

PR #10034, opened 2025-04-10
Recipe and Input class definitions with e2e export

tarun292:
Based on the discussion in https://github.com/pytorch/executorch/discussions/9027

This PR adds the executorch.export API and all the supporting components required for it. At a high level the executorch.export API takes in a model, example inputs and a recipe, then underneath the hood executes all the steps required to export, quantize and lower the model based on the recipe.

The pipeline consists of a staged setup where each major step in the process such as Export, Quantization etc. is listed as a separate stage and a chain of these is formed and then executed. The result of this will be that we'll get a PTE file which can then be executed on device.

The major new components added in this PR are:
- class definition for ExportSession and ExportRecipe
- Definitions for each stage in the process
- executorch.export API which will return a session object that the user can then use to get access to the PTE file, run the model via pybindings, print delegation info etc.


export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+        artifact_dir=artifact_dir,
+    )
+    
+    if apply_quantization:
+        manager.export(apply_quantization=True)
+    
+    return manager
+
+
+class ExportSession:
```
digantdesai, line 67: This is a good idea, I like it. But just a pet peeve, let's refactor this before starting, and organize the logic so that it can scale better, else I fear it will end up becoming a giant bowl of spaghetti.

Suggestion, make a `class Stage(ABC)` and derive others from it like `Quantize(Stage)` and `Export(Stage)` and then Session just manages the pipeline with `Stage`s. 

See - https://github.com/pytorch/executorch/blob/409447d75a1524c1acc8f8ea894c2e13dd723a79/backends/xnnpack/test/tester/tester.py#L424

export/_recipe.py
```diff
@@ -0,0 +1,83 @@
...
+                              during edge lowering
+        transform_check_ir_validity: Whether to check IR validity during transformation
+        partitioners: Optional list of partitioners for model partitioning
+        executorch_backend_config: Optional backend configuration for ExecuTorch
+        mode: Export mode (debug or release)
+    """
+
+    name: Optional[str] = None
+    quantizer: Optional[Quantizer] = None
+    edge_compile_config: Optional[EdgeCompileConfig] = ( # pyre-ignore[11]: Type not defined
```
digantdesai, line 62: mypy?

export/_recipe.py
```diff
@@ -0,0 +1,83 @@
...
+    ] = None
+    edge_transform_passes: Optional[Sequence[PassType]] = None
+    transform_check_ir_validity: bool = True
+    partitioners: Optional[list[Partitioner]] = None
+    executorch_backend_config: Optional[ExecutorchBackendConfig] = ( # pyre-ignore[11]: Type not defined
+        None
+    )
+    mode: Mode = Mode.RELEASE
+
+    def get_quantizer(self) -> Optional[Quantizer]:
```
digantdesai, line 76: nit property?

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+from executorch.runtime import Runtime, Verification
+from tabulate import tabulate
+from torch.ao.quantization import allow_exported_model_train_eval
+from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
+from torch.export import export_for_training, ExportedProgram
+
+from ._recipe import ExportRecipe
+
+
+def export(
```
digantdesai, line 20: nit, WDYT about? I understand we want to do `s/torch.export/executorch.export` but it feels a bit confusing given we are returning something not analogous to the model which torch.export does.
```suggestion
def create_export_session(
# or
Session.__init__
# or
@classmethod
Session.create()
```

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+    model: Union[nn.Module, Dict[str, nn.Module]],
+    example_inputs: Union[List[tuple[torch.Tensor, ...]], Dict[str, List[tuple[torch.Tensor, ...]]]],
+    export_recipe: ExportRecipe,
+    name: Optional[str] = None,
```
digantdesai, line 24: what is the rationale for this?

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+        model: The PyTorch model(s) to export, either a single model or a dictionary
+              mapping method names to models
+        example_inputs: Example inputs for the model(s), either a list of input tuples
+                      or a dictionary mapping method names to lists of input tuples
+        export_recipe: Contains the configuration for the export process
+        name: Optional name for the export
+        dynamic_shapes: Optional dynamic shape specifications
+        constant_methods: Optional dictionary of constant methods
+        artifact_dir: Optional directory to store artifacts
+        apply_quantization: Whether to apply quantization during export, defaults to False
```
digantdesai, line 46: this feels a bit confusing

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+
+    Returns:
+        A configured ExportSession instance with the export process completed if requested
+    """
+    manager = ExportSession(
```
digantdesai, line 51: nit
```suggestion
    session = ExportSession(
```

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+
+from ._recipe import ExportRecipe
+
+
+def export(
+    model: Union[nn.Module, Dict[str, nn.Module]],
+    example_inputs: Union[List[tuple[torch.Tensor, ...]], Dict[str, List[tuple[torch.Tensor, ...]]]],
+    export_recipe: ExportRecipe,
+    name: Optional[str] = None,
+    dynamic_shapes: Optional[Union[Any, Dict[str, Any]]] = None,
```
digantdesai, line 25: why is this not on `Session.export()`?

digantdesai (approved):
Looks good at a high level. Left some comments.

export/_export.py
```diff
@@ -0,0 +1,357 @@
...
+)
+from executorch.runtime import Runtime, Verification
+from tabulate import tabulate
+from torch.ao.quantization import allow_exported_model_train_eval
+from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
+from torch.export import export_for_training, ExportedProgram
+
+from ._recipe import ExportRecipe
+
+
```
mergennachin, line 19: We should annotate with executorch.exir._warnings.experimental decorators

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+from torch.ao.quantization import allow_exported_model_train_eval
+from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
+from torch.export import export_for_training, ExportedProgram
+from torchao.quantization import quantize_
+from torchao.utils import unwrap_tensor_subclass
+
+from .recipe import ExportRecipe
+
+
+class Stage(ABC):
```
jackzhxng, line 25: I like this 👍🏻

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+    def run(self, **kwargs) -> None:
+        """
+        Executes this stage with the given inputs.
+
+        Each concrete stage class implements this method with specific parameter names.
+        """
+        pass
+
+    @abstractmethod
+    def get_outputs(self) -> Any:
```
jackzhxng, line 52:
```suggestion
    def get_artifacts(self) -> Any:
```

Think this conveys the intent of this function better when the user hasn't read the docstring

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+    def __init__(self, quantizer: Any) -> None:
+        self._quantizer = quantizer
+        self._quantized_models: Dict[str, nn.Module] = {}
+        self._model_dict: Dict[str, nn.Module] = {}
+        self._exported_models_dict: Dict[str, nn.Module] = {}
+        self._example_inputs_dict: Dict[str, List[tuple[torch.Tensor, ...]]] = {}
+
+    @property
+    def name(self) -> str:
+        return "quantize"
```
jackzhxng, line 334: Would be worth to comment somewhere that quantization may also happen in `SourceTransformStage`

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        Raises:
+            RuntimeError: If the executorch program manager is not initialized
+        """
+        if self._executorch_program_manager is None:
+            raise RuntimeError(
+                "Executorch program manager is not initialized. Run export() first."
+            )
+        return self._executorch_program_manager.buffer
+
+    def get_example_input(
```
jackzhxng, line 800: I feel like this and `run_method` seem a bit confusing to have in the ExportSession?

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+            self._model = result
+        elif stage_name == "export":
+            self._exported_program = result
+        elif stage_name == "edge_transform_and_lower":
+            self._edge_program_manager = result
+            self._delegation_info = self._stages[stage_name].delegation_info
+        elif stage_name == "executorch":
+            self._executorch_program_manager = result
+
+    def _run_pipeline(self, start_stage: str) -> None:
```
jackzhxng, line 693: Why do we need this start_stage stuff, can we just not take any parameters and just run the all stages in self._stages?

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+
+        Raises:
+            RuntimeError: If the executorch program manager is not initialized
+        """
+        if self._executorch_program_manager is None:
+            raise RuntimeError("Executorch program manager is not initialized.")
+        return self._executorch_program_manager
+
+
+class SourceTransformStage(Stage):
```
jackzhxng, line 277: NIT: have this and QuantizeStage declared in the order of the export process, this one declared at the top

jackzhxng (approved):

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+
+    def __init__(
+        self,
+        pre_edge_transform_passes: Optional[
+            Callable[[ExportedProgram], ExportedProgram]
+        ] = None,
+    ) -> None:
+        self._exported_program: Dict[str, ExportedProgram] = {}
+        self._pre_edge_transform_passes = pre_edge_transform_passes
+        self._model_dict: Dict[str, nn.Module] = {}
```
kimishpatel, line 75: is this to capture more than one Module to be exported? why do we need this?

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        """
+        # Store inputs
+        self._model_dict = models.get("model", {})
+
+        if export_config is not None:
+            self._example_inputs_dict = export_config.get("example_inputs", {})
+            self._dynamic_shapes_dict = export_config.get("dynamic_shapes", {})
+
+        # Check if we need to do export (if _exported_program is empty)
+        if not self._exported_program:
```
kimishpatel, line 105: What if you want to rexport? For different input shapes or export args? If ExportStange is strictly for one export session and to export with different args you need a new ExportStage than I will make it immutable.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        self._stages["executorch"] = ExecutorchStage(
+            backend_config=self._export_recipe.executorch_backend_config
+        )
+
+        # Define the pipeline as a list of stages in execution order
+        self._pipeline = ["export", "edge_transform_and_lower", "executorch"]
+
+        # If quantization is available, add it after source transform
+        if "quantize" in self._stages:
+            self._pipeline.insert(0, "quantize")
```
kimishpatel, line 566: As I said elsewhere I think quantize should not be responsible for running export. We should run export, run quantizer, prepare, convert, and run export again. This call can be managed within here by detecting the need to run export again if we have pt2e quantize requested

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+            self._current_stage_index = i
+
+            # Get the primary input and configuration parameters for this stage
+            primary_input, config_params = self._get_stage_inputs(stage_name)
+
+            # Run the stage
+            result = self._run_stage(stage_name, primary_input, config_params)
+
+            # Store the result
+            self._store_stage_result(stage_name, result)
```
kimishpatel, line 730: Yeah at thsi point i expect jsut iterate through stages and run without needing to figure out input and storage stage outputs. Outputs of each stage should be there in `get_output/artifact` if needed

export/export.py
kimishpatel: Yeah I agree. We should not have this start_stage thing. We should just construct pipeline that can just be run with each stage in sequence. Construction of the pipeline should take care of apprpriate deps

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+
+        # Access the first element of the list for this method
+        example_inputs_list = self._example_inputs[method_name]
+        if not example_inputs_list:
+            raise ValueError(f"Example inputs list for method {method_name} is empty")
+
+        # The original code expects this to be a tuple of tensors
+        return self._example_inputs[method_name][0]
+
+    def run_method(
```
kimishpatel, line 827: why is this part of this class

kimishpatel (commented):
Left quite a few comments

kimishpatel (commented):
I didnt "request changes" if you feel strongly about landing, but I think we should consider addressing these comments

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        self._dynamic_shapes_dict: Dict[str, Any] = {}
+
+    @property
+    def name(self) -> str:
+        return "export"
+
+    def run(
+        self,
+        models: Dict[str, Any],
+        export_config: Optional[Dict[str, Any]] = None,
```
jackzhxng, line 86: Looking at this PR again, why isn't this a part of the ExportRecipe?

export/export.py
tarun292: Updated.

export/export.py
tarun292: The way i'm thinking is that export session lets the users do almost all the things they'd like to with an executorch model like load the model and run via pybindings, get reference inputs etc. and that's why I added these helper utils.

export/export.py
tarun292: Addressed in latest version.

export/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
tarun292: Thought quite a bit about this, the main reasoning to call this export too is because we're consuming the export call within the API. One of the main points of this API is that if someone wants to target ExecuTorch they don't even need to learn about torch.export (although many of them probably will).

export/export.py
tarun292: Yep a user could possibly pass in multiple methods to be exported and wanted to have support for that from the beginning. Something we struggled with in ModAI by trying to add it later.

export/export.py
tarun292: This wasn't needed, removed it. One export session is intended to be only for one set of input args.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+    """
+
+    def __init__(
+        self,
+        partitioners: Optional[List[Any]] = None,
+        transform_passes: Optional[Sequence[Callable[[Any], Optional[Any]]]] = None,
+        compile_config: Optional[Any] = None,
+    ) -> None:
+        self._partitioners = partitioners
+        self._transform_passes = transform_passes
```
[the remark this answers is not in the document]
tarun292: Transform passes are what we run after to_edge e.g. edge_manager.transform(), and partitioners are the ones we use in to_backend.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        self._compile_config = compile_config
+        self._edge_program_manager: Optional[EdgeProgramManager] = None
+        self._delegation_info = None
+        self._exported_program: Dict[str, ExportedProgram] = {}
```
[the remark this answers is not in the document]
tarun292: This is needed in later stages right after we export and do quantization and to_edge etc.

export/export.py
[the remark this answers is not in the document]
tarun292: Thought a bit about this, it doesn't necessarily mean all the needed source transforms will be squeezed in here, but this will be needed for quantize_ and maybe other things in the future so it made sense to have a source transform stage.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+
+        Args:
+            models: Dictionary mapping method names to PyTorch models
+            **kwargs: Additional keyword arguments (not used)
+        """
+        # Store the original models
+        self._transformed_models = models
+
+        # Check if there's a quantization recipe with ao_base_config
+        if self._quantization_recipe and self._quantization_recipe.ao_base_config:
```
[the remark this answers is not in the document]
tarun292: This is an optional config right, if this is not specified we can just continue without quantizing.

export/export.py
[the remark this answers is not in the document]
tarun292: Thought a bit about this, structuring it this way doesn't necessarily bring us a lot of benefits as `EagerQuantRecipe` and `ExportQuantRecipe` don't necessarily have a lot of common things which they can inherit from by having a base `QuantizationRecipe` class.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        self._model_dict = model_data["model"]
+        self._exported_models_dict = {}
+
+        # Initialize with empty dictionaries
+        self._example_inputs_dict = {}
+        dynamic_shapes_dict = {}
+
+        if calibration_config is not None:
+            self._example_inputs_dict = calibration_config.get("example_inputs", {})
+            dynamic_shapes_dict = calibration_config.get("dynamic_shapes", {})
```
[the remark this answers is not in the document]
tarun292: Yep definitely in the plans for next udpate.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+            model.eval()
+
+            # Export the model for training
+            kwargs = {}
+            if dynamic_shapes is not None:
+                kwargs["dynamic_shapes"] = dynamic_shapes
+
+            self._exported_models_dict[method_name] = export_for_training(
+                model, self._example_inputs_dict[method_name][0], **kwargs
+            ).module()
```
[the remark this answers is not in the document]
tarun292: Updated in latest version.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        # Process inputs
+        for method_name, _ in self._model_dict.items():
+            # Use the pre-exported model
+            captured_model = self._exported_models_dict[method_name]
+
+            # Prepare the model for quantization
+            prepared_model = prepare_pt2e(captured_model, self._quantizer)  # type: ignore
+
+            # Allow the exported model to switch between train and eval modes
+            allow_exported_model_train_eval(prepared_model)
```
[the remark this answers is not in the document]
tarun292: This is something we noticed that was needed internally for dropout and batchnorm.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+    )
+    session.export()
+
+    return session
+
+
+@experimental(
+    "This API and all of its related functionality such as ExportSession and ExportRecipe are experimental."
+)
+class ExportSession:
```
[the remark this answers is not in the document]
tarun292: This is an active session where users can do things like run the model via pybindings, print delegation info etc. that's why i decided to call it session rather than pipeline.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        self._constant_methods = constant_methods
+        self._artifact_dir = artifact_dir
+        self._export_recipe = export_recipe
+
+        # Initialize stages
+        self._stages = {}
+
+        # Create the source transform stage if a quantization recipe is provided
+        if self._export_recipe.quantization_recipe is not None:
+            self._stages["source_transform"] = SourceTransformStage(
```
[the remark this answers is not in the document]
tarun292: Within this stage in the future i wanted to be able to support running source transforms too if needed.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+            )
+
+            # Create the quantize stage if a quantizer is provided
+            if self._export_recipe.quantization_recipe is not None:
+                quantizer = self._export_recipe.quantization_recipe.get_quantizer()
+                if quantizer is not None:
+                    self._stages["quantize"] = QuantizeStage(quantizer=quantizer)
+
+        # Create the export stage
+        self._stages["export"] = ExportStage(
```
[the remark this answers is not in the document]
tarun292: Updated in latest version.

export/export.py
tarun292: We need to store this because it's used in all the later stages such as to_edge etc.

export/export.py
```diff
@@ -0,0 +1,866 @@
...
+        # - edge_transform_and_lower: exported_programs, transform_config
+        # - executorch: edge_program, backend_options
+        # - quantize: model_data, calibration_config
+        # But we use generic parameter names here for simplicity
+        self._stages[stage_name].run(primary_input, config_params)
+
+        # Return the outputs
+        return self._stages[stage_name].get_outputs()
+
+    def _get_stage_inputs(self, stage_name: str) -> tuple[Any, Dict[str, Any]]:
```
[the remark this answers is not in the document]
tarun292: I thought a lot about this too, in order to deal with quantize_ and later on other things such as custom op replacement in eager mode we do need to do it and it's quite a bit of complexity that'll be better hidden from the users.

export/export.py
tarun292: This means that the quantization recipe doesn't necessarily have configs to use with quantize_ but might still have pt2e quantization. In that case it doesn't make sense to throw an error right?

export/export.py
tarun292: I think that's a valid point, but i've felt like having all the quantization configuration related items in one class would generally be easier for recipe authors. We when creating the pipeline decide what source transforms need to be run and what is in pt2e. This isn't set in stone though, we'll definitely iterate on this as we add more recipes in XNNPack, CoreML etc. and we can arrive at a different structure over the next few weeks if the current one isn't flexible enough.

export/export.py
tarun292: This contains example inputs and dynamic shapes which shouldn't be a part of the recipe.

tarun292:
@kimishpatel @jackzhxng thanks for the detailed review, really appreciate it! Addressed all of your comments:

In the latest version the major changes i made are:
- Removed start_stage API and string input for stage name, rather just have the run_pipeline function now which runs through the entire pipeline. Changed quite a bit how the pipeline is constructed now.
- The pipeline is a list of stages now and not a dictionary.
- Export happens independently of quantization now.
- Added get_artifact which gets the output of each stage.

digantdesai:
Thanks @tarun292 - love the pipeline design.

export/export.py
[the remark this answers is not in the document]
kimishpatel: I am against having SourceTranformStage just for quantization. It doesnt feel right. In fact quantize_ is not a source transformation. So I am not sure why quantization is wrapped up here at all.

export/export.py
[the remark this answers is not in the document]
kimishpatel: > deal with quantize_ and later on other things such as custom op replacement in eager mode

Again eager mode should not belong to this. It is coupling unrelated things together.

export/export.py
kimishpatel: Not sure I follow what you mean.

PR #10040, opened 2025-04-10
Add support for the default float, int and bool types for executorch::runtime::etensor::Scalar to` method

### LINKED ISSUE #9500: ExecuTorch Scalar to() supports fewer types than c10::Scalar, breaking source compatibility
### 🐛 Describe the bug

some_scalar.to<float>() is legal for c10::Scalar, but not ExecuTorch's Scalar.

### Versions

main

ahbpp:
### Summary
* Added support for `float`, `::executorch::runtime::etensor::Half`, and `::executorch::runtime::etensor::BFloat16` for `to` method of ExecuTorch's Scalar 
* Replaced separate `to` methods with `to##name()` macros (same as [PyTorch c10::Scalar](https://github.com/pytorch/pytorch/blob/main/c10/core/Scalar.h#L115)). 
* Added constructor for `float` type. 
* This PR does not cover `uint16/32/64_t` types (since 'uint64_t' can overflow' int64_t', separate logic is required) and does not include `executorch::runtime::etensor::complex`. 
* I haven’t added overflow checks before because  I'm not sure if I can copy (c10 dir in executorch) [`c10::overflows.h` from torch](https://github.com/pytorch/pytorch/blob/main/c10/util/overflows.h).  (I will add an overflow check after I figure out what is the right way to do this :))

I need a bit further guidance related to adding support for more complex types and handling overflows:
1. I'm thinking about whether it’s better to create a separate `TypeCast.h` for typecasting (similar to PyTorch c10). Would this be better than having the casting code in the `Scalar` class?
2. Can I copy and use `c10::overflows.h` from pytorch c10 to handle overflows?

Fixes #9500 

P.S. If you have any concerns about this PR, please let me know, I will fix :)

### Test plan

I added tests in' runtime/core/portable_type/test/scalar_test.cpp' for additional types types.

### DIFF
--- a/runtime/core/portable_type/scalar.h
+++ b/runtime/core/portable_type/scalar.h
@@ -10,6 +10,8 @@
 
 #include <executorch/runtime/core/portable_type/bfloat16.h>
 #include <executorch/runtime/core/portable_type/half.h>
+#include <executorch/runtime/core/portable_type/complex.h>
+#include <executorch/runtime/core/portable_type/scalar_type.h>
 #include <executorch/runtime/core/tag.h>
 #include <executorch/runtime/platform/assert.h>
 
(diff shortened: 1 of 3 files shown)


JacobSzwejbka:
Sorry for the very late reply. Im not aware of any need for ExecuTorch to support these types in Scalar. I dont think we allow serializing them. Can you explain the motivation of the PR?

Issue #9971, opened 2025-04-08
[Request impl] Gracefully error out in ETDump


Gasoonjia:
### Context and Target:
In the `etdump_flatcc.cpp` file, there are instances where the class may directly raise an error and shut down the system when encountering issues. An example can be found [here](https://github.com/pytorch/executorch/blob/main/devtools/etdump/etdump_flatcc.cpp#L69).
The preferred approach is to handle these errors gracefully by returning an error code to the user, allowing them to decide how to address the issue.

### TODO:

- Modify the code to return an error code to the user instead of crashing the system when an error occurs.
If applicable, use the `ET_CHECK_OR_RETURN_ERROR` macro, as seen [here](https://github.com/pytorch/executorch/blob/main/devtools/etdump/etdump_flatcc.cpp#L367).
Alternatively, manually return an error code, as demonstrated [here](https://github.com/pytorch/executorch/blob/main/devtools/etdump/data_sinks/buffer_data_sink.cpp#L24), if there's extra work like `free` need to be done.

- Update the return type of the function as necessary and ensure corresponding tests are updated.

Consider breaking this task into two or three pull requests for better structure and clarity.

zhongmingyuan:
I want to work on it as a learning opportunity. May I take this issue?

zhenyan-zhang-meta:
Please take it, reference my PR as a good starting point.

PR #10056, opened 2025-04-10
[#9971] Gracefully error out in ETDump for get_flatbuffer_scalar_type

zhenyan-zhang-meta:
* __->__ #10056

Following https://github.com/pytorch/executorch/issues/9971

- Update get_flatbuffer_scalar_type return type to Result<T>
- Iteratively update functions that calling the functions with result type changed:
  - Check returns, if with an error, pass above the error.
  - If unable to pass error, update the return type as Result<T>

Differential Revision: [D72771753](https://our.internmc.facebook.com/intern/diff/D72771753/)

### DIFF
--- a/runtime/core/event_tracer.h
+++ b/runtime/core/event_tracer.h
@@ -314,7 +314,7 @@ class EventTracer {
    * @param[in] evalue_type Indicates what type of output this is logging e.g.
    * an intermediate output, program output etc.
    */
-  virtual void log_evalue(
+  virtual Result<bool> log_evalue(
       const EValue& evalue,
       LoggedEValueType evalue_type) = 0;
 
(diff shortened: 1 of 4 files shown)


runtime/core/event_tracer.h
```diff
@@ -314,7 +314,7 @@ class EventTracer {
    * @param[in] evalue_type Indicates what type of output this is logging e.g.
    * an intermediate output, program output etc.
    */
-  virtual void log_evalue(
```
Gasoonjia, line 317: Please update the comment for result value

Gasoonjia (approved):
LGTM! Thanks for update!
Please update the comment before landing it.

PR #10043, opened 2025-04-10
[arm][ez] Add `xfail` for `norm` tests

SS-JIA:
Summary:
## Context

https://github.com/pytorch/executorch/pull/9938 made it so that `linalg_vector_norm` is now decomposed when exporting to Edge. However, this broke some tests in the arm delegate because export passes cannot handle the decomposed operator sequence. To account for this, add `xfail` for the failing tests since `linalg_vector_norm` is not supported in TOSA yet.

## Changes

Add `xfail` for `norm` tests in `test_torch_functions.py`

Test Plan:
## Test Plan

Check CI that failing test is recovered.

PR #10057, opened 2025-04-10
Arm backend: Add test_arm_baremetal.sh checks

perheld:
Test_arm_baremetal.sh didnt warn about missing setup_path.sh like other scripts. This patch makes it more like others.

Change-Id: I124df660ce873d98fd067ce2441a396ac3c183fc


perheld:
Strange error. `FAILED backends/arm/test/models/test_torch_functions.py::test_torch_fns_MI[norm] - Exception: An error occurred when running the 'KeepDimsFalseToSqueezePass' pass after the following passes:`

I cant see how its related.

zingo:
The error might go away when https://github.com/pytorch/executorch/pull/10043 is merged

zingo (approved):

PR #10064, opened 2025-04-10
Llama exported model from executorch-community

mergennachin:
We have the recipe and .pte file on ExecuTorch-Community on HF. So let's just use that.


lucylq (approved):
Wow, nice! This is great

PR #10071, opened 2025-04-10
allow not memory planning mutable buffers

JacobSzwejbka:
Summary:
Config option to not memory plan mutable buffers.

This when paired with a future runtime PR will allow users to retrieve buffers by name in the runtime and then set their dataptr.

Differential Revision: D72749868

cc @angelayi

### DIFF
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -1640,13 +1640,26 @@ class _TopLevelEmitter(_Emitter):
                 else:
                     spec.extra_tensor_info.fully_qualified_name = fqn
                     spec.extra_tensor_info.location = TensorDataLocation.EXTERNAL
-            if self.emitter_state.emit_mutable_buffer_names and is_mutable_buffer:
-                if spec.extra_tensor_info is None:
-                    spec.extra_tensor_info = ExtraTensorInfo(
-                        fully_qualified_name=fqn, location=TensorDataLocation.SEGMENT
+
+            if is_mutable_buffer:
+                # Emit names if we are supposed to.
+                if self.emitter_state.emit_mutable_buffer_names:
+                    if spec.extra_tensor_info is None:
+                        spec.extra_tensor_info = ExtraTensorInfo(
+                            fully_qualified_name=fqn,
+                            location=TensorDataLocation.SEGMENT,
+                        )
+                    else:
+                        spec.extra_tensor_info.fully_qualified_name = fqn
+                # if We aren't emitting the name then it needs to be memory planned.
+                elif spec.mem_id is None or spec.mem_offset is None:
+                    raise InternalError(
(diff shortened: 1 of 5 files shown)


exir/emit/_emitter.py
```diff
@@ -1640,13 +1640,26 @@ def placeholder(  # noqa: C901
                 else:
                     spec.extra_tensor_info.fully_qualified_name = fqn
                     spec.extra_tensor_info.location = TensorDataLocation.EXTERNAL
-            if self.emitter_state.emit_mutable_buffer_names and is_mutable_buffer:
-                if spec.extra_tensor_info is None:
-                    spec.extra_tensor_info = ExtraTensorInfo(
-                        fully_qualified_name=fqn, location=TensorDataLocation.SEGMENT
+
+            if is_mutable_buffer:
```
JacobSzwejbka, line 1644: This is the actual core logic change, the rest of the changes are mostly piping a flag around

exir/emit/test/test_emit.py
```diff
@@ -1838,8 +1838,40 @@ def forward(self, x):
         ep = to_edge(ep)
         # Lower the graph to executorch.
         ep = ep.to_executorch(
-            config=ExecutorchBackendConfig(emit_mutable_buffer_names=True)
+            config=ExecutorchBackendConfig(
```
tarun292, line 1841: Maybe also add a __post_init__ in the `ExecutorchBackendConfig` dataclass that asserts if `emit_mutable_buffer_names` is `False` and `alloc_mutable_buffers` is also False.

exir/emit/_emitter.py
```diff
@@ -1640,13 +1640,26 @@ def placeholder(  # noqa: C901
...
+                # Emit names if we are supposed to.
+                if self.emitter_state.emit_mutable_buffer_names:
+                    if spec.extra_tensor_info is None:
```
tarun292, line 1647: A little confused by this as i haven't kept track of the tensor info changes, what does it mean if `extra_tensor_info` is `None` and if it isn't we overwrite the fqn again? Maybe add a small comment here too.

tarun292 (approved):

exir/emit/test/test_emit.py
JacobSzwejbka: I considered this but since the MemoryPlanningPass is user configureable there is no guarantee memory_planning_pass.alloc_mutable_buffers exists

exir/emit/_emitter.py
JacobSzwejbka: None just means no one has created one, if its not none they might not have populated the fqn so I populate it. I guess I could check what it is before overwriting it but the name is unique so it should always be safe to do this.

PR #10093, opened 2025-04-11
Qualcomm AI Engine Direct - Add rewrite function of observer

chunit-quic:
- Add function to rewrite observer after prepare_pt2e
- Add corresponding test case

### DIFF
--- a/backends/qualcomm/tests/test_qnn_delegate.py
+++ b/backends/qualcomm/tests/test_qnn_delegate.py
@@ -49,6 +49,7 @@ from executorch.backends.qualcomm.utils.utils import (
     generate_qnn_executorch_compiler_spec,
     PyQnnManagerAdaptor,
     QnnPartitioner,
+    rewrite_prepared_observer,
     skip_annotation,
     to_edge_transform_and_lower_to_qnn,
     update_spill_fill_size,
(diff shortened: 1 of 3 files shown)


chunit-quic:
Hi @cccclai,

This PR is for the requested functionality to overwrite a quanization parameters after calibration. It can handle shared observer too. Thank you!

cccclai:
@chunit-quic thanks for putting up the pr. cc: @sxu @billmguo

cccclai:
Mind sharing a bit more details for the request? I probably miss it

chunit-quic:
> Mind sharing a bit more details for the request? I probably miss it

No problem. :)
We received some feature requests in a mail thread regarding quantization requirements. This particular request is for the following purpose:
```
The QNN Quantizer should allow users to override quantization parameters for specific tensors, regardless of the data ranges observed during calibration or QAT. This override must respect the transitive closures established by SharedQuantizationSpec.
```

After a brief discussion with the team, we concluded that the rewriting stage should occur after calibration but before conversion. That's essentially the background.

sxu:
Looks like this would work to me, however it would be great if someone from AO can confirm this kind of overriding act postprocessing submodule is enough and there's no other references to them.

sxu:
Maybe @jerryzh168 can confirm?

sxu (approved):

cccclai:
@kirklandsign can you help checking the CI signal for this one too? I remember there are some errors.

billmguo (approved):

cccclai:
Hey there some errors likely due to the pytorch pin update. Mind rebasing again? Sorry for the inconvinience

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -2913,6 +2914,37 @@ def test_qnn_backend_dynamic_shape(self):
...
+        new_obs = observer.FixedQParamsObserver(
+            scale=0.004,
+            zero_point=0,
+            dtype=torch.uint8,
+            quant_min=0,
+            quant_max=255,
+            qscheme=torch.per_tensor_affine,
+        )
+
+        rewrite_prepared_observer(prepared, {"activation_post_process_2": new_obs})
```
jerryzh168, line 2939: @sxu does your callback work for this?

cccclai:
> No problem. Just rebased. Pleasefel free to let me know if there is any issue. Thank you!

Did you rebase? Seems like there is no commit..

chunit-quic:
Ops, Sorry I rebased locally without push. Thank you for pointing out! Done.

chunit-quic:
Ops, Sorry I rebased locally without push. Thank you for pointing out! Done.

cccclai:
There seems to be a lint error, and that might be reason merging fails. Since https://github.com/pytorch/executorch/pull/11049 is landed, can you apply the according changes and see if the error is gone?

chunit-quic:
> There seems to be a lint error, and that might be reason merging fails. Since #11049 is landed, can you apply the according changes and see if the error is gone?

Fixed. Thanks a lot for pointing out the error!

kimishpatel:
> > Mind sharing a bit more details for the request? I probably miss it
> 
> No problem. :) We received some feature requests in a mail thread regarding quantization requirements. This particular request is for the following purpose:
> 
> > The QNN Quantizer should allow users to override quantization parameters for specific tensors, regardless of the data ranges observed during calibration or QAT. This override must respect the transitive closures established by SharedQuantizationSpec.
> 
> After a brief discussion your team, we concluded that the rewriting stage should occur after calibration but before conversion. That's essentially the background.

I dont quite follow why we cannot do this by allowing module specific customization? FOr example can I not say `quanitzer.set_module_qconfig(....)`. And @jerryzh168 I really think we should make this part of the base class. Quantizers that dont implement this will fallback to baseclass which will just raise error.

PR #10095, opened 2025-04-11
Use certifi certs for buck download

GregoryComer:
### Summary
Depending on how Python is installed and managed, it is common for certificates to not be installed out of box on Macs. This can cause the download of buck2 as part of the build process to fail with a certificate error.

While the ultimate solution for this is for the user to install certs in their environment, we can smooth this over a bit for build by pointing urllib at the certifi package certs, which we install as part of install_requirements.

### Test plan
I worked with @nil-is-all to verify that this fix allowed him to build and install ExecuTorch from a clean environment. Without this change, he was seeing cert failure errors during buck download.

### DIFF
--- a/tools/cmake/resolve_buck.py
+++ b/tools/cmake/resolve_buck.py
@@ -9,6 +9,7 @@
 import argparse
 import os
 import platform
+import ssl
 import stat
 import sys
 import urllib.request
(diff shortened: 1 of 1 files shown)


GregoryComer:
Link lint job failures appear to be pre-existing.

jathu (approved):

larryliu0820 (approved):

PR #10098, opened 2025-04-11
Qualcomm AI Engine Direct - Mimi Enablement Stage 2

winskuo-quic:
### Summary

- Support Mimi Encoder
- Support Mimi Decoder
- Support OP: CDist
- Workaround for 0D tensor scenario since QNN does not support 0D tensor

Commands to Execute the model
10 sec as a batch (1 inference for 10 sec audio):
`python examples/qualcomm/oss_scripts/moshi/mimi.py -b build-android -s $DEVICE -m SM8650 --chunks_per_batch 125`

80s as a batch (125 inferences for 10 sec audio)
`python examples/qualcomm/oss_scripts/moshi/mimi.py -b build-android -s $DEVICE -m SM8650`

### Test plan

- UT for CDist
- UT for 0D tensor


winskuo-quic:
Hi @cccclai, @iseeyuan, @billmguo,
This PR is to enable mimi without streaming mode.
I have listed to execution commands in the summary section.
Please have a look.
Thanks

backends/qualcomm/_passes/remove_0d_tensor.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class Remove0DTensor(ExportPass):
```
cccclai, line 12: Does the logic in https://github.com/pytorch/executorch/blob/main/backends/transforms/rank_0_to_rank_1.py can be applied here?

winskuo-quic: Hi @cccclai,
Thanks for reviewing the PR.
I believe the pass you are suggesting is to change input from 0d to 1d tensor.
However, for our case, this 0D tensor happened during `select` op in the middle of graph. We just removed this `select` op since it does not affect the logic of the graph.
The exact point where 0D tensor occurs in mimi is under `moshi/quantization/core_vq.py`, where is tries to create a 1D tensor and retrieve index 0.
image

cccclai: I see, I guess for this specific logic, it's more efficient to remove the select op. 

A more generic way to handle 0-d tensor is to convert it to 1-d tensor. Like following 
```
class Rank0ToRank1Pass(ExportPass):
    """
    Replace Rank-0 Tensor to Rank-1 Tensor for all the inputs.
    """

    def __init__(self) -> None:
        super().__init__()

    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
            // we can make the pass configurable to decide what kind of node we want to reshape to 1-d tensor
            if node.op == "call_function" and node.meta["val"].shape == ():
                node.meta["val"] = node.meta["val"].reshape(1, 1)
        graph_module.recompile()
```

cccclai (approved):
Thank you for enabling mimi!

PR #10106, opened 2025-04-11
Arm backend: Add check to not partition ops with float64 input

YufengShi-dudu:
- Float64 placeholders are not supported in the Arm backend. They will cause a crash when processed in the process_placeholder function. This patch rejects Float64 placeholders early to prevent crashes during the partition.

Change-Id: I7e7c61836a7a50f29c18252819cf5e537f3b595a


zingo:
MacOS fails unrelated

zingo (approved):

zingo (approved):

PR #10135, opened 2025-04-14
Arm backend: Add support for TOSA 1.0 serializer

per:
Adapt serialization and TOSA graph handling to be able to handle 1.0. Also install TOSA pip package for 1.0 alongside 0.80.

### Test plan
Validate that 0.80 TOSA version test still work with the 1.0 package installed.

### DIFF
--- a/backends/arm/tosa_backend.py
+++ b/backends/arm/tosa_backend.py
@@ -13,7 +13,7 @@
 import logging
 from typing import cast, final, List
 
-import tosa_tools.v0_80.serializer.tosa_serializer as ts  # type: ignore
+import executorch.backends.arm.tosa_specification as tosa_specification
 
 from executorch.backends.arm.arm_backend import get_tosa_spec
 from executorch.backends.arm.operators.node_visitor import get_node_visitors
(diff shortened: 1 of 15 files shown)


zingo (approved):

backends/arm/tosa_backend.py
```diff
@@ -88,7 +88,22 @@ def preprocess(  # noqa: C901
 
         # Converted output for this subgraph, serializer needs path early as it emits
         # const data directly. Path created and data written only in debug builds.
+        if isinstance(tosa_spec, tosa_specification.Tosa_0_80):
+            import tosa_tools.v0_80.serializer.tosa_serializer as ts  # type: ignore
+        elif isinstance(tosa_spec, tosa_specification.Tosa_1_00):
+            import serializer.tosa_serializer as ts  # type: ignore
```
digantdesai, line 94: hmm, might need to revert, keep you posted.

digantdesai, line 94: IIUC in the future, the serializer will support older versions of TOSA, and we will delete the v0.80 serializer, right? If yes, what is the ETA?

per: Yes, the 0.80 should be deleted when 1.0 has landed and is stable.

PR #10137, opened 2025-04-14
Arm backend: Remove no-op repeat nodes in ConvertExpandCopyToRepeatPass

martinlsm:
Update ConvertExpandCopyToRepeatPass to eliminate Repeat nodes where all repetition counts are one, as these are functionally equivalent to a simple copy. This reduces unnecessary operations in the graph.

This patch affects two unit tests in
backends/arm/test/ops/test_expand.py. The models (test subjects) in those unit tests are in the affected cases optimized such that the only computation node (torch.Tensor.expand) they each contain is optimized away by the modified ConvertExpandCopyToRepeatPass. This will make Vela complain saying that the model is empty and the TOSA file cannot be compiled. Set the affected unit tests to xfail since they are expected to be resolved later in another patch.

### DIFF
--- a/backends/arm/_passes/convert_expand_copy_to_repeat.py
+++ b/backends/arm/_passes/convert_expand_copy_to_repeat.py
@@ -1,15 +1,17 @@
-# Copyright 2024 Arm Limited and/or its affiliates.
-# All rights reserved.
+# Copyright 2024-2025 Arm Limited and/or its affiliates.
 #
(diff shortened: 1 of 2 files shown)


oscarandersson8218:
@martinlsm Linting issues need to be addressed. Apart from that it looks good to me.

zingo (approved):

PR #10144, opened 2025-04-14
Runtime API to retrieve attributes

JacobSzwejbka:
Summary: Scan implementation today which is kind of slow. I dont think people wil regularly be scanning for tons of tensors though so perf is not super sensitive. Can always add a cache later if necessary.

Differential Revision: D72802654

cc @larryliu0820 @lucylq

### DIFF
--- a/runtime/executor/method.cpp
+++ b/runtime/executor/method.cpp
@@ -1593,6 +1593,37 @@ EValue& Method::mutable_input(size_t i) {
   return mutable_value(get_input_index(i));
 }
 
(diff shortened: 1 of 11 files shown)


larryliu0820 (approved):

JacobSzwejbka:
@cptspacemanspiff We ended up pursuing this approach over the memory planning one to share state between entry points. I think this one is easier to maintain (fewer foot guns setting it up as well), though we can continue supporting both approaches.

PR #10145, opened 2025-04-14
Add error message for empty string in filedataloader.

JacobSzwejbka:
Summary: Special case empty string/ nullpointer to provide a better error message in that case.

Differential Revision: D72981574


cccclai (approved):
Seems reasonable

PR #10155, opened 2025-04-14
import complex.h from c10

manuelcandales:
Summary: Import complex type and utils from c10 in order to do arithmetic with complex numbers in ExecuTorch kernels.

Reviewed By: digantdesai

Differential Revision: D72197508


swolchok (approved):
seems legit as long as the c10 headers are exact copies

I thought this would be bc-breaking; didn't realize that real_ and imag_ on c10::complex are public.

PR #10162, opened 2025-04-14
Move benchmarking workflow cli from testinfra to executorch

Camyll:
Request from Executorch team to run android-perf or apple-perf workflows via commandline

https://github.com/pytorch/executorch/actions/workflows/android-perf.yml
https://github.com/pytorch/executorch/actions/workflows/apple-perf.yml

HOW TO:

1. Be in repo pytorch/executorch. Navigate to extension/benchmark folder.
2. Make sure you have your GITHUB_TOKEN set (classic personal access token, not fine-grained) (export GITHUB_TOKEN=<YOUR_TOKEN>)
3. run python benchmark_workflow_cli.py with either --android or --ios as a required argument
4. use --help to see other optional arguments (--branch, --models, --devices, --benchmark-configs)

cc @guangy10 @huydhn @kirklandsign @shoumikhin


extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+def parse_args() -> Any:
+    parser = ArgumentParser(
+        "Run Android and iOS tests on AWS Device Farm via github actions workflow run"
+    )
+
+    parser.add_argument(
+        "--branch",
+        type=str,
+        default="main",
```
guangy10, line 25: This arg must be set by the user explicitly. Since this script is primarily for developers, most likely it should run on a developer's branch instead of main.

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        required=False,
+        help="what gh branch to use in pytorch/executorch",
```
guangy10, line 27: We should make it clear that the branch must be non-fork, otherwise it won't work

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+            "google_pixel_8_pro",
+        ],
+        help="specific devices to run on. Default is s22 for android and iphone 15 for ios.",
+    )
+
+    parser.add_argument(
+        "--benchmark_configs",
+        type=str,
+        required=False,
+        choices=["xplat", "android", "ios"],
```
guangy10, line 71: This doesn't look correct.

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        action="store_true",
+        required=False,
+        help="run the test on iOS",
+    )
+
+    parser.add_argument(
+        "--models",
+        type=str,
+        required=False,
+        default="llama",
```
guangy10, line 48: Similar to branch. Users should specify what model(s) they want to benchmark with. Do not provide default seems better to me

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        help="the model to run on. Default is llama. See https://github.com/pytorch/executorch/blob/0342babc505bcb90244874e9ed9218d90dd67b87/examples/models/__init__.py#L53 for more model options",
```
guangy10, line 49: It could be a valid hugging face model_id as well, e.g. ["meta-llama/Llama-3.2-1B"](https://huggingface.co/meta-llama/Llama-3.2-1B)

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        action="store_true",
+        required=False,
+        help="run the test on Android",
+    )
+    app_type.add_argument(
+        "--ios",
+        action="store_true",
+        required=False,
+        help="run the test on iOS",
+    )
```
guangy10, line 42: nit: to be consistent, can we rename it to "platform"?

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+    args, unknown = parser.parse_known_args()
+    if len(unknown) > 0:
+        info(f"detected unknown flags: {unknown}")
+    return args
+
+
+def run_workflow(app_type, branch, models, devices, benchmark_configs):
+    dispatch_hook = "/dispatches"
+    if app_type == "android":
```
guangy10, line 84: platform

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        url = f"https://api.github.com/repos/pytorch/executorch/actions/workflows/android-perf.yml"
+    else:
```
guangy10, line 86: Should error out explicitly if a not supported platform

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        url = f"https://api.github.com/repos/pytorch/executorch/actions/workflows/apple-perf.yml"
+
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"Bearer {GITHUB_TOKEN}",
+        "X-GitHub-Api-Version": "2022-11-28",
+    }
```
guangy10, line 93: What are these? Maybe add some comment to explain a bit

extension/benchmark/benchmark_workflow_cli.py
guangy10: Also, if multiple models are allowed, you should explain how users should specify it in that case

extension/benchmark/benchmark_workflow_cli.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
```
guangy10, line 1: Suggest moving this script to `extension/benchmark/scripts`, and rename the script to something more concise

guangy10 (changes requested):
Do not forget to include the instructions to the `README.md`

extension/benchmark/benchmark_workflow_cli.py
Camyll: In the GH workflow there is a default value (main) which is why I didn't require this argument. Should I make it default to the user's branch?

extension/benchmark/benchmark_workflow_cli.py
Camyll: The language in the github workflow implied that this parameter is a list of strings. Does that seem correct? That's why I had it as the names of the lists rather than the values themselves

Camyll: image

extension/benchmark/benchmark_workflow_cli.py
Camyll: Leaving it with no default for now since I couldn't use the GH utils to grab the branch name in this folder

extension/benchmark/benchmark_workflow_cli.py
Camyll: changed to list

yangw-dev (approved):

Camyll:
@guangy10 can you remove changes requested as the PR has been updated? Thanks!

huydhn (approved):

huydhn:
There are lots of unrelated failures, so you probably want to rebase it and also fix lint before landing (I could help with that)

PR #10192, opened 2025-04-15
Experiment with private rooted Pixel 3 devices

huydhn:
I'm creating a new experimental workflow to run benchmarks on private rooted Pixel 3 devices.  We have only 2 devices there though, so I think let's reduce the number of model we want to cover for now.

### Testing

https://github.com/pytorch/executorch/actions/runs/14465442714

### DIFF
--- a/.github/workflows/android-perf-private-device-experiment.yml
+++ b/.github/workflows/android-perf-private-device-experiment.yml
@@ -1,4 +1,4 @@
-name: android-perf (private devices)
+name: android-perf-private-devices
 
 on:
   schedule:
--- a/.github/workflows/android-perf-private-device-experiment.yml
+++ b/.github/workflows/android-perf-private-device-experiment.yml
@@ -1,4 +1,4 @@
-name: android-perf-private-devices
+name: android-perf (private devices)
 
 on:
   schedule:
--- a/.github/workflows/android-perf-private-device-experiment.yml
+++ b/.github/workflows/android-perf-private-device-experiment.yml
@@ -46,7 +46,7 @@ on:
         type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref_name }}-${{ github.ref_type == 'branch' && github.sha }}-${{ github.event_name == 'workflow_dispatch' }}-${{ github.event_name == 'schedule' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/android-perf-private-device-experiment.yml
```diff
@@ -0,0 +1,62 @@
...
+
+jobs:
+  android:
+    uses: ./.github/workflows/android-perf.yml
+    secrets: inherit
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      models: ${{ inputs.models }}
```
huydhn, line 60: @guangy10 @kirklandsign Appreciate if you could pass me the list of models and benchmark configs that you want to cover here. We have only 2 devices there though, so I think let's reduce the number of model we want to cover for now.

guangy10: Let's pick one non-genAI model (mv3, xnnpack_q8) and one genAI model (llama3.2-1b, spinquant & qlora)

mv3,meta-llama/Llama-3.2-1B-Instruct-SpinQuant_INT4_EO8,meta-llama/Llama-3.2-1B-Instruct-QLORA_INT4_EO8

.github/workflows/android-perf-private-device-experiment.yml
```diff
@@ -0,0 +1,62 @@
...
+      devices: google_pixel_3_private_rooted
```
guangy10, line 61: Unfortunately we don't have data points of using the same type of device on public pool, can't do any comparison, so we would mainly rely on the new data on the private & rooted device and compare those across different runs.

guangy10 (commented):
I recall @kirklandsign mentioned that the metrics instability is frequently seen on nightly batched jobs, on-demand with only 1-2 models are running much stable. 

@huydhn In this new schedule, and to gather more data points, we may want to run this workflow more often (every 2 hours maybe? Should have enough time to cooldown from previous run)

@huydhn Do we need to merge this PR? I thought we can reuse the existing workflow instead of creating a new fork

huydhn:
> @huydhn Do we need to merge this PR? I thought we can reuse the existing workflow instead of creating a new fork

This uses the existing workflow but gives us the flexibility to use a different schedule and default inputs

guangy10 (approved):

PR #10208, opened 2025-04-15
Implement _fft_c2r core ATen op

pssrawat:
Summary: Add ff2_c2r

Differential Revision: D73006888

### DIFF
--- a/kernels/optimized/cpu/op_fft_c2r.cpp
+++ b/kernels/optimized/cpu/op_fft_c2r.cpp
@@ -10,38 +10,35 @@
 #include <executorch/runtime/core/span.h>
 
 namespace torch::executor::native {
-Tensor& opt_fft_r2c_out(
+Tensor& opt_fft_c2r_out(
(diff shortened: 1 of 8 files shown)


kernels/optimized/cpu/op_fft_c2r.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <executorch/runtime/core/span.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+#include <pocketfft_hdronly.h>
+
+#include <optional>
+
+namespace torch::executor::native {
+
+// TODO: contents of this anonymous namespace are copy/pasted from
```
swolchok, line 18: A lot of this looks copy/pasted from op_fft_r2c.cpp . Please don't do that; instead, create a header that both op_fft_r2c.cpp and this file can include.

swolchok (changes requested):

swolchok (approved):
seems OK now

PR #10214, opened 2025-04-15
Update size threshhold

JacobSzwejbka:
Summary: These are the current actual threshholds

Differential Revision: D73056265


JacobSzwejbka:
Do these numbers look like what you expect after your size work @swolchok ?

lucylq (approved):
thanks for updating these!

swolchok:
> Do these numbers look like what you expect after your size work @swolchok ?

no particular expectations.

> Im seeing a pretty huge regression from https://github.com/pytorch/executorch/actions/runs/14476679027/job/40603913130?pr=10144
> 
> and I have no idea why.

I recommend using bloaty.

Issue #8728, opened 2025-02-26
Fix stable/ links in readmes for moved pages


GregoryComer:
### 📚 The doc issue

With the new documentation flow, I've moved, renamed, or split out a number of docs. There are hard-coded links in various readme files in the ET repo to stable/ docs. These are tricky to handle, because we can't easily relative link from readmes in the repo to docs (it likely should point to stable). So if we point to the new locations, it's broken now. If we don't, it breaks on the next release (when the new docs are promoted to stable).

To solve this, I see two options:
1. Fix the links during the 0.6 release process.
2. Set up redirects to handle the main moved pages (getting-started-setup being the most common one).

I'm leaning towards (2), then fixing the links permanently after 0.6 hits stable.

cc @mergennachin @byjlw

GregoryComer:
[#10221](https://github.com/pytorch/executorch/pull/10221) is merged. I'm cherry-picking this onto the 0.6 release branch and can close this out once that lands.

PR #10221, opened 2025-04-16
Add redirects for relocated docs

GregoryComer:
### Summary
We've moved a number of doc pages in 0.6. This PR adds the sphinx-reredirects extension, which adds a sphinx feature to define redirected doc pages. I've used this to declare redirects for key relocated doc pages to preserve SEO and incoming external links.

Solves #8728.

### Test plan
I've locally built and served the docs and navigated to each of the redirected URLs. I confirmed that all redirect as expected.

cc @mergennachin @byjlw

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -62,6 +62,7 @@ extensions = [
     "myst_parser",
     "sphinx_design",
     "sphinx_gallery.gen_gallery",
+    "sphinx_reredirects",
 ]
 
 if not FBCODE:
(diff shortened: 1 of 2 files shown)


lucylq (approved):

docs/source/conf.py
```diff
@@ -195,6 +196,24 @@
...
+    "tutorials/export-to-executorch-tutorial": "../using-executorch-export.html",
+    "running-a-model-cpp-tutorial": "using-executorch-cpp.html",
+    "build-run-vulkan": "backends-vulkan.html",
+    "executorch-arm-delegate-tutorial": "backends-arm-ethos-u.html",
+    "build-run-coreml": "backends-coreml.html",
+    "build-run-mediatek-backend": "backends-mediatek.html",
+    "build-run-mps": "backends-mps.html",
+    "build-run-qualcomm-ai-engine-direct-backend": "backends-qualcomm.html",
+    "build-run-xtensa": "backends-cadence.html",
+    "apple-runtime": "using-executorch-ios.html",
```
shoumikhin, line 213: Not sure we have any of these left

docs/source/conf.py
```diff
@@ -195,6 +196,24 @@
     "torch": ("https://pytorch.org/docs/stable/", None),
 }
 
+# Redirects for moved pages
+redirects = {
```
shoumikhin, line 200: Perhaps a simple find\replace would work as well?

docs/source/conf.py
GregoryComer: Yeah, the goal of this change is to make these old links point to the new pages. That way any external links (and SEO) will continue to work.

docs/source/conf.py
```diff
@@ -195,6 +196,24 @@
...
+    "running-a-model-cpp-tutorial": "using-executorch-cpp.html",
+    "build-run-vulkan": "backends-vulkan.html",
+    "executorch-arm-delegate-tutorial": "backends-arm-ethos-u.html",
+    "build-run-coreml": "backends-coreml.html",
+    "build-run-mediatek-backend": "backends-mediatek.html",
+    "build-run-mps": "backends-mps.html",
+    "build-run-qualcomm-ai-engine-direct-backend": "backends-qualcomm.html",
+    "build-run-xtensa": "backends-cadence.html",
+    "apple-runtime": "using-executorch-ios.html",
+    "tutorials/devtools-integration-tutorial": "../using-executorch-troubleshooting.html",
```
Gasoonjia, line 214: I don't think this is correct. original path is correct, it will not show up until running `tutorial_source/devtools-integration-tutorial.py`. Let me fix that in my PR.

Gasoonjia: yes your redirection is working, but the page you redirect to is not the one we need.

this is the one you redirected to: https://pytorch.org/executorch/0.6/using-executorch-troubleshooting.html
but this is the one we need: https://pytorch.org/executorch/0.6/tutorials/devtools-integration-tutorial.html

GregoryComer: Sure. Feel free to update the redirect link if you prefer that page.

Issue #9341, opened 2025-03-17
RFC: Decoder only LLM runner API


larryliu0820:
## **Objective**

Design and implement a set of C++ and Python APIs for running decoder-only text-only LLMs that closely resemble HuggingFace transformers' interfaces. The APIs will serve as an entry point for users who have already exported `llm.pte` and have corresponding `tokenizer.model` artifacts ready, using the `extension.llm` Python APIs, and will allow users to implement custom KV cache managers.

## **Background**

ExecuTorch already provides an `extension.llm` module with Python APIs for exporting LLM models to ET format (`llm.pte`). ExecuTorch also provides simple C++ runtime components such as<code> [text_token_generator.h](https://github.com/pytorch/executorch/blob/main/extension/llm/runner/text_token_generator.h)</code> and <code>[text_decoder_runner.h](https://github.com/pytorch/executorch/blob/main/extension/llm/runner/text_decoder_runner.h)</code> However, there is currently no standardized API for running these exported models, also it’s not easy for a user to close the gap between using Python AOT export API to writing C++ code to actually run the pte files.

Currently, users need to implement custom runners using the low-level `IRunner` interface defined in `irunner.h`. This requires significant boilerplate code and deep understanding of the PyTorch Edge internals. Additionally, KV cache management is currently handled by the `StaticAttentionIOManager` class, which provides limited flexibility for custom implementations.

## **Motivation**

1.  **Ease of adoption**: By providing APIs that resemble HuggingFace transformers, we can lower the barrier to entry for users already familiar with that ecosystem.
1.  **Standardization**: A consistent API will make it easier to share code and examples across the community.
1.  **Extensibility**: Allowing users to implement custom KV cache managers in both Python for quick iteration speed and C++, for enabling optimization for specific hardware or use cases.
1.  **Python integration**: Adding Python bindings will make the APIs accessible to a wider audience and enable seamless integration with existing Python workflows.

## **Overview**

The proposed design consists of the following components:

*  **C++ API**: A set of C++ classes that provide a HuggingFace-like interface for loading and running LLMs.
    1. `DecoderLLMRunner` class with `load()` static method
    2. `GenerationConfig` class for configuring generation parameters
    3. `IIOManager` interface for custom KV cache implementations
*  **Python Bindings**: Python bindings for the C++ API using pybind11.
    1. Python `DecoderLLMRunner` class with `load()` method
    2. Python `GenerationConfig` class
    3. Python `IIOManager` interface
[...]

PR #10228, opened 2025-04-16
Introduce GenerationConfig

### LINKED ISSUE #8495: [etLLM] extension/llm should have unit tests

TSIA

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @mergennachin @cccclai @helunwencser @jackzhxng

larryliu0820:
Summary:
Started to implement https://github.com/pytorch/executorch/issues/9341
Started to fix https://github.com/pytorch/executorch/issues/8495

This PR introduces `GenerationConfig` which contains the configs that can be changed across different invocations of `generate()`. 

For example, `temperature` is moved out from the runner constructor for it's not tied to the runner instance but instead should be adjustable every time we call `generate()`.

Similarly we put `echo` and `warming` into the config.

We also allow both `seq_len` and `max_new_tokens` to be passed through the config and we determine the value of `max_new_tokens` based on these 2 config values, pte file metadata as well as the number of prompt tokens.

Differential Revision: D73091676


examples/models/llama/main.cpp
```diff
@@ -73,13 +73,14 @@ int32_t main(int32_t argc, char** argv) {
   }
 #endif
   // create llama runner
-  example::Runner runner(model_path, tokenizer_path, temperature);
+  example::Runner runner(model_path, tokenizer_path);
 
   if (warmup) {
-    runner.warmup(prompt, seq_len);
+    runner.warmup(prompt, /*max_new_tokens=*/seq_len);
```
iseeyuan, line 79: Would it be added in the internal runner as well?

extension/llm/runner/irunner.h
```diff
@@ -6,42 +6,114 @@
...
 
+// Configuration struct for generation parameters
+struct GenerationConfig {
+  // Temperature for sampling (higher = more random)
+  float temperature = 0.8f;
+
+  // Whether to echo the input prompt in the output
+  bool echo = true;
+
+  // Maximum number of total tokens
```
iseeyuan, line 33: More doc string on this: a limit that users can set. It must not be larger than max_context_len from the model. (something like that, if I understand it correctly).

iseeyuan (approved):

examples/mediatek/executor_runner/mtk_llama_runner.h
```diff
@@ -43,11 +43,9 @@ class MTKLlamaRunner : public executorch::extension::llm::IRunner {
   Error load();
   Error generate(
       const std::string& prompt,
-      int32_t seq_len = 128,
+      executorch::extension::llm::GenerationConfig config,
```
cccclai, line 46: Can we leave mtk code as is? It's currently in unstable stage given the new apk release and upcoming code change.

examples/models/llama/main.cpp
larryliu0820: which internal runner?

PR #10237, opened 2025-04-16
qnn runner: add memory consumption logging

nirvanagth:
Summary:
We want to log the memory consumption 
follow the llama runner implementation https://github.com/pytorch/executorch/blob/main/examples/models/llama/runner/runner.cpp#L187

Differential Revision: D72747341

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
+++ b/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
@@ -214,6 +214,11 @@ Error Runner::load() {
   bos_id_ = tokenizer_->bos_tok();
   eos_id_.insert(tokenizer_->eos_tok());
 
+  ET_LOG(
+      Info,
+      "RSS after loading model: %f MiB (0 if unsupported)",
+      executorch::extension::llm::get_rss_bytes() / 1024.0 / 1024.0);
+
   // create sampler
   sampler_ = std::make_unique<Sampler>(
       vocab_size_,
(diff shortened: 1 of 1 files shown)


nirvanagth:
cc @cccclai if you have suggestion how to properly log memory consumption on qnn

cccclai:
Hi team, I remember there are different ways to check memory, and for htp, we may need to profile memory in a specific way. Can you instruct us how to do this?

examples/qualcomm/oss_scripts/llama/runner/runner.cpp
```diff
@@ -214,6 +214,11 @@ Error Runner::load() {
   bos_id_ = tokenizer_->bos_tok();
   eos_id_.insert(tokenizer_->eos_tok());
 
+  ET_LOG(
+      Info,
+      "RSS after loading model: %f MiB (0 if unsupported)",
+      executorch::extension::llm::get_rss_bytes() / 1024.0 / 1024.0);
```
haowhsu-quic, line 220: If `libqnn_executorch_backend.so` will be linked by multiple processes simultaneously, measuring PSS will be closer to the real memory usage. But looks like `getrusage` has no direct way to get such info.
If the log could be recorded exclusively (no other qnn delegate APPs are running at the same time), maybe it would be fine.

nirvanagth: thanks for the feedback. I agree using PSS would be more accurate but I couldn't find example in the codebase to log it. I think since the log explicitly mentioned RSS and we are ok with it as long as it's overestimating rather than underestimating the memory footprint. do you mind approve this

cccclai: Does it reflect the memory usage for htp?

haowhsu-quic: I think we also need to take DMA buffers (`dmabuf_dump`) into consideration. Don't know if there is a way to do it programmatically.

PR #10248, opened 2025-04-16
Support dynamically quantized 2D convolutions

### LINKED ISSUE #9021: Support Dynamically Quantized Convolutions
We wish to have some initial support for Dynamically Quantized Convolutions. 

Let's first write a test to drive development of the feature. Let's add a test here first:
https://github.com/pytorch/executorch/blob/e9c2315aacaa36fbd5aa8fe060d798fdc3f08695/backends/xnnpack/test/ops/test_conv2d.py#L4

let's just do 2d convolutions for now. Take a look at how we test dqlinears in general: https://github.com/pytorch/executorch/blob/e9c2315aacaa36fbd5aa8fe060d798fdc3f08695/backends/xnnpack/test/ops/test_linear.py#L327

And let's try to add a test here. Since we're adding quantizer support, we should make sure that after 
```
.quantize()
.export()
```
we should check that a choose_q_param node is in the graph. Now, after we've added this test, when we run it with 
`python -m unittest backends.xnnpack.test.ops.test_conv2d....` It should fail because it can't find the choose_q_params node. Let's first start by enabling the quantizer to properly annotate convolutions.

https://github.com/pytorch/executorch/blob/e9c2315aacaa36fbd5aa8fe060d798fdc3f08695/backends/xnnpack/quantizer/xnnpack_quantizer.py#L266

Since we're first starting with conv2d, we should only annotate dynamically quantized convs if they are 2d. We can add a check that the len(outputpadding) == 2 somewhere here:
https://github.com/pytorch/executorch/blob/main/backends/xnnpack/quantizer/xnnpack_quantizer_utils.py#L295

Now that we have it annotated, it should bass through the test that's checking for the choose_q_param. Now we just need to update our partitioner to allow DynamicallyQuantizedConvolutions:

https://github.com/pytorch/executorch/blob/e9c2315aacaa36fbd5aa8fe060d798fdc3f08695/backends/xnnpack/partition/config/gemm_configs.py#L396

Again it would be nice to check in our constraints that if we detect a dynamically quantized convolution, and it is 1d, then we don't partition. After this the test should be passing. There may be some more lingering issues with the wiring, if that's the case feel free to reach out in the discord group:

https://discord.com/channels/1334270993966825602/1336777807509979188

cc @digantdesai @cbilgin

keyprocedure:
### Summary
Add initial support for dynamically quantized Conv2d in XNNPACK:
- Add `conv` to `DYNAMIC_OPS` for annotation
- Update partitioner to support dynamically quantized Conv2d
- Add checks to ensure only 2D, non-depthwise dynamically quantized convs are partitioned and annotated
- Update NHWC permute node insertion to trace back to original input for dynamically quantized inputs
- Compute `num_nonbatch_dims` based on whether the node feeds into a conv
- Remove the `num_nonbatch_dims` check from XNNCompiler
- Add unit tests for channels-last permute and single, sequential, and parallel dynamically quantized 2D convs 

Fixes #9021 

### Test plan
```bash
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d_seq
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d_parallel
python -m unittest backends.xnnpack.test.passes.test_channels_last_tagged_reshape.TestChannelsLastTaggedReshapePass.test_dq_conv2d_channels_last_tagged_reshape_pass

### CODE REVIEW

File: backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py [hunk 1]
```diff
@@ -283,14 +284,26 @@ def input_to_nhwc(
             ]
         else:
             # Need to create NHWC node
+            # Check if input uses dynamic quantization
```
mcr229, line 287: nice! the change made here looks simplier than i expected. Though i suspect it was still pretty hard to navigate and figure out.

Do you mind adding a new test here as well:
https://github.com/pytorch/executorch/blob/d7f74bd4adf10950224d2d975a6e23e92e7be6f3/backends/xnnpack/test/passes/test_channels_last_tagged_reshape.py

Essentially we just check if there is a dynamically quantized convolution, then we place one permute before the dynamic_q chain and one permute after the conv.

backends/xnnpack/operators/quant_params.py
```diff
@@ -145,9 +145,14 @@ def quantize_tensor(self, tensor: torch.Tensor) -> torch.Tensor:
...
         q_input = quant_node.args[0]  # fp32 input
         assert isinstance(q_input, torch.fx.Node)
-        # TODO - materialize this from the quant_node scale count and val shape
         num_nonbatch_dims = 1
 
+        # Compute non-batch dimensions (shape length - 1), defaulting to 1
+        q_input_val = q_input.meta.get("val", None)
+        q_input_shape = getattr(q_input_val, "shape", None)
+        if q_input_shape is not None:
+            num_nonbatch_dims = max(len(q_input_shape) - 1, 1)
```
mcr229, line 154: this isn't necessarily the case for linear layers, the input to these will always have 1 nonbatch dimension.:
(x, y ,z, input_channels). The rank of the tensor can be arbitrary, and we always interpret the last dimension as input channels, and every other dimension as a batch dimension.

for now let's just add a check if convolution then 3 if linear then 1. This issue stems from the fact that we are injecting per_tensor quant nodes, but our intent is to do per_batch. We are adding in affine quantization which should tell us how many batch and how many non-batch dimensions there are in the quant node, so later on we will fix it to use that, but for now, it might just be hard code the conv and linear case :(

backends/xnnpack/quantizer/xnnpack_quantizer_utils.py
```diff
@@ -323,6 +323,19 @@ def _do_annotate_conv(
...
+        if (
+            quantization_config
+            and quantization_config.input_activation
+            and quantization_config.input_activation.is_dynamic
+        ):
+            weight_val = weight.meta.get("val", None)
+            weight_shape = getattr(weight_val, "shape", None)
+
+            # Skip if not a 4D weight tensor (i.e. not conv2d)
+            if weight_shape is not None and len(weight_shape) != 4:
```
mcr229, line 336: nice, let's also add a skip if the convolution is depthwise, since XNNPACK can't handle that case yet.

backends/xnnpack/runtime/XNNCompiler.cpp
```diff
@@ -1172,7 +1167,7 @@ Error defineStaticTransposeNode(
   ET_CHECK_OR_RETURN_ERROR(
       status == xnn_status_success,
       Internal,
-      "Failed to create sigmoid node %i with code: %s",
+      "Failed to create static transpose node %i with code: %s",
```
mcr229, line 1170: thanks :)

backends/xnnpack/test/ops/test_conv2d.py
```diff
@@ -169,6 +173,20 @@ def get_inputs(self):
         return (torch.randn(2, 2, 4, 4),)
 
 
+class Conv2dDynamicQuant(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv = torch.nn.Conv2d(3, 10, 3)
+        self.conv.weight.requires_grad = False
+        self.conv.bias.requires_grad = False
```
mcr229, line 181: is this necessary?

backends/xnnpack/test/ops/test_conv2d.py
```diff
@@ -169,6 +173,20 @@ def get_inputs(self):
         return (torch.randn(2, 2, 4, 4),)
 
 
+class Conv2dDynamicQuant(torch.nn.Module):
```
mcr229, line 176: can you just add two more, where we have:
1. two convolutions in sequence: `inp --> conv --> conv --> out`
2. two convolutions running in parallel: 
```
inp1 --> conv --> out1
   \
     --> conv2 --> out2
```

backends/xnnpack/partition/config/gemm_configs.py
```diff
@@ -358,6 +358,11 @@ def check_constraints(self, node: torch.fx.Node, ep: ExportedProgram) -> bool:
             why(node, "Only support 1D + 2D Conv")
             return False  # Only support 1D + 2D Conv
 
+        precision = self._detect_precision(node)
+        if precision == ConfigPrecisionType.DYNAMIC_QUANT and len(conv_stride) != 2:
```
mcr229, line 362: can you also add the depthwise check here?

mcr229 (changes requested):
Overall looks really good. Just a few minor comments.

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
keyprocedure: The time to lines of code written ratio was definitely off the charts haha 

Test added in `test_channels_last_tagged_reshape.py`

backends/xnnpack/operators/quant_params.py
keyprocedure: Good to know. I added a check to determine if the node feeds into a conv and set the non-batch dimensions to 3 if it does

backends/xnnpack/test/ops/test_conv2d.py
keyprocedure: Good catch, removed

backends/xnnpack/test/ops/test_conv2d.py
keyprocedure: Added the two unit tests

backends/xnnpack/partition/config/gemm_configs.py
keyprocedure: Added the depthwise check. If it looks good, I can move the check to a helper function since it's being used in `op_conv2d.py`, `gemm_configs.py`, and `xnnpack_quantizer_utils.py`

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
mcr229: yea, this is a good datapoint, I want to help improve the contributability flow for the XNNPACK Backend, and some of that definitely means improving/refactoring passes like this which are way too complex.

backends/xnnpack/quantizer/xnnpack_quantizer_utils.py
mcr229: i do believe the default is 1 if it is not an arg

backends/xnnpack/test/ops/test_conv2d.py
```diff
@@ -169,6 +173,55 @@ def get_inputs(self):
         return (torch.randn(2, 2, 4, 4),)
 
 
+class Conv2dDQ(torch.nn.Module):
```
digantdesai, line 176: nit
```suggestion
class Conv2d(torch.nn.Module):
```

backends/xnnpack/test/ops/test_conv2d.py
```diff
@@ -223,6 +276,37 @@ def _test(
...
+        dynamic_shapes=None,
+    ):
+        quant_config = get_symmetric_quantization_config(
+            is_per_channel=True,
+            is_dynamic=True,
+        )
+
+        DynamicallyQuantizedPartitioner = XnnpackPartitioner(
+            config_precisions=ConfigPrecisionType.DYNAMIC_QUANT,
+            per_op_mode=True,
```
digantdesai, line 292: Nit: also check without this?

backends/xnnpack/test/ops/test_conv2d.py
```diff
@@ -699,3 +783,15 @@ def forward(self, x):
...
             .run_method_and_compare_outputs(qtol=1)
         )
+
+    def test_dq_conv2d(self) -> None:
+        model = Conv2dDQ()
+        self._test_dq(model)
+
+    def test_dq_conv2d_seq(self) -> None:
+        model = Conv2dDQSeq()
+        self._test_dq(model, conv_count=2)
```
digantdesai, line 793: Nit: get conv count from the model rather than hardcoding

mcr229:
looks good, let's rebase and just let the CI run thank you!

digantdesai (approved):
Thank you!

PR #10250, opened 2025-04-17
Qualcomm AI Engine Direct - Fix the bug in rms_norm builder

shewu-quic:
As the title

### DIFF
--- a/backends/qualcomm/builders/op_rms_norm.py
+++ b/backends/qualcomm/builders/op_rms_norm.py
@@ -81,8 +81,9 @@ class RmsNormVisitor(NodeVisitor):
             {},  # kwargs
         )
         if quant_attrs := node.meta.get(QCOM_QUANT_ATTRS):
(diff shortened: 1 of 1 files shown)


cccclai:
It seems CI is passing, how do you find out the bug?

shewu-quic:
> It seems CI is passing, how do you find out the bug? What will be affected?

We checked the graph and noticed that the offset of rms_norm is set to zero after the qnn_partitioner. Upon further investigation, we discovered that we didn't copy the quant_attr for the bias node, which results in modifying the original node's quant_attr. I think CI pass due to the offset is zero for rms_norm in llama model. In fact, our internal CI also pass.

cccclai (approved):
Thanks for the fix!

cccclai:
> I think CI pass due to the offset as zero for rms_norm in llama model

Hmm, feel like it's quite unlikely to have zero offset for rms_norm in llama model..

PR #10258, opened 2025-04-17
NXP backend: Enable initial unit tests workflow

skywall:
### Summary
Initial workflow definition for running unit tests for Neutron (NXP) backend. 

### DIFF
--- a/backends/nxp/run_unittests.sh
+++ b/backends/nxp/run_unittests.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+set -eux
+
+SCRIPT_DIR=$(dirname $(readlink -fm $0))
+EXECUTORCH_DIR=$(dirname $(dirname $SCRIPT_DIR))
+
+cd $EXECUTORCH_DIR
+
+# '-c /dev/null' is used to ignore root level pytest.ini.
+PYTHONPATH=`cd ..; pwd` pytest -c /dev/null backends/nxp/tests/
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -716,3 +716,29 @@ jobs:
       build-mode: Release
       build-tool: cmake
       docker-image: executorch-ubuntu-22.04-clang12
+
+  unittest-nxp-neutron:
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-clang12
+      submodules: 'true'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
+      script: |
+        set -eux
+        
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+        
+        # Build and install Executorch
+        PYTHON_EXECUTABLE=python \
+        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON" \
+        .ci/scripts/setup-linux.sh --build-tool "cmake"
+        
+        # Run pytest
+        PYTHON_EXECUTABLE=python bash backends/nxp/run_unittests.sh


backends/nxp/run_unittests.sh
```diff
@@ -0,0 +1,14 @@
...
+# LICENSE file in the root directory of this source tree.
+set -eux
+
+SCRIPT_DIR=$(dirname $(readlink -fm $0))
+EXECUTORCH_DIR=$(dirname $(dirname $SCRIPT_DIR))
+
+cd $EXECUTORCH_DIR
+
+# '-c /dev/null' is used to ignore root level pytest.ini.
+PYTHONPATH=`cd ..; pwd` pytest -c /dev/null backends/nxp/tests/
```
digantdesai, line 14: Do we have to use `pytest`? Can we start with just the unit tests?

.github/workflows/pull.yml
```diff
@@ -513,6 +513,32 @@ jobs:
...
+      script: |
+        set -eux
+        
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+        
+        # Build and install Executorch
+        PYTHON_EXECUTABLE=python \
+        CMAKE_ARGS="-DEXECUTORCH_BUILD_PYBIND=ON -DEXECUTORCH_BUILD_NXP_NEUTRON=ON" \
```
digantdesai, line 536: Curious, why do we need to enable `PYBIND`?

digantdesai (changes requested):

backends/nxp/run_unittests.sh
skywall: We use pytest quite extensively - fixtures, mocking/spying, test parametrization, test skipping etc. Tests in Robert's PR are also based on pytest so it would require significant refactoring - https://github.com/pytorch/executorch/pull/10196/files#diff-5103e16cc6c48b276ba322c235272453e0442b6d19193200cd4a0e95706a605c . So if there's chance to stick with pytest, we would prefer it.

.github/workflows/pull.yml
skywall: Sorry, this is my misunderstanding of how `EXECUTORCH_BUILD_PYBIND` should be used. I have considered it as a "mechanism to build backend-specific dependencies", because I needed to build `portable_lib` and also `libquantized_ops_aot_lib`. Removed.

backends/nxp/run_unittests.sh
digantdesai: We are trying people to discourage using pytest because it is yet-another-thing while we already have unit-tests. Is it possible to slowly migrate away from this going forward? I.e. new tests won't use pytest features etc.?

skywall: Yes, we can go this way (new features covered by unittest). A bit sad to miss out on pytest's powers, but I understand the desire to avoid yet another dependency.

We still have some more code to upstream (beside Robert's PR) so there will be a few more PR's with pytest tests, but we're internally switching to unittest.

skywall: @digantdesai We have replaced pytest in out newest code, but some functionality (spying) has to be substituted by library "kgb" (https://pypi.org/project/kgb/). Is combination unittest+kgb preferred then using pytest?

robert-kalmar: @digantdesai , based on our conversation, let me refrase the question @skywall brings up. With abandoning pytest some functionality will get lost, e.g. spying. So, whether the future test shall be limited to unittest capabilities or additional packages like the kgb is OK. 
So essentially what limitation you envision for unit test creation?

digantdesai (approved):

.github/workflows/pull.yml
```diff
@@ -513,6 +513,32 @@ jobs:
         # Run pytest without simulator
         backends/arm/test/test_arm_baremetal.sh test_pytest
 
+  unittest-nxp-neutron:
```
digantdesai, line 516: On one quick follow up, can we move this to trunk.yaml instead of here given early days for the delegate, we don't want to disrupt the regular devx.

digantdesai (approved):

.github/workflows/pull.yml
digantdesai: yeah once we get more tests/mature backend we can make it on "every pr"

digantdesai: FYI we have to add a `ci: trunk` label to run these jobs.

digantdesai:
Running CI, once green I will merge. Thanks again.

skywall:
Rebased to include tests fix: https://github.com/pytorch/executorch/commit/e7d39c239879cf6e84d6d9db869a5d4c495b10ca

JakeStevens:
> @digantdesai Do you please have an idea why the CI fails? Looking at the logs of [trunk / unittest-nxp-neutron / linux-job (push)](https://github.com/pytorch/executorch/actions/runs/15049212213/job/42299479862?pr=10258) I don't see any problems with job definition itself, but I'm not sure why other similar jobs pass. Maybe some caching issue after moving from pull.yaml to trunk.yaml?

Error is here:

Error:
  2025-05-15 15:42:42,467 [ExecuTorch] ERROR: Failed to query buck for sources. Failed command:

     buck2 cquery inputs(deps('//examples/models/llama/runner:runner'))
     
     
 Seems like may be unrelated, might be fixed by now. I think the lintrunner issue is also resolved. So perhaps rebase and try again? I think should be green then

skywall:
Rebased. @JakeStevens Can you please re-run CI?

digantdesai:
Rerunning it, seems unrelated, sorry about that. If you want less "turbulence" I would say feel free to disable things you don't want to build through CMAKE_ARGS for ET

PR #10259, opened 2025-04-17
Replace third-party/pkg_resources:pkg_resources with third-party/pypi/setuptools:setuptools [4/11]

kkolur76:
Differential Revision: D73180095


manuelcandales (approved):

swolchok:
Can you provide more information (on the internal diff if not appropriate to share) about why this change is needed? I have no context here

PR #10274, opened 2025-04-17
[exir] Allow verifiers in _transform

digantdesai:
* #10276
* __->__ #10274

This is to allow users to transform an ExportedProgram using passes in places where it may result in a dialect that is not compliant with the original creation context.

For example, if an ExportedProgram was created in an edge dialect and now needs to be run and transformed in a way that is not compliant with the EdgeDialectVerifier, such as in a delegate preprocess() function, then the user may want to override the verifier with their own or simply disable it.

Differential Revision: [D73205727](https://our.internmc.facebook.com/intern/diff/D73205727/)


larryliu0820 (approved):
Can you add a unit test

PR #10276, opened 2025-04-17
[mps] Disable dialect verifier under mps preprocess

digantdesai:
* __->__ #10276
* #10274

As title. With reverting dim_order ops, we are producing an illegal IR, which is OK given we are inside MPS preprocess function which shouldn't adhere to edge_ir constraints.

Differential Revision: [D73205726](https://our.internmc.facebook.com/intern/diff/D73205726/)


cccclai (approved):
Seems reasonable

PR #10280, opened 2025-04-17
Fix undefined fht_float in Apple OS

derekxu:
Differential Revision: D73213645


larryliu0820:
Can you change the title and description to describe why you make this change?

derekxu:
> Can you change the title and description to describe why you make this change?

@larryliu0820 done

larryliu0820 (approved):

extension/llm/custom_ops/spinquant/third-party/FFHT/targets.bzl
```diff
@@ -19,6 +19,9 @@ def define_common_targets():
             "DEFAULT": [],
             "ovr_config//cpu:arm64": ["fht_neon.c"],
             "ovr_config//cpu:x86_64": ["fht_avx.c"],
+            "ovr_config//os:iphoneos": ["fht_avx.c"],
+            "ovr_config//os:macos-x86_64": ["fht_avx.c"],
+            "ovr_config//os:macos-arm64": ["fht_neon.c"],
```
swolchok, line 24: can you explain more about why this is necessary? the added stuff seems redundant with what is already there. also, in particular, fht_avx.c is not the correct implementation on arm64 iOS.

swolchok (changes requested):

PR #10285, opened 2025-04-17
[Executorch][llama] Allow custom sdpa op replacement pass to leverage attention mask

kimishpatel:
* #10286
* __->__ #10285
* #10284
* #10283

Previously we assumed that the custom sdpa always does causal attention.
This diff adds option to this module swap pass to make custom sdpa leverage
attention mask instead of causal.

Differential Revision: [D73222736](https://our.internmc.facebook.com/intern/diff/D73222736/)


examples/models/llama/source_transformation/sdpa.py
```diff
@@ -47,34 +63,54 @@ def forward(
...
-        output = torch.ops.llama.custom_sdpa(
-            q,
-            k,
-            v,
-            input_pos[0].item(),
-            None,  # Attention mask
-            0,  # dropout probability. Ignored by the code
-            True,  # is_causal
-        )
+        if self.use_attention_mask:
```
lucylq, line 66: nit, maybe this could be:

if not self.use_attention_mask:
    mask = None
    
output = torch.ops.llama.custom_sdpa(
                q,
                k,
                v,
                input_pos[0].item(),
                mask,  # Attention mask
                0,  # dropout probability. Ignored by the code
                False,  # is_causal
            )

lucylq (approved):

examples/models/llama/source_transformation/sdpa.py
kimishpatel: Note that is_causal is True in one case, while not true in the other

PR #10293, opened 2025-04-18
Support pre-quantization via torchao quantize_

metascroy:
Checkpoints saved with torchao quantized subclasses can be loaded with the PR

### DIFF
--- a/examples/models/llama/model.py
+++ b/examples/models/llama/model.py
@@ -18,6 +18,7 @@ from executorch.examples.models.checkpoint import (
 from executorch.examples.models.llama.llama_transformer import Transformer
 
 from executorch.examples.models.llama.model_args import ModelArgs
+from torchao.utils import TorchAOBaseTensor
 
 try:
     from .fairseq2 import convert_to_llama_checkpoint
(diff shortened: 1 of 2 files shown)


jackzhxng (commented):
Weren't these changes in another PR?

metascroy:
> Weren't these changes in another PR?

They were, but that PR had to be landed with a pin bump PR and a diff from torchao simultaneously due to some dependencies between them (especially for internal CI).  These changes were causing issues, so I decided to split them out into another PR once all the dust settled on the previous change.

jackzhxng (approved):

examples/models/llama/model.py
```diff
@@ -257,6 +258,9 @@ def __init__(self, **kwargs):
                 strict=False,
                 assign=True,
             )  # self.model_ = Transformer(gptconf)
+            for param in self.model_.parameters():
+                if isinstance(param, TorchAOBaseTensor):
+                    param.requires_grad = False
```
guangy10, line 263: Is this a temp hack? Can we make it not tie to the modeling code, better dealing with it inside one of our public API or making it a source to source transformation?

PR #10295, opened 2025-04-18
Exporting start_time in InstructionEvent to Inspector

juyoung-meta:
Summary:
ETDump profiled data does not include start_time information when
Inspector analyzes Events. Although InstructionEvent's ProfileEvent member
contains start_time and end_time information, they are used to only
elpased_time which is exposed to Inspect. However, without the information, we
are unable to visualize the time-series view of operator executions. This diff
is to expose the start_time to Inspector so that it can access the start_time
information.

Differential Revision: D72740782


Gasoonjia:
should be merged already. CLOSED

PR #10303, opened 2025-04-18
Refactor export_delegated_program

lucylq:
* #10330
* #10328
* #10327
* #10305
* __->__ #10303

Currently, we generate every combination of inputs for each module with the export_delegate_program script:
- extract_segments=True, False
- delegate_alignment=None,1024

Planning to add another flag, 'external_constants', which will move constants into a separate file to test program-data separation for delegated programs.

This test only requires pte, ptd, with default settings. So refactoring the export script to only generate based on the args, and update genrule to generate what the test requires.

Differential Revision: [D73278562](https://our.internmc.facebook.com/intern/diff/D73278562/)


swolchok (commented):
no general Python complaints; leaving accept for folks with context

JacobSzwejbka (approved):

PR #10310, opened 2025-04-18
Lint docs before building

shoumikhin:
Run link checks

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -14,18 +14,21 @@ on:
     - cron: '0 0 * * *'
 
 jobs:
-  lint-docs:
+  check-urls:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
       - name: Check URLs
         run: bash ./scripts/check_urls.sh
+
+  check-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kirklandsign (approved):

.github/workflows/doc-build.yml
```diff
@@ -14,7 +14,18 @@ on:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Check URLs
+        run: bash ./scripts/check_urls.sh
+      - name: Check Links
+        run: bash ./scripts/check_links.sh
+
   build:
+    needs: lint-docs
```
kirklandsign, line 28: Remove?

.github/workflows/doc-build.yml
```diff
@@ -14,7 +14,18 @@ on:
...
 jobs:
+  lint-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Check URLs
+        run: bash ./scripts/check_urls.sh
+      - name: Check Links
+        run: bash ./scripts/check_links.sh
```
kirklandsign, line 25: My suggestion is we can split into two jobs: check-urls and check-links. So we can run in parallel and get signals a little bit faster

.github/workflows/doc-build.yml
shoumikhin: You think we can still build the docs with broken links? Well, probably yes :)

PR #10325, opened 2025-04-21
New embedding quant fusion

metascroy:
Summary:
The diff adds new quant fusion passes to recognize 2, 4, and 8 bit quantized embeedings (per group and per channel) and fuses them to ExecuTorch kernels.  This makes torchao's quantize_ integrate with ExecuTorch:

```
 quantize_(
    model,
    IntxWeightOnlyConfig(weight_dtype=torch.int4, granularity=PerGroup(32)),
    lambda m, fqn: isinstance(m, torch.nn.Embedding)
)

# lower model to executorch
```

For the model to lower, we need to run QuantFusionPass.  For subbyte, we also need to run constant_prop_pass.  (See new unit tests for examples).  In follow-up diffs, we will enable these passes by default in to_executorch before the memory passing and out-variant passes.

Differential Revision: D73381542

### DIFF
--- a/exir/passes/_quant_patterns_and_replacements.py
+++ b/exir/passes/_quant_patterns_and_replacements.py
@@ -25,6 +25,8 @@ __all__ = [
 
 from torch import Tensor
 from torch.library import custom_op
+
+
 @custom_op("quant_fusion::_pack_embedding_weight", mutates_args=())
 def _pack_embedding_weight(weight: Tensor, bitwidth: int) -> Tensor:
     num_embeddings, embedding_dim = weight.shape
@@ -32,9 +34,7 @@ def _pack_embedding_weight(weight: Tensor, bitwidth: int) -> Tensor:
     if bitwidth == 2:
         assert embedding_dim % 4 == 0, "embedding_dim must be divisible by 4"
         weight_range_shifted = weight.add(2).view(torch.uint8)
-        weight_view = weight_range_shifted.view(
-            num_embeddings, embedding_dim // 4, 4
-        )
+        weight_view = weight_range_shifted.view(num_embeddings, embedding_dim // 4, 4)
         weight_0 = weight_view[:, :, 0]
         weight_1 = weight_view[:, :, 1] << 2
         weight_2 = weight_view[:, :, 2] << 4
@@ -53,7 +53,7 @@ def _pack_embedding_weight(weight: Tensor, bitwidth: int) -> Tensor:
         return packed_weight
     elif bitwidth == 8:
         return weight
-    
+
     raise RuntimeError(f"Unsupported bitwidth {bitwidth}")
 
 
@@ -64,7 +64,12 @@ def _(weight, bit_width):
     num_embeddings, embedding_dim = weight.shape
     values_per_byte = 8 // bit_width
     assert embedding_dim % values_per_byte == 0
-    return torch.empty(num_embeddings, embedding_dim // values_per_byte, dtype=torch.uint8, device=weight.device)
+    return torch.empty(
+        num_embeddings,
+        embedding_dim // values_per_byte,
+        dtype=torch.uint8,
+        device=weight.device,
+    )
 
 
 # TODO: extending an existing library that is defined in OSS might be a bit
@@ -114,9 +119,10 @@ def embedding_weight_checks(weight, weight_scales, weight_zero_points):
     assert (
         weight_zero_points is None or weight_zero_points.dtype == weight_scales.dtype
     ), "Expecting weight_zero_points to be None or have same dtype as weight_scales"
-    assert (
-        weight_zero_points is None or weight_zero_points.dim() in [1, 2]
-    ), f"Expecting weight_zero_points tensor to be None or have dim()==1, but found {weight_zero_points.dim()}"
+    assert weight_zero_points is None or weight_zero_points.dim() in [
+        1,
+        2,
+    ], f"Expecting weight_zero_points tensor to be None or have dim()==1, but found {weight_zero_points.dim()}"
     assert weight_zero_points is None or weight_zero_points.size(0) == weight.size(
         0
     ), f"Expecting weight_zero_points tensor to be None or have same number of rows as weights, but found {weight.size()} and {weight_zero_points.size()}"
@@ -278,6 +284,7 @@ def embedding_2bit(
     )
     return torch.ops.aten.embedding.default(weight, indices)
 
+
 @register_fake("quantized_decomposed::embedding_2bit")
 def _(
     weight: torch.Tensor,
@@ -286,11 +293,12 @@ def _(
     weight_quant_min: int,
     weight_quant_max: int,
     indices: torch.Tensor,
-    ):
+):
     num_embeddings, packed_embedding_dim = weight.shape
     embedding_dim = packed_embedding_dim * 4
     embedding = torch.nn.Embedding(num_embeddings, embedding_dim, device=weight.device)
     return embedding(indices)
+
 
 @register_fake("quantized_decomposed::embedding_2bit.out")
 def embedding_2bit_out_meta(
@@ -310,6 +318,7 @@ def embedding_2bit_out_meta(
         weight_quant_max,
         indices,
     )
+
 
 @impl(quantized_decomposed_lib, "embedding_2bit.dtype", "CompositeExplicitAutograd")
 def embedding_2bit_dtype(
@@ -352,6 +361,7 @@ def embedding_2bit_dtype(
     )
     return torch.ops.aten.embedding.default(weight, indices)
 
+
 @register_fake("quantized_decomposed::embedding_2bit.dtype")
 def _(
     weight: torch.Tensor,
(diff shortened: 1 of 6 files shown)


exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -22,6 +22,56 @@
...
+        weight_3 = weight_view[:, :, 3] << 6
+        packed_weight = weight_0 + weight_1 + weight_2 + weight_3
+        return packed_weight
+    elif bitwidth == 4:
+        assert embedding_dim % 2 == 0, "embedding_dim must be divisible by 2"
+        weight_range_shifted = weight.add(8).view(torch.uint8)
+        weight_view = weight_range_shifted.view(
+            weight.shape[0], weight.shape[1] // 2, 2
+        )
+        weight_even = weight_view[:, :, 0] * 16  # left shift 4
```
kimishpatel, line 50: Why * 16 here but shift by 4 above?

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -22,6 +22,56 @@
...
+        weight_odd = weight_view[:, :, 1]
+        packed_weight = weight_even + weight_odd
```
kimishpatel, line 52: bitwise OR

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -296,6 +362,22 @@ def embedding_2bit_dtype(
     return torch.ops.aten.embedding.default(weight, indices)
 
 
+@register_fake("quantized_decomposed::embedding_2bit.dtype")
```
kimishpatel, line 365: this is needed?

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -438,6 +535,22 @@ def embedding_4bit_dtype(
...
+    weight_scales: torch.Tensor,
+    weight_zero_points: Optional[torch.Tensor],
+    weight_quant_min: int,
+    weight_quant_max: int,
+    indices: torch.Tensor,
+    dtype: Optional[torch.dtype],
+) -> torch.Tensor:
+    num_embeddings, packed_embedding_dim = weight.shape
+    embedding_dim = packed_embedding_dim * 2
+    embedding = torch.nn.Embedding(num_embeddings, embedding_dim, device=weight.device)
```
kimishpatel, line 550: very small nit: small refactor can abstract these 3 lines out.

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -873,6 +986,186 @@ def replacement(x, dim, start, end, x_scale, x_zero_point, x_qmin, x_qmax):
...
+            zero_point_dtype_cast,
+            -128,
+            127,
+            indices,
+        )
+
+    def embedding_byte_dtype_pattern(
+        indices, int_data, group_size, scale, zero_point, output_dtype
+    ):
+        dq = torch.ops.torchao.dequantize_affine.default(
```
kimishpatel, line 1012: what does "INT" mean here?

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -873,6 +986,186 @@ def replacement(x, dim, start, end, x_scale, x_zero_point, x_qmin, x_qmax):
...
+    List[Tuple[Callable, Callable, List[Callable]]]
+):
+    def embedding_byte_pattern(indices, int_data, group_size, scale, zero_point):
+        dq = torch.ops.torchao.dequantize_affine.default(
+            int_data, [1, group_size], scale, zero_point, torch.int8, -128, 127
+        )
+        return torch.ops.aten.embedding.default(dq, indices)
+
+    def embedding_byte_replacement(indices, int_data, group_size, scale, zero_point):
+        zero_point_dtype_cast = torch.ops.aten.to.dtype(zero_point, scale.dtype)
```
kimishpatel, line 999: embedding byte ops take float zero point?

exir/tests/TARGETS
```diff
@@ -298,6 +298,8 @@ python_unittest(
         "//caffe2:torch",
         "//executorch/exir:lib",
         "//executorch/exir/passes:quant_fusion_pass",
+        "//pytorch/ao:torchao",
+        "//executorch/exir/passes:constant_prop_pass",
```
kimishpatel, line 302: shouldnt you call const prop in the QuantFusionPass?

exir/tests/test_quant_fusion_pass.py
```diff
@@ -373,3 +376,96 @@ def forward(self, indices):
...
             # )
+
+    def test_embedding_torchao(self) -> None:
+        for bit_width, test_dtype_variant, test_per_group in zip(
+            [2, 4, 8], [True, False], [True, False]
+        ):
+            self._test_embedding_torchao(bit_width, test_dtype_variant, test_per_group)
+
+    def _test_embedding_torchao(
+        self, bit_width: int, test_dtype_variant: bool, test_per_group: bool
```
kimishpatel, line 387: test_dtype_variant as bool feels like a bit of misnomer

kimishpatel (changes requested):
sending back for question on const prop

exir/passes/_quant_patterns_and_replacements.py
metascroy: The ones in the ET kernels do: https://github.com/pytorch/executorch/blob/main/kernels/quantized/cpu/embeddingxb.cpp#L200

exir/passes/_quant_patterns_and_replacements.py
metascroy: It's the zero point domain.  It means the zero points are integers.  But this is an arg torchao is going to change in their quant primitives when they clean them up.

exir/passes/_quant_patterns_and_replacements.py
metascroy: I think so?  You cannot dynamo trace code that doesn't have meta kernels registered.

exir/passes/_quant_patterns_and_replacements.py
metascroy: Isn't * 16 the same as shift by 4?

I just copied the code from here: https://github.com/pytorch/executorch/blob/main/examples/models/llama/source_transformation/quantize.py#L659-L683

But I can clean it up and change to using left shift.

exir/passes/_quant_patterns_and_replacements.py
```diff
@@ -234,6 +285,21 @@ def embedding_2bit(
...
+    weight: torch.Tensor,
+    weight_scales: torch.Tensor,
+    weight_zero_points: Optional[torch.Tensor],
+    weight_quant_min: int,
+    weight_quant_max: int,
+    indices: torch.Tensor,
+):
+    num_embeddings, packed_embedding_dim = weight.shape
+    embedding_dim = packed_embedding_dim * 4
+    embedding = torch.nn.Embedding(num_embeddings, embedding_dim, device=weight.device)
```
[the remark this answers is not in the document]
metascroy: I thought this would be easier because the output size depends on the embedding_dim and the shape of indices.  Rather than repeat the logic to construct the right shape, I thought of just using the fake kernel that already exists in nn.Embedding.

exir/tests/TARGETS
metascroy: Unfortunately QuantFusionPass works on a graph module and const_prop_pass works on an exported_program.  I thought we could enable both in to_executorch by default, rather than have users call the passes separately like is done in the unit test.

exir/tests/test_quant_fusion_pass.py
metascroy: I'll change to use_dtype_variant?

exir/passes/_quant_patterns_and_replacements.py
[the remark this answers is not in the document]
kimishpatel: ok sounds good. i am ok with it

exir/passes/_quant_patterns_and_replacements.py
kimishpatel: yeah sorry I think I asked because I wasnt sure how it was working before without it. But yeah you do need meta kernel

exir/tests/TARGETS
kimishpatel: I think const prop has some nuances which may make it non-trivial for excample for q/dq nodes.

My concern here would be the perf cliff for the uninitiated.

Can we just not const prop manually? instead of even introducing this op in the graph in the first place?

exir/tests/test_quant_fusion_pass.py
kimishpatel: i think i misunderstood this probably because it was a bit late when i was reviewin git. You just meant to check `.dtype` variant of the op and I thought you were checking fp16 vs fp32 vs other dtype variants

exir/tests/TARGETS
metascroy: Manual const propagation would still require updating the signature of the exported program because we're changing the weights.  So it couldn't be done on the graph module.

In terms of perf cliff, it will not lower to ExecuTorch without const propagation because the pack embedding op has no out-variant.

kimishpatel: I see. ok. but lets plan to have this cost prop done appropriately. I highly doubt that this can be done transparently. For example for quantized models' have dq on weights that might get const propagated

kimishpatel (approved):
On const prop question, lets follow up in a separate pr

PR #10326, opened 2025-04-21
Use dependency injection for runner

larryliu0820:
Summary:
Pass in runner components, move most of the instantiation logic from `load()` to a new static API `create()`.

This adds testability to runner components.

Differential Revision: D73165546

### DIFF
--- a/examples/models/llama/runner/runner.cpp
+++ b/examples/models/llama/runner/runner.cpp
@@ -4,15 +4,13 @@
  *
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
+ * @lint-ignore-every CLANGTIDY facebook-hte-Deprecated
  */
 
 // A simple llama2 runner that includes preprocessing and post processing logic.
 // The module takes in a string as input and emits a string as output.
 
 #include <executorch/examples/models/llama/runner/runner.h>
-
-#include <algorithm>
-#include <ctime>
 
 #include <executorch/extension/llm/runner/util.h>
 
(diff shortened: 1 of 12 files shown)


examples/models/llama/runner/runner.h
```diff
@@ -30,18 +30,22 @@ namespace example {
 
 class ET_EXPERIMENTAL Runner : public executorch::extension::llm::IRunner {
  public:
-  explicit Runner(
+  // Static factory method to create a Runner instance
+  static std::unique_ptr<Runner> create(
```
lucylq, line 34: most users will likely interact with runner via this API?

swolchok (commented):
I am generally opposed to both tests that use mocks and dependency injection.

I am generally opposed to mocks because they tend to produce [change detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). For the case at hand, why not create a Runner and see if it decodes the right number of tokens?

I am generally opposed to dependency injection because pass-from-above is typically all you need in order to accomplish the usual stated goals. In this particular diff, I don't immediately understand what Runner::create is accomplishing; I don't see any sort of flag that causes it to return a different Runner in a test environment, nor do the tests use it.

examples/models/llama/runner/runner.cpp
```diff
@@ -35,129 +32,161 @@ static constexpr auto kMaxSeqLen = "get_max_seq_len";
 static constexpr auto kMaxContextLen = "get_max_context_len";
 static constexpr auto kVocabSize = "get_vocab_size";
 static constexpr auto kUseKVCache = "use_kv_cache";
-static constexpr auto kUseSDPAWithKVCache = "use_sdpa_with_kv_cache";
 } // namespace
 
-Runner::Runner(
+std::unique_ptr<Runner> Runner::create(
```
swolchok, line 37: Specifically, it looks like create() could just be another Runner constructor; it's not clear to me why it has to be a static method instead.

examples/models/llama/runner/runner.h
```diff
@@ -30,18 +30,22 @@ namespace example {
...
       const std::string& model_path,
       const std::string& tokenizer_path,
-      std::optional<const std::string> data_path = std::nullopt);
+      std::optional<const std::string> data_path = std::nullopt,
+      float temperature = -1.0f);
 
-  [[deprecated(
-      "This constructor is deprecated. Use the constructor without temperature parameter instead.")]]
+  // Constructor with dependency injection
   explicit Runner(
```
swolchok, line 41: looks like I was confused in my original comment. This form of dependency injection (pass from above) is fine; I thought `Runner::create` was going to be part of a trend of giving everything a create method and therefore building our own dependency injection framework.

larryliu0820: yeah - create() is merely a util to replace the old constructor which takes in all the strings and instantiate components. 

From a high level I want to eventually build something like this:

Given a specific model (Llama or Qwen), there's a logic to determine the architecture (could be hardcoded and listed in a yaml file or whatever), in Llama and Qwen's case we know the model type is decoder only text only LLM, therefore we need `TextTokenGenerator` and `TextPrefiller` for it. If we see a Llava model we can instantiate a `ImagePrefiller`.

Then we can also look at the tokenizer artifacts being passed in from create() and tell that this is a HF tokenizer or a Tiktoken tokenizer or anything else, then instantiate the actual object.

Eventually we pass the `TextTokenGenerator`, `TextPrefiller` and `Tokenizer` objects to the `Runner` and it should just work.

larryliu0820:
> I am generally opposed to both tests that use mocks and dependency injection.
> 
> I am generally opposed to mocks because they tend to produce [change detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). For the case at hand, why not create a Runner and see if it decodes the right number of tokens?
> 
> I am generally opposed to dependency injection because pass-from-above is typically all you need in order to accomplish the usual stated goals. In this particular diff, I don't immediately understand what Runner::create is accomplishing; I don't see any sort of flag that causes it to return a different Runner in a test environment, nor do the tests use it.

I generally agree that we should not write behavior change detector. Though the tests I'm adding is slightly different, I'm not guarding the internal logic of the runner but rather focus on:

* If we pass in a seq_len into generate() will we generate exact number of tokens.
* If the runner sees a EOS token can it stop generate next token.

I didn't test the implementation details, but just want to make sure the API behavior is expected.

> why not create a Runner and see if it decodes the right number of tokens?

It's hard to test with real model because 1. it's slow, 2. it's hard to see if we generate the exact number of tokens because we are getting string not tokens.

Hope that makes sense.

examples/models/llama/runner/runner.h
larryliu0820: This is for allowing existing call sites to still get a Runner instance by passing in these strings.

See comment below, in the end I want to build some logic that takes in the model type and tokenizer type, and be able to get different runners.

iseeyuan (approved):

PR #10328, opened 2025-04-21
Add pass to tag external constants for delegates

lucylq:
* #10330
* __->__ #10328

generate pte+ptd file for a delegated linear example

Differential Revision: [D73281924](https://our.internmc.facebook.com/intern/diff/D73281924/)


exir/passes/external_constants_pass.py
```diff
@@ -74,3 +76,30 @@ def external_mutable_weights_pass(
                     node.meta["constant_tag"] = "_default_external_constant"
                     mutated = True
     return PassResult(gm, mutated)
+
+
+def xnnpack_external_constants_pass(
```
JacobSzwejbka, line 81: why is this xnnpack specific.

JacobSzwejbka (commented):
Does the main schema support named data?

lucylq:
> Does the main schema support named data?

Yes it does now with weight sharing: https://github.com/pytorch/executorch/blob/main/schema/program.fbs#L437

exir/passes/external_constants_pass.py
lucylq: Hmn, it doesn't have to be xnnpack-specific but I think it has to be delegate-specific; if we use the same tag as portable ('constant_tag'), then constants that are not lowered will be tagged and error out in the emitter. eg. for linear, we have weight, bias and arg0 as constants. weight and bias are lowered, but arg0 makes it through to the emitter and errors out there. 

I'm not sure if there will be similar issues with other delegates, so made it xnnpack-specific to be safe.

lucylq: cc @mcr229 maybe worthwhile to not tag arg0 (by checking node name for linear/conv) and make the pass more generic?

exir/passes/external_constants_pass.py
```diff
@@ -74,3 +86,34 @@ def external_mutable_weights_pass(
                     node.meta["constant_tag"] = "_default_external_constant"
                     mutated = True
     return PassResult(gm, mutated)
+
+
+def delegate_external_constants_pass(
+    gm: GraphModule,
+    ep: ExportedProgram,
+    filter_fn: Optional[Callable[[torch.fx.Node], str]] = None,
```
digantdesai, line 94: nit
```suggestion
    gen_tag_fn: Optional[Callable[[torch.fx.Node], str]] = None,
```

exir/passes/external_constants_pass.py
```diff
@@ -74,3 +86,34 @@ def external_mutable_weights_pass(
...
+) -> PassResult:
+    """
+    Tag external constants before to_backend. Tagged constants will be saved
+    to an external file.
```
digantdesai, line 98: This function doesn't know how the tags will be used.

```suggestion
    Tag external constants before to_backend.
```

exir/passes/external_constants_pass.py
```diff
@@ -74,3 +86,34 @@ def external_mutable_weights_pass(
...
+
+    Note: this pass must be run after run_decompositions(), as tags on
+    constants are removed then.
+
+    Args:
+        gm: GraphModule to tag.
+        ep: ExportedProgram, to distinguish if a node is a constant.
+        filter_fn: node -> str callable indicating the file (str) that a node should be saved to.
```
digantdesai, line 106: the comment doesn't make sense.. file?, this function doesn't know or cares about the return string content

backends/xnnpack/operators/node_visitor.py
```diff
@@ -592,8 +592,13 @@ def get_serialized_buffer_index(
         xnn_graph.constant_data.append(
             ConstantDataOffset(offset=UINT64_MAX, size=size, named_key=named_key)
         )
+
+        external_tag = tensor.meta.get("delegate_constant_tag", None)
```
digantdesai, line 596: Nit: add logger.info?

exir/passes/external_constants_pass.py
lucylq: updated, thanks!

exir/passes/external_constants_pass.py
lucylq: thanks!

mcr229 (approved):

PR #10338, opened 2025-04-21
Clarify ownership of runner components

larryliu0820:
Summary:
The ownership of these components need some clarification.

* `Module` should be shared by `TextDecoderRunner` and potentially `TextPrefiller` (or `ImagePrefiller` in multimodal runner).
* `TextDecoderRunner` should be shared by the `TextPrefiller` and `TextTokenGenerator`.
* `Tokenizer` should be owned by the `Runner` as well as `TextTokenGenerator`.

Differential Revision: D73399600


kirklandsign (approved):

examples/models/llama/tokenizer/llama_tiktoken.cpp
```diff
@@ -11,51 +11,85 @@
...
-  // pad the rest of the special tokens with reserved tokens
-  ssize_t reserved_special_token_num = 2;
-  while (special_tokens->size() < kSpecialTokensSize) {
-    special_tokens->emplace_back(
-        "<|reserved_special_token_" +
-        std::to_string(reserved_special_token_num++) + "|>");
+// Compile-time special tokens selection using templates
+template <Version V>
+struct SpecialTokensSelector {
+  static std::unique_ptr<std::vector<std::string>> get();
```
swolchok, line 24: I would call this `create`; `get` connotes that it's somehow returning a thing that already exists

examples/models/llava/runner/llava_text_decoder_runner.h
```diff
@@ -17,7 +17,8 @@ namespace example {
 class ET_EXPERIMENTAL LlavaTextDecoderRunner
     : public executorch::extension::llm::TextDecoderRunner {
  public:
-  explicit LlavaTextDecoderRunner(executorch::extension::Module* module)
+  explicit LlavaTextDecoderRunner(
+      std::shared_ptr<executorch::extension::Module> module)
       : TextDecoderRunner(module, true) {}
```
swolchok, line 22: need to `std::move(module)`

extension/llm/runner/image_prefiller.h
```diff
@@ -21,7 +21,8 @@ namespace llm {
 // Assuming kv cache and parallel prefill are enabled.
 class ET_EXPERIMENTAL ImagePrefiller {
  public:
-  explicit ImagePrefiller(::executorch::extension::Module* module)
+  explicit ImagePrefiller(
+      std::shared_ptr<::executorch::extension::Module> module)
       : module_(module) {}
```
swolchok, line 26: ditto move

extension/llm/runner/text_decoder_runner.cpp
```diff
@@ -21,7 +21,9 @@ namespace llm {
 // NOTE: we observed ~2x loading performance increase on iPhone 15
 // and a ~5% improvement on Galaxy S22 by switching to
 // FileDataLoader instead of MmapDataLoader + UseMlockIgnoreErrors.
-TextDecoderRunner::TextDecoderRunner(Module* module, bool use_kv_cache)
+TextDecoderRunner::TextDecoderRunner(
+    std::shared_ptr<Module> module,
+    bool use_kv_cache)
     : module_(module), use_kv_cache_(use_kv_cache) {}
```
swolchok, line 27: move

extension/llm/runner/text_prefiller.cpp
```diff
@@ -17,7 +17,7 @@ namespace extension {
 namespace llm {
 
 TextPrefiller::TextPrefiller(
-    TextDecoderRunner* text_decoder_runner,
+    std::shared_ptr<TextDecoderRunner> text_decoder_runner,
```
swolchok, line 20: need to move this on line 24 below

extension/llm/runner/text_token_generator.h
```diff
@@ -21,17 +21,19 @@ namespace llm {
...
+      std::shared_ptr<::tokenizers::Tokenizer> tokenizer,
+      std::shared_ptr<TextDecoderRunner> text_decoder_runner,
       bool use_kv_cache,
       std::unique_ptr<std::unordered_set<uint64_t>>&& eos_ids,
-      Stats* stats)
+      std::shared_ptr<Stats> stats)
       : tokenizer_(tokenizer),
         text_decoder_runner_(text_decoder_runner),
         eos_ids_(std::move(eos_ids)),
         use_kv_cache_(use_kv_cache),
```
swolchok, line 32: move

swolchok (changes requested):
please also get the information about ownership from your summary into the code as comments so we can tell what things are supposed to be shared between

examples/models/llava/runner/llava_text_decoder_runner.h
larryliu0820: Do we do std::move for shared_ptr as well? I always thought only unique_ptr uses it.

larryliu0820:
> please also get the information about ownership from your summary into the code as comments so we can tell what things are supposed to be shared between

Instead of using shared_ptr I added comments to explain the ownership. Still use raw pointers for Tokenizer and TextDecoderRunner etc.

swolchok (changes requested):

examples/models/llama/tokenizer/llama_tiktoken.cpp
```diff
@@ -11,51 +11,71 @@
...
+// Helper function to create a Tiktoken with the given version
+template <Version V>
+std::unique_ptr<Tiktoken> create_tiktoken() {
+  std::unique_ptr<std::vector<std::string>> special_tokens =
+      example::SpecialTokensSelector<V>::create();
   return std::make_unique<Tiktoken>(
-      _get_special_tokens(version), kBOSTokenIndex, kEOSTokenIndex);
+      std::move(special_tokens), kBOSTokenIndex, kEOSTokenIndex);
+}
+} // namespace detail
```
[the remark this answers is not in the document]
larryliu0820: The reason is I wanted to use this to support tokenizer for different models. I guess it doesn't belong to the current PR so I'll separate it out.

examples/models/llama/runner/runner.h
```diff
@@ -71,7 +71,7 @@ class ET_EXPERIMENTAL Runner : public executorch::extension::llm::IRunner {
       text_token_generator_;
 
   // stats
-  ::executorch::extension::llm::Stats stats_;
+  std::unique_ptr<::executorch::extension::llm::Stats> stats_;
```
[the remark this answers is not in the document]
larryliu0820: Ah that's because TextTokenGenerator needs to share the same stats with Runner, TextTokenGenerator(stats) but also Runner constructor is going to take in a TextTokenGenerator. Runner(text_token_generator)

Therefore the only way I can think of is to:

```cpp
auto stats = std::make_unique<Stats>();
auto text_token_generator = std::make_unique<TextTokenGenerator>(stats.get());
auto runner = std::make_unique<Runner>(std::move(stats), std::move(text_token_generator));
```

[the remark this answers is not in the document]
larryliu0820: of course I can put this change into next PR.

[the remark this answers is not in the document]
swolchok: if you want to avoid the heap allocation, you can have the Runner hand a pointer to its Stats object to the TextTokenGenerator, or you can have the runner create the TextTokenGenerator and leave the constructor as-is.

swolchok (approved):
LGTM

PR #10345, opened 2025-04-22
[Android] New config API for Llm init and generate

kirklandsign:
Add a config for init, and generate() similar to GenerationConfig


larryliu0820 (approved):
Thank you!

PR #10347, opened 2025-04-22
Support dynamically quantized 2D convolutions

### LINKED ISSUE #9021: Support Dynamically Quantized Convolutions

keyprocedure:
### Summary
Add initial support for dynamically quantized Conv2d in XNNPACK:
- Add `conv` to `DYNAMIC_OPS` for annotation
- Update partitioner to support dynamically quantized Conv2d
- Add checks to ensure only 2D, non-depthwise dynamically quantized convs are partitioned and annotated
- Update NHWC permute node insertion to trace back to original input for dynamically quantized inputs
- Compute `num_nonbatch_dims` based on whether the node feeds into a conv
- Remove the `num_nonbatch_dims` check from XNNCompiler
- Add unit tests for channels-last permute and single, sequential, and parallel dynamically quantized 2D convs 

Fixes #9021 

### Test plan
```bash
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d_seq
python -m unittest backends.xnnpack.test.ops.test_conv2d.TestConv2d.test_dq_conv2d_parallel
python -m unittest backends.xnnpack.test.passes.test_channels_last_tagged_reshape.TestChannelsLastTaggedReshapePass.test_dq_conv2d_channels_last_tagged_reshape_pass


mcr229 (approved):
looks good! let's wait for CI to run and then we can merge. Thank you for the contribution!

mcr229:
lintrunner failure isn't relevant and is fixed on main. Merging changes

PR #10350, opened 2025-04-22
Arm backend: Remove build_quantized_ops_aot_lib.sh

AdrianLundell:
This lib is built by default when installing executorch, so use that instead of building our own version. The the version built by executorch however has an extra dependency on _portable_lib.cpython-310, which needs to be loaded as well.

Also removes all mentions of this script in related files.


AdrianLundell:
Unrelated lint failure in unmodified file.

zingo (approved):

PR #10351, opened 2025-04-22
Arm backend: Populate __init__.py for quantizer and Arm root

oscarandersson8218:
Populate backends/arm/quantizer/__init__.py and backends/arm/__init__.py


zingo (approved):

oscarandersson8218:
@digantdesai feel free to revert this if necessary.

digantdesai:
Yeah I haven't looked into where the circular dependency is coming from.. this is when I try to update the buck files.

re. `is_tosa` etc. it might be OK but feel it is cleaner to import from one module namespace when they are in the same python file. We can move them into other file if you don't want to expose them to the end users. 

Also I see main, operators, quantizer, _passes have __init__.py, are you planning to refactor others beyond main and quantizer here?

PR #10372, opened 2025-04-22
[Feature Request] Add job_arn  to benchmark result

yangw-dev:
Add device job_arn to track different device jobs.

### DIFF
--- a/.github/scripts/extract_benchmark_results.py
+++ b/.github/scripts/extract_benchmark_results.py
@@ -536,7 +536,9 @@ def extract_benchmark_result_from_artifact(
         )
     if not benchmark_results:
         return []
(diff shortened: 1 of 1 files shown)


huydhn (approved):

huydhn:
Please find lint before merging

huydhn:
Just FYI, I'm also adding the device pool name and id here so that I can differentiate between a S22 device from a public v.s. a S22 device from the private pool.  Atm, I think they have the same name.

yangw-dev:
> Just FYI, I'm also adding the device pool name and id here so that I can differentiate between a S22 device from a public v.s. a S22 device from the private pool. Atm, I think they have the same name.

sounds good!

Issue #10364, opened 2025-04-22
Introduce Protected Method Getter in Extension.Module


zhenyan-zhang-meta:
# Context

This issue is the first step of https://github.com/pytorch/executorch/discussions/9638.

In https://github.com/pytorch/executorch/discussions/9638, we want to have `extension.Module` as the single source of implementation in `pybindings`, which means that `pybindings.PyModule` should use `extension.Module` rather than its own `pybindings.Module`.

The issue is that `pybindings.PyModule` is dependent on the `method` getter from `pybindings.Module`, which `extension.Module` do not have. Since we don't want to expose `method` getter in `extension.Module`, we have to protect the getter, wrap the functions that is dependent on it and use the protected getter there, ultimately decouple `pybindings` from a `method` getter.

# Proposal

Although we are decoupling the `method` getter from `pybindings` implementation, `method` itself is still needed for bundled programs. To keep having the `method` getter while not exposing it, we can create a protected method getter to confine its usage inside a wrapper class inheriting `extension.Module` that we are about to create.

```cpp
// In extension.module
class Module {
    // Everything else keeps same

protected:
  /**
   * Get a method by method name.
   *
   * @param[in] method_name The name of the method to get.
   *
   * @returns A Result object containing either a pointer to the requested
   *          method or an error to indicate failure.
   */
  ET_NODISCARD inline runtime::Result<Method*> get_method(
      const std::string& method_name);
}
```

PR #10374, opened 2025-04-22
[ExecuTorch][#10364] Add Protected Method Getter in `extension.Module`

zhenyan-zhang-meta:
* __->__ #10374

# Context

This issue is a step of https://github.com/pytorch/executorch/discussions/9638. In the discussion, we want to unblock having `extension/Module` as the single source of implementation, which means that `pybindings/PyModule` should use `extension/Module` rather than its own.

Although we are decouping method getter from `pybindings` implementation, method getter itself is still needed. To keep having the method getter while not exposing it, we can create a protected method getter and confine it's usage inside child classes that we are about to create.

# Proposal

Add a protected `get_method` to `extension.Module`, taking method name string as an input.

Differential Revision: [D73473766](https://our.internmc.facebook.com/intern/diff/D73473766/)

### DIFF
--- a/extension/module/module.cpp
+++ b/extension/module/module.cpp
@@ -302,5 +302,15 @@ runtime::Error Module::set_output(
       output_tensor.mutable_data_ptr(), output_tensor.nbytes(), output_index);
 }
 
+ET_NODISCARD inline runtime::Result<Method*> Module::get_method(
+    const std::string& method_name) {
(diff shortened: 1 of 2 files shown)


zhenyan-zhang-meta:
https://github.com/pytorch/executorch/issues/10364

Gasoonjia (approved):
LGTM!

PR #10376, opened 2025-04-23
Minibench refactor

kirklandsign:
Allow running generic model benchmark before LLM

Now all models use the same org.pytorch.minibench.BenchmarkActivity


guangy10 (approved):
Only the linter issue to fix

PR #10385, opened 2025-04-23
Arm backend: Allow --quantize in non delegated using aot_arm_compiler

zingo:
This fix a bug where the --quantize flag didn't work if --delegate was not set. So if compiling for non Ethos-U targets you always got float models.


oscarandersson8218 (approved):
Unrelated lintrunner failure

PR #10391, opened 2025-04-23
Arm backend: Add unit test for DeiT-Tiny model on TOSA-MI backend

martinlsm:
Add unit test that runs the DeiT-Tiny model on the TOSA-MI profile. The output performance is not great at the moment. An internal ticket has been created to debug this problem later.

### Test plan
Command used to test the patch:
`python -m pytest -s -c /dev/null --log-level=DEBUG "backends/arm/test/models/test_deit_tiny_arm.py"`

### DIFF
--- a/backends/arm/test/models/test_deit_tiny_arm.py
+++ b/backends/arm/test/models/test_deit_tiny_arm.py
@@ -1,0 +1,44 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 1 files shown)


oscarandersson8218:
@martinlsm We may need to increase `atol` or mark the test as flaky?

martinlsm:
> @martinlsm We may need to increase `atol` or mark the test as flaky?

I increased a little bit now. Hopefully enough.

zingo (approved):

PR #10395, opened 2025-04-23
Revert "Arm backend: Populate __init__.py for quantizer and Arm root"

digantdesai:
Reverts pytorch/executorch#10351


zingo (approved):

jackzhxng (approved):
Temporary revert to unblock difftrain

PR #10397, opened 2025-04-23
Revert "Support dynamically quantized 2D convolutions"

jackzhxng:
Breaking internal tests, reverting temporarily to unblock diff train


jackzhxng:
@mcr229 to land https://github.com/pytorch/executorch/pull/10347 separately from diff train

kirklandsign (approved):

PR #10403, opened 2025-04-23
Add CI workflow to check c10 is synced with PyTorch

swolchok:
This script checks that all files in our c10 copy are in sync with our pinned version of PyTorch.

NOTE TO REVIEWERS: as part reviewing this PR, please weigh in on whether this is a sensible policy. Instinctively, I wanted to allow temporary divergence from the pinned PyTorch version, but I'm unable to articulate why. It now seems like this is the only sensible policy, as our c10 copy is logically part of our PyTorch pin.

Test Plan: 1) detected us being out of sync when I wrote this PR
2) CI to run it and make sure that we are clean now

### DIFF
--- a/.ci/scripts/check_c10_sync.sh
+++ b/.ci/scripts/check_c10_sync.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set -exu
+ls pytorch/.git || git clone https://github.com/pytorch/pytorch.git
+pushd pytorch
+git checkout "$(< ../.ci/docker/ci_commit_pins/pytorch.txt)"
+popd
+"$(dirname "${BASH_SOURCE[0]}")"/diff_c10_mirror_with_pytorch.sh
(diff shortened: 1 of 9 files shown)


swolchok:
* (to be filled)

.ci/scripts/check_c10_sync.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
mergennachin, line 1: where is this file being used?

.ci/scripts/compare_dirs.sh
```diff
@@ -0,0 +1,53 @@
...
+fi
+dir1="$1"
+dir2="$2"
+
+if [ ! -d "$dir1" ] || [ ! -d "$dir2" ]; then
+    echo "Error: Both directories must exist" >&2
+    exit 1
+fi
+
+while IFS= read -r -d '' file; do
```
mergennachin, line 26: Thoughts on continuing after errors (instead of exiting on first mismatch)

mergennachin (commented):
Yeah, this looks like a good policy

.ci/scripts/check_c10_sync.sh
swolchok: whoops! supposed to be the thing that the workflow calls

.ci/scripts/compare_dirs.sh
swolchok: done

swolchok:
ugh, I don't recall clicking the merge button on this PR. will re-send

swolchok:
>  clicking the merge button

accidentally sent this out with the order of the stack reversed and I guess that's how github chose to reflect what I did. replacement is https://github.com/pytorch/executorch/pull/10413

PR #10412, opened 2025-04-23
[Android] Support 16bit output data as raw data byte[]

### LINKED ISSUE #9881: Support BFloat16 dtype in Android Tensor API

Currently, the ExecuTorch Android Java tensor class does not support BFloat16 dtype tensors. This prevents Java users from passing BFloat16 tensors in or out of the model. BFloat16 is increasingly common for GenAI, so this will be a useful feature to have.

To implement support, we will need to logic to [Tensor.java](https://github.com/pytorch/executorch/blob/main/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java). This task will require the following:
- Create a `Tensor_bfloat16` implementation, similar to the existing `Tensor_int8` and `Tensor_float32`.
- Update the dtype switches in Tensor.java to support bfloat16.
- Add corresponding tests to [TestTensor.java](https://github.com/pytorch/executorch/blob/a884e9274de4ada2641f1ec23f68178799bc0914/extension/android/executorch_android/src/test/java/org/pytorch/executorch/TensorTest.java#L25).

CC https://github.com/pytorch/executorch/issues/6571

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @kirklandsign @cbilgin

### LINKED ISSUE #10371: [Android] Add java layer fp16 support
### 🐛 Describe the bug

Right now, it only has fp32 support. No fp16 or bf16 support.

### Versions

0.6

cc @cbilgin

kirklandsign:
In java, when the returned dtype is fp16 or bf16, instead of crash, use byte[] to represent these raw data, and let user parse the byte[]

We don't support fp16 or bf16 as input dtype. User can convert to fp32.

Resolved: https://github.com/pytorch/executorch/issues/9881
Resolved: https://github.com/pytorch/executorch/issues/10371

cc @cbilgin @guangy10 @huydhn @shoumikhin

### DIFF
--- a/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java
+++ b/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java
@@ -714,7 +714,7 @@ public abstract class Tensor {
       tensor = new Tensor_int8(data, shape);
     } else if (DType.HALF.jniCode == dtype) {
       tensor = new Tensor_raw_data_16b(data, shape, DType.HALF);
-    } else if (DType.HALF.jniCode == dtype) {
+    } else if (DType.BFLOAT16.jniCode == dtype) {
       tensor = new Tensor_raw_data_16b(data, shape, DType.BFLOAT16);
     } else {
       throw new IllegalArgumentException("Unknown Tensor dtype");
--- a/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java
+++ b/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java
@@ -630,6 +630,44 @@ public abstract class Tensor {
     }
   }
 
+  /**
+   * A generic holder for 16b (fp16/bfloat16) tensor. Not intended for general use. User can only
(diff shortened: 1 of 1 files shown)


extension/android/executorch_android/src/main/java/org/pytorch/executorch/Tensor.java
```diff
@@ -674,6 +712,10 @@ private static Tensor nativeNewTensor(
       tensor = new Tensor_uint8(data, shape);
     } else if (DType.INT8.jniCode == dtype) {
       tensor = new Tensor_int8(data, shape);
+    } else if (DType.HALF.jniCode == dtype) {
+      tensor = new Tensor_raw_data_16b(data, shape, DType.HALF);
+    } else if (DType.HALF.jniCode == dtype) {
```
phaiting, line 717: This is same check as L715, should be DType.BFLOAT16.jniCode == dtype?

kirklandsign:
Per discussion, we use https://github.com/pytorch/executorch/pull/10414 instead. Don't support it as output data type.

PR #10414, opened 2025-04-23
[Android] Add Tensor_unsupoprted return type instead of crash

### LINKED ISSUE #10371: [Android] Add java layer fp16 support

kirklandsign:
Create a Tensor_unsupoprted object, which user can query its dtype and shape, but no other op supported. It will give an error saying user should re-export a model which Java supports.

Resolved: https://github.com/pytorch/executorch/issues/10371


phaiting (approved):

guangy10 (approved):

larryliu0820 (approved):
Thank you for addressing comments

PR #10390, opened 2025-04-23
Arm backend: Update node visitors to support TOSA 1.0

per:
### Summary
Updates node visitors to support TOSA 1.0 specification. 

### Test plan
Tested through public and internal CI.

cc @digantdesai @freddan80 @zingo @oscarandersson8218

PR #10425, opened 2025-04-24
Arm backend: Update more node visitors to support TOSA 1.0

oscarandersson8218:
### Summary
Updates more node visitors to support TOSA 1.0 specification.

### Test plan
Tested through public and internal CI.


zingo (approved):

digantdesai:
Yeah we don't have 1.0 internally yet, I can pull it in. I wanted to discuss with @per how this is going to evolve going forward, but we didn't sync yesterday :(

backends/arm/operators/op_cos.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-unsafe
+from typing import List
+
+import serializer.tosa_serializer as ts  # type: ignore
```
digantdesai, line 9: this can be inside the class and it might work

backends/arm/operators/op_upsample_nearest2d.py
```diff
@@ -20,19 +19,23 @@
...
 
     def define_node(
         self,
         node: torch.fx.Node,
-        tosa_graph: ts.TosaSerializer,
+        tosa_graph: Any,
         inputs: List[TosaArg],
         output: TosaArg,
     ) -> None:
+        import tosa_tools.v0_80.serializer.tosa_serializer as ts  # type: ignore
```
digantdesai, line 37: want to move 
`from tosa_tools.v0_80.tosa.ResizeMode import ResizeMode` here as well? Also do the same for v1.0 i.e. `from tosa.ResizeMode import ResizeMode`?

oscarandersson8218: Yes, good catch!

oscarandersson8218:
> Yeah we don't have 1.0 internally yet, I can pull it in. I wanted to discuss with @per how this is going to evolve going forward, but we didn't sync yesterday :(

I'm struggling to see how this is any different from https://github.com/pytorch/executorch/pull/10390 with regards to TOSA 1.0. Is it the sin and cos which are causing the issues?

PR #10440, opened 2025-04-24
Re-sync extension/pytree/function_ref.h with LLVM

swolchok:
We had an old version, and that version didn't work quite right when
handed a const ref to a callable. Just resync and note divergences
from the original in the comment. Our old version also had an
`explicit` constructor, which is *not* desirable because it
complicates callsites. It was also a divergence from both LLVM and
[the upcoming C++26
std::function_ref](https://en.cppreference.com/w/cpp/utility/functional/function_ref/function_ref).

The test made invalid use of a FunctionRef -- you can't keep one
around past the lifetime of the function object it refers to, but the
test held onto a temporary capturing lambda in an object. I fixed the
test.


swolchok:
* (to be filled)

extension/pytree/test/function_ref_test.cpp
```diff
@@ -76,9 +52,8 @@ TEST(FunctionRefTest, FunctionPointer) {
...
   EXPECT_EQ(val, 1);
 
-  Item item(0, one);
-  EXPECT_EQ(item.get(), 1);
-
-  Item item1(0, &one);
-  EXPECT_EQ(item1.get(), 1);
+  val = 0;
+  FunctionRef<void(int32_t&)> ref2(one);
+  ref(val);
```
digantdesai, line 57: s/ref/ref2 perhaps?

digantdesai (approved):

swolchok:
bypassing two apparent tests flakes and merging.

PR #10441, opened 2025-04-24
Move fully-featured FunctionRef from extension/pytree to ExecuTorch core

swolchok:
I just re-synced it with LLVM, and it seems harmless. We previously simplified FunctionRef because of an ARM baremetal toolchain issue around placement of a lambda (#555), but the current implementation has no lambda.

Motivation: I want to be able to use it in the next PR in thread_parallel_interface.h (the "threadpool active" mode) and threadpool in order to save a ton of size that std::function is currently wasting by virtue of supporting ownership and copying.


swolchok:
* (to be filled)

digantdesai (approved):

swolchok:
unit tests look good despite three apparent flakes. will merge after rebase.

Issue #10375, opened 2025-04-22
Add `extension.BundledModule` to Wrap `extension.Module` with Bundled Program Logic


zhenyan-zhang-meta:
# Context

This issue is a step of https://github.com/pytorch/executorch/discussions/9638 and a sequel of https://github.com/pytorch/executorch/issues/10364.

In https://github.com/pytorch/executorch/discussions/9638, we want to have `extension.Module` as the single source of implementation in `pybindings`, which means that `pybindings.PyModule` should use `extension.Module` rather than its own `pybindings.Module`.

The issue is that `pybindings.PyModule` is dependent on the `method` getter from `pybindings.Module`, which `extension.Module` do not have. Since we don't want to expose `method` getter in `extension.Module`, we have to protect the getter, wrap the functions that is dependent on it and use the protected getter there, ultimately decouple `pybindings` from a `method` getter.

# Proposal

Now that we have a protected `method_` to use, we can introduce a `extension.BundledModule`, a child class inheriting `extension.Module` which wraps up bundled program logic that is dependent on the `method` getter.

```cpp
class BundledModule : public Module {
   
  // Load a specific method with the input value at the given `testset_idx` from the bundle to the method.
  ET_NODISCARD runtime::Error load_bundled_input(
      const std::string& method_name,
      const size_t testset_idx);

   // Verify the output of a specific method with the expected output from the program bundle at the given `testset_idx`.
   ET_NODISCARD inline runtime::Result<bool> verify_output_with_bundled(const std::string& method_name, const size_t bundled_output_index, std::vector<runtime::Evalue>> method_output; other_comparsion_configs);
}
```

PR #10449, opened 2025-04-24
[ExecuTorch][#10375] Add `extension.BundledModule` to Wrap `extension.Module` with Bundled Program Logic

zhenyan-zhang-meta:
* __->__ #10449

https://github.com/pytorch/executorch/issues/10375

# Context

This issue is a step of https://github.com/pytorch/executorch/discussions/9638.

In https://github.com/pytorch/executorch/discussions/9638, we want to have `extension.Module` as the single source of implementation in `pybindings`, which means that `pybindings.PyModule` should use `extension.Module` rather than its own `pybindings.Module`.

The issue is that `pybindings.PyModule` is dependent on the `method` getter from `pybindings.Module`, which `extension.Module` do not have. Since we don't want to expose `method` getter in `extension.Module`, we have to protect the getter, wrap the functions that is dependent on it and use the protected getter there, ultimately decouple `pybindings` from a `method` getter.

# Proposal

Now that we have a protected `method` getter, we can introduce a `extension.BundledModule`, a child class inheriting `extension.Module` which wraps up bundled program logic that is dependent on the `method` getter.

Differential Revision: [D73564125](https://our.internmc.facebook.com/intern/diff/D73564125/)


extension/module/bundled_module.cpp
```diff
@@ -0,0 +1,112 @@
...
+    std::unique_ptr<runtime::DataLoader> data_map_loader)
+    : Module(
+          program_data_loader(bundled_program_ptr),
+          std::move(memory_allocator),
+          std::move(temp_allocator),
+          std::move(event_tracer),
+          std::move(data_map_loader)),
+      bundled_program_ptr_(bundled_program_ptr) {}
+
+runtime::Result<std::unique_ptr<BundledModule>> BundledModule::from_file(
```
Gasoonjia, line 44: It would be great if we can follow the `extension.module`'s pattern, which uses constructor instead of factory function, and leverage `is_loaded` function for sanity check.

Gasoonjia (approved):
current version is good enough for unblocking our requirement. Thanks @zhenyan-zhang-meta for your great work!

extension/module/bundled_module.cpp
zhenyan-zhang-meta: Yeah we can follow up in another PR to iterate on it. Will land this diff as discussed and write up potential improvements in https://github.com/pytorch/executorch/issues/10375.

Issue #10454, opened 2025-04-24
[Android] Migrate tests to kotlin


kirklandsign:
### 🚀 The feature, motivation and pitch

Android Studio can do it one click.

https://github.com/search?q=repo%3Apytorch%2Fexecutorch+test+language%3AJava&type=code

However, need to configure gradle and CI env?

cc @cbilgin

PR #10458, opened 2025-04-25
Java test use kotlin

kirklandsign:
Want to migrate as in https://github.com/pytorch/executorch/issues/10454


kirklandsign:
Already done 😄

PR #10473, opened 2025-04-25
Arm backend: Broaden exception handling for unsupported ops

Sebastian-Larsson:
Previously, only AssertionError, RuntimeError, and ValueError were caught during node translation failures in the TOSA backend. This patch broadens the exception catch to all Exception types to ensure robust error reporting via `dbg_fail()` even for unexpected errors.

This change helps avoid silent crashes or unhelpful tracebacks during backend lowering of unsupported or misconfigured nodes.


zingo (approved):

zingo:
macos fails unrelated

PR #10474, opened 2025-04-25
Arm backend: Update op_view for TOSA 1.0

oscarandersson8218:
Make op_view lower rank 0 outputs instead of raising an exception. Rank 0 outputs are triggered by test_amin.py and test_amax.py .


zingo (approved):

zingo:
macOS fail is inrelated

PR #10484, opened 2025-04-25
Add pass to remove unused parameters in to_edge

GregoryComer:
Summary:
Currently, ExecuTorch will serialize any parameters in the exported program, regardless of whether they are actually used. Exporting with strict=True will remove unused parameters, but strict=False will not. Export recently switched to non-strict as the default behavior.

This causes PTE bloat when doing pt2e quantization (unquantized weights are left in the graph) or sometimes when exporting multiple methods (encode and decoder, for example).

This PR adds a new pass (`remove_unused_parameters_pass`) to strip unused parameters from the `ExportedProgram`. It is run as part of `to_executorch`. Parameters are considered unused if there are no uses of the placeholder node. Parameters are removed by stripping them from the state_dict, input specs, and graph.

As a question for reviewers, should we run this pass earlier, as part of to_edge? My rationale for running as part of to_executorch was that it could theoretically clean up anything else left by partitioning and lowering, but I'm not aware of any concrete use cases for this.

Differential Revision: D73654202

cc @JacobSzwejbka @angelayi

### DIFF
--- a/exir/passes/remove_unused_parameters_pass.py
+++ b/exir/passes/remove_unused_parameters_pass.py
@@ -1,0 +1,55 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import copy
+
+import torch
+
+from torch.export.exported_program import ExportedProgram, InputKind
+
+
+def remove_unused_parameters_pass(
+    ep: ExportedProgram,
+) -> ExportedProgram:
+    """
+    Remove unused parameters from the exported program.
+    """
+
+    placeholder_nodes = {
+        node.target: node
+        for node in ep.graph_module.graph.nodes
+        if node.op == "placeholder"
+    }
+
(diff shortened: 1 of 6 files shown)


JacobSzwejbka:
> It is run as part of to_executorch

Can we run it in to_edge instead? 

We should remove unused things as soon as possible imo.

exir/passes/remove_unused_parameters_pass.py
```diff
@@ -0,0 +1,54 @@
...
+    ep.graph_module.recompile()
+    return ep
+
+
+def _is_parameter_used(
+    ep: ExportedProgram, parameter: str, placeholder_nodes: dict[str, torch.fx.Node]
+) -> bool:
+    placeholder_node = placeholder_nodes.get(parameter)
+    if placeholder_node is None:
+        # Shouldn't happen, but in this case, leave the parameter to be safe.
```
JacobSzwejbka, line 51: throw an exception?

JacobSzwejbka (approved):

JacobSzwejbka (approved):

PR #10487, opened 2025-04-25
Add direct copy fast path for portable copy op

GregoryComer:
Summary:
The PR adds a direct memcpy fast-path for portable copy and copy_ ops. This speeds up copy significantly in cases where no broadcasting is needed. Note that the copy op always checks that dim order and dtype match, so this should be sound in all cases where the shape matches (no broadcasting).

This is most noticeable when copying buffer mutations back, such as transformer KV cache when managing the cache as a mutable buffer. Prior to this change, an encoder/decoder model was taking ~25% of the total runtime copying KV cache back after permuting. With this change, the copy becomes significantly cheaper.

I benchmarked a simple model on S23 and Pixel 5:
```
class TestModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.register_buffer("buffer", torch.zeros((2, 10, 1024, 1024)))

    def forward(self, x):
        self.buffer.add_(x)
        return self.buffer


model = TestModel()
inputs = (torch.randn(2, 10, 1024, 1024),)

lowered = to_edge_transform_and_lower(
    torch.export.export(model, inputs),
    partitioner=[XnnpackPartitioner()],
).to_executorch()
```

S23, average of 50 runs, time in copy_:
4.1ms vs 22.3ms

Pixel 5, average of 50 runs, time in copy_:
12.1ms vs 66.6ms

This is approximately a ~5.5x speedup of the copy operator.

Differential Revision: D73656456

### DIFF
--- a/kernels/portable/cpu/op_copy.cpp
+++ b/kernels/portable/cpu/op_copy.cpp
@@ -46,17 +46,24 @@ Tensor& copy_out(
   // @lint-ignore CLANGTIDY facebook-hte-CArray
   static constexpr const char op_name[] = "copy.out";
 
-  ET_SWITCH_REALHBBF16_TYPES(in.scalar_type(), ctx, "copy.out", CTYPE, [&]() {
-    utils::apply_bitensor_elementwise_fn<CTYPE, op_name>(
-        [](ET_UNUSED const CTYPE _, const CTYPE val_src) { return val_src; },
-        ctx,
-        in,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        src,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        out,
-        utils::SupportedTensorDtypes::REALHBBF16);
-  });
+  // Use direct copy fast path if broadcast is not needed and tensors are
+  // non-empty
+  if (internal::sizes_match_ignoring_leading_1s(out.sizes(), src.sizes()) &&
+      src.numel() > 0) {
+    std::memcpy(out.mutable_data_ptr(), src.const_data_ptr(), src.nbytes());
+  } else {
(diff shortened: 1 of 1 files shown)


kernels/portable/cpu/op_copy.cpp
```diff
@@ -46,17 +46,23 @@ Tensor& copy_out(
...
-        in,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        src,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        out,
-        utils::SupportedTensorDtypes::REALHBBF16);
-  });
+  // Use direct copy fast path if broadcast is not needed and tensors are
+  // non-empty
+  if (tensors_have_same_shape(out, src) && src.numel() > 0) {
```
swolchok, line 51: AIUI the requirement is looser than this; leading 1s can be ignored. how about using sizes_match_ignoring_leading_1s?

kernels/portable/cpu/op_copy.cpp
```diff
@@ -79,17 +85,23 @@ Tensor& copy_(
...
-        in,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        src,
-        utils::SupportedTensorDtypes::REALHBBF16,
-        in,
-        utils::SupportedTensorDtypes::REALHBBF16);
-  });
+  // Use direct copy fast path if broadcast is not needed and tensors are
+  // non-empty
+  if (tensors_have_same_shape(in, src) && src.numel() > 0) {
```
swolchok, line 90: ditto

swolchok (approved):

GregoryComer:
Overriding lint failure for land - broken trunk.

PR #10497, opened 2025-04-25
Use generated files in module test

lucylq:
* __->__ #10497

Instead of checking in the PTE/PTD files. This helps us keep up to date with aot changes, especially with incoming flat_tensor refactor.

Main changes here are with the model. Instead of using AddModule from examples/portable, use ModuleAdd from test/models/export_program, like our other tests. There is a shape difference (1 tensor vs 2 tensors+alpha), so we have to change all the inputs.

Differential Revision: [D73470865](https://our.internmc.facebook.com/intern/diff/D73470865/)


JacobSzwejbka (approved):
thanks!

extension/module/test/module_test.cpp
```diff
@@ -26,9 +27,9 @@ class ModuleTest : public ::testing::Test {
     if (const char* env = std::getenv("RESOURCES_PATH")) {
       resources_path = env;
```
shoumikhin, line 28: Is this path used for anything anymore?

lucylq: ah it's not. I'll remove it, thanks

kirklandsign (approved):

PR #10501, opened 2025-04-26
Allow removing permute pairs in addition to transpose pairs

mcremon-meta:
Summary:
As titled. Gets us 27% better cycles on Activity Classification (at opt level 3).

Can be improved further, task is T222295719

Reviewed By: zonglinpeng

Differential Revision: D73619452

### DIFF
--- a/backends/cadence/aot/fuse_ops.py
+++ b/backends/cadence/aot/fuse_ops.py
@@ -14,7 +14,7 @@ import math
 import operator
 from collections import deque
 from numbers import Number
-from typing import cast, Sequence
+from typing import cast
 
 # Import these for the cadence function signatures.
 import executorch.backends.cadence.aot.ops_registrations  # noqa: F401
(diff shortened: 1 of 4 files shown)


dulinriley (approved):

backends/cadence/aot/fuse_ops.py
```diff
@@ -907,42 +907,28 @@ def can_fuse_for_chain(
...
+        non_unit_dims_permuted = [
+            final_shape[dim] for dim in range(len(final_shape)) if final_shape[dim] != 1
+        ]
 
-        return is_equivalent(
-            cast(torch.fx.Node, producer.args[0]).meta["val"].shape,
-            get_dims(producer),
-            get_dims(consumer),
-        )
+        return non_unit_dims == non_unit_dims_permuted
```
ThomasJannaud, line 931: it seems that we are checking that the combination of 2 permutations is an identity based on the start and final size of the shapes, instead of on the permutation itself
(so if the input shape is [4 , 4, 4, 4] then this will always trigger)

ThomasJannaud:
closed in favor of #10566

PR #10514, opened 2025-04-28
Backends arm: Bump ethos-u/core_platform

perheld:
Bump core_platform to a newer version that includes the patch moving .rodata into DDR thus making the patch reduntant.

But, lurking inside that patch was a add of .got into DDR section so that needs its own patch to be handled.

Once the 0001 patch was fixed the 0002 patch needed some updates.

Change-Id: I453376b8ac8e05005983d15589ab8e42d3c7e0db


zingo (approved):

zingo:
lint error unrelated

PR #10519, opened 2025-04-28
Refactor _get_source_transforms to remove args

jackzhxng:
Refactor `_get_source_transforsm` to remove args

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -661,10 +661,37 @@ def _prepare_for_llama_export(args) -> LLMEdgeManager:
     logging.info(f"Checkpoint dtype: {edge_manager.model.checkpoint_dtype}")
     edge_manager = edge_manager.set_output_dir(output_dir_path).source_transform(
         _get_source_transforms(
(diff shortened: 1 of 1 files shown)


examples/models/llama/export_llama_lib.py
```diff
@@ -1196,12 +1259,25 @@ def _get_source_transforms(  # noqa
         transformations based on the given checkpoint first. In those cases,
         this wil be a no-op.
         """
-        modelname = f"{modelname}_e"
+        # Create a mock args object with the necessary attributes
```
iseeyuan, line 1262: Is it to limit the extend of changes? It's better than Namespace, but I'm wondering if eventually we need to group args, and there can still be passing the entire args even not all fields are needed.

iseeyuan (approved):
It's a great improvement from passing Namespace directly. Thanks @jackzhxng !

examples/models/llama/export_llama_lib.py
jackzhxng: Yup this is just to limit the changes, after I do a final refactor of the quant source transform functions we can remove this

PR #10523, opened 2025-04-28
Use python3 in .lintrunner.toml

shoumikhin:


kirklandsign (approved):

GregoryComer:
@shoumikhin FYI This breaks conda setup in a few places. Is it possible to conditionally use the correct python?

shoumikhin:
@GregoryComer how to repro? Fwiw, I don't use conda. This change is to make linter work on a Mac.

PR #10528, opened 2025-04-28
Pass one NDM to backend init

lucylq:
* #10529
* __->__ #10528

Take external NDM if it exists, otherwise internal.

Note: by default, xnnpack uses the named_data_map. Constants are not stored with the delegated blob anymore.

Differential Revision: [D73679710](https://our.internmc.facebook.com/intern/diff/D73679710/)


runtime/executor/method.cpp
```diff
@@ -805,6 +805,21 @@ Error Method::init(
...
+    int named_data_map_items =
+        (named_data_map != nullptr) ? named_data_map->get_num_keys().get() : 0;
+    int pte_map_items =
+        (pte_data_map != nullptr) ? pte_data_map->get_num_keys().get() : 0;
+
+    // TODO: implement ndm merge.
+    ET_CHECK_OR_RETURN_ERROR(
+        (pte_map_items == 0 || named_data_map_items == 0),
+        NotImplemented,
+        "NamedDataMap merge not supported. Both pte_data_map and named_data_map are non-empty, don't know what to do.");
```
JacobSzwejbka, line 817: nit: remove the dont know what to do part haha. Direct people seeing this crash to open an issue

runtime/executor/method.cpp
```diff
@@ -805,6 +805,21 @@ Error Method::init(
       pte_data_map = pte_data_map_res.get();
     }
 
+    int named_data_map_items =
```
JacobSzwejbka, line 808: we shouldnt use raw int anywhere. size_t or a fixed size prim like int32

JacobSzwejbka (changes requested):

mcr229 (approved):

PR #10541, opened 2025-04-29
Use std::string_view and std::optional  

cyyever:
### Summary
This PR moves ``executorch::aten::{string_view,optional}`` to std counterparts.


GregoryComer:
I had proposed this change previously, but @dbort brought up that the std types throw exceptions, which may be a problem for no-exception environments. Has there been more discussion on this point?

cyyever:
Currently `executorch::aten::{string_view,optional}` are aliases of std types, so they are not `noexcept`, they throw.

kernels/optimized/third-party/eigen
swolchok, line 1: revert?

swolchok (changes requested):
seems fine, but please put the submodule back and then ping me for accept

kernels/optimized/third-party/eigen
cyyever: No, I didn't touch third-party

cyyever:
@swolchok Ping

swolchok (approved):
thanks!

swolchok:
this appears to have changed numerics for test-llava-runner-linux: https://github.com/pytorch/executorch/actions/runs/15386371381/job/43336641649?pr=10541

however, that test is known to segfault from time to time. rerunning it one more time to see if the output change persists.

swolchok:
@larryliu0820 are you the llava PoC? this diff appears to change numerics slightly but only for the llava test; should that be blocking?

swolchok:
hm, this really shouldn't change numerics; I'm pretty sure that torch::executor::optional and torch::executor::string_view are the std equivalents in all cases. we can verify that by sending an interim PR that forces the exec_aten types to be typedefs for the std types in all cases.

swolchok:
OK, we landed #11321 , rebasing to see what CI has to say about that. (sorry for the delay; I was out for a couple days and am catching back up.)

cyyever:
@swolchok The current failures are unrelated

swolchok:
> @swolchok The current failures are unrelated

I agree; I'll rerun them once test-llava-runner-linux (very slow, part of the same workflow) is done

swolchok:
nothing alarming showing up internally! merging.

PR #10560, opened 2025-04-29
Add complex dtype support to mul

tarun292:
Summary: Adding support for complex dtypes in op_mul. In this diff i'm only adding support in the optimized path and not in the core portable kernels as that needs support in a bunch of the element-wise macro's that i'm not really familiar with. For now this unblocks our internal use cases.

Differential Revision: D73877440

### DIFF
--- a/kernels/test/op_mul_test.cpp
+++ b/kernels/test/op_mul_test.cpp
@@ -322,6 +322,38 @@ class OpMulOutTest : public OperatorTest {
     EXPECT_TENSOR_CLOSE(op_mul_out(a, b, out), expected);
     EXPECT_TENSOR_CLOSE(op_mul_out(b, a, out), expected);
   }
+
+  template <typename CTYPE, ScalarType DTYPE>
+  void test_complex_dtype() {
+    TensorFactory<DTYPE> tf;
+    const std::vector<int32_t> sizes = {2, 2};
+
+    // Create complex tensors with real and imaginary parts
+    Tensor x =
+        tf.make(sizes, {CTYPE(1, 2), CTYPE(3, 4), CTYPE(5, 6), CTYPE(7, 8)});
+
+    Tensor y =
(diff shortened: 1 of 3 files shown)


kernels/test/op_mul_test.cpp
```diff
@@ -472,6 +514,19 @@ TEST_F(OpMulOutTest, BothScalarInputBroadcastTest) {
...
 
+TEST_F(OpMulOutTest, AllComplexDtypesSupported) {
+#define TEST_ENTRY(ctype, dtype) test_complex_dtype<ctype, ScalarType::dtype>();
+  if (torch::executor::testing::SupportedFeatures::get()->is_aten) {
+    ET_FORALL_COMPLEX_TYPES(TEST_ENTRY);
+  } else {
+    // Disabling this for portable mode for now because currently we only support
+    // complex dtypes in optimized mode. Portable mode needs support in elementwise_ops.h
+    // for complex dtypes, which we don't have yet.
+    //ET_FORALL_COMPLEXH_TYPES(TEST_ENTRY);
```
manuelcandales, line 525: The fact that you are commenting out this line means none of the changes you added in this PR are being tested. I agree you shouldn't be blocked by adding support in the elementwise macros. You have two possible solutions:
a) Add your optimized-only tests under https://github.com/pytorch/executorch/tree/main/kernels/optimized/test
b) Enable complex in portable op_mul without messing around with the elementise macros, just add an if/else and add similar code to the one you added in the optimized op.
I think b) will be way easier

manuelcandales (changes requested):
Test your code changes either by doing a) or b)

kernels/test/op_mul_test.cpp
tarun292: @manuelcandales took approach b and updated the PR.

manuelcandales (approved):

PR #10561, opened 2025-04-29
[MPS] Add portable grid_sampler_2d implementation + tests

DenisVieriu97:
Upstream grid_sampler_2d (cpu implementation) and the tests implemented by @versi379

Summary of changes:
-  Add grid_sampler_2d implementation
- Add tests for grid_sampler_2d tests

cc @cccclai , @shoumikhin


tarun292:
@cccclai @manuelcandales bumping up for a review.

cccclai:
It's portable op implementation, can @manuelcandales or @swolchok take a look?

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+#include <utility>
+
+#include <c10/util/irange.h>
+
+namespace torch {
+namespace executor {
+namespace native {
```
swolchok, line 19: nit: we have C++17, new code should use nested namespaces (`namespace torch::executor::native {`)

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+using Tensor = exec_aten::Tensor;
+using ScalarType = executorch::aten::ScalarType;
+using SizesType = executorch::aten::SizesType;
```
swolchok, line 23: nit: sort

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+// Transform normalized coordinates to pixel space
+inline float unnormalize_coord(float coord, int64_t size, bool align_corners) {
```
swolchok, line 26: for internal functions like this one: remove inline, put in anonymous namespace

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+    // -1 and 1 correspond to the centers of the first and last pixels
+    return ((coord + 1.0f) / 2.0f) * (size - 1);
+  } else {
+    // -1 and 1 correspond to the boundary of the image
+    return ((coord + 1.0f) * size - 1.0f) / 2.0f;
+  }
+}
+
+// Compute source index and interpolation weight
+inline std::pair<int64_t, float>
```
swolchok, line 37: ditto inline

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+  if (size <= 1)
+    return 0;
+
+  int64_t double_size = 2 * size - 2;
+  if (double_size <= 0)
+    return 0;
+
+  // Handle negative coordinates
+  int64_t abs_coord = std::abs(coord);
+  abs_coord = abs_coord % double_size;
```
swolchok, line 56: super nit: I see a couple places in this PR where you could be terser by performing more than one operation per statement. for example, here would be better as `int64_t abs_coord = std::abs(coord) % double_size;`

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+    int64_t height,
+    int64_t width,
+    int64_t padding_mode) {
+  // Handle out-of-bounds coordinates
+  if (h < 0 || h >= height || w < 0 || w >= width) {
+    if (padding_mode == 0) { // Zeros
+      return 0;
+    } else if (padding_mode == 1) { // Border
+      h = std::min(std::max(h, static_cast<int64_t>(0)), height - 1);
+      w = std::min(std::max(w, static_cast<int64_t>(0)), width - 1);
```
swolchok, line 81: this looks like it's just std::clamp https://en.cppreference.com/w/cpp/algorithm/clamp

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+    Tensor& out) {
+  const int64_t N = input.size(0);
+  const int64_t C = input.size(1);
+  const int64_t out_H = grid.size(1);
+  const int64_t out_W = grid.size(2);
+
+  // Check for 4D input and grid
+  ET_KERNEL_CHECK(ctx, (input.dim() == 4), InvalidArgument, out);
+  ET_KERNEL_CHECK(ctx, (grid.dim() == 4), InvalidArgument, out);
+  ET_KERNEL_CHECK(ctx, (grid.size(3) == 2), InvalidArgument, out);
```
swolchok, line 219: please use ET_KERNEL_CHECK_MSG here (and below as well) to make it easy for people to debug these failures. The condition will be reported, but you need to tell them the actual value of input.dim()/grid.dim()/grid.size(3).

also please remove the comments that repeat the code

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+class OpGridSampler2dTest : public OperatorTest {
+ protected:
+  // Test grid_sampler_2d with bilinear interpolation
+  void test_grid_sampler_2d_bilinear() {
+    TensorFactory<ScalarType::Float> tf;
+
+    // Create a 1x1x2x2 input tensor
+    const std::vector<int32_t> input_sizes = {1, 1, 2, 2};
+    std::vector<float> input_data = {1.0, 2.0, 3.0, 4.0};
+    Tensor input = tf.make(input_sizes, input_data);
```
swolchok, line 27: nit: you don't need to name input_sizes and input_data here, just inline the initializer lists

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+    // Create a 1x1x1x2 grid tensor
+    const std::vector<int32_t> grid_sizes = {1, 1, 1, 2};
+    std::vector<float> grid_data = {0.0, 0.0};
+    Tensor grid = tf.make(grid_sizes, grid_data);
```
swolchok, line 32: ditto, and use tf.zeros instead of specifying data with all zeros

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+    // Create output tensor with expected shape 1x1x1x1
+    const std::vector<int32_t> output_sizes = {1, 1, 1, 1};
+    Tensor out = tf.zeros(output_sizes);
```
swolchok, line 36: ditto

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+    EXPECT_TENSOR_CLOSE(out, tf.make(output_sizes, expected_output));
+  }
+
+  // Test grid_sampler_2d with border padding mode
+  void test_grid_sampler_2d_border_padding() {
+    TensorFactory<ScalarType::Float> tf;
+
+    // Create a 1x1x4x4 input tensor
+    const std::vector<int32_t> input_sizes = {1, 1, 4, 4};
+    std::vector<float> input_data = {
```
swolchok, line 127: nit: you could use std::iota to make this a lot shorter
https://en.cppreference.com/w/cpp/algorithm/iota

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+    torch::executor::aten::grid_sampler_2d_outf(
+        context_,
+        input,
+        grid,
+        interpolation_mode,
+        padding_mode,
+        align_corners,
+        out);
+
+    // Check non-zero
```
swolchok, line 247: this comment just repeats the code, delete (and similar comments throughout that repeat the code)

kernels/test/op_grid_sampler_2d_test.cpp
```diff
@@ -0,0 +1,374 @@
...
+    EXPECT_NE(out.data_ptr<float>()[3], 0.0f);
+  }
+
+  // Test grid_sampler_2d with align_corners
+  void test_grid_sampler_2d_align_corners() {
+    TensorFactory<ScalarType::Float> tf;
+
+    // Create a 1x1x4x4 input tensor
+    const std::vector<int32_t> input_sizes = {1, 1, 4, 4};
+    std::vector<float> input_data = {
```
swolchok, line 260: ditto iota

shim_et/xplat/executorch/kernels/portable/op_registration_util.bzl
```diff
@@ -1330,4 +1337,4 @@ def portable_source_list():
 
 def portable_header_list():
     """All the header file names from //executorch/kernels/portable/cpu/"""
-    return ["selective_build.h", "scalar_utils.h", "math_constants.h", "vec_ops.h"]
+    return ["selective_build.h", "scalar_utils.h", "math_constants.h", "vec_ops.h"]
```
swolchok, line 1340: not 100% sure, but it looks like github is saying you removed the end-of-file newline? if so, please put it back

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+  // Check padding mode is valid (0=zeros, 1=border, 2=reflection)
+  ET_KERNEL_CHECK(
+      ctx, (padding_mode >= 0 && padding_mode <= 2), InvalidArgument, out);
+
+  // Check for output shape
+  ET_KERNEL_CHECK(
+      ctx,
+      (out.size(0) == N && out.size(1) == C && out.size(2) == out_H &&
+       out.size(3) == out_W),
```
swolchok, line 241: shouldn't we be attempting to resize output to the desired size so that we can handle dynamic shapes smoothly? @manuelcandales

swolchok (changes requested):
I'm not familiar with this operator and its mathematical function, but here are some general C++ comments

kernels/portable/cpu/op_grid_sampler_2d.cpp
manuelcandales: Yes.
@DenisVieriu97 please follow this [pattern](https://github.com/pytorch/executorch/blob/main/kernels/portable/cpu/op_pixel_shuffle.cpp#L85-L95)

kernels/test/targets.bzl
```diff
@@ -335,4 +336,4 @@ def define_common_targets():
     _common_op_test("op_view_as_real_copy_test", ["aten", "portable"])
     _common_op_test("op_view_copy_test", ["aten", "portable"])
     _common_op_test("op_where_test", ["aten", "portable"])
-    _common_op_test("op_zeros_test", ["aten", "portable"])
+    _common_op_test("op_zeros_test", ["aten", "portable"])
```
manuelcandales, line 339: newline?

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+    int64_t padding_mode,
+    bool align_corners,
+    Tensor& out) {
+  const int64_t N = input.size(0);
+  const int64_t C = input.size(1);
+  const int64_t out_H = grid.size(1);
+  const int64_t out_W = grid.size(2);
+
+  // Check for 4D input and grid
+  ET_KERNEL_CHECK(ctx, (input.dim() == 4), InvalidArgument, out);
```
manuelcandales, line 217: @DenisVieriu97 do you mind refactoring all of the InvalidArgument checks into a helper function dedicated to check the arguments, which returns True only if all checks pass, and logs when a check doesn't pass. Here is an [example](https://github.com/pytorch/executorch/blob/main/kernels/portable/cpu/util/upsample_util.cpp#L15-L39)
and then you would be calling
`ET_KERNEL_CHECK(ctx, check_grid_sampler_2d_args(...) == 4), InvalidArgument, out);`

kernels/portable/cpu/op_grid_sampler_2d.cpp
```diff
@@ -0,0 +1,257 @@
...
+  const int64_t channel_stride = height * width;
+  const int64_t height_stride = width;
+
+  return input_data
+      [n * batch_stride + c * channel_stride + h * height_stride + w];
+}
+
+// Process grid sampling with specified input, grid, and modes
+template <typename T>
+void grid_sampler_2d_impl(
```
manuelcandales, line 99: This implementation requires default dim order. Add a check in your check_args function, to check that the dim order is the default.

PR #10563, opened 2025-04-29
Introducing NXP Neutron runtime

jirioc:
### Summary

Introducing NXP Neutron runtime.


tarun292:
@digantdesai are you the POC for reviewing this on our end?

backends/nxp/runtime/NeutronDriver.h
```diff
@@ -0,0 +1,252 @@
...
+  void (*wait)(uint32_t channel);
+} NeutronConfig;
+
+/* Invalid handle, returned by neutronModelPrepare() if an error occurred. */
+#define NEUTRON_INVALID_HANDLE NULL
+
+/// - Initialize the Neutron Driver library, setting initial values, do memory
+/// allocation
+///   for internal data structures, do memory mapping.
+NeutronError neutronInit();
```
digantdesai, line 149: can these functions be in a `neutron` namespace? 
Where are we calling this specific function to init the lib?

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -0,0 +1,411 @@
...
+#include <executorch/runtime/core/error.h>
+#include <executorch/runtime/core/evalue.h>
+
+#include "NeutronDriver.h"
+#include "NeutronErrors.h"
+
+using namespace std;
+
+namespace torch {
+namespace executor {
```
digantdesai, line 20: neuton namespace?

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -0,0 +1,411 @@
...
+  NeutronModelConfig mcfg;
+  NeutronDataConfig dcfg;
+  NeutronModelHandle nmh = NULL;
+  const uint8_t* inputTranspositionFlags;
+  const uint8_t* outputTranspositionFlags;
+} NeutronConfig;
+
+// Applied on outputs.
+template <typename T>
+void transposeToChannelFirst(
```
digantdesai, line 62: OK for now, but I wonder is there another way to do this, i.e. set the dim_order on the output and let the portable or someone else take care of this?

digantdesai (commented):
Is this still in progress? 

At a high level looks OK to me.

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -0,0 +1,411 @@
...
+    //    cfg->numoutputs
+    //    cfg->mcfg.microcode
+    //    cfg->mcfg.weights
+    //    cfg->mcfg.kernels
+    const uint8_t* transpositionFlags =
+        static_cast<const uint8_t*>(processed->data());
+    uint32_t numInputs = transpositionFlags[INPUT_TENSOR_FORMAT_LEN_POS];
+    uint32_t numOutputs = transpositionFlags[OUTPUT_TENSOR_FORMAT_LEN_POS];
+    cfg->inputTranspositionFlags =
+        INPUT_TENSOR_FORMAT_ARRAY_ADDR(transpositionFlags);
```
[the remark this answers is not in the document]
jirioc: The logic is as follows: The partitioner keeps track of the channel order of each partition. If the embedded partition is incompatible with the surrounding order, the flags are added to perform the change. Similar approach was taken by other vendors [here](https://github.com/pytorch/executorch/blob/main/backends/arm/runtime/EthosUBackend.cpp#L423).

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -0,0 +1,411 @@
...
+ public:
+  NeutronBackend() {}
+
+  ~NeutronBackend() = default;
+
+  virtual bool is_available() const override {
+    return true;
+  }
+
+  Result<DelegateHandle*> init(
```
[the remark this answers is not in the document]
jirioc: Neutron init should be called once, whereas this init is called for each delegated partition. Therefore, at the moment the main.cpp is the place for neutron init.

backends/nxp/runtime/NeutronBackend.cpp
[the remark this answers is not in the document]
JakeStevens: I see, because this is only hitting the Neutron sub graph, and so we know the order, q/dq is taken care of outside of the delegate, and that is what is interfacing with our sensor streams bascially. Got it.

backends/nxp/runtime/NeutronDriver.h
JakeStevens: It is called in surrounding rtos code, much like ethosu_init

backends/nxp/runtime/NeutronBackend.cpp
JakeStevens: From what @jirioc explained below, I think this could be removed by inserting a transpose node instead of setting a flag

JakeStevens (commented):
LGTM, still outstanding seems to be:

1) namespace for neutron backend
2) consider removing NeutronDriver/NeutronErrors.h from ET repo and pull in to link separately like ARM does

digantdesai:
Any update on this?

robert-kalmar:
> LGTM, still outstanding seems to be:
> 
>     1. namespace for neutron backend
Thanks for advice, will do
>     2. consider removing NeutronDriver/NeutronErrors.h from ET repo and pull in to link separately like ARM does
So far we don't have the NeutronDriver/NeutronErrors.h. It is part of the ExecuTorch example in the [MCUXpresso SDK](https://github.com/nxp-mcuxpresso/mcuxsdk-middleware-executorch/tree/release/mcux/backends/nxp/runtime)  but I can imagine to have it distributed alongside with the Neutron Converter simulator or as a separated part of the SDK. Let us check internally and update as we discuss it internally.

digantdesai (commented):
I guess transpose is the only fix pending? I am ok with checking in header file for now.

JakeStevens:
> I guess transpose is the only fix pending? I am ok with checking in header file for now.

FWIW, it is how ARM does it currently as well. 

https://www.internalfb.com/code/fbsource/fbcode/executorch/backends/arm/runtime/EthosUBackend.cpp

If there are well supported tests for the runtime, this might be a good candidate for a task for new contributors interested in ET.

backends/nxp/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright 2024 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.19)
+project(neutron_backend)
```
jathu, line 7: Why does this need to be its own project?

backends/nxp/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+# Source root directory for executorch.
+if(NOT EXECUTORCH_ROOT)
+  set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
+endif()
+
+include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
```
jathu, line 16: If remove the standalone project, we can simplify this since the root project handles all the loading

```suggestion
```

backends/nxp/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+set(_common_include_directories ${EXECUTORCH_ROOT}/.. ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10)
```
jathu, line 18:
```suggestion
set(
  _common_include_directories 
  ${PROJECT_SOURCE_DIR}/.. 
  ${PROJECT_SOURCE_DIR}/runtime/core/portable_type/c10
)
```

robert-kalmar:
> > I guess transpose is the only fix pending? I am ok with checking in header file for now.
> 
> FWIW, it is how ARM does it currently as well.
> 
> https://www.internalfb.com/code/fbsource/fbcode/executorch/backends/arm/runtime/EthosUBackend.cpp
> 
> If there are well supported tests for the runtime, this might be a good candidate for a task for new contributors interested in ET.

Compared to Arm's approach we carry the transposition information with us in the Neutron Payload's header, as we get this by running the "tensor format inference" instead of Dim Orders.

backends/nxp/CMakeLists.txt
jathu: Sorry I'm new to cmake, but wouldn't `PROJECT_SOURCE_DIR` just be the first project it comes across? So the root level `executorch/CMakeList.txt` in this case

digantdesai (approved):

backends/nxp/CMakeLists.txt
jirioc: Correct. I am talking about situations where backends/nxp/CMakeLists.txt is included into the application which is in a separate repo constituting its own project.

digantdesai:
Let's re-merge the CI PR, this PR depends on that (touches trunk.yaml), then we can merge this. Thanks.

jirioc:
The runtime commits have been squashed and rebased.
The build test is now also working: https://github.com/pytorch/executorch/actions/runs/15912049098/job/44881583257?pr=10563

.github/workflows/trunk.yml
```diff
@@ -271,6 +271,36 @@ jobs:
...
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
+      script: |
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+
+        # Build
+        cmake -DEXECUTORCH_BUILD_NXP_NEUTRON=ON -Bcmake-out .
+        cmake --build cmake-out --target executorch_delegate_neutron --config Release
```
digantdesai, line 293: I guess compiling for x86 is OK for now

jirioc: Yes, it is.

PR #10573, opened 2025-04-30
Do not raise error when quant primitives are left after partitioner

metascroy:
Currently _sanity_check_graph_for_non_decomp_ops raises an Exception if a delegate asks an op for preservation, but doesn't lower it.

In general, this is a sensible thing to do, but for quant primitives, it is less sensible.  Since what gets lowered are patterns involving the quant primitives and FP32 ops.

XNNPACK asks that quant primitives be preserved, and so if a quant primitive is not lowered (e.g., it is part of embedding quant), an error is thrown.

In this PR, we:

* Define a central location for _QUANT_PRIMITIVES (with TODO task of moving this to torchao)
* Use these _QUANT_PRIMITIVES to avoid raising an error in _sanity_check_graph_for_non_decomp_ops
* Use _QUANT_PRIMITIVES in tracer.py to no decompose during to_edge and const_prop_pass to not constant propagate (this logic existed previously, but is being rewritten using the central _QUANT_PRIMITIVES list).

### DIFF
--- a/exir/TARGETS
+++ b/exir/TARGETS
@@ -15,8 +15,8 @@ python_library(
         ":types",
         "//caffe2:torch",
         "//executorch/exir/operator:convert",
+        "//executorch/exir/operator::util",
         "//executorch/extension/pytree:pylib",
-        "//pytorch/ao:torchao",
     ],
 )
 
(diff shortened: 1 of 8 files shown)


mcr229:
so are quant primitives by default preserved now? meaning i don't have to explicitly ask them to be preserved in XNNPACK?

metascroy:
> so are quant primitives by default preserved now? meaning i don't have to explicitly ask them to be preserved in XNNPACK?

The new ones should be as of earlier this week.

mcr229 (approved):

PR #10576, opened 2025-04-30
[ExecuTorch] Separate `extension.Module` Namespaces from Aten and non-Aten

zhenyan-zhang-meta:
* #10450
* __->__ #10576
* #10449

# Context

Separate `extension.Module` Namespaces to be `executorch::extension::module` and `executorch::extension::module::aten`, otherwise if a package relies on both aten and non-aten of the same implementation and the namespace is the same, there will be duplicate symbol issue like:

```
ld.lld: error: duplicate symbol: vtable for executorch::extension::Module
>>> defined at {redacted}/executorch/extension/module/__module__/__stripped__/module.cpp.pic.stripped.o:(vtable for executorch::extension::Module)
>>> defined at {redacted}/executorch/extension/module/__module_aten__/__stripped__/module.cpp.pic.stripped.o:
```

# Proposal

Doing something similar to what we already did for `bundled_program` in https://github.com/pytorch/executorch/pull/10307

Since `extension.Module` is a public API, we introduce a namespace alias, so that existing use cases won't get affected. Since namespace alias doesn't create additional symbols, there won't be duplicate symbol issue.

Differential Revision: [D73903870](https://our.internmc.facebook.com/intern/diff/D73903870/)


shoumikhin:
Something seems off, not sure we can just introduce macros like that in public APIs.
Need more details on how we got here and potentially amend the precedent in #10307 too.

zhenyan-zhang-meta:
@shoumikhin This is a fix to unblock future issues where we see targets using both aten and non-aten together and there will be issues like:
```
ld.lld: error: duplicate symbol: vtable for executorch::extension::Module
>>> defined at {redacted}/executorch/extension/module/__module__/__stripped__/module.cpp.pic.stripped.o:(vtable for executorch::extension::Module)
>>> defined at {redacted}/executorch/extension/module/__module_aten__/__stripped__/module.cpp.pic.stripped.o:
```

Also cc @larryliu0820 if you have other context that could justify separating namespaces.

Also cc @Gasoonjia

shoumikhin:
Sorry, I still don't get the bigger picture and need to look closer at where we're heading by introducing this pattern.
It looks like a workaround and I wonder if there are any alternatives to tackle this?
Also, @swolchok any thoughts?

swolchok:
I'm skeptical of the original goal to support mixing ATen and non-ATen mode builds of ExecuTorch files in the same binary. I suspect that that is just asking for ODR violations to crop up eventually.

zhenyan-zhang-meta:
@swolchok

> I suspect that that is just asking for ODR violations to crop up eventually.

Is this a yes to this change, since this change fixes the ODR violation of this module.

Seems like similar things has been done for all deps of `pybindings` that has an aten and non-aten module, I found one more: https://github.com/pytorch/executorch/pull/9894

cc @larryliu0820 @iseeyuan

zhenyan-zhang-meta:
@swolchok What we could do if the module dependent on both aten and non-aten is not owned by us? Should we discard this change, roll back https://github.com/pytorch/executorch/pull/9894 and https://github.com/pytorch/executorch/pull/10307, and let the module owner fix their part on failed tests?

shoumikhin:
I'd rolled back https://github.com/pytorch/executorch/pull/9894 and https://github.com/pytorch/executorch/pull/10307 and considered a different approach when we don't mix aten with not-aten

larryliu0820 (approved):
Not harmful and if it unblocks things why not

larryliu0820:
oh oops please address @shoumikhin 's comment

zhenyan-zhang-meta:
Closing this PR since it is not good to have aten and non-aten dependencies of a same model, thus we don't need to reconcile.

PR #10579, opened 2025-04-30
NXP backend: Create NeutronAtenPassManager with initial BatchNorm fusing passes

skywall:
### Summary
Add NeutronAtenPassManager that by default runs passes for optimal model conversion. Initially, only passes for BatchNorm + Conv/Linear fusing are added. I haven't intentionally moved passes into backend/transforms because added passes are executed on Aten ops and not Edge ops. Please let me know if I should follow different approach.

This pass manager is planned to be included in Neutron backend's pipeline - https://github.com/pytorch/executorch/pull/10196/files#diff-b3b0844670da322bb4c245ae440e8c71378d631bb94334148322cdf758369baaR51

### DIFF
--- a/backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
+++ b/backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
@@ -1,0 +1,146 @@
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from typing import Optional
+
+import torch
+from torch.export.unflatten import _assign_attr, _AttrKind
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch.nn.parameter import Parameter
+from torch.nn.utils import fuse_conv_bn_weights
+
+
+class FuseBatchNormWithConvPass(PassBase):
+    """The executorch batch normalization carries out the following computation [1].
+
+        (x - mean) / sqrt(var + eps) * W + B
+
+    Which can be expressed as
+
+        x * (W / sqrt(var + eps)) + (B - mean * (W / sqrt(var + eps)))
+
+    So the batch norm can be done as 1 multiplication and 1 addition, provided that the parameters are static,
+     and the terms can be precomputed. If there is a `Conv` operator before the batch normalization, this scale and
+     bias can be statically integrated into the weights and bias of the `Conv`, which allows the batch norm to be
+     completely removed.
+
+
+                               │
+                 ┌─────────────▼─────────────┐
+                 │ aten.conv1d | aten.conv2d │
+                 └─────────────┬─────────────┘
+                               │                                                                │
+         ┌─────────────────────▼─────────────────────┐        replace with        ┌─────────────▼─────────────┐
+         │              aten.batch_norm              │       ──────────────►      │ aten.conv1d | aten.conv2d │
+         └─────────────────────┬─────────────────────┘                            └─────────────┬─────────────┘
+                               │                                                                ▼
+                               ▼
+
+    [1] https://github.com/pytorch/executorch/blob/v0.5.0-rc2/kernels/portable/cpu/op_native_batch_norm.cpp#L118-L128
+    """
+
+    def _get_tensor_constant_from_node(self, graph_module, node) -> Parameter | None:
+        """Get the static data from a given node. If it doesn't have any data, return `None`."""
+        if node is None or node.op != "get_attr":
+            return None
+
+        target_atoms = node.target.split(".")
+        attr_itr = graph_module
+        for atom in target_atoms:
+            if not hasattr(attr_itr, atom):
+                return None
+            attr_itr = getattr(attr_itr, atom)
+        return attr_itr
+
+    def call(self, graph_module: GraphModule) -> Optional[PassResult]:
+        def _is_batch_norm(node_: Node) -> bool:
+            return (
+                node_.op == "call_function"
+                and node_.target == torch.ops.aten.batch_norm.default
+            )
+
+        def _is_conv(node_: Node):
+            is_conv = node_.op == "call_function" and node_.target in (
+                torch.ops.aten.conv1d.default,
+                torch.ops.aten.conv2d.default,
+            )
+            has_single_user = len(node.users) == 1
+
+            return is_conv and has_single_user
+
+        made_changes = False
+
+        if not any(map(_is_batch_norm, graph_module.graph.nodes)):
+            return PassResult(
+                graph_module, made_changes
+            )  # No batch norm nodes in the model.
+
+        for node in graph_module.graph.nodes:
+            if not _is_batch_norm(node):
+                continue  # Not BatchNorm.
+
+            bn_node = node
+
+            if not _is_conv(bn_node.args[0]):
+                continue  # Something other than a Conv node comes before the BatchNorm.
+
+            conv_node = bn_node.args[0]
+            conv_weight_node = conv_node.args[1]
+            conv_bias_node = conv_node.args[2] if len(conv_node.args) > 2 else None
+
+            # conv args: input, weight, bias, stride, padding, dilation, ...
+            conv_w = self._get_tensor_constant_from_node(graph_module, conv_weight_node)
+            conv_b = self._get_tensor_constant_from_node(graph_module, conv_bias_node)
+
+            # batch norm args: input, weight, bias, running mean, training, running var, momentum, eps
+            bn_w = self._get_tensor_constant_from_node(graph_module, bn_node.args[1])
+            bn_b = self._get_tensor_constant_from_node(graph_module, bn_node.args[2])
+            bn_rm = self._get_tensor_constant_from_node(graph_module, bn_node.args[3])
+            bn_rv = self._get_tensor_constant_from_node(graph_module, bn_node.args[4])
+            bn_eps = bn_node.args[7]
+
+            if any(
+                t is None for t in (conv_w, bn_rm, bn_rv)
+            ):  # The other inputs can be None.
+                continue  # The data is not static. Leave this BatchNorm as is (probably a rare case).
+            fused_weight, fused_bias = fuse_conv_bn_weights(
+                conv_w, conv_b, bn_rm, bn_rv, bn_eps, bn_w, bn_b
+            )
+
+            # Update the weight and bias for Conv.
+            conv_args = list(conv_node.args)
+            if len(conv_args) == 2:
+                # Fill in the default bias argument.
+                conv_args.append(None)
+
+            weight_attr_name = conv_weight_node.target
+            _assign_attr(
+                fused_weight, graph_module, weight_attr_name, _AttrKind.PARAMETER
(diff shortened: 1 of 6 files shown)


backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
```diff
@@ -0,0 +1,147 @@
...
+                graph_module, made_changes
+            )  # No batch norm nodes in the model.
+
+        for node in graph_module.graph.nodes:
+            if not _is_batch_norm(node):
+                continue  # Not BatchNorm.
+
+            bn_node = node
+
+            if not _is_linear(bn_node.args[0]):
```
JakeStevens, line 82: here and above, you can be defensive and also make sure args len == 1. there is nothing stopping someone from adding a skip connection after conv/linear and before batchnorm, or some other op. Can't say i've seen it in practice though

backends/nxp/aten_passes/neutron_aten_pass_manager.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Callable
+
+import torch
+
+from executorch.backends.nxp.aten_passes.fuse_batch_norm_with_conv_pass import (
```
JakeStevens, line 10: nit: can become unwieldly since there are a few other permutations of "batch norm" fusion, maybe just one file for fuse_batch_norm would be good.

JakeStevens:
LGTM but @digantdesai  thoughts on the aten pass setup?

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
skywall: Added check for number of users of Conv/Linear op.

skywall:
I have also disabled some tests that require https://github.com/pytorch/executorch/pull/10258 . We'll update PR that will be merged last.

backends/nxp/aten_passes/neutron_aten_pass_manager.py
skywall: Agree with you. Maybe we can create subdirectories in `backend/transforms` based on what dialect/ir is pass targeting?

backends/nxp/tests/test_batch_norm_fusion.py
```diff
@@ -0,0 +1,211 @@
...
+    "input_shape", [[4, 6, 8], [2, 4, 6, 8]], ids=lambda x: f"{len(x)}D"
+)
+def test_batch_norm_conv_fusing(bias: bool, input_shape: list[int]):
+    example_input = (torch.ones(*input_shape),)
+
+    module = ConvBatchNormModule(bias, len(input_shape), 4)
+    program = torch.export.export_for_training(module, example_input, strict=True)
+    og_module = program.module()
+
+    pm = NeutronAtenPassManager()
```
digantdesai, line 93: curious why does it have to be before `export`? Typically this is done in neutron.preprocess(). The reason is, it is not easy to run these nodes if Neutron partitioner fails to pick them up. There are several examples of this in Arm, XNNPACK etc.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
```diff
@@ -0,0 +1,150 @@
...
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch.nn.parameter import Parameter
+from torch.nn.utils import fuse_linear_bn_weights
+
+
+class FuseBatchNormWithLinearPass(PassBase):
+    """The executorch batch normalization carries out the following computation [1].
+
+        (x - mean) / (var + eps) * W + B
```
digantdesai, line 18:
```suggestion
        (x - mean) / sqrt(var + eps) * W + B
```

backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
```diff
@@ -0,0 +1,146 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from typing import Optional
+
+import torch
+from torch.export.unflatten import _assign_attr, _AttrKind
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch.nn.parameter import Parameter
+from torch.nn.utils import fuse_conv_bn_weights
```
digantdesai, line 12: glad you are using this as is

backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+
+            # Update the weight and bias for Conv.
+            conv_args = list(conv_node.args)
+            if len(conv_args) == 2:
+                # Fill in the default bias argument.
+                conv_args.append(None)
+
+            weight_attr_name = conv_weight_node.target
+            _assign_attr(
```
digantdesai, line 119: you can do this without using internal methods, and insert new params and delete old params https://github.com/pytorch/executorch/blob/main/backends/transforms/utils.py#L65

backends/nxp/tests/test_batch_norm_fusion.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+    # Verify that the behavior has not changed.
+    input_data = torch.randn(input_shape, dtype=torch.float32)
+    out1 = og_module(input_data).detach().numpy()
+    out2 = graph_module_out(input_data).detach().numpy()
+    assert np.allclose(out1, out2, atol=1.2e-7)
+
+
+# TODO - enable within PR #10196
```
digantdesai, line 150: move it to that PR?

digantdesai (commented):
Looks good, if we can move this either in `NeutronQuantizer.transform_for_annotation` (unless you are waiting for the other PR) with or move to `Neutron.Preprocess` it will be ready to go.

backends/nxp/tests/test_batch_norm_fusion.py
skywall: What `export` do you have in mind here? Export to lowered program?

These passes are executed before quantization on graph with nodes in ATen IR. If we try to run them later, pass will be much more complex because Quantize/Dequantize nodes will be already present in the graph.

backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
skywall: That would require to propagate exported program (via constructor?) to the pass. I don't find that particularly clean because passes are built upon GraphModules as you've mentioned in some of the previous PR's. We will end up with `ExportedProgram` being passed via constructor and `GraphModule` within `call()` function what doesn't sound right to me.

backends/nxp/tests/test_batch_norm_fusion.py
skywall: PR was merged, so I have included tests here.

backends/nxp/aten_passes/fuse_batch_norm_with_conv_pass.py
digantdesai: Yeah I guess this is for when you do run this pass in `delegate.preprocess`

backends/nxp/tests/test_batch_norm_fusion.py
digantdesai: Yeah not export_for_training but the export after quantization. 

And yeah typically backend specific graph modifications are done in `preprocess`, unless done specifically for quantization purposes. 

The rationale is, you can compose partitioners/backends together and not have to worry about someone else messing with the graph or you failing to partition your modified graph. For example Neutron + Cortex_m delegates can be composed by the user.

See other delegates doing bn fusion - https://github.com/search?q=repo%3Apytorch%2Fexecutorch+FuseBatchNorm&type=code 

I guess you are following what Cadence backend is doing, the main difference between Neutron and Cadence is that Cadence is not a delegate.

robert-kalmar: The `BN` + `Conv` fusion is done as Neutron does not (yet) have kernel for BN, but typically the BN is used with Convolution or Matrix multiplication, where it can be fused to. And because of the quantization we do the fusion on ATen graph level, so that the `NeutronQuantizer` already quantized with respect to the fused batchnorm. 

This reminds me about our discussion of quantizing the Conv/Matmul + BN+ (Relu) Activation together as one "cluster" (i.e. not apply observers on Conv and Relu separatelly). This would enable to move the batchnorm fusion even on the Edge dialect transformation in the NeutronBackend.
@skywall , can you please add a such a remark to the passes to address it with the quantizer update? And as suggested by @digantdesai, is is possible to move the transformation to the `NeutronQuantizer.transform_for_annotation` method as immediate resolution?

robert-kalmar: Appart from that, the rationale of backends composition makes completelly sense, I am curious how the quantizers composition is expected to be used by the user. Let say when combining the Neutron & Cortex-m backend the user will apply the backends per performance gain (that is e.g. Neutron first, cortex_m second) and apply the corresponding quantizers in the same order?

skywall:
Moved `NeutronAtenPassManager` call into `NeutronQuantizer.transform_for_annotation`.

skywall:
@digantdesai I just realized, BN + Conv fusion is already happening in `prepare_pt2e` [function call](https://github.com/pytorch/pytorch/blob/main/torch/ao/quantization/quantize_pt2e.py#L95) so our pass is doing kinda duplicate work. I don't think it's good practice to introduce optimizations randomly into the pipeline (`prepare_pt2e` function call) and it seems to me like possible candidate for removal in the future. What is your opinion? Should we leave the pass also here or we can rely on PyTorch that fusion will stay there as is.

backends/nxp/tests/test_batch_norm_fusion.py
digantdesai: yeah ok for now, please create an issue, and put it in the code as a TODO. The rationale is if you fuse before and for some reason can't lower to neutron, there is no core-aten op in the edge-ir as "conv+bn+relu" so it will fail as missing op vs. conv, bn, and relu individually will run OK.

digantdesai:
> I don't think it's good practice to introduce optimizations randomly into the pipeline

100% agree, Can you create an issue and tag @jerryzh168 and me?

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -202,4 +205,5 @@ def __init__(self):
     def transform_for_annotation(
         self, model: torch.fx.GraphModule
     ) -> torch.fx.GraphModule:
-        return model
+        pass_runner = NeutronAtenPassManager()
```
digantdesai, line 208: Thanks!

digantdesai (approved):
LGTM, please create issues and leave it as TODOs. Thanks.

digantdesai:
Arm test seems like an infra issue.

PR #10584, opened 2025-04-30
Qualcomm AI Engine Direct - multi-method support

haowhsu-quic:
### Summary
- refactor to adopt multi-method change
- framework patch to meet multi-method use case

### Test plan
```bash
python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_multi_graphs -s $device_sn -b build-android -m SM8750
```


exir/backend/backend_api.py
```diff
@@ -204,11 +205,38 @@ def _insert_lowered_submodule(
     owning_graph_module = call_submodule_node.graph.owning_module
     # call delegate args should only use user_inputs
     call_delegate_args = []
+    # handle getitem node in multi-method scenario
```
cccclai, line 208: Can you share what issues you run into?

haowhsu-quic: The scenario happens when there are common nodes shared by multiple delegated subgraphs (e.g. frequency sin/cos in sharded llama):
The example graph looks like below:
partitioned_0
When replacing submodule `fused_qnn_1`, the original name finding mechanism are trying to match between:
- submodule_program.graph_signature.user_inputs: `('aten_mean_dim', 'aten_select_int', 'aten_select_int_1')`
- call_submodule_node.all_input_nodes: `[aten_mean_dim, getitem_1, getitem_2]`

Which makes `getitem` node dangling as following, since the names could not be correctly mapped:
partitioned_1
The patch here is trying to find the original graph with real output names and use index of `getitem` to have them in correct order.

And I think another issue is about validation in `_unsafe_adjust_original_program`. Since the partitioned sub graphs in multi-method scenario have already been turned into submodules (like first diagram). That behavior will make `original_program._validate()` fail in `_unsafe_adjust_original_program`.
This does not happen in single method lowering because sub graphs are turned into submodules one by one and replaced into `executorch_call_delegate`.
But I cannot find an appropriate way to pass official CI, could you give me some hint? thank you!

cccclai: @mcr229 @angelayi  can you help double check the logic here?

haowhsu-quic: Thank you! There is another issue mentioned in last section, could you help check it as well?

> And I think another issue is about validation in `_unsafe_adjust_original_program`. Since the partitioned sub graphs in multi-method scenario have already been turned into submodules (like first diagram). That behavior will make `original_program._validate()` fail in `_unsafe_adjust_original_program`. This does not happen in single method lowering because sub graphs are turned into submodules one by one and replaced into `executorch_call_delegate`. But I cannot find an appropriate way to pass official CI, could you give me some hint? thank you!

mcr229: https://github.com/pytorch/executorch/pull/10708

haowhsu-quic: Thank you for helping resolve the issues.

larryliu0820: > Which makes getitem node dangling as following, since the names could not be correctly mapped:

I'm curious if DCE can prune the dangling getitem nodes. Have you tried to run the DCE pass right after the mapping is done?

cccclai:
It needs to rebase..

cccclai (approved):

exir/backend/backend_api.py
cccclai: I think the graph with dangling getitem nodes is not the correct behavior, so it's not about removing them.

kirklandsign (approved):

PR #10595, opened 2025-04-30
Update using-executorch-ios.md

shoumikhin:


mergennachin (approved):
@metascroy - i believe we can just update the docs page?

metascroy (approved):

PR #10596, opened 2025-04-30
Update doc-build.yml

metascroy:
Add ability to build docs with manual workflow

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -89,7 +89,7 @@ jobs:
 
   upload-gh-pages:
     needs: build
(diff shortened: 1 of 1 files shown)


mergennachin (approved):
What is `workflow_dispatch`?

metascroy:
> workflow_dispatch

I think it corresponds to the "run workflow" button under actions in the github web UI.  Maybe @kirklandsign can confirm?

kirklandsign:
> > workflow_dispatch
> 
> I think it corresponds to the "run workflow" button under actions in the github web UI. Maybe @kirklandsign can confirm?

That's correct. Have you validated? Different trigger has different ref names

PR #10612, opened 2025-05-01
[Executorch][llm] Enable local global attention in export_llama script

kimishpatel:
* __->__ #10612
* #10611
* #10610
* #10609
* #10608

Added a new option of --local_global_attention that takes in pattern of sizes to determine which layers are using local sliding window attention.
For example, [0, 256, 256, 0, 256, 256] can be used for 6 layers transformer. Or you can also use [0, 256, 256] as pattern you want
to repeat.

Differential Revision: [D73891423](https://our.internmc.facebook.com/intern/diff/D73891423/)


digantdesai (approved):

jackzhxng (commented):
From my understanding local global attention is tied closely to the model, and isn't something you can adjust given a particular checkpoint unlike these other export options such as `qmode`, `enable_kv_cache`, etc. I think it would be better to have this in [model_args](https://github.com/pytorch/executorch/blob/main/examples/models/llama/model_args.py) instead so that we can represent it in the config json file, so it is part of model configuration, not export configuration

kimishpatel:
> From my understanding local global attention is tied closely to the model, and isn't something you can adjust given a particular checkpoint unlike these other export options such as `qmode`, `enable_kv_cache`, etc. I think it would be better to have this in [model_args](https://github.com/pytorch/executorch/blob/main/examples/models/llama/model_args.py) instead so that we can represent it in the config json file, so it is part of model configuration, not export configuration

You can also make all the layers just do sliding window attention for example, so you can view this also as pure optimization. 

I havent thought a lot of about making this a model arg vs. export arg. I think your point definitely makes sense but I think it can be configurable as well if you want to support infinite generation as many runners do.

jackzhxng:
Hmm can a model trained without local(/global) attention run fine with local attention to take advantage of performance optimizations? If so and we view this as an export optimization then it might make sense and we can apply it to any arbitrary transformer. Main concern for me is that when we have this as an export option, then for models that require local/global attention become coupled to this export CLI option, Instead of just being able to represent it in the config json file, [like is done in Huggingface](https://huggingface.co/google/gemma-3-1b-it/blob/main/config.json#L32)

kimishpatel:
> Hmm can a model trained without local(/global) attention run fine with local attention to take advantage of performance optimizations? If so and we view this as an export optimization then it might make sense and we can apply it to any arbitrary transformer. Main concern for me is that when we have this as an export option, then models that require local/global attention become coupled to this export CLI option, instead of just being able to represent it in the config json file, [like is done in Huggingface](https://huggingface.co/google/gemma-3-1b-it/blob/main/config.json#L32)

I think you can even when the model is not trained. Using sliding window (e.g. attention sink https://arxiv.org/abs/2309.17453) is used to support much larger context. e.g. you may have a model trained with 128k context but you cant fully support that with such large kv cache

jackzhxng (approved):

PR #10623, opened 2025-05-01
Refactor attention v2

lucylq:
* __->__ #10623

Pull attention creation out of Transformer/TransformerBlock. Instead, pass the layers into Transformer.

The motivation is to customize linear layers in attention for LoRA (eg. make wq into a LoraLinear instead of a regular linear). In the next diff (D73517350), we pull wq,wk,wv,wo out of the attention and pass those in as well.

This allows us to customize attention parameters without passing in ModelArgs and doing the customization deep inside attention.py.

I think this modularizes our attention/transformer components, though also means that users have to do some more work to construct the attention layers and pass it to transformer.

It follows the torchtune structure more closely, eg. https://github.com/pytorch/torchtune/blob/main/torchtune/models/llama3_2/_component_builders.py#L221

Differential Revision: [D73538697](https://our.internmc.facebook.com/intern/diff/D73538697/)


iseeyuan (approved):
It makes sense to me to pass attention to the transformer, and unblock LoRA.

examples/models/llama/llama_transformer.py
```diff
@@ -83,25 +84,30 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
 
 class TransformerBlock(nn.Module):
-    def __init__(self, layer_id: int, args: ModelArgs, rope: Rope):
+    def __init__(self, args: ModelArgs, attention: Attention):
```
iseeyuan, line 87: Can you add doc string on each argument, especially the attention? I think it makes sense to me that Attention type is required, so that the API of user-defined attention is compatible with our transformer.

examples/models/llama/llama_transformer.py
```diff
@@ -117,7 +138,15 @@ def forward(self, x, freqs_cos, freqs_sin, attn_options: ForwardOptions):  # x:
 
 
 class Transformer(nn.Module):
-    def __init__(self, params: ModelArgs):
+    def __init__(self, params: ModelArgs, layers: nn.ModuleList, rope: Rope):
```
jackzhxng, line 141: I think if you are going to do this, might as well lift all of the major model components out as well, such as the embedding layer and rms norm, even though they are not customizable by model args at the moment

examples/models/llama/llama_transformer.py
```diff
@@ -212,3 +239,23 @@ def forward(
             return logits, attn_options_update
 
         return logits
+
+
+def construct_transformer(model_args: ModelArgs) -> Transformer:
```
jackzhxng, line 244: Why not `@classmethod`?

examples/models/llama/llama_transformer.py
lucylq: I think we can, but would prefer to do that in a separate PR if it's something we want to do. Is there a use-case, or more to make Transformer even more modular?

jackzhxng: Up to you, no use case atm, just for modularity. Just feels a bit weird to me seeing layers and rope be the only lifted inputs for Transformer

examples/models/llama/llama_transformer.py
lucylq: discussed offline; construct_transformer is likely going to be more high-level; not quite at model-creation, but will contain eg. lora instantiation so may not make sense for it to be part of the transformer class itself.

PR #10638, opened 2025-05-01
Avoid directly calling the map operator

ydwu4:
Summary: Calling the higher order operator directly is not recommended since it skips the safety check.

Differential Revision: D74036464


angelayi (approved):

JacobSzwejbka:
is this true for cond? 

I should stop pointing people to torch.cond and instead to torch_cond and similar for all control flow ops?

PR #10652, opened 2025-05-02
Update install script and building from source docs

### LINKED ISSUE #9772: Return "platform not supported" when using PyTorch on intel-based Macbooks

Currently, when a user installs requirements for ExecuTorch on their Intel-chip Macbook by running the `./install_requirements.sh `after git-cloning the ExecuTorch repo, the installation fails for torch versions, throwing an error and a stack trace.

`ERROR: Could not find a version that satisfies the requirement torch==2.7.0.dev20250310 (from versions: 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, 2.2.0, 2.2.1, 2.2.2)
ERROR: No matching distribution found for torch==2.7.0.dev20250310`

Image

### Alternatives

1) Return a message stating "this platform is not supported" while running PyTorch on Intel Macbooks
2) Updating documentation to add a disclaimer or in pre-requisites that PyTorch does not work anymore on Intel Macbooks.

### Additional context

_No response_

### RFC (Optional)

_No response_

keyprocedure:
### Summary
- Updated `install_requirements.sh` to improve compatibility with source-built PyTorch and Intel macOS systems.
- Added checks to prevent install commands from altering existing PyTorch builds by adding the `--no-deps` flag.
- Removed `torchaudio` and `torchvision` dependencies  for Intel macOS systems to avoid compatibility issues.

Fixes #9772 

### Test plan
Tested executorch installations with source-built PyTorch on:
- Intel-macOS with PyTorch 2.7.0, Python 3.12.9
- ARM64 macOS with PyTorch 2.8.0a0, Python 3.12.9  

Validated by exporting and running the MobileNet V2 image classification model, following the examples on the [Getting Started with ExecuTorch](https://pytorch.org/executorch/stable/getting-started.html) page.

cc @larryliu0820 @jathu

### DIFF
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -61,7 +61,7 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
    # Install ExecuTorch pip package and its dependencies, as well as
    # development tools like CMake.
    # If developing on a Mac, make sure to install the Xcode Command Line Tools first.
-   # Intel-based macOS systems require building PyTorch from source (see below)
+   # Intel-based macOS systems require building PyTorch, Torchvision, and Torchaudio from source (see below)
    ./install_executorch.sh
    ```
   
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -16,7 +16,7 @@ Linux (x86_64)
 - Ubuntu 20.04.6 LTS+
 - RHEL 8+
 
-macOS (x86_64/M1/M2)
+macOS (x86_64/ARM64)
 - Big Sur (11.0)+
 
 Windows (x86_64)
@@ -56,12 +56,20 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
    conda create -yn executorch python=3.10.0 && conda activate executorch
    ```
 
-## Install ExecuTorch pip package from Source
+## Install ExecuTorch pip package from source
    ```bash
    # Install ExecuTorch pip package and its dependencies, as well as
    # development tools like CMake.
    # If developing on a Mac, make sure to install the Xcode Command Line Tools first.
+   # Intel-based macOS systems require building PyTorch from source (see below)
    ./install_executorch.sh
+   ```
+  
+  Use the [`--use-pt-pinned-commit` flag](https://github.com/pytorch/executorch/blob/main/install_executorch.sh) to install Executorch with an existing PyTorch build.  
+  See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.
+
+   ```bash
+   ./install_executorch.sh --use-pt-pinned-commit
    ```
 
    Not all backends are built into the pip wheel by default. You can link these missing/experimental backends by turning on the corresponding cmake flag. For example, to include the MPS backend:
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -61,22 +61,22 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
    # Install ExecuTorch pip package and its dependencies, as well as
    # development tools like CMake.
    # If developing on a Mac, make sure to install the Xcode Command Line Tools first.
-   # Intel-based macOS systems require building PyTorch, Torchvision, and Torchaudio from source (see below)
+   # Intel-based macOS systems require building PyTorch from source (see below)
    ./install_executorch.sh
    ```
   
-  See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.  
+   See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.  
 
-  Use the [`--use-pt-pinned-commit` flag](../../install_executorch.py) to install ExecuTorch with an existing PyTorch build:
+   Use the [`--use-pt-pinned-commit` flag](../../install_executorch.py) to install ExecuTorch with an existing PyTorch build:
 
    ```bash
    ./install_executorch.sh --use-pt-pinned-commit
    ```
 
-  For Intel-based macOS systems, use the [`--use-pt-pinned-commit --minimal` flags](../../install_executorch.py):
-  ```bash
-  ./install_executorch.sh --use-pt-pinned-commit --minimal
-  ```
+   For Intel-based macOS systems, use the [`--use-pt-pinned-commit --minimal` flags](../../install_executorch.py):
+   ```bash
+   ./install_executorch.sh --use-pt-pinned-commit --minimal
+   ```
 
    Not all backends are built into the pip wheel by default. You can link these missing/experimental backends by turning on the corresponding cmake flag. For example, to include the MPS backend:
 
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -65,12 +65,18 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
    ./install_executorch.sh
    ```
   
-  Use the [`--use-pt-pinned-commit` flag](https://github.com/pytorch/executorch/blob/main/install_executorch.sh) to install Executorch with an existing PyTorch build.  
-  See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.
+  See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.  
+
+  Use the [`--use-pt-pinned-commit` flag](../../install_executorch.py) to install ExecuTorch with an existing PyTorch build:
 
    ```bash
    ./install_executorch.sh --use-pt-pinned-commit
    ```
+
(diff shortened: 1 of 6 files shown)


install_requirements.py
```diff
@@ -139,6 +152,24 @@ def install_requirements(use_pytorch_nightly):
...
+def is_intel_mac_os():
+    # Returns True if running on Intel macOS
+    if platform.system().lower() == "darwin" and platform.machine().lower() in (
+        "x86",
+        "x86_64",
+        "i386",
+    ):
+        print(
+            "ERROR: Prebuilt PyTorch wheels are no longer available for Intel-based macOS.\n"
+            "Please build from source by following https://pytorch.org/executorch/0.6/using-executorch-building-from-source.html",
```
mergennachin, line 166: https://pytorch.org/executorch/0.6/ -> change this to https://pytorch.org/executorch/main/

mergennachin, line 166: The print message should appear inside this block.

```
  if use_pytorch_nightly and is_intel_mac:
```

`is_intel_mac_os()` is just a getter function that should do any printing.

install_requirements.py
```diff
@@ -75,45 +75,58 @@ def python_is_compatible():
...
+
+    # Only pip install torchvision and torchaudio if not building PyTorch from source on Intel macOS,
+    # to avoid installing incompatible prebuilt wheels from PyPI.
+    if not is_intel_mac:
+        EXIR_REQUIREMENTS.append(
+            f"torchvision==0.22.0.{NIGHTLY_VERSION}"
+        )  # For testing.
+
+        EXAMPLES_REQUIREMENTS = [
+            f"torchaudio==2.6.0.{NIGHTLY_VERSION}",
```
mergennachin, line 102: what about this logic? 

```
f"torchaudio==2.6.0.{NIGHTLY_VERSION}" if use_pytorch_nightly else "torchaudio"
```

mergennachin (changes requested):
Thank you @keyprocedure 

Overall looks great. Thank you for testing end-to-end. Glad that it works. 

Please see inline comments below.

install_requirements.py
keyprocedure: I had removed it for testing and forgot to restore it. Updated the code :)

keyprocedure:
Hi @mergennachin, I’ve made the requested changes — could you please take a look when you get a chance?

mergennachin:
Thanks @keyprocedure

Just started the full CI tests

keyprocedure:
@mergennachin 
It looks like CI is failing because the `--no-deps` flag prevents the required modules from being installed. I used it to avoid torch from being upgraded, but that's leading to missing modules

Is the expectation that any torch version should be usable, or only ones compatible with the current dependencies?

If any torch version should be usable, would it make more sense to use a temporary `constraints.txt` file to pin the torch version instead of the `--no-deps` flag? That may still lead to breakages, but those conflicts would be clearer during the install process

mergennachin:
@keyprocedure - First of all, thanks for being patient with this PR.

Looking a bit more into it. Why do we need `--no-deps` in the first place? 

Presumably, when you install torch from source manually as first step, it should install a newer version from source. And in the install_requirements.py step, perhaps it won't try to upgrade it. Do you know which package is transitively trying to upgrade it?

keyprocedure:
@mergennachin Sorry for the delay 

It looks like the torch version is actually downgraded to fit the dependency requirements. I'm not sure which package is causing the version change, but I recently tried to install executorch with the latest torch build without the no-deps flag and after running `./install_executorch.sh --use-pt-pinned-commit`, torch gets downgraded to 2.7.0:

```
Installing collected packages: zstd, sentencepiece, pytz, antlr4-python3-runtime, xxhash, tzdata, tqdm, tomli, safetensors, regex, python-dateutil, pycryptodomex, pyarrow, propcache, pillow, omegaconf, multidict, lxml, hf-xet, hf-transfer, fsspec, frozenlist, dill, cmake, aiohappyeyeballs, yarl, `torch`, tiktoken, pandas, multiprocess, kagglehub, huggingface_hub, blobfile, aiosignal, torchvision, torchdata, torchaudio, tokenizers, aiohttp, transformers, torchsr, timm, datasets, torchtune
Attempting uninstall: torch
Found existing installation: torch 2.8.0a0+git9a8c42f
Uninstalling torch-2.8.0a0+git9a8c42f:
Successfully uninstalled torch-2.8.0a0+git9a8c42f
```

keyprocedure:
If the downgrade is being caused by something that I'm doing incorrectly during the install process/isolated to my build, I can remove the no-deps flag from the pip install command for this PR. I'm open to any suggestions/guidance.

mergennachin:
Hi @keyprocedure 

I did some digging and found out that torchvision and torchaudio were in fact downgrading. 

torch, torchvision and torchaudio are in lock step dependency.

I did some refactoring in (https://github.com/pytorch/executorch/pull/11653). torchvision and torchaudio are not absolute necessity. Only used for examples.

Once that PR is landed, we can update the Intel instruction to be as follows:

```
Build torch from source
./install_executorch.sh --use-pt-pinned-commit --minimal
```

keyprocedure:
Hey @mergennachin, I really appreciate your follow up! 

Perfect, that'll simplify things. 
I'll update the Intel instructions in this PR once #11653 lands.

mergennachin:
Hi @keyprocedure 

I just landed the PR, please rebase. Hopefully it will simplify a lot of things in this PR

keyprocedure:
@mergennachin I've rebased and updated the Intel macOS install instructions to `./install_executorch.sh --use-pt-pinned-commit --minimal`. Feel free to let me know if there's anything else that needs to be   changed.

docs/source/using-executorch-building-from-source.md
```diff
@@ -56,13 +56,21 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
...
    # Install ExecuTorch pip package and its dependencies, as well as
    # development tools like CMake.
    # If developing on a Mac, make sure to install the Xcode Command Line Tools first.
+   # Intel-based macOS systems require building PyTorch, Torchvision, and Torchaudio from source (see below)
    ./install_executorch.sh
    ```
+  
+  Use the [`--use-pt-pinned-commit` flag](https://github.com/pytorch/executorch/blob/main/install_executorch.sh) to install Executorch with an existing PyTorch build.  
+  See the [PyTorch instructions](https://github.com/pytorch/pytorch#installation) on how to build PyTorch from source.
+
```
mergennachin: Use relative paths

docs/source/using-executorch-ios.md
```diff
@@ -114,6 +114,13 @@ python3 -m venv .venv && source .venv/bin/activate && pip install --upgrade pip
 ./backends/apple/coreml/scripts/install_requirements.sh
 ```
 
+- **Intel-based macOS** systems require building PyTorch, Torchvision, and Torchaudio from source:
```
mergennachin: Let's not have two instructions. Let's consolidate Intel based instructions in "building from source section" and call it a day

docs/source/using-executorch-building-from-source.md
```diff
@@ -56,13 +56,21 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
...
+   ```bash
+   ./install_executorch.sh --use-pt-pinned-commit
+   ```
```
mergennachin, line 74: Insert the Intel-based instruction under here... 

Something like: 

For Intel-based instructions, use `./install_executorch.sh --use-pt-pinned-commit --minimal`

docs/source/using-executorch-building-from-source.md
```diff
@@ -56,13 +56,21 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
...
+## Install ExecuTorch pip package from source
    ```bash
    # Install ExecuTorch pip package and its dependencies, as well as
    # development tools like CMake.
    # If developing on a Mac, make sure to install the Xcode Command Line Tools first.
+   # Intel-based macOS systems require building PyTorch, Torchvision, and Torchaudio from source (see below)
    ./install_executorch.sh
    ```
+  
+  Use the [`--use-pt-pinned-commit` flag](https://github.com/pytorch/executorch/blob/main/install_executorch.sh) to install Executorch with an existing PyTorch build.
```
mergennachin: s/Executorch/ExecuTorch/g

mergennachin (commented):
See inline

docs/source/using-executorch-building-from-source.md
keyprocedure: I updated the `flag` links to use relative paths to the `.py` file.
For the `PyTorch instructions` link, since it points to a doc in the PyTorch repo, is there a way to reference that using a relative path, or should we keep the current path?

docs/source/using-executorch-ios.md
keyprocedure: Good point, moved the instructions.

keyprocedure:
I removed the mention of building `Torchvision` and `Torchaudio` from source in the [building-from-source](https://github.com/pytorch/executorch/pull/10652/files#diff-05258271df7e81c17add5e54131250685b6014736847d203c54b51d559e1d29eR64) doc for `Intel-macOS` builds to simplify the instructions, since they're not required to install ExecuTorch.

mergennachin (approved):
Thank you @keyprocedure

keyprocedure:
Happy to help!

PR #10653, opened 2025-05-02
Arm backend: Add support to neg.default

fumchin:
This patch adds support to aten.neg.default via TOSA NEGATE and added test cases.


zingo (approved):

zingo:
macos fails unrelated

PR #10654, opened 2025-05-02
Arm backend: Add support for single input matmul

oscarandersson8218:
### Summary

AnnotateDecomposedMatmul makes sure that a decomposed matmul will two dq-nodes before and a q-node after it's mm/bmm-node. Previously it assumed that the partition always had two input nodes (two dq-nodes), but this is not the case for a single input matmul, e.g. torch.matmul(x, x). In such a case we must copy the dq-node and insert it before the mm/bmm's two inputs.

```
Before pass:
         -> expand -> view ->
       /                      \
x -> dq                        bmm -> view -> q
       \                      /
         -> expand -> view ->

After pass:
   -> expand -> view -> dq
 /                        \
x                          bmm -> q -> view
 \                        /
   -> expand -> view -> dq
```


zingo:
MacOS fails is unrelated

zingo (approved):

PR #10655, opened 2025-05-02
Arm backend: Refactor TosaArg to use TosaSpecification

oscarandersson8218:
### Summary
Make TosaArg aware of which TosaSpecification is used. This is used to map torch-dtypes to TOSA-dtypes.


zingo:
@oscarandersson8218 This seem to need a rebase

zingo (approved):

PR #10659, opened 2025-05-02
support phi4 in ios demo app

metascroy:
As titled

### DIFF
--- a/examples/demo-apps/apple_ios/LLaMA/README.md
+++ b/examples/demo-apps/apple_ios/LLaMA/README.md
@@ -37,7 +37,7 @@ Download already exported LLaMA/LLaVA models along with tokenizers from [Hugging
 
 4. To run on a device, ensure you have it set up for development and a provisioning profile with the `increased-memory-limit` entitlement. Update the app's bundle identifier to match your provisioning profile with the required capability.
 
-5. After successfully launching the app, copy the exported ExecuTorch model (`.pte`) and tokenizer (`.model`) files to the iLLaMA folder. Three models are currently supported at the moment - Llama, Qwen3, Phi4-mini, and Llava multimodal. Please ensure that your model `.pte` file starts with `llama`, `qwen3`, `phi4` or `llava` so that the app selects the correct model type.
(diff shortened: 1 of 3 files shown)


examples/demo-apps/apple_ios/LLaMA/README.md
```diff
@@ -32,12 +32,12 @@ Download already exported LLaMA/LLaVA models along with tokenizers from [Hugging
...
     open examples/demo-apps/apple_ios/LLaMA/LLaMA.xcodeproj
     ```
-    
+
 3. Click the Play button to launch the app in the Simulator.
 
 4. To run on a device, ensure you have it set up for development and a provisioning profile with the `increased-memory-limit` entitlement. Update the app's bundle identifier to match your provisioning profile with the required capability.
 
-5. After successfully launching the app, copy the exported ExecuTorch model (`.pte`) and tokenizer (`.model`) files to the iLLaMA folder. Three models are currently supported at the moment - Llama, Qwen3, and Llava multimodal. Please ensure that your model `.pte` file starts with `llama`, `qwen3`, or `llava` so that the app selects the correct model type.
+5. After successfully launching the app, copy the exported ExecuTorch model (`.pte`) and tokenizer (`.model`) files to the iLLaMA folder. Three models are currently supported at the moment - Llama, Qwen3, Phi4-mini, and Llava multimodal. Please ensure that your model `.pte` file starts with `llama`, `qwen3`, `phi4` or `llava` so that the app selects the correct model type.
```
jackzhxng, line 40:
```suggestion
5. After successfully launching the app, copy the exported ExecuTorch model (`.pte`) and tokenizer (`.model`) files to the iLLaMA folder. Four models are currently supported at the moment - Llama, Qwen3, Phi4-mini, and Llava multimodal. Please ensure that your model `.pte` file starts with `llama`, `qwen3`, `phi4` or `llava` so that the app selects the correct model type.
```

jackzhxng (approved):
Any special tokens to handle?

metascroy:
> Any special tokens to handle?

no, the <end> token is handled by the runner.cpp logic because it's in the PTEs metadata.

But we should really move this logic (and the template) to the HF tokenizer.  The tokenizer.json contains information about both stop tokens and template.

jackzhxng:
Yeah we should definitely do that, but I think the prompt templating should be outside of the tokenizer

PR #10660, opened 2025-05-02
Use std::align_alloc in file_data_loader

lucylq:
Summary:
Issue with aligned buffers: P1800967583

The alignment requested is 16, and std::max_align_t is also 16. This means we do not need to pad the size to meet any alignment.

However, the buffer we get from malloc is aligned to 8, not 16. When we try to align the buffer to 16, we overflow the original buffer (as it wasn't padded) and error out.

Seems like malloc is not guaranteed to return 8 or 16 byte-aligned buffers, so also a bit hard to test definitively. So far we've only seen this when the buffer size is small (size 2, 4)
```
The malloc(), calloc(), realloc(), and reallocarray() functions
return a pointer to the allocated memory, which is suitably
aligned for any type that fits into the requested size or less.
```

Use std::aligned_alloc (C++17) to ensure buffer is aligned.

Differential Revision: D74041198

### DIFF
--- a/runtime/platform/compiler.h
+++ b/runtime/platform/compiler.h
@@ -171,6 +171,101 @@
 using ssize_t = ptrdiff_t;
 #endif
 
+/**
+ * Platform-specific aligned memory allocation and deallocation.
+ *
+ * Usage:
+ *   void* ptr = ET_ALIGNED_ALLOC(alignment, size);
+ *   // use ptr...
+ *   ET_ALIGNED_FREE(ptr);
+ *
+ * Note: alignment must be a power of 2 and size must be an integral multiple of
+ * alignment.
+ */
+#if defined(_MSC_VER)
+#include <malloc.h>
+#define ET_ALIGNED_ALLOC(alignment, size) \
+  _aligned_malloc(((size + alignment - 1) & ~(alignment - 1)), (alignment))
+#define ET_ALIGNED_FREE(ptr) _aligned_free(ptr)
+#elif defined(__APPLE__)
+#include <stdlib.h> // For posix_memalign and free
+inline void* et_apple_aligned_alloc(size_t alignment, size_t size) {
+  void* ptr = nullptr;
+  // The address of the allocated memory must be a multiple of sizeof(void*).
+  if (alignment < sizeof(void*)) {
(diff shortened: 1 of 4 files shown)


larryliu0820 (approved):
Is it possible to add some tests?

lucylq:
> Is it possible to add some tests?

alignment is covered by the file_data_loader_tests: https://github.com/pytorch/executorch/blob/3ca6cbbff564d52bb5c206539d25fadb5f1eb854/extension/data_loader/test/file_data_loader_test.cpp#L68

I think we shouldn't have this error now that we've moved to aligned_alloc (hopefully), probably the main thing is to make sure oss ci passes on macos.

Are you thinking of a different test though?

swolchok:
> Seems like malloc is not guaranteed to return 8 or 16 byte-aligned buffers

it is [*supposed* to be guaranteed](https://en.cppreference.com/w/cpp/memory/c/malloc), but [reality does not always conform to standards](https://github.com/jemalloc/jemalloc/issues/1533#issuecomment-507915829) :\

runtime/platform/compiler.h
```diff
@@ -171,6 +171,101 @@
...
+  // The address of the allocated memory must be a multiple of sizeof(void*).
+  if (alignment < sizeof(void*)) {
+    alignment = sizeof(void*);
+  }
+  if (posix_memalign(
+          &ptr, alignment, (size + alignment - 1) & ~(alignment - 1)) != 0) {
+    return nullptr;
+  }
+  return ptr;
+}
```
swolchok, line 203: would be nice to include citations for why we need all these varied solutions despite std::aligned_alloc being required in C++17.  https://gitlab.com/gromacs/gromacs/-/issues/3968 is one example but I'm curious where you got this from.

lucylq: Ah from: https://www.internalfb.com/code/fbsource/[0afaf0bdc6ee692ec9a9b3e3c57bca4af2930abc]/third-party/pocket_fft/pocketfft_hdronly.h?lines=156%2C165

And mostly from ci tests - xtensa seems to only have malloc available? Thanks for the link!

PR #10670, opened 2025-05-03
Handle unsupported pybind inputs non-fatally

GregoryComer:
Summary: Currently, passing unsupported python types to pybind method execution, such as lists, dicts, or tuples, will crash the kernel due to hitting an assert. This PR updates the logic to raise an exception, which gets nicely bubbled up to the notebook. This gives the user a nicer error message and does not crash the bento/jupyter process.

Differential Revision: D74118509


tarun292 (approved):

tarun292:
This is great. Thanks for adding this.

PR #10675, opened 2025-05-04
Introduce PAL function table

GregoryComer:
Summary:
This is part 1 of the implementation of the PAL changes described in https://github.com/pytorch/executorch/discussions/10432. This PR introduces a struct (`pal_table`) to hold function pointers to the PAL function implementations. There is a singleton instance of this struct, which is initialized with the weak/strong et_pal_ functions - maintaining backwards compatibility with the existing override mechanism.

I've then added wrapper functions for the PAL into the executorch::runtime namespace which dispatch through the function table. It is intended that callers use these functions instead of the "raw" et_pal_ methods in order to correctly dispatch through the function table. In the following PR, I update ET callers to do this.

Differential Revision: D74121895

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -371,7 +371,7 @@ jobs:
         size=${arr[4]}
         # threshold=48120 on devserver with gcc11.4
         # todo(lfq): update once binary size is below 50kb.
-        threshold="55504"
+        threshold="55584"
         if [[ "$size" -le "$threshold" ]]; then
           echo "Success $size <= $threshold"
         else
@@ -406,7 +406,7 @@ jobs:
         output=$(ls -la cmake-out/test/size_test)
         arr=($output)
         size=${arr[4]}
-        threshold="51656"
+        threshold="51728"
         if [[ "$size" -le "$threshold" ]]; then
           echo "Success $size <= $threshold"
         else
--- a/runtime/platform/platform.cpp
+++ b/runtime/platform/platform.cpp
@@ -1,0 +1,160 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/runtime/platform/log.h>
+#include <executorch/runtime/platform/platform.h>
+#include <cstdlib>
+
+namespace executorch::runtime {
+
+namespace {
+/**
+ * The singleton instance of the PAL function table.
+ */
+PalImpl pal_impl = {
+    et_pal_init,
+    et_pal_abort,
+    et_pal_current_ticks,
+    et_pal_ticks_to_ns_multiplier,
+    et_pal_emit_log_message,
+    et_pal_allocate,
+    et_pal_free,
+    __FILE__};
+
+/**
+ * Tracks whether the PAL has been overridden. This is used to warn when
+ * multiple callers override the PAL.
+ */
+bool is_pal_overridden = false;
+} // namespace
+
+PalImpl PalImpl::create(
+    pal_emit_log_message_method emit_log_message,
+    const char* source_filename) {
+  return PalImpl::create(
+      nullptr, // init
+      nullptr, // abort
+      nullptr, // current_ticks
+      nullptr, // ticks_to_ns_multiplier
+      emit_log_message,
+      nullptr, // allocate
+      nullptr, // free
+      source_filename);
+}
+
+PalImpl PalImpl::create(
+    pal_init_method init,
+    pal_abort_method abort,
+    pal_current_ticks_method current_ticks,
+    pal_ticks_to_ns_multiplier_method ticks_to_ns_multiplier,
+    pal_emit_log_message_method emit_log_message,
+    pal_allocate_method allocate,
+    pal_free_method free,
+    const char* source_filename) {
+  return PalImpl{
+      init,
+      abort,
+      current_ticks,
+      ticks_to_ns_multiplier,
+      emit_log_message,
+      allocate,
+      free,
+      source_filename};
+}
+
+/**
+ * Override the PAL functions with user implementations. Any null entries in the
+ * table are unchanged and will keep the default implementation.
+ */
+bool register_pal(PalImpl impl) {
+  if (is_pal_overridden) {
+    ET_LOG(
+        Error,
+        "register_pal() called multiple times. Subsequent calls will override the previous implementation. Previous implementation was registered from %s.",
+        impl.source_filename != nullptr ? impl.source_filename : "unknown");
+  }
+  is_pal_overridden = true;
+
+  if (impl.abort != nullptr) {
+    pal_impl.abort = impl.abort;
+  }
+
+  if (impl.current_ticks != nullptr) {
+    pal_impl.current_ticks = impl.current_ticks;
+  }
+
+  if (impl.ticks_to_ns_multiplier != nullptr) {
+    pal_impl.ticks_to_ns_multiplier = impl.ticks_to_ns_multiplier;
+  }
+
+  if (impl.emit_log_message != nullptr) {
+    pal_impl.emit_log_message = impl.emit_log_message;
+  }
+
+  if (impl.allocate != nullptr) {
+    pal_impl.allocate = impl.allocate;
+  }
+
+  if (impl.free != nullptr) {
+    pal_impl.free = impl.free;
+  }
+
+  if (impl.init != nullptr) {
+    pal_impl.init = impl.init;
+    if (pal_impl.init != nullptr) {
+      pal_impl.init();
+    }
+  }
+
+  return true;
+}
+
+const PalImpl* get_pal_impl() {
+  return &pal_impl;
+}
+
+void pal_init() {
+  pal_impl.init();
+}
+
+ET_NORETURN void pal_abort() {
+  pal_impl.abort();
+  // This should be unreachable, but in case the PAL implementation doesn't
+  // abort, force it here.
+  std::abort();
+}
+
+et_timestamp_t pal_current_ticks() {
+  return pal_impl.current_ticks();
+}
+
+et_tick_ratio_t pal_ticks_to_ns_multiplier() {
+  return pal_impl.ticks_to_ns_multiplier();
+}
+
+void pal_emit_log_message(
+    et_timestamp_t timestamp,
+    et_pal_log_level_t level,
+    const char* filename,
+    const char* function,
+    size_t line,
+    const char* message,
+    size_t length) {
+  pal_impl.emit_log_message(
+      timestamp, level, filename, function, line, message, length);
+}
+
+void* pal_allocate(size_t size) {
+  return pal_impl.allocate(size);
+}
+
+void pal_free(void* ptr) {
+  pal_impl.free(ptr);
+}
+
+} // namespace executorch::runtime
--- a/runtime/platform/platform.h
+++ b/runtime/platform/platform.h
@@ -11,6 +11,10 @@
  * Platform abstraction layer to allow individual platform libraries to override
  * symbols in ExecuTorch. PAL functions are defined as C functions so a platform
  * library implementer can use C in lieu of C++.
+ *
+ * The et_pal_ methods should not be called directly. Use the corresponding
+ * methods in the executorch::runtime namespace instead to appropriately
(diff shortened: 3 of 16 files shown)


runtime/platform/platform.h
```diff
@@ -126,12 +142,134 @@ void et_pal_emit_log_message(
...
 
 } // extern "C"
+
+namespace executorch {
+namespace runtime {
+
+/**
+ * Table of pointers to platform abstraction layer functions.
+ */
+struct pal_table {
```
GregoryComer, line 163: Any preferences on naming this? I'm not 100% happy with pal_table.

runtime/platform/test/executor_pal_override_test.cpp
```diff
@@ -16,79 +17,6 @@
 using namespace ::testing;
 using executorch::runtime::LogLevel;
 
-class PalSpy : public PlatformIntercept {
```
GregoryComer, line 19: This was refactored into a separate header to share with the new tests.

runtime/platform/platform.h
```diff
@@ -126,12 +142,134 @@ void et_pal_emit_log_message(
...
+  const char* source_filename;
+};
+
+/**
+ * Override the PAL functions with user implementations. Any null entries in the
+ * table are unchanged and will keep the default implementation.
+ *
+ * Returns true if the registration was successful, false otherwise.
+ */
+bool register_pal(pal_table impl);
```
GregoryComer, line 199: This API is returning bool primarily because using Error would introduce a circular header dependency and there's no current way for this call to fail. I'm open to switching to error, but would require a little bit of a header refactor.

GregoryComer:
I see the size regression. I may need to add a build time flag to gate this out. @digantdesai is it acceptable to enable this feature by default?

runtime/platform/test/executor_pal_runtime_override_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+        nullptr, // abort
+        pal_current_ticks,
+        pal_ticks_to_ns_multiplier,
+        pal_emit_log_message,
+        pal_allocate,
+        pal_free,
+        __FILE__));
+  }
+
+  PalSpy _spy;
```
digantdesai, line 88: what's spy?

runtime/platform/test/executor_pal_runtime_override_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+    auto success = executorch::runtime::register_pal(_original_pal_impl);
+    if (!success) {
+      throw std::runtime_error("Failed to restore PAL implementation.");
+    }
+  }
+
+  void RegisterSpy() {
+    active_spy = &_spy;
+
+    executorch::runtime::register_pal(executorch::runtime::pal_table(
```
digantdesai, line 77: can we do a test with c++ static init, which is how I expect people to use it?

Also is there any platform which overrides the default methods?

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return &pal_impl;
+}
+
+void pal_init() {
+  pal_impl.init();
+}
+
+ET_NORETURN void pal_abort() {
+  pal_impl.abort();
+  // This should be unreachable, but in case the PAL implementation doesn't
```
digantdesai, line 117: was it there before?

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+  // abort, force it here.
+  std::abort();
+}
+
+et_timestamp_t pal_current_ticks() {
+  return pal_impl.current_ticks();
```
digantdesai, line 123: Can't this be nullptr?

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  if (impl.init != nullptr) {
+    pal_impl.init = impl.init;
+    if (pal_impl.init != nullptr) {
+      pal_impl.init();
+    }
+  }
+
+  return true;
```
digantdesai, line 104: assert the whole table is non-nullptr?

runtime/platform/test/executor_pal_runtime_override_test.cpp
digantdesai, line 77: Thinking out loud, the PAL resolution earlier would happen at compile/link time, worst case at load time through weak symbol resolution. While with C++ static init (before we enter main) or worst manual call during init would happen way after the process starts running. So if there are things we want to intercept, we can't use stuff from PAL like timestamp or print calls, this is very important for something like Linux kernel which jumps through some hoops to get printk enabled as soon as possible.

I wonder if this is important to us, not now but also in the future looking at bare metal deployments, not so much for rich OSes.

runtime/platform/test/executor_pal_runtime_override_test.cpp
larryliu0820: Seems like a test implementation

runtime/platform/test/executor_pal_runtime_override_test.cpp
larryliu0820: Yeah we should have a test for static init.

runtime/platform/platform.h
psiddh: Do you think this is better ? PlatformAbstractionLayerTbl

runtime/platform/test/executor_pal_runtime_override_test.cpp
GregoryComer: Yeah, this is an existing test class that's used in the other PAL tests. I reused it here for simplicity.

runtime/platform/test/executor_pal_runtime_override_test.cpp
GregoryComer: Makes sense. Ultimately, the existing weak mechanism still works and will be set up at link time. Tbh what I'd prefer is if ET had an explicit init call where we could pass in library config, but it's probably a bit late for that.

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+      emit_log_message(emit_log_message),
+      allocate(allocate),
+      free(free),
+      source_filename(source_filename) {}
+
+/**
+ * Override the PAL functions with user implementations. Any null entries in the
+ * table are unchanged and will keep the default implementation.
+ */
+bool register_pal(pal_table impl) {
```
psiddh, line 64: It looks like the expectation here is, that calling entity is responsible for thread synchronization. If so should we leave an explicit comment ?

runtime/platform/test/executor_pal_runtime_override_test.cpp
GregoryComer: Thanks. I'll add one.

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+  pal_impl.emit_log_message(
+      timestamp, level, filename, function, line, message, length);
+}
+
+void* pal_allocate(size_t size) {
+  return pal_impl.allocate(size);
+}
+
+void pal_free(void* ptr) {
+  pal_impl.free(ptr);
```
psiddh, line 147: nits: Add assert ?  'assert(pal_impl != nullptr);'

runtime/platform/test/executor_pal_runtime_override_test.cpp
GregoryComer: > Also is there any platform which overrides the default methods?

Are you asking about in production? I believe we have Meta embedded consumers who do. I don't know of any mobile users, though this is in part because mobile users are often writing Java/Obj/Swift and don't have any way to do this without switching to C++ (I intend to solve this, at least for log routing, as a follow-up).

runtime/platform/test/executor_pal_runtime_override_test.cpp
digantdesai: No I am wondering about the name lol

runtime/platform/test/executor_pal_static_runtime_override_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+void* pal_allocate(size_t size) {
+  return spy.allocate(size);
+}
+
+void pal_free(void* ptr) {
+  spy.free(ptr);
+}
+
+// Statically register PAL impleementation.
+bool registration_result = executorch::runtime::register_pal(
```
digantdesai, line 50: how do we know this doesn't get removed by the linker in some different build config, without a `static` qualifier?

runtime/platform/test/executor_pal_runtime_override_test.cpp
GregoryComer: I assume it's called Spy as it's implementing a spy pattern for testing - where it observes (counts, in this case) method calls made on it. Not sure who originally added it.

runtime/platform/platform.cpp
GregoryComer: In general, we've taken the approach that the burden of thread synchronization is on the caller for all core runtime APIs (https://docs.pytorch.org/executorch/stable/runtime-overview.html#threading-considerations). This is largely due to not all platforms supporting threading, though there has been some discussion of incorporating locking in the PAL, which can be a no-op on bare-metal platforms. But as of now, we have a general policy of no locking in the core runtime, so I'm hesitant to call it out explicitly as we'd end up documenting it on every core runtime API.

There is a race possible with multiple global constructors, which I don't love, but we do at least log a warning when the override happens multiple times.

runtime/platform/platform.cpp
GregoryComer: This should be statically impossible, since we initialize to the default implementations and only override with non-null entries, but I'm not opposed to adding the assert if desired.

runtime/platform/platform.cpp
```diff
@@ -0,0 +1,151 @@
...
+    pal_impl.free = impl.free;
+  }
+
+  if (impl.init != nullptr) {
+    pal_impl.init = impl.init;
+    if (pal_impl.init != nullptr) {
+      pal_impl.init();
+    }
+  }
+
```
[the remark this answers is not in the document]
GregoryComer: To clarify - you're suggesting we print the address of the function that is being overridden? I'm happy to add it, just not sure how useful that is in general vs adding to log spam in debug mode.

runtime/platform/platform.cpp
GregoryComer: Currently, pal_impl is actually not a pointer.

runtime/platform/platform.h
```diff
@@ -53,19 +57,22 @@ typedef struct {
  * to initialize any global state. Typically overridden by PAL implementer.
  */
 void et_pal_init(void) ET_INTERNAL_PLATFORM_WEAKNESS;
+typedef void (*et_pal_init_t)(void);
```
swolchok, line 60: use using

runtime/platform/platform.h
```diff
@@ -126,12 +142,136 @@ void et_pal_emit_log_message(
...
  *
  * @param[in] ptr Pointer to memory to free. May be nullptr.
  */
 void et_pal_free(void* ptr) ET_INTERNAL_PLATFORM_WEAKNESS;
+typedef void (*et_pal_free_t)(void* ptr);
 
 } // extern "C"
+
+namespace executorch {
+namespace runtime {
```
swolchok, line 158: use nested namespaces

.github/workflows/pull.yml
```diff
@@ -371,7 +371,7 @@ jobs:
         size=${arr[4]}
         # threshold=48120 on devserver with gcc11.4
         # todo(lfq): update once binary size is below 50kb.
-        threshold="51408"
+        threshold="55584"
```
GregoryComer, line 374: Note that the "true" size increase is around ~80 bytes, due to some instability in the size test on older toolchains. See https://github.com/pytorch/executorch/pull/11217/files. This is confirmed with manual builds on newer toolchains and Meta app build size checks.

.github/workflows/pull.yml
```diff
@@ -406,7 +406,7 @@ jobs:
         output=$(ls -la cmake-out/test/size_test)
         arr=($output)
         size=${arr[4]}
-        threshold="47560"
+        threshold="51720"
```
GregoryComer: Similar to above, note that the "true" size increase is around ~64 bytes. See https://github.com/pytorch/executorch/pull/11217/files.

larryliu0820 (approved):

PR #10697, opened 2025-05-05
Handle avg_pool2d with padding == 0 as no padding

3l1:
Differential Revision: D74117402


digantdesai (approved):
stamping to unblock, left a comment on the diff

limintang (approved):

PR #10698, opened 2025-05-05
[CMake] llm_runner use executorch_core

kirklandsign:
### Summary
We only need executorch_core. We don't need prim ops for runner library. runner.exe can add prim_ops separately

Change:
```
set(_extension_llm_runner__srcs
-    extension/tensor/tensor_ptr.cpp
-    extension/tensor/tensor_ptr_maker.cpp
)```

### Test plan
CI

cc @larryliu0820 @jathu

### CODE REVIEW

File: tools/cmake/cmake_deps.toml [hunk 1]
```diff
@@ -197,8 +196,7 @@ filters = [
   ".cpp$",
 ]
 deps = [
-  "extension_flat_tensor_schema",
```
lucylq, line 200: was this unecessary?

kirklandsign: Let me add it back. Probably shouldn't remove.

lucylq (approved):

shoumikhin (approved):

PR #10701, opened 2025-05-05
Add input size validation to Module.execute

### LINKED ISSUE #10510: Module.execute does not check input count
### 🐛 Describe the bug

The execute function in module.cpp will cause silent memory corruption if too many input EValues are passed.

https://github.com/pytorch/executorch/blob/976fe484c811277252756a39a9b6c76fd8c6e3cb/extension/module/module.cpp#L233-L244

If input_values.size() > inputs.size(), the indexing operation on line 242 will write past the end of the array. I encountered this when I updated a model definition but hadn't updated the C++ runner code yet and it caused a memory-related crash when destructing the module.

We should check for input_values.size() <= inputs.size() and log an error + return an error code (Error::InvalidArgument) if the check fails.

As an aside, passing fewer inputs in may or may not be a desirable feature, given that some may be memory planned or unchanged between calls. We can ignore this case for the sake of this bug as it's not a correctness issue, but it might be good to error out for clarity if too few inputs are passed.

### Versions

Repro-able on main, though it's likely been like this for a long time (or forever).

keyprocedure:
### Summary
Adds input size validation to `Module.execute` to prevent possible silent memory corruption when too many EValue inputs are passed.

Fixes #10510 

### Test plan
- Added unit test `TestExecuteWithTooManyInputs`
- Verified by successfully running all `module_test.cpp` tests, except `TestPTD` (did not have access to `ModuleLinear.ptd`)
- To run locally:  
  - Bypass `is_fbcode` guard in `targets.bzl` and redirect test file paths to use a locally exported `ModuleAdd.pte` file
  - Build and run tests via:
  
  ```
  buck2 build //extension/module/test:test
  buck2 run //extension/module/test:test


shoumikhin (approved):

keyprocedure:
Hi @shoumikhin, the CI failures are unrelated to this PR; they're due to`test_llava_export` and `test_position_embeddings.py`. Could you please merge when you get a chance?

PR #10712, opened 2025-05-06
Arm backend: Add non-interactive in git hook

perheld:
The pre-push script is a nice to have script as a git hook for your work with the arm backend.

In fact, it is so nice that we would like it to be part of CI jobs but to make that happen it needs to be non-interactive.

This patch adds support for launching the script and never ending up in a scenario where the user has to interact with it.

Change-Id: I345a3a71f1b81446228150baf4db28bf4a47b0f1


zingo:
llama error unrelated

zingo (approved):

PR #10713, opened 2025-05-06
Arm backend: Create op utility function for num input verification

Sebastian-Larsson:
Create function for validating that the number of inputs is correct in each node visitor.


Sebastian-Larsson:
Failing CI seems unrelated

zingo (approved):

zingo:
MacOS fails unrelated

PR #10714, opened 2025-05-06
Arm backend: Allocate the scratch buffer runtime rather than in the pte

gggekov:
This change lowers the size of the pte and allows you to allocate the scratch buffer in an array, usually in the SRAM, for more efficient memory usage on a MCU

### DIFF
--- a/backends/arm/arm_vela.py
+++ b/backends/arm/arm_vela.py
@@ -73,8 +73,8 @@ def vela_compile(tosa_flatbuffer: bytes, args: List[str], verbose: bool = False)
             np_path = os.path.join(tmpdir, "output", "out_vela.npz")
         else:
             np_path = os.path.join(tmpdir, "output", "out_sg0_vela.npz")
-        blocks = b""
 
+        blocks = b""
         with np.load(np_path, allow_pickle=False) as data:
             # Construct our modified output_blocks with data in a form easily
             # digested on the device side
@@ -92,7 +92,7 @@ def vela_compile(tosa_flatbuffer: bytes, args: List[str], verbose: bool = False)
             if not isinstance(data["scratch_shape"][0], np.int64):
                 raise RuntimeError("Expected scratch to be int64")
             block_length = int(data["scratch_shape"][0])
-            bin_blocks["scratch_data"] = b"\x00" * block_length
+            bin_blocks["scratch_size"] = struct.pack("<I", block_length)
 
             # Capture inputs and outputs
             bin_blocks["inputs"] = vela_bin_pack_io("input", data)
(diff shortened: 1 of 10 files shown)


backends/arm/arm_vela.py
```diff
@@ -92,7 +92,7 @@ def vela_compile(tosa_flatbuffer: bytes, args: List[str], verbose: bool = False)
             if not isinstance(data["scratch_shape"][0], np.int64):
                 raise RuntimeError("Expected scratch to be int64")
             block_length = int(data["scratch_shape"][0])
-            bin_blocks["scratch_data"] = b"\x00" * block_length
+            bin_blocks["scratch_size"] = struct.pack("<I", block_length)
```
digantdesai, line 95: Nice

digantdesai (approved):
Make sure the CI is green, thanks.

gggekov:
Yes, I am adding support for Dedicated_Sram for U85 & changing the default mem mode we test on U85. This is the proper fix for the fail we see for inception_v4. With this fix, we will place the NN & scratch buffer in the DDR and use the SRAM as a cache. The reason for the failure is that the scratch_buffer for inception_v4 is around 2.6-2.7MB, we allocate the scratch buffer in the SRAM, but on the CS-300 we only 2MB of SRAM. Will update the pr soon.

zingo:
failed test are unrelated

kirklandsign:
Hi @gggekov @zingo this is breaking our internal test. 

undefined reference to `executorch::backends::arm::ethosu_fast_scratch_size'

Seems it's defined in arm_executor_runner only. Any suggestions what we should do on our side? Otherwise OK to revert this?

kirklandsign:
Could you please review https://github.com/pytorch/executorch/pull/10958 for a temp fix

gggekov:
Hi @kirklandsign, 
Thanks for the message, added a comment in https://github.com/pytorch/executorch/pull/10958

backends/arm/runtime/EthosUBackend.cpp
```diff
@@ -181,23 +183,33 @@ class EthosUBackend final : public ::executorch::runtime::BackendInterface {
...
     EXECUTORCH_PROF_END(event_tracer, event_tracer_local_scope);
 
+    MemoryAllocator* temp_allocator = context.get_temp_allocator();
+    // Use a temporary allocator for the intermediate tensors of the
+    // computation. The allocator is released in runtime/executor/method.cpp at
+    // the end of the execution of the Ethos-U custom delegate
+    char* ethosu_scratch =
+        static_cast<char*>(temp_allocator->allocate(handles.scratch_data_size));
+    extern size_t ethosu_fast_scratch_size;
+    extern unsigned char* ethosu_fast_scratch;
```
digantdesai, line 193: Ok this is not the cleanest. Let me think of a better way to do this.

Issue #10716, opened 2025-05-06
[Build Presets] Create foundation and default configurations


jathu:
Create the foundation for presets, and setup the default configurations.

jathu:
This is done.

PR #10731, opened 2025-05-06
Create a helper to define overridable configs

jathu:
### Summary
Context: https://github.com/pytorch/executorch/discussions/10661. As part https://github.com/pytorch/executorch/issues/10716, this creates a helper to define ET configs.

#### Why define_overridable_config?

This is a simple wrapper to configure an option via the user or fallback to a default value. Instead of having people do this manually, this is convenient and consistent. More specifically:

- If the option is set via the CLI or `set`, then save that value in the cache — with our own description
- If the option is unset, then fallback to the default value

The user options will be loaded *before* loading the `default.cmake` file. This means, when we load the default file, we can use the user set options or fallback on defaults. Then finally we can do a final sweet of option validations. After this point, all options will be frozen and immutable.

Moreover, having a helper function will allow us to enforce standards on options. For example, naming standards done in this PR.

#### Migration plan

For now, I've only migrated a simple option `EXECUTORCH_ENABLE_LOGGING`. We'll incrementally migrate remaining options in future PRs. 

### Test plan

For changes to `EXECUTORCH_ENABLE_LOGGING`, I changed the release modes and checked the finaly `CMakeCache.txt` file. Rely on CI for full tests.

For python tests:

```
$ python3 -m unittest discover -s tools/cmake -p "*.py"
```

I will include this test in CI in the next diff.

cc @larryliu0820

### DIFF
--- a/tools/cmake/common/__init__.py
+++ b/tools/cmake/common/__init__.py
@@ -1,0 +1,144 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import os
+import shutil
+import subprocess
+import tempfile
+import unittest
+from functools import cache
+from typing import Any, Dict, List, Optional
+
+# Files to copy from this directory into the temporary workspaces.
+TESTABLE_CMAKE_FILES = [
+    "preset.cmake",
+]
+
+
+# If KEEP_WORKSPACE is set, then keep the workspace instead of deleting it. Useful
+# when debugging tests.
+@cache
+def _keep_workspace() -> bool:
+    keep_workspace_env = os.environ.get("KEEP_WORKSPACE")
+    if keep_workspace_env is None:
+        return False
+    return keep_workspace_env.lower() not in ("false", "0", "no", "n")
+
+
+# Create a file tree in the current working directory (cwd). The structure of the
+# tree maps to the structure of the file tree. The key of the tree is the name
+# of the folder or file. If the value is dict, it creates a folder. If the value
+# is a string, it creates a file.
+#
+# Example:
+#
+#     {
+#       "README.md": "this is a read me file",
+#       "build": {
+#         "cmake": {
+#           "utils.cmake": "this is a cmake file",
+#         }
+#       }
+#     }
+# Results in:
+#
+#     ├── README.md
+#     └── build
+#         └── cmake
+#             └── utils.cmake
+#
+def _create_file_tree(tree: Dict[Any, Any], cwd: str) -> None:
+    for name, value in tree.items():
+        if isinstance(value, str):
+            file_path = os.path.join(cwd, name)
+            assert not os.path.exists(file_path), f"file already exists: {file_path}"
+            os.makedirs(cwd, exist_ok=True)
+            with open(file_path, "w") as new_file:
+                new_file.write(value)
+        elif isinstance(value, dict):
+            new_cwd = os.path.join(cwd, name)
+            os.makedirs(new_cwd, exist_ok=True)
+            _create_file_tree(tree=value, cwd=new_cwd)
+        else:
+            raise AssertionError("invalid tree value", value)
+
+
+# Get the key/value pair listed in a CMakeCache.txt file.
+@cache
+def _list_cmake_cache(cache_path: str) -> Dict[str, str]:
+    result = {}
+    with open(cache_path, "r") as cache_file:
+        for line in cache_file:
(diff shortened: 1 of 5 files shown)


tools/cmake/common/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+                if ":" in key:
+                    key, value_type = key.split(":")
+                result[key.strip()] = _CacheValue(
+                    value_type=value_type,
+                    value=value.strip(),
+                )
+    return result
+
+
+class CMakeTestCase(unittest.TestCase):
```
larryliu0820, line 95: Wow thanks for writing the tests! I didn't know CMake is testable

tools/cmake/common/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+            assert os.path.exists(
+                source_path
+            ), f"{testable_cmake_file} does not exist in {source_path}"
+            destination_path = os.path.join(self.workspace, testable_cmake_file)
+            os.makedirs(os.path.dirname(destination_path), exist_ok=True)
+            shutil.copy(source_path, destination_path)
+
+        _create_file_tree(tree=tree, cwd=self.workspace)
+
+    def assert_file_content(self, relativePath: str, expectedContent: str) -> None:
```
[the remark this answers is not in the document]
jathu: Since this is a _helper_ to test if a file contains a certain content, I think `assert` might be appropriate no?

tools/cmake/common/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+        path = os.path.join(self.workspace, relativePath)
+        self.assertTrue(os.path.exists(path), f"expected path does not exist: {path}")
+
+        with open(path, "r") as path_file:
+            self.assertEqual(path_file.read(), expectedContent)
+
+    def run_cmake(
```
larryliu0820, line 129:
```suggestion
    def test_run_cmake_gives_correct_result(
```

tools/cmake/common/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+            self.assertNotEqual(result.returncode, 0)
+            actual_error = result.stderr.decode("utf-8")
+            self.assertTrue(
+                error_contains in actual_error, f"Actual error: {actual_error}"
+            )
+        else:
+            self.assertEqual(result.returncode, 0)
+            self.assertTrue(os.path.exists(os.path.join(self.workspace, "cmake-out")))
+
+    def assert_cmake_cache(
```
larryliu0820, line 154:
```suggestion
    def test_cmake_cache_works_correctly(
```

tools/cmake/common/preset.cmake
```diff
@@ -0,0 +1,29 @@
...
+  enforce_executorch_config_name(${NAME})
+
+  if(NOT "${VALUE_TYPE}" STREQUAL "STRING" AND NOT "${VALUE_TYPE}" STREQUAL "BOOL")
+    message(FATAL_ERROR "Invalid config (${NAME}) value type '${VALUE_TYPE}', must be either STRING or BOOL")
+  endif()
+
+  if(DEFINED ${NAME})
+    set(${NAME} ${${NAME}} CACHE ${VALUE_TYPE} ${DESCRIPTION} FORCE)
+  else()
+    set(${NAME} ${DEFAULT_VALUE} CACHE ${VALUE_TYPE} ${DESCRIPTION})
```
larryliu0820, line 27: Should this be 
```
option(EXECUTORCH_ENABLE_LOGGING "Build with ET_LOG_ENABLED"
       ${_default_release_disabled_options}
)
```

larryliu0820 (approved):

tools/cmake/common/__init__.py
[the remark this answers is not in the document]
larryliu0820: ah ok make sense

tools/cmake/common/preset.cmake
larryliu0820: talked about this offline, it seems cmake is limiting options to be boolean so set() here is fine.

tools/cmake/common/preset_test.py
```diff
@@ -0,0 +1,225 @@
...
+                "CMakeLists.txt": _cmake_lists_txt,
+                "example": {
+                    "CMakeLists.txt": _example_cmake_lists_txt,
+                },
+            }
+        )
+        self.run_cmake()
+        self.assert_cmake_cache("EXECUTORCH_TEST_MESSAGE", "set value", "STRING")
+
+    def testdefine_overridable_config_set_override_after(self):
```
larryliu0820, line 158: typo

PR #10732, opened 2025-05-06
openvino_backend doesn't need to be static only

kirklandsign:
We can allow BUILD_SHARED_LIBS=ON and build shared library.


kirklandsign:
Hi @ynimmaga could use review from openvino team

PR #10735, opened 2025-05-06
[CMake] Avoid extension_module have dupe flat_tensor cpp

kirklandsign:
Currently, _extension_module__srcs has srcs from flat_tensor. This is bad.

```
set(_extension_module__srcs
    extension/flat_tensor/flat_tensor_data_map.cpp
    extension/flat_tensor/serialize/flat_tensor_header.cpp
    extension/module/module.cpp
)
```

We need to update deps. We need to export extension_flat_tensor to install directory.


lucylq (approved):
thanks!

PR #10740, opened 2025-05-07
Fix BUCK

kirklandsign:
Differential Revision: D74290346


guangy10 (approved):

guangy10:
Should you merge from OSS or codev in this case? If from codev, would cause an out-of-sync issue?

PR #10750, opened 2025-05-07
Update aot_arm_compiler.py

Juanfi8:
Found an error when trying to use the aot_arm_compiler.py with a custom model defined in an external python file:

```python
import torch

class ModelUnderTest(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return x +x
    
ModelInputs = (torch.ones(5),)
```
Error message: TypeError: Module.eval() missing 1 required positional argument: 'self'

This patch fixes the problem.

### DIFF
--- a/examples/arm/aot_arm_compiler.py
+++ b/examples/arm/aot_arm_compiler.py
@@ -113,7 +113,7 @@ def get_model_and_inputs_from_name(
         spec = importlib.util.spec_from_file_location("tmp_model", model_name)
         module = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(module)
-        model = module.ModelUnderTest
+        model = module.ModelUnderTest()
         if example_inputs is None:
             example_inputs = module.ModelInputs
     # Case 4: Model is in an saved model file torch.save(model)


zingo:
Hi thanks for your PR, this highlight a "bad" think in our current setup and help texts,  as we just used this method to test model internally I see we lack some sort of documentation/description on how file should look like. Down the line I was hoping to move out the small models in the aot compiler to small model file snipets like yours as a better example then how it works today when they are baked into it.

The idea was to set/use ModelUnderTest and ModelInputs as pre defined varable in the pyton file 
e,g, you file could be changed to this:

```
import torch

class mymodel(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return x +x

ModelUnderTest=mymodel()
ModelInputs = (torch.ones(5),)
```

If that is OK for you then the PR should not be needed :) But if you would like to submit a model like this as a new example maybe with a short text in the readm on how to run it we would appreciate it, if not we will hopefully (soon:ish) do it anyway down the line.  

We are also open to other maybe better ways to handle this if you have any input that would be really nice, I feel that the ModelUnderTest and ModelInputs is a bit of a hack, kind of OK and solves the problem but maybe we could do it better :) It would be really cool if for some huggingface or torchaudion/video models you could just sue them and don't need to patch the files, but lets see what we manage to do :)

PR #10755, opened 2025-05-07
: constant fold None

ThomasJannaud:
Summary: Constant folding should fold 'None' and consider it a constant

Differential Revision: D74280168


JacobSzwejbka:
Sorry for the late review. Can you add some context on this? Does const_prop not trigger if one of the args is None? Can you add a unit test that wouldve failed before this change?

ThomasJannaud:
closed in favor of #10762

PR #10756, opened 2025-05-07
move pattern

kirklandsign:
Differential Revision: D74344187


swolchok (changes requested):
isn't it bc-breaking to move all this stuff?

PR #10760, opened 2025-05-07
Make constant_folding's _DEFAULT_SKIP_TARGETS public

ThomasJannaud:
Summary: This follows D73513516 and allows customers of constant_prop_pass to use previous defaults

Differential Revision: D74349918


JacobSzwejbka (approved):
looks a bit weird since both sets are the same here but I guess its fine

PR #10762, opened 2025-05-07
: constant fold None

ThomasJannaud:
Summary:
Constant folding should fold 'None' and consider it a constant

This goes with D74350331 and D74349918 but keeping things separate

Differential Revision: D74350331

### DIFF
--- a/exir/passes/constant_prop_pass.py
+++ b/exir/passes/constant_prop_pass.py
@@ -61,6 +61,8 @@ def is_const(
         )
     elif isinstance(arg, _PRIMITIVE_TYPES):
         return True
+    elif arg is None:
+        return True
     elif not isinstance(arg, torch.fx.Node):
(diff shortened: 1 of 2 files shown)


exir/passes/constant_prop_pass.py
```diff
@@ -61,6 +61,8 @@ def is_const(
         )
     elif isinstance(arg, _PRIMITIVE_TYPES):
         return True
+    elif arg is None:
```
JacobSzwejbka, line 64: was the problem before that const prop would skip nodes with None in the args?

metascroy: yes, I think the issue is None is not a constant, and so if an op has args that include constants and None, it doesn't get constant folded.

metascroy (approved):

PR #10781, opened 2025-05-09
Remove FLATC_EXECUTABLE and the ability to bring your own flatc

jathu:
### Summary

* Remove the ability to set flatc using FLATC_EXECUTABLE and always build it from source
  * Also remove, the now redundant, `EXECUTORCH_BUILD_FLATC`
* Move the flatc configuration into `third-party/CMakeLists.txt` to keep things organized
* Allow `FLATBUFFERS_MAX_ALIGNMENT` to be configurable

### Test plan

CI

cc @larryliu0820

### DIFF
--- a/third-party/CMakeLists.txt
+++ b/third-party/CMakeLists.txt
@@ -1,0 +1,49 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+# MARK: - flatbuffers
+
+# We use ExternalProject to build flatc from source to force it target the host.
+# Otherwise, flatc will target the project's toolchain (i.e. iOS, or Android).
+ExternalProject_Add(
+  flatbuffers_external_project
+  PREFIX ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers_external_project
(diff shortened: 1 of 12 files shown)


third-party/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+add_dependencies(flatc flatbuffers_external_project)
+if(WIN32)
+  # flatbuffers does not use CMAKE_BUILD_TYPE. Internally, the build forces Release
+  # config, but from CMake's perspective the build type is always Debug.
+  set_target_properties(flatc PROPERTIES IMPORTED_LOCATION ${BINARY_DIR}/$<CONFIG>/flatc.exe)
+else()
+  set_target_properties(flatc PROPERTIES IMPORTED_LOCATION ${BINARY_DIR}/flatc)
+endif()
+
+# TODO: re-enable once flatbuffers is added as a subdirectory.
```
larryliu0820, line 44: Why do we need this

larryliu0820 (approved):

third-party/CMakeLists.txt
jathu: I want to stop directly reaching into flatbuffers, and instead use the `flatbuffer` library target:

https://github.com/pytorch/executorch/blob/bf505276b6ea0e020b582edf8b6b9342e9d15d8f/runtime/executor/test/CMakeLists.txt#L158-L158

I'll do it in some future diff

kirklandsign (approved):

PR #10548, opened 2025-04-29
Arm backend: Fix sigmoid int16 and int32 flakyness

oscarandersson8218:
Step vela pin and remove all related pytest.mark_flaky in sigmoid int16 and int32 tests.

cc @digantdesai @freddan80 @per @zingo

PR #10782, opened 2025-05-09
Cortex-M: Use q/dq ops in Arm Ethos Runner

digantdesai:
### Summary
* Adds MVE version of [de]quantize_per_tensor
* Adds basic support to link Cortex-M backend op (replacing ET quantized ops) with Arm Ethos runner
* Use them in FVP tests for M55/M85
* No selective build yet, the plan is to extend this such that we can have, merged_yaml with ops used my model(s) from Cortex-M-backend and Portable Lib.

### Test plan
```./examples/arm/run.sh --target=ethos-u85-256 --model_name=lstm --bundleio```

### DIFF
--- a/backends/cortex_m/ops/op_dequantize_per_tensor.cpp
+++ b/backends/cortex_m/ops/op_dequantize_per_tensor.cpp
@@ -73,16 +73,23 @@ void check_dequantize_args(
 /**
  * Scalar implementation of quantization for a single value.
  */
-template <typename K, typename T>
-T dequantize_val(
+template <typename Q, typename F>
+F dequantize_val(
     float scale,
     int32_t zero_point,
-    K value,
+    Q qvalue,
     int64_t quant_min,
     int64_t quant_max) {
   (void)quant_min;
   (void)quant_max;
-  return static_cast<T>((static_cast<int32_t>(value) - zero_point) * scale);
+  F value = static_cast<F>((static_cast<int32_t>(qvalue) - zero_point) * scale);
+  printf(
+      "Scalar DQ: float value: %f, quantized value: %d, scale: %f, zp: %d\n",
+      value,
+      qvalue,
+      scale,
+      zero_point);
+  return value;
 }
 
 } // namespace
@@ -98,6 +105,13 @@ Tensor& dequantize_per_tensor_out(
     Tensor& out) {
   // Ignore context for now
   (void)context;
+
+  printf(
+      "DQ: scale: %f, zp: %d, qmin: %d, qmax: %d\n",
+      scale,
+      zero_point,
+      quant_min,
+      quant_max);
 
   // Resize output tensor to match input dimensions
   torch::executor::Error err = resize_tensor(out, input.sizes());
@@ -118,11 +132,11 @@ Tensor& dequantize_per_tensor_out(
   float* out_data = out.mutable_data_ptr<float>();
   const size_t numel = input.numel();
 
-#if defined(HAS_HELIUM_SIMD)
+#if defined(HAS_HELIUM_SIMD) && 0
 // Helium MVE implementation for float32 to int8 quantization
 #Error "Implement MVE version!"
 #else
-  // Scalar implementation for float32 to int8 quantization
+  // Scalar implementation for int8 to float32 quantization
   for (size_t i = 0; i < numel; i++) {
     out_data[i] =
         dequantize_val<int8_t, float>(scale, zp, input_data[i], qmin, qmax);
--- a/backends/cortex_m/ops/op_dequantize_per_tensor.cpp
+++ b/backends/cortex_m/ops/op_dequantize_per_tensor.cpp
@@ -29,6 +29,7 @@ namespace {
  */
 void check_dequantize_args(
     const Tensor& input,
+    int64_t zero_point,
     int64_t quant_min,
     int64_t quant_max,
     ScalarType dtype,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: examples/arm/aot_arm_compiler.py [hunk 1]
```diff
@@ -759,7 +840,9 @@ def to_edge_no_delegate(exported_program, args, model: torch.nn.Module, example_
             output_name = os.path.join(args.output, output_name)
 
     if args.bundleio:
-        save_bpte_program(exec_prog, original_model, output_name)
+        # Realize the quantization impact on numerics when generating reference output
```
AdrianLundell, line 843: The problematic thing with comparing to the quantized model is that you don't catch how changes to the quantization affect accuracy compared to the original. Is this tested somewhere else?

backends/cortex_m/ops/op_dequantize_per_tensor.cpp
```diff
@@ -106,29 +111,71 @@ Tensor& dequantize_per_tensor_out(
...
 #if defined(HAS_HELIUM_SIMD)
-// Helium MVE implementation for float32 to int8 quantization
-#Error "Implement MVE version!"
-#else
-  // Scalar implementation for float32 to int8 quantization
-  for (size_t i = 0; i < numel; i++) {
-    out_data[i] =
-        dequantize_val<int8_t, float>(scale, zp, input_data[i], qmin, qmax);
+  // Helium MVE implementation for int8 to float quantization
+  static uint8x16_t voffset{
```
AdrianLundell, line 126: These implementations are actually very similar to what we have done on our side except that we used predication for the tail handling, is this something you have considered?

backends/cortex_m/ops/op_quantize_per_tensor.cpp
```diff
@@ -75,18 +75,18 @@ void check_quantize_args(
 /**
  * Scalar implementation of quantization for a single value.
  */
-template <typename T, typename K>
-T quantize_val(
-    float inv_scale,
+template <typename Q, typename F>
```
AdrianLundell, line 78: Why template this when only int8/f32 is supported?

backends/cortex_m/ops/op_quantize_per_tensor.cpp
```diff
@@ -123,16 +123,97 @@ Tensor& quantize_per_tensor_out(
...
+      0x6,
+      0xE,
+      0x3,
+      0xB,
+      0x7,
+      0xF};
+
+  float32x4_t inv_scale_vec = vdupq_n_f32(inv_scale);
+
+  // Magic number for float to int conversion, round to nearest even integer
```
AdrianLundell, line 150: Really cool trick, I didn't know about this. This looks better than the rounding I have suggested.

examples/arm/executor_runner/CMakeLists.txt
```diff
@@ -505,7 +504,18 @@ set_property(
   PROPERTY IMPORTED_LOCATION
            "${ET_BUILD_DIR_PATH}/kernels/quantized/libquantized_kernels.a"
 )
-
+add_library(cortex_m_ops_lib STATIC IMPORTED)
```
AdrianLundell, line 507: Would be nice with a comment here to explain the discrepancy to how the kernel registration library is used here compared to how it is described here https://docs.pytorch.org/executorch/stable/kernel-library-custom-aten-kernel.html#yaml-entry-api-high-level-architecture

examples/arm/aot_arm_compiler.py
digantdesai: You are right, IMO that should be tested independently?

A - original FP32 model
B - quantized model
C - lowered model

(1) A vs. B captures quantization noise
(2) B vs. C captures implementation noise

A vs. C isn't really useful in case of quantization esp when you are trying vet the implementation, which is the goal here with bundled-io to test the implementation on FVP, while (1) can be done using random op impls AoT or on x86.

backends/cortex_m/ops/op_quantize_per_tensor.cpp
digantdesai: Just copy-paste artifact, it shouldn't cause any size or perf overhead.

backends/cortex_m/ops/op_quantize_per_tensor.cpp
digantdesai: Thanks. This isn't too uncommon for f32 -> i32.

examples/arm/executor_runner/CMakeLists.txt
digantdesai: Yeah this isn't doing any selection yet. I left a comment on the top level PR.

examples/arm/aot_arm_compiler.py
AdrianLundell: Agreed, I don't think we need to change something here as long as we keep this in mind because I feel like we are lacking somewhat in (1) right now afaik.

backends/cortex_m/ops/op_quantize_per_tensor.cpp
AdrianLundell: Sure, I'm just thinking it could cause confusion about what is supported but no strong opinion here

examples/arm/executor_runner/CMakeLists.txt
AdrianLundell: Ack

backends/cortex_m/ops/op_quantize_per_tensor.cpp
AdrianLundell: I guess in CMSIS-NN everything was integer already so I didn't have to worry about that, but nice to learn.

AdrianLundell (approved):

digantdesai:
@zingo - `test_sigmoid_u85_BI_add_sigmoid` failure is related or flaky? I can try to reproduce locally.

zingo:
> @zingo - `test_sigmoid_u85_BI_add_sigmoid` failure is related or flaky? I can try to reproduce locally.
> 
> Update: Landing this given I don't think we are using these q/dq ops when lowering through tester, which I can do in the follow up PRs. Also U55 tests are passing.

There is a PR where we hope to have fixed the flakey sigmoid, and update vela that you brought into Meta for test/merge a few days ago. It will hopefully fix the flakeyness. Until it is merged i think rerun the test is the best option (or ignore it)

zingo:
This is the sigmoid fix PR mentioned: https://github.com/pytorch/executorch/pull/10548

PR #10783, opened 2025-05-09
Arm backend: Suppress colors in pre-push if non-interactive

perheld:
Dont try to print with colors in the pre-push script if the script is non-interactive. This is to avoid getting broken output in the CI which doesnt support colors.

Change-Id: I58b6a45de8b8403fa39ecbe299d380f3ca9fba62


zingo (approved):

zingo:
Macos fail unrelated

Issue #10677, opened 2025-05-04
[CMake] optimized_kernels and quantized_kernel will depend on portable_kernels.


kirklandsign:
### 🚀 The feature, motivation and pitch

For example, 
https://github.com/pytorch/executorch/blob/94f7b10be3d26739787d2e2ab4e2f73abc7b4e26/kernels/optimized/cpu/op_native_layer_norm.cpp#L128-L133
here it uses `check_layer_norm_args`, and it's defined in 
https://github.com/pytorch/executorch/blob/94f7b10be3d26739787d2e2ab4e2f73abc7b4e26/kernels/portable/cpu/util/normalization_ops_util.cpp#L76-L85

So it means
Either optimized_kernels/quantized_kernel depends on portable_kernels privately, or we split these helpers into its own lib.

Right now, it depends on user of optimized_kernels to pull in portable_kernels as well, so it's "it happens to work"

cc @larryliu0820 @jathu

swolchok:
@manuelcandales

kirklandsign:
Close for now. At least now the dependency is correct. Ideally we should factor out the common code

PR #10788, opened 2025-05-09
Make a separate target for kernel utils

kirklandsign:
Currently it's built into portable kernels only. Other kernels used it, and the runner must include portable kernel to avoid undefined symbols. We need to split these utils out. A fix for https://github.com/pytorch/executorch/issues/10677 without BC breaking.


tools/cmake/cmake_deps.toml
```diff
@@ -56,6 +56,20 @@ excludes = [
 deps = [
   "executorch",
   "executorch_core",
+  "extension_threadpool",
+  "kernels_util_all_deps",
+]
+
+[targets.kernels_util_all_deps]
+buck_targets = [
+  "//kernels/portable/cpu/util:all_deps",
```
JacobSzwejbka, line 65: if its used by other kernels can we move it out of portable/

JacobSzwejbka (approved):

tools/cmake/cmake_deps.toml
kirklandsign: I tried. So many things can be broken, so I wouldn't bother right now.

PR #10807, opened 2025-05-11
Update export and build from source docs

GregoryComer:
### Summary
This PR makes some small updates to the docs in a few places. In the Build from Source page, I've added an explicit command to generate a test model (add.pte) to use with the executor_runner.

In the Exporting and Lowering page, I've added some details on advanced export and lowering techniques, including state management, dynamic control flow, and multi-method .ptes.

### Test plan
I've built and viewed the docs locally to validate the changes.

cc @mergennachin @byjlw

### DIFF
--- a/docs/source/using-executorch-export.md
+++ b/docs/source/using-executorch-export.md
@@ -169,6 +169,76 @@ outputs = method.execute([input_tensor])
 
 For more information, see [Runtime API Reference](executorch-runtime-api-reference.md).
 
+## Advanced Topics
+
+While many models will "just work" following the steps above, some more complex models may require additional work to export. These include models with state and models with complex control flow or auto-regressive generation.
+See the [Llama model](https://github.com/pytorch/executorch/tree/main/examples/models/llama) for example use of these techniques.
+
+### State Management
+
+Some types of models maintain internal state, such as KV caches in transformers. There are two ways to manage state within ExecuTorch. The first is to bring the state out as model inputs and outputs, effectively making the core model stateless. This is sometimes referred to as managing the state as IO.
+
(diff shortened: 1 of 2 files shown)


mergennachin:
@JacobSzwejbka  - can you take a look at the new content?

docs/source/using-executorch-export.md
```diff
@@ -169,6 +169,76 @@ outputs = method.execute([input_tensor])
...
+    - Do not reassign the buffer variable. Instead, use `copy_` to update the entire buffer content.
+- Mutable buffers are not shared between multiple methods within a .pte.
+- In-place operations are replaced with non-in place variants, and the resulting tensor is written back at the end of the method execution. This can be a performance bottleneck when using `index_put_`.
+- Buffer mutations are not supported on all backends and may cause graph breaks and memory transfers back to CPU.
+
+Support for mutation is expiremental and may change in the future.
+
+### Dynamic (Data-Dependent) Control Flow
+
+Control flow is considered data-dependent if the path taken is not fixed at export-time. This is commonly the case when if or loop conditions depend on the value of a Tensor, such as a generator loop that terminates when an
```
JacobSzwejbka, line 196: >Control flow is considered data-dependent if the path taken is not fixed at export-time.

This isn't true. 

Its data dependent if its dependent on the data contained within the tensor rather than its shape. But shape based control flow can still be dynamic.

docs/source/using-executorch-export.md
```diff
@@ -169,6 +169,76 @@ outputs = method.execute([input_tensor])
...
+lowered = to_edge_transform_and_lower({
+    "encode": encode_ep,
+    "decode": decode_ep,
+}).to_executorch()
+```
+
+At runtime, the method name can be passed to `load_method` and `execute` on the `Module` class.
+
+Multi-method .ptes have several caveats:
+- Methods are individually memory-planned. Activation memory is not current re-used between methods.
```
JacobSzwejbka, line 220: Maybe worth mentioning you could write a custom plan to change this, or share the same arenas at runtime.

JacobSzwejbka (changes requested):

GregoryComer:
@JacobSzwejbka I've updated the wording in the two areas you pointed out. Any concerns with the updated wording?

JacobSzwejbka (approved):

byjlw (approved):

PR #10809, opened 2025-05-12
Arm backend: Rescale fixes for TOSA 1.0

oscarandersson8218:
### Summary
- Adds support for INT16 rescales for TOSA 1.0.
- Updates names of const-tensors for TOSA 1.0.

### Test plan
Tested through public and internal CI.


zingo (approved):

zingo:
Retriggered faild arm trunk job to check if it is just flakey

PR #10816, opened 2025-05-12
Update backends-coreml.md

metascroy:
Adds new "common issues" section to CoreML backend


GregoryComer (approved):
Thanks for adding this. In the long-term, is it feasible for us to transform the dtypes in the delegate when lowering to avoid needing a workaround?

metascroy:
> Thanks for adding this. In the long-term, is it feasible for us to transform the dtypes in the delegate when lowering to avoid needing a workaround?

Long term, we should upstream a change to coremltools to do this (but I don't think we should do it in ET).

PR #10827, opened 2025-05-12
Default to file load mode in module

GregoryComer:
Summary:
We've seen a number of people hit issues with mlock as the default load mode. In particular, mlock will fail above a certain PTE size and may require elevated permissions on some systems. We should not default to using mlock and can leave at as an opt-in optimization.

Differential Revision: D74607072

### DIFF
--- a/extension/module/module.h
+++ b/extension/module/module.h
@@ -53,7 +53,7 @@ class Module {
    */
   explicit Module(
       const std::string& file_path,
-      const LoadMode load_mode = LoadMode::MmapUseMlock,
+      const LoadMode load_mode = LoadMode::File,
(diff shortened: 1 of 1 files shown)


shoumikhin:
How about changing it to `File` by default?

GregoryComer:
> How about changing it to `File` by default, so that everything else becomes explicit and people know what they do? Btw, It is the default for [ObjC/Swift bindings](https://github.com/pytorch/executorch/blob/main/extension/apple/ExecuTorch/Exported/ExecuTorchModule.mm#L82).

That's fine with me. I'll swap it to File. I initially went with MMap because peak memory is reduced by quite a bit, but I'm okay to leave that as an opt-in optimization.

shoumikhin (approved):

PR #10831, opened 2025-05-12
[ET-VK] custom memory pools

SS-JIA:
* __->__ #10831

## Context

Currently, in ET-VK all memory allocations are done through the default memory pool of VMA (Vulkan Memory Allocator). This has some consequences:

* Memory Pool Behaviour (e.g. block size, allocation algorithm, etc.) cannot be configured on a per-model basis
* When a model is unloaded, GPU memory is not guaranteed to be freed since VMA may elect to hold on to the allocated memory for future use

This PR introduces the ability for models to use a custom VMA memory pool to allocate memory for tensors belonging to that model. This addresses the two points listed above, and makes it possible for VMA to release all memory used for a model once the model is unloaded. Additionally, using a custom VMA memory pool is optional.

Differential Revision: [D74617843](https://our.internmc.facebook.com/intern/diff/D74617843/)

### DIFF
--- a/backends/vulkan/runtime/api/Context.cpp
+++ b/backends/vulkan/runtime/api/Context.cpp
@@ -40,6 +40,10 @@ Context::Context(size_t adapter_i, const ContextConfig& config)
       cmd_mutex_{},
       cmd_(VK_NULL_HANDLE, 0u),
       submit_count_{0u},
+      // Custom memory pools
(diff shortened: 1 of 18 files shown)


igs-raiderler:
The performance optimizations here are quite clever. Nice engineering work!

PR #10839, opened 2025-05-13
Support prequant qwen3

metascroy:
As titled

### DIFF
--- a/examples/models/qwen3/convert_weights.py
+++ b/examples/models/qwen3/convert_weights.py
@@ -117,7 +117,7 @@ def main():
     parser.add_argument("output", type=str, help="Path to the output checkpoint")
 
     args = parser.parse_args()
-    convert_weights(args.input_dir, args.output)
+    convert_weights(args.input_dir_or_checkpoint, args.output)
 
 
 if __name__ == "__main__":
--- a/examples/models/qwen3/convert_weights.py
+++ b/examples/models/qwen3/convert_weights.py
@@ -13,6 +13,7 @@ from torchtune.models.convert_weights import get_mapped_key
 _QWEN_3_FROM_META = {
     "tok_embeddings.weight": "model.embed_tokens.weight",
     "norm.weight": "model.norm.weight",
+    "output.weight": "lm_head.weight",
     "layers.{}.attention.wk.weight": "model.layers.{}.self_attn.k_proj.weight",
     "layers.{}.attention.k_norm_fn.weight": "model.layers.{}.self_attn.k_norm.weight",
     "layers.{}.attention.wq.weight": "model.layers.{}.self_attn.q_proj.weight",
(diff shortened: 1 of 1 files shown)


examples/models/qwen3/convert_weights.py
```diff
@@ -47,20 +48,19 @@ def qwen_3_tune_to_meta(state_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.
...
-        # Tied embeddings for 0.6b and 4b models.
-        if key == "lm_head.weight":
-            continue
         new_key = get_mapped_key(key, inverted_mapping_dict)
         converted_state_dict[new_key] = value
 
-    converted_state_dict["output.weight"] = converted_state_dict[
-        "tok_embeddings.weight"
-    ]
+    # If lm_head.weight is not present, assume tied embeddings (e.g., 0.6b and 4b models)
```
jackzhxng, line 54:
```suggestion
    # If lm_head.weight is not present, assume tied embeddings (0.6b, 1.7b, and 4b models)
```

examples/models/qwen3/convert_weights.py
```diff
@@ -47,20 +48,19 @@ def qwen_3_tune_to_meta(state_dict: Dict[str, torch.Tensor]) -> Dict[str, torch.
...
+    if "lm_head.weight" not in state_dict:
```
jackzhxng, line 55: lm_head is present in the hf checkpoints even if they are tied embeddings, it will just be the same weights as the tok_embeddings

examples/models/qwen3/convert_weights.py
```diff
@@ -101,14 +110,14 @@ def main():
         description="Convert Qwen3 weights to Meta format."
     )
     parser.add_argument(
-        "input_dir",
+        "input_dir_or_checkpoint",
```
jackzhxng, line 113: not a big fan of specifying something that could be either a dir or a file, if `pytorch_model.bin` is the filename for the quantized checkpoint going forward i'd rather just specify the checkpoint and search for that

jackzhxng (commented):
have you tested the regular qwen3 export?

examples/models/qwen3/convert_weights.py
metascroy: how do you usually download the directories from HF?

examples/models/qwen3/convert_weights.py
metascroy: Hmmm, are you sure it's there?  I thought when config. tie_word_embeddings = true, it might not be there, but gets materialized during a tie_weights() command on the HF model. 

In any case, if it is there, it's covered by the regular loop through keys and this logic is not executed.  If it's not there, this sets lm_head's weight to the embeddings.

metascroy:
> have you tested the regular qwen3 export?

I can test if there is no CI for it.

examples/models/qwen3/convert_weights.py
jackzhxng: Yeah it's here, https://huggingface.co/Qwen/Qwen3-0.6B/tree/main?show_file_info=model.safetensors. Also I remember seeing it while debugging the checkpoint. But sure, in that case could you reword the comment?

examples/models/qwen3/convert_weights.py
jackzhxng: e.g. ` huggingface-cli download ibm-granite/granite-3b-code-instruct-128k`

jackzhxng:
Yeah if you test it and it works, feel free to merge

jackzhxng (approved):

examples/models/qwen3/convert_weights.py
metascroy: I changed to use directories.

metascroy:
> Yeah if you test it and it works, feel free to merge pending comments

Checked convert script works with original Qwen3 on HF

examples/models/qwen3/convert_weights.py
metascroy: Reworded the comment a little.

I'm not convinced by https://huggingface.co/Qwen/Qwen3-0.6B/tree/main?show_file_info=model.safetensors because it's just metadata, and doesn't prove anything about what is stored in the file.

It does not look like lm_head is present in the safetensors when I unpack them locally.  Perhaps you looked at the checkpoint after running your script?  (Which copied the embedding tensors into lm_head).

jackzhxng: Ok, if you double checked then that's fine!

PR #10845, opened 2025-05-13
Arm backend: Update NEGATE with TOSA 1.0 support

per:
### Summary
Add the serialization to TOSA 1.0 where the attributes has moved to input tensors instead.

### Test plan


zingo:
I rebased this to make sure it works after another big PR when in.

digantdesai (approved):

PR #10846, opened 2025-05-13
Arm backend: Add test for DeiT Tiny for TOSA BI

martinlsm:
Add test case for the DeiT Tiny model for the TOSA BI profile. At this time the output of the model differs from the reference implementation by a mean absolute error of around 2.5, which is too high. An internal ticket has been raised to resolve this issue.

### DIFF
--- a/backends/arm/test/models/test_deit_tiny_arm.py
+++ b/backends/arm/test/models/test_deit_tiny_arm.py
@@ -11,7 +11,10 @@ import timm
 
 import torch
 
-from executorch.backends.arm.test.tester.test_pipeline import TosaPipelineMI
+from executorch.backends.arm.test.tester.test_pipeline import (
+    TosaPipelineBI,
+    TosaPipelineMI,
+)
 
 from timm.data import IMAGENET_INCEPTION_MEAN, IMAGENET_INCEPTION_STD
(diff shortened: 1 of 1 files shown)


backends/arm/test/models/test_deit_tiny_arm.py
```diff
@@ -42,3 +45,16 @@ def test_deit_tiny_tosa_MI():
...
+
+
+def test_deit_tiny_tosa_BI():
+    pipeline = TosaPipelineBI[input_t](
+        deit_tiny,
+        model_inputs,
+        aten_op=[],
+        exir_op=[],
+        use_to_edge_transform_and_lower=True,
+        atol=3.0,  # This needs to go down: MLETORCH-956
```
digantdesai, line 57: Is it stable though? As opposed to flaky, else we can mark it as Flaky..

digantdesai (approved):
More transformers! Let's go.

backends/arm/test/models/test_deit_tiny_arm.py
martinlsm: @digantdesai  I believe it is rather stable. Been looping it several times myself and the numerical error looks to be consistent.

PR #10849, opened 2025-05-13
Arm backend: Update operator support for TOSA-1.0+INT+u55

per:
### Summary
Make is_U55_subset part of the base TOSA specification class as the
subset is not tied to a specific specification. Using the
is_U55_subset attribute the supported checks for U55 subset are
updated to catch TOSA 1.0 as well.

### Test plan


zingo:
Test fails unrelated fails on other PRs as well

zingo (approved):

PR #10850, opened 2025-05-13
Arm backend: Fix mypy linting in pre-push

AdrianLundell:
- Adds ./src/ to MYPYPATH to find imports for editable install
- Make --revision point to the commit before the commit to be linted. Since lintrunner compares files diffing between --revision and HEAD, pointing it to the the latest commit (== HEAD) results in no files being linted.

### DIFF
--- a/backends/arm/scripts/pre-push
+++ b/backends/arm/scripts/pre-push
@@ -76,7 +76,7 @@ for COMMIT in ${COMMITS}; do
 
     # lintrunner on latest patches.
     echo -e "${INFO} Lintrunner"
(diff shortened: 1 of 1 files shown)


AdrianLundell:
Unrelated error [trunk / test-models-linux-aarch64 (linear, portable, linux.arm64.2xlarge) / linux-job](https://hud.pytorch.org/pr/pytorch/executorch/10850#42138191326) ([gh](https://github.com/pytorch/executorch/actions/runs/14998329659/job/42138191326)): "error: incomplete-download"

zingo (approved):

PR #10855, opened 2025-05-13
Build flatcc for the host

jathu:
### Summary
* Similar to what we did with `flatc`, let's build `flatcc_cli` for the host using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html)
* Move the `flatcc` definitions to be under `third-party/`
* Move the `etdump` and `bundled_program` definitions to be under their respective folders
* Clean up the build files and remove redundant things

### Test plan
CI

```
$ ./install_executorch.sh

$ ./scripts/build_apple_frameworks.sh

$ rm -rf cmake-out \
   && cmake --preset macos-arm64 \
   && cmake --build cmake-out --parallel

$ rm -rf cmake-out \
   && cmake -DEXECUTORCH_BUILD_DEVTOOLS=ON -Dprotobuf_BUILD_TESTS=OFF -DEXECUTORCH_ENABLE_EVENT_TRACER=ON --preset macos-arm64 \
   && cmake --build cmake-out --parallel
```

cc @larryliu0820


kirklandsign (approved):

jathu:
cc @zingo I see that arm [build scripts](https://github.com/pytorch/executorch/blob/f1ef702ff35fdeb92c5affa0e9f8e1e4472ff731/backends/arm/scripts/build_executorch.sh#L117) set `FLATCC_EXECUTABLE`. We should be able to remove that now. I'll leave the change up to you

zingo:
Thanks for this change, we have an extra build and copy of that executable we now can clean up 🙏

PR #10858, opened 2025-05-13
Allow setting thread count from Java

GregoryComer:
Summary: Allow passing in an explicit thread count through the Java bindings.

Differential Revision: D74676049

### DIFF
--- a/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Module.java
+++ b/extension/android/executorch_android/src/main/java/org/pytorch/executorch/Module.java
@@ -50,6 +50,19 @@ public class Module {
    * @return new {@link org.pytorch.executorch.Module} object which owns the model module.
    */
   public static Module load(final String modelPath, int loadMode) {
+    return load(modelPath, loadMode, 0);
+  }
+
+  /**
+   * Loads a serialized ExecuTorch module from the specified path on the disk.
+   *
+   * @param modelPath path to file that contains the serialized ExecuTorch module.
+   * @param loadMode load mode for the module. See constants in {@link Module}.
+   * @param numThreads the number of threads to use for inference. A value of 0 defaults to a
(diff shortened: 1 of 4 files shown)


extension/android/executorch_android/src/main/java/org/pytorch/executorch/Module.java
```diff
@@ -52,7 +52,22 @@ public static Module load(final String modelPath, int loadMode) {
...
-    return new Module(new NativePeer(modelPath, loadMode));
+    return new Module(new NativePeer(modelPath, loadMode, 0));
+  }
+
+  /**
+   * Loads a serialized ExecuTorch module from the specified path on the disk.
+   *
+   * @param modelPath path to file that contains the serialized ExecuTorch module.
+   * @param loadMode load mode for the module. See constants in {@link Module}.
+   * @param numThreads number of threads to use for inference.
```
kirklandsign, line 63: Specify when the value is 0?

extension/android/executorch_android/src/main/java/org/pytorch/executorch/Module.java
```diff
@@ -52,7 +52,22 @@ public static Module load(final String modelPath, int loadMode) {
     if (!NativeLoader.isInitialized()) {
```
kirklandsign, line 52: Remove?

kirklandsign (approved):

extension/android/executorch_android/src/main/java/org/pytorch/executorch/Module.java
GregoryComer: I left the original constructor there for BC, though we could also just use a default value in the new constructor to maintain source compatibility. Are there any other ABI-style implications? What is your preference?

kirklandsign (approved):

PR #10870, opened 2025-05-14
Arm backend: Check in tosa.fbs for TOSA 0.80 and 1.0

oscarandersson8218:
### Summary
Add schema files for 0.80 and 1.0. This will enable dump_artifact and dump_operator_distribution for both 0.80 and 1.0. Also bumps reference model.

### Test plan


zingo (approved):

zingo:
lintrunner lama, llava and llama macos fails are unrelated.

PR #10872, opened 2025-05-14
Arm backend: Add validation for same dtype to operators

Sebastian-Larsson:
When applicable, check that the data types of inputs to a given operator are the same.


zingo (approved):

zingo:
lintrunner llama and llava fails unrelated,

PR #10878, opened 2025-05-14
Dtype selective build for optimized ops

lucylq:
* #10990
* __->__ #10878
* #10877

Add dtype selective build for optimized ops. Follows the same process as portable, where we copy the source files and rebuild the library.

1. Generalize copy genrule for portable/optimized/source/header.
2. Copy optimized source files + headers.
3. Build optimized ops using source files, dependencies, portable header.
4. Add test, confirm that we can run addmul with float dtypes (when we remove, the test fails).

Differential Revision: [D74688554](https://our.internmc.facebook.com/intern/diff/D74688554/)


iseeyuan (approved):
LGTM. What's the binary size save for add_mul.pte in the test?

PR #10892, opened 2025-05-15
Introduce export LLM config

jackzhxng:
Introduce dataclasses to configure `export_llama` (and later `export_llm`). To be used with OmegaConf's `DictConfig` and Hydra.

### DIFF
--- a/examples/models/llama/config/llm_config.py
+++ b/examples/models/llama/config/llm_config.py
@@ -1,0 +1,151 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-unsafe
+
+"""
+Configurations for exporting Llama.
+
+Uses dataclases, which integrate with OmegaConf and Hydra.
+"""
+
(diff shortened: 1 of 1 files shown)


examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,151 @@
...
+@dataclass
+class BaseConfig:
+    """
+    These are specific to the specific model, e.g. whether it’s Qwen3 0.6B or Phi-4-mini.
+    for each of these different models, you can expect each of these fields to change.
+    """
+
+    model_class: str = "llama"
+    params: Optional[str] = None
+    checkpoint: Optional[str] = None
```
lucylq, line 29:
```
adapter_checkpoint: Optional[str] = None
```

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,151 @@
...
+    checkpoint_dir: Optional[str] = None  # For sharded checkpoint
+    tokenizer_path: Optional[str] = None
+    metadata: Optional[str] = None
+    fairseq2: bool = False  # For legacy internal use cases
```
lucylq, line 33:
```
# lora args
lora_rank: Optional[int] = None 
lora_alpha: Optional[int] = None
# Eg. q_proj, k_proj, v_proj, output_proj
lora_target_modules: Optional[List] = None
```

PR #10901, opened 2025-05-15
Arm backend: Add arange.default dummy tests

AdrianLundell:
Since we use unittests for coverage on the Edge IR op set which includes arange.default, we need tests for it even though it is decomposed to indicate that it has not been missed.


AdrianLundell:
Failures unrelated, this patch only introduces four new skipped tests, so very unlikely to cause any issues.

zingo (approved):

PR #10902, opened 2025-05-15
Arm backend: Update parse_test_name script

AdrianLundell:
- Make TARGETS a list to preserve order.
- Replace edge_ops with a dict mapping test op names to edge op names, e.g. op_name_map["split_tensor"] = "split_copy.Tensor"

Both changes are slightly less efficient compared to using sets, but more convenient to use when handling the results.


AdrianLundell:
Failures unrelated, this patch only touches a python script  run in our pre-push hook.

zingo (approved):

PR #10916, opened 2025-05-15
Fix lint

cccclai:
Differential Revision: D74825457


kirklandsign (approved):

jackzhxng (changes requested):
We shouldn't be fixing the lint this way, we are in the process of explicitly removing args from export_llama_lib. I can look into this

PR #7570, opened 2025-01-09
Work around flatccrt issue in CI script

benkli01:
When running build-qnn-sdk.sh in certain circumstances the libflatccrt.a cannot be found. To work around this we can build the QNN backend twice, second time with option `--no_clean`, to make sure the library is found.

Fixes #7300

Issue #7300, opened 2024-12-12
Missing libflatccrt.a when building executor_runner


benkli01:
### 🐛 Describe the bug

There is a CI failure in the QNN tests of PR https://github.com/pytorch/executorch/pull/5027:
```gmake[2]: *** No rule to make target '../third-party/flatcc/lib/libflatccrt.a', needed by 'executor_runner'.  Stop.```

To reproduce:

1. Checkout the changes from PR https://github.com/pytorch/executorch/pull/5027
2. Download Android NDK and QNN SDK according to [this guide](https://pytorch.org/executorch/stable/build-run-qualcomm-ai-engine-direct-backend.html)
3. Set up these environment variables accordingly: `EXECUTORCH_ROOT`, `QNN_SDK_ROOT`, `ANDROID_NDK_ROOT`
4. (not sure if required) Set env vars `LD_LIBRARY_PATH=$QNN_SDK_ROOT/lib/x86_64-linux-clang/:${LD_LIBRARY_PATH:+"$LD_LIBRARY_PATH"}` and `export PYTHONPATH=$EXECUTORCH_ROOT/..`
5. Run the following command from the executorch root: `backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release` (used by [CI here](https://github.com/pytorch/executorch/blob/main/.ci/scripts/build-qnn-sdk.sh#L17))

Some observations:

- The issue disappears for me locally when running step 5 above without `--job_number 2`.
- Replacing `${FLATCCRT_LIB}` with `flatccrt` or removing the dependency altogether (relying on inheritance from etdump) from files `CMakeLists.txt` and `examples/qualcomm/executor_runner/CMakeLists.txt` does not change the behavior.
- Possible workaround: When running step 5 above twice but adding `--no_clean` the second time the problem seems to be gone.

### Versions

PyTorch version: 2.5.0+cpu
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.2 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: version 3.30.3
Libc version: glibc-2.35

Python version: 3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-6.8.0-1019-aws-x86_64-with-glibc2.35
[...]

PR #10918, opened 2025-05-15
Fix libflatccrt race

jathu:
### Summary

Seems like there is a race in building `libflatccrt.a`. This issue has existed for a while: https://github.com/pytorch/executorch/issues/7300. It was temporarily mitigated in https://github.com/pytorch/executorch/pull/7570 by just reducing the parallelism. 

In this diff I attempt to fix it. This is just my assumption of what is wrong. Given flatccrt builds a debug version with a `_d` suffix, if the target isn't depended on (i.e. some target don't use the conditional target name) then the order of how the lib is built causes a race. So for now, always use the non-debug version.

Given it's a race, I was never able to repro the issue locally — I can't guarantee this is the problem. However, it seems my recent changes in https://github.com/pytorch/executorch/pull/10855 has increased the frequency of the problem in CI.  

### Test plan

CI

cc @larryliu0820

### DIFF
--- a/.ci/scripts/build-qnn-sdk.sh
+++ b/.ci/scripts/build-qnn-sdk.sh
@@ -11,17 +11,12 @@ set -o xtrace
 
 build_qnn_backend() {
   echo "Start building qnn backend."
-  export ANDROID_NDK_ROOT=/opt/ndk
-  export QNN_SDK_ROOT=/tmp/qnn/2.28.0.241029
+  export ANDROID_NDK_ROOT=${ANDROID_NDK_ROOT:-/opt/ndk}
+  export QNN_SDK_ROOT=${QNN_SDK_ROOT:-/tmp/qnn/2.28.0.241029}
   export EXECUTORCH_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../.." && pwd)"
 
-  # Workaround to avoid issues around missing flatccrt library (depending on the
-  # number of jobs used), see issue #7300:
-  # Build twice (second time with `--no_clean`) to make sure libflatccrt.a is
-  # available.
-  # TODO: Remove this workaround once the underlying issue is fixed.
-  bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release || \
-  bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release --no_clean
+  parallelism=$(( $(nproc) - 1 ))
+  bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number ${parallelism} --release
 }
 
 set_up_aot() {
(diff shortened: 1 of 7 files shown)


kirklandsign (approved):

.ci/scripts/build-qnn-sdk.sh
```diff
@@ -11,17 +11,12 @@ set -o xtrace
...
+  export QNN_SDK_ROOT=${QNN_SDK_ROOT:-/tmp/qnn/2.28.0.241029}
   export EXECUTORCH_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../.." && pwd)"
 
-  # Workaround to avoid issues around missing flatccrt library (depending on the
-  # number of jobs used), see issue #7300:
-  # Build twice (second time with `--no_clean`) to make sure libflatccrt.a is
-  # available.
-  # TODO: Remove this workaround once the underlying issue is fixed.
-  bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release || \
-  bash backends/qualcomm/scripts/build.sh --skip_aarch64 --job_number 2 --release --no_clean
```
larryliu0820, line 24: Is this change related?

examples/apple/mps/CMakeLists.txt
```diff
@@ -97,12 +97,6 @@ if(NOT CMAKE_TOOLCHAIN_FILE MATCHES ".*(iOS|ios\.toolchain)\.cmake$")
   list(TRANSFORM _mps_executor_runner__srcs PREPEND "${EXECUTORCH_ROOT}/")
   add_executable(mps_executor_runner ${_mps_executor_runner__srcs})
 
-  if(CMAKE_BUILD_TYPE MATCHES "Debug")
-    set(FLATCC_LIB flatccrt_d)
-  else()
-    set(FLATCC_LIB flatccrt)
-  endif()
-
```
larryliu0820, line 105: Have you confirmed that debug build will also build `flatccrt`?

larryliu0820 (approved):

kirklandsign (approved):

.ci/scripts/build-qnn-sdk.sh
jathu: Yes, see the comment right above — this was the way the race was mitigated before in https://github.com/pytorch/executorch/pull/7570

examples/apple/mps/CMakeLists.txt
jathu: Yes, it seems to have both libs

shewu-quic: Hi @jathu,
[In QNN build script,](https://github.com/pytorch/executorch/blob/6cceab6209ff7409b63c4700186ed949b82462b7/backends/qualcomm/scripts/build.sh#L33) debug build is default setting. We encounter the link error which is related to find library -lflatccrt. It seems to build libflatccrt_d instead of flatccrt in debug build. May I check is it safe to use flatccrt instead of ${FLATCCRT_LIB} .
 
image

Error message
```
ld.lld: error: unable to find library -lflatccrt
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [executor_runner/CMakeFiles/qnn_executor_runner.dir/build.make:184: executor_runner/qnn_executor_runner] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:5174: executor_runner/CMakeFiles/qnn_executor_runner.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....

```

cccclai:
@jathu, looks like this causing issue to build qnn backend in debug mode. Looks like the script to build qnn sdk only covers release mode though

PR #10920, opened 2025-05-15
Update remove clone to drop no-op q/dq

JakeStevens:
Summary:
After removing clone, we may be left with no-op quantize operations.

This diff updates the pass in backend/transforms to remove these, if they exist

Differential Revision: D74832417


digantdesai (approved):

JakeStevens:
Failing tests pre-existing. moshi from undefined libtorchaudio symbol, openvino from backend import. Landing.

PR #10927, opened 2025-05-15
Make test_no_replace_quant_permute_dequant_with_requantize use GraphBuilder.

eigen-k:
Summary: Use GraphBuilder to create the model for unit testing.

Differential Revision: D74842294

### DIFF
--- a/backends/cadence/aot/tests/test_fusion_ops_passes.py
+++ b/backends/cadence/aot/tests/test_fusion_ops_passes.py
@@ -295,11 +295,12 @@ class TestFusionPasses(TestFusionPassesBase):
             args=(permute, 4.5, 6, 0, 127, torch.int8),
         )
         builder.output(dequant)
-        graph_module = FuseQuantDequantToRequantizePass(
+        original_graph = builder.get_graph_module()
+        converted_graph = FuseQuantDequantToRequantizePass(
             force_quant_dequant_fusion=False
-        )(builder.get_graph_module()).graph_module
+        )(original_graph).graph_module
         self.check_op_counts(
-            graph_module,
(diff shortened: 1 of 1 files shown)


backends/cadence/aot/tests/test_fusion_ops_passes.py
```diff
@@ -281,25 +281,23 @@ def forward(self, x):
...
+        permute = builder.call_operator(
+            op=exir_ops.edge.aten.permute_copy.default, args=(quant, [2, 0, 1, 3])
+        )
+        dequant = builder.call_operator(
+            op=exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
+            args=(permute, 4.5, 6, 0, 127, torch.int8),
+        )
+        builder.output(dequant)
+        graph_module = FuseQuantDequantToRequantizePass(
+            force_quant_dequant_fusion=False
```
mcremon-meta, line 299: any reason to add the flag here?

mcremon-meta (approved):

backends/cadence/aot/tests/test_fusion_ops_passes.py
eigen-k: Although the parameter defaults to False, I chose to set it explicitly to False to avoid any ambiguity, given that the adjacent test has it set to True. Does it seem redundant?

mcremon-meta: I have no issue with it, was just wondering if there was a reason, but I noticed the other test being True after this review :)

zonglinpeng (approved):

PR #10932, opened 2025-05-16
Create a pybind preset

jathu:
### Summary

In this diff we create a `pybind` preset that is copied with configurations from [setup.py](https://github.com/pytorch/executorch/blob/7175ca420dc5a173f8635da976457bf6f17bbbc1/setup.py). In upcoming diffs, we will completely gut setup.py and point directly to this preset.

### Test plan

```
$ cmake --list-presets
Available configure presets:

  "macos-arm64"  - Build everything buildable on macOS arm64
  "pybind" - Build pybindings exported in the wheel
```
```
$ cmake --preset pybind && cmake --build cmake-out --parallel
```

cc @larryliu0820

### DIFF
--- a/kernels/quantized/CMakeLists.txt
+++ b/kernels/quantized/CMakeLists.txt
@@ -10,10 +10,6 @@
 # ~~~
 cmake_minimum_required(VERSION 3.19)
 
-option(EXECUTORCH_BUILD_KERNELS_QUANTIZED_AOT
-       "Build the optimized ops library for AOT export usage" OFF
-)
-
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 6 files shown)


tools/cmake/preset/default.cmake
```diff
@@ -227,13 +237,41 @@ define_overridable_option(
...
   "Build the executor_runner executable"
   BOOL ${_default_executorch_build_executor_runner}
 )
 
+# NB: Enabling this will serialize execution of delegate instances Keeping this
+# OFF by default to maintain existing behavior, to be revisited.
+define_overridable_option(
+  EXECUTORCH_XNNPACK_SHARED_WORKSPACE
+  "Enable workspace sharing across different delegate instances"
+  BOOL ON
```
larryliu0820, line 254: Shouldn't this be OFF by default, based on the comment?

jathu: It's currently on by default:

https://github.com/pytorch/executorch/blob/a0d9c7e68fa9c36c62b6849fd606622ff72a7c0a/backends/xnnpack/CMakeLists.txt#L28-L32

I'm just moving the option definition here from `backends/xnnpack/CMakeLists.txt`

kernels/quantized/CMakeLists.txt
```diff
@@ -91,7 +87,13 @@ if(NOT CMAKE_GENERATOR STREQUAL "Xcode"
     # pybindings.
     if(TARGET portable_lib)
       add_library(quantized_pybind_kernels_lib ${_quantized_kernels__srcs})
-      target_link_libraries(quantized_pybind_kernels_lib PRIVATE portable_lib)
+      target_link_libraries(
+        quantized_pybind_kernels_lib
+        PRIVATE
+          portable_lib
+          executorch_core
+          kernels_util_all_deps
```
larryliu0820, line 95: why are we depending on more libs?

jathu: These seem to be implicitly used — without them, builds are failing. Note that `quantized_kernels` uses the same sources and includes these as dependencies:

https://github.com/pytorch/executorch/blob/a0d9c7e68fa9c36c62b6849fd606622ff72a7c0a/kernels/quantized/CMakeLists.txt#L140-L141

larryliu0820 (approved):

PR #10938, opened 2025-05-16
Add a android log implementation

kirklandsign:


runtime/platform/targets.bzl
```diff
@@ -41,10 +41,12 @@ def define_common_targets():
...
         name = "platform_private",
-        srcs = _select_pal({
-            "minimal": ["default/minimal.cpp"],
-            "posix": ["default/posix.cpp"],
-        }),
+        srcs = select({
+            "ovr_config//os:android": ["default/android.cpp"],
+            "DEFAULT": _select_pal({
+                "minimal": ["default/minimal.cpp"],
+                "posix": ["default/posix.cpp"],
```
larryliu0820, line 48: I'm wondering if this should be another "android" entry

runtime/platform/targets.bzl
```diff
@@ -41,12 +41,15 @@ def define_common_targets():
...
+        })}),
         deps = [
             ":pal_interface",
+            "fbsource//third-party/toolchains:log",
```
larryliu0820, line 52: This will break in OSS buck right? Can you also hide it under ovr_config android?

runtime/platform/targets.bzl
kirklandsign: > However, it might also cause some confusion for users since executorch.pal_default is now ignored

That makes sense, but ovr_config//os:android is automatic.

jathu (approved):
build changes lgtm

PR #10944, opened 2025-05-16
Arm backend: Refactor test_scalars to new naming standard

AdrianLundell:


AdrianLundell:
Errors unrelated, this is just a arm unittest change

zingo (approved):

PR #10950, opened 2025-05-16
Arm backend: test_conformer.py emergency fix

zingo:
### Summary
Some tests was changed from expected fail to pass in a recent PR but they turned out to not work.
Lets back them out again and investigate.
This make the unittest-arm-backend-with-no-fvp (test_pytest_models) test work again.

### Test plan
By hand and in GitHub

### DIFF
--- a/backends/arm/test/models/test_conformer.py
+++ b/backends/arm/test/models/test_conformer.py
@@ -63,6 +63,7 @@ class TestConformer(unittest.TestCase):
             )
         )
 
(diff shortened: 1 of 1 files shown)


zingo:
Hmm this is probably not the proper fix. Tests pass sometime som Im wondering if it's a rosource problem as we run the models on the same machine in parallell. Need to test some more.

zingo:
This was not a problem, close without merge

PR #10951, opened 2025-05-16
Use "pybind" preset when building a wheel

jathu:
### Summary

Now that we have presets for pybind, lets use that and eliminate the need to set cmake configurations in `setup.py`. Moreover, we now stop relying on defaults and fully rely on the cmake configurations. This includes relying on `CMakeCache.txt` to determine what features are enabled to determine what extensions to copy.

### Test plan

CI +

```
$ rm -rf pip-out && python setup.py bdist_wheel
$ rm -rf pip-out && ./install_executorch.sh
```

cc @larryliu0820


kirklandsign (approved):

backends/openvino/scripts/openvino_build.sh
```diff
@@ -48,16 +48,11 @@ main() {
...
-                           -DEXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL=ON \
-                           -DEXECUTORCH_ENABLE_LOGGING=ON \
-                           -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
-                           -DEXECUTORCH_BUILD_PYBIND=ON"
+                           -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON"
         export CMAKE_BUILD_ARGS="--target openvino_backend"
 
         # Build the package
-        pip install . --no-build-isolation
+        ./install_executorch.sh
```
larryliu0820, line 55: ah so good

setup.py
```diff
@@ -678,237 +650,169 @@ def initialize_options(self):
...
+                [
+                    "cmake",
+                    *cmake_configuration_args,
+                    "--preset",
+                    "pybind",
+                    "-B",
+                    cmake_cache_dir,
+                ],
+                check=True,
+            )
```
larryliu0820, line 703: Much better!

larryliu0820 (approved):
Thank you for making CMake much cleaner

PR #10953, opened 2025-05-16
Arm backend: Do not run model unit tests in parallel

zingo:
Runing to many of the larger models in parallel can cause some resource problems on the machine..

cc @digantdesai @freddan80 @per @oscarandersson8218

PR #10958, opened 2025-05-17
Arm backend: Remove fast scratch part for now

kirklandsign:
Summary: Fix CI

Differential Revision: D74939323

### DIFF
--- a/backends/arm/runtime/EthosUBackend.cpp
+++ b/backends/arm/runtime/EthosUBackend.cpp
@@ -189,8 +189,6 @@ class EthosUBackend final : public ::executorch::runtime::BackendInterface {
     // the end of the execution of the Ethos-U custom delegate
     char* ethosu_scratch =
         static_cast<char*>(temp_allocator->allocate(handles.scratch_data_size));
-    extern size_t ethosu_fast_scratch_size;
-    extern unsigned char* ethosu_fast_scratch;
     ET_LOG(
         Debug,
         "EthosUBackend::execute: Running program data:\n  cmd %p %zu\n  weight %p %zu\n  scratch %p %zu\n  fast scratch %p %zu\n",
@@ -200,8 +198,8 @@ class EthosUBackend final : public ::executorch::runtime::BackendInterface {
         handles.weight_data_size,
         ethosu_scratch,
         handles.scratch_data_size,
-        ethosu_fast_scratch,
-        ethosu_fast_scratch_size);
+        nullptr,
+        0);
 
     // Write argument values (from EValue tensor) into Ethos-U scratch
     // TODO(MLETORCH-123): Optimise into direct write from Vela into the SRAM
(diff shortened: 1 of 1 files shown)


zingo:
Hi, sorry for the break, from our (Arm) point just do what is easier for you to just make it work. Revert or this fix and we will look into this when back to work next week.

To understand the problem you see better and if you know, is it so that you have a different runner then our example runner you are testing with?

zingo (approved):
Ok, to just make it work for you.

kirklandsign:
> To understand the problem you see better and if you know, is it so that you have a different runner then our example runner you are testing with?

Seems that our internal CI has a different runner than the example runner.

zingo:
** > Seems that our internal CI has a different runner than the example runner.**

Thank that makes sense as you get this problem, and something we will try to keep in mind.

zingo:
Regarding the failed Arm unit model test it is unrelated an a fix for it is (hopefully) here 
https://github.com/pytorch/executorch/pull/10953

kirklandsign:
Hi @zingo seems that the test-arm-backend (test_models_ethos-u85) / linux-job always failed after this patch. 

https://github.com/pytorch/executorch/actions/runs/15101696584/job/42443518087

Could you please suggest how to fix?

gggekov:
Hi @kirklandsign, 
Yes, when you pass **nullptr** & **0** as base address/base address size for the U85 in Dedicated_Sram memory mode, the inference hangs as you see in your CI trace - that behaviour is expected. Dedicated_Sram means that the NPU uses the fast scratch buffer as a buffer to store the most commonly accessed intermediate tensors. With your fix, you are still testing Dedicated_Sram on U85, but you are not providing fast scratch array, hence the NPU hangs. 

Can you please expand a bit more on what you mean when you say you use a different arm_executor_runner.cpp in your internal test suite - do you have a way to define the ethosu_fast_scratch/ethosu_fast_scratch_size to be 0 or 384KB depending on if we test U55(Shared_Sram) or U85(Dedicated_Sram) in your internal arm_executor_runner.cpp?  

The U85 is designed to enable NNs where the scratch buffer is too big to fit into the SRAM and hence we place it in the DDR. In order to achieve good performance, it is important to still "dedicate" small amount of SRAM where the NPU can R/W the most commonly accessed tensors of the NN. That behaviour is enabled by the Dedicated_Sram memory mode, so it's important we support it properly.

kirklandsign:
@zingo @hsharma35 Sorry I am not the best contact for this issue. I don't work on this and just ran into an internal error. Could use @digantdesai 's help

digantdesai:
I think weak symbol override is fine. As long as we don't tightly couple the delegate runtime vs. app runner.

> I don't think we need a backend specific configuration to enable the Dedicated_Sram mode on the U85

Generally speaking, I see this mechanism as a cleaner mechanism to pass information from outside to a delegate especially when it originates at runtime. For compile-time information we can use other mechanisms like compile_spec, preprocessor macros/variables, etc.

gggekov:
FYI - here is a pr(https://github.com/pytorch/executorch/pull/11459) with the suggested fix with the weak symbol

PR #10967, opened 2025-05-19
Arm backend: Add support for BN fusing during QAT

oscarandersson8218:
Makes it possible to annotate patterns with more than two operators. This allows us to annotate patterns: conv -> bn and conv -> bn -> relu to be able to fold away BN after training in QAT. Also adds support for QAT in Tester class.


mcr229 (approved):

zingo:
Rebased after some Arm test fixes

zingo (approved):

PR #10975, opened 2025-05-19
Enable OSS coreml python tests

metascroy:
As titled


JacobSzwejbka (approved):

JacobSzwejbka:
Only concern is mac env tests are expensive, but obviously testing stuff like this is what they are for.

shoumikhin (approved):

pytest.ini
```diff
@@ -47,6 +47,8 @@ addopts =
     --ignore=backends/xnnpack/test/ops/test_sdpa.py
     backends/xnnpack/test/passes
     backends/xnnpack/test/serialization
+    # backends/apple/coreml
```
jathu, line 50: nit: delete
```suggestion
```

pytest.ini
```diff
@@ -47,6 +47,8 @@ addopts =
...
+    backends/apple/coreml/test
```
jathu, line 51: This will fail on non-Apple platforms. We need to see if we can selectively include this only on macOS

metascroy: @jathu yeah, I saw this in the CI failures.  Is there currently a good way to specify this granularity in ET's testing framework?

jathu: Maybe try adding a `--ignore` option for linux here:

https://github.com/pytorch/executorch/blob/cb3eba0d7f630bc8cec0a9cc1df8ae2f17af3f7a/.ci/scripts/unittest-linux-cmake.sh#L10C1-L10C7

PR #10996, opened 2025-05-20
Implement index_put inplace

manuelcandales:
Differential Revision: D74765433


kernels/portable/cpu/op_index_put.cpp
```diff
@@ -154,6 +157,175 @@ Tensor& index_put_out(
...
+        values, {expected_values_size, expected_values_dim}),
+    "Special case in-place index_put requires values to match input shape except for indexed dim");
+#endif // ET_LOG_ENABLED
+
+  return true;
+}
+
+} // namespace
+
+Tensor& index_put_(
```
JacobSzwejbka, line 267: is it worth making the out variant call the inplace one on out?

JacobSzwejbka (approved):

PR #11006, opened 2025-05-20
Add GELU support to XNNPACK

leafs1:
### Summary
Add GELU support to XNNPACK

### Test plan
Wrote tests and asserted that xnnpack delegate would run for GELU op


.gitignore
```diff
@@ -40,3 +40,18 @@ xcuserdata/
 .swiftpm/
 *.xcworkspace/
 *.xcframework/
+
+# misc
+/.vscode/
+backends/apple/coreml/executorchcoreml.cpython-312-darwin.so
+/cmake_wrapper.sh
+/data/
+/devtools/bundled_program/serialize/
```
mcr229, line 49: these seem a little suspicious? maybe @shoumikhin or @larryliu0820  can comment on some of these, i'm guessing they come from the --editable path.

Do you mind putting this up as a separate PR? I see it's a separate commit, so would be a nice time to try out ghstack

backends/xnnpack/operators/op_gelu.py
```diff
@@ -0,0 +1,52 @@
...
+from executorch.backends.xnnpack.serialization.xnnpack_graph_schema import (
+    XNNGelu,
+    XNNGraph,
+    XNode,
+)
+from executorch.backends.xnnpack.utils.utils import get_input_node
+
+
+@register_node_visitor
+class LogVisitor(NodeVisitor):
```
mcr229, line 23: wrong name?

.gitignore
leafs1: When building locally all of these directories got generated, I just wanted to add them to gitignore so i can add files to git easier. I'll make a separate pr

mcr229 (approved):

PR #11015, opened 2025-05-20
migrate convert/prepare to torchao

metascroy:
This PR migrates the convert/prepare pt2e APIs to torchao, and adds a linter to prevent using the APIs from torch.ao.

It also migrates the example quantizer to fully use torchao APIs.

### DIFF
--- a/backends/example/example_quantizer.py
+++ b/backends/example/example_quantizer.py
@@ -11,9 +11,9 @@ import torch
 from executorch.backends.example.example_operators.ops import module_to_annotator
 from executorch.backends.xnnpack.quantizer.xnnpack_quantizer_utils import OperatorConfig
 from torch import fx
-from torch.ao.quantization.observer import HistogramObserver, MinMaxObserver
-from torch.ao.quantization.pt2e.graph_utils import find_sequential_partitions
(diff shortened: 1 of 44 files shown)


jerryzh168 (approved):

backends/example/example_quantizer.py
```diff
@@ -11,9 +11,9 @@
 from executorch.backends.example.example_operators.ops import module_to_annotator
 from executorch.backends.xnnpack.quantizer.xnnpack_quantizer_utils import OperatorConfig
 from torch import fx
-from torch.ao.quantization.observer import HistogramObserver, MinMaxObserver
-from torch.ao.quantization.pt2e.graph_utils import find_sequential_partitions
-from torch.ao.quantization.quantizer import QuantizationSpec, Quantizer
+from torchao.quantization.pt2e.graph_utils import find_sequential_partitions
+from torchao.quantization.pt2e.observer import HistogramObserver, MinMaxObserver
```
jerryzh168, line 15: do we want to import from `pt2e` directly? or do you feel specifying `graph_utils` / `observer` are better

metascroy: I like graph_utils/observer personally

But people can use top-level too if they want

PR #11018, opened 2025-05-20
Build Windows pybind in CI

jathu:
### Summary

Screenshot 2025-05-22 at 11 17 18 AM

### Test plan
CI

cc @larryliu0820

### DIFF
--- a/.github/workflows/build-presets.yml
+++ b/.github/workflows/build-presets.yml
@@ -65,4 +65,30 @@ jobs:
 
         ./install_requirements.sh > /dev/null
         cmake --preset ${{ matrix.preset }}
-        cmake --build cmake-out -j$(( $(nproc) - 1 ))
+
+  windows:
+    uses: pytorch/test-infra/.github/workflows/windows_job.yml@main
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: [pybind]
+    with:
+      job-name: build
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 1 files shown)


SS-JIA (approved):

.github/workflows/build-presets.yml
```diff
@@ -66,3 +66,30 @@ jobs:
...
+        preset: [pybind]
+    with:
+      job-name: build
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      submodules: recursive
+      timeout: 90
+      script: |
+        set -eux
+        conda init powershell
+        powershell -Command "& {
```
SS-JIA, line 84: very nice! So it still doesn't work via bash? If so, very curious why powershell specifically is needed...

jathu: Yeah there are some compiler errors when using bash. I think this is good for now, we can look into the bash issue later

.github/workflows/build-presets.yml
```diff
@@ -66,3 +66,30 @@ jobs:
         ./install_requirements.sh > /dev/null
         cmake --preset ${{ matrix.preset }}
         cmake --build cmake-out --parallel
+
+  windows:
+    uses: pytorch/test-infra/.github/workflows/windows_job.yml@main
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: [pybind]
```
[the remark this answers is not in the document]
jathu: I'll follow up, just want to get this in first

PR #11028, opened 2025-05-21
Add new export LLM config

### LINKED ISSUE #9449: [etLLM] New config system to export_llama

jackzhxng:
* #11162
* #11081
* #11029
* __->__ #11028

Differential Revision: [D75263991](https://our.internmc.facebook.com/intern/diff/D75263991)


jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263991).

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,270 @@
...
+class PreqMode(str, Enum):
+    PREQ_8DA4W = "8da4w"
+    PREQ_8DA4W_OUT_8DA8W = "8da4w_output_8da8w"
+
+
+@dataclass
+class BaseConfig:
+    """
+    These are specific to the specific model, e.g. whether it’s Qwen3 0.6B or Phi-4-mini.
+    For each of these different models, you can expect each of these fields to change.
```
lucylq: Find this docstring a bit confusing.

> For each of these different models, you can expect each of these fields to change.

Do we expect variation in the fields for each model (ie, user/we add different fields), or the values?

Suggestion:
```
The BaseConfig sets the parameters required to construct the model in eager mode. Expect the values of each field to be different for each model (e.g. Qwen3 0.6B and Phi-4-mini will have different checkpoints).
```

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,270 @@
...
+from enum import Enum
+from typing import List, Optional
+
+
+################################################################################
+################################## BaseConfig ##################################
+################################################################################
+
+
+class ModelType(str, Enum):
```
lucylq, line 27: Add a docstring here - could you add one for each class?

```
Describes the model family. Within each ModelType enum, there can be variation in the weight classes. For example, LLAMA3 comes in 1B, 3B and 8B. 
```

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,270 @@
...
+    preq_group_size: int = 32
+    preq_embedding_quantize: str = "8,0"
+
+
+################################################################################
+################################# ModelConfig ##################################
+################################################################################
+
+
+class DtypeOverride(str, Enum):
```
lucylq, line 100: Can we add docstring here as well, and also clarify when the model is converted? Iirc there's some logic to decide when to convert?

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+@dataclass
+class MPSConfig:
+    enabled: bool = False
+
+
+@dataclass
+class BackendConfig:
+    xnnpack: XNNPackConfig = field(default_factory=XNNPackConfig)
```
lucylq, line 446: Wdyt about making these optional configs, set to None?

I think that could make it easier to tell when a backend is enabled or not, instead of looking through each backend config to check if enabled=False.

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,270 @@
...
+class ModelConfig:
+    """
+    These are not necessarily specific to the model, but are needed to finish off
+    the rest of the model configuration in eager. You can think of these like
+    optimizations / actual configurations. The same ModelConfig can be applied
+    to different models.
+    """
+
+    dtype_override: DtypeOverride = DtypeOverride.FP32
+    enable_dynamic_shape: bool = True
```
lucylq, line 148: Dynamic shapes default to true - does the user have to provide the dynamic shapes for export, or is it available for each ModelType?

jackzhxng: Yeah, its the same for each model

examples/models/llama/config/llm_config.py
jackzhxng: Mind if I follow up with this change? I would like to keep the changes in this diff as small as possible (ie as close to the original args as possible) to make it easier to pass all the internal CI

examples/models/llama/config/llm_config.py
lucylq: Ok, asking because if a user brings their own model, we have to make sure that dynamic shapes exists, right? Maybe it's useful to add a comment specifying that.

lucylq: ah, looks like it's provided by the builder, not on a per-model basis: https://github.com/pytorch/executorch/blob/main/extension/llm/export/builder.py#L186

examples/models/llama/config/llm_config.py
lucylq: Sure, sounds good.

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+@dataclass
+class ExportConfig:
+    """
+    Configures properties relevant to the export process.
+    """
+
+    max_seq_length: int = 128
+    max_context_length: int = 128
```
larryliu0820, line 215: This seems worth a `__post_init__` check? I don't think max_context_length can be larger than max_seq_length

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,312 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
larryliu0820, line 1: We should start to write tests for this. We can make sure the `__post_init__()` hook is able to error out when inputs are invalid.

larryliu0820 (approved):

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+    STATIC_LLAMA = "static_llama"
+    QWEN2_5 = "qwen2_5"
+    QWEN3_0_6B = "qwen3-0_6b"
+    QWEN3_1_7B = "qwen3-1_7b"
+    QWEN3_4B = "qwen3-4b"
+    PHI_4_MINI = "phi_4_mini"
+    SMOLLM2 = "smollm2"
+
+
+class PreqMode(str, Enum):
```
kimishpatel, line 43: worth tagging as deprecated

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+    """
+    Configurations specific to the model, e.g. whether it’s Qwen3 or Phi-4-mini,
+    and are the minimal set of parameters needed to load the pretrained
+    eager model and its weights.
+
+    Attributes:
+        model_class: Which model to to export.
+        params: Model parameters, such as n_layers, hidden_size, etc.
+            If left empty will use defaults specified in model_args.py.
+        checkpoint: Path to the checkpoint file.
```
kimishpatel, line 66: can this be hf path as well?

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+            If left empty, the model will be initialized with random weights.
+        checkpoint_dir: Path to directory containing sharded checkpoint files.
+        tokenizer_path: Path to the tokenizer file.
+        metadata: Json string containing metadata information.
+            e.g. '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}'
+        use_lora: Rank of the LoRA, if set to 0 then this means no LoRA. For use with QAT.
```
kimishpatel, line 72: lora in our case is really tied to QAT model that we released, right? It is not independently applicable to any model? If so I think we want to tied this to QAT checkpoints specifically for llama3_2

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+        fairseq2: For legacy internal use cases, this is safe to ignore.
```
kimishpatel, line 73: prefix with _ to differentiate between this and other supported params

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+        preq_mode: Legacy option to specify how prequantized weights are loaded.
+            Going forward, ExecuTorch supports loading weights prequantized through
+            TorchAo as-is, without any special handling.
+        preq_group_size: Legacy option to specify the group size of prequantized weights.
```
kimishpatel, line 77: I thik we probably want to couple these together. If you are loading pre-quantized checkpoint that group size cannot be set independently, right? So maybe having a separate dataclass that captures all the params and maps it by name is better. ALthough I presume you have to keep this for BC?

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+
+    FP32 = "fp32"
+    FP16 = "fp16"
+    BF16 = "bf16"
+
+
+@dataclass
+class ModelConfig:
+    """
+    Configurations not necessarily specific to the model, but are needed to
```
kimishpatel, line 115: Then why call it modelconfig?

kimishpatel: Maybe call LoweringConfig? or ExportConfig although not all options are export related

examples/models/llama/config/llm_config.py
jackzhxng: Is that the case? If so I'll add something to the __post_init__ to verify and update the comment

examples/models/llama/config/llm_config.py
jackzhxng: Yeah, just keeping for BC. I think ideally we are moving away from preq to tho right?

examples/models/llama/config/llm_config.py
jackzhxng: This is more like: what other modifications do I want to make to the model in eager that _aren't_ specific to the model itself (e.g. NOT checkpoint, model architecture, tokenizer) and can be shared across different models.

examples/models/llama/config/llm_config.py
jackzhxng: No, but at the moment if you specify a non-llama `model_class` and don't specify checkpoint, it will download from HF. Worth adding this comment

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+            token generation.
+        use_shared_embeddings: whether the embedding/output weights should be
+            shared. Only available with torchao kernels, e.g. when
+            qmode set to use a "torchao:8da(\\d+)w" pattern.
+        use_sdpa_with_kv_cache: Whether to use flash attention by substituting
+            for our custom SDPA op. Note that the naming is poor and this
+            doesn't actually have anything to do with the kv_cache at the moment.
+        expand_rope_table: Temporary workaround to expand sin/cos table in head
+            dim to take vectorized path in optimized kernels.
+        use_attention_sink: Whether to use attention sink to support multi-round
```
kimishpatel, line 133: I would also considering pruning options that we dont intend to support. attention sink at the moment does not have performant implementation and I would rather hide it somewhere than to expose it. Reduce maintenance burden

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+class ExportConfig:
+    """
+    Configures properties relevant to the export process.
+
+    Attributes:
+        max_seq_length: Maximum length of sequence to evaluate.
+        max_context_length: Maximum of context for the model to remember.
+        output_dir: Output dir to save the exported .pte file to.
+        output_name: File name to override the exported .pte file.
+        so_library: Shared library to specify custom quantized operators.
```
kimishpatel, line 209: I dont think we need so_library anymore. Remove it in follow up

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+        export_only: Whether to stop right after torch.export() and
```
kimishpatel, line 210: was this for debug?

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+    """
+    Type of backend-specific Pt2e quantization strategy to use.
+
+    Pt2e uses a different quantization library that is graph-based
+    compared to `qmode`, which is also specified in the QuantizationConfig
+    and is source transform-based.
+    """
+
+    XNNPACK_DYNAMIC = "xnnpack_dynamic"
+    XNNPACK_DYNAMIC_QC4 = "xnnpack_dynamic_qc4"
```
kimishpatel, line 271: we probably can remove these two

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+    Configures how the model should be quantized (PTQ).
+
+    Attributes:
+        qmode: Quantization mode using TorchAo, expressed as a string.
+            See the __post_init__ validation for available qmode options.
+        embedding_quantize: Type of embedding quantization.
+            Must be of the format '<bitwidth>,<groupsize>', e.g., '8,1024'.
+        pt2e_quantize: Quantization mode using pt2e, which is an alternative
+            to TorchAo that uses backend-aware graph mode quantization rather
+            than source transformation quantization.
```
kimishpatel, line 300: I would want to hide these details from users

examples/models/llama/config/llm_config.py
```diff
@@ -0,0 +1,495 @@
...
+        extended_ops: Whether to match more types of ops to delegates to XNNPack.
+    """
+
+    enabled: bool = False
+    extended_ops: bool = False
+
+
+class CoreMLQuantize(str, Enum):
+    B4W = "b4w"
+    C4W = "c4w"
```
kimishpatel, line 379: maybe favor this pattern over the ones in pt2e_quantizer

examples/models/llama/config/test_llm_config.py
```diff
@@ -0,0 +1,104 @@
...
+        qcfg = QuantizationConfig(qmode="int8")
+
+        with self.assertRaises(ValueError):
+            LlmConfig(model=model_cfg, quantization=qcfg)
+
+
+class TestValidConstruction(unittest.TestCase):
+
+    def test_valid_llm_config(self):
+        LlmConfig(
```
kimishpatel, line 73: are these configs constructible from json? I think that would e the best

PR #11029, opened 2025-05-21
Introduce hydra framework with backwards compatibility

### LINKED ISSUE #9449: [etLLM] New config system to export_llama

jackzhxng:
* #11162
* #11081
* __->__ #11029
* #11028

Differential Revision: [D75263989](https://our.internmc.facebook.com/intern/diff/D75263989)


jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263989).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263989).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263989).

jackzhxng:
@jackzhxng has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D75263989).

examples/models/llama/TARGETS
```diff
@@ -82,6 +82,8 @@ runtime.python_binary(
     ],
     deps = [
         ":export_library",
+        ":export_llama_args",
+        ":export_llama_hydra",
```
larryliu0820: Instead of having 2 new libraries, why not just add the source files into `export_library`? Is there a case where we want to use these new libraries separately?

examples/models/llama/config/llm_config_utils.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+
+from executorch.examples.models.llama.config.llm_config import LlmConfig
+
+
+def convert_args_to_llm_config(args: argparse.Namespace) -> LlmConfig:
```
larryliu0820: I think this should be a static method of `LlmConfig`. For example `LlmConfig.from_args()`

examples/models/llama/config/targets.bzl
```diff
@@ -13,3 +13,19 @@ def define_common_targets():
             "@EXECUTORCH_CLIENTS",
         ],
     )
+
+    runtime.python_library(
+        name = "llm_config_utils",
```
larryliu0820: Again I think this should be merged into `llm_config`.

examples/models/llama/export_llama_args.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
larryliu0820, line 1: Why do we have to have `export_llama_args.py` and `export_llama_hydra.py`? Can we merge them into `export_llama.py`?

jackzhxng: So the reason is due to how Hydra works, it needs to decorate the main() function. So this is the only way to still optionally use the old CLI and the new Hydra CLI

larryliu0820 (approved):

examples/models/llama/export_llama.py
```diff
@@ -4,30 +4,50 @@
...
-    parser = build_args_parser()
-    args = parser.parse_args()
-    export_llama(args)
+
+    use_hydra, remaining_args = parse_hydra_arg()
+    if use_hydra:
+        # The import runs the main function of export_llama_hydra with the remaining args
+        # under the Hydra framework.
+        sys.argv = [arg for arg in sys.argv if arg != "--hydra"]
+        print(f"running with {sys.argv}")
```
kimishpatel, line 40: did you mean to print this?

examples/models/llama/export_llama.py
```diff
@@ -4,30 +4,50 @@
...
+        runpy.run_module(
```
kimishpatel, line 41: not just use importlib?

kimishpatel: oh nevermind. you are running this module not importing it

examples/models/llama/export_llama_hydra.py
```diff
@@ -0,0 +1,27 @@
...
+
+from executorch.examples.models.llama.config.llm_config import LlmConfig
+from executorch.examples.models.llama.export_llama_lib import export_llama
+from hydra.core.config_store import ConfigStore
+
+cs = ConfigStore.instance()
+cs.store(name="llm_config", node=LlmConfig)
+
+
+@hydra.main(version_base=None, config_name="llm_config")
```
kimishpatel, line 21: i dont think i know enough about hydra but is there a test that showcases the use?

PR #11031, opened 2025-05-21
serialize scales as bf16 and serialize in Named Data Map

mcr229:
XNNPACK Currently uses BF16 scales for running GEMMS with groupwise quantized weights. Currently we serialize scales as FP32, and then convert them to BF16 before passing to XNNPACK. We can save both memory and file size by serializing the scales as BF16 first. 

As an additional step here, we move the serialization of scales both for channelwise and groupwise quantized weights into the named data map. In the future, if we want to swap data that could be a potential feature because scales are no longer tied to the XNNPACK payload but can be swappable through the ptd file.

cc @lucylq for the scale serialization

### Llama Experiments
```
-rw-r--r--  1 maxren  staff  1746392320 May 20 16:49 llama3_fp32_scales.pte
-rw-r--r--  1 maxren  staff  1707798912 May 20 18:47 llama3_bf16_scales.pte
```

we see ~40 mb reduction in model size.

### DIFF
--- a/backends/xnnpack/operators/node_visitor.py
+++ b/backends/xnnpack/operators/node_visitor.py
@@ -274,19 +274,46 @@ class NodeVisitor:
 
         return dtype
 
-    def get_quant_params(self, quant_params: QuantParams) -> XNNQuantParams:
+    def get_quant_params(
+        self, quant_params: QuantParams, xnn_graph: XNNGraph
+    ) -> XNNQuantParams:
         if quant_params.per_channel:
             scale = cast(torch.Tensor, quant_params.scale)
+            buffer_idx = len(xnn_graph.constant_data)
+            num_scales = scale.numel()
+
+            if quant_params.is_per_channel_group:
+                scale = scale.to(torch.bfloat16)
+
+            num_bytes = scale.untyped_storage().nbytes()
+            scale_array = ctypes.cast(
+                scale.untyped_storage().data_ptr(),
+                ctypes.POINTER(ctypes.c_char * num_bytes),
+            ).contents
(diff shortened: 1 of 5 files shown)


GregoryComer (approved):

backends/xnnpack/operators/node_visitor.py
```diff
@@ -274,19 +274,46 @@ def get_per_channel_dtype(
...
+    def get_quant_params(
+        self, quant_params: QuantParams, xnn_graph: XNNGraph
+    ) -> XNNQuantParams:
         if quant_params.per_channel:
             scale = cast(torch.Tensor, quant_params.scale)
+            buffer_idx = len(xnn_graph.constant_data)
+            num_scales = scale.numel()
+
+            if quant_params.is_per_channel_group:
+                scale = scale.to(torch.bfloat16)
```
digantdesai, line 286: I think this should be a flag (default=bf16), which lines up with QB4W xnnpack flag to use bf16. We can error out at AoT for fp32 given we can't run that yet.

backends/xnnpack/serialization/runtime_schema.fbs
```diff
@@ -63,7 +65,9 @@ table PerChannelGroupQuant {
   scale:[float];
   channel_dim:int;
   group_size:int;
-  scale_bf16:[ushort];
+  scale_bf16:[ushort] (deprecated);
```
digantdesai, line 68: curious, why mark this as deprecated but not float if we are moving to ndm for evreythig?

mcr229: because this actually was never used, since we never added the export path to serialize into this field. scale[float] is still used by older versions

PR #11035, opened 2025-05-21
Qualcomm AI Engine Direct - GA PVT

shewu-quic:
Summary:
- Add PVT example script
- Add the test for PVT


cccclai (approved):
Wow does it run out of box?

PR #11038, opened 2025-05-21
Arm backend: Add constant ops unittests.

AdrianLundell:
Adds unittests for eye, ones and zeros.

These ops are supported by the ComputeConstantOpsAOT pass, add unittests for coverage. This exposes an edge case bug similar as for the clone, where a partition with one single constant op results in an empty network.


zingo:
Unrelated fails
pull / unittest-editable / macos / macos-job (pull_request)

zingo (approved):

PR #11039, opened 2025-05-21
GA models benchmark script

haowhsu-quic:
### Test plan
```bash
# build xnnpack
cd executorch
export ANDROID_NDK_ROOT=path/to/ndk
./build_xnnpack.sh
```
```bash
# benchmark example for xnnpack
cd optimum
optimum-cli export executorch --model "microsoft/swin-tiny-patch4-window7-224" --task "image-classification" --recipe "xnnpack" --output_dir optimum/xnn_artifacts/swin
cd executorch
python benchmark.py -b xnn -p path/to/optimum/xnn_artifacts/swin/model.pte -s $DEVICE --iteration 150
```
```bash
# benchmark example for qnn
cd executorch
python benchmark.py -b qnn -p path/to/swin_qnn_q8.pte -s $DEVICE --iteration 400
```

Statistics will be something like:
```
load: 272.587000
1st: 4.648000
avg: 3.960253
peak_mem: 88766
avg_mem: 84006.000
```

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -1,0 +1,1257 @@
+import argparse
+import numpy as np
+import os
+import random
+import subprocess
+import torch
+
+from executorch.examples.models.llama.evaluate.eager_eval import EagerEvalWrapper
+
+
+### GLOBALS
+qnn_sdk = os.getenv("QNN_SDK_ROOT")
+workspace = "/data/local/tmp/et_ga_benchmark"
+memory_script_file = "peak_memory.sh"
+perf_file = "statistics.txt"
+seed = 1126
+###
+
+
+def image_classification_eval(
+    backend,
+    soc_model,
+    device,
+    host,
+    pte_path,
+    module,
+    inputs,
+    targets,
+    artifact_dir,
+):
+    from executorch.examples.qualcomm.utils import (
+        make_output_dir,
+        SimpleADB,
+        topk_accuracy,
+    )
+    import numpy as np
+    from pathlib import Path
+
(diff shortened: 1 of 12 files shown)


benchmark.py
```diff
@@ -0,0 +1,211 @@
+import argparse
```
cccclai, line 1: This is helpful. Should we add this script to the devtools as a cli tool so others can also use it? And expand to other backend like vulkan as well.

cccclai:
> How are we generating pte file for QNN. It will be good to have that recipe and ideally if it is available on optimum-cli that would be great

Those are listed in https://github.com/pytorch/executorch/tree/main/examples/qualcomm/oss_scripts

build_xnnpack.sh
```diff
@@ -0,0 +1,40 @@
...
+  -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
+  -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
+  -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
+  -DEXECUTORCH_BUILD_XNNPACK=ON \
+  -DEXECUTORCH_ENABLE_LOGGING=ON \
+  -DPYTHON_EXECUTABLE=python \
+  -B$BUILD_FOLDER .
+
+cmake --build $BUILD_FOLDER -j9 --target install --config $BUILD_TYPE
+
```
digantdesai: This should generate extecutorch libraries, now you need to build a runner which uses this.

PR #11044, opened 2025-05-21
Add LoRA linear definition

lucylq:
* __->__ #11044

^
Add lora linear definition. Pull out linears from attention, and allow custom linear (eg. lora linear) to be passed in.

If none, construct linear (current behaviour).

Differential Revision: [D73953776](https://our.internmc.facebook.com/intern/diff/D73953776/)


examples/models/llama/llama_transformer.py
```diff
@@ -255,7 +256,67 @@ def construct_transformer(model_args: ModelArgs) -> Transformer:
...
+                rank=model_args.r,
+                alpha=model_args.lora_alpha,
+                dropout=0.0,
+                use_bias=model_args.attention_qkv_bias,
+            )
+            if model_args.target_modules is not None
+            and "output_proj" in model_args.target_modules
+            else None
+        )
+        if model_args.attention_type == "static":
```
jackzhxng: Nit, should do the wq, wk, wv, wo construction inside the else

jackzhxng (approved):

examples/models/llama/llama_transformer.py
```diff
@@ -255,7 +256,67 @@ def construct_transformer(model_args: ModelArgs) -> Transformer:
...
+            # Static attention constructs ModuleLists for qkvo and
+            # populates them with MHA attention layers; do not pass in
+            # wq, wk, wv, wo.
+            attention = cls(model_args, layer_id, rope)
```
JacobSzwejbka: Why are you doing this in llama_transformer at all? You have model args passed to the constructor couldnt you just construct these tensors in the constructor rather then passing them to init?

Not having to add attention specific logic to llama_transformer.py is the intention of the attention registry, so if we have to break that abstraction I think we need to elaborate why

JacobSzwejbka (changes requested):

examples/models/llama/llama_transformer.py
lucylq: The intention is to make building a transformer composable, e.g., you can pass in any type of attention, with any type of linears.

If construct them inside the transformer constructor, then the transformer has to understand what all the different options are. i.e., we didn't want to put LoRA logic inside MHA attention itself, as it isn't tied to MHA specifically.

JacobSzwejbka: Attention already has to understand what all the different options are. Thats why model args is passed into it no?

JacobSzwejbka: > Attention already has to understand what all the different options are. Thats why model args is passed into it no?

As in its already possible to generate invalid configs where someone sticks model options in and then uses an attention that doesnt support those options.

lucylq: LoRA specifically doesn't have to be coupled with the attention implementation (eg. MHA). I can see why it makes sense to do it inside if other optimizations are also applied within eg. MHA though ..

larryliu0820: @JacobSzwejbka my original concern on this is that model config might be non-standardized. Now that we have Jack's new config system maybe this is less of a concern.

JacobSzwejbka: This set up is already dependent on model config though. The lora constructors use tons of args from the config

JacobSzwejbka (approved):

Issue #9580, opened 2025-03-25
Export lora llama3b with adapter weights


lucylq:
### 🚀 The feature, motivation and pitch

For program-data separation lora demo

cbilgin:
ETA: Roughly goaling for 2 weeks.

lucylq:
https://github.com/pytorch/executorch/pull/11044

PR #11045, opened 2025-05-21
Export a lora model

lucylq:
* #12724
* __->__ #11045
* #11044

^

Program+data combined currently, using the lora linear definition.
Issue: https://github.com/pytorch/executorch/issues/9580

Differential Revision: [D75153377](https://our.internmc.facebook.com/intern/diff/D75153377/)


examples/models/llama/export_llama_lib.py
```diff
@@ -235,6 +235,18 @@ def build_args_parser() -> argparse.ArgumentParser:
         help="checkpoint directory. Use with a sharded checkpoint, not for the standard llama2 model. Note, checkpoint_dir takes precedence over checkpoint if both are set.",
     )
 
+    parser.add_argument(
+        "--adapter_checkpoint",
```
digantdesai, line 239: Nit
```suggestion
        "--lora_adapter_checkpoint",
```

examples/models/llama/model.py
```diff
@@ -130,6 +133,21 @@ def __init__(self, **kwargs):
             with open(params_path, "r") as f:
                 params = json.loads(f.read())
 
+        # Get adapter checkpoint and config.
+        adapter_checkpoint = {}
+        adapter_config = {}
+        adapter_checkpoint_path = kwargs.get("adapter_checkpoint", None)
+        if adapter_checkpoint_path:
```
digantdesai, line 140: nit: use os.exists or similar from Path?

digantdesai (commented):
tests?

lucylq:
> tests?

Let me add one. Maybe with phi-3-mini or a smaller model.

jackzhxng (approved):

extension/llm/export/config/llm_config.py
```diff
@@ -90,6 +96,8 @@ class BaseConfig:
     params: Optional[str] = None
     checkpoint: Optional[str] = None
     checkpoint_dir: Optional[str] = None
+    adapter_checkpoint: Optional[str] = None
```
JacobSzwejbka, line 99: can you name these the same as their corresponding foundational weight fields? I guess thats params and checkpoint? So call it lora_params and lora_checkpoint?

extension/llm/export/config/llm_config.py
```diff
@@ -73,10 +73,16 @@ class BaseConfig:
...
+        adapter_checkpoint: Path to the adapter.pt file from torchtune. Used if
+            the model has trained LoRA adapters. Must provide
+            adapter_config.json.
+        adapter_config: Path to the adapter_config.json file from torchtune.
+            Used if the model has trained LoRA adapters. Must provide adapter.pt.
         tokenizer_path: Path to the tokenizer file.
         metadata: Json string containing metadata information.
             e.g. '"{\"get_bos_id\":128000, \"get_eos_ids\":[128009, 128001]}"'
-        use_lora: Rank of the LoRA, if set to 0 then this means no LoRA. For use with QAT.
+        use_lora: Only for use with QAT. Rank of the LoRA adapter, disabled
```
JacobSzwejbka, line 84: You didnt add this, but why is this boolean named field an int, and why does it correspond with qat?

extension/llm/export/config/llm_config.py
lucylq: These come directly from the adapter_config.json output by torchtune, see example here: https://huggingface.co/lucylq/llama3_1B_lora/blob/main/adapter_config.json

I guess we'd have to do some sort of patching to make sure they match up to use lora-prefixed names instead

extension/llm/export/config/llm_config.py
lucylq: cc @cccclai ?

JacobSzwejbka (approved):

extension/llm/export/config/llm_config.py
cccclai: Didn't add it myself either, I think it's likely from Lunwen - I believe it's for the llama 3.2 1b QAT checkpoint which include LoRA, so make sure we don't break llama3.2 QAT model if we use this flag somewhere else

cccclai: https://github.com/pytorch/executorch/pull/5823

jackzhxng (approved):

PR #11058, opened 2025-05-21
Add ExecuTorchRuntime.java skeleton

kirklandsign:
Summary: No API yet. Just factor out NativeLoader.loadLibrary from Module.java and LlmModule.java, to allow mock.

Differential Revision: D75180081


shoumikhin (approved):

GregoryComer:
@kirklandsign Some Meta apps started seeing a high load failure rate. My hypothesis is that the load might actually be getting skipped since the ExecuTorchRuntime instance may be getting optimized out of the Module call, since it's not actually used. I'm re-adding the so load in Module.java in https://github.com/pytorch/executorch/pull/11385 to try to resolve this for now, but I could use your help to find the long term solution once you're back. Thanks!

PR #11073, opened 2025-05-22
Use one-less cores in preset workflows

jathu:
### Summary
must go faster

image

### Test plan
CI

cc @larryliu0820


kirklandsign (approved):

larryliu0820 (approved):
Change 900 to 90?

PR #11097, opened 2025-05-23
Qualcomm AI Engine Direct - GA FocalNet

winskuo-quic:
### Summary
- GA Model: FocalNet
- Properly annotate `AdaptiveAvgPool1D`

#### Accuracy
top1: ~60%
top5: ~85%

#### Speed
SM8750: 2.2ms/inf

#### Script
python examples/qualcomm/oss_scripts/focalnet.py -b build-android -H $HOST -s $DEVICE -m $MODEL --dataset ../imagenet-mini/val/

### Test plan
E2E UT

### DIFF
--- a/backends/qualcomm/scripts/build.sh
+++ b/backends/qualcomm/scripts/build.sh
@@ -30,7 +30,7 @@ CMAKE_X86_64="build-x86"
 BUILD_AARCH64="true"
 CMAKE_AARCH64="build-android"
 CLEAN="true"
-BUILD_TYPE="Debug"
+BUILD_TYPE="RelWithDebInfo"
 BUILD_JOB_NUMBER="16"
 
 if [ -z PYTHON_EXECUTABLE ]; then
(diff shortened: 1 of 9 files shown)


backends/qualcomm/scripts/build.sh
```diff
@@ -30,7 +30,7 @@ CMAKE_X86_64="build-x86"
 BUILD_AARCH64="true"
 CMAKE_AARCH64="build-android"
 CLEAN="true"
-BUILD_TYPE="Debug"
+BUILD_TYPE="RelWithDebInfo"
```
cccclai, line 33: Is there any specific reason for this change?

cccclai (approved):
Thanks

backends/qualcomm/scripts/build.sh
winskuo-quic: The reason behind is because of this PR: https://github.com/pytorch/executorch/pull/10918.
If we run `build.sh` with `Debug` build, it will get the following error: ` error: unable to find library -lflatccrt`.

PR #11105, opened 2025-05-23
Add event_tracer to MIL for Backend Init Context

cmt0:
Summary: Allow for intended usage of event tracer in BackendInitContext.

Differential Revision: D75308778


cccclai (approved):
Good catch

PR #11107, opened 2025-05-23
Fix triu/tril CoreML lowering error in to_edge_transform_and_lower

metascroy:
This PR fixes a CoreML lowering issue with triu/tril when using to_edge_transform_and_lower.

It also changes the warning logging in _remove_invalid_ops_for_not_decompose  to log only once per call, rather than once per op.


cccclai (approved):
Looks reasonable

PR #11118, opened 2025-05-23
support channels last inputs in xnnpack

leafs1:
### Summary

Support channels last inputs in xnnpack

### Test plan

Wrote tests taking combinations of channels last/first inputs with ops requiring inputs of both forms.


backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -333,7 +344,10 @@ def input_to_nchw(
             # do anything else here
             self.mark_as_nchw_node(input_node)
 
-        if self.is_nchw_node(input_node):
+        if input_node.op == "placeholder":
```
mcr229, line 347: do we need to check that this is a placeholder? Can we just skip if the input is already contiguous?

backends/xnnpack/runtime/XNNExecutor.cpp
```diff
@@ -106,20 +106,25 @@ ET_NODISCARD Error XNNExecutor::prepare_args(EValue** args) {
...
           InvalidArgument,
           "XNNPACK backend accepts tensors with at most %d dims, but got %zu",
           XNN_MAX_TENSOR_DIMS,
           num_dims);
-      for (int d = 0; d < num_dims; ++d) {
-        dims[d] = tensor->size(d);
+
+      bool is_channels_last =
+          executorch::runtime::is_channels_last_dim_order(dim_order, num_dims);
+      if (is_channels_last) {
```
mcr229, line 119: can we just always convert? meaning something like:
```
for(int i = 0; i < dim_order.size(); ++i){
    dims[i] = dim_order[i];
}
```
if dim_order is 0, 1, 2, 3 then this is normal, otherwise channels last dim order is 0, 2, 3, 1

backends/xnnpack/runtime/XNNExecutor.cpp
```diff
@@ -220,8 +225,25 @@ ET_NODISCARD Error XNNExecutor::resize_outputs(EValue** args) const {
...
-      expected_output_size[d] = static_cast<SizesType>(dims[d]);
+    executorch::aten::DimOrderType dim_order[kTensorDimensionLimit];
+    Error errr =
+        ET_RUNTIME_NAMESPACE::get_dim_order(*out_tensor, dim_order, num_dim);
+    ET_CHECK_OR_RETURN_ERROR(
+        errr == Error::Ok,
+        Internal,
+        "Failed to retrieve dim order from tensor!");
+
+    bool is_channels_last =
```
mcr229, line 236: same as above.

backends/xnnpack/test/passes/test_channels_last_tagged_reshape.py
```diff
@@ -58,6 +58,88 @@ def test_fp32_channels_last_tagged_reshape_pass(self):
...
+            y = self.linear1(x)
+            return self.conv1(y)
+
+    def test_conv_linear_dim_order_swaps_on_nhwc_input(self):
+        tester = Tester(
+            self.LinearConv().eval(),
+            (torch.randn(1, 3, 6, 4).to(memory_format=torch.channels_last),),
+        )
+
+        tester.export().to_edge_transform_and_lower().to_executorch().serialize().run_method_and_compare_outputs()
```
mcr229, line 77: nice can we refactor this a little bit? since every test basically has the same stuff just different module and different inputs

mcr229 (changes requested):

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -269,7 +277,10 @@ def input_to_nhwc(
             # serializing graph, but don't do anything else here
             self.mark_as_nhwc_node(input_node)
 
-        if self.is_nhwc_node(input_node):
+        if input_node.name == "x":
```
mcr229, line 280: i don't think this is right?

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -333,7 +344,10 @@ def input_to_nchw(
             # do anything else here
             self.mark_as_nchw_node(input_node)
 
-        if self.is_nchw_node(input_node):
+        if input_node.name == "x":
```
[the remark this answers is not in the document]
leafs1: The idea is that since the input nodes arent tagged, we need to check the metadata to see if they're continuous or not.

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
GregoryComer: @leafs1 Just want to ping on this one. In general, we can't assume the name of inputs. Can we leverage other graph metadata to determine when to do the conversion? IIRC this logic was around the inputs to the partition and was originally checking to see if it was a placeholder. It might have gotten changed due to some other feedback, but I think that might be the best way to handle it?

leafs1: I agree, I switched back the logic to check for placeholder ops. Since we know that placeholder ops are inputs and that inputs don't get tagged we need to rely on the metadata.

mcr229 (approved):

PR #11119, opened 2025-05-23
Android preset

kirklandsign:


scripts/build_android_library.sh
```diff
@@ -36,28 +36,13 @@ build_android_native_library() {
 
   cmake . -DCMAKE_INSTALL_PREFIX="${CMAKE_OUT}" \
     -DCMAKE_TOOLCHAIN_FILE="${ANDROID_NDK}/build/cmake/android.toolchain.cmake" \
-    -DANDROID_ABI="${ANDROID_ABI}" \
+    --preset "android-${ANDROID_ABI}" \
     -DANDROID_PLATFORM=android-26 \
```
larryliu0820, line 40: Can you add this to preset as well?

CMakePresets.json
```diff
@@ -6,6 +6,36 @@
       "hidden": true,
       "binaryDir": "${sourceDir}/cmake-out"
     },
+    {
+      "name": "android-arm64-v8a",
+      "displayName": "Build everything buildable on android arm64-v8a",
```
larryliu0820: This doesn't sound accurate right? Could be "Build executorch core and JNI bindings on android arm64-v8a"?

larryliu0820:
Can you add a CI job for android preset? Also cc @jathu if you are interested.

scripts/build_android_library.sh
kirklandsign: We need to build twice, one for x86_64, one for arm64-v8a.

CMakePresets.json
kirklandsign: Good point!

kirklandsign:
> Can you add a CI job for android preset? Also cc @jathu if you are interested.

Not for our preset test series. We are using a separate sh to build it right now and always using this preset anyway so we don't really want the preset specific test.

scripts/build_android_library.sh
larryliu0820: I think you can set one as default and override another one

larryliu0820 (approved):

scripts/build_android_library.sh
```diff
@@ -36,28 +36,13 @@ build_android_native_library() {
...
-    -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
-    -DEXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL=ON \
-    -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
     -DEXECUTORCH_BUILD_EXTENSION_TRAINING=ON \
-    -DEXECUTORCH_BUILD_KERNELS_OPTIMIZED=ON \
-    -DEXECUTORCH_BUILD_KERNELS_QUANTIZED=ON \
-    -DEXECUTORCH_BUILD_KERNELS_LLM="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
     -DEXECUTORCH_BUILD_LLAMA_JNI="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
     -DEXECUTORCH_BUILD_NEURON="${EXECUTORCH_BUILD_NEURON}" \
     -DNEURON_BUFFER_ALLOCATOR_LIB="${NEURON_BUFFER_ALLOCATOR_LIB}" \
```
jathu: Can't we remove all the options? i.e. `EXECUTORCH_BUILD_DEVTOOLS` is still there

PR #11124, opened 2025-05-24
Enable all platform tests in CMake

GregoryComer:
### Summary
Enable platform_death_test and clock_test in CMake. They were previously disabled. They seem to run fine locally and will be nice to run in OSS with the upcoming PAL changes.

### Test plan
I built and ran the tests locally without issue. I'm also relying on CI to validate that the tests are healthy.


mergennachin:
What's up the CI failures? Looks legit

PR #11131, opened 2025-05-26
Add new dependency library for vulkan tests

fffrog:
**Background**:

We removed the inline flag of the symbol `torch::jit::GetBackendMetaSerialization()` from and moved it to .cpp in order to make the dynamic registration related to PyTorch serialization work properly.

However, after testing, it was found that Vulkan's tests depend on the symbol named `torch::jit::GetBackendMetaSerialization()`, so after the above changes, it is necessary to explicitly link libtorch_cpu (which contains the torch::jit::GetBackendMetaSerialization() symbol)

For more information, see [PR](https://github.com/pytorch/pytorch/pull/147095), especially this [comment](https://github.com/pytorch/pytorch/pull/147095#issuecomment-2788206845)


larryliu0820:
@SS-JIA can you take a look?

SS-JIA:
IIRC this should have been resolved via a separate solution landed by @ezyang in Pytorch. Closing.

PR #11133, opened 2025-05-26
Arm backend: Use _pytree tree_flatten in model evaluator

SaoirseARM:
### Summary
Updates model evaluator to use tree_flatten for output tensors when checking int8 deviation. 
* Removes in script function which flattens the tensors.


zingo (approved):

zingo:
unittest fails are unrelated and fails on main also

PR #11141, opened 2025-05-27
Arm backend: Fix decompose_meandim_pass bug

AdrianLundell:
Previously this pass could insert avg_pool ops with configurations not supported for Ethos-U55. Check this using the AvgPool2dSupported checker and decompose fully using sum for non supported cases.

Also modifies the test cases to test this, +more varied shapes.


zingo (approved):

zingo:
non Arm unitests fails are unrelated and fails on main also

PR #11143, opened 2025-05-27
Arm backend: Add support for 5D tensors

oscarandersson8218:
Updates memory-format handling to consider 5D tensors. Adds a guard to not partition nodes with ranks > 5. This is because the current memory format pass does not handle such ranks in a good way.


zingo (approved):

zingo:
unittest fails are unrelated (and on main also)

PR #11149, opened 2025-05-27
Arm backend: Update minimal example notebook

AdrianLundell:


zingo:
Nxp and llama unit test fails are unrelated and happens on main also

zingo (approved):

PR #11156, opened 2025-05-27
Fix returning too early when setting options

jathu:
### Summary
Fixes a bug where we return too early since we are using a macro. This is only apparent if `set_overridable_option` is used outside of a top-level CMake file (i.e. different scope)

### Test plan
unittest + CI

### DIFF
--- a/tools/cmake/common/preset.cmake
+++ b/tools/cmake/common/preset.cmake
@@ -89,11 +89,9 @@ endmacro()
 # Set an overridable option.
 macro(set_overridable_option NAME VALUE)
   # If the user has explitily set the option, do not override it.
-  if(DEFINED ${NAME})
-    return()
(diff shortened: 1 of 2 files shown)


tools/cmake/common/preset.cmake
```diff
@@ -89,11 +89,9 @@ endmacro()
 # Set an overridable option.
 macro(set_overridable_option NAME VALUE)
   # If the user has explitily set the option, do not override it.
-  if(DEFINED ${NAME})
-    return()
```
larryliu0820, line 93: `return` not exiting the macro?

jathu: Return was exiting the scope that `set_overridable_option` was used in since it's a macro. We now just avoid that overall

larryliu0820 (approved):

PR #11179, opened 2025-05-28
adjust platform compatibility for qnn backend

h-friederich:
Summary: The Qualcomm dependencies are only available for Android, so let's mark the target accordingly.

Differential Revision: D75139951


cccclai (approved):
Internal only change

PR #11196, opened 2025-05-28
Rename Apple simulator libs to maintain backwards compatibility

jathu:
### Summary

Given people hardcode these lib values, let's just maintain backwards compatibility when generating the frameworks.

### Test plan

Built the framework and checked the simulator suffix.


shoumikhin:
Looks like a workaround? Can we address the root cause instead?

shoumikhin (approved):

PR #11204, opened 2025-05-28
Add vectorized_math.h

swolchok:
Set of math functions that work on both scalars and at::vec::Vectorized, to be used in #9432.


swolchok:
* (to be filled)

kernels/portable/cpu/util/test/vectorized_math_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include <cstdint>
+
+#ifndef ET_USE_PYTORCH_HEADERS
+#error "This test requires ET_USE_PYTORCH_HEADERS!"
+#endif // ET_USE_PYTORCH_HEADERS
+
+TEST(VectorizedMathTest, BasicUnary) {
+  __at_align__ float result_floats[at::vec::Vectorized<float>::size()] = {0};
+  const auto x_vec = at::vec::Vectorized<float>::arange(0, 1);
```
manuelcandales, line 23: why are all these tests using `arange(0, 1)`, i.e. a single number instead of an actual vector of numbers?

swolchok: it's not a single number; the arguments are start and step, not start and end. https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cpu/vec/vec_base.h#L255

manuelcandales (approved):

PR #9584, opened 2025-03-25
Add SupportedTensorDtypes::BOOL

swolchok:

PR #11206, opened 2025-05-28
Add mixed integer precision test for op_mul

swolchok:
This tests a possibly-surprising result: int8(100) * int8(100) with output type of long is 16 in ATen, even though the output type can hold 10000.


swolchok:
* #9386
* __->__ #11206
* #9584

kimishpatel:
wait what? I dont get why is it 16?

manuelcandales (approved):

manuelcandales:
> wait what? I dont get why is it 16?

because 100 * 100 mod 256 = 16

this is how ATen works, because of functionalization
`mul.out(x, y, out=out)` must be equivalent to `z = mul(x, y); out = z.to(out.dtype)`
if `x` & `y` are `int8`, then `z` is also `int8`, and so multiplication turns into multiplication modulo 256
this is the downside of functionalization, it is sort of a compromise

PR #9527, opened 2025-03-23
Added support for bias in optimized linear operation

DavitGrigoryan132:
### Summary

#8234 

This PR is adding a functionality to use bias in optimized linear operation

PR #11210, opened 2025-05-29
Add support for bias in optimized op_linear.cpp.

hsharma35:
Summary: Diff uses `op_add_sub_impl` to add bias after optimized gemm call.

Differential Revision: D75491158

### DIFF
--- a/kernels/optimized/cpu/op_linear.cpp
+++ b/kernels/optimized/cpu/op_linear.cpp
@@ -6,17 +6,69 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <array>
+
+#include <c10/util/irange.h>
+
 #include <executorch/kernels/optimized/blas/CPUBlas.h>
+#include <executorch/kernels/optimized/vec/functional_base.h>
+#include <executorch/kernels/optimized/vec/vec_base.h>
 #include <executorch/kernels/portable/cpu/util/matmul_ops_util.h>
 #include <executorch/runtime/kernel/kernel_includes.h>
-
-#include <array>
 
 namespace torch {
 namespace executor {
 namespace native {
 
-using Tensor = executorch::aten::Tensor;
+namespace {
+using ::executorch::aten::Tensor;
+using ::executorch::cpublas::gemm;
+using ::executorch::cpublas::TransposeType;
+using ::executorch::runtime::toString;
+using ::executorch::vec::map;
+using ::executorch::vec::Vectorized;
+
+// Use vector store to initialize with scalar bias.
+template <typename scalar_t>
+void initialize_scalar(
+    const ssize_t out_numel,
+    const scalar_t init,
+    scalar_t* out) {
+  using Vec = Vectorized<scalar_t>;
+
+  // Initialize a vector with the scalar initial value.
+  Vec init_vec(init);
+
+  ssize_t d = 0;
+  for (; d < out_numel - (out_numel % Vec::size()); d += Vec::size()) {
+    // Vector-length store.
+    init_vec.store(out + d);
+  }
+  if (out_numel - d > 0) {
+    // Sub-vector-length store.
+    init_vec.store(out + d, static_cast<int>(out_numel - d));
+  }
+}
+
+// Use std::memcpy to initialize with vector bias.
+template <typename scalar_t>
+void initialize_to_vector(
+    const ssize_t n,
+    const ssize_t m,
+    const scalar_t* bias,
+    scalar_t* out) {
+  // Output is a n x m x scalar_t, while bias is m x scalar_t.
+  const size_t row_size = static_cast<size_t>(m) * sizeof(scalar_t);
+  for (const auto col : c10::irange(n)) {
+    std::memcpy(
+        // Point to Column `col` of the output tensor.
+        out + col * m,
+        bias,
+        row_size);
+  }
+}
+
+} // namespace
 
 Tensor& opt_linear_out(
     RuntimeContext& ctx,
@@ -24,12 +76,6 @@ Tensor& opt_linear_out(
     const Tensor& mat2,
     const optional<Tensor>& bias,
     Tensor& out) {
-  ET_KERNEL_CHECK_MSG(
-      ctx,
-      !bias.has_value(),
-      InvalidArgument,
-      out,
-      "bias not supported yet in linear");
   ET_KERNEL_CHECK(ctx, check_linear_args(in, mat2, out), InvalidArgument, out);
 
   size_t output_ndim = 0;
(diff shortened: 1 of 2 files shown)


kernels/optimized/cpu/op_linear.cpp
```diff
@@ -72,7 +70,13 @@ Tensor& opt_linear_out(
             m);
       });
 
-  return out;
+  if (!bias.has_value()) {
+    return out;
+  }
+
+  // Add bias.
+  return opt_add_sub_out_impl<false, op_name>(
```
swolchok, line 78: Hmm. I don't have a great immediate intuition for whether this is better than the approach in #9527. Can you talk about (and perhaps add a comment explaining) why we shouldn't just rely on the accumulation in the GEMV/GEMM kernel? Possibly it just doesn't matter. Again CC @kimishpatel

shim_et/xplat/executorch/kernels/optimized/op_registration_util.bzl
```diff
@@ -213,7 +213,11 @@ OPTIMIZED_ATEN_OPS = (
     op_target(
         name = "op_linear",
         deps = [
+            ":binary_ops",
+            ":add_sub_impl",
```
swolchok, line 217: sort

zonglinpeng (approved):

kernels/optimized/cpu/op_linear.cpp
```diff
@@ -50,8 +47,9 @@ Tensor& opt_linear_out(
   for (int ii = 0; ii < in.dim() - 1; ++ii) {
     flattened_input_dim *= in.sizes()[ii];
   }
+  static constexpr const char op_name[] = "linear.out";
   ET_SWITCH_REAL_TYPES_AND2(
-      Half, BFloat16, in.scalar_type(), ctx, "mm.out", CTYPE, [&]() {
```
swolchok, line 54: Hmm, this is tricky. linear isn't a core ATen op so I  *think* we aren't supposed to do this; instead I *think* if there is a bias it's addmm.out and if not it's mm.out. This matters for selective build configurations and depends on how XNNPACK does it, so CC @kimishpatel / @digantdesai / @mcr229  ?

kernels/optimized/cpu/op_linear.cpp
hsharma35: Ah I didn't dig deeper into alternatives. This approach seemed to need the least code change. The approach in #9527 is likely faster. 
Can we push this PR in now and then let #9527 merge whenever possible? I just want bias support to unblock meta internal stuff.

kernels/optimized/cpu/op_linear.cpp
kimishpatel: @larryliu0820 can you say something about how this impacts selective build? Are we relying on mm.out to apply selective dtype rule for all instances of op_mm?
@swolchok are you also saying that dtype selective build looks at op_mm in the edge ir graph and then apply selective build rule that match "mm.out" in the op name string?

kernels/optimized/cpu/op_linear.cpp
kimishpatel: Yeah I think #9527 is likely faster. If it is faster then why not just land that @hsharma35 ?

kernels/optimized/cpu/op_linear.cpp
larryliu0820: I think this is a right fix. For linear op we should use "linear.out" here. Selective build will have the following impact with this change: before this change, if a user select a dtype (say float16) for `mm.out`, it will apply to linear.out *by mistake*, like a side effect. After this change, any binary that relies this side effect will have a small binary size regression, because the key changes from "mm.out" to "linear.out". Though I would argue that this is still a right fix and we should fix whatever binary size regression by properly selecting "linear.out".

larryliu0820 (approved):

kernels/optimized/cpu/op_linear.cpp
kimishpatel: Thanks @larryliu0820 that was my understanding too. But @hsharma35 can we please use the other diff as the right fix. I dont think we should really land this as is if we have a better fix available. If this is urgent for you, lets prioritze the other one

kimishpatel (changes requested):
I want to block this for now in favor of landing the other fix for bias. mm.out -> linear.out is the right change though

kernels/test/op_linear_test.cpp
```diff
@@ -88,6 +96,70 @@ TEST_F(OpLinearOutTest, AllDtypesSupported) {
...
+  // Initialize input tensors.
+  constexpr int kReduceDim = 4;
+  constexpr int kDimX = 3, kDimY = 5;
+  constexpr int kValueX = 1;
+  constexpr int kValueY = 2;
+  constexpr int kValueBias = 4;
+  Tensor x = tf.full({kDimX, kReduceDim}, kValueX);
+  Tensor y = tf.full({kDimY, kReduceDim}, kValueY);
+  // Same size as output.
+  Tensor b = tf.full({kDimX, kDimY}, kValueBias);
```
kimishpatel, line 153: I dont think your implementation supports this. In fact for linear bias cannot be 2d. Can you update test so taht expected values not the same for all elements. In that case you likely would have caught the issue. This test should otherwise actually fail

kimishpatel (changes requested):
ok so this diff does incorporate "idea" from the other diff so seems ok to me. but would like the test case to be updated

kernels/test/op_linear_test.cpp
hsharma35: Good catch! I added this test to check if the aten implementation works. Surprisingly, the aten implementation can accept a 2d bias.
This works too:
```
x = torch.rand(3, 4)
w = torch.rand(5, 4)
b = torch.rand(3, 5)
out = torch.ops.aten.linear.default(x, w, b)
```
I've removed the 2D test. If we want to add support for 2D bias, it should be in a follow-up PR.

kernels/optimized/cpu/op_linear.cpp
```diff
@@ -6,30 +6,76 @@
...
+template <typename scalar_t>
+void initialize_to_vector(
+    const ssize_t n,
+    const ssize_t m,
+    const scalar_t* bias,
+    scalar_t* out) {
+  // Output is a n x m x scalar_t, while bias is m x scalar_t.
+  const size_t row_size = static_cast<size_t>(m) * sizeof(scalar_t);
+  for (const auto col : c10::irange(n)) {
+    std::memcpy(
```
kimishpatel, line 63: To handle 2d bias, you need to fix this. Bias pointer is not advancing

kimishpatel (approved):
I am accepting this but do note that you want either some random values or "not-all-outputs-are-same" values. The test as is, is quite weak IMO.

kernels/optimized/cpu/op_linear.cpp
hsharma35: hmm, it's more complicated than that. bias can be 1xm, nx1, or nxm.

PR #11217, opened 2025-05-29
check for overflow in calculate_nbytes

JacobSzwejbka:
Summary: Got a fuzzer error around overflow here

Differential Revision: D75544079

### DIFF
--- a/runtime/executor/method_meta.cpp
+++ b/runtime/executor/method_meta.cpp
@@ -55,12 +55,24 @@ Result<Tag> get_tag(
 size_t calculate_nbytes(
     Span<const int32_t> sizes,
     executorch::aten::ScalarType scalar_type) {
-  ssize_t n = 1;
+  size_t n = 1;
+  size_t prev_n = 1;
   for (size_t i = 0; i < sizes.size(); i++) {
+    prev_n = n;
     n *= sizes[i];
+    // Check for overflow
+    ET_CHECK(sizes[i] == 0 || n / sizes[i] == prev_n);
   }
-  // Use the full namespace to disambiguate from c10::elementSize.
-  return n * executorch::runtime::elementSize(scalar_type);
+
+  size_t elem_size = executorch::runtime::elementSize(scalar_type);
+
+  prev_n = n;
+  n = n * elem_size;
+
+  // Check for overflow
+  ET_CHECK(elem_size == 0 || n / elem_size == prev_n);
+
+  return n;
 }
 
 } // namespace
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -371,7 +371,7 @@ jobs:
         size=${arr[4]}
         # threshold=48120 on devserver with gcc11.4
         # todo(lfq): update once binary size is below 50kb.
-        threshold="51408"
+        threshold="55504"
         if [[ "$size" -le "$threshold" ]]; then
           echo "Success $size <= $threshold"
         else
@@ -406,7 +406,7 @@ jobs:
         output=$(ls -la cmake-out/test/size_test)
         arr=($output)
         size=${arr[4]}
-        threshold="47560"
+        threshold="51656"
         if [[ "$size" -le "$threshold" ]]; then
           echo "Success $size <= $threshold"
         else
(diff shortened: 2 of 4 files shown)


runtime/executor/method_meta.cpp
```diff
@@ -55,12 +55,28 @@ Result<Tag> get_tag(
...
     executorch::aten::ScalarType scalar_type) {
-  ssize_t n = 1;
+  size_t n = 1;
+  size_t prev_n = 1;
   for (size_t i = 0; i < sizes.size(); i++) {
+    prev_n = n;
     n *= sizes[i];
+    // Check for overflow
+    ET_CHECK_MSG(
+      sizes[i] == 0 || n / sizes[i] == prev_n,
```
lucylq, line 65: nit: instead of storing a var for prev_n, wdyt about:
```
ET_CHECK_MSG(sizes[i] == 0 || ((n*sizes[i])/sizes[i]) == n))
```

runtime/executor/method_meta.cpp
```diff
@@ -55,12 +55,28 @@ Result<Tag> get_tag(
...
   }
-  // Use the full namespace to disambiguate from c10::elementSize.
-  return n * executorch::runtime::elementSize(scalar_type);
+
+  size_t elem_size = executorch::runtime::elementSize(scalar_type);
+
+  prev_n = n;
+  n = n * elem_size;
+
+  // Check for overflow before multiplying by element size
```
lucylq, line 74: Update the comment, or do it before line 72?

Could do the same thing, 
```
ET_CHECK_MSG(elem_size == 0 || (((n*elem_size)/elem_size) == n);

n = n * elem_size;
```

lucylq (approved):
Thanks for adding this! Could you also make the same changes to tensor_layout?

https://github.com/pytorch/executorch/blob/14c3b310ee873dbe9f8decf37bf32096b4abcf81/runtime/core/tensor_layout.cpp#L19

runtime/executor/method_meta.cpp
```diff
@@ -55,12 +55,24 @@ Result<Tag> get_tag(
...
     Span<const int32_t> sizes,
     executorch::aten::ScalarType scalar_type) {
-  ssize_t n = 1;
+  size_t n = 1;
+  size_t prev_n = 1;
   for (size_t i = 0; i < sizes.size(); i++) {
+    prev_n = n;
     n *= sizes[i];
+    // Check for overflow
+    ET_CHECK(sizes[i] == 0 || n / sizes[i] == prev_n);
```
swolchok, line 64: is it necessary to perform a division here? could we check `n >= prev_n` instead?

runtime/executor/method_meta.cpp
```diff
@@ -55,12 +55,24 @@ Result<Tag> get_tag(
...
-  // Use the full namespace to disambiguate from c10::elementSize.
-  return n * executorch::runtime::elementSize(scalar_type);
+
+  size_t elem_size = executorch::runtime::elementSize(scalar_type);
+
+  prev_n = n;
+  n = n * elem_size;
+
+  // Check for overflow
+  ET_CHECK(elem_size == 0 || n / elem_size == prev_n);
```
swolchok, line 73: ditto

runtime/executor/method_meta.cpp
JacobSzwejbka: ah. yes...

JacobSzwejbka: Ah ok no it doesnt work. I can give a counter example. 

To get the numbers smaller lets say Im accumulating in uint16_t and Im taking the product of int8_t::max

n: 127 prev_n: 1

n: 16129 prev_n: 127

n: 16767 prev_n: 16129 (overflowed at 65535 from logical value 2048383)

swolchok: hmm. I'm tempted to ask if we should just import [c10/util/safe_numerics.h](https://github.com/pytorch/pytorch/blob/main/c10/util/safe_numerics.h) (which, note to self, requires llvmMathExtra.h) instead of doing division then. it is annoying that it only has 64-bit safe_multiplies functions, but we could add a size_t-based one (in executorch for now if you don't want to wait for pin bump, followed by adding it in pytorch, getting a pin bump, and deleting the executorch one).

JacobSzwejbka: https://github.com/pytorch/executorch/issues/11370

PR #11231, opened 2025-05-29
Expose gen_oplist_lib to clients

Vysarat:
Summary: I'm working on a linter that checks a model's ops against those available in our runtime environment (see D75110565), and I'd like to call `gen_oplist` programmatically. I can't do that because `gen_oplist_lib` is not visible outside the `executorch` subdir. If there are no objections, I'd like to expose it to clients.

Differential Revision: D75630096

### DIFF
--- a/codegen/tools/targets.bzl
+++ b/codegen/tools/targets.bzl
@@ -14,6 +14,7 @@ def define_common_targets(is_fbcode = False):
         base_module = "executorch.codegen.tools",
         visibility = [
             "//executorch/...",
+            "@EXECUTORCH_CLIENTS",
         ],
         deps = [
             "//executorch/codegen:gen_lib",


codegen/tools/targets.bzl
```diff
@@ -14,6 +14,7 @@ def define_common_targets(is_fbcode = False):
         base_module = "executorch.codegen.tools",
         visibility = [
             "//executorch/...",
+            "@EXECUTORCH_CLIENTS",
```
JacobSzwejbka, line 17: Im fine with this but you want a yaml? Would it be better if you refactored to take a dep on an in memory represenation?

@larryliu0820  Is this representation stable for people outside ET to take a dep?

Vysarat: I agree that skipping the yaml would make things cleaner. I think that can be fixed in a follow-up PR.

larryliu0820: Agreed let's do the hard work of exposing an API that returns the object in memory. Instead of a follow-up, can you do it first and it becomes clear on what library/target to expose publicly?

JacobSzwejbka: Also just in general it seems like all you'd need are the dtype dim order combinations for every op

PR #11238, opened 2025-05-29
Add MPS runner test

jathu:
### Summary
TSIA

### Test plan

CI

```
$ .ci/scripts/test_model.sh mv3 cmake mps
```

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -219,7 +219,7 @@ test_model_with_coreml() {
   DTYPE=float16
 
   "${PYTHON_EXECUTABLE}" -m examples.apple.coreml.scripts.export --model_name="${MODEL_NAME}" --compute_precision "${DTYPE}" --use_partitioner
-  EXPORTED_MODEL=$(find "." -type f -name "${MODEL_NAME}*.pte" -print -quit)
+  EXPORTED_MODEL=$(find "." -type f -name "${MODEL_NAME}*coreml*.pte" -print -quit)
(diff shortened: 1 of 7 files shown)


.ci/scripts/test_model.sh
```diff
@@ -219,7 +219,7 @@ test_model_with_coreml() {
   DTYPE=float16
 
   "${PYTHON_EXECUTABLE}" -m examples.apple.coreml.scripts.export --model_name="${MODEL_NAME}" --compute_precision "${DTYPE}" --use_partitioner
-  EXPORTED_MODEL=$(find "." -type f -name "${MODEL_NAME}*.pte" -print -quit)
+  EXPORTED_MODEL=$(find "." -type f -name "${MODEL_NAME}*coreml*.pte" -print -quit)
```
manuelcandales, line 222: why is this coreml?

jathu: Note that this is under the `test_model_with_coreml` method — so it makes sense to hone the search to coreml models

mergennachin:
> There are no reasons not to support MPS out-of-the box. I will update docs in a follow-up PR.

It's not as stable compared to coreml. Let's not do this for now.

jathu:
@mergennachin good point, I didn't realize MPS wasn't P0. I've repurposed the diff to include MPS runner tests, so we can still use that in the future.

shoumikhin (approved):

larryliu0820 (approved):

Issue #10819, opened 2025-05-12
Consolidate executor_runners


kirklandsign:
### 🚀 The feature, motivation and pitch

Now we have so many executor runner targets, one for each backend. We should use a main one, and the link backends

metascroy:
I think this is a good idea.  I added CoreML support for the executor runner in March, and it looks like ScottW added XNNPACK support there as well in March: https://github.com/pytorch/executorch/blob/main/CMakeLists.txt#L799-L846

cccclai:
What if different backends have different ways plan input/output memory?

kirklandsign:
Is it possible that we make executor_runner generic?

VarunSai02:
Hi, I’d like to work on this issue. I noticed that the current main branch now has the executorch_backends interface target and the generic executor_runner links against it, so part of the proposed consolidation appears to already be implemented. Is the remaining scope primarily migrating/removing the backend-specific runners, or are there particular backends that still need to be integrated? I’d be happy to take one of them.

kirklandsign:
Some backends like QNN might have specific features. 

Most other backends are generic.

PR #11230, opened 2025-05-29
Used cmake-format on CMakeLists.txt

BujSet:
### Summary
`Fixes #<10736>`  : https://github.com/pytorch/executorch/issues/10736#event-17545114889

### Changes I Made

`cmake-format -i CMakeLists.txt`

`cmake-lint CMakeLists.txt`

The linting reported an [E1122] Duplicate keyword argument error for variable DESTINATION. I deleted the earlier setting of this variable, as the comment seemed to imply that `lib` should *not* be the appropriate destination.

### Test plan
Successfully ran the commands in the [building from source](https://github.com/BujSet/executorch/blob/main/docs/source/using-executorch-building-from-source.md) guide. I did try running the test commands from the [Testing](https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#testing), however, these reported error on a clean install. No new errors were reported when I made these changes.

PR #11239, opened 2025-05-29
Consolidating executor runners in CMake - xnnpack

BujSet:
### Summary
Continuing the process of consolidating the executor runners to top level CMakeLists.txt file. Completing this process for the xnnpack backend in the cmake build flow. 

`Partially Fixes #<10819>` : https://github.com/pytorch/executorch/issues/10819.

### Test plan
Successfully ran the commands in the [building from source](https://github.com/BujSet/executorch/blob/main/docs/source/using-executorch-building-from-source.md) guide. Updated the github actions to not use the generic `executor_runner` and remove the invocation for the backend-specific `xnn_executor_runner`. Also ran the following formatinng and lint checks:

```cmake-format -i backends/xnnpack/CMakeLists.txt```

```cmake-lint backends/xnnpack/CMakeLists.txt```

Includes updates to documentation and tutorials (flows from examples and tutorials were verify to ensure build functionality).

Note: the buck flow side of things was unmodified, and support for that flow remains for a future change.

### CODE REVIEW

File: backends/xnnpack/CMakeLists.txt [hunk 1]
```diff
@@ -153,7 +132,6 @@ endif()
 
 install(
   TARGETS xnnpack_backend
-  DESTINATION lib
```
larryliu0820, line 156: Why change this?

.ci/scripts/test_model.sh
```diff
@@ -150,10 +137,10 @@ test_model_with_xnnpack() {
   if [[ "${BUILD_TOOL}" == "buck2" ]]; then
     buck2 run //examples/xnnpack:xnn_executor_runner -- --model_path "${OUTPUT_MODEL_PATH}"
   elif [[ "${BUILD_TOOL}" == "cmake" ]]; then
-    if [[ ! -f ${CMAKE_OUTPUT_DIR}/backends/xnnpack/xnn_executor_runner ]]; then
-      build_cmake_xnn_executor_runner
+    if [[ ! -f ${CMAKE_OUTPUT_DIR}/executor_runner ]]; then
+      build_cmake_executor_runner
```
larryliu0820, line 141: You will have to make sure `EXECUTORCH_BUILD_XNNPACK` is ON, right? Maybe let `build_cmake_executor_runner` take a backend string and turn the option on accordingly? This can be extended to other backends.

.ci/scripts/test_model.sh
```diff
@@ -150,10 +137,10 @@ test_model_with_xnnpack() {
   if [[ "${BUILD_TOOL}" == "buck2" ]]; then
     buck2 run //examples/xnnpack:xnn_executor_runner -- --model_path "${OUTPUT_MODEL_PATH}"
```
larryliu0820, line 138: Also after this PR the internal (BUCK) behavior is different than CMake. I think it's fine for now but maybe add a TODO here?

backends/xnnpack/CMakeLists.txt
BujSet: I ran `cmake-lint` on this file, and it reported `[E1122] Duplicate keyword argument DESTINATION` because this variable is overridden two lines below to what I believe is the correct destination. It's the same error and fix as in the [formatting PR-L496](https://github.com/pytorch/executorch/pull/11230/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL469).

.ci/scripts/test_model.sh
BujSet: Added in a comment for it.

larryliu0820 (approved):

PR #11244, opened 2025-05-30
Save the representative intputs into the ETRecord object

Juntian777:
Summary: The diff saves representative inputs from a BundledProgram object passed into generate_etrecord, specifically targeting the "forward" method. It iterates through the method_test_suites of the bundled_program, and if it finds a suite with the method_name "forward", it returns the inputs from the first test case.

Differential Revision: D75637400


Gasoonjia (approved):
LGTM!

PR #11245, opened 2025-05-30
Remove outdated editable mode warning from building from source docs

### LINKED ISSUE #11229: Remove the outdated warning in building from source documentation
### 📚 The doc issue

Here https://docs.pytorch.org/executorch/main/using-executorch-building-from-source.html

We have a warning section 

Image

This warning is not necessary since editable mode now works with code like:

```bash
python -c "from executorch.exir import CaptureConfig"
```

We can completely remove this section.

### Suggest a potential alternative/fix

_No response_

knobby-tires:
The warning about modules not being importable in editable mode is no longer necessary since this issue has been resolved.

Fixes #11229

### Summary
Removes the outdated warning section from the building from source documentation as requested in the issue. The warning is no longer needed because editable mode now works correctly with imports like `from executorch.exir import CaptureConfig`.

### Test plan
This is a documentation-only change.

### DIFF
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -91,15 +91,6 @@ Or alternatively, [install conda on your machine](https://conda.io/projects/cond
 
    If C++ files are being modified, you will still have to reinstall ExecuTorch from source.
 
-> **_WARNING:_**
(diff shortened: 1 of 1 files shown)


mergennachin (approved):
Thank you!

PR #11249, opened 2025-05-30
Arm backend: Avoid subtraction error on boolean tensors in error diff logging

fumchin:
This patch add a dtype check before diff calculation to avoid runtime errors since boolean tensors do not support arithmetic operations like subtraction. 
Reports the mismatch count for boolean tensors.

for example: 
```
############### ERROR DIFFERENCE ###############
BATCH 0
(BOOLEAN tensor) 0 / 8 elements differ (0.00%)
BATCH 1
(BOOLEAN tensor) 8 / 8 elements differ (100.00%)
BATCH 2
(BOOLEAN tensor) 8 / 8 elements differ (100.00%)
################################################
...
...
AssertionError: Output 0 (bool tensor) does not match reference output.
```


zingo:
Hi @digantdesai  this PR touched file outside Arm backend, so we would appreciate a review if OK

zingo:
Hi @mergennachin  this PR touched file outside Arm backend, so we would appreciate a review if OK

mergennachin (approved):

PR #11260, opened 2025-05-30
Don't depend on `aten_headers_for_executorch` when running with HTP simulator since it already gets this dep from xplat

sammarden-meta:
Differential Revision: D75488427

This prevents linker errors when building with `-c user.enable_fllm_on_htp_sim=true`.


swolchok (changes requested):
hopefully https://github.com/pytorch/pytorch/pull/155858 should supersede this

PR #11266, opened 2025-05-30
Removed minimum target macOS 11.0 branching

NeelakshSharma:
### Summary
Removed minimum target macOS 11.0 branching.

Fixes #11259 [Issue](https://github.com/pytorch/executorch/issues/11259)


jathu:
Thanks for the fix @NeelakshSharma! Just kicked off some tests, if everything is green we can land it

shoumikhin (approved):

PR #11275, opened 2025-06-02
Qualcomm AI Engine Direct - Fix error of avg pool2d

chunit-quic:
- Fix the ceil_mode if condition of avg_pool2d


cccclai (approved):
Thanks for the fix

PR #11278, opened 2025-06-02
Arm backend: Remove special host build of flatcc

zingo:
### Summary
This was a workaround as host version of flattcc was not generated from the build as it should and you got the target version. Later when flattcc was needed on the host it was the wrong version and you got an error.
Now this has been solved in the project and flattcc is only build for the host so there is no need to force a special build like this anymore.

### Test plan
Tested in local build and on GitHub


zingo:
Unrelated fail 

mergennachin:
@jathu looks good?

jathu (approved):
awesome 👏

PR #11280, opened 2025-06-02
Arm backend: Make pre-push non-interactive on one argument

perheld:
Calling pre-push as a git hook actually supplies it with 2 arguments, remote-name and remote-url. So the logic of automatically going non-interactive when there are any arguments is flawed.

This commit makes sure pre-push only goes non-interactive when there is only one argument.

Change-Id: I21cc05b24521bb075c2f74fe4ed7f773b797b085


zingo:
Unrelated fail 

zingo (approved):

PR #11281, opened 2025-06-02
Arm backend: Reduce atol in vector_norm test

oscarandersson8218:
pipeline.change_args("run_method_and_compare_outputs",...) was used to change qtol, but as no inputs were passed in the call run_method_and_compare_outputs produced new inputs with a different distribution (randn), than what was used for calibration (rand).


zingo:
Fail unrelated
[trunk / test-llama-torchao-lowbit / macos-job (push)]

zingo (approved):

PR #11282, opened 2025-06-02
Arm backend: aot_arm_compiler.py: Create output dir

martinlsm:
Create output directory for the generated BundledProgram if it does not already exist. This will save the user time by not having to rerun the program due to it complaining that the directory does not exist.


zingo:
Unrelated fail 

zingo (approved):

PR #11285, opened 2025-06-02
[RFC] Fetch gflags dynamically

jathu:
### Summary
I recently learned about [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html). This seems like a nicer solution than using submodules. Thoughts?

### Test plan

CI


larryliu0820 (approved):

third-party/CMakeLists.txt
```diff
@@ -14,6 +14,13 @@ if(BUILD_TESTING)
   add_subdirectory(googletest)
 endif()
 
+FetchContent_gflags()
+# Create a symlink from the gflags source directory to third-party/gflags
+message(STATUS "gflags source dir: ${gflags_SOURCE_DIR}")
```
jathu: nit: lets not log, it's too noisy — do we really need to tell them every time?

larryliu0820: I think this is very helpful for debugging

mergennachin:
Why add such as complication? What are we optimizing for? What are the trade-offs?

gflags doesn't seem very large (if you're optimizing for disk space)

mergennachin:
See some previous context here: https://github.com/pytorch/executorch/pull/7305

jathu:
Discussed with @larryliu0820 offline and we decided to punt on this until our buck workflow is removed from cmake.

PR #11288, opened 2025-06-02
[1/N] Add backend option

cccclai:
Summary: Introduce backend option as discussed in https://github.com/pytorch/executorch/discussions/10216

Differential Revision: D75770142

Step 1: Introducd Backend Option class

In later stage, it will be plugged in with the rest of the stack.

### DIFF
--- a/runtime/backend/backend_option.h
+++ b/runtime/backend/backend_option.h
@@ -1,0 +1,169 @@
+#include <executorch/runtime/core/error.h>
+#include <cstddef>
+#include <cstring>
+
+namespace executorch {
+namespace ET_RUNTIME_NAMESPACE {
+
+// Strongly-typed option key template
+// Wraps a string key with type information for type-safe option access
+template <typename T>
+struct OptionKey {
+  const char* key; // String identifier for the option
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Supported option data types
+enum class OptionType { BOOL, INT, STRING };
+
+// Union-like container for option values (only one member is valid per option)
+struct OptionValue {
+  bool bool_value; // Storage for boolean values
+  int int_value; // Storage for integer values
+  const char* string_value; // Storage for string values
+};
+
+// Represents a single backend configuration option
+struct BackendOption {
+  const char* key; // Name of the option
+  OptionType type; // Data type of the option
+  OptionValue value; // Current value of the option
+};
+
+// Fixed-capacity container for backend options with type-safe access
+// MaxCapacity: Maximum number of options this container can hold
+template <size_t MaxCapacity>
+class BackendOptions {
+ public:
+  // Initialize with zero options
+  BackendOptions() : size(0) {}
+
+  // Type-safe setters ---------------------------------------------------
+
+  /// Sets or updates a boolean option
+  /// @param key: Typed option key
+  /// @param value: Boolean value to set
+  void set_option(OptionKey<bool> key, bool value) {
+    set_option_internal(key.key, OptionType::BOOL, {.bool_value = value});
+  }
+
+  /// Sets or updates an integer option
+  /// @param key: Typed option key
+  /// @param value: Integer value to set
+  void set_option(OptionKey<int> key, int value) {
+    set_option_internal(key.key, OptionType::INT, {.int_value = value});
+  }
+
+  /// Sets or updates a string option
+  /// @param key: Typed option key
+  /// @param value: Null-terminated string value to set
+  void set_option(OptionKey<const char*> key, const char* value) {
+    set_option_internal(key.key, OptionType::STRING, {.string_value = value});
+  }
+
+  // Type-safe getters ---------------------------------------------------
+
+  /// Retrieves a boolean option value
+  /// @param key: Typed option key
+  /// @param out_value: Reference to store retrieved value
+  /// @return: Error code (Ok on success)
+  executorch::runtime::Error get_option(OptionKey<bool> key, bool& out_value)
+      const {
+    OptionValue val{};
+    executorch::runtime::Error err =
+        get_option_internal(key.key, OptionType::BOOL, val);
+    if (err == executorch::runtime::Error::Ok) {
+      out_value = val.bool_value;
+    }
+    return err;
+  }
+
+  /// Retrieves an integer option value
+  /// @param key: Typed option key
+  /// @param out_value: Reference to store retrieved value
+  /// @return: Error code (Ok on success)
+  executorch::runtime::Error get_option(OptionKey<int> key, int& out_value)
+      const {
+    OptionValue val{};
+    executorch::runtime::Error err =
+        get_option_internal(key.key, OptionType::INT, val);
+    if (err == executorch::runtime::Error::Ok) {
+      out_value = val.int_value;
+    }
+    return err;
+  }
+
+  /// Retrieves a string option value
+  /// @param key: Typed option key
+  /// @param out_value: Reference to store retrieved pointer
+  /// @return: Error code (Ok on success)
+  executorch::runtime::Error get_option(
+      OptionKey<const char*> key,
+      const char*& out_value) const {
+    OptionValue val{};
+    executorch::runtime::Error err =
+        get_option_internal(key.key, OptionType::STRING, val);
+    if (err == executorch::runtime::Error::Ok) {
+      out_value = val.string_value;
+    }
+    return err;
+  }
+
+ private:
+  BackendOption options[MaxCapacity]{}; // Storage for options
+  size_t size; // Current number of options
+
+  // Internal helper to set/update an option
+  void
+  set_option_internal(const char* key, OptionType type, OptionValue value) {
+    // Update existing key if found
+    for (size_t i = 0; i < size; ++i) {
+      if (strcmp(options[i].key, key) == 0) {
+        options[i].type = type;
+        options[i].value = value;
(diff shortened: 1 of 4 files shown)


JacobSzwejbka:
Will review tomorrow

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Supported option data types
+enum class OptionType { BOOL, INT, STRING };
+
+// Union-like container for option values (only one member is valid per option)
+struct OptionValue {
+  bool bool_value; // Storage for boolean values
+  int int_value; // Storage for integer values
```
JacobSzwejbka, line 22: int64_t

mergennachin:
I don't think we need to define a generic BackendOption that takes arbitrary key value. Each backend owners could define their options instead. Vulkan will define its own config structure, Core ML will define its own config structure. Wdyt? 

@JacobSzwejbka, @mcr229, @kimishpatel 

Something like this:

```
BackendInterface { // parent interface
  virtual Error setOption(const BackendOptions& options) = 0;
}

struct BackendOptions {  // Empty parent struct
  virtual ~BackendOptions() = default;
};
```

Each individual backends define their own configs

```
struct VulkanBackendOptions : public BackendOptions {
  const char* metadata[20] = "";
  int thread_count = 1;
  // ... other Vulkan-specific options
};

```

```
VulkanBackend extends BackendInterface {
  Error setOption(const BackendOptions& options) override {
    auto* vulkan_opts = dynamic_cast<const VulkanBackendOptions*>(&options);
  }
}
```

// User code

```
VulkanBackendOptions opts;
opts.thread_count = 5;
auto backend = std::make_unique<VulkanBackend>();
backend->setOption(opts);  // 
```

cccclai:
> I don't think we need to define a generic BackendOption that takes arbitrary key value. Each backend owners could define their options instead. Vulkan will define its own config structure, Core ML will define its own config structure. Wdyt?
> 
> @JacobSzwejbka, @mcr229, @kimishpatel
> 
> Something like this:
> 
> ```
> BackendInterface { // parent interface
>   virtual Error setOption(const BackendOptions& options) = 0;
> }
> 
> struct BackendOptions {  // Empty parent struct
>   virtual ~BackendOptions() = default;
> };
> ```
> 
> Each individual backends define their own configs
> 
> ```
> struct VulkanBackendOptions : public BackendOptions {
>   const char* metadata[20] = "";
>   int thread_count = 1;
>   // ... other Vulkan-specific options
> };
> ```
> 
> ```
> VulkanBackend extends BackendInterface {
>   Error setOption(const BackendOptions& options) override {
>     auto* vulkan_opts = dynamic_cast<const VulkanBackendOptions*>(&options);
>   }
> }
> ```
> 
> // User code
> 
> ```
> VulkanBackendOptions opts;
> opts.thread_count = 5;
> auto backend = std::make_unique<VulkanBackend>();
> backend->setOption(opts);  // 
> ```

I thought about this idea at the beginning, but it means that users code needs to link to a specific backend, otherwise the runner code will fail to build. Currently the same user code can be backend agnostic, meaning the same code can still run with different backends with/without linking the backend. As the example, the portable runner can work with/without xnnpack https://github.com/pytorch/executorch/blob/main/examples/xnnpack/executor_runner/targets.bzl. 

It also exposes the backend specific symbol to users, and the contract interface is a bit looser. Previously, The contract is like backend <-> ET <-> backend, and if we change it to this, it can be user <-> backend. And users can pass arbitrary things to backend. Users can run backend->init, backend->execute in their own code too, and it will be out of our control.

With the proposed solution (https://github.com/pytorch/executorch/discussions/10216), the user code can still be backend agnostic, and the backend option is still optional. We are still on the hook of the contract.

cccclai:
Btw I was thinking using ghstack such that I can export a stack of PRs...but it means this PR will be closed and a new PR will be created. I will copy the comments over just so we keep the history.

Edit: the new pr is here https://github.com/pytorch/executorch/pull/11389, I'm closing this PR and please leave comments in the new PR

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+struct OptionKey {
+  const char* key; // String identifier for the option
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Supported option data types
+enum class OptionType { BOOL, INT, STRING };
+
+// Union-like container for option values (only one member is valid per option)
+struct OptionValue {
```
digantdesai, line 20: Why not use a union? Or structure with a template?

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
+#include <executorch/runtime/core/error.h>
+#include <cstddef>
+#include <cstring>
+
+namespace executorch {
+namespace ET_RUNTIME_NAMESPACE {
```
kimishpatel, line 6: why this capitalized?

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+// Strongly-typed option key template
+// Wraps a string key with type information for type-safe option access
+template <typename T>
+struct OptionKey {
+  const char* key; // String identifier for the option
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Supported option data types
+enum class OptionType { BOOL, INT, STRING };
```
kimishpatel, line 17: Just use std::varaint

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+// Union-like container for option values (only one member is valid per option)
+struct OptionValue {
+  bool bool_value; // Storage for boolean values
+  int int_value; // Storage for integer values
+  const char* string_value; // Storage for string values
+};
+
+// Represents a single backend configuration option
+struct BackendOption {
+  const char* key; // Name of the option
```
kimishpatel, line 28: you can also do
```
struct BackendOptions {
  const char* key;
  std::variant<int, bool, char *> value;
}

File: runtime/backend/backend_option.h [hunk 6]
```diff
@@ -0,0 +1,169 @@
...
+ public:
+  // Initialize with zero options
+  BackendOptions() : size(0) {}
+
+  // Type-safe setters ---------------------------------------------------
+
+  /// Sets or updates a boolean option
+  /// @param key: Typed option key
+  /// @param value: Boolean value to set
+  void set_option(OptionKey<bool> key, bool value) {
```
kimishpatel, line 46: Just make this `set_option<T>(BackendOption<T> key_value)...`

If tomorrow you expand BackendOption to have other types like float or int8 you just have to update allowed types in BackendOpton.

runtime/backend/backend_option.h
kimishpatel: or std::variant that can hold only bool, int and char *?

runtime/backend/backend_option.h
kimishpatel: and assert to on allowable types

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+        get_option_internal(key.key, OptionType::STRING, val);
+    if (err == executorch::runtime::Error::Ok) {
+      out_value = val.string_value;
+    }
+    return err;
+  }
+
+ private:
+  BackendOption options[MaxCapacity]{}; // Storage for options
+  size_t size; // Current number of options
```
kimishpatel, line 114: size_ for private variables.

runtime/backend/backend_option.h
```diff
@@ -0,0 +1,169 @@
...
+    if (size < MaxCapacity) {
+      options[size++] = {key, type, value};
+    }
+  }
+
+  // Internal helper to get an option value with type checking
+  executorch::runtime::Error get_option_internal(
+      const char* key,
+      OptionType expected_type,
+      OptionValue& out) const {
```
kimishpatel, line 137: Is it necessary that type is specified? If we use std::variant we can just make this `get_optional_internal(const char* key, BackendOption& option);` User can be responsible for type check. Not sure why we need to provide it

kimishpatel (commented):
Left a few comments

runtime/backend/backend_option.h
cccclai: It will introduce std library I assume?

runtime/backend/backend_option.h
cccclai: union can be a good option, let me try

runtime/backend/backend_option.h
cccclai: It's used across the whole ExecuTorch stack, but I find now it's not needed so switch to `executorch::runtime` namespace

runtime/backend/backend_option.h
cccclai: `BackendOptions` is a list of `BackendOption`, and each `BackendOption` can be different type (Like `cpu: {thread: 2, profile: true}`). If we make `BackendOption` a template, then `BackendOptions` can only hold the same type of `BackendOption`, which is not we would like.

runtime/backend/backend_option.h
cccclai: std::variant is part of the std library, and this code needs to stay in core

cccclai:
Address comments in https://github.com/pytorch/executorch/pull/11389, please take a look

PR #11289, opened 2025-06-02
Make the intermediate output capturer more extensible

Juntian777:
Summary: Make the intermediate output capturer more extensible by adding a NodeFilter class. Add one rule to the intermediate output capturer to skip the getitem node for now.

Differential Revision: D75699297


Gasoonjia (approved):
LGTM.

PR #11290, opened 2025-06-02
Remove remaining reference to torchgen.gen_executorch

### LINKED ISSUE #11232: Move codegen from using `torchgen` to `executorch.codegen`

We already moved `torchgen.executorch` (in pytorch/pytorch) to `executorch.codegen` (in pytorch/executorch).

In [Codegen.cmake](https://github.com/pytorch/executorch/blob/main/tools/cmake/Codegen.cmake#L95) we are still using `torchgen.gen_executorch` so now it's time to change it to `executorch.codegen`.

Image

Notice that at this stage we haven't installed `executorch` so we might need to run the script in a different way (e.g., `from executorch.codegen import gen` may not work since executorch is not installed yet).

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @mergennachin @byjlw

larryliu0820:
Summary:
Fixes #11232
This is part of an effort to remove ExecuTorch related codegen code in pytorch/pytorch's torchgen module. We need to remove `torchgen.gen_executorch` in ExecuTorch and use `executorch.codegen.gen` instead.

Differential Revision: D75800153

### DIFF
--- a/codegen/gen.py
+++ b/codegen/gen.py
@@ -8,15 +8,32 @@ from pathlib import Path
 from typing import Any, Callable, TextIO, TYPE_CHECKING
 
 import yaml
-from executorch.codegen.api import et_cpp
-from executorch.codegen.api.custom_ops import (
-    ComputeNativeFunctionStub,
-    gen_custom_ops_registration,
-)
-from executorch.codegen.api.types import contextArg, ExecutorchCppSignature
-from executorch.codegen.api.unboxing import Unboxing
-from executorch.codegen.model import ETKernelIndex, ETKernelKey, ETParsedYaml
-from executorch.codegen.parse import ET_FIELDS, parse_et_yaml, parse_et_yaml_struct
+
+try:
+    from executorch.codegen.api import et_cpp
+    from executorch.codegen.api.custom_ops import (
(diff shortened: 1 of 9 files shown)


codegen/gen.py
```diff
@@ -8,15 +8,32 @@
...
+    from executorch.codegen.api.custom_ops import (
+        ComputeNativeFunctionStub,
+        gen_custom_ops_registration,
+    )
+    from executorch.codegen.api.types import contextArg, ExecutorchCppSignature
+    from executorch.codegen.api.unboxing import Unboxing
+    from executorch.codegen.model import ETKernelIndex, ETKernelKey, ETParsedYaml
+    from executorch.codegen.parse import ET_FIELDS, parse_et_yaml, parse_et_yaml_struct
+except ImportError:
+    # If we build from source, executorch.codegen is not available.
```
lucylq, line 23: noob q: why is this the case?

lucylq (approved):

codegen/gen.py
larryliu0820: To build ET from source we need selective build and generated bindings. So these scripts will be called before they are being installed into site-packages.

PR #11295, opened 2025-06-02
Implement GLU using internal views to avoid copying

swolchok:
GLU requires slicing the input Tensor into two halves. Currently, we accomplish this by copying; ExecuTorch does not support views in general because it requires Tensors to be contiguous. However, nothing stops us from implementing [the ATen that uses views](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/GatedLinearUnit.cpp#L35) entirely internally to the op.

To support this, I added `support_noncontiguous_tensors` as an optional template argument to BroadcastIndexesRange and plumbed it through to the elementwise_util functions as an optional SupportNonContiguousTensors parameter.


swolchok:
* __->__ #11295

kernels/portable/cpu/op_glu.cpp
```diff
@@ -120,11 +34,63 @@ void slice_tensor(
...
+          const auto one = static_cast<decltype(val_a)>(1.0);
+          return val_a * (one / (one + std::exp(-val_b)));
+        },
+        ctx,
+        first_half,
+        utils::SupportedTensorDtypes::FLOATHBF16,
+        second_half,
+        utils::SupportedTensorDtypes::FLOATHBF16,
+        out,
+        utils::internal::SupportNoncontiguousTensors());
```
manuelcandales, line 92: why didn't you pass the support_non_contiguous_tensors here as a template parameter?

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -295,7 +341,37 @@ inline void apply_bitensor_elementwise_fn(
...
+    typename Op>
+inline void apply_bitensor_elementwise_fn(
+    const Op& compute_fun,
+    KernelRuntimeContext& ctx,
+    const Tensor& a,
+    SupportedTensorDtypes a_dtypes,
+    const Tensor& b,
+    SupportedTensorDtypes b_dtypes,
+    const Tensor& out,
+    SupportNoncontiguousTensors) {
```
manuelcandales, line 369: Why do you need to add this overload, with this confusing anonymous argument, rather than adding a template parameter with a default value = false (i.e. just pass along the support_noncontiguous_tensors template parameter, in the original implementation above)

swolchok: > adding a template parameter with a default value = false

I tried this first, but AFAICT there's nowhere to put that template parameter -- template parameters with defaults need to go at the end, and then specifying a value for the parameter would require you to explicitly specify the type `Op`, which requires uglifying your code significantly if `Op` is a lambda type.

> confusing anonymous argument

this type of thing is in the standard. For example, [std::in_place_t](https://en.cppreference.com/w/cpp/utility/in_place.html).

swolchok:
noting that CI is green

GregoryComer (approved):

kernels/portable/cpu/util/elementwise_util.h
swolchok: thankfully this is in `namespace internal`, so if we come up with a better idea, we can fix it later. Given that, I will go ahead and land based on Gregory's accept.

kernels/portable/cpu/op_glu.cpp
swolchok: the argument exists because specifying it as a template parameter is messy; see other thread

PR #11310, opened 2025-06-03
Arm backend: Support for dynamic shapes and fix resize bugs

Burton2000:
Dynamic shapes are detected when converting torch shapes to TOSA shapes, when they contain SymInts rather than regular ints. These are converted to -1s as this is how TOSA represents dynamic shapes.

The resize op needs special handling for dynamic shapes, and will only work when all of the TOSA parameters (scale n/d, offset, border) work out to be constant values independent of the shape.

Also fix bug where align_corners was always set to True, when it should be False

Signed-off-by: Richard Burton <richard.burton@arm.com>
Co-authored-by: Robert Hughes @Rob-Hughes-Arm

### DIFF
--- a/backends/arm/operators/op_upsample_bilinear2d.py
+++ b/backends/arm/operators/op_upsample_bilinear2d.py
@@ -49,15 +49,18 @@ class UpsampleBilinear2dVisitor_0_80(NodeVisitor):
         input_dtype = inputs[0].dtype
 
         # tosa_shape output is NHWC, take HW
-        input_size_yx = torch.tensor(
-            tosa_shape(inputs[0].shape, inputs[0].dim_order)[1:3]
-        )
-        # Ignore scale and size parameters, directly use the output size as
-        # we only support static shapes currently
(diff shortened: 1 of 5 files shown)


zingo:
This PR caused a strange lintrunner error so I rebased it to check if it helped.

zingo:
@Burton2000 unfortunately it didn't help, it seems this PR cause some lint problems. Without looking at it it might be that one of you changes in a generic class get error when used so if you tests locally you  might not see it if you only check changed files in the patch e.g. you might need to use lintrunner  without args if you used something like 
`
lintrunner --revision 'HEAD^' -a`

Burton2000:
Looking better now?

zingo:
> Looking better now?

Yes but there is a fail in a Arm backend test now :(

zingo:
It was 
 backends/arm/test/models/test_deit_tiny_arm.py
 It seems a bit flakey I think the atol might need a bump as it seem to fail now and then, I retriggered the test for you.

zingo:
MacOS fails are unrelated as this delegate don't build or run on it.

zingo (approved):

PR #11314, opened 2025-06-03
Reusable test framework the Inspector tests

Juntian777:
Summary:
This Diff introduces a reusable test framework. The inspector_test_utils.py file provides methods to instantiate the model, retrieve expected outputs, and assert the correctness of actual outputs in a extensible way.

Also, update the intermediate_output_capturer_test to take advantage of the reusable framework and make the intermediate_output_capturer_test more extensible by using reusable setup method, making it easier to add new models and tests.

Differential Revision: D75803288


Gasoonjia (approved):
LGTM. Thanks for continue work

PR #11316, opened 2025-06-03
Milestone2.2: Optimize transposes in XNNPACK partition by removing redundant to_copy ops

leafs1:
### Summary
Optimize transposes in XNNPACK partition by adding a new remove_redundant_ops_pass that check for dim order conversion ops that cancel each other. The pass supports non-quantized conversions and also quantized graphs. In the quantized graph case, the conversion nodes and wrapping q/dq nodes will be removed. I also refactored the channels_last_tagged_reshape_pass code by modularizing some functions and adding some setter/getter functions.

This change will improve speed/memory at runtime by not executing redundant to_copy ops that would be there otherwise.

### Test plan
Created a TestChannelsLastTaggedReshapePass class which constructs graphs with multiple redundant to_copy ops in different positions and in quantized/non-quantized graphs. These redundant ops are either explicitly stated or generated via other passes. I asserted their removal after the passes finished.


backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -257,6 +265,22 @@ def insert_copy_and_assign_partner_nodes_quantization_sensitive(
             # in that case
             self.make_partners(original_input, copy_node)
 
+    def input_dim_order(
+        self, input_node: torch.fx.Node, input_order: InputDimOrder
+    ) -> bool:
+        if input_node.name == "x":
```
GregoryComer: Can you replace this with checking if the input_node is a placeholder?

backends/xnnpack/test/passes/test_remove_redundant_ops_pass.py
```diff
@@ -0,0 +1,104 @@
...
+)
+from executorch.backends.xnnpack._passes.convert_to_linear import ConvertToLinearPass
+from executorch.backends.xnnpack._passes.remove_redundant_ops_pass import (
+    RemoveRedundantOpsPass,
+)
+from executorch.backends.xnnpack.test.tester import RunPasses, Tester
+from executorch.exir.passes.memory_format_ops_pass import DimOrderOpsRevertPass
+
+
+class TestChannelsLastTaggedReshapePass(unittest.TestCase):
```
GregoryComer: Can we add a test that includes implicitly created dim order conversions? This will check to make sure that user created and pass-created converts get optimized out correctly. I expected it will work, but it would be nice to cover it since this is a common use case.

Maybe something like:
to_channels_last
upsample_nearest2d (not partitioned)
to_channels_first
conv

backends/xnnpack/_passes/remove_redundant_ops_pass.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # Store first subsequent visitation of to_copy node
+        prev = None
+        for node in original_nodes:
+            if len(node.all_input_nodes) == 0:
+                continue
+
+            # If we encounter a to_copy node, check if it is preceded by an opposite to_copy node
+            if node.target == exir_ops.edge.aten._to_copy.default:
+                if prev and ChannelsLastTaggedReshapePass.is_nchw_node(
```
GregoryComer: I think that there may be cases where the using the previous node in the iteration order might not actually be the first arg, especially in more complex graphs. Can you try replacing prev with node.args[0]? That should be sound in all cases.

backends/xnnpack/_passes/remove_redundant_ops_pass.py
```diff
@@ -0,0 +1,48 @@
...
+import torch
+from executorch.backends.xnnpack._passes.channels_last_tagged_reshape_pass import (
+    ChannelsLastTaggedReshapePass,
+)
+from executorch.backends.xnnpack._passes.xnnpack_pass import XNNPACKPass
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import PassResult
+
+
+class RemoveRedundantOpsPass(XNNPACKPass):
```
digantdesai: Nit: rename this as `RemoveRedundantCopyPass` or something? This is too generic to infer what its doing

backends/xnnpack/_passes/remove_redundant_copy_pass.py
```diff
@@ -0,0 +1,48 @@
...
+    def call(self, graph_module: torch.fx.GraphModule):
+        graph = graph_module.graph
+        original_nodes = list(graph.nodes)
+
+        for node in original_nodes:
+            if len(node.all_input_nodes) == 0:
+                continue
+
+            # If we encounter a to_copy node, check if its input is also a to_copy node with opposite format
+            if node.target == exir_ops.edge.aten._to_copy.default:
```
GregoryComer: I thought of one more edge case while reading this code. We should probably check to make sure that the second copy is the only user of the first. It's possible to have two copies in a row, but something else could use the output of the first. It's unlikely, but would lead to an invalid graph in this case.

leafs1: Added a check, thanks for this find

GregoryComer (approved):

backends/xnnpack/test/passes/test_remove_redundant_copy_pass.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def setUp(self):
+        torch._dynamo.reset()
+
+    def run_tester(self, module, inputs):
+        tester = Tester(
+            module.eval(),
+            inputs,
+        )
+        tester.export().to_edge_transform_and_lower().to_executorch().serialize().run_method_and_compare_outputs()
```
mcr229, line 39: for complicated paths, can we also try quantized models?

backends/xnnpack/test/passes/test_remove_redundant_copy_pass.py
```diff
@@ -0,0 +1,132 @@
...
+
+    class ChannelsLastToContiguous(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.conv1 = torch.nn.Conv2d(3, 3, 3)
```
mcr229, line 44: can we check conv1d as well.

mcr229 (changes requested):
Just a few extra test cases to make sure things look ok

mcr229 (approved):

PR #11317, opened 2025-06-03
Mitigate coremltool issue with python 3.13

metascroy:


docs/source/backends-coreml.md
```diff
@@ -195,6 +195,12 @@ Note that if the ExecuTorch model has graph breaks, there may be multiple extrac
...
-During lowering to the CoreML backend, you might see an error like: "ValueError: In op, of type [X], named [Y], the named input [Z] must have the same data type as the named input x. However, [Z] has dtype fp32 whereas x has dtype fp16."
+### During lowering
+1. "ValueError: In op, of type [X], named [Y], the named input [Z] must have the same data type as the named input x. However, [Z] has dtype fp32 whereas x has dtype fp16."
 
 This happens because the model is in FP16, but CoreML interprets some of the arguments as FP32, which leads to a type mismatch.  The solution is to keep the PyTorch model in FP32.  Note that the model will be still be converted to FP16 during lowering to CoreML unless specified otherwise in the compute_precision [CoreML CompileSpec](#coreml-compilespec).  Also see the [related issue in coremltools](https://github.com/apple/coremltools/issues/2480).
+
+2. coremltools/converters/mil/backend/mil/load.py", line 499, in export
+    raise RuntimeError("BlobWriter not loaded")
+
+If you're using Python 3.13, try reducing your python version to Python 3.12.  coremltools does not support Python 3.13.
```
swolchok: should we link to https://github.com/apple/coremltools/issues/2487 so people can easily tell when this goes stale?

swolchok (approved):

jathu:
What if we just hard enforce 3.12 as the upper bound version?

https://github.com/pytorch/executorch/blob/5ef38d3c7c8f0fd3ab0eaeaa37fed87a55a7c8f5/pyproject.toml#L51-L52

metascroy:
> What if we just hard enforce 3.12 as the upper bound version?
> 
> https://github.com/pytorch/executorch/blob/5ef38d3c7c8f0fd3ab0eaeaa37fed87a55a7c8f5/pyproject.toml#L51-L52

I think this is a good idea

metascroy:
> > What if we just hard enforce 3.12 as the upper bound version?
> > https://github.com/pytorch/executorch/blob/5ef38d3c7c8f0fd3ab0eaeaa37fed87a55a7c8f5/pyproject.toml#L51-L52
> 
> I think this is a good idea

@jathu added 3.13 upper bound to pyproject.toml.  Can you have a look?

jathu (approved):

pyproject.toml
```diff
@@ -49,7 +49,8 @@ classifiers = [
 ]
 
 # Python dependencies required for use.
-requires-python = ">=3.10"
+# coremltools has issue with python 3.13, see https://github.com/apple/coremltools/issues/2487 
+requires-python = ">=3.10,<=3.12"
```
jathu: I guess we should change this to sub 13
```suggestion
requires-python = ">=3.10,<3.13"
```

reneleonhardt:
Interestingly enough, I can install coremltools==8.3.0 and 9.0b1, but not executorch when I revert requires-python, so binary wheels from cibuildwheel would be greatly appreciated as always 😄

```shell
.venv/bin/python' '-c' 'import importlib.util; print(importlib.util.find_spec('torch').submodule_search_locations[0])'
      Error while adding torch to CMAKE_PREFIX_PATH. Exit code: 1
      Output:
      
      CMake Error at tools/cmake/Utils.cmake:135 (message):
        Error:
      
          File "<string>", line 1, in <module>
            import importlib.util; print(importlib.util.find_spec('torch').submodule_search_locations[0])
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      
        AttributeError: 'NoneType' object has no attribute
        'submodule_search_locations'
      
        tools/cmake/Utils.cmake:109 (get_torch_base_path)
        CMakeLists.txt:470 (find_package_torch_headers)
      
      
      -- Configuring incomplete, errors occurred!
```

By the way, if I move up EXECUTORCH_ROOT_PATH in backends/apple/coreml/scripts/install_requirements.sh this seems to work to remove manual syncing:
```shell
COREMLTOOLS_VERSION="$(grep coremltools== $EXECUTORCH_ROOT_PATH/pyproject.toml | grep -Eo '[0-9]+\.[0-9]+[^ ;"]*')"
```

PR #11342, opened 2025-06-03
[llm] Add a generic text only LLM runner

larryliu0820:
* __->__ #11342

Introducing `text_llm_runner`. This can be used to run all text only decoder only LLM models supported by ExecuTorch.

* Metadata is being read out from the .pte file and being used to construct the runner object.
* examples/models/llama/runner.h[.cpp] only contains a simple wrapper around `text_llm_runner.h[.cpp]`.

In next PRs I will move examples/models/phi-3-mini/runner to use the generic runner.

Will look into QNN and MediaTek runners as well.

Differential Revision: [D75910889](https://our.internmc.facebook.com/intern/diff/D75910889/)


swolchok:
looks like a duplicate of https://github.com/pytorch/executorch/pull/11343 ? should probably close one

lucylq (approved):

PR #11343, opened 2025-06-03
[llm] Add a generic text only LLM runner

larryliu0820:
Introducing `text_llm_runner`. This can be used to run all text only decoder only LLM models supported by ExecuTorch.

* Metadata is being read out from the .pte file and being used to construct the runner object.
* examples/models/llama/runner.h[.cpp] only contains a simple wrapper around `text_llm_runner.h[.cpp]`.

In next PRs I will move examples/models/phi-3-mini/runner to use the generic runner.

Will look into QNN and MediaTek runners as well.

Differential Revision: [D75910889](https://our.internmc.facebook.com/intern/diff/D75910889/)

ghstack-source-id: 288004703
Pull Request resolved: https://github.com/pytorch/executorch/pull/11342

### Summary

### Test plan


extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+      const GenerationConfig& config,
+      std::function<void(const std::string&)> token_callback = {},
+      std::function<void(const Stats&)> stats_callback = {}) override;
+  ::executorch::runtime::Error warmup(
+      const std::string& prompt,
+      int32_t max_new_tokens);
+  void stop() override;
+
+ private:
+  bool shouldStop_{false};
```
mergennachin, line 119: is this even used?

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+  ::executorch::runtime::Error load() override;
+  ::executorch::runtime::Error generate(
+      const std::string& prompt,
+      const GenerationConfig& config,
+      std::function<void(const std::string&)> token_callback = {},
+      std::function<void(const Stats&)> stats_callback = {}) override;
+  ::executorch::runtime::Error warmup(
+      const std::string& prompt,
+      int32_t max_new_tokens);
+  void stop() override;
```
mergennachin: docblock for public methods

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+  // Constructor with dependency injection
+  explicit TextLLMRunner(
+      std::unordered_map<std::string, int64_t> metadata,
+      std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
+      std::unique_ptr<::executorch::extension::Module> module,
+      std::unique_ptr<TextDecoderRunner> text_decoder_runner,
+      std::unique_ptr<TextPrefiller> text_prefiller,
+      std::unique_ptr<TextTokenGenerator> text_token_generator,
+      std::unique_ptr<Stats> stats,
+      float temperature = -1.0f);
```
mergennachin, line 60: docblock for the dependencies too

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+std::unordered_set<uint64_t> get_eos_ids(
+    tokenizers::Tokenizer* tokenizer,
+    Module* module);
+
+std::unique_ptr<TextLLMRunner> create_text_llm_runner(
+    const std::string& model_path,
+    std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
+    std::optional<const std::string> data_path = std::nullopt,
+    float temperature = -1.0f);
+
```
mergennachin: are you sure you want to expose this public functions as core APIs?

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+    size_t bos_token_index = 0,
+    size_t eos_token_index = 1);
+
+std::unordered_map<std::string, int64_t> get_llm_metadata(
+    tokenizers::Tokenizer* tokenizer,
+    Module* module);
+
+std::unordered_set<uint64_t> get_eos_ids(
+    tokenizers::Tokenizer* tokenizer,
+    Module* module);
```
mergennachin: isn't this an internal helper function? if so, move to the .cpp file

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,109 @@
...
+};
+
+std::unique_ptr<tokenizers::Tokenizer> load_tokenizer(
+    const std::string& tokenizer_path,
+    std::unique_ptr<std::vector<std::string>> special_tokens = nullptr,
+    std::optional<std::string> pattern = std::nullopt,
+    size_t bos_token_index = 0,
+    size_t eos_token_index = 1);
+
+std::unordered_map<std::string, int64_t> get_llm_metadata(
```
mergennachin: same...

extension/llm/runner/text_llm_runner.h
mergennachin: if you want to expose, please tag as ET_EXPERIMENTAL.

the current ET_EXPERIMENTAL is only applicable to the class

mergennachin (commented):
See inline

extension/llm/runner/text_llm_runner.h
larryliu0820: Yes android and ios are using the `stop()` API

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,183 @@
...
+ * @param tokenizer_path Path to the tokenizer file
+ * @param special_tokens Optional list of special tokens to add to the tokenizer
+ * @param pattern Optional regex pattern for tokenization
+ * @param bos_token_index Index of the beginning-of-sequence token
+ * @param eos_token_index Index of the end-of-sequence token
+ * @return std::unique_ptr<tokenizers::Tokenizer> Initialized tokenizer instance
+ */
+ET_EXPERIMENTAL std::unique_ptr<tokenizers::Tokenizer> load_tokenizer(
+    const std::string& tokenizer_path,
+    std::unique_ptr<std::vector<std::string>> special_tokens = nullptr,
```
mergennachin, line 157: instead of using smart pointer, how about this? 

```
const std::optional<std::vector<std::string>>& tokens
```

mergennachin: That way, you're passing by const reference and not using any pointers

extension/llm/runner/text_llm_runner.h
```diff
@@ -0,0 +1,183 @@
...
+   */
+  explicit TextLLMRunner(
+      std::unordered_map<std::string, int64_t> metadata,
+      std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
+      std::unique_ptr<::executorch::extension::Module> module,
+      std::unique_ptr<TextDecoderRunner> text_decoder_runner,
+      std::unique_ptr<TextPrefiller> text_prefiller,
+      std::unique_ptr<TextTokenGenerator> text_token_generator,
+      std::unique_ptr<Stats> stats,
+      float temperature = -1.0f);
```
mergennachin, line 60: if it's deprecated, why include this as part of the API?

If there's a user code, they can always just use create_text_llm_runner, or can we migrate the user code to use GenerationConfig directly?

examples/models/llama/runner/runner.h
```diff
@@ -19,74 +19,20 @@
...
-      std::unique_ptr<::executorch::extension::Module> module,
-      std::unique_ptr<::executorch::extension::llm::TextDecoderRunner>
-          text_decoder_runner,
-      std::unique_ptr<::executorch::extension::llm::TextPrefiller>
-          text_prefiller,
-      std::unique_ptr<::executorch::extension::llm::TextTokenGenerator>
-          text_token_generator,
-      std::unique_ptr<::executorch::extension::llm::Stats> stats,
-      float temperature = -1.0f);
+std::unique_ptr<llm::TextLLMRunner> create_llama_runner(
```
mergennachin, line 29: do you think we should keep example::create_llama_runner method at all? 

should we just replace with llm::load_tokenizer and llm::create_llm_runner  for all the current callsites of example::create_llama_runner

It seems unnecessary indirection, and we would like to showcase these low-level APIs in our examples, as opposed to having yet another indirection (i.e., example::create_llama_runner)

i'm fine if you want to this in a follow-up PR, but I think we should do it very soon.

mergennachin (approved):
See inline @larryliu0820 and thank you!

extension/llm/runner/text_llm_runner.h
larryliu0820: This is for JNI where it doesn't expose a temperature argument in generate() API. I think @kirklandsign is going to refactor JNI API (and demo app) to fully deprecate temperature to the constructor.

examples/models/llama/runner/runner.h
larryliu0820: Yeah don't have a strong opinion on this. I can do a follow up.

extension/llm/runner/text_llm_runner.h
larryliu0820: Wouldn't that trigger a copy? I think the tokenizer will have to hold the memory of these tokens.

larryliu0820:
Close, duplicate of #11342

Issue #10887, opened 2025-05-14
Get rid of fixed number cmake --build -j in sh and docs


kirklandsign:
### 📚 The doc issue

Right now we have a few places where the number of `cmake --build -j` is fixed.
https://github.com/search?q=repo%3Apytorch%2Fexecutorch+%22+-j%22&type=code

Ideally let's use 
mac: `$(sysctl -n hw.ncpu) - 1`
linux: `$(nproc) - 1`

Update these docs and scripts.

larryliu0820:
> I would like to be assigned this issue if possible, thank you.

Feel free to take it and submit PR.

MichaelIspas:
Thank you @larryliu0820, I've created a pull request #11351

PR #11351, opened 2025-06-04
Replace fixed number cmake --build -j in sh and docs #10887

MichaelIspas:
## Summary
Replaces fixed number of cmake --build -j entries with:

mac: -j$(($(sysctl -n hw.ncpu) - 1))
linux: -j$(($(nproc) - 1))

## Changes
Implemented changes to the best of my ability to .sh, .md, and .yml, could you kindly review if any changes were unnecessary or not relevant? I am a CS student learning how to work with FLOSS communities so I am happy to make any adjustments if needed, thank you.

### DIFF
--- a/.ci/docker/common/install_openssl.sh
+++ b/.ci/docker/common/install_openssl.sh
@@ -17,7 +17,7 @@ tar xf "${OPENSSL}.tar.gz"
 pushd "${OPENSSL}" || true
 ./config --prefix=/opt/openssl -d "-Wl,--enable-new-dtags,-rpath,$(LIBRPATH)"
 # NOTE: openssl install errors out when built with the -j option
-make -j6; make install_sw
+make -j$((nproc) - 1); make install_sw
 # Link the ssl libraries to the /usr/lib folder.
 ln -s /opt/openssl/lib/lib* /usr/lib
 popd || true
--- a/.ci/scripts/test_llama.sh
+++ b/.ci/scripts/test_llama.sh
@@ -158,7 +158,7 @@ cmake_install_executorch_libraries() {
         -DEXECUTORCH_BUILD_QNN="$QNN" \
         -DQNN_SDK_ROOT="$QNN_SDK_ROOT" \
         -Bcmake-out .
-    cmake --build cmake-out -j9 --target install --config "$CMAKE_BUILD_TYPE"
+    cmake --build cmake-out -j$(($(nproc) - 1)) --target install --config "$CMAKE_BUILD_TYPE"
 }
 
 cmake_build_llama_runner() {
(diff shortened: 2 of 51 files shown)


executorch.sln
```diff
@@ -0,0 +1,110 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
larryliu0820, line 1: Can you remove this file?

larryliu0820 (approved):
Looks good, thanks for your contribution! Please get rid of the additional file.

.ci/docker/common/install_openssl.sh
```diff
@@ -17,7 +17,7 @@ tar xf "${OPENSSL}.tar.gz"
 pushd "${OPENSSL}" || true
 ./config --prefix=/opt/openssl -d "-Wl,--enable-new-dtags,-rpath,$(LIBRPATH)"
 # NOTE: openssl install errors out when built with the -j option
-make -j6; make install_sw
+make -j$((nproc) - 1); make install_sw
```
jathu, line 20: I don't think this is valid syntax is it? You need an additional parentheses

```suggestion
make -j$(($(nproc) - 1)); make install_sw
```

.ci/scripts/test_llama.sh
```diff
@@ -158,7 +158,7 @@ cmake_install_executorch_libraries() {
         -DEXECUTORCH_BUILD_QNN="$QNN" \
         -DQNN_SDK_ROOT="$QNN_SDK_ROOT" \
         -Bcmake-out .
-    cmake --build cmake-out -j9 --target install --config "$CMAKE_BUILD_TYPE"
+    cmake --build cmake-out -j$(($(nproc) - 1)) --target install --config "$CMAKE_BUILD_TYPE"
```
jathu, line 161: This script is also used to test locally on macOS, so we cannot just use nproc. 


Overall what I would suggest is abstracting out this logic into a common script, and perhaps sharing the logic. For example, create `./tools/common_build.sh`

```bash
core_count() {
  if [[ "$(uname)" == "Darwin" ]]; then
    num_cores=$(sysctl -n hw.ncpu)
  else
    num_cores=$(nproc)
  fi
  echo $((num_cores - 1))
}
```

jathu (changes requested):

jathu:
Thanks for the contribution @MichaelIspas! I have some feedback, mainly that most of these scripts are used on macOS and Linux (and also Windows in bash), so we need to support both systems

nil-is-all:
Hi @MichaelIspas could you address the code changes requested and the CI tests after the fixes?

nil-is-all:
@MichaelIspas pinging again to address this PR, thanks

PR #8612, opened 2025-02-21
irangeify most of runtime/core/exec_aten

swolchok:
Had to skip scalar_type_util.h because it needs the constexpr patch to land in PyTorch (and for us to update our pin to pick it up).

PR #11360, opened 2025-06-04
Fix -Wformat issues from #8612

swolchok:
Built locally with -Wformat -Werror=format on a recent version of clang. (Apparently, clang12 which we use on CI is not new enough?)


swolchok:
* __->__ #11360

swolchok:
lucy sent this as #11362

PR #11376, opened 2025-06-04
remove raied error to solve ci issue.

Gasoonjia:
Summary: etrecord shouldn't raised error when user provide model doesn't have forward function. Remove the raised error to solve ci issue.

Differential Revision: D75983673


Juntian777 (approved):
LGTM

PR #11388, opened 2025-06-05
fix unittest-buck2 -- buck test doesn't imply buck build

swolchok:
When I wrote this, I mistakenly thought that `buck test` would build all non-test targets it was passed, but it doesn't. Fix that, and fix a bunch of builds that we weren't guarding so that CI stays green.


swolchok:
* __->__ #11388

lucylq (approved):
thanks!

swolchok:
noting that tests were green before rebase, but we need to wait for unittest-buck2 to be green before landing

swolchok:
unittest-buck is green, merging

PR #11389, opened 2025-06-05
[1/N] Add BackendOptions class

cccclai:
* #11865
* #11391
* #11390
* __->__ #11389

Introduce backend option as discussed in https://github.com/pytorch/executorch/discussions/10216

Step 1: Introducd Backend Option class

In later stage, it will be plugged in with the rest of the stack. BackendOptions is pretty much a list of BackendOption, and backend option is a key value pair. The key is a string, and the value can be 3 different types, including bool, string and int.

Differential Revision: [D75993712](https://our.internmc.facebook.com/intern/diff/D75993712/)

Differential Revision: [D75993712](https://our.internmc.facebook.com/intern/diff/D75993712)


cccclai:
Previous comment in the old PR: https://github.com/pytorch/executorch/pull/11288

From @JacobSzwejbka : 

> use int64_t for int type.

cccclai:
From @mergennachin: 

> 
> I don't think we need to define a generic BackendOption that takes arbitrary key value. Each backend owners could define their options instead. Vulkan will define its own config structure, Core ML will define its own config structure. Wdyt? 
> 
> @JacobSzwejbka, @mcr229, @kimishpatel 
> 
> Something like this:
> 
> ```
> BackendInterface { // parent interface
>   virtual Error setOption(const BackendOptions& options) = 0;
> }
> 
> struct BackendOptions {  // Empty parent struct
>   virtual ~BackendOptions() = default;
> };
> ```
> 
> Each individual backends define their own configs
> 
> ```
> struct VulkanBackendOptions : public BackendOptions {
>   const char* metadata[20] = "";
>   int thread_count = 1;
>   // ... other Vulkan-specific options
> };
> 
> ```
> 
> ```
> VulkanBackend extends BackendInterface {
>   Error setOption(const BackendOptions& options) override {
>     auto* vulkan_opts = dynamic_cast<const VulkanBackendOptions*>(&options);
>   }
> }
> ```
> 
> // User code
> 
> ```
> VulkanBackendOptions opts;
> opts.thread_count = 5;
> auto backend = std::make_unique<VulkanBackend>();
> backend->setOption(opts);  // 
> ```

cccclai:
My response: 

> I thought about this idea at the beginning, but it means that users code needs to link to a specific backend, otherwise the runner code will fail to build. Currently the same user code can be backend agnostic, meaning the same code can still run with different backends with/without linking the backend. As the example, the portable runner can work with/without xnnpack https://github.com/pytorch/executorch/blob/main/examples/xnnpack/executor_runner/targets.bzl. 
> 
> It also exposes the backend specific symbol to users, and the contract interface is a bit looser. Previously, The contract is like backend <-> ET <-> backend, and if we change it to this, it can be user <-> backend. And users can pass arbitrary things to backend. Users can run backend->init, backend->execute in their own code too, and it will be out of our control.
> 
> With the proposed solution (https://github.com/pytorch/executorch/discussions/10216), the user code can still be backend agnostic, and the backend option is still optional. We are still on the hook of the contract.

kimishpatel:
I think what @mergennachin suggested makes sense. If user code is setting backend specific option they already have to know what backend they are talking about, unless some options are backend agnostic.

Other piece of the puzzle is how the user will update this. And from #10216 it seems that we will pipe `update` (I would call set_backend_options) method that will set all the options. But part that is not clear to me is how would user indicate which backend they setting this option for. So somehow we need a string identifer for the backends that can be used to figure out which backend's update method be called. @mergennachin in this case we cannot quite have what you proposed. Imagine you have a pointer to an instance of `VulkanBackendOptions`. You pass this to runtime's update method which accepts `BackendOptions*`.  Eventually you want to call VulkanBackend's update options method so you need to cast BackendOptions* to VulkanBackendOptions*. You have two choices either do static_cast or dynamic_cast. The later requires rtti which we dont allow. For static_cast, it has to be safe and you need to know that BackendOption* is actually pointing to VulkanBackendOption. To do that you need to have BackendOptions store tag containing what is the type of the derived class. If we do that, then what you suggest can work.

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+#include <executorch/runtime/core/error.h>
+#include <cstddef>
+#include <cstring>
+#include <variant>
+
+namespace executorch {
+namespace runtime {
+
+// Strongly-typed option key template
+template <typename T>
```
kimishpatel: this class doesnt need to be templated. No use of T

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+};
+
+// Union replaced with std::variant
+using OptionValue = std::variant<bool, int, const char*>;
+
+struct BackendOption {
+  const char* key; // key is the name of the backend option, like num_threads,
+                   // enable_profiling, etc
+  OptionValue
+      value; // value is the value of the backend option, like 4, true, etc
```
kimishpatel: format?

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+struct OptionKey {
+  const char* key;
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Union replaced with std::variant
+using OptionValue = std::variant<bool, int, const char*>;
+
+struct BackendOption {
+  const char* key; // key is the name of the backend option, like num_threads,
```
kimishpatel: I am a bit concerned about storing pointer to the key. Users may do things like
```
std::string my_key("my_key");
BackendOption().set_option<int>(my_key.c_str(), 10);
``` 

When that goes out of scope we have a problem. I suggest instead use something like `const char[kMaxKeyLength] key`

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+                   // enable_profiling, etc
+  OptionValue
+      value; // value is the value of the backend option, like 4, true, etc
+};
```
kimishpatel: The other thing I realize is why do we need BackendOption as a separate struct?

Can we not just have `BackendOptions` class instantiate array of keys and values? The interface there is set_option with key name and value, get_option using key name. User never really needs to get BackendOption

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+
+template <size_t MaxCapacity>
+class BackendOptions {
+ public:
+  // Initialize with zero options
+  BackendOptions() : size_(0) {}
+
+  // Type-safe setters
+  template <typename T>
+  void set_option(OptionKey<T> key, T value) {
```
kimishpatel: Bassed on suggestion above 
```suggestion
  void set_option(const char [kMaxKeyLength]& key, T value) {
```

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+    for (size_t i = 0; i < size_; ++i) {
+      if (strcmp(options_[i].key, k) == 0) {
+        options_[i].value = value;
+        return;
+      }
+    }
+    // Add new option if space available
+    if (size_ < MaxCapacity) {
+      options_[size_++] = BackendOption{k, value};
+    }
```
kimishpatel: set error if no capacity

runtime/backend/backend_options.h
```diff
@@ -0,0 +1,93 @@
...
+  return OptionKey<bool>(k);
+}
+
+constexpr OptionKey<int> IntKey(const char* k) {
+  return OptionKey<int>(k);
+}
+
+constexpr OptionKey<const char*> StrKey(const char* k) {
+  return OptionKey<const char*>(k);
+}
```
kimishpatel: Dont think you need these

kimishpatel (changes requested):
made some suggestions

runtime/backend/backend_options.h
cccclai: It's just helper functions to simplify users' code.

runtime/backend/backend_options.h
cccclai: Users don't need get backend option, but the dispatcher needs it and dispatch to the actual backend.

runtime/backend/backend_options.h
cccclai: Yeah good point. Will update

runtime/backend/backend_options.h
cccclai: Actually you're right. The backend options map is for dispatching. and each backend will get an array ref of backend options.

runtime/backend/options.h
```diff
@@ -0,0 +1,103 @@
...
+#include <cstring>
+#include <variant>
+
+namespace executorch {
+namespace runtime {
+
+// Strongly-typed option key template
+template <typename T>
+struct OptionKey {
+  const char* key;
```
mergennachin: Good practice to add

```
using value_type = T;
```

runtime/backend/options.h
```diff
@@ -0,0 +1,103 @@
...
+  constexpr explicit OptionKey(const char* k) : key(k) {}
+};
+
+// Union replaced with std::variant
+using OptionValue = std::variant<bool, int, const char*>;
+
```
mergennachin, line 28: Document the safety and lifecycle of string types

All string keys and values must have static storage duration (string literals, static const char arrays, or global constants). The BackendOptions class does NOT take ownership of strings.

runtime/backend/options.h
```diff
@@ -0,0 +1,103 @@
...
+  return OptionKey<bool>(k);
+}
+
+constexpr OptionKey<int> IntKey(const char* k) {
+  return OptionKey<int>(k);
+}
+
+constexpr OptionKey<const char*> StrKey(const char* k) {
+  return OptionKey<const char*>(k);
+}
```
mergennachin: Can you force to accept only string literals with this?

```
template <size_t N>
constexpr OptionKey<const char*> StrKey(const char (&literal)[N]) {
  return OptionKey<const char*>(literal);
}
```

runtime/backend/options.h
```diff
@@ -0,0 +1,103 @@
...
+
+template <size_t MaxCapacity>
+class BackendOptions {
+ public:
+  // Initialize with zero options
+  BackendOptions() : size_(0) {}
+
+  // Type-safe setters
+  template <typename T>
+  void set_option(OptionKey<T> key, T value) {
```
mergennachin: rename this to set_option_impl and make it a private function. And add type safe public APIs to enforce key is always safe string... And value is also safe in the case of string.

```
  // String options - only accepts string literals/arrays
  template <size_t KeyN, size_t ValueN>
  Error set_option(
    const char (&key)[KeyN], 
    const char (&value)[ValueN]
  ) noexcept {
    return set_option_impl(key, static_cast<const char*>(value));
  }
  
  // Non-string options
  template <size_t N>
  Error set_option(const char (&key)[N], bool value) noexcept {
    return set_option_impl(key, value);
  }
  
  template <size_t N>
  Error set_option(const char (&key)[N], int value) noexcept {
    return set_option_impl(key, value);
  }
```

mergennachin (commented):
See inline

runtime/backend/options.h
```diff
@@ -0,0 +1,103 @@
...
+#include <cstddef>
+#include <cstring>
+#include <variant>
+
+namespace executorch {
+namespace runtime {
+
+// Strongly-typed option key template
+template <typename T>
+struct OptionKey {
```
kimishpatel: I dont think we need OptionKey as a separate class. Just use `const char *` in the `BackendOption` class.

However, the thing that worries me the most is by owning `const char*` we are giving users a footgun. They can allocate pointer to the backend option name on stack, pass the pointer around and it will go out-of-scope. If this results in segfault that might be your best option. Worst can lead to undefined behaviors.

I would suggest that we just store `const char[kMaxLegthOfBackendOptionName]`.

runtime/backend/backend_options.h
cccclai: updated

cccclai:
@kimishpatel @mergennachin thanks for the feedback! I've addressed most of them. If there is anything I miss, let me know

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+   * If the key already exists, updates its value. Otherwise, adds a new option.
+   *
+   * @tparam N The length of the key string (automatically deduced)
+   * @param key The option key (must be a string literal or array)
+   * @param value The boolean value to set
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  template <size_t N>
+  Error set_option(const char (&key)[N], bool value) noexcept {
+    ET_CHECK_MSG(N <= kMaxOptionKeyLength, "Option key is too long");
```
mergennachin: static_assert(N <= kMaxOptionKeyLength, "Option key is too long");

for compile time check, as opposed to runtime check

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+   * If the key already exists, updates its value. Otherwise, adds a new option.
+   *
+   * @tparam N The length of the key string (automatically deduced)
+   * @param key The option key (must be a string literal or array)
+   * @param value The integer value to set
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  template <size_t N>
+  Error set_option(const char (&key)[N], int value) noexcept {
+    ET_CHECK_MSG(N <= kMaxOptionKeyLength, "Option key is too long");
```
mergennachin: .

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+   * NOT take ownership of the string - it only stores the pointer.
+   *
+   * @tparam N The length of the key string (automatically deduced)
+   * @param key The option key (must be a string literal or array)
+   * @param value The string value to set (must have static storage duration)
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  template <size_t N>
+  Error set_option(const char (&key)[N], const char* value) noexcept {
+    ET_CHECK_MSG(N <= kMaxOptionKeyLength, "Option key is too long");
```
mergennachin: .

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+      if (strcmp(options_[i].key, key) == 0) {
+        options_[i].value = value;
+        return Error::Ok;
+      }
+    }
+    // Add new option if space available
+    if (size_ < MaxCapacity) {
+      BackendOption new_option;
+      strncpy(new_option.key, key, kMaxOptionKeyLength - 1);
+      new_option.key[kMaxOptionKeyLength - 1] = '\0';
```
mergennachin:
```
const size_t key_len = std::strlen(key);
const size_t copy_len = std::min(key_len, kMaxOptionKeyLength - 1);
std::memcpy(new_option.key, key, copy_len);
new_option.key[copy_len] = '\0';
```

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+   * Default constructor - initializes with zero options.
+   */
+  BackendOptions() : size_(0) {}
+
+  /**
+   * Returns a const view of all stored options as a Span.
+   *
+   * @return A const Span containing all BackendOption entries
+   */
+  executorch::runtime::Span<BackendOption> view() const {
```
mergennachin:
```
executorch::runtime::Span<const BackendOption> view() const
```

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+ * @tparam MaxCapacity The maximum number of options that can be stored
+ */
+template <size_t MaxCapacity>
+class BackendOptions {
+ public:
+  /**
+   * Default constructor - initializes with zero options.
+   */
+  BackendOptions() : size_(0) {}
+
```
mergennachin, line 58: Potentially have this

```
  /**
   * Copy constructor
   */
  BackendOptions(const BackendOptions& other) : size_(other.size_) {
    for (size_t i = 0; i < size_; ++i) {
      options_[i] = other.options_[i];
    }
  }

  /**
   * Copy assignment operator
   */
  BackendOptions& operator=(const BackendOptions& other) {
    if (this != &other) {
      size_ = other.size_;
      for (size_t i = 0; i < size_; ++i) {
        options_[i] = other.options_[i];
      }
    }
    return *this;
  }
```

mergennachin (approved):
See inline, approving but inline comments should be addressed

kimishpatel:
@mergennachin I am concerned about lifecycle and potential footgun it introduces when setting string values. So I would like to resolve that before we land

runtime/backend/options.h
kimishpatel: You resolved this comment but I dont see it being addressed

mergennachin:
@kimishpatel 

Can we just always require string value to be a string literals, static const char arrays, or global constants in the API like this? 

https://github.com/pytorch/executorch/pull/11389#discussion_r2159218778

runtime/backend/options.h
kimishpatel: For the actual value itself, I think we should do this where std::variant itself holds array of chars rather than pointer to char. Like this https://godbolt.org/z/odTc9ex96

runtime/backend/options.h
kimishpatel: Looks like you addressed but kept `OptionKey`. please remove

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+  executorch::runtime::Span<BackendOption> view() const {
+    return executorch::runtime::Span<BackendOption>(options_, size_);
+  }
+
+  /**
+   * Returns a mutable view of all stored options as a Span.
+   *
+   * @return A mutable Span containing all BackendOption entries
+   */
+  executorch::runtime::Span<BackendOption> view() {
```
kimishpatel, line 82: maybe make it explicit that this is a mutable view.

runtime/backend/options.h
kimishpatel: why is there KeyLen here? Can it not be just max length? Like why are we templating this on keylength? If it is to allow for users to construct `const char[10]` vs. `const char [5]`, I would say just make recommended use via `const char[kMaxOptionKeyLength]`? Feels like that is simler

runtime/backend/options.h
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @tparam T The expected type of the option value (bool, int, or const char*)
+   * @tparam KeyLen The length of the key string (automatically deduced)
+   * @param key The option key to look up
+   * @param out Reference to store the retrieved value
+   * @return Error::Ok if found and type matches, Error::NotFound if key doesn't
+   * exist, Error::InvalidArgument if type doesn't match
+   */
+  template <typename T, size_t KeyLen>
+  Error get_option(const char (&key)[KeyLen], T& out) const {
```
kimishpatel, line 148: same as before on keylen

runtime/backend/test/backend_options_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  bool bval;
+  int ival;
+
+  // Boolean get should fail - type was overwritten to INT
+  EXPECT_EQ(options.get_option("flag", bval), Error::InvalidArgument);
+
+  // Integer get should succeed with correct value
+  EXPECT_EQ(options.get_option("flag", ival), Error::Ok);
+  EXPECT_EQ(ival, 123);
+}
```
kimishpatel, line 95: Add tests for mutability of the options now that you have mutable view

kimishpatel (changes requested):
I think we should make option itself have char array instead of char*

kimishpatel:
> @kimishpatel
> 
> Can we just always require string value to be a string literals, static const char arrays, or global constants in the API like this?
> 
> [#11389 (comment)](https://github.com/pytorch/executorch/pull/11389#discussion_r2159218778)

Yeah those are fine options. I just want API to make it clear, like what you suggested in the comment (although I think length specifier, for option and value, being template param seems a bit weird. Just have a max longest length as some const)

runtime/backend/test/backend_options_test.cpp
cccclai: Added

runtime/backend/options.h
cccclai: Updated

cccclai:
New comments are addressed, let me know your thoughts

runtime/backend/options.h
```diff
@@ -0,0 +1,217 @@
...
+   * Sets a boolean option value for the given key.
+   * If the key already exists, updates its value. Otherwise, adds a new option.
+   *
+   * @tparam N The length of the key string (automatically deduced)
+   * @param key The option key (must be a string literal or array)
+   * @param value The boolean value to set
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  template <size_t N>
+  Error set_option(const char (&key)[N], bool value) noexcept {
```
kimishpatel, line 96: Do we need N? It will just create N different methods for many possible lengths. My suggestion is to remove N as template param and expect callsites to be something like
```
char my_option_name[kMaxOptionKeyLength] = "use_gpu";
...set_option(my_option_name, True)
```

runtime/backend/options.h
```diff
@@ -0,0 +1,217 @@
...
+   * @param value The string value to set (must have static storage duration)
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  template <size_t N>
+  Error set_option(const char (&key)[N], const char* value) noexcept {
+    static_assert(N <= kMaxOptionKeyLength, "Option key is too long");
+    // Create a fixed-size array and copy the string
+    std::array<char, kMaxOptionValueLength> arr;
+    strncpy(arr.data(), value, kMaxOptionValueLength - 1);
+    arr[kMaxOptionValueLength - 1] = '\0'; // Ensure null termination
```
kimishpatel, line 134: Using the suggestion I mentioned above will also not require you to do this. ALthough its not a bit deal

kimishpatel (approved):
Looks good. I still think we should not need N or KeyLen template params. So I would suggest to fix that.

cccclai:
> > @kimishpatel
> > Can we just always require string value to be a string literals, static const char arrays, or global constants in the API like this?
> > [#11389 (comment)](https://github.com/pytorch/executorch/pull/11389#discussion_r2159218778)
> 
> Yeah those are fine options. I just want API to make it clear, like what you suggested in the comment (although I think length specifier, for option and value, being template param seems a bit weird. Just have a max longest length as some const)

I remove the template N, given that it might generate too many code and causing too much size regression. However the cost is that the key length check is runtime instead of compile time, which I think maybe fine

kimishpatel:
> > > @kimishpatel
> > > Can we just always require string value to be a string literals, static const char arrays, or global constants in the API like this?
> > > [#11389 (comment)](https://github.com/pytorch/executorch/pull/11389#discussion_r2159218778)
> > 
> > 
> > Yeah those are fine options. I just want API to make it clear, like what you suggested in the comment (although I think length specifier, for option and value, being template param seems a bit weird. Just have a max longest length as some const)
> 
> I remove the template N, given that it might generate too many code and causing too much size regression. However the cost is that the key length check is runtime instead of compile time, which I think maybe fine

There shouldnt be any key lengths check really though? Your fn would be `get_option(const char optin_name[kMaxLength]...)`

runtime/backend/options.h
```diff
@@ -0,0 +1,220 @@
...
+   * Sets a boolean option value for the given key.
+   * If the key already exists, updates its value. Otherwise, adds a new option.
+   *
+   * @tparam N The length of the key string (automatically deduced)
+   * @param key The option key (must be a string literal or array)
+   * @param value The boolean value to set
+   * @return Error::Ok on success, Error::InvalidArgument if storage is full
+   */
+  Error set_option(const char* key, bool value) noexcept {
+    ET_CHECK_MSG(
```
kimishpatel: See my comment. I didt mean this

runtime/backend/options.h
cccclai: Turns out that we pretty much need it to be mutable, because the way options work is we create one options, and option, then pass it to set_option/get_option. I remove the const one.

cccclai:
Discuss offline with @kimishpatel and experiement with `set_option(const char optin_name[kMaxLength]...)`, looks like array length is still not known at compile time, hence switching back to template base

PR #11390, opened 2025-06-05
[2/N] Add option context

cccclai:
* #11865
* #11391
* __->__ #11390
* #11389

For future needs without breacking API BC, in case we need to pass more information to the update API

Differential Revision: [D75919212](https://our.internmc.facebook.com/intern/diff/D75919212/)

Differential Revision: [D75919212](https://our.internmc.facebook.com/intern/diff/D75919212)


kimishpatel (commented):
I am not sure on the use of this. I understand you said this is introduced for BC, but can you elaborate on the BC aspect of it? Is it to avoid makign changes to client call sites? If this struct is populated then I imagine it will require client callsites to change anyway, if they want to make use of it. So maybe a bit more on this in the summary would be good.

If you do want to keep it then maybe get_option and set_option should provide both overloads? One with backendoptioncontext and one without..

cccclai:
> I am not sure on the use of this. I understand you said this is introduced for BC, but can you elaborate on the BC aspect of it? Is it to avoid makign changes to client call sites? If this struct is populated then I imagine it will require client callsites to change anyway, if they want to make use of it. So maybe a bit more on this in the summary would be good.
> 
> If you do want to keep it then maybe get_option and set_option should provide both overloads? One with backendoptioncontext and one without..

Yeah to avoid making change to client call sites. I was thinking it's for pupulating runtime info from ET to the backend during set option. We can take a look at how init context helps on passing new info from ET to backend without causing BC.

Like when we pass event tracer: https://github.com/pytorch/executorch/commit/0c4053e00a629e4996ef73653f62eebcbeb7c599

And the named data map: 
https://github.com/pytorch/executorch/commit/492366a239fb0974d065f05ee9a37ea468ad63c2

kimishpatel (approved):

PR #11391, opened 2025-06-05
[3/N] Add get_option/set_option function in backend interface

cccclai:
* #11865
* __->__ #11391
* #11390
* #11389

Add update function in backend interface class. The update function will receive the backend options from dispatched by the ET runtime.

ET runtime's logic: loop over each backend and it's corresponding backend options, dispatch the backend options to the corresponding backend

Next step, will add update API in the method and then module
@exported-using-ghexport

Differential Revision: [D75919242](https://our.internmc.facebook.com/intern/diff/D75919242/)

Differential Revision: [D75919242](https://our.internmc.facebook.com/intern/diff/D75919242)


runtime/backend/backend_options.h
```diff
@@ -117,6 +118,10 @@ class BackendOptions {
     return err;
   }
 
+  executorch::runtime::ArrayRef<BackendOption> view() const {
```
kimishpatel: maybe more meaningful name. `get_backend_options` or `get_ref_to_backend_options`

runtime/backend/interface.h
```diff
@@ -99,6 +101,20 @@ class BackendInterface {
...
 
+  /**
+   * Responsible update the backend status, if any. The backend options are passed in
+   * by users, and the backend can update its internal status based on the options.
+   *
+   * @param[in] context Runtime context if any. Currently it's not used.
+   * @param[in] args A list of BackendOptions passed in by users.
+   * @retval Error::Ok if successful.
+   */
+   ET_NODISCARD virtual Error update(
```
kimishpatel: I would jsut call it set_backend_options which seems more appropriate

runtime/backend/interface.h
```diff
@@ -99,6 +101,20 @@ class BackendInterface {
...
+    BackendUpdateContext& context,
+    const executorch::runtime::ArrayRef<BackendOption>& backend_options) const {
+        return Error::Ok;
+      };
```
kimishpatel: what is the implementation of this? If it is in the next diff, can you pull it here?

runtime/backend/interface.h
```diff
@@ -99,6 +101,20 @@ class BackendInterface {
...
+  /**
+   * Responsible update the backend status, if any. The backend options are passed in
+   * by users, and the backend can update its internal status based on the options.
+   *
+   * @param[in] context Runtime context if any. Currently it's not used.
+   * @param[in] args A list of BackendOptions passed in by users.
+   * @retval Error::Ok if successful.
+   */
+   ET_NODISCARD virtual Error update(
+    BackendUpdateContext& context,
```
kimishpatel: what is supposed to be the role of BackendUpdateContext?

runtime/backend/test/backend_interface_update_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+     executorch::runtime::runtime_init();
+     mock_backend = std::make_unique<MockBackend>();
+    //  static Error register_success = register_executor_backend();
+    }
+
+  std::unique_ptr<MockBackend> mock_backend;  
+  BackendOptions<5> options;  
+};
+
+TEST_F(BackendInterfaceUpdateTest, HandlesInvalidOption) {
```
kimishpatel, line 115: There are all these tests but I am not sure if they are meanigfully testing anything? do you need all of them?

runtime/backend/interface.h
cccclai: It's a place holder for future API changes. Currently have the context for init and execute

runtime/backend/interface.h
cccclai: It's the base function. If backends didn't implement this function, this function will be called. If we set it to virtual, then we need all backends to implement this fucntion.

runtime/backend/interface.h
cccclai: I kind of like this API name more, but we can discuss about it.

runtime/backend/test/backend_interface_update_test.cpp
cccclai: I think it's helpful to make sure we know each option inside the backend is updated successfully, and they are isolated.

kimishpatel (approved):

PR #11394, opened 2025-06-05
Arm backend: Replace asserts with exceptions in passes

Sebastian-Larsson:
- annotate_channels_last_dim_order_pass: raise on dim_order length != 4 and != 5
- remove_clone_pass: ValueError if args count != 1
- fold_qdq_with_annotated_qparams_pass: RuntimeError if dq_op mismatch or unexpected meta key presence
- convert_split_to_slice: ValueError if split sizes don't match dim
- insert_table_ops: ValueError if input/output quant params count != 1


zingo:
Cortex-M size test error is unrelated as this only changes python file and nothing on the target.
Also this test fails on main since https://github.com/pytorch/executorch/pull/10675

zingo (approved):

PR #11396, opened 2025-06-05
NXP backend: Enable initial unit tests workflow

robert-kalmar:
### Summary
Initial workflow definition for running unit tests for Neutron (NXP) backend.

Follow up for the https://github.com/pytorch/executorch/pull/11109, our IT updated the eIQ PyPI, so the neutron-converter python package shall be reachable using the executorch's self-hosted agents too.

@digantdesai , @mergennachin as the problem was only reproduced with the self-hosted agents used by the ExecuTorch/PyTorch, I need your help to verify the IT change. Can you please let run the "trunk" workflow? 

cc @digantdesai @JakeStevens

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -686,3 +686,32 @@ jobs:
       build-mode: Release
       build-tool: cmake
       docker-image: executorch-ubuntu-22.04-clang12
+
+  unittest-nxp-neutron:
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-clang12
+      submodules: 'recursive'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
+      script: |
+        set -eux
+        
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+        
+        # Build and install Executorch
+        PYTHON_EXECUTABLE=python \
+        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON" \
+        .ci/scripts/setup-linux.sh --build-tool "cmake"
+        
+        # Install test requirements
+        pip install -r backends/nxp/requirements-tests.txt
+        
+        # Run pytest
+        PYTHON_EXECUTABLE=python bash backends/nxp/run_unittests.sh
--- a/backends/nxp/run_unittests.sh
+++ b/backends/nxp/run_unittests.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+set -eux
+
+SCRIPT_DIR=$(dirname $(readlink -fm $0))
+EXECUTORCH_DIR=$(dirname $(dirname $SCRIPT_DIR))
+
+cd $EXECUTORCH_DIR
+
+# '-c /dev/null' is used to ignore root level pytest.ini.
+PYTHONPATH=`cd ..; pwd` pytest -c /dev/null backends/nxp/tests/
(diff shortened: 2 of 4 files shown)


backends/nxp/run_unittests.sh
```diff
@@ -0,0 +1,14 @@
...
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+set -eux
+
+SCRIPT_DIR=$(dirname $(readlink -fm $0))
+EXECUTORCH_DIR=$(dirname $(dirname $SCRIPT_DIR))
+
+cd $EXECUTORCH_DIR
```
digantdesai, line 11: Don't we need to install the NXP sdk here? See my PR with retries, I am on phone so can't look it up.

digantdesai:
Looks good, left a comment, and add a trunk label, hopefully it will be all green now.

.github/workflows/trunk.yml
```diff
@@ -686,3 +686,29 @@ jobs:
...
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-clang12
+      submodules: 'true'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
```
cccclai: @kirklandsign mentioned it's better to have a separate docker image for a specific backend to avoid unstable CI jobs.

.github/workflows/trunk.yml
```diff
@@ -686,3 +686,29 @@ jobs:
...
+
+  unittest-nxp-neutron:
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: executorch-ubuntu-22.04-clang12
+      submodules: 'true'
```
robert-kalmar: Must be set to "recursive" since 7e034cae.

backends/nxp/run_unittests.sh
robert-kalmar: Done. The test requirements are installed in the trunk.yaml: 
```
# Install test requirements
pip install -r backends/nxp/requirements-tests.txt
```

.github/workflows/trunk.yml
robert-kalmar: ✅

.github/workflows/trunk.yml
robert-kalmar: Sounds reasonable.  
For curiosity, the github job runs from a clean docker container, right? It does not reuse already instantiated containers. (I don't see this explicitly from the log, but as the docker image was pulled I guess clean docker container is used) 
Where I am heading if we can run the trunk workflow, after my fixes, to chase one error at a time, or we must isolate the build into its own docker container to not break containers for other jobs. 

@cccclai , @kirklandsign, can you elaborate more about "unstable CI jobs". The background it to not mix the backends SDK, what can be potentially installed during the docker image creation (https://github.com/pytorch/executorch/blob/main/.ci/docker/ubuntu/Dockerfile#L85)? If I understand the script correctly currently only Android  really uses this option, right?

robert-kalmar: Addressed is second PR: https://github.com/pytorch/executorch/pull/11473

robert-kalmar:
@digantdesai , we are green. 
The tests (including trunk workflow) passed. https://github.com/pytorch/executorch/actions/runs/15498620828/job/43641195770?pr=11396#step:15:11806. 

The failing trunk/test-arm-cortex-m-size-test seems unrelated. It failed on slight size increase for the `cmake-out/test/size_test` application (https://github.com/pytorch/executorch/actions/runs/15498620828/job/43641195433?pr=11396#step:15:15498). Maybe someone from Arm to confirm? 

The @cccclai  request is addressed in separate PR https://github.com/pytorch/executorch/pull/11473 to have a baseline here. Started the tests, if pass, we can merge the PR 11473 instead of this as it contains the `NXP backend: Enable initial unit tests workflow` commit too.

digantdesai:
Cc @GregoryComer for Arm size test.
LGTM otherwise.

digantdesai (approved):

digantdesai:
Thanks you guys. Since this is on CI trunk for now, make sure to manually apply trunk label to NXP PRs. Also going forward let's ensure new PRe include tests for the CI so we don't regress silently.

PR #11430, opened 2025-06-05
Use Memcpy in copy_utils

ethansfng:
Summary:
standard elementwise copy in copy_utils is inefficient, use memcpy instead

Rollback Plan:

Differential Revision: D76061894

### DIFF
--- a/kernels/portable/cpu/util/copy_ops_util.h
+++ b/kernels/portable/cpu/util/copy_ops_util.h
@@ -27,10 +27,17 @@ void _as_strided_copy(
     ArrayRef<int64_t> stride,
     int64_t dim) {
   // the last dimension, copy data
+  const int64_t stride_dim = stride.at(dim);
   if (dim == static_cast<int64_t>(size.size()) - 1) {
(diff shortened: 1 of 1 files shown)


kernels/portable/cpu/util/copy_ops_util.h
```diff
@@ -28,9 +28,15 @@ void _as_strided_copy(
...
-      input_data += stride.at(dim);
+    const size_t num_elements = size.at(dim);
+    // use memcpy for contiguous memory
+    if (stride.at(dim) == 1) {
+      memcpy(output_data, input_data, num_elements * sizeof(CTYPE));
+    } else {
+      for (const auto i : c10::irange(num_elements)) {
+        output_data[i] = *input_data;
+        input_data += stride.at(dim);
+      }
```
swolchok: please cache stride.at(dim) in a local. If CTYPE is ever int64_t or char, C++ aliasing rules mean the compiler won't be able to prove that the write to output_data doesn't affect stride.at(dim) and thus it will have to re-load on every loop iteration.

swolchok (approved):

PR #11432, opened 2025-06-05
Plumbing for dtype selection in cmake flow - header generation

BujSet:
### Summary
This change introduces a new `cmake` command line option that can be enable via `-DEXECUTORCH_SELECT_DTYPE=ON.` For now , enabling this flag does *not* change the generated binary size. This change simply add the header file generation to the cmake build flow. Next steps will include automatically have the cmake build flow transplant the contents of the generated header file directly into the `kernels/portable/cpu/selective_build.h` file. This step must manually be done by a user to enable directed dtype pruning.

### Test plan
Changes were tested via `bash examples/selective_build/test_dtype_selective_build.sh cmake`. The resultant header file generated by the flow is now created and retained at `cmake-out/examples/selective_build/select_build_lib/selected_op_variants.h`.

### DIFF
--- a/examples/selective_build/CMakeLists.txt
+++ b/examples/selective_build/CMakeLists.txt
@@ -61,6 +61,10 @@ option(EXECUTORCH_SELECT_OPS_LIST "Register a list of ops, separated by comma"
 option(EXECUTORCH_SELECT_ALL_OPS
        "Whether to register all ops defined in portable kernel library." OFF
 )
+
+# Option to enable dtype selective build
+option(EXECUTORCH_SELECT_DTYPE "Enable dtype selection during build." OFF
+)
 # ------------------------------- OPTIONS END --------------------------------
 
 #
@@ -108,16 +112,30 @@ gen_selected_ops(
   "${EXECUTORCH_SELECT_OPS_LIST}"
   INCLUDE_ALL_OPS
   "${EXECUTORCH_SELECT_ALL_OPS}"
+  DTYPE_SELECT
+  "${EXECUTORCH_SELECT_DTYPE}"
 )
 
 generate_bindings_for_kernels(
-  LIB_NAME "select_build_lib" FUNCTIONS_YAML
-  ${EXECUTORCH_ROOT}/kernels/portable/functions.yaml CUSTOM_OPS_YAML
+  LIB_NAME
+  "select_build_lib"
+  FUNCTIONS_YAML
+  ${EXECUTORCH_ROOT}/kernels/portable/functions.yaml
+  CUSTOM_OPS_YAML
   "${_custom_ops_yaml}"
+  DTYPE_SELECT
+  "${EXECUTORCH_SELECT_DTYPE}"
 )
 
 gen_operators_lib(
-  LIB_NAME "select_build_lib" KERNEL_LIBS ${_kernel_lib} DEPS executorch_core
+  LIB_NAME
+  "select_build_lib"
+  KERNEL_LIBS
+  ${_kernel_lib}
+  DEPS
+  executorch_core
+  DTYPE_SELECT
+  "${EXECUTORCH_SELECT_DTYPE}"
 )
 
 list(TRANSFORM _executor_runner__srcs PREPEND "${EXECUTORCH_ROOT}/")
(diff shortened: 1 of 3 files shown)


examples/selective_build/CMakeLists.txt
```diff
@@ -108,16 +112,30 @@ gen_selected_ops(
   "${EXECUTORCH_SELECT_OPS_LIST}"
   INCLUDE_ALL_OPS
   "${EXECUTORCH_SELECT_ALL_OPS}"
+  DTYPE_SELECT
```
lucylq, line 115: nit: could we use `DTYPE_SELECTIVE_BUILD` to mirror the buck API here?

examples/selective_build/CMakeLists.txt
```diff
@@ -108,16 +112,30 @@ gen_selected_ops(
...
 generate_bindings_for_kernels(
-  LIB_NAME "select_build_lib" FUNCTIONS_YAML
-  ${EXECUTORCH_ROOT}/kernels/portable/functions.yaml CUSTOM_OPS_YAML
+  LIB_NAME
+  "select_build_lib"
+  FUNCTIONS_YAML
+  ${EXECUTORCH_ROOT}/kernels/portable/functions.yaml
+  CUSTOM_OPS_YAML
   "${_custom_ops_yaml}"
+  DTYPE_SELECT
```
lucylq, line 126: I think we only need this option in gen_selected_ops at the moment, so let's leave it out of the other APIs for now?

examples/selective_build/test_selective_build.sh
```diff
@@ -94,6 +94,7 @@ test_cmake_select_all_ops() {
     rm -rf ${build_dir}
     retry cmake -DCMAKE_BUILD_TYPE=Release \
             -DEXECUTORCH_SELECT_ALL_OPS=ON \
+            -DEXECUTORCH_SELECT_DTYPE=ON \
```
lucylq, line 97: Let's add these in once the E2E is working - currently, it won't enable dtype selective build so may be misleading

tools/cmake/Codegen.cmake
```diff
@@ -12,18 +12,20 @@
...
-  set(arg_names LIB_NAME OPS_SCHEMA_YAML ROOT_OPS INCLUDE_ALL_OPS)
+  set(arg_names LIB_NAME OPS_SCHEMA_YAML ROOT_OPS INCLUDE_ALL_OPS DTYPE_SELECT)
   cmake_parse_arguments(GEN "" "" "${arg_names}" ${ARGN})
 
   message(STATUS "Generating operator lib:")
   message(STATUS "  LIB_NAME: ${GEN_LIB_NAME}")
   message(STATUS "  OPS_SCHEMA_YAML: ${GEN_OPS_SCHEMA_YAML}")
   message(STATUS "  ROOT_OPS: ${GEN_ROOT_OPS}")
   message(STATUS "  INCLUDE_ALL_OPS: ${GEN_INCLUDE_ALL_OPS}")
+  message(STATUS "  DTYPE_SELECT: ${GEN_DTYPE_SELECT}")
```
jathu, line 23: Where is `GEN_DTYPE_SELECT` being defined?

tools/cmake/Codegen.cmake
```diff
@@ -78,6 +98,13 @@ function(generate_bindings_for_kernels)
   # By default selective build output is selected_operators.yaml
   set(_oplist_yaml ${_out_dir}/selected_operators.yaml)
 
+  # If dtype selective build is enable, force header file to be preserved
+  if(GEN_DTYPE_SELECT)
```
lucylq, line 102: We can leave this out for now, as it'll be preserved once we link the header into the portable library.

tools/cmake/Codegen.cmake
BujSet: `GEN_DTYPE_SELECT` is toggled `ON` or `OFF` via the command line switch `-DEXECUTORCH_SELECT_DTYPE=ON`. The option is defined in `examples/selective_build/CMakeLists.txt`.

examples/selective_build/CMakeLists.txt
BujSet: I was running into an issue where if the `selected_op_variants.h` file wasn't include in the sources list in the `gen_operators_lib` function, `cmake` was smart enough to delete it. The casing in that function conditionally adds it to the source list if the flag is set.

PR #11450, opened 2025-06-06
Benchmark HF optimum-executorch

guangy10:
Benchmark LLMs from [`optimum-executorch`](https://github.com/huggingface/optimum-executorch). With all the work recently happening in `optimum-executorch`, we are able to boost the out-of-the-box performance. Putting these models on benchmark infra to gather perf numbers and understand the remaining perf gaps between the in-house generated model via export_llama. 

We are able to do apple-to-apple comparison for CPU backend by introducing quant, custom SPDA, custom KV cache to native Hugging Face models in `optimum-executorch`: `hf_xnnpack_custom_spda_kv_cache_8da4w ` represents the recipe used by optimum-et, `et_xnnpack_custom_spda_kv_cache_8da4w ` is the counterpart for etLLM.

Here are the benchmark jobs in our infra:
 - Android: https://github.com/pytorch/executorch/actions/runs/15597347625
 - Apple: https://github.com/pytorch/executorch/actions/runs/15597340954
 
Fixed in benchmark apps (Android and iOS) to support using HF's tokenizers


guangy10:
@huydhn ~~In the apple's workflow, though I have specified the python version to be "3.11", it still install the python "3.13". Then when trying to pip install executorch, it ended with no package found. That's because we only publish with python 3.10, 3.11, and 3.12. https://github.com/pytorch/executorch/actions/runs/15500604843/job/43647388676#step:9:13372~~

Okay, it turns out that I need to run install with `${CONDA_RUN}`

guangy10:
~~Another issue. @huydhn Why isn't the artifact uploaded though it's zipped and moved to the dest dir successfully?
https://github.com/pytorch/executorch/actions/runs/15502766906/job/43653221066#step:20:1~~

This is fixed. It turns out the artifacts must be moved from the ET root dir to the artifacts-to-be-uploaded, mv from other dir under ET root will end up with the above issue.

.ci/scripts/gather_benchmark_configs.py
```diff
@@ -32,7 +32,8 @@
 BENCHMARK_CONFIGS = {
     "xplat": [
         "xnnpack_q8",
-        "hf_xnnpack_fp32",
+        "hf_xnnpack_custom_spda_kv_cache_8da4w",
+        "et_xnnpack_custom_spda_kv_cache_8da4w",
```
kimishpatel, line 36: whats the difference between these two?

.github/workflows/android-perf.yml
```diff
@@ -272,6 +272,21 @@ jobs:
...
+                      --model llama3_2 \
+                      --checkpoint "${DOWNLOADED_PATH}/consolidated.00.pth" \
+                      --params "${DOWNLOADED_PATH}/params.json" \
+                      -kv \
+                      --use_sdpa_with_kv_cache \
+                      -d fp32 \
+                      -X \
+                      --xnnpack-extended-ops \
+                      -qmode 8da4w -G 32 -E 8,0 \
+                      --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' \
```
kimishpatel, line 287: are these for llama_3_2?

.ci/scripts/gather_benchmark_configs.py
guangy10: Here is the explaination in the PR summary: `hf_xnnpack_custom_spda_kv_cache_8da4w`  represents the recipe used by optimum-et, `et_xnnpack_custom_spda_kv_cache_8da4w`  is the counterpart for etLLM.

.github/workflows/android-perf.yml
guangy10: @kimishpatel Yeah, for llama_3_2.

guangy10:
I'm seeing jobs hitting API Limits in AWS Device Farm. We lifted it for public AWS devices, @huydhn  do we need to do same and separately for new devices in private pools? https://github.com/pytorch/executorch/actions/runs/15504512047

guangy10:
Both benchmark jobs are finished successfully, but upon checking the benchmark_results.json, they are empty. https://github.com/pytorch/executorch/actions/runs/15540702059/job/43754028987 @kirklandsign any idea why? 

And all these runs: https://github.com/pytorch/executorch/actions/runs/15543294199. I would expect those to fail due to the issues in the tokenizer support in llama runner.

guangy10:
Disable passing tokenizer to the Android app will make it work for Qwen3 from both etLLM and optimum-et as showin here: https://github.com/pytorch/executorch/actions/runs/15546916104/job/43772530729. 
There must be issues when integrating HF tokenizer with the Android benchmark app. cc: @kirklandsign @jackzhxng 


The iOS app can not run the optimum-et generated PTE even after disabling the tokenizer. That is, run it as a regular PTE doesn't work as expected. https://github.com/pytorch/executorch/actions/runs/15540727931/job/43752973136. cc: @shoumikhin

guangy10:
@kimishpatel Here I can see the reported raw latency for Qwen3-0.6B from both etLLM and optimum-et: https://github.com/pytorch/executorch/actions/runs/15546916104/job/43772530729. The numbers are not making sense to me, it shows the optimum-et generated PTE is 5x faster on same Samsung Galaxy S22 5G. I suspect if it's because the etLLM model is not exported with the same config we're using for optimum-et.

.github/workflows/android-perf.yml
guangy10: @kimishpatel @jackzhxng can you confirm if this is the correct config we should use to export Qwen3 via etLLM path? The perf numbers reported here doesn't make sense to me https://github.com/pytorch/executorch/pull/11450#issuecomment-2957501752

kimishpatel: I dont know for qwen3. Can you compare the file sizes for the two? Also use `--xnnpack-extended-ops`

kimishpatel: Oh nevermind. you are using the option i mentioned

kimishpatel: This is not the command for qwen, right? That one is the one below?

kimishpatel: For this one I dont see hf counterpart

.github/workflows/apple-perf.yml
```diff
@@ -278,6 +279,21 @@ jobs:
...
+                --params "${DOWNLOADED_PATH}/params.json" \
+                -kv \
+                --use_sdpa_with_kv_cache \
+                -d fp32 \
+                -X \
+                --xnnpack-extended-ops \
+                -qmode 8da4w -G 32 -E 8,0 \
+                --metadata '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}' \
+                --output_name="${OUT_ET_MODEL_NAME}.pte"
+              ls -lh "${OUT_ET_MODEL_NAME}.pte"
```
kimishpatel, line 298: If possible please refactor these in a later PR. It is fewer lines to review for change like this

kimishpatel (commented):
Looks good to me but do the benchmark numbers get reported to a dashboard? It will be easier to track numbers that way.

Also lets validate the numbers before landing

extension/benchmark/apple/Benchmark/Tests/LLaMA/LLaMATests.mm
```diff
@@ -60,7 +60,7 @@ @implementation LLaMATests
 + (NSDictionary<NSString *, BOOL (^)(NSString *)> *)predicates {
   return @{
     @"model" : ^BOOL(NSString *filename){
-      return [filename hasSuffix:@".pte"] && [filename.lowercaseString containsString:@"llama"];
+      return [filename hasSuffix:@".pte"] && [filename.lowercaseString containsString:@"llm"];
```
guangy10, line 63: This is likely the issue that causing no TPS reported neither for Qwen model, neither for etLLM generated one nor for the optimum-et generated one. Reschedule a new run with this fix here: https://github.com/pytorch/executorch/actions/runs/15549765343

.github/workflows/android-perf.yml
guangy10: Yeah, it's for Qwen. I compared the command in this doc: https://github.com/pytorch/executorch/blob/9e9815c94f427f55a8240f01f7de3f84b9580618/examples/models/qwen3/README.md?plain=1#L19-L30
The only difference is `-G 32 -E 8,0`. I think we will need it because we did the same embedding quant for optimum-et models.

guangy10: > For this one I dont see hf counterpart

The HF counterpart should be just get via optimum-cli down at line 365

.github/workflows/apple-perf.yml
guangy10: Yeah, this should be refactored out to a script. .yml isn't the ideal place to handle the complex export configs like this.

guangy10:
> Looks good to me but do the benchmark numbers get reported to a dashboard? It will be easier to track numbers that way.
> 
> Also lets validate the numbers before landing

@kimishpatel Yeah, the numbers are reported to the dashboard, but as I mentioned it doesn't make sense the optimum-et generated model runs 5x faster than the etLLM one lol. Need to debug that, but here are the numbers on dash for Qwen3-0.6B (you can filter by "backends" to see the etLLM and optimum-et recipes): https://hud.pytorch.org/benchmark/llms?startTime=Tue%2C%2003%20Jun%202025%2003%3A35%3A48%20GMT&stopTime=Tue%2C%2010%20Jun%202025%2003%3A35%3A48%20GMT&granularity=day&lBranch=optimum_et_benchmark&lCommit=d12c6f6e7a834e8c10042ed6e22852c88232fd03&rBranch=optimum_et_benchmark&rCommit=d12c6f6e7a834e8c10042ed6e22852c88232fd03&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=Qwen%2FQwen3-0.6B&backendName=All%20Backends&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms

.github/workflows/android-perf.yml
guangy10: > Can you compare the file sizes for the two?

PTE size from etLLM: 470M. See this line in CI: https://github.com/pytorch/executorch/actions/runs/15549915868/job/43778415688#step:15:12848
PTE size from optimum-et: 506M. See this line in CI: https://github.com/pytorch/executorch/actions/runs/15549915868/job/43778415678#step:15:13591 

@kimishpatel The size makes sense to me. The performance doesn't, unless the PTE is not exported with correct config via etLLM, can't explain why optimum-et is 5x faster than etLLM. cc: @jackzhxng

extension/benchmark/apple/Benchmark/Tests/Tests.xcconfig
```diff
@@ -21,3 +21,5 @@ HEADER_SEARCH_PATHS = $(inherited) \
 
 LIBRARY_SEARCH_PATHS = $(inherited) \
     $(TEMP_DIR)/cmake/lib
+
+OTHER_CFLAGS = $(inherited) -DSUPPORT_REGEX_LOOKAHEAD=1
```
guangy10: @shoumikhin Does this look correct?

guangy10: Scheduled a new run here: https://github.com/pytorch/executorch/actions/runs/15574097109/job/43856956077?pr=11450

guangy10:
All optimum-et generated PTEs (for Qwen3, Gemma3, Llama3.2-1B, Olmo-1B, SmollmV2-135M) can be benchmarked on Android app, but it requires to disable tokenizer because it's not working, hence only the raw latency is reported, no TPS. 

Here is workflow showing running these models on public AWS Android phones: https://github.com/pytorch/executorch/actions/runs/15550573270. FYI, more accurate perf numbers will be collected from the private devices. We're using the public device for quick validation to avoid over-heating the private device as there is only one instance in the private pool atm.

We can view the perf on the [dashboard UI](https://hud.pytorch.org/benchmark/llms?startTime=Tue%2C%2003%20Jun%202025%2005%3A06%3A17%20GMT&stopTime=Tue%2C%2010%20Jun%202025%2005%3A06%3A17%20GMT&granularity=day&lBranch=optimum_et_benchmark&lCommit=b42ba369d03a30c4de7db8d1098b9b517c221e5f&rBranch=optimum_et_benchmark&rCommit=b42ba369d03a30c4de7db8d1098b9b517c221e5f&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=All%20Models&backendName=hf_xnnpack_custom_spda_kv_cache_8da4w&modeName=All%20Modes&dtypeName=All%20DType&deviceName=Samsung%20Galaxy%20S22%205G%20(Android%2013)&archName=All%20Platforms). cc: @kimishpatel @cbilgin 

We should prioritize this fix the bug in the app.

.github/workflows/apple-perf-private-device-experiment.yml
```diff
@@ -18,7 +18,7 @@ on:
         description: Models to be benchmarked
         required: false
         type: string
-        default: mv3,meta-llama/Llama-3.2-1B-Instruct-SpinQuant_INT4_EO8,meta-llama/Llama-3.2-1B-Instruct-QLORA_INT4_EO8
+        default: google/gemma-3-1b-it,Qwen/Qwen3-0.6B,HuggingFaceTB/SmolLM2-135M,meta-llama/Llama-3.2-1B,allenai/OLMo-1B-hf
```
guangy10, line 21: @huydhn FYI, I will have to offload the private device by temporarily removing some models, otherwise we would hit the API limits. Please add models and other existing runs when we you refactor the workflow to run as many as possible given the #instances in the private pool

.github/workflows/android-perf.yml
kimishpatel: is there a debug build vs release build difference for apps? You have to try and repro this locally. If it doesnt repro locally likely there is a benchmark infra issue

jackzhxng:
@guangy10 for the reported 5x higher numbers compared to etLLM, could this be comparing the results for the HF pte running non-LLM benchmark tests vs the etLLM pte. running LLM benchmark test? Just a guess.

extension/benchmark/apple/Benchmark/Tests/Tests.xcconfig
guangy10: @jackzhxng Is the cflag added correctly? Does it require additional linker flag? 
```
  E tokenizers:re2_regex.cpp:26] Failed to compile regex: ((?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\r\n\p{L}\p{N}]?\p{L}+|\p{N}| ?[^\s\p{L}\p{N}]+[\r\n]*|\s*[\r\n]+|\s+(?!\S)|\s+), error: invalid perl operator: (?!
  E tokenizers:regex.cpp:49] RE2 doesn't support lookahead patterns. Link with `regex_lookahead` to enable support.
  ```

guangy10:
> @guangy10 for the reported 5x higher numbers compared to etLLM, could this be comparing the results for the HF pte running non-LLM benchmark tests vs the etLLM pte. running LLM benchmark test? Just a guess.

Now I can TPS reported from the Android app. I manually pull the results and it turns seems the TPS is almost same between the etLLM generated Qwen3 and the optimum-et generated Qwen3:
Screenshot 2025-06-11 at 1 52 50 PM

guangy10:
For some reason doesn't see the results from the hf recipe got uploaded [here](https://github.com/pytorch/executorch/actions/runs/15595747531). Upon manual checking the results, same perf on iOS devices as well.
Screenshot 2025-06-11 at 5 41 52 PM

huydhn (approved):

guangy10:
One nit that will follow up in a separate PR is that we should allow bypass overriding installed et and deps when installing optimum-executorch. This is because in optimum-executorch the stack is running in dev mode with pinned nightly instead of source for stability, however, in ET, we would want to benchmark continuously to catch issues on trunk, so it should be using ET from source instead. I will need to address this discrepancy in both `optimum-executorch` and `executorch`

PR #11457, opened 2025-06-06
Enforce required cmake flags as explicit requirements

jathu:
### Summary

The build file currently turns on a bunch of implicitly dependent flags, perhaps unnecessarily. Let's make these dependencies explicit now.

For example, previously if `EXECUTORCH_BUILD_EXTENSION_TRAINING` was turned on, it would then turn on its own set of dependencies. This is good from a developer experience point of view because users would need to just turn on one flag. However, from a build point of view, it makes it harder to maintain and understand what flags are conflicting with one another. Now, if you want to build the training extension, you will have to explicitly turn on the required flags. However, with our build presets — most developers should not have to do to this!

I have created a follow up ticket to actually comb through the existing flags and trim the fat if necessary: https://github.com/pytorch/executorch/issues/11458

### Test plan

CI

cc @larryliu0820


.ci/scripts/build-qnn-sdk.sh
```diff
@@ -32,6 +32,8 @@ set_up_aot() {
       -DQNN_SDK_ROOT=${QNN_SDK_ROOT} \
       -DEXECUTORCH_BUILD_DEVTOOLS=ON \
       -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
+      -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
+      -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
```
larryliu0820, line 36: Why are these needed?

CMakeLists.txt
```diff
@@ -336,7 +310,7 @@ if(EXECUTORCH_USE_CPP_CODE_COVERAGE)
            " -fprofile-instr-generate -fcoverage-mapping"
     )
   else()
-    message(ERROR
+    message(FATAL_ERROR
```
larryliu0820, line 313: What's the difference?

tools/cmake/preset/default.cmake
```diff
@@ -276,9 +276,72 @@ define_overridable_option(
...
+    EXECUTORCH_BUILD_EXTENSION_DATA_LOADER
+    EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR
+)
+
+check_required_options_on(
+  IF_ON
+    EXECUTORCH_BUILD_KERNELS_CUSTOM
+  REQUIRES
+    EXECUTORCH_BUILD_KERNELS_OPTIMIZED
+)
```
larryliu0820, line 318: This doesn't seem right? Why custom kernels depend on optimized?

.ci/scripts/build-qnn-sdk.sh
jathu: Because EXECUTORCH_BUILD_EXTENSION_MODULE requires them to be on! Previously we were implicitly turning them on:

```cmake
if(EXECUTORCH_BUILD_EXTENSION_MODULE)
  set(EXECUTORCH_BUILD_EXTENSION_DATA_LOADER ON)
  set(EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR ON)
endif()
```

Where as now we fail if it's not explicitly turned on:

```cmake
check_required_options_on(
  IF_ON
    EXECUTORCH_BUILD_EXTENSION_MODULE
  REQUIRES
    EXECUTORCH_BUILD_EXTENSION_DATA_LOADER
    EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR
)
```

CMakeLists.txt
jathu: I guess this stop the configuration completely so the user knows that coverage is not supported

tools/cmake/preset/default.cmake
jathu: I am not sure, however this just assumes the logic that existed before:

```cmake
if(EXECUTORCH_BUILD_KERNELS_CUSTOM)
  set(EXECUTORCH_BUILD_KERNELS_OPTIMIZED ON)
endif()
```

larryliu0820: This is hilarious and yeah we should do a cleanup once preset is done

larryliu0820 (approved):

PR #11462, opened 2025-06-06
[4/N] Add backend options map

cccclai:
* #11758
* __->__ #11462
* #11391
* #11390
* #11389

This is to manage the backend <-> BackendOptions map. Users will create the bakcend options map, and ET runtime will read the backend name, and dispatch the list of backend options to each backend.
@exported-using-ghexport

Differential Revision: [D76149466](https://our.internmc.facebook.com/intern/diff/D76149466/)

Differential Revision: [D76149466](https://our.internmc.facebook.com/intern/diff/D76149466)


runtime/backend/backend_options.h
```diff
@@ -6,94 +6,92 @@
...
- 
- constexpr OptionKey<const char*> StrKey(const char* k) {
-   return OptionKey<const char*>(k);
- }
- 
- } // namespace runtime
- } // namespace executorch
- 
\ No newline at end of file
+#pragma once
```
kimishpatel: I am not sure whats happening in these diffs and why they show up as so many deleted lines and so many added lines

runtime/backend/backend_options.h
```diff
@@ -6,94 +6,92 @@
...
+          return Error::Ok;
+        }
+        return Error::InvalidArgument;
+      }
+    }
+    return Error::NotFound;
+  }
+  executorch::runtime::ArrayRef<BackendOption> view() const {
+    return executorch::runtime::ArrayRef<BackendOption>(options_, size_);
+  }
```
kimishpatel: is this method needed?

runtime/backend/backend_options_map.h
```diff
@@ -0,0 +1,74 @@
...
+namespace executorch {
+namespace runtime {
+
+struct Entry {
+  const char* backend_name;
+  ArrayRef<BackendOption> options;
+};
+
+template <size_t MaxBackends>
+class BackendOptionsMap {
```
kimishpatel: why do we need this?

runtime/backend/backend_options.h
cccclai: Ah it's the weird formatting. Let me update so it's easier to review. I've been manually running lintrunner to get rid of the weird formating

cccclai: Changes in backend_options.h is not needed...clean up the PR here

runtime/backend/backend_options_map.h
cccclai: This is to manage the mapping between backend_name and backend options. Like
{
    "XNNPACK": ["profile": true, "thread": 4],
    "QNN": ["power": "high"],
}

kimishpatel: Can you not just use set_option and get_option? We discussed this by allowing users to directly set backend options, without having to plumb it through methods/program etc., and I belive you did that? In that case do we still need this?

cccclai: backend options map is for setting multiple backends, users code will be like following
```
// users' code
BackendOptions backend_options = {
    {"CoreMLBackend", {{"power", "high"}}},
    {"QNNBackend", {{"backend", "CPU_AND_ANE"}}}
};

// Optional
update(backend_options);

module.load(file);
// Optional
update(backend_options);
    
module.execute(inputs);

```

kimishpatel (approved):
I am not 100% convinced we need this, but not strongly opposed to it. Leave it to the author

PR #11463, opened 2025-06-06
[5/N] Add update in method

cccclai:
* #11534
* #11533
* __->__ #11463
* #11462
* #11391
* #11390
* #11389

Expose the update API in method. This API will take BackendOptionMap as input and call backend_interface->update under the hood.

Differential Revision: [D76149467](https://our.internmc.facebook.com/intern/diff/D76149467/)


runtime/executor/test/method_update_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+ 
+ bool StubBackend::registered_ = false;
+ StubBackend StubBackend::singleton_;
+ 
+ class MethodUpdateTest : public ::testing::Test {
+  protected:
+   void load_program() {
+    // Since these tests cause ET_LOG to be called, the PAL must be initialized
+    // first.
+    executorch::runtime::runtime_init();
```
kimishpatel, line 140: This already happens in Setup too

runtime/executor/method.cpp
```diff
@@ -1512,6 +1513,26 @@ Error Method::experimental_step() {
   return step();
 }
 
+Error Method::update(executorch::runtime::ArrayRef<executorch::runtime::Entry> backend_option) {
+  for (const auto& entry : backend_option) {
+    const char* backend_name = entry.backend_name;
+    auto backend_options = entry.options;
+
+    auto backend_class = get_backend_class(backend_name);
```
kimishpatel, line 1521: does backend class give you a singleton instance of backend interface? If so why do we need to pipe it up via Method class. Just make something like `executorch::backend::set_options(BackendOptions&)` (I dont think BackendOptions constructor exist? but probably would be good to add)

cccclai: No it doesn't. I did only for the test backend I added, just so I can check if the config inside the backend is indeed changed in the test.

digantdesai: I agree. Allowing it to pass via Method or Module hints that it is private to a method which in fact it isn't.

PR #11469, opened 2025-06-09
Allow passing `install_executorch.py` args to `backends/openvino/scripts/openvino_build.sh`

KamilMatejuk:
### Summary
Executorch installation with OpenVINO backend by default accepts only `build_type` (either `--cpp_runtime` or `--enable_python`. However, internally it runs `install_executorch.sh` which COULD accept arguments:
* `--use-pt-pinned-commit`
* `--editable`

but there is no way to set them. This PR aims to pass those arguments from running `openvino_build.sh` into `install_executorch.sh`.

### Test plan
Tested commands:
```
bash backends/openvino/scripts/openvino_build.sh --enable_python
bash backends/openvino/scripts/openvino_build.sh --enable_python --use-pt-pinned-commit
```


JacobSzwejbka:
some merge conflicts. Im asking around for the openvino poc

PR #11475, opened 2025-06-09
Arm backend: Introduce TOSA aware context

per:
Introduce a 'TosaLoweringContext' to track the TOSA specification being used in the backend to be able to make decisions in the pipeline.


per:
Unrelated test failures.

zingo (approved):

PR #11479, opened 2025-06-09
[ET] enabling half dtype input for quantization

ahmtox:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #11553
* #11557
* #11483
* #11369
* #11556
* #11555
* #11554
* #11482
* #11481
* #11480
* #11368
* #11367
* #11366
* #11575
* #11365
* #11552
* __->__ #11479

Improving the cpu implementation op_quantize to support input half dtype and adding additional testing

Differential Revision: [D76053764](https://our.internmc.facebook.com/intern/diff/D76053764/)

PR #11367, opened 2025-06-04
[ET-VK][Ops] quantize_per_token.default test setup

ahmtox:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #11553
* #11557
* #11483
* #11369
* #11556
* #11555
* #11554
* #11482
* #11481
* #11480
* #11368
* __->__ #11367
* #11366
* #11575
* #11365
* #11552
* #11479

Creating quantize_per_token testing framework along with a reference implementation for testing

Differential Revision: [D75607854](https://our.internmc.facebook.com/intern/diff/D75607854/)

PR #11366, opened 2025-06-04
[ET-VK][Ops] quantize ops skeleton test framework

ahmtox:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #11553
* #11557
* #11483
* #11369
* #11556
* #11555
* #11554
* #11482
* #11481
* #11480
* #11368
* #11367
* __->__ #11366
* #11575
* #11365
* #11552
* #11479

Skeleton framework that is needed to build out the quantize_per_tensor and quantize_per_token operators based on cpu implementation

Differential Revision: [D75959066](https://our.internmc.facebook.com/intern/diff/D75959066/)

PR #11365, opened 2025-06-04
[ET-VK] double, short, and uint16 dtype runtime support

ahmtox:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #11553
* #11557
* #11483
* #11369
* #11556
* #11555
* #11554
* #11482
* #11481
* #11480
* #11368
* #11367
* #11366
* #11575
* __->__ #11365
* #11552
* #11479

Creating support for double, short, and uint16 for quantization ops. Registering the short keyword since theres already support.

Also changing the cpu implementation to support half

Differential Revision: [D75959063](https://our.internmc.facebook.com/intern/diff/D75959063/)

PR #11481, opened 2025-06-09
[ET-VK][Ops] dequantize_per_tensor.default test setup

ahmtox:
* #11553
* #11557
* #11483
* #11369
* #11556
* #11555
* #11554
* #11482
* __->__ #11481
* #11480
* #11368
* #11367
* #11366
* #11575
* #11365
* #11552
* #11479

Creating dequantize_per_tensor testing framework along with a reference implementation for testing

Differential Revision: [D76267054](https://our.internmc.facebook.com/intern/diff/D76267054/)


backends/vulkan/test/op_tests/dequantize_test.cpp
```diff
@@ -314,3 +314,337 @@ void check_dequantize_args(
...
+    std::cout << "reference:" << std::endl;
+    std::cout << reference_out << std::endl;
+    std::cout << "vulkan:" << std::endl;
+    std::cout << vk_out << std::endl;
+  }
+
+  ASSERT_TRUE(output_correct);
+}
+
+// Test cases for dequantize_per_tensor
```
manuelcandales, line 503: also add tests for int32 and int64

manuelcandales (changes requested):

backends/vulkan/test/op_tests/dequantize_test.cpp
ahmtox: theres no real support for int64 in the shader. right now dequantize matches the IO for quantize which include half, float and then uint8, int8, and int32 as common dtypes and to avoid generating excess variants. if int64 is a necessary variant I'll make those changes in a future diff.

there is a [diff](https://github.com/pytorch/executorch/pull/11553) that adds double support (though really just fp32 underneath)

i will add more tests in this diff.

SS-JIA: As @ahmtox has pointed out, to reduce binary size Vulkan kind of takes the opposite approach of the portable kernels, which means supporting the *minimum* dtypes necessary i.e. basically what we would expect to encounter in a "real" model.

manuelcandales (approved):

PR #11483, opened 2025-06-09
[ET-VK][Ops] dequantization op shaders and impl

ahmtox:
* #11553
* #11557
* __->__ #11483
* #11369

Creating the dequantize_per_tensor and dequantize_per_token logic shaders and impl which are linked with the testing framework.

Differential Revision: [D76267107](https://our.internmc.facebook.com/intern/diff/D76267107/)


backends/vulkan/test/op_tests/dequantize_test.cpp
```diff
@@ -749,14 +753,40 @@ TEST(
 
 TEST(
     VulkanDequantizePerTensorTest,
-    test_reference_dequantize_per_tensor_int16_to_float) {
```
manuelcandales: why did you delete this int16 test?

ahmtox: I deleted int16 because it's not included as a variant in the yaml. I'll remove this from the origin commit. I'm trying to only support variants that would seemingly be used, which I think would be uint8, int8, and int32

manuelcandales (approved):

PR #11484, opened 2025-06-09
Use nullptr to represent fallback kernels

rascani:
Summary:
`KernelKey`s can be constructed as either a "fallback" or "specialized" key. The "fallback" type uses the default constructor, and the "specialized" takes a specially formatted string. Internally, this was represented as a pointer to the optional key string and a bool for whether it is a fallback kernel. As it would not make sense to construct a "specialized" kernel without a key string, this diff eliminates the bool `is_fallback_` in favor of using `kernel_key_data_ == nullptr` to represent fallback kernels.

Each `KernelKey` is nested within the `Kernel` data structure, which makes up the list of registered kernels. As the default size of the registered_kernels array is 2000 kernel entries, this diff can reduce the size of the `registered_kernels` array by 8 KB.

This diff also changes the backing storage buffer for `registered_kernels_data` to ensure that there is enough space for each Kernel element in the array to be aligned according to `alignas(Kernel)`.

Differential Revision: D76201866

### DIFF
--- a/runtime/kernel/operator_registry.cpp
+++ b/runtime/kernel/operator_registry.cpp
@@ -35,9 +35,12 @@ constexpr uint32_t kMaxRegisteredKernels = kMaxOperators * kMaxKernelsPerOp;
 // require constructing them at init time. Since we don't care about the values
 // until we add each entry to the table, allocate static zeroed memory instead
 // and point the table at it.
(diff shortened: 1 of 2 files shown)


larryliu0820 (approved):
Thank you, this is a good change.

PR #11485, opened 2025-06-09
Avoid use-after-return caused by double move

tamird:
Summary:

Previously, `Result<T>`'s constructor used `std::move(val)` to initialize the
value. This resulted in an unnecessary extra move and destructor call on a
moved-from stack object, triggering an HWASAN stack tag mismatch when the
moved-from object was later destructed.

Replacing `std::move(val)` with `std::forward<T>(val)` avoids the extra move
while preserving correct semantics. This ensures only one move occurs and
avoids lifetime violations that can lead to tag mismatches under HWASAN.

Differential Revision: D76271359

### DIFF
--- a/runtime/core/result.h
+++ b/runtime/core/result.h
@@ -70,8 +70,9 @@ class Result final {
   /// Value copy constructor.
   /* implicit */ Result(const T& val) : value_(val), hasValue_(true) {}
 
-  /// Value move constructor.
-  /* implicit */ Result(T&& val) : value_(std::move(val)), hasValue_(true) {}
+  /// Value forwarding constructor.
+  /* implicit */ Result(T&& val)
+      : value_(std::forward<T>(val)), hasValue_(true) {}
 
   /// Result move constructor.
   /* implicit */ Result(Result&& rhs) noexcept : hasValue_(rhs.hasValue_) {


runtime/core/result.h
```diff
@@ -70,8 +70,9 @@ class Result final {
   /// Value copy constructor.
   /* implicit */ Result(const T& val) : value_(val), hasValue_(true) {}
 
-  /// Value move constructor.
-  /* implicit */ Result(T&& val) : value_(std::move(val)), hasValue_(true) {}
+  /// Value forwarding constructor.
+  /* implicit */ Result(T&& val)
+      : value_(std::forward<T>(val)), hasValue_(true) {}
```
JacobSzwejbka, line 75: Can you add a test that wouldve caught the previous bad behavior.

JacobSzwejbka (commented):
Thanks for the catch!

runtime/core/result.h
tamird: Sure. Where are the tests for this type?

lucylq: Can you try adding to this file: https://github.com/pytorch/executorch/blob/2dedc9e0e39047269b7762652b593c5c53883168/runtime/core/test/error_handling_test.cpp

swolchok (changes requested):
looks great, needs test, happy to accept with the test

tamird:
> looks great, needs test, happy to accept with the test

I tried writing a test, but I have failed to induce the problematic behavior. I think it shows up only in the presence of stack tagging as performed by hwasan.

runtime/core/result.h
swolchok, line 75: I misread what was going on here originally. This isn't a template function (though it is a member of a template class), so this really is an rvalue reference. I think that the original code is probably correct; std::forward is for arguments to template functions (https://en.cppreference.com/w/cpp/utility/forward.html).

tamird:
This turned out to be a red herring.

Issue #7441, opened 2024-12-26
Enable mypy typechecker tracker


mergennachin:
### 🚀 The feature, motivation and pitch

Currently there's no python typechecker enabled on OSS. Let's enable mypy gradually

### Alternatives

Another alternative is to use pyre. But we want to be consistent with pytorch/pytorch which uses mypy.

### Additional context

Progress: 

- [ ] exir
- [ ] backends
- [ ] examples
- [ ] extension
- [ ] devtools
- [x] test https://github.com/pytorch/executorch/pull/7497
- [x] util https://github.com/pytorch/executorch/pull/7496
- [x] scripts https://github.com/pytorch/executorch/pull/7442
- [x] profiler https://github.com/pytorch/executorch/pull/7494
- [x] docs https://github.com/pytorch/executorch/pull/7493
- [x] codegen  https://github.com/pytorch/executorch/pull/7493

mergennachin:
https://github.com/pytorch/executorch/pull/7776

robert-kalmar:
backend/nxp - Issue here https://github.com/pytorch/executorch/issues/13518

PR #11489, opened 2025-06-09
Typecheck 25% of exir directory

mergennachin:
Used Claude Code mainly. Used https://github.com/pytorch/executorch/issues/7441 as a starting context. Asked for an overall plan and it came up with this https://gist.github.com/mergennachin/dee5de3ae244d68baa34f46ed3afabe6

Started with simpler files and simple fixes first. Just hit the throttling limit for now. But this is a good first step as is

Test Plan: CI


exir/delegate.py
```diff
@@ -151,7 +153,7 @@ def is_lowered_module(obj: Any) -> bool:
     def get_lowered_module_name(
         root: torch.nn.Module,
         # pyre-ignore: Undefined or invalid type [11]: Annotation `LoweredBackendModule` is not defined as a type.
-        lowered_module: LOWERED_BACKEND_MODULE_TYPE,  # noqa
```
cccclai, line 154: Is this correct?

mergennachin: Yeah, it was a bug before. 

LOWERED_BACKEND_MODULE_TYPE was not a python type, but just a string

https://github.com/pytorch/executorch/blob/2dedc9e0e39047269b7762652b593c5c53883168/exir/delegate.py#L42

.lintrunner.toml
```diff
@@ -315,6 +315,56 @@ include_patterns = [
     # 'examples/**/*.py',
     'examples/openvino/**/*.py',
     # 'exir/**/*.py',
+    # Phase 1: Start with simplest exir files (Batch 1)
```
lolpack, line 318: These should be added to the `mypy.ini` as well?

exir/delegate.py
cccclai: I remember I explicitly add it to decouple backend_api.py and delegate.py...it's supposed to be this type https://github.com/pytorch/executorch/blob/fad25ad78a8b3a18e2fb3e85bc02e96c7011dee4/exir/lowered_backend_module.py#L54

.lintrunner.toml
mergennachin: It's already there, i think 

https://github.com/pytorch/executorch/blob/main/.mypy.ini#L18

exir/delegate.py
mergennachin: Why are we having all these pyre-ignores all over for LoweredBackendModule?

cccclai: My guess for the pyre-ignores is likely due to pyre can't recognize string types. However we have to do that because delegate.py and lowered_backend_module.py is cross dependency (not really cross dependency, but delegate.py needs the forward declaration from lowered_backend_module.py. This is supported in c++, but not PyThon)

PR #11493, opened 2025-06-09
Configure conflicting ARM baremetal cmake flags

jathu:
### Summary

Similar to https://github.com/pytorch/executorch/pull/11457, but now we also want to warn against unsupported/conflicting flags. For example, when using `EXECUTORCH_BUILD_ARM_BAREMETAL`, we cannot enable certain flags.

### Test plan

CI

```
$ cmake -DEXECUTORCH_BUILD_ARM_BAREMETAL=ON --preset macos

CMake Error at tools/cmake/common/preset.cmake:142 (message):
  Both 'EXECUTORCH_BUILD_ARM_BAREMETAL' and
  'EXECUTORCH_BUILD_EXTENSION_DATA_LOADER' can't be ON
  tools/cmake/preset/default.cmake:352 (check_conflicting_options_on)
  CMakeLists.txt:87 (include)
```

cc @larryliu0820 @digantdesai @freddan80 @per @zingo @oscarandersson8218


zingo (approved):

zingo:
Thanks, if this combination don't work we should try to fix it :)

swolchok:
>   Both 'EXECUTORCH_BUILD_ARM_BAREMETAL' and
  'EXECUTORCH_BUILD_EXTENSION_DATA_LOADER' can't be ON
  
  The ARM size test builds with EXECUTORCH_BUILD_ARM_BAREMETAL, and size_test.cpp requires FileDataLoader (and, currently, incorrectly doesn't enable it). Is it OK to remove this constraint?

jathu:
@swolchok I'm not sure, you can try it. I tried retaining the existing behavior where `EXECUTORCH_BUILD_EXTENSION_DATA_LOADER` was turned off if baremetal was on

PR #11496, opened 2025-06-09
Dont require method loading for meta meta in module api

JacobSzwejbka:
Summary: Program load is still required but this is very cheap. Method load is not needed.

Differential Revision: D76278890


shoumikhin (approved):

shoumikhin:
Lgtm as long as all tests pass.
Maybe add CI/flow label to trigger everything.
And iirc there are some docs and doc comments to amend a bit

PR #11504, opened 2025-06-09
Extracting dtype information from model PTE in OSS

### LINKED ISSUE #11434: Enable selective build model API in OSS
Currently, gen_oplist pulls in an internal-only script to scrape models. Write this in python in OSS.

BujSet:
### Summary
Previously, exporting kernel dtypes was accomplished via `gen_oplist.py` but was only supported in internal-only scripts. This functionality of  scraping models for dtypes to make them available for the dtype selective build is now implemented for OSS and internal, removing dependance on internal only scripts.

 `Fixes #11434` : [11434](https://github.com/pytorch/executorch/issues/11434)

### Test plan
Manually tested on the `add_mul.pte` and `mv3.pte` example models. After building executorch, ran `python3 -m codegen.tools.gen_oplist --model_file_path add_mul.pte --output_path add_mul_model.txt`. The results `add_mul_model.txt` file is a YAML file that now includes dtype info in the kernel metadata section, as shown below:

```
build_features: []
custom_classes: []
et_kernel_metadata:
  aten::add:
  - v1/6;0,1|6;0,1|6;0,1|6;0,1
  aten::mm:
  - v1/6;0,1|6;0,1|6;0,1|6;0,1
include_all_non_op_selectives: false
include_all_operators: false
kernel_metadata: {}
operators:
  aten::add:
    debug_info:
    - add_mul.pte
    include_all_overloads: false
    is_root_operator: true
    is_used_for_training: true
  aten::mm:
    debug_info:
    - add_mul.pte
    include_all_overloads: false
    is_root_operator: true
    is_used_for_training: true
```

Comparable results were seen on the `mve3.pte` model (generated via `python -m examples.portable.scripts.export --model_name="mv3"`).

CI flow is also updated to enable unittests for `gen_oplist.py` for the selective build process.

### DIFF
--- a/codegen/tools/gen_oplist.py
+++ b/codegen/tools/gen_oplist.py
@@ -15,26 +15,10 @@ import yaml
 
 try:
     from executorch.codegen.parse import strip_et_fields
-    from executorch.exir._serialize import _deserialize_pte_binary
-    from executorch.exir.schema import (
-        EValue,
-        KernelCall,
-        OptionalTensorList,
-        Tensor,
-        TensorList
-    )
 except ImportError:
     # If we build from source, executorch.codegen is not available.
     # We can use relative import instead.
     from ..parse import strip_et_fields
-    from ..exir._serialize import _deserialize_pte_binary
-    from ..exir.schema import (
-        EValue,
-        KernelCall,
-        OptionalTensorList,
-        Tensor,
-        TensorList
-    )
 
 from torchgen.gen import LineLoader, parse_native_yaml_struct
 from torchgen.selective_build.operator import SelectiveBuildOperator
@@ -105,6 +89,7 @@ def _get_operators(model_file: str) -> List[str]:
     with open(model_file, "rb") as f:
         buf = f.read()
 
+    from executorch.exir._serialize import _deserialize_pte_binary
     model = _deserialize_pte_binary(buf)
     operators = []
     for execution_plan in model.execution_plan:
@@ -114,12 +99,6 @@ def _get_operators(model_file: str) -> List[str]:
     return operators
 
 
-def _get_dtypes_from_non_list(evalue: EValue):
-    kernel_key = ""
-    if isinstance(evalue, Tensor):
-        dim_order = ",".join(map(str, evalue.dim_order))
-        kernel_key += f"{evalue.scalar_type};{dim_order}|"
-    return kernel_key
 
 
 def _get_kernel_metadata_for_model(model_file: str) -> Dict[str, List[str]]:
@@ -133,6 +112,20 @@ def _get_kernel_metadata_for_model(model_file: str) -> Dict[str, List[str]]:
         buf = f.read()
     op_kernel_key_list: Dict[str, List[str]] = {}
 
+    from executorch.exir._serialize import _deserialize_pte_binary
+    from executorch.exir.schema import (
+        EValue,
+        KernelCall,
+        OptionalTensorList,
+        Tensor,
+        TensorList
+    )
+    def _get_dtypes_from_non_list(evalue: EValue):
+        kernel_key = ""
+        if isinstance(evalue, Tensor):
+            dim_order = ",".join(map(str, evalue.dim_order))
+            kernel_key += f"{evalue.scalar_type};{dim_order}|"
+        return kernel_key
     model = _deserialize_pte_binary(buf)
     for execution_plan in model.execution_plan:
         for chain in execution_plan.chains:
--- a/codegen/tools/gen_oplist.py
+++ b/codegen/tools/gen_oplist.py
@@ -9,7 +9,7 @@ import json
 import os
 import sys
 from enum import IntEnum
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Set
 
 import yaml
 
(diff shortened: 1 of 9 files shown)


codegen/tools/gen_oplist.py
```diff
@@ -86,59 +86,88 @@ class KernelType(IntEnum):
...
+                        KernelType.OPTIONAL_TENSOR_LIST,
+                    ]:
+                        dim_order = ",".join(map(str, io_metadata.dim_order))
+                        kernel_key += f"{io_metadata.dtype};{dim_order}|"
+                op_kernel_key_list[op_name].append(kernel_key[:-1])
+
+    except ModuleNotFoundError:
+        from executorch.exir._serialize import _deserialize_pte_binary
+        model = _deserialize_pte_binary(buf)
+        for execPlan in model.execution_plan:
```
lucylq: nit: can we use snake case to match the rest of the file, maybe `execution_plan` instead of `execPlan`.

codegen/tools/gen_oplist.py
```diff
@@ -86,59 +86,88 @@ class KernelType(IntEnum):
...
+            for chain in execPlan.chains:
+                for instr in chain.instructions:
+                    op_name = execPlan.operators[instr.instr_args.op_index].name
+                    if op_name not in op_kernel_key_list:
+                        op_kernel_key_list[op_name] = []
+                    version = "v1"
+                    kernel_key = version + "/"
+                    # TODO what happens when tensors have different types withina single kernel/ is that even allowed?
```
lucylq: If I understand the question correctly, it should be allowed. eg. one op_add kernel may take in floats, and another may take ints. They'll appear as another KernelMetadata entry. 

I think we can also have op_add take a float and an int tensor, and I guess some sort of casting is done? cc @manuelcandales  

This looks good, you can probably remove the TODO.

codegen/tools/gen_oplist.py
```diff
@@ -86,59 +86,88 @@ class KernelType(IntEnum):
...
+                    for tensor_arg in instr.instr_args.args:
+                        val = execPlan.values[tensor_arg].val
+
+                        # TODO is there a better way to do this?
+                        if type(val).__name__ == "Tensor":
```
lucylq: How about: `if isinstance(val, Tensor):`

Can you also add TensorList, OptionalTensorList?

You can import these types from [schema.py](https://github.com/pytorch/executorch/blob/a7a9130922f75814c1deef3bddaa9c33d1ed6967/exir/schema.py)

lucylq:
Thanks for adding this! Can we check against mv3.pte, which has some more operators than addmul?

`python -m examples.portable.scripts.export --model_name="mv3"`

Result from internal: P1836375803

codegen/tools/gen_oplist.py
BujSet: Yea, my question was more about within a single kernel, can types be mixed. From the execution plan for the `add_mul` model data, it seems like this is true:

```
values
    EValue(val=Tensor(scalar_type=<ScalarType.FLOAT: 6>, storage_offset=0, sizes=[2, 2], dim_order=[0, 1], requires_grad=False, layout=0, data_buffer_idx=0, allocation_info=AllocationDetails(memory_id=1, memory_offset_low=48, memory_offset_high=0), shape_dynamism=<TensorShapeDynamism.STATIC: 0>, extra_tensor_info=None))
    EValue(val=Tensor(scalar_type=<ScalarType.FLOAT: 6>, storage_offset=0, sizes=[2, 2], dim_order=[0, 1], requires_grad=False, layout=0, data_buffer_idx=0, allocation_info=AllocationDetails(memory_id=1, memory_offset_low=0, memory_offset_high=0), shape_dynamism=<TensorShapeDynamism.STATIC: 0>, extra_tensor_info=None))
    EValue(val=Tensor(scalar_type=<ScalarType.FLOAT: 6>, storage_offset=0, sizes=[2, 2], dim_order=[0, 1], requires_grad=False, layout=0, data_buffer_idx=0, allocation_info=AllocationDetails(memory_id=1, memory_offset_low=32, memory_offset_high=0), shape_dynamism=<TensorShapeDynamism.STATIC: 0>, extra_tensor_info=None))
    EValue(val=Tensor(scalar_type=<ScalarType.FLOAT: 6>, storage_offset=0, sizes=[2, 2], dim_order=[0, 1], requires_grad=False, layout=0, data_buffer_idx=0, allocation_info=AllocationDetails(memory_id=1, memory_offset_low=16, memory_offset_high=0), shape_dynamism=<TensorShapeDynamism.STATIC: 0>, extra_tensor_info=None))
    EValue(val=Tensor(scalar_type=<ScalarType.FLOAT: 6>, storage_offset=0, sizes=[2, 2], dim_order=[0, 1], requires_grad=False, layout=0, data_buffer_idx=0, allocation_info=AllocationDetails(memory_id=1, memory_offset_low=0, memory_offset_high=0), shape_dynamism=<TensorShapeDynamism.STATIC: 0>, extra_tensor_info=None))
    EValue(val=Int(int_val=1))]
operators
    Operator(name='aten::mm', overload='out')
    Operator(name='aten::add', overload='out')
instructions
    Instruction(instr_args=KernelCall(op_index=0, args=[0, 1, 3, 3]))
    Instruction(instr_args=KernelCall(op_index=1, args=[3, 2, 5, 4, 4]))
inputs
    [0, 1, 2]
outputs
    [4]
```

It appears that the second kernel call is to the add operator, where arguments 0, 1, 3, 4, and 5 are `Float` tensors, but argument 2 is an `Int` (maybe a multiplicative scalar?). The if guard ignores non-tensor types, so I think you're right that some sort of casting happens elsewhere.

tools/cmake/Codegen.cmake
```diff
@@ -12,14 +12,22 @@
 include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
 
 function(gen_selected_ops)
+<<<<<<< Updated upstream
   set(arg_names LIB_NAME OPS_SCHEMA_YAML ROOT_OPS INCLUDE_ALL_OPS)
+=======
```
lucylq: have to resolve the merge conflict?

BujSet: Yea, think I had a git snafu. This shouldn't be a part of this changeset.

codegen/tools/targets.bzl
```diff
@@ -28,7 +28,7 @@ def define_common_targets(is_fbcode = False):
         main_module = "executorch.codegen.tools.gen_oplist",
         deps = [
             ":gen_oplist_lib",
-        ],
+        ] + [] if runtime.is_oss else ["//exir:lib"], # TODO This is a hack to run on fbcode. These targets depends on exir which is only available in fbcode.
```
larryliu0820: I think it should be //executorch/exir:lib

PR #11506, opened 2025-06-09
[xnn update prep] deprecate sdpa

mcr229:
Differential Revision: [D77265464](https://our.internmc.facebook.com/intern/diff/D77265464)


mcr229:
* #11856
* #11507
* __->__ #11506

backends/xnnpack/runtime/XNNCompiler.cpp
```diff
@@ -2097,7 +2061,6 @@ DefineNodeFunc getDefineNodeFunc(fb_xnnpack::XNodeUnion nodeType) {
     _DEFINE(Concatenate4)
     _DEFINE(Concatenate5)
     _DEFINE(StaticSlice)
-    _DEFINE(ScaledDotProductAttention)
```
digantdesai, line 1791: Not updating schema for marking deprecated? `XNNScaledDotProductAttention`

backends/xnnpack/operators/op_sdpa.py
```diff
@@ -1,111 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: Should we continue partitioning this and lower it as decomposed?
Else this will be a BC breaking change.

digantdesai (changes requested):

backends/xnnpack/runtime/XNNCompiler.cpp
mcr229: so XNNPACK removed the operator from their codebase, so for next update we need to delete. I can mark the operator in the schema as deprecated though.

backends/xnnpack/operators/op_sdpa.py
mcr229: So ever since my new partitioner, SDPA has not been delegated. I assumed that no models in production or since h ave really used any sdpa implementation (also because i heard our sdpa is slow). so I believe this is safe to remove. I can import it internally to make sure

mcr229:
@mcr229 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D77265464).

mcr229:
@mcr229 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D77265464).

backends/xnnpack/operators/op_sdpa.py
digantdesai: What I was thinking was we should handle decomp of SDPA inside XNNPACK AoT so that we don't regress for current perf. Lowering pieces is OK but can cause unexpected perf drops with slight changes.
And I guess when the constraint fails it will get decomp and we will lower it in pieces anyway, right?

digantdesai (approved):
Stamping to unblock rebase, but we should revisit this under the larger task of replacing custom SDPA op with XNNPACK's in the future.

kernels/portable/cpu/util/dtype_util.cpp
```diff
@@ -27,8 +27,6 @@ bool check_tensor_dtype(
       return executorch::runtime::tensor_is_floating_type(t);
     case SupportedTensorDtypes::INTB:
       return executorch::runtime::tensor_is_integral_type(t, true);
-    case SupportedTensorDtypes::BOOL:
```
swolchok, line 30: why is this change here? bad rebase?

PR #11510, opened 2025-06-10
Bump tokenizers

jackzhxng:


guangy10:
What has been fixed? If it's fixing the reported tokenizer issue, can we at least rerun the app with previously failed model+tokenizer in the CI?

kimishpatel (approved):

jackzhxng:
@guangy10 some recent tokenizer changes broke the cmake for the regex lookahead, so this fixes it

mergennachin (approved):

Issue #11307, opened 2025-06-03
Complie issue cause lost #include


barbecacov:
### 🐛 Describe the bug

sh executorch/backends/qualcomm/scripts/build.sh;
when complie executorch/examples/qualcomm/oss_scripts/llama/;
it reports "no template named 'unordered_map' in namespace 'std' in https://github.com/pytorch/executorch/blob/main/examples/qualcomm/oss_scripts/llama/runner/rpc_mem.h#L59"

if i add 

#include <unordered_map> 

it is ok;

please fix the bug, thank you

### Versions

please fix the bug, thank you

mcr229:
looks like this is coming from qualcomm. @cccclai  mind taking a look here?

cccclai:
Thank you for the report, send the patch in https://github.com/pytorch/executorch/pull/11515

PR #11515, opened 2025-06-10
Fix missing import during build

cccclai:
Summary: Reported in https://github.com/pytorch/executorch/issues/11307 that the unordered_map fails to compile in due to https://github.com/pytorch/executorch/blob/main/examples/qualcomm/oss_scripts/llama/runner/rpc_mem.h#L59

Reviewed By: mcr229

Differential Revision: D76278059


shewu-quic (approved):
Thanks for your help! :)

mergennachin (approved):

PR #11499, opened 2025-06-09
Refactor FlatTensor

lucylq:
Summary:
This diff is the stack at D73209986 combined.

Landing as one diff/PR as it contains BC-breaking changes throughout the stack.

Differential Revision: D76285590

PR #11517, opened 2025-06-10
Arm backend: Add missing ops to annotator

oscarandersson8218:
- Adds missing operators to annotator. These are needed to improve numerical precision for llama and deit_tiny.
- Extends 5D support to runtime by permuting input/output to/from channels_last.
- Adds unit tests for operators with 5D tensors.


mansnils (approved):

zingo:
Test fails in test-arm-cortex-m-size-test seem unrelated and seem to happen after this got merged: 
https://github.com/pytorch/executorch/pull/11499

zingo:
> Test fails in test-arm-cortex-m-size-test seem unrelated and seem to happen after this got merged: #11499

Well it was another error and not the same, sorry, we are looking into this.

PR #11519, opened 2025-06-10
Arm backend: Replace asserts with exceptions in quantizer module

Sebastian-Larsson:
Replace all assert statements in the quantizer module with appropriate exceptions (ValueError/TypeError) so misuse raises clear errors instead of AssertionError.


zingo (approved):

zingo:
Test fail in cortex-m is unrelated

PR #11520, opened 2025-06-10
Arm backend: Bump cortex-m size test threshold

zingo:
Size to runtime has grown and this started to fail after some code was improved/added in the "Refactor FlatTensor" change: https://github.com/pytorch/executorch/pull/11499 and we got ~50 Bytes over the limit. 

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -262,7 +262,7 @@ jobs:
         output=$(ls -la ${elf})
         arr=($output)
         size=${arr[4]}
-        threshold="103068" # ~100KiB
+        threshold="103268" # ~100KiB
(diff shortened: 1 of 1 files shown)


zingo:
https://github.com/pytorch/executorch/pull/11499 seem to have bumped the bin size slightly for the cortex-m build, tjis bumps the test threshold (assuming the bigger size was ok)

mergennachin (approved):

digantdesai (approved):
While I understand this change, but it would be nice in the future if we track what caused this to fail.

zingo:
Good idea There are some code added in https://github.com/pytorch/executorch/pull/11499 when we started to fail, Ill add a pointer to the commit msg for future tracking.

PR #11532, opened 2025-06-10
Replace debug handle with `from_node` to trace operator transformation

Gasoonjia:
Summary:
X-link: https://github.com/pytorch/ao/pull/2339

This diff replace the debug handle  with `from_node` infrastructure, which is a first class citizen in exported program and used to trace the node-level transformation. For simplify the progress, we are trying to reuse the debug handle infrastructure by generating debug handle from from_node info via hasing.

After this change user no longer need to invoke `generate_numeric_debug_handle` for debugging. Also the original pipeline will still work under current scenario.

Reviewed By: jerryzh168

Differential Revision: D76168997

### DIFF
--- a/backends/xnnpack/test/quantizer/test_pt2e_quantization.py
+++ b/backends/xnnpack/test/quantizer/test_pt2e_quantization.py
@@ -7,7 +7,7 @@
 # pyre-unsafe
 
 from collections import Counter
-from typing import Dict, Tuple
+from typing import Tuple
 
 import torch
 from executorch.backends.xnnpack.quantizer.xnnpack_quantizer import (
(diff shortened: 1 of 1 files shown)


backends/xnnpack/test/quantizer/test_pt2e_quantization.py
```diff
@@ -723,75 +723,101 @@ def test_save_load(self) -> None:
...
-            )
+    def _extract_debug_handles_with_prev_decomp_op(self, model) -> dict[str, int]:
+        prev_decomp_op_to_debug_handle_map: dict[str, int] = {}
 
-        bfs_trace_with_node_process(model, _assert_node_has_debug_handle)
+        def _extract_debug_handles_with_prev_decomp_op_from_node(node):
+            nonlocal prev_decomp_op_to_debug_handle_map
+            if FROM_NODE_KEY in node.meta:
+                prev_decomp_op = str(node.meta.get("nn_module_stack"))
+                debug_handle = _generate_debug_handle_from_node(node)
```
digantdesai, line 757: is `from_node` a new thing?

digantdesai (approved):
This is mainly updating the test, so ok with merging this.

backends/xnnpack/test/quantizer/test_pt2e_quantization.py
Gasoonjia: it is introduced after torch 2.6 and we will use that to trace the node transformation from torch.export to torchao to edge.

PR #11534, opened 2025-06-10
[7/N] Add sugar syntax for module.update

cccclai:
* __->__ #11534
* #11533
* #11463
* #11462
* #11391
* #11390
* #11389

The update API in method is supposed to be portable, but we can make it more user friendly for the update API in module. Add a bit sugar syntax in module to improve UX. Such that user can update backend option in module like following:
```
Module module(stub_model_path_);
int new_num_threads = 4;

const auto update_result = module.update("forward",
      {
        {"StubBackend", {{IntKey("NumberOfThreads"), new_num_threads}}
      },
  );
```

Differential Revision: [D76242292](https://our.internmc.facebook.com/intern/diff/D76242292/)


extension/module/dynamic_backend_options_map.h
```diff
@@ -0,0 +1,43 @@
+
+#pragma once
+
+#include <vector>
+#include <string>
+#include <initializer_list>
+#include <executorch/runtime/backend/backend_options.h>
```
shoumikhin:
```suggestion
#include <executorch/runtime/backend/backend_options.h>

#include <vector>
#include <string>
#include <initializer_list>
```

extension/module/module.h
```diff
@@ -15,6 +15,7 @@
 #include <vector>
 
 #include <executorch/runtime/executor/program.h>
+#include <executorch/extension/module/dynamic_backend_options_map.h>
```
shoumikhin:
```suggestion
#include <executorch/extension/module/dynamic_backend_options_map.h>
#include <executorch/runtime/executor/program.h>
```

extension/module/module.h
```diff
@@ -483,6 +484,12 @@ class Module {
     return update("forward", backend_options);
   }
 
+  ET_EXPERIMENTAL ET_NODISCARD inline runtime::Error update(
```
shoumikhin: Any comments on what this API does?

extension/module/dynamic_backend_options_map.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace executorch {
+namespace runtime {
+
+class DynamicBackendOptionsMap {
```
shoumikhin: Can we somehow use only standard containers for this map in order to skip this extra class? E.g. `std::vector<std::pair<const char*, ArrayRef<BackendOption>>>` or something?

cccclai: That's an option, not sure how simple it is compared with update API exposed in https://github.com/pytorch/executorch/pull/11533 If that one is simple enough for users, we may not need to add this PR. What do you think?

shoumikhin: If I understand this PR correctly, it just makes it easier for devs to use the other update() API so that they don't need to keep the backend options preallocated and pass refs to them, but employ standard containers to keep the memory alive?

What I propose here is to skip any custom classes to keep the backend options and just use the standard containers directly. Then inside this new convenient update() override do the due conversion and call the original update().

extension/module/module.h
cccclai: Update the comments

extension/module/dynamic_backend_options_map.h
cccclai: Ah I see, update the PR based on the suggestion

PR #11537, opened 2025-06-10
Use safe_numerics util from PyTorch

### LINKED ISSUE #11370: Import and extend c10/util/safe_numerics.h

Pytorch code referenced in title: https://github.com/pytorch/pytorch/blob/main/c10/util/safe_numerics.h

In https://github.com/pytorch/executorch/pull/11217 we added a check for an overflow in nybtes. Rather then performing division every check itd be nicer if we had the faster checks in safe_numerics.h. There are probably other places in ET we should be checking for overflows like this.

safe_numerics.h currently only has int64 and uint64 so we'd need to extend it for size_t. 

Could just write our own size_t form first and then resolve merging with pytorch/pytorch later.

cc @swolchok for suggesting it

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

aaron-ang:
Fixes #11370

### Summary
We import `safe_numerics.h` from PyTorch and extend `mul_overflows` for `size_t`. Then, we apply it in `calculate_nbytes` to perform faster overflow checks.

### Test plan
Use existing tests since we are doing a drop-in replacement for overflow checks.

### DIFF
--- a/kernels/portable/cpu/util/normalization_ops_util.cpp
+++ b/kernels/portable/cpu/util/normalization_ops_util.cpp
@@ -14,6 +14,10 @@
 namespace torch {
 namespace executor {
 
+#ifdef __clang__
+#pragma clang diagnostic ignore "-Wvla-extension"
+#endif
+
 using Tensor = executorch::aten::Tensor;
 
 bool check_batch_norm_args(
--- a/runtime/core/portable_type/c10/c10/util/safe_numerics.h
+++ b/runtime/core/portable_type/c10/c10/util/safe_numerics.h
@@ -1,0 +1,90 @@
+#pragma once
+#include <c10/macros/Macros.h>
+
+#include <cstdint>
+
+// GCC has __builtin_mul_overflow from before it supported __has_builtin
+#ifdef _MSC_VER
+#define C10_HAS_BUILTIN_OVERFLOW() (0)
+#include <c10/util/llvmMathExtras.h>
+#include <intrin.h>
+#else
+#define C10_HAS_BUILTIN_OVERFLOW() (1)
+#endif
+
+namespace c10 {
+
+C10_ALWAYS_INLINE bool add_overflows(uint64_t a, uint64_t b, uint64_t* out) {
+#if C10_HAS_BUILTIN_OVERFLOW()
+  return __builtin_add_overflow(a, b, out);
+#else
+  unsigned long long tmp;
+#if defined(_M_IX86) || defined(_M_X64)
+  auto carry = _addcarry_u64(0, a, b, &tmp);
+#else
+  tmp = a + b;
+  unsigned long long vector = (a & b) ^ ((a ^ b) & ~tmp);
+  auto carry = vector >> 63;
+#endif
+  *out = tmp;
+  return carry;
+#endif
+}
+
+C10_ALWAYS_INLINE bool mul_overflows(uint64_t a, uint64_t b, uint64_t* out) {
+#if C10_HAS_BUILTIN_OVERFLOW()
+  return __builtin_mul_overflow(a, b, out);
+#else
+  *out = a * b;
+  // This test isnt exact, but avoids doing integer division
+  return (
+      (c10::llvm::countLeadingZeros(a) + c10::llvm::countLeadingZeros(b)) < 64);
+#endif
+}
+
+C10_ALWAYS_INLINE bool mul_overflows(int64_t a, int64_t b, int64_t* out) {
+#if C10_HAS_BUILTIN_OVERFLOW()
+  return __builtin_mul_overflow(a, b, out);
+#else
+  volatile int64_t tmp = a * b;
+  *out = tmp;
+  if (a == 0 || b == 0) {
+    return false;
+  }
+  return !(a == tmp / b);
+#endif
+}
+
+template <typename It>
+bool safe_multiplies_u64(It first, It last, uint64_t* out) {
+#if C10_HAS_BUILTIN_OVERFLOW()
+  uint64_t prod = 1;
+  bool overflow = false;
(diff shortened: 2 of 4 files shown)


runtime/core/portable_type/c10/c10/util/safe_numerics.h
```diff
@@ -0,0 +1,112 @@
...
+  if (a == 0 || b == 0) {
+    return false;
+  }
+  return !(a == tmp / b);
+#endif
+}
+
+C10_ALWAYS_INLINE bool mul_overflows(size_t a, size_t b, size_t* out) {
+#if C10_HAS_BUILTIN_OVERFLOW()
+  return __builtin_mul_overflow(a, b, out);
```
JacobSzwejbka, line 73: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

Oh it just supports it natively lol

runtime/executor/method_meta.cpp
```diff
@@ -56,23 +56,20 @@ size_t calculate_nbytes(
     Span<const int32_t> sizes,
     executorch::aten::ScalarType scalar_type) {
   size_t n = 1;
-  size_t prev_n = 1;
   for (size_t i = 0; i < sizes.size(); i++) {
-    prev_n = n;
-    n *= sizes[i];
-    // Check for overflow
-    ET_CHECK(sizes[i] == 0 || n / sizes[i] == prev_n);
+    size_t next_n;
```
JacobSzwejbka, line 60: nit: declare outside loop

runtime/core/portable_type/c10/c10/util/safe_numerics.h
JacobSzwejbka: Is this function the only new part and the rest copy pasted from pytorch/pytorch?

aaron-ang: I extended the `add_overflows` and `mul_overflows` methods.

kernels/portable/cpu/util/normalization_ops_util.cpp
```diff
@@ -14,6 +14,10 @@
 namespace torch {
 namespace executor {
 
+#ifdef __clang__
+#pragma clang diagnostic ignore "-Wvla-extension"
```
JacobSzwejbka, line 18: whys this needed?

JacobSzwejbka:
Oh on 
> check-c10-sync / check-c10-sync (pull_request)

We track that anything under c10 mirrors exactly (this is a temporary thing iiuc until we can introduce a hard dep on pytorch/pytorch directly). If you want you could try submitting this to pytorch/pytorch instead and then mirror the file from there, or you can just move this to any other util spot that looks reasonable in ET to dodge this issue, and then we will deal with upstreaming it later. 

Moving it will also probably help with the c10 namespace issues you were having.

kernels/portable/cpu/util/normalization_ops_util.cpp
aaron-ang: I faced this error when compiling (`./install_executorch.sh`) with Clang on MacOS.

jathu: @aaron-ang can you paste the error?

aaron-ang: @jathu My bad. I faced this error not when installing Executorch, but when running tests locally using `sh test/build_size_test.sh` as mentioned in https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#testing.
```sh
executorch/kernels/portable/cpu/util/normalization_ops_util.cpp:110:37: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
  110 |   executorch::aten::SizesType shape[ndim];
      |                                     ^~~~
executorch/kernels/portable/cpu/util/normalization_ops_util.cpp:110:37: note: read of non-const variable 'ndim' is not allowed in a constant expression
executorch/kernels/portable/cpu/util/normalization_ops_util.cpp:88:10: note: declared here
   88 |   size_t ndim = normalized_shape.size();
      |          ^
1 error generated.
gmake[2]: *** [kernels/portable/cpu/util/CMakeFiles/kernels_util_all_deps.dir/build.make:219: kernels/portable/cpu/util/CMakeFiles/kernels_util_all_deps.dir/normalization_ops_util.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:750: kernels/portable/cpu/util/CMakeFiles/kernels_util_all_deps.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```
My syntax is also wrong: it should be `ignored` instead of `ignore`, but surprisingly compiles fine.

aaron-ang:
hi @huydhn why is this issue closed? I need to update the code from PyTorch upstream. It is different from this branch.
Edit: I have incorporated the changes. Could you please reopen the PR?

huydhn:
What!  This makes no sense.  I have no intention of closing this PR, as I didn't even know about this before you pointed it out.  Something doesn't feel right

huydhn:
And I couldn't reopen the PR, the option is grayed out :(

Screenshot 2025-06-30 at 18 31 54

huydhn:
While I'm checking, maybe it's easier to recreate this in a different PR

malfet:
> What! This makes no sense. I have no intention of closing this PR, as I didn't even know about this before you pointed it out. Something doesn't feel right

https://github.com/huydhn/pytorch/commit/3dda80e990121eaf156014fffe6e2a4602c8b195 claimes to fix this issue, therefore github closed it (issues and pulls are the same entities in GitHub parlance)

PR #11545, opened 2025-06-11
Improve performance of _to_dim_order_copy

swolchok:
We had a problem similar to the one motivating BroadcastIndexesRange: we were doing full coordinate recalcuation on every iteration of a loop. Thankfully, it is straightforward to use BroadcastIndexesRange (and the new support_noncontiguous_tensors option, which I realized should be renamed) to fix this.


swolchok:
* __->__ #11545

manuelcandales (approved):

PR #11547, opened 2025-06-11
fix platform inconsistencies

h-friederich:
Summary: When underlying dependencies have platform restrictions we need to propagate them to dependencts.

Differential Revision: D76402988


cccclai (approved):
Thanks

Issue #10447, opened 2025-04-24
Extend `PyBundledModule` with `extension.BundledModule`


zhenyan-zhang-meta:
# Context

This issue is a step of https://github.com/pytorch/executorch/discussions/9638 and a sequel of https://github.com/pytorch/executorch/issues/10375.

In https://github.com/pytorch/executorch/discussions/9638, we want to have `extension.Module` as the single source of implementation in `pybindings`, which means that `pybindings.PyModule` should use `extension.Module` rather than its own `pybindings.Module`.

# Proposal

Now that we have `extension.BundledModule` ready, we want to test it out by having our existing `PyBundledModule` to extend it, and let `verify_result_with_bundled_expected_output` to use it, so that we can test out the whole thing with https://github.com/pytorch/executorch/blob/fb45e19055a92d2a91a4d4b7008e135232cbb14b/devtools/bundled_program/test/test_end2end.py

mergennachin:
Current implementation uses BundledModule; the requested end-to-end test passes without skips.

PR #11565, opened 2025-06-11
[ExecuTorch][#10447] Extend `PyBundledModule` with `extension.BundledModule`

Gasoonjia:
* __->__ #11565
* #11566

# Context

This issue is a step of https://github.com/pytorch/executorch/discussions/9638.

In https://github.com/pytorch/executorch/discussions/9638, we want to have `extension.Module` as the single source of implementation in `pybindings`, which means that `pybindings.PyModule` should use `extension.Module` rather than its own `pybindings.Module`.

# Proposal

Now that we have `extension.BundledModule` ready, we want to test it out by having our existing `PyBundledModule` to extend it, and let `verify_result_with_bundled_expected_output` to use it, so that we can test out the whole thing with https://github.com/pytorch/executorch/blob/fb45e19055a92d2a91a4d4b7008e135232cbb14b/devtools/bundled_program/test/test_end2end.py

Differential Revision: [D73564127](https://our.internmc.facebook.com/intern/diff/D73564127/)


lucylq (approved):
Thanks for adding this!

PR #11566, opened 2025-06-11
solve duplicate dependency issues to unblock pybinding backend migration

Gasoonjia:
* #11565
* __->__ #11566

This diff creates different namespace for Module and BundledModule under different mode (aten or portable) to hacky solve the duplicated symbol issues in D73564127 since the their are too many complex dependency issues to solve, and lots of users are asking for pybinding backend migration.

In the future this impl should be replaced by shim-like layers.

Differential Revision: [D76405701](https://our.internmc.facebook.com/intern/diff/D76405701/)


larryliu0820 (approved):

shoumikhin:
I vaguely remember a discussion with @swolchok et al. about this hack, did we eventually decide to go with it? Will it be temporary workaround with a timeline to revert when a proper solution is introduced?

swolchok:
> I vaguely remember a discussion with @swolchok et al. about this hack, did we eventually decide to go with it? Will it be temporary workaround with a timeline to revert when a proper solution is introduced?

yeah I don't think we should be doing this and I also vaguely recall talking about it previously, but I don't have a reference to the specific previous discussion.

Gasoonjia:
@shoumikhin  and @swolchok yes we do have discussion previously and this is sort of a hacky way. But in our codebase there're too much existing targets that depends on both aten and portable mode at the same time and some weird dependency situation we can not solve. So maybe currently we can land it first and then working on some shim-like layer to solve the dependency issue completely. 

cc. @larryliu0820  @cccclai @lucylq

PR #11568, opened 2025-06-11
braodcast in op_le

kimishpatel:
* __->__ #11568

Differential Revision: [D76456398](https://our.internmc.facebook.com/intern/diff/D76456398/)

### DIFF
--- a/kernels/optimized/cpu/op_le.cpp
+++ b/kernels/optimized/cpu/op_le.cpp
@@ -6,12 +6,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <executorch/kernels/optimized/cpu/binary_ops.h>
 #include <executorch/kernels/optimized/vec/functional.h>
 #include <executorch/kernels/optimized/vec/vec.h>
 #include <executorch/kernels/portable/cpu/scalar_utils.h>
 #include <executorch/kernels/portable/cpu/util/broadcast_util.h>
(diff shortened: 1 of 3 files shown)


manuelcandales (approved):
I think we need to add a check at the top of this op that the dim orders must be the same, because I don't think any of this support different dim orders. But this was a preexisting issue, so, won't hold you on that.

kernels/optimized/cpu/op_le.cpp
```diff
@@ -79,52 +81,41 @@ Tensor& opt_le_tensor_out(
...
-              a.const_data_ptr<CTYPE>(),
-              b.const_data_ptr<CTYPE>(),
-              a.numel());
-        });
+  // Check for optimized broadcast paths
+  auto selected_optimized_path = select_optimized_path(a, b, out);
+  printf("selected_optimized_path: %d\n", static_cast<int>(selected_optimized_path));
+  if (selected_optimized_path == ElementwiseOptimizedPath::kTreatAs1d) {
+    // Resize for dynamic shape
+    auto error = resize_tensor(out, a.sizes());
```
manuelcandales, line 89: this is introducing the same bug you are fixing for op_mul in https://github.com/pytorch/executorch/pull/11590

manuelcandales (changes requested):

manuelcandales (approved):

PR #11582, opened 2025-06-11
[do not land] selective build size test

lucylq:
```
bash examples/selective_build/test_selective_build.sh cmake

check file at cmake-out/examples/selective_build/selective_build_test
```
Results (all are stripped)
```
-rwxr-xr-x 1 lfq users 2730352 Jun 11 16:39 all_ops_s
-rwxr-xr-x 1 lfq users 260464 Jun 11 16:18 sb_add_s
-rwxr-xr-x 1 lfq users 231792 Jun 11 16:20 sb_add_float_s
-rwxr-xr-x 1 lfq users 219504 Jun 11 16:28 sb_add_none_s
```

### DIFF
--- a/examples/selective_build/test_selective_build.sh
+++ b/examples/selective_build/test_selective_build.sh
@@ -110,20 +110,17 @@ test_cmake_select_all_ops() {
 }
 
 test_cmake_select_ops_in_list() {
-    echo "Exporting MobilenetV2"
-    ${PYTHON_EXECUTABLE} -m examples.portable.scripts.export --model_name="mv2"
+    # echo "Exporting MobilenetV2"
+    # ${PYTHON_EXECUTABLE} -m examples.portable.scripts.export --model_name="mv2"
 
     local example_dir=examples/selective_build
     local build_dir=cmake-out/${example_dir}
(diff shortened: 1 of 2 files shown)


BujSet:
# Results from Binary Size Tests

Experiments with the model API introduced in PR [11601](https://github.com/pytorch/executorch/pull/11601). Binary size reports were collected on the `selective_build_test` binary after running strip:

| Model | Binary size before dtype selection | Binary size after dtype selection |
| -------- | :---: | :---:|
| MobilenetV2 | 587KB | 467KB |
| MobilenetV3 | 795KB | 519KB |
| add_mul | 339KB | 279KB |
| add | 335 KB | 271KB |
| softmax | 255KB | 255KB |

# Notes:

It seems that the exported model do not always contain all the information needed to include all the dtypes.

## Example Failures:

* MobilenetV2 requires `double` for the `hardtanh` operator, however, the extracted yaml file does not relay this information.
* MobilenetV3 requires `long` for the `clamp` operator, however, the extracted yaml file does not relay this information. Additionally manually including this dtype in `kernels/portable/cpu/selective_build.h` does not seem to work (program crashes reporting that operator `clamp` is missing dtype 6 (long).
* Not sure why `softmax` shows no improvement.

PR #11590, opened 2025-06-12
Fix bug in optimized mul's broadcast handling

kimishpatel:
Summary:
When we have two tensors that match exactly in size but not in dims, the output resize does not work correctly in cases like this a[6] b[1, 1, 6] -> out [1, 1, 6] but current code tried to resize using a.sizes()

Test Plan:
tests added

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/kernels/optimized/cpu/op_mul.cpp
+++ b/kernels/optimized/cpu/op_mul.cpp
@@ -111,14 +111,11 @@ Tensor& opt_mul_out(
 
   auto selected_optimized_path = select_optimized_path(a, b, out);
   if (selected_optimized_path == ElementwiseOptimizedPath::kTreatAs1d) {
-    // Resize for dynamic shape
-    auto error = resize_tensor(out, a.sizes());
-    ET_KERNEL_CHECK_MSG(
+    ET_KERNEL_CHECK(
         ctx,
-        error == Error::Ok,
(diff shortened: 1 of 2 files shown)


kernels/optimized/cpu/op_mul.cpp
```diff
@@ -112,7 +112,12 @@ Tensor& opt_mul_out(
   auto selected_optimized_path = select_optimized_path(a, b, out);
   if (selected_optimized_path == ElementwiseOptimizedPath::kTreatAs1d) {
     // Resize for dynamic shape
-    auto error = resize_tensor(out, a.sizes());
+    // broadcasting rule dictate that the tensor whose dimensionality matches
+    // the output is the one that we should pick.
+    // e.g. a[6] and b[1, 1, 6] would imply out = [1, 1, 6], but if we resize
+    // via a we will crash
```
manuelcandales: this is a good catch, but I think the original error was to use resize_tensor() in the first place when this op was updated to support broadcasting. The right function to use is `resize_to_broadcast_target_size(a, b, out)`, which is already being used inside some of the if/else. This should be moved out and to the top, before the case work begins.
by the way, this same error is now present in your other PR for `le`

manuelcandales (approved):

manuelcandales (approved):

PR #11596, opened 2025-06-12
Refactor XNNPACK tester to extract delegate-independent tester classes

GregoryComer:
### Summary
Refactor the XNNPACK tester to split out reusable base components from XNNPACK-specific parts. I've relocated the base classes to backends/test/harness.

I've kept the tester structure pretty much unchanged, except for replacing stage names with an enum.

It looks like Arm tests are directly importing for XNNPACK's tester currently. Ideally, we'll want to refactor to have their own stage implementations, but I've left that as a follow-up to minimize changes for the initial refactor.

### Test plan
CI


GregoryComer:
Note to reviewers: let me know if you have strong feelings on string vs enum keys for stages. I originally switched it to a enum to try to get away from the local stage registration in XNNPACK. But I could put it back to use string keys everywhere. CC @mcr229

cccclai:
I'm proposing moving this code to https://github.com/pytorch/executorch/tree/main/exir/backend under `executorch/exir/backend/test_framework`, such that the strucutre of the backend is indeed a list of backends. I've brought it up to @digantdesai about this before.

backends/test/harness/tester.py
```diff
@@ -0,0 +1,453 @@
...
+            ],
+            StageType.TO_EDGE: [
+                StageType.PARTITION,
+                StageType.RUN_PASSES,
+            ],
+            StageType.RUN_PASSES: [
+                StageType.PARTITION,
+                StageType.TO_EDGE_TRANSFORM_AND_LOWER,
+            ],
+            # TODO Make this Stage optional
```
mcr229, line 65: can you just do this by making TO_EDGE either go to TO_EXECUTORCH or PARTITION. Also I know its a bit of a headache now, but I wash oping to change the name of partition to TO_BACKEND. If its not too much of a hassle to do. otherwise it's a very minor nit

mcr229: I'm guessing majority of this is copied from the XNNPACK Tester, so looks fine to me.

mcr229 (approved):
This is a very nice refactor Gregory. Thanks for doing it

GregoryComer:
> I'm proposing moving this code to https://github.com/pytorch/executorch/tree/main/exir/backend under `executorch/exir/backend/test_framework`, such that the strucutre of the backend is indeed a list of backends. I've brought it up to @digantdesai about this before.

Sounds good. I don't have a particularly strong opinion on this. I'm happy to discuss more. If we move it, I'll put out another PR to do this since I have a pretty large stack on top of this.

backends/test/harness/tester.py
GregoryComer: > can you just do this by making TO_EDGE either go to TO_EXECUTORCH or PARTITION. Also I know its a bit of a headache now, but I wash oping to change the name of partition to TO_BACKEND. If its not too much of a hassle to do. otherwise it's a very minor nit

Makes sense. I'll put out a follow-up PR to rename partition to to_backend. The other TODO was moved from the previous location, but seems easy to fix, so I'll do it as well.

PR #11601, opened 2025-06-12
Support for Model API for Selective Build in OSS

BujSet:
### Summary
Added support for parsing the selected operators and dtypes directly from an exported model to a `selected_operators.yaml` file. Previously, operators had to be specified in string list or in a schema yaml file for the selective build process. Now, a model can be specified at build time (e.g. `-DEXECUTORCH_SELECT_OPS_FROM_MODEL="./mv2.pte"`) to parse and select operators.

### Test plan
Added a test for this in `examples/selective_build/test_selective_build.sh`. When run with `bash examples/selective_build/test_selective_build.sh cmake` a resultant `selected_operators.yaml` file is created at `cmake-out/examples/selective_build/select_build_lib/selected_operators.yaml` that conatins the selected operators as well as their dtypes.

### DIFF
--- a/examples/selective_build/test_selective_build.sh
+++ b/examples/selective_build/test_selective_build.sh
@@ -166,6 +166,7 @@ test_cmake_select_ops_in_model() {
     ${PYTHON_EXECUTABLE} -m examples.portable.scripts.export --model_name="mv2"
     local example_dir=examples/selective_build
     local build_dir=cmake-out/${example_dir}
+    rm -rf ${build_dir}
     retry cmake -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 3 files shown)


examples/selective_build/test_selective_build.sh
```diff
@@ -161,6 +161,28 @@ test_cmake_select_ops_in_yaml() {
     rm "./custom_ops_1.pte"
```
lucylq, line 161: Can you try adding this at the end of each cmake test?

```
echo "Cleaning build directory ${build_dir}/"
rm -rf ${build_dir}/
```

And see if that resolves the test errors

lucylq (approved):
Looks great, thanks for adding this!

PR #11603, opened 2025-06-12
Define ET_HAS_EXCEPTIONS macro

swolchok:
* #11604
* __->__ #11603

To support passing ET_USE_PYTORCH_HEADERS only when exceptions are enabled.

Differential Revision: [D76470039](https://our.internmc.facebook.com/intern/diff/D76470039/)


runtime/platform/compiler.h
```diff
@@ -171,6 +171,14 @@
 using ssize_t = ptrdiff_t;
 #endif
 
+#ifdef __EXCEPTIONS
+#define ET_HAS_EXCEPTIONS 1
+#elif defined(_HAS_EXCEPTIONS) && _HAS_EXCEPTIONS
```
JacobSzwejbka, line 176: just curious where do __EXCEPTIONS and _HAS_EXCEPTIONS come from?

JacobSzwejbka (approved):

runtime/platform/compiler.h
JacobSzwejbka: Oh its in the test plan internally. From the compilers.

PR #11610, opened 2025-06-12
Update default models to be benchmarked continuously

guangy10:
### Summary
- Promoted Qwen3-0.6B to be the default as it's not only small enough to run quickly but also covers most of advanced changes in both etLLM and [optimum-executorch](https://github.com/huggingface/optimum-executorch).
- Removed tinyllama as nobody would care about its perf. We shouldn't use device farm for correctness testing
- ~~Added google/gemma-3-1b-it to apple perf (private)~~

### DIFF
--- a/.github/workflows/android-perf-private-device-experiment.yml
+++ b/.github/workflows/android-perf-private-device-experiment.yml
@@ -18,7 +18,7 @@ on:
         description: Models to be benchmarked
         required: false
         type: string
-        default: google/gemma-3-1b-it,Qwen/Qwen3-0.6B,HuggingFaceTB/SmolLM2-135M,meta-llama/Llama-3.2-1B,allenai/OLMo-1B-hf
+        default: Qwen/Qwen3-0.6B
       devices:
         description: Target devices to run benchmark
         required: false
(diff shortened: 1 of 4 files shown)


.github/workflows/android-perf-private-device-experiment.yml
```diff
@@ -57,6 +57,6 @@ jobs:
       id-token: write
       contents: read
     with:
-      models: ${{ inputs.models || 'Qwen/Qwen3-0.6B' }}
+      models: ${{ inputs.models || 'google/gemma-3-1b-it,Qwen/Qwen3-0.6B,HuggingFaceTB/SmolLM2-135M,meta-llama/Llama-3.2-1B,allenai/OLMo-1B-hf' }}
```
guangy10: @huydhn I only want to run one model (`Qwen/Qwen3-0.6B`) on PR/trunk, but run the all listed models here in the scheduled job, how can I do it?

guangy10:
Have to delist the `google/gemma-3-1b-it` from continuous benchmarking. 

@kimishpatel @cbilgin  FYI, Still can't run `google/gemma-3-1b-it` on-device due to this bug https://github.com/pytorch/executorch/issues/11618. The issue wasn't caught on Android side because of another bug in the Android benchmark app here https://github.com/pytorch/executorch/issues/11620.

shoumikhin (approved):

guangy10:
Can't merge, try rebase with no code changes

PR #11619, opened 2025-06-12
Support for Header File Generation in DTYPE Selective Build in OSS

BujSet:
### Summary

Part of the process for enabling the dtype selective build flow requires a header file to be generated that includes information about kernel operators and their dtypes. This change adds support for generating this header file when operators are specified. This functionality is enabled via the command line flag: `-DEXECUTORCH_DTYPE_SELECTIVE_BUILD=ON`. When enabled, the resultant `selected_op_variants.h` file will be generated.

### DIFF
--- a/examples/selective_build/CMakeLists.txt
+++ b/examples/selective_build/CMakeLists.txt
@@ -66,9 +66,8 @@ option(EXECUTORCH_SELECT_ALL_OPS
 option(EXECUTORCH_SELECT_OPS_FROM_MODEL "Enable op selection from pte during build." OFF
 )
 
-# Option to enable dtype header selective build process
+# Option to enable dtype selective build. Note: must be using selective build model API.
 option(EXECUTORCH_DTYPE_SELECTIVE_BUILD "Enable dtype selective build." OFF
->>>>>>> 6ae80f260 (Support for header file generation with dtype)
 )
 # ------------------------------- OPTIONS END --------------------------------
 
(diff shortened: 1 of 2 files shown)


examples/selective_build/CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ option(EXECUTORCH_SELECT_OPS_LIST "Register a list of ops, separated by comma"
 option(EXECUTORCH_SELECT_ALL_OPS
        "Whether to register all ops defined in portable kernel library." OFF
 )
+
+# Option to enable dtype header selective build process
```
lucylq: nit: `Option to enable dtype selective build. Note: must be using selective build model API.`

tools/cmake/Codegen.cmake
```diff
@@ -53,6 +58,23 @@ function(gen_selected_ops)
...
+    set(_gen_opvariant_command "${PYTHON_EXECUTABLE}" -m codegen.tools.gen_selected_op_variants
+                          --yaml-file=${_oplist_yaml}
+                          --output-dir=${CMAKE_CURRENT_BINARY_DIR}/${GEN_LIB_NAME}/
+    )
+    message("Command - ${_gen_opvariant_command}")
+    add_custom_command(
+      COMMENT "Generating selected_op_variants.h for ${GEN_LIB_NAME}"
+      OUTPUT ${_opvariant_h}
+      COMMAND ${_gen_opvariant_command}
+      DEPENDS ${_optlist_yaml} ${_codegen_tools_srcs}
```
lucylq: nit: oplist_yaml

lucylq (approved):
Looks great, thanks for adding this! We can remove the warning message after the e2e.

tools/cmake/Codegen.cmake
BujSet: Sorry, what is the change that you wanted here? `${_oplist_yaml}` is an output of the previous command, which, when using the model API, will include the dtypes. So I believe it should be listed in the `DEPENDS`, as the command will need it to generate the header file, right?

lucylq: Oh I think it's a typo, `optlist_yaml` --> `oplist_yaml`

PR #11624, opened 2025-06-12
Mapping between runtime and aot intermediate outputs

Juntian777:
Summary:
This PR introduces a function map_runtime_aot_intermediate_outputs that maps runtime intermediate outputs to AOT intermediate outputs by identifying overlapping(between aot and runtime) debug handles and combining them into a single key that could handle below various mapping scenarios.

1. No Overlaps: There are no overlapping debug handles between AOT and runtime outputs.
2. 1-to-1 Mapping: A straightforward mapping where one AOT debug handle corresponds directly to one runtime debug handle.
3. 1-to-N Mapping: A single AOT debug handle maps to multiple runtime debug handles.
4. N-to-1 Mapping: Multiple AOT debug handles map to a single runtime debug handle.
5. N-to-N Mapping: More intricate scenarios where multiple AOT and runtime debug handles form a chain of overlaps

In all cases where multiple debug handles are involved (N-to-1, 1-to-N, or N-to-N), the function merges these into a single combined debug handle and retains only the last intermediate output for each mapping.

In order to handle all cases, the code first does some pre-processing to the input, then constructs a graph of nodes representing debug handles and outputs, identifies connected components using DFS, and creates mappings for overlapping components, merging debug handles and retaining the last output.

This function will be used later in the Inspector Numerical Comparator class to create the mapping.

Differential Revision: D76442807

### DIFF
--- a/devtools/inspector/_inspector_utils.py
+++ b/devtools/inspector/_inspector_utils.py
@@ -8,6 +8,7 @@
 
 import math
 import sys
+from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Dict, IO, List, Mapping, Optional, Tuple, TypeAlias, Union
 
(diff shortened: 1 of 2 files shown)


Gasoonjia (approved):
LGTM

PR #11638, opened 2025-06-13
Arm backend: Test and track resnet18 and resnet50 models in CI

zingo:
### Test plan
This adds tests for resnet18 and resnet50 for Ethos-U


zingo:
broken tests are broken on main also

perheld (approved):
LGTM

PR #11653, opened 2025-06-13
Install examples and domain libraries as last (optional) step

mergennachin:
Install requirements is installing so many things that's not necessary for core executorch runtime. For example, it's not necessary to install packages that is used in examples. 

Ideally, developers should just run to install core ET. 

```
install_requirement.sh
pip install . -e
```

### DIFF
--- a/install_executorch.py
+++ b/install_executorch.py
@@ -16,6 +16,7 @@ import sys
 from contextlib import contextmanager
 
 from install_requirements import (
+    install_optional_example_requirements,
     install_requirements,
     python_is_compatible,
     TORCH_NIGHTLY_URL,
@@ -157,6 +158,13 @@ def _parse_args() -> argparse.Namespace:
         "picked up without rebuilding the wheel. Extension libraries will be "
         "installed inside the source tree.",
     )
+    parser.add_argument(
+        "--minimal",
+        "-m",
+        action="store_true",
+        help="Only installs necessary dependencies for core executorch and skips "
+        " packages necessary for running example scripts.",
+    )
     return parser.parse_args()
 
 
@@ -182,9 +190,13 @@ def main(args):
     # This option is used in CI to make sure that PyTorch build from the pinned commit
     # is used instead of nightly. CI jobs wouldn't be able to catch regression from the
     # latest PT commit otherwise
-    install_requirements(use_pytorch_nightly=not args.use_pt_pinned_commit)
-    os.execvp(
-        sys.executable,
+    use_pytorch_nightly = not args.use_pt_pinned_commit
+
+    # Step 1: Install core dependencies first
+    install_requirements(use_pytorch_nightly)
+
+    # Step 2: Install core package
+    cmd = (
         [
             sys.executable,
             "-m",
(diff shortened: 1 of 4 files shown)


requirements-dev.txt
```diff
@@ -1,7 +1,11 @@
...
+
 cmake>=3.19, <4.0.0  # For building binary targets in the wheel.
 pip>=23  # For building the pip package.
 pyyaml  # Imported by the kernel codegen tools.
 setuptools>=63  # For building the pip package contents.
 tomli  # Imported by extract_sources.py when using python < 3.11.
 wheel  # For building the pip package archive.
 zstd  # Imported by resolve_buck.py.
+lintrunner==0.12.7
+lintrunner-adapters==0.12.4
```
jathu, line 11: Since you're doing this, why not just consolidate all of [requirements-lintrunner.txt](https://github.com/pytorch/executorch/blob/main/requirements-lintrunner.txt) to this and deleting that? I never understood why we split that — linting should be required for development

install_executorch.py
```diff
@@ -199,6 +205,30 @@ def main(args):
...
 
+    # Step 3: Extra (optional) packages that is only useful for running examples.
+    DOMAIN_LIBRARIES = [
+        (
+            f"torchvision==0.23.0.{NIGHTLY_VERSION}"
+            if use_pytorch_nightly
+            else "torchvision"
+        ),
+        f"torchaudio==2.8.0.{NIGHTLY_VERSION}" if use_pytorch_nightly else "torchaudio",
+    ]
```
jathu: What is the benefit of doing this here vs. in `install_requirements`? When developing the wheel people will run `install_executorch.py` multiple times, so now this will cause pip to attempt to install these examples multiple time (even though it won't cause it's already installed)

requirements-dev.txt
mergennachin: requirements-lintrunner.txt is only for CI, users don't need to install. `lintrunner init` should just install those. ideally, we can move this file to .ci/ folder.

But yeah, I moved the lintrunner and lintrunner-adapters to here. So that people don't need to install 

https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

install_executorch.py
mergennachin: Even today, when people run install_executorch.py, it will transitively call install_requirements.py anyway.

The reason i did it here is to install the non core optional steps as last time. That way, people can do, 

```
./install_requirements.py
pip install .e
```

to install minimal core package.

This is useful for running on Intel Mac OS for instance. We can compile torch from source and use it as part of ET. Now today, we will require people to also compile torchvision and torchaudio from source on top of torch. See context here: https://github.com/pytorch/executorch/pull/10652

install_executorch.py
```diff
@@ -180,9 +181,13 @@ def main(args):
     # This option is used in CI to make sure that PyTorch build from the pinned commit
     # is used instead of nightly. CI jobs wouldn't be able to catch regression from the
     # latest PT commit otherwise
-    install_requirements(use_pytorch_nightly=not args.use_pt_pinned_commit)
-    os.execvp(
```
mergennachin, line 186: I had to change back from  os.execvp to subprocess since install_optional_example_requirement is never reachable because os.execvp completely replaces the current process with the pip install command

install_executorch.py
```diff
@@ -196,8 +201,12 @@ def main(args):
             "-v",
             "--extra-index-url",
             TORCH_NIGHTLY_URL,
-        ],
+        ]
     )
+    subprocess.run(cmd, check=True)
+
+    # Step 3: Extra (optional) packages that is only useful for running examples.
+    install_optional_example_requirement(use_pytorch_nightly)
```
jathu: If this is optional, do we want to stick this behind a flag?

jathu (approved):

PR #11654, opened 2025-06-13
Implement load_into for Mmap Data Loader

### LINKED ISSUE #11561: Implement load_into for mmap data loader

load_into is used if mutable state has a meaningful initial value. This primarily happens with on-device training but could also happen if some model had read before write state that we serialized an initial value for. 

For this task we need to implement this function https://github.com/pytorch/executorch/blob/main/runtime/core/data_loader.h#L111 in https://github.com/pytorch/executorch/blob/main/extension/data_loader/mmap_data_loader.h

example of implementation in filedataloader https://github.com/pytorch/executorch/blob/main/extension/data_loader/file_data_loader.h#L75 

Then add a corresponding test to https://github.com/pytorch/executorch/blob/main/extension/data_loader/test/mmap_data_loader_test.cpp

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @larryliu0820 @lucylq

keyprocedure:
### Summary
- Adds `load_into()` to support memory mapped loading directly into the caller's buffer.  
- Enables copying a specific byte range from the file without creating an internal buffer.
- Resuses input validation from `load()`.

Fixes #11561

### Test plan
- Added unit tests to `MmapDataLoaderTest` for `load_into()` to validate copying for both aligned and offset data.
- All Mmap Data Loader tests pass via: 
```
./build-ninja/extension/data_loader/test/extension_data_loader_test --gtest_filter='MmapDataLoaderTest.*'
```
- Full set of Data Loader tests pass via:
```
./build-ninja/extension/data_loader/test/extension_data_loader_test
```


extension/data_loader/mmap_data_loader.cpp
```diff
@@ -268,5 +280,101 @@ Result<size_t> MmapDataLoader::size() const {
...
+  ET_CHECK_OR_RETURN_ERROR(
+      pages != MAP_FAILED,
+      AccessFailed,
+      "Failed to map %s: mmap(..., size=%zd, ..., fd=%d, offset=0x%zx)",
+      file_name_,
+      range.size,
+      fd_,
+      range.start);
+
+  if (mlock_config_ == MlockConfig::UseMlock ||
```
JacobSzwejbka: I don't think we need to worry about the page getting swapped out in this entry point. Can probably skip mlocking.

extension/data_loader/mmap_data_loader.cpp
```diff
@@ -268,5 +280,101 @@ Result<size_t> MmapDataLoader::size() const {
...
+  }
+
+  // Nothing to copy
+  if (size == 0) {
+    return Error::Ok;
+  }
+
+  // Find the range of pages that covers the requested region.
+  Range range =
+      get_overlapping_pages(static_cast<uintptr_t>(offset), size, page_size_);
```
JacobSzwejbka, line 304: I dont really understand whats going on here but I see it originates from the load() fn so I guess its fine.

JacobSzwejbka (changes requested):

extension/data_loader/mmap_data_loader.cpp
JacobSzwejbka: oh offset=0 is guaranteed to be the start of a page for files. Ok I get it now then

extension/data_loader/mmap_data_loader.cpp
keyprocedure: Sounds good, removed mlocking from load_into()

extension/data_loader/mmap_data_loader.h
```diff
@@ -95,6 +95,17 @@ class MmapDataLoader final : public executorch::runtime::DataLoader {
 
   ET_NODISCARD executorch::runtime::Result<size_t> size() const override;
 
+  ET_NODISCARD executorch::runtime::Error validate_input(
```
JacobSzwejbka: nit: Can you make this private

JacobSzwejbka (changes requested):

extension/data_loader/mmap_data_loader.h
keyprocedure: Done, moved to private section

JacobSzwejbka (approved):

JacobSzwejbka:
just letting ci run again and then will merge

keyprocedure:
CI failure looks unrelated: `MimiModel` is missing `_context_for_encoder_decoder`

JacobSzwejbka:
Thanks for contributing!

keyprocedure:
I'm happy to help! Thanks for the review.

GregoryComer:
FYI I temporarily reverted this due to a meta-internal build issue. Sorry about that. I'm going to re-merge it shortly, preserving commit authorship. We'll also fix the CI so that it builds with -Wshadow.

keyprocedure:
Got it. Let me know if you'd like me to change anything.

PR #11660, opened 2025-06-13
Use elementwise_util instead of functional_util in op_neg, delete optimized op

swolchok:
Allows vectorization (checked assembly to make sure; it's different from the old optimized op because the optimized op wasn't parallelized), so we can delete the optimized op as well.


swolchok:
* __->__ #11660

swolchok:
converting to draft because we need to check assembly before we can commit this

swolchok:
ring-attention failure is a known issue on main, merging

swolchok:
correction, need to remove optimized op_neg test

PR #11666, opened 2025-06-14
Use wrappers from xnnpack.h for unary and binary ops (#11584)

### LINKED ISSUE #11584: Move to binary/unary subgraph

For binary(mul, add, sub) and unary operators,(abs, rsqrt,) we define them in our XNNPAK subgraph like this 

Binary:
https://github.com/pytorch/executorch/blob/07be1ffc0efe53119799d8257ee561dfa2a00c0e/backends/xnnpack/runtime/XNNCompiler.cpp#L676

Unary:
https://github.com/pytorch/executorch/blob/07be1ffc0efe53119799d8257ee561dfa2a00c0e/backends/xnnpack/runtime/XNNCompiler.cpp#L1747

However in XNNPACK we have moved to use the more broadly:
Define Unary: https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L580
Define Binary: https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L1146

let's try to use those instead

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

abhinaykukkadapu:
### Summary
This PR uses `xnn_define_binary` and `xnn_define_unary` to define XNNPack ops, instead of separately calling the individual definitions. 

Further changes:
1. Removes individual node definitions for unary and binary ops
2. Creates a wrapper macro to generate function defs for individual ops  using `xnn_define_binary` and `xnn_define_unary` inside.

Fixes #11584

### Test plan
```
## Build steps
cmake -DEXECUTORCH_BUILD_XNNPACK=ON ..
cmake --build cmake-out -j9

Tests ran:
./test/run_oss_cpp_tests.sh
.
.
.
100% tests passed, 0 tests failed out of 86
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,9 @@ xcuserdata/
 *.xcworkspace/
 *.xcframework/
 
+# clangd
+.cache/
+
 # misc
 /.vscode/
 *.so
--- a/backends/xnnpack/runtime/XNNCompiler.cpp
+++ b/backends/xnnpack/runtime/XNNCompiler.cpp
@@ -11,6 +11,7 @@
 #include <executorch/backends/xnnpack/serialization/schema_generated.h>
 #include <executorch/extension/threadpool/threadpool.h>
 #include <executorch/runtime/executor/pte_data_map.h>
+#include <xnnpack.h>
 #include <string>
 #include <unordered_map>
 #include <vector>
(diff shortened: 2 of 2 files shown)


abhinaykukkadapu:
lint runner failure seems unrelated. Local lint run is successful:

```
$ lintrunner -a
Warning: Could not find a lintrunner config at: '.lintrunner.private.toml'. Continuing without using configuration file.
  FLAKE8 success!
  CMAKE success!
  CLANGFORMAT success!
  UFMT success!
  ETCAPITAL success!
  NEWLINE success!
  NOTORCHINC success!
  NOSTDINC success!
  MYPY success!
  LICENSELINT success!
  TORCH_AO_IMPORT success!
ok No lint issues.
Successfully applied all patches.
```

mcr229:
Seems like a transient issue with lint runner for now. Thanks for taking this up. In XNNCompiler we have large switch statements that look like this:

https://github.com/pytorch/executorch/blob/cf252a8c6037a3009fc324690faae3c33a7caad3/backends/xnnpack/runtime/XNNCompiler.cpp#L2058

since the define_binary/unary all look like pretty much the same implementation, would it be possible to refactor some of these _DEFINE into _DEFINE_UNARY, _DEFINE_BINARY So that we don't need custom defineSubNode, defineAddNode. I know for a couple unary configs there is some variation with alpha so there is custom logic there, but this would defintely help a lot for readability.

abhinaykukkadapu:
@mcr229 please let me know if you suggest any more tests.

mcr229 (approved):

Issue #10961, opened 2025-05-19
XNN Int8 ReLU gives incorrect outputs


GregoryComer:
### 🐛 Describe the bug

When running standalone ReLU ops in int8 on XNNPACK, there appears to be some sort of correctness or memory safety issue causing incorrect outputs. Playing around with various inputs, there doesn't seem to be a clear pattern, which makes me suspect that there is an issue handling the tensor memory or dtype somewhere.

Repro:
```py
import torch
from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
from executorch.exir import to_edge_transform_and_lower
from executorch.runtime import Runtime

class Model(torch.nn.Module):
    def forward(self, x):
        return torch.nn.ReLU()(x)

inputs = (
    torch.Tensor([-128, -127, -126, -10, 5]).to(torch.int8),

)
et_program = to_edge_transform_and_lower(
    torch.export.export(Model(), inputs),
    partitioner=[XnnpackPartitioner()]
).to_executorch()

print(et_program.exported_program())

runtime = Runtime.get()
program = runtime.load_program(et_program.buffer)
method = program.load_method("forward")
et_out = method.execute(inputs)
ref_out = Model()(*inputs)

print(f"ET:  {et_out}")
print(f"Ref: {ref_out}")
```

Output:
```
ET:  [tensor([0, 0, 0, 0, 0], dtype=torch.int8)]
Ref: tensor([0, 0, 0, 0, 5], dtype=torch.int8)
```

### Versions

N/A

cc @digantdesai @mcr229 @cbilgin

digantdesai:
Can you dump a json from [here](https://github.com/pytorch/executorch/blob/33d47904c5781a87cb644bc09205c71b3147f61b/backends/xnnpack/serialization/xnnpack_graph_serialize.py#L310-L313)?

Also do we wire int8 to qu8 with fake scale/zp? And lastly what if you try quantizing fp32 with bogus scale/zp through all ones or something as calibration input?

mergennachin:
Original reproducer and all 256 int8 values produce exact results through CPU fallback.

PR #11685, opened 2025-06-14
Only support int8 and quant dtypes for quant operators

### LINKED ISSUE #11581: Update Partitioner Dtype Logic

Currently when validating dtypes for our partitions we allow int8:
https://github.com/pytorch/executorch/blob/07be1ffc0efe53119799d8257ee561dfa2a00c0e/backends/xnnpack/partition/config/xnnpack_config.py#L219

This is because we support this dtype in quant operations. However we don't support this dtype in any other setting! Let's add some code in this check to only support int8 dtype if the operator is a quant operator.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

aaron-ang:
### Summary
Fixes #11581. Check that node is a quant op before including `int8` and other quant `dtypes` in `valid_dtypes`.

### Test plan
Still working on creating tests.


aaron-ang:
@mcr229 @digantdesai  I'm running into some issues creating tests. Could you please provide some guidance on some reference files, etc.?

backends/xnnpack/partition/config/xnnpack_config.py
```diff
@@ -220,9 +225,12 @@ def _check_node_has_valid_dtype(self, node):
         valid_dtypes = {
             torch.float32,
             torch.float16,
-            torch.int8,
             torch.qint8,
         }
+        # Only allow int8 for quantization operations
```
digantdesai: What about running non quantized int8 tensors? IIRC we do support those, no?

aaron-ang: I was under the impression that `int8` is only supported for quant ops based on #11581, hence this PR. Has that rule changed?

digantdesai: Good call out. Thanks.

backends/xnnpack/partition/config/xnnpack_config.py
```diff
@@ -220,9 +225,12 @@ def _check_node_has_valid_dtype(self, node):
         valid_dtypes = {
             torch.float32,
             torch.float16,
-            torch.int8,
             torch.qint8,
```
digantdesai: what about this guy here?

aaron-ang: That makes sense, I've moved it into the conditional check and added other quant dtypes.

mcr229:
@GregoryComer can you point how to run the int8 tests that were failing in the new backend tester? Aside from that as long as existing tests are passing this change should be fine.

aaron-ang:
rebased and fixed lint

GregoryComer:
@mcr229 Here's one concrete repro from the tester: https://github.com/pytorch/executorch/issues/10961. I'll have an easier general repro soon for running the test suite.

digantdesai:
@GregoryComer  does this fix your test?

digantdesai (approved):

aaron-ang:
@digantdesai @mcr229 is there anything needed on my end to support this?

PR #11724, opened 2025-06-16
Arm backend: Added decomposition for MaxPool2d with dilation > 0.

wwwind:
Arm backend: Added decomposition for MaxPool2D operator with dilation > 0

Change-Id: I77efb3077c3a8f48450669fd461bfda5acc86520


zingo (approved):

zingo:
linter fail was detect on main also

PR #11733, opened 2025-06-16
Use MAP_SHARED to allow sharing memory between processing

shoumikhin:


mergennachin (approved):
Looks like safe enough.

- It is using PROT_READ in ::mmap()
- It is using open(file_name, O_RDONLY);

PR #10982, opened 2025-05-19
Script for benchmark stability assessment

guangy10:
### Summary
The custom script for ET benchmark stability assessment.
```
pip install openpyxl tabulate matplotlib
```
Then
```
python .ci/scripts/analyze_benchmark_stability.py \
    Benchmark\ Dataset\ with\ Private\ AWS\ Devices.xlsx \
    --reference_file Benchmark\ Dataset\ with\ Public\ AWS\ Devices.xlsx
```

Datasets:
  - [Benchmark Dataset with Private AWS Devices.xlsx](https://github.com/user-attachments/files/20596657/Benchmark.Dataset.with.Private.AWS.Devices.xlsx)
  - [Benchmark Dataset with Public AWS Devices.xlsx](https://github.com/user-attachments/files/20596666/Benchmark.Dataset.with.Public.AWS.Devices.xlsx)

The generated analysis:
```
Analyzing latency stability from primary file: /Users/guangyang/Desktop/Benchmark Dataset with Private AWS Devices.xlsx
Using reference file for comparison: /Users/guangyang/Desktop/Benchmark Dataset with Public AWS Devices.xlsx


====================================================================================================
===== LOADING PRIMARY DATASETS (Private) ==========================================================
====================================================================================================

Loading dataset: llama3_qlora+s22_android13
Loading dataset: llama3_spinq+s22_android13
Loading dataset: mv3_qnn+s22_android13
Loading dataset: mv3_xnnq8+s22_android13
Loading dataset: llama3_qlora+s22ultra_android14
Loading dataset: llama3_spinq+s22ultra_android14
Loading dataset: mv3_qnn+s22ultra_android14
Loading dataset: mv3_xnnq8+s22ultra_android14
Loading dataset: mv3_xnnq8+pixel3_rooted_android
Loading dataset: llama3_qlora+iphone15max_ios17
Loading dataset: llama3_spinq+iphone15max_ios17
Loading dataset: mv3_xnnq8+iphone15max_ios17
Loading dataset: mv3_coreml+iphone15max_ios17
Loading dataset: mv3_mps+iphone15max_ios17
Loading dataset: llama3_qlora+iphone15_ios18
Loading dataset: llama3_spinq+iphone15_ios18
Loading dataset: mv3_xnnq8+iphone15_ios18
Loading dataset: mv3_coreml+iphone15_ios18
Loading dataset: mv3_mps+iphone15_ios18


====================================================================================================
===== LOADING REFERENCE DATASETS (Public) =========================================================
[...]

PR #11734, opened 2025-06-16
Add script to fetch benchmark results for execuTorch

yangw-dev:
# Summary
Provide methods and script to fetch all execuTorch benchamrk data from HUD API into two dataset,private and public, the script will:
- fetch all data from HUD API from input time range in UTC
- clean out records and tables with only FAILURE_REPORT due to job-level failures
- get all private table metrics, generate `table_name` and find intersected public table metrics
- generate private and public table groups
- output data

OutputType:
- run with excel-sheet export
- run with csv export
- run with dataframe format print
- run with json format print

See more guidance in README.md

 the data is similar to the excel sheet generated manually  in https://github.com/pytorch/executorch/pull/10982
The result should be the same as the hud per model datatable:
image

## helper methods: common.py
provide common.py helper method to convert back csv and excel sheets back to {"groupInfo":{}, "df":df.DataFrame} format.

# run with
``` bash
python3 .ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py \
--startTime "2025-04-29T09:48:57" \
--endTime "2025-05-13T22:00:00" \
--outputType "excel" \
--models "mv3"

python3 .ci/scripts/benchmark_tooling/analyze_benchmark_stability.py \
--primary-file private.xlsx \
--reference-file public.xlsx
```
Generate excel files:
[private.xlsx](https://github.com/user-attachments/files/20844977/private.xlsx)
[public.xlsx](https://github.com/user-attachments/files/20844978/public.xlsx)

For instance you can find result for mv3 xnnq_q8 S22 Ultra android 14:
```

Latency Stability Analysis: table10 (Primary)
================================================================================
Model: mv3(xnnpack_q8)
Device: Samsung Galaxy S22 Ultra 5G (private)(Android 14)

Dataset Overview:
  - Number of samples: 88
  - Date range: 2025-04-29 09:48:57+00:00 to 2025-05-13 21:08:36+00:00

Central Tendency Metrics:
  - Mean latency: 2.91 ms
  - Median latency (P50): 2.54 ms
  - Mean trimmed latency: 2.41 ms
  - Median trimmed latency: 2.15 ms

Dispersion Metrics:
  - Standard deviation: 1.14 ms
  - Coefficient of variation (CV): 39.08%
  - Interquartile range (IQR): 0.82 ms
  - Trimmed standard deviation: 0.76 ms
  - Trimmed coefficient of variation: 31.60%

Percentile Metrics:
  - P50 (median): 2.54 ms
  - P90: 3.88 ms
  - P95: 4.60 ms
  - P99: 5.91 ms

Inter-Jitter Metrics (variability between runs):
  - Max/Min ratio: 5.6103
  - P99/P50 ratio: 2.3319
  - Mean rolling std (window=5): 0.79 ms

Intra-Jitter Metrics (variability within runs):
  - Mean trimming effect ratio: 15.37%
  - Max trimming effect ratio: 38.83%

Stability Assessment:
  - Overall stability score: 0.0/100
  - Overall stability rating: Poor

Interpretation:
  The benchmark shows poor stability (score: 0.0/100) with significant
  variation between runs (CV: 39.08%).
  Performance is unpredictable and may lead to inconsistent user experience.

  The significant difference between raw and trimmed means suggests
  considerable intra-run jitter (15.4%) with occasional outliers within benchmark runs.

  The max/min ratio of 5.61 indicates
  substantial performance differences between the best and worst runs.

  The P99/P50 ratio of 2.33 suggests
  occasional latency spikes that could affect tail latency sensitive applications.
```


yangw-dev:
FYI, this method can be more general, but since only execuTorch is using it, i just make it execuTorch specific @huydhn

yangw-dev:
the excel sheet has limit of sheet name len < 31, which can be easy to break in the future. @huydhn @guangy10 , I think instead of generate one file per category, maybe we can generate list of excel files stored in folders [private, public]


But right now with the hard-coded abbreviation, this works fine. THe excel sheet option is there in case people want to use it.

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Run with default output (CLI):
+```bash
+python3 get_benchmark_analysis_data.py --startTime "2025-06-11T00:00:00" --endTime "2025-06-17T18:00:00"
+```
+
+Additional options:
+- `--silent`: Hide processing logs, show only results
+- `--outputType df`: Display results in DataFrame format
+- `--outputType excel --excelDir "{YOUR_LOCAL_DIRECTORY}"`: Generate Excel files (`res_private.xlsx` and `res_public.xlsx`)
```
huydhn: Would a CSV file suffice? One less dependency I guess, but Excel is fine too if it's easier to use

yangw-dev: CSV  can't save data for multiple sheets, i can add csv as option too, it just have to be multiple files when generate the data

huydhn (approved):
Stamped to unblock!  Let's start using the script and improve it along the way

guangy10:
> We configured a fixed list of matching names to list limited tables

I think we should make it more flexible as there are always new models, recipes, devices added. For example, we recently add more models [(see on dash)](https://hud.pytorch.org/benchmark/llms?startTime=Tue%2C%2010%20Jun%202025%2019%3A31%3A51%20GMT&stopTime=Tue%2C%2017%20Jun%202025%2019%3A31%3A51%20GMT&granularity=day&lBranch=main&lCommit=f3a6c187053ded347af36db1b09c2f4538b30b91&rBranch=main&rCommit=5365c5559accc7a0d522eacdfb7385ce8914ef53&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=All%20Models&backendName=hf_xnnpack_custom_spda_kv_cache_8da4w&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms) from `huggingface/optimum-executorch` to the benchmark infra, and the list will keep expanding.

Similarly with new "devices" or "backends" available, we want to be able to query the results via the script as well.

> the excel sheet has limit of sheet name len < 31, which can be easy to break in the future. @huydhn @guangy10 , I think instead of generate one file per category, maybe we can generate list of excel files stored in folders [private, public]

Yeah noticed the limits when I manually created the excel sheet. Ideally I'd like to get rid of the excel sheet by wiring the outputs from db to the analysis script directly. Given what is currently supported in this PR, what does the workflow look like if I want to rerun the analysis? That is, how is this script interfaced to the analysis script?

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,52 @@
...
+Run with default output (CLI):
+```bash
+python3 .ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py --startTime "2025-06-11T00:00:00" --endTime "2025-06-17T18:00:00"
+```
+
+Additional options:
+- `--silent`: Hide processing logs, show only results
+- `--outputType df`: Display results in DataFrame format
+- `--outputType excel --outputDir "{YOUR_LOCAL_DIRECTORY}"`: Generate Excel file with multiple sheets (`res_private.xlsx` and `res_public.xlsx`)
+- `--outputType csv --outputDir "{YOUR_LOCAL_DIRECTORY}"`: Generate CSV files in folders (`private` and `public`)
```
guangy10: Oh now I see how it's used with the analysis script

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+    ["llama3", "spinq", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "xnnpack_q8", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "coreml", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "mps", "iphone_15_pro_max", "ios_17"],
+    ["llama3", "qlora", "iphone_15", "ios_18.0"],
+    ["llama3", "spinq", "iphone_15", "ios_18.0"],
+    ["mv3", "xnnpack_q8", "iphone_15", "ios_18.0"],
+    ["mv3", "coreml", "iphone_15", "ios_18.0"],
+    ["mv3", "mps", "iphone_15", "ios_18.0"],
+]
```
guangy10: Instead of hard-coded it, can we get this list dynamically by querying the db? The models, backends, devices are keep changing.

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+    ["mv3", "xnnpack_q8", "s22_ultra_5g", "android_14"],
+    ["mv3", "xnnpack_q8", "pixel3_rooted", "android"],
+    ["llama3", "qlora", "iphone_15_pro_max", "ios_17"],
+    ["llama3", "spinq", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "xnnpack_q8", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "coreml", "iphone_15_pro_max", "ios_17"],
+    ["mv3", "mps", "iphone_15_pro_max", "ios_17"],
+    ["llama3", "qlora", "iphone_15", "ios_18.0"],
+    ["llama3", "spinq", "iphone_15", "ios_18.0"],
+    ["mv3", "xnnpack_q8", "iphone_15", "ios_18.0"],
```
guangy10: nit about the os versioning. we should cover minor version for all/none to be consistent.

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+
+# The abbreviations used to generate the short name for the benchmark result table
+# this is used to avoid the long table name issue when generating excel file (<=31 characters)
+ABBREVIATIONS = {
+    "samsung": "",
+    "galaxy": "",
+    "5g": "",
+    "xnnpackq8": "xnnq8",
+    "iphone15promax": "iphone15max",
+}
```
guangy10: This is totally fine or if you want to make it more robust, we can create such an ABBREVIATIONS and use it in both the query script (encode) and analysis script (decode)

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+            "%Y-%m-%dT%H:%M:%S"
+        )
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            f"Invalid datetime format for '{value}'. Expected: YYYY-MM-DDTHH:MM:SS"
+        )
+
+
+def argparser():
+    parser = argparse.ArgumentParser()
```
guangy10: Optionally can we allow filtering by model, backends, devices?

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+    return parser.parse_args()
+
+
+class ExecutorchBenchmarkFetcher:
+    """
+    Fetch and process benchmark data from HUD API for ExecutorchBenchmark.
+
+    Usage:
+        fetcher = ExecutorchBenchmarkFetcher()
+        fetcher.run(start_time, end_time, private_device_matching_list, public_device_matching_list)
```
guangy10: Note that the reference list (from public devices) is optional when running the analysis script. In the analysis we do both inter- and intra- analysis, if one model+config is only available from private devices, we can still use it to run the intra-private analysis. In fact intra-private analysis is what we would keep running in the long-term after shutdown all the public devices.

guangy10 (changes requested):

guangy10:
Thanks for making the benchmark analysis semi-automated!

guangy10:
You will need to fix the linter error as well. Here is the guide: https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
```diff
@@ -0,0 +1,591 @@
...
+    repo: str
+    benchmark_name: str
+    start_time: str
+    end_time: str
+    group_table_by_fields: list
+    group_row_by_fields: list
+
+
+BASE_URLS = {
+    "local": "http://localhost:3000",
```
guangy10, line 106: Is this for testing purpose only?

guangy10:
Maybe consider adding some tests for the script.

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
yangw-dev: yah, I can do that

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
yangw-dev: sounds good, will update!

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
yangw-dev: yah! I use it when I run the api endpoint locally :)!  maybe change it to `dev`

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
yangw-dev: yah, I think currently when we run this, 

it will list all the combination name now (public and private > 200 combination), I can add a method to list all of them, so you can generate the json, and when query with run, you can pass the json file or json string, will it work?

guangy10: > it will list all the combination name now (public and private > 200 combination

@yangw-dev Why do we have so many? The source of truth is data from private. Maybe you can split the query into two, once query to the private, which shouldn't return many results, given we only have very few devices there. Then the 2nd query to the public would just trying to match with entries in the private. We do not care entries that are only available in public.

yangw-dev: SO from  startTime 2025-06-11T00:00:00 to endTime 2025-06-17T00:00:00 

it returns 60 tables for private devices
table is unique by (mode_name, backend_name, device_name, arch(iso/android version)

yangw-dev: sounds good! I will try to do that

.ci/scripts/benchmark_tooling/get_benchmark_analysis_data.py
yangw-dev: let's just use the fullname then, but in that case, if we genearte csv or excel file, I will leave the details of the table in the first row and first column, 
so you need to iterate each file/sheet and check the details

yangw-dev:
> You will need to fix the linter error as well. Here is the guide: https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

I think the lintrunner error is not caued by my pr seems, but others

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+This script analyzes the stability of benchmark data, comparing the results of private and public devices.
+
+#### Quick Start
+
+```bash
+python .ci/scripts/benchmark_tooling/analyze_benchmark_stability.py \
+    "Benchmark Dataset with Private AWS Devices.xlsx" \
+    --reference_file "Benchmark Dataset with Public AWS Devices.xlsx"
+```
```
guangy10: This is outdated with new update in this PR.

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,130 @@
...
+- `--no-silent`: Show processing logs (default: only show results & minimum logging)
+- `print-all-table-info`: show all cleaned table infos, this helps user to pick the correct format of filters
+
+##### Output Options:
+- `--outputType`: Choose output format (default: "print")
+  - `print`: Display results in console
+  - `json`: Generate JSON file
+  - `df`: Display results in DataFrame format: `{'private': List[{'groupInfo':Dict,'df': DF},...],'public':List[{'groupInfo':Dict,'df': DF}]`
+  - `excel`: Generate Excel files with multiple sheets, the field in first row and first column contains the json string of the raw metadata
+  - `csv`: Generate CSV files in separate folders, the field in first row and first column contains the json string of the raw metadata
```
guangy10: - It seems like there are no major difference. Do we need both? I'm okay if analysis script can accept either format.
- With the additional metadata embedded, the analysis script should be updated accordingly otherwise it breaks

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,130 @@
...
+- `--outputDir`: Directory to save output files (default: current directory)
+
+##### Filtering Options:
+Notice, the filter needs full name matchings with correct format, to see all the options of the filter choices, please run the script with `--print-all-table-info`, and pay attention to section `Full list of table info from HUD API` with the field 'info', which contains normalized data we use to filter records from the original metadata 'groupInfo'.
+The filter block any record if it does not in any of the filter keywords.
+
+- `--devices`: Filter by specific device names (e.g., "samsung-galaxy-s22-5g", "samsung-galaxy-s22plus-5g")
```
guangy10: 1. Is this field always refer to private device?
2. What is the format to specify multiple devices? comma-separated or something else, should it be clarified here?
3. All supported devices can be found [here](https://github.com/pytorch/executorch/blob/main/.ci/scripts/gather_benchmark_configs.py#L20-L29) (Note it's mixing of private and public), should `--help` list all supported devices so users can copy/paste easily.
4. We should be consistent on names of these devices in all places. https://github.com/pytorch/executorch/blob/main/.ci/scripts/gather_benchmark_configs.py#L20-L29

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,130 @@
...
+- `--backends`: Filter by specific backend names (e.g.,  "qnn-q8" , ""llama3-spinquan)
+- `--models`: Filter by specific model names (e.g "mv3" "meta-llama-llama-3.2-1b-instruct-qlora-int4-eo8")
```
guangy10: I don't know where `meta-llama-llama-3.2-1b-instruct-qlora-int4-eo8` comes from, but this model name is different from what has been used elsewhere. How could users know what the is the right model name to specify if they want to query results for Qwen3? We should be consistent on the model names as well

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,130 @@
...
+  - `excel`: Generate Excel files with multiple sheets, the field in first row and first column contains the json string of the raw metadata
+  - `csv`: Generate CSV files in separate folders, the field in first row and first column contains the json string of the raw metadata
+- `--outputDir`: Directory to save output files (default: current directory)
+
+##### Filtering Options:
+Notice, the filter needs full name matchings with correct format, to see all the options of the filter choices, please run the script with `--print-all-table-info`, and pay attention to section `Full list of table info from HUD API` with the field 'info', which contains normalized data we use to filter records from the original metadata 'groupInfo'.
+The filter block any record if it does not in any of the filter keywords.
+
+- `--devices`: Filter by specific device names (e.g., "samsung-galaxy-s22-5g", "samsung-galaxy-s22plus-5g")
+- `--backends`: Filter by specific backend names (e.g.,  "qnn-q8" , ""llama3-spinquan)
```
guangy10: Same comment as I left for devices and models. The name must be consistent otherwise it's not possible to figure out the correct backend name

guangy10 (changes requested):
Thank you for iterating on this! There are few critical issues to be fixed in order to merge

.ci/scripts/benchmark_tooling/README.md
yangw-dev: Hi please see the common.py, I also provide tools to basically read those files properly, see README.md for guidance.

I'm ok to just still wit one :)!

.ci/scripts/benchmark_tooling/README.md
yangw-dev: I think this can be general, I can associate with the list to provide list of choices when user try to add to it, if underscore needs to be consistent, i can change some thing, but

.ci/scripts/benchmark_tooling/README.md
yangw-dev: this is generated from removing suspicious symble such as '/',  

we got it directly form the database where executorch runs put it in benchmark tables, one option I list in README.md is call the script with --list-all-table-info , this will give the filter name, but if we want to be consistent, I need a proper mapping provided bv you guys

.ci/scripts/benchmark_tooling/README.md
yangw-dev: do you have a list of backend name with correct format? This is not my area, would be great if you guys can provide that

guangy10: https://hud.pytorch.org/benchmark/llms?repoName=pytorch%2Fexecutorch

.ci/scripts/benchmark_tooling/README.md
yangw-dev: I can remove it now

.ci/scripts/benchmark_tooling/README.md
yangw-dev: I will leave it since it also helps me debug for a bit

.ci/scripts/benchmark_tooling/README.md
yangw-dev: updated

.ci/scripts/benchmark_tooling/README.md
yangw-dev: now everything is same as the HUD ones, except device_pools

yangw-dev:
@guangy10 
I updated logics 
 Fetch and process benchmark data from HUD API for ExecutorchBenchmark.

    This class provides methods to:
    1. Fetch all benchmark data for a specified time range
    2. Get all private device info within the time range
    3. Filter the private device data if filter is provided, for backends and models we expect human-readable models, notice, everything is first based on private devices.
    4. Then use the filtered private device data to find matched the public device data using [model, backend, device, arch]
    3. Export results in various formats (JSON, DataFrame, Excel, CSV)

.ci/scripts/benchmark_tooling/README.md
yangw-dev: I changed everything to private devices now since we first filter wiht private devices, i only add two private_device_pool options there

.ci/scripts/benchmark_tooling/README.md
guangy10: Removing it doesn't solve the problem. The guidance here should show instructions of how to pipe the data generated from `get_benchmark_analysis_data` to `analyze_benchmark_stability`. This is missing and I doubt if it can work as expected given the output format change in `get_benchmark_analysis_data`, but the input format is not updated in `analyze_benchmark_stability`.

guangy10 (changes requested):
This may not be able to work together with the analysis script.

.ci/scripts/benchmark_tooling/README.md
yangw-dev: I think you need to modify the code form analyze_benchmark_stability to make it seamless, cuz the limitation of the excel sheet name is <=31, 
the next step is make this workable with the analyze_benchmark_stability

guangy10:
Just to clarify, the whole purpose of this PR is to make it easier to analysis the stability of the benchmark by end of H1, and that should be part of the core of the benchmark infra. The work is incomplete if the newly added script can not work with the analysis script or require additional changes to be handled separately. The DevX is not getting better.

yangw-dev:
> Just to clarify, the whole purpose of this PR is to make it easier to analysis the stability of the benchmark by end of H1, and that should be part of the core of the benchmark infra. The work is incomplete if the newly added script can not work with the analysis script or require additional changes to be handled separately. The DevX is not getting better.
 
please review this again, this script is synced with analysis script with excel output now, test mv3 data results, and post the analysis example in the comment, for samples i saw, the result generated from the new script is similar to the results post in https://github.com/pytorch/executorch/pull/10982

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,171 @@
...
+  - `df`: Display results in DataFrame format: `{'private': List[{'groupInfo':Dict,'df': DF},...],'public':List[{'groupInfo':Dict,'df': DF}]`
+  - `excel`: Generate Excel files with multiple sheets, the field in first row and first column contains the JSON string of the raw metadata
+  - `csv`: Generate CSV files in separate folders, the field in first row and first column contains the JSON string of the raw metadata
+- `--outputDir`: Directory to save output files (default: current directory)
+
+##### Filtering Options:
+
+- `--private-device-pools`: Filter by private device pool names (e.g., "samsung-galaxy-s22-5g", "samsung-galaxy-s22plus-5g")
+- `--backends`: Filter by specific backend names (e.g., "qnn-q8", "llama3-spinquan")
+- `--models`: Filter by specific model names (e.g., "mv3", "meta-llama-llama-3.2-1b-instruct-qlora-int4-eo8")
```
guangy10: The examples are outdated. Please fix before merging.
nit: Can we rename `--private-device-pools` to `--devices` to be more generic? Users don't need to know private/public, and there will be no public pool in the long-term.

guangy10 (approved):

.ci/scripts/benchmark_tooling/README.md
yangw-dev: updated! I think it should be device_pools now cuz the names we use here are actually device pools instead of specific device

for specific device, can have a followup pr if it's needed

.ci/scripts/benchmark_tooling/README.md
```diff
@@ -0,0 +1,171 @@
...
+  - `df`: Display results in DataFrame format: `{'private': List[{'groupInfo':Dict,'df': DF},...],'public':List[{'groupInfo':Dict,'df': DF}]`
+  - `excel`: Generate Excel files with multiple sheets, the field in first row and first column contains the JSON string of the raw metadata
+  - `csv`: Generate CSV files in separate folders, the field in first row and first column contains the JSON string of the raw metadata
+- `--outputDir`: Directory to save output files (default: current directory)
+
+##### Filtering Options:
+
+- `--device-pools`: Filter by private device pool names (e.g., "samsung-galaxy-s22-5g", "samsung-galaxy-s22plus-5g")
+- `--backends`: Filter by specific backend names (e.g.,"xnnpack_q8")
+- `--models`: Filter by specific model names (e.g., "mv3", "meta-llama-llama-3.2-1b-instruct-qlora-int4-eo8")
```
guangy10, line 82: Note that the examples names are still incorrect

guangy10 (approved):
linter error to fix

PR #11735, opened 2025-06-16
deduplicate torch ao debugger tests between pytorch/ao and ExecuTorch

Gasoonjia:
Summary: This diff deduplicates numeric debugging tests on XnnPack quantizer between torchao and ExecuTorch.

Differential Revision: D76634915

### DIFF
--- a/backends/xnnpack/test/quantizer/test_pt2e_quantization.py
+++ b/backends/xnnpack/test/quantizer/test_pt2e_quantization.py
@@ -6,6 +6,8 @@
 
 # pyre-unsafe
 
+import unittest
+
 from collections import Counter
 from typing import Dict, Tuple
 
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):

backends/xnnpack/test/quantizer/test_pt2e_quantization.py
```diff
@@ -722,7 +724,7 @@ def test_save_load(self) -> None:
 
 instantiate_parametrized_tests(TestQuantizePT2E)
 
-
+@unittest.skip("TODO: Reenable it after debug infrature finish update")
```
digantdesai, line 727: This is to avoid updating two places? I was hoping you did this other way around given the XNNPACK Quantizer in Torch is scheduled to get deprecated.

Gasoonjia: yes i will remove the xnnpack-related test for numberic debugging on ao side. Just due to some sync issue i need to skip our test first, and I will bring it back asap.

PR #11745, opened 2025-06-17
Fix for overriding installed executorch when running optimum-et models

guangy10:
### Summary
Running `install_dev.py` for `optimum-executorch` will force overriding installed `executorch` and torch deps to the pinned nightly in `optimum-executorch`. In ExecuTorch CI including the benchmark, we would want to always run the optimum-executorch models with ExecuTorch from source to catch issues/regressions.  

### Test plan
Verified the installed deps in the CI and benchmark jobs


huydhn (approved):
Oh, good catch!

PR #11747, opened 2025-06-17
Update PAL override docs for new APIs

GregoryComer:
### Summary
Update PAL docs in using-executorch-runtime-integration.md to describe the new PAL override mechanism introduced in https://github.com/pytorch/executorch/pull/10675. I've left the docs in for the weak symbol override mechanism, as well, though I've recommended use of the runtime APIs.

### Test plan
I built the docs locally to validate form and content.

### DIFF
--- a/docs/source/using-executorch-runtime-integration.md
+++ b/docs/source/using-executorch-runtime-integration.md
@@ -24,7 +24,49 @@ For non-POSIX-compliant systems, a minimal no-op PAL implementation is provided.
 
 ### Overriding the PAL
 
-Overriding the default PAL implementation is commonly done to route logs to a user-specified destination or to provide PAL functionality on embedded systems. To override one or more PAL methods, take the following steps:
+Overriding the default PAL implementation is commonly done to route logs to a user-specified destination or to provide PAL functionality on embedded systems. The PAL can be overriden usinn runtime APIs or at link time. Prefer the runtime API unless you specifically need link-time overrides.
+
+### Runtime PAL Registration
+
+To register a custom PAL implementation, take the following steps:
+
+- Include
+  [`executorch/runtime/platform/platform.h`](https://github.com/pytorch/executorch/blob/main/runtime/platform/platform.h)
+  in one of your application's `.c` or `.cpp` files.
+- Create an instance of the [PalImpl](https://github.com/pytorch/executorch/blob/7b39a0ce63bfb5124d4d29cfb6c8af85a3c580ba/runtime/platform/platform.h#L163) struct.
+  - Set one or more fields to custom PAL function implementations. Leave fields as null to use the default platform implementation.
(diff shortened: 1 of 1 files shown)


docs/source/using-executorch-runtime-integration.md
```diff
@@ -24,7 +24,47 @@ For non-POSIX-compliant systems, a minimal no-op PAL implementation is provided.
...
+  std::cerr << "[" << filename << ":" << line << "] " << message << std::endl;
+}
+
+runtime::PalImpl build_pal() {
+  return runtime::PalImpl::create(emit_log_message, __FILE__);
+}
+
+// Update PAL to redirect logs.
+ET_UNUSED bool registration_result = runtime::register_pal(build_pal());
+```
```
larryliu0820: Is this how you do static initialization? It should be a static variable right?

GregoryComer: @larryliu0820 I updated the example code to be slightly more clear. The original snippet is inside an anonymous namespace, which serves the same purpose of translation-unit scoping for the variable. But it is a global constructor in the same way. I've included the anonymous namespace in the snippet so that this is more clear.

docs/source/using-executorch-runtime-integration.md
```diff
@@ -24,7 +24,49 @@ For non-POSIX-compliant systems, a minimal no-op PAL implementation is provided.
...
+- Include
+  [`executorch/runtime/platform/platform.h`](https://github.com/pytorch/executorch/blob/main/runtime/platform/platform.h)
+  in one of your application's `.c` or `.cpp` files.
+- Create an instance of the [PalImpl](https://github.com/pytorch/executorch/blob/7b39a0ce63bfb5124d4d29cfb6c8af85a3c580ba/runtime/platform/platform.h#L163) struct.
+  - Set one or more fields to custom PAL function implementations. Leave fields as null to use the default platform implementation.
+  - The PalImpl struct provides a [create](https://github.com/pytorch/executorch/blob/7b39a0ce63bfb5124d4d29cfb6c8af85a3c580ba/runtime/platform/platform.h#L168) method for this purpose.
+- Call `executorch::platform::register_pal(pal_impl)` to register the implementation.
+  - This can be done from as as a global constructor, as in the example below.
+
+Here is a complete example from [pybindings.cpp](https://github.com/pytorch/executorch/blob/7b39a0ce63bfb5124d4d29cfb6c8af85a3c580ba/extension/pybindings/pybindings.cpp#L4), where logs are redirected to show up properly in a Python notebook environment.
```
larryliu0820: This link is pointing to a comment though

larryliu0820 (approved):

docs/source/using-executorch-runtime-integration.md
GregoryComer: Ah, oops. Good catch. Thanks.

PR #11748, opened 2025-06-17
Add XNNPACK backend option for workspace sharing

GregoryComer:
### Summary
Add a backend option for XNNPACK to enable runtime control of workspace sharing. I've added 3 mode options - Disabled, PerModel, and Global. PerModel shares the workspace between all CALL_DELEGATE instances in a model, keyed by memory allocator address (see below). Global uses a single workspace instance.

I've written the code to allow for the active workspace mode to be safely changed at any time. The workspace instance is resolved at delegate instance init time (model load) and is stored in the XNNExecutor instance. This design will also allow us to set per-model sharing options in the future. I've introduced a wrapper class (XNNWorkspace) to help with synchronization.

With regard to the PerModel behavior, I am using the address of the runtime allocator to disambiguate the model. This is not ideal in the long-run, but there is some larger discussion around generating IDs in a coherent manner in multithreaded environments without synchronization in the core runtime. This might require PAL changes (exposing a thread ID, for example), so I intend to come back to this.

It should be possible to transparently update this logic in the future. The program ID can collide or change without affecting correctness, but may increase memory (for collisions) or enforce extra synchronization (if unstable between delegate instances in a method).

I'd like to add a PerMethod mode as a follow-up. This should be keyed to the specific method instance (not name), such that multiple method instances for the same method can be loaded for execution on different threads without forcing synchronization, but still allow sharing between call delegate instances in each method instance. This will require a unique method identifier.

### Test plan
CI. I've also added a set of dedicated tests for getting/setting the option, running PTEs in each mode, switching modes at runtime, and I've also updated the multithreaded stress test to run in each mode.

### DIFF
--- a/backends/xnnpack/runtime/XNNCompiler.cpp
+++ b/backends/xnnpack/runtime/XNNCompiler.cpp
@@ -1891,9 +1891,8 @@ ET_NODISCARD Error XNNCompiler::compileModel(
   xnn_weights_cache_t weights_cache_ptr = nullptr;
 #endif
 
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-  ET_CHECK_OR_RETURN_ERROR(
-      workspace != nullptr, Internal, "Failed to initialize XNNPACK workspace");
+  // NOLINTBEGIN(facebook-hte-NullableDereference) - weights cache is allowed to
+  // be null
   status = xnn_create_runtime_v4(
       subgraph.get(),
       weights_cache_ptr,
@@ -1901,14 +1900,7 @@ ET_NODISCARD Error XNNCompiler::compileModel(
       ::executorch::extension::threadpool::get_pthreadpool(),
       runtime_flags,
       &runtime_ptr);
-#else
-  status = xnn_create_runtime_v3(
-      subgraph.get(),
-      weights_cache_ptr,
-      ::executorch::extension::threadpool::get_pthreadpool(),
-      runtime_flags,
-      &runtime_ptr);
-#endif
+  // NOLINTEND(facebook-hte-NullableDereference)
 
   ET_CHECK_OR_RETURN_ERROR(
       xnn_status_success == status,
--- a/backends/xnnpack/runtime/XNNPACKBackend.cpp
+++ b/backends/xnnpack/runtime/XNNPACKBackend.cpp
@@ -7,7 +7,10 @@
  */
 
 #include <executorch/backends/xnnpack/runtime/XNNCompiler.h>
+#include <executorch/backends/xnnpack/runtime/XNNPACKBackend.h>
 #include <executorch/backends/xnnpack/runtime/XNNWeightsCache.h>
+#include <executorch/backends/xnnpack/runtime/XNNWorkspace.h>
+#include <executorch/backends/xnnpack/runtime/XNNWorkspaceManager.h>
 #include <executorch/runtime/backend/interface.h>
 #include <executorch/runtime/core/error.h>
 #include <executorch/runtime/core/evalue.h>
@@ -21,14 +24,18 @@
 namespace executorch {
 namespace backends {
 
+using executorch::backends::xnnpack::WorkspaceSharingMode;
+using executorch::backends::xnnpack::XNNWorkspace;
 using executorch::backends::xnnpack::delegate::XNNWeightsCache;
 using executorch::ET_RUNTIME_NAMESPACE::Backend;
 using executorch::ET_RUNTIME_NAMESPACE::BackendExecutionContext;
 using executorch::ET_RUNTIME_NAMESPACE::BackendInitContext;
+using executorch::ET_RUNTIME_NAMESPACE::BackendOptionContext;
 using executorch::ET_RUNTIME_NAMESPACE::CompileSpec;
 using executorch::ET_RUNTIME_NAMESPACE::DelegateHandle;
 using executorch::ET_RUNTIME_NAMESPACE::NamedDataMap;
 using executorch::runtime::ArrayRef;
+using executorch::runtime::BackendOption;
 using executorch::runtime::Error;
 using executorch::runtime::EValue;
 using executorch::runtime::FreeableBuffer;
@@ -51,23 +58,8 @@ class XnnpackBackend final
       return;
     }
 
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    // Create a workspace for the XNNExecutor to use. This workspace will be
-    // shared across all delegate instances.
-    ET_LOG(Debug, "Creating XNN workspace");
-    xnn_workspace_t workspace = nullptr;
-    status = xnn_create_workspace(&workspace);
-    if (status != xnn_status_success) {
-      ET_LOG(
-          Error,
-          "Failed to create XNN workspace, XNNPACK status: 0x%x",
-          (unsigned int)status);
-      workspace = nullptr;
-      return;
-    }
-    workspace_.reset(workspace);
-    ET_LOG(Debug, "Created XNN workspace: %p", workspace_.get());
-#endif // ENABLE_XNNPACK_SHARED_WORKSPACE
+    // Workspace manager is initialized with the appropriate default mode in its
+    // constructor
   }
 
   bool is_available() const override {
@@ -85,11 +77,12 @@ class XnnpackBackend final
     }
 
     const NamedDataMap* named_data_map = context.get_named_data_map();
-    // thread safe. This can heppen when multiple threads call init() on
+    // thread safe. This can happen when multiple threads call init() on
     // the same backend instance.
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
+
+    auto program_id =
+        reinterpret_cast<uintptr_t>(context.get_runtime_allocator());
+    auto workspace = ET_UNWRAP(get_or_create_workspace(program_id));
 
 #ifdef ENABLE_XNNPACK_WEIGHTS_CACHE
     const std::lock_guard<std::mutex> lock_weight_cache(weights_cache_mutex_);
@@ -97,17 +90,19 @@ class XnnpackBackend final
         context.get_runtime_allocator(), named_data_map);
 #endif
 
+    auto [workspace_lock, workspace_ptr] = workspace->acquire();
+
     // Executor has been allocated but not constructed, ensure that runtime_ is
     // nullptr by constructing it in place here. NOTE: Since we use placement
     // new and since this type is not trivially destructible, we must call the
     // destructor manually in destroy().
-    new (executor) xnnpack::delegate::XNNExecutor;
+    new (executor) xnnpack::delegate::XNNExecutor(workspace);
     Error err = xnnpack::delegate::XNNCompiler::compileModel(
         processed->data(),
         processed->size(),
         executor,
         weights_cache_.get(),
-        workspace_.get(),
+        workspace_ptr,
         named_data_map);
     // This backend does not need its processed data after compiling the model.
     processed->Free();
@@ -130,13 +125,11 @@ class XnnpackBackend final
       Span<EValue*> args) const override {
     auto executor = static_cast<xnnpack::delegate::XNNExecutor*>(handle);
 
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
-
 #ifdef ENABLE_XNNPACK_WEIGHTS_CACHE
     const std::lock_guard<std::mutex> lock_weights_cache(weights_cache_mutex_);
 #endif
+
+    auto [raii_lock, _] = executor->get_workspace().acquire();
 
     // Prepare Inputs/Outputs and Propagate Input Shapes
     Error err = executor->prepare_args(args);
@@ -158,13 +151,6 @@ class XnnpackBackend final
 
   void destroy(DelegateHandle* handle) const override {
     if (handle != nullptr) {
-      // This is needed to serialize access to xnn_delete_runtime which is not
-      // thread safe. This can heppen when multiple threads call destroy() on
-      // the same backend instance.
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-      const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
-
       auto executor = static_cast<xnnpack::delegate::XNNExecutor*>(handle);
 
 #ifdef ENABLE_XNNPACK_PROFILING
@@ -176,18 +162,84 @@ class XnnpackBackend final
           weights_cache_mutex_);
       weights_cache_->delete_packed_data(executor->get_packed_data_names());
 #endif
+
+      // This is needed to serialize access to xnn_delete_runtime which is not
+      // thread safe. This can heppen when multiple threads call destroy() on
+      // the same backend instance.
+      auto [raii_lock, _] = executor->get_workspace().acquire();
+
       // XNNExecutor is not trivially destructible. Since this was constructed
       // manually in init(), we must destroy it manually here.
       executor->~XNNExecutor();
     }
   }
 
+  Error get_option_internal(
+      BackendOptionContext& context,
+      executorch::runtime::Span<executorch::runtime::BackendOption>&
+          backend_options) const {
+    // Intentionally not locking here as it is not required.
+
+    // Verify that the expected option key is present and modify the value
+    for (size_t i = 0; i < backend_options.size(); ++i) {
+      if (strcmp(
+              backend_options[i].key,
+              xnnpack::workspace_sharing_mode_option_key) == 0) {
+        // Set the value to what was stored by set_option
+        backend_options[i].value =
+            static_cast<int>(workspace_manager_.get_sharing_mode());
+      }
+    }
+
+    return Error::Ok;
+  }
+
+  Error get_option(
+      BackendOptionContext& context,
+      executorch::runtime::Span<executorch::runtime::BackendOption>&
+          backend_options) override {
+    return get_option_internal(context, backend_options);
+  }
+
+  Error set_option(
+      BackendOptionContext& context,
+      const executorch::runtime::Span<executorch::runtime::BackendOption>&
+          backend_options) override {
+    if (backend_options.size() > 0) {
+      for (const auto& option : backend_options) {
+        if (strcmp(option.key, xnnpack::workspace_sharing_mode_option_key) ==
+            0) {
+          if (auto* val = std::get_if<int>(&option.value)) {
+            if (*val < 0 ||
+                *val > static_cast<int>(WorkspaceSharingMode::Count)) {
+              ET_LOG(
+                  Error,
+                  "XNNPACK workspace sharing mode must be between 0 and %d, inclusive, but was %d.",
+                  static_cast<int>(WorkspaceSharingMode::Count),
+                  *val);
+              return Error::InvalidArgument;
+            }
+
+            ET_LOG(
+                Debug, "Setting XNNPACK workspace sharing mode to %d.", *val);
+            auto status = workspace_manager_.set_sharing_mode(
+                static_cast<WorkspaceSharingMode>(*val));
+            if (status != Error::Ok) {
+              return status;
+            }
+          } else {
+            ET_LOG(Error, "XNNPACK workspace sharing mode must be an integer.");
+            return Error::InvalidArgument;
+          }
+        }
+      }
+    }
+    return Error::Ok;
+  }
+
  private:
-  // This is a global workspace for all delegate instances.
-  mutable std::mutex workspace_mutex_;
-  std::unique_ptr<xnn_workspace, decltype(&xnn_release_workspace)> workspace_{
-      nullptr,
-      &xnn_release_workspace};
+  // Workspace manager for handling workspace sharing modes
+  mutable xnnpack::XNNWorkspaceManager workspace_manager_;
 
   // Weights cache is global to all delegate instances.
   mutable std::mutex weights_cache_mutex_;
@@ -195,13 +247,21 @@ class XnnpackBackend final
       std::make_unique<XNNWeightsCache>();
 
   // Lock Hiearchy for Mutexes:
-  // workspace_mutex_
   // weights_cache_mutex_
+  // workspace_meta_mutex_
+  // workspace_mutex_ (owned by executor)
+
+  // Retrieve a workspace for the given method ID, depending on the sharing
+  // mode.
+  Result<std::shared_ptr<XNNWorkspace>> get_or_create_workspace(
+      uintptr_t program_id) const {
+    return workspace_manager_.get_or_create_workspace(program_id);
+  }
 };
 
 namespace {
-auto cls = XnnpackBackend();
-Backend backend{"XnnpackBackend", &cls};
+auto backend_instance = XnnpackBackend();
+Backend backend{xnnpack::xnnpack_backend_key, &backend_instance};
 static auto success_with_compiler = register_backend(backend);
 } // namespace
 
--- a/backends/xnnpack/runtime/XNNPACKBackend.h
+++ b/backends/xnnpack/runtime/XNNPACKBackend.h
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch::backends::xnnpack {
+/// The key for the backend. This is used to register the backend, check
+/// availability, and get/set options.
+const char xnnpack_backend_key[] = "XnnpackBackend";
+
+/// The key for the workspace sharing option. See the WorkspaceSharingMode enum
+/// for a description of the associated functionality.
+const char workspace_sharing_mode_option_key[] = "workspace_sharing_mode";
+
+/// Workspace sharing mode. This is a backend option that can be set via the
+/// set_option API to control memory sharing between CALL_DELEGATE instances.
+/// This is useful for reducing memory consumption.
+enum class WorkspaceSharingMode {
+  /// No workspace sharing. Each CALL_DELEGATE instance will have its own
+  /// workspace (memory arena).
+  Disabled = 0,
+
+  /// All CALL_DELEGATE instances in a given program will share a workspace.
+  /// This reduces memory consumption
+  /// for methods with multiple delegate calls, at the cost of only allowing one
+  /// method to execute at a time.
+  PerModel = 1,
+
+  /// All CALL_DELEGATE instances accross all loaded methods will share a
+  /// workspace. This reduces memory
+  /// consumption by overlapping activation memory between methods but enforces
+  /// synchronization between
+  /// methods. If multiple methods are run concurrently, it may block as only
+  /// one delegate call occur
+  /// at a time. Additionally, the workspace does not shrink when a method is
+  /// unloaded, so memory will
+  /// only be reclaimed when all XNNPACK-delegated methods are unloaded.
+  Global = 2,
+
+  /// The number of workspace sharing modes. This is not a valid mode and is
+  /// only used for tracking the
+  // maximum enum value.
+  Count,
+};
+} // namespace executorch::backends::xnnpack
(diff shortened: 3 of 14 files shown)


backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -86,9 +75,29 @@ class XnnpackBackend final
     const NamedDataMap* named_data_map = context.get_named_data_map();
     // thread safe. This can heppen when multiple threads call init() on
     // the same backend instance.
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
+
+    std::unique_lock<std::mutex> lock(workspace_mutex_, std::defer_lock);
+    if (enable_shared_workspace_) {
+      lock.lock();
```
digantdesai: I guess this will protect us from someone doing `method::init()` on the same program from different threads?

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -160,9 +170,10 @@ class XnnpackBackend final
       // This is needed to serialize access to xnn_delete_runtime which is not
       // thread safe. This can heppen when multiple threads call destroy() on
```
digantdesai: s/heppen/happen

backends/xnnpack/runtime/XNNPACKBackend.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch::backends::xnnpack {
+ET_EXPERIMENTAL void set_workspace_sharing_enabled(bool enable);
```
digantdesai: So from priority, compile_time > runtime? And at runtime, it's one shot at the start up? What about some thread turning it on vs. other off? I guess it's on them?

digantdesai (commented):
Looks good. Please add a test at least to make sure we don't regress on the API and its semantics?

cccclai:
backend options is landed btw

GregoryComer:
> backend options is landed btw

Thanks. I'll rework to use those.

backends/xnnpack/runtime/XNNCompiler.cpp
```diff
@@ -1866,24 +1866,24 @@ ET_NODISCARD Error XNNCompiler::compileModel(
...
+        runtime_flags,
+        &runtime_ptr);
+  } else {
+    status = xnn_create_runtime_v3(
+        subgraph.get(),
+        weights_cache_ptr,
+        ::executorch::extension::threadpool::get_pthreadpool(),
+        runtime_flags,
+        &runtime_ptr);
+  }
```
mcr229: do we need both runtime_v4 and runtime_v3? Can we just pass in nullptr to v4 and it gives us the same results?

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -51,21 +52,9 @@ class XnnpackBackend final
     }
 
 #ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
```
mcr229: do we still need this anymore? Can we just default enable_shared_workspace to false. And rely solely on the runtime to enable workspace sharing?

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -129,9 +138,10 @@ class XnnpackBackend final
       EValue** args) const override {
     auto executor = static_cast<xnnpack::delegate::XNNExecutor*>(handle);
 
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
+    std::unique_lock<std::mutex> lock(workspace_mutex_, std::defer_lock);
```
mcr229: if there are multiple models on the same process, we'd be forced to run all these models serially right? Since each model will have to execute their graph using the same workspace? Should we do some more thinking into how workspaces should be managed among different methods?

GregoryComer: Yeah. I'm not too worried about this for most use cases, as they'll be fighting for compute anyways, but it would be good to not have that limitation. The issue that I'm actually most concerned about is not reclaiming memory.

I'm going to re-work this diff to use backend options, and maybe we should tackle workspace per model at the same time. I'll take a look at the backend interface and propose a way to identify the method that the partition came from.

digantdesai: > Should we do some more thinking into how workspaces should be managed among different methods?

We discussed it here - https://discord.com/channels/1334270993966825602/1336777807509979188/1375180079679668296

GregoryComer:
Reworking this to use the backend options API. It's WIP again. I don't see a way to revert to draft, but I'm validating CI with the initial changes.

GregoryComer:
Changes are code complete. I'm waiting on full CI run, but this should be ready for review again.

CC @mcr229 @digantdesai

backends/xnnpack/runtime/XNNPACKBackend.h
GregoryComer: I've updated for backend options. To maintain the current behavior, I'm setting global mode if ENABLE_XNNPACK_SHARED_WORKSPACE is set, though we probably want to re-evaluate this in the future, as per-model might be better.

The runtime backend option is global, but I've written the code in such way that we can set it per-method/model if/when backend options support this.

backends/xnnpack/runtime/XNNCompiler.cpp
GregoryComer: Updated.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Ideally, we can get rid of it, yeah. I can take that as a follow-up.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: I've reworked the design since this comment, so marking as resolved.

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -86,27 +82,30 @@ class XnnpackBackend final
     const NamedDataMap* named_data_map = context.get_named_data_map();
     // thread safe. This can heppen when multiple threads call init() on
     // the same backend instance.
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
+
+    auto program_id =
+        reinterpret_cast<uintptr_t>(context.get_runtime_allocator());
```
digantdesai, line 84: Feels like needless tangling of assumptions. Can't runtime hide this under an abstraction so delegates don't have to make such assumptions?

If I pass the same `MallocMemoryAllocator` across various methods this pointer could be same for all the models in a given process.

cc @cccclai

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
           weights_cache_mutex_);
       weights_cache_->delete_packed_data(executor->get_packed_data_names());
 #endif
+
+      // This is needed to serialize access to xnn_delete_runtime which is not
+      // thread safe. This can heppen when multiple threads call destroy() on
+      // the same backend instance.
+      auto [lock, _] = executor->get_workspace().acquire();
```
digantdesai: Nit
```suggestion
      auto [_raii_lock, _] = executor->get_workspace().acquire();
```

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+    return Error::Ok;
+  }
+
+  Error set_option(
+      BackendOptionContext& context,
+      const executorch::runtime::Span<executorch::runtime::BackendOption>&
+          backend_options) override {
+    if (backend_options.size() > 0) {
+      for (const auto& option : backend_options) {
+        if (strcmp(option.key, xnnpack::workspace_sharing_mode_option_key) ==
```
digantdesai, line 210: feels like need a bit of restructuring. should we wrap xnnpack options in a wrapper which also implements setter and getter? so then this just becomes a look up and set on it.

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+                  Error,
+                  "XNNPACK workspace sharing mode must be between 0 and %d, inclusive, but was %d.",
+                  static_cast<int>(WorkspaceSharingMode::Count),
+                  *val);
+              return Error::InvalidArgument;
+            }
+
+            ET_LOG(
+                Debug, "Setting XNNPACK workspace sharing mode to %d.", *val);
+            workspace_sharing_mode_ = static_cast<WorkspaceSharingMode>(*val);
```
digantdesai: if this can be called at any time i.e
(1) before init
(2) after init, but before first execute
(3) after first execute
(4) after last execute but before destroy
(5) after destroy but before dector

so should we make a general state machine to find where we are in 1 - 5 and then each option checks if its ok to set or to a particular value given which state we are in.

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
       // XNNExecutor is not trivially destructible. Since this was constructed
       // manually in init(), we must destroy it manually here.
       executor->~XNNExecutor();
     }
   }
 
+  Error get_option(
+      BackendOptionContext& context,
+      executorch::runtime::Span<executorch::runtime::BackendOption>&
+          backend_options) override {
```
digantdesai:
```suggestion
          backend_options) const override {
```

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+      return get_global_workspace();
+    } else {
+      ET_LOG(
+          Error, "Invalid workspace sharing mode: %d.", static_cast<int>(mode));
+      return Error::Internal;
+    }
+  }
+
+  // Retrieve the global workspace, lazy initializing it if needed.
+  Result<std::shared_ptr<XNNWorkspace>> get_global_workspace() const {
```
digantdesai:
```suggestion
  Result<std::shared_ptr<XNNWorkspace>> get_or_create_global_workspace() const {
```

backends/xnnpack/runtime/XNNPACKBackend.h
```diff
@@ -0,0 +1,44 @@
...
+enum class WorkspaceSharingMode {
+  /// No workspace sharing. Each CALL_DELEGATE instance will have its own
+  /// workspace (memory arena).
+  Disabled = 0,
+
+  /// All CALL_DELEGATE instances in a given program will share a workspace.
+  /// This reduces memory consumption
+  /// for methods with multiple delegate calls, at the cost of only allowing one
+  /// method to execute at a time.
+  PerModel = 1,
```
digantdesai, line 26: what about multiple methods in the same PTE running in parallel? and a workspace per method? May be contrived but still want to hear your thoughts.

digantdesai (commented):
still reviewing but want to drop this before I get back to this

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: I agree, generally, though after discussing with the runtime group, this ends up being a larger pain due to thread sychronization in the core. @JacobSzwejbka suggested using signed PTE signatures in the future. Alternatively, we can make larger PAL changes to allow for generating unique IDs in multithreaded environments. But I don't want to be blocked on this, since this feature is somewhat overdue. We should 100% continue to pursue a cleaner solution in parallel, but it should be able to be implemented transparently as a follow-up. Thoughts?

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Yeah, I was thinking about introducing an abstraction around backend options, but didn't want to bloat this PR. I can take this a follow-up, if no objection.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: The current design associates the mode with the executor (implicitly via the XNNWorkspace object), so it should be safe to set in all scenarios. I believe almost all of these scenarios are covered by the tests that load two models while changing mode in between. I can add a little more coverage there, if desired.

backends/xnnpack/runtime/XNNPACKBackend.h
GregoryComer: Yeah, I'd like to do this as a follow-up. I considered adding a per-method (instance) sharing mode (see PR description), but we need a way to identify method instances (as opposed to the name). I think that makes sense as the default behavior going forward, to avoid breaking the use cases that rely on loading the same method multiple times to execute from different threads. Thoughts?

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: I think the base class method is non-const, so we might not be able to do this. I'll double-check.

backends/xnnpack/runtime/XNNPACKBackend.cpp
cccclai: > MallocMemoryAllocator across various methods this pointer

Is it supported? I think each method will have their dedicated allocator, otherwise the memory might be corrupted

digantdesai: > this ends up being a larger pain due to thread sychronization in the core

Not sure I understand but i don't want to block you. Please leave a TODO(#ISSUE) comment and I am ok with moving on for now.


> we can make larger PAL changes to allow for generating unique IDs in multithreaded environments

yeah that seems like the right place to generate a unique id for the runtime based on `uuid = fn(pid, tid, pte_id, method_id, more..)`

>  @JacobSzwejbka suggested using signed PTE signatures in the future.

What if a PTE loaded twice by the same process but don't want to share workspace.

>  I think each method will have their dedicated allocator, otherwise the memory might be corrupted

Can you elaborate? IIUC this is just malloc() under the hood. `a->malloc()` twice vs. `a->malloc()` and `b->malloc()`. You can cause the internal vector.emplace to cause race but if you lock at the method load not sure what can be a problem.

backends/xnnpack/runtime/XNNPACKBackend.cpp
digantdesai: TODO(#issue) please

backends/xnnpack/runtime/XNNPACKBackend.cpp
digantdesai: I have to read the code again but I am thinking doing something silly like changing a mode through backend option on every inference, would that work as expected?

backends/xnnpack/runtime/XNNPACKBackend.cpp
digantdesai: write a local `_get_option(..) const` and call it from this? We want compiler to make sure we don't accidently update the option in this method.

backends/xnnpack/runtime/XNNPACKBackend.h
digantdesai: Sounds good.

backends/xnnpack/test/runtime/test_workspace_sharing.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::optional<WorkspaceSharingMode> mode1,
+    std::optional<WorkspaceSharingMode> mode2) {
+  // Load and run two models, verifying that the output tensors are correct,
+  // optionally setting sharing mode.
+
+  if (mode1) {
+    set_and_check_workspace_sharing_mode(*mode1);
+  }
+
+  Module mod1(std::getenv("ET_XNNPACK_GENERATED_ADD_LARGE_PTE_PATH"));
```
digantdesai, line 119: How will it work on the OSS CI?

I know the pte generation is only in buck IIRC, but we should fix that. Worst case we can use some HF uploaded PTEs.

backends/xnnpack/runtime/XNNWorkspace.h
```diff
@@ -0,0 +1,60 @@
...
+    if (status != xnn_status_success) {
+      ET_LOG(
+          Error,
+          "Failed to create XNN workspace, XNNPACK status: 0x%x",
+          (unsigned int)status);
+      return runtime::Error::Internal;
+    }
+
+    return std::make_shared<XNNWorkspace>(
+        WorkspacePtr(workspace, &xnn_release_workspace));
```
digantdesai, line 59: I think this works well (1) for global when the ref count goes to 0, and (2) for per_model when the individual user ref count goes to 0. If we are switching mode, then when will these "duplicated" workspaces will get released? Should we explicitly detect a mode switch and release the old one? This will make switch cost high but peak memory low.

digantdesai: Alternative is to not allow switching after init :p

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+    return global_workspace_;
+  }
+
+  // Get or create a workspace for the given program ID.
+  Result<std::shared_ptr<XNNWorkspace>> get_model_workspace(
+      uintptr_t program_id) const {
+    std::scoped_lock<std::mutex> lock(workspace_meta_mutex_);
+
+    // Check for an existing (live) workspace for this program.
+    auto match = model_workspaces_.find(program_id);
```
digantdesai: Thinking out loud, worst case if we have collision in the program_id we will end up collobering memory during inference since two runtimes will go update the workspace.

backends/xnnpack/test/runtime/test_workspace_sharing.cpp
```diff
@@ -0,0 +1,169 @@
...
+  // the other mode.
+
+  std::array<WorkspaceSharingMode, 3> modes = {
+      WorkspaceSharingMode::Disabled,
+      WorkspaceSharingMode::PerModel,
+      WorkspaceSharingMode::Global};
+
+  for (auto i = 0; i < modes.size(); ++i) {
+    for (auto j = i + 1; j < modes.size(); ++j) {
+      run_and_validate_two_models(modes[i], modes[j]);
```
digantdesai, line 95: Nice!

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+    std::shared_ptr<XNNWorkspace> workspace = {};
+    if (match != model_workspaces_.end()) {
+      if (auto live_workspace = match->second.lock()) {
+        workspace = live_workspace;
+      }
+    }
+
+    // Allocate a new workspace if needed.
+    if (!workspace) {
+      ET_LOG(Debug, "Creating workspace for program %" PRIuPTR ".", program_id);
```
digantdesai: Nit move this down and print the workspace ptr?

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -175,32 +165,179 @@ class XnnpackBackend final
...
+    }
+  }
+
+  // Retrieve the global workspace, lazy initializing it if needed.
+  Result<std::shared_ptr<XNNWorkspace>> get_global_workspace() const {
+    std::scoped_lock<std::mutex> lock(workspace_meta_mutex_);
+
+    // Lazy init global workspace.
+    if (!global_workspace_) {
+      ET_LOG(Debug, "Instantiating global workspace.");
```
digantdesai: Nit: move this down and print `gloabla_workspace_` ptr

digantdesai (commented):
Some more comments. I will work with you to make sure we can land this today

backends/xnnpack/runtime/XNNPACKBackend.cpp
cccclai: Are a and b the allocator used by two methods? And `a->malloc` means the allocator used by method A and `b->malloc` means the allocator used by method B?

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13189.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13190. I added some brief comments in the issue, but it would be nice create a reusable abstraction that multiple backends can leverage. But I don't know if we can get it in the core runtime (space + embedded constraints), and I don't want to add it as an extension (I don't think it makes sense). Any thoughts?

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: I'll add a little more test coverage for this case. It should work, as the workspace is effectively "locked" at load time. So the option only comes into play loading the model. Even if it's changed later, it will use the workspace shared_ptr stored in the executor instance.

backends/xnnpack/runtime/XNNPACKBackend.h
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13192.

backends/xnnpack/runtime/XNNWorkspace.h
GregoryComer: My intent is that workspaces get freed whenever all users are unloaded. Each executor instance holds a shared_ptr, and the top-level backend uses weak pointers to allow them to get cleaned up automatically. It would be nice to assert this behavior in a test. I might refactor the workspace management logic into a dedicated class to make this easier.

The global workspace is currently not released, but it probably should be. I'll update to hold a weak pointer and re-create when needed.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Yeah, I don't love this, but it should be safe in all cases, at least. Either you end up using more memory than needed or it enforces extra synchronization. I'd like to push more on the method/program ID.

backends/xnnpack/test/runtime/test_workspace_sharing.cpp
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13194.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Updated.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Updated.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Done.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Done.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Done.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: I've added some extra coverage for running while changing modes.

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Updated.

backends/xnnpack/runtime/XNNWorkspace.h
GregoryComer: Updated to allow release and re-creation of the global workspace.

backends/xnnpack/runtime/XNNPACKBackend.cpp
digantdesai: a, b are pointers to `MallocMemoryAllocator` instance(s).

backends/xnnpack/runtime/XNNWorkspaceManager.h
```diff
@@ -0,0 +1,94 @@
...
+  // A mutex guarding global_workspace_ and model_workspaces_. Note that this
+  // mutex only guards the top-level definitions, not the contents of the
+  // workspace. The contents of the workspace are guarded by the workspace's own
+  // mutex in the XNNWorkspace class.
+  mutable std::mutex workspace_meta_mutex_;
+
+  // A global workspace for all delegate instances, if global sharing is
+  // enabled. Lazy initialized. Stored as a weak pointer to allow automatic
+  // cleanup when all references are released.
+  mutable std::weak_ptr<XNNWorkspace> global_workspace_;
```
digantdesai, line 75: nit: can this be in `model_workspaces_` with `program_id<uintptr_t> == 0` or something, this will simplify `get_or_create` methods, and also other logic which treats global workspace different when its is just a shared `program_id` in reality.

backends/xnnpack/runtime/XNNPACKBackend.cpp
digantdesai: OK with this for now, as discussed offline with Gregory.

backends/xnnpack/test/runtime/test_workspace_sharing.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::optional<WorkspaceSharingMode> mode1,
+    std::optional<WorkspaceSharingMode> mode2) {
+  // Load and run two models, verifying that the output tensors are correct,
+  // optionally setting sharing mode.
+
+  if (mode1) {
+    set_and_check_workspace_sharing_mode(*mode1);
+  }
+
+  Module mod1(std::getenv("ET_XNNPACK_GENERATED_ADD_LARGE_PTE_PATH"));
```
digantdesai, line 119: nit mod/mode/model/module :P

```suggestion
  Module module_for_pte_1(std::getenv("ET_XNNPACK_GENERATED_ADD_LARGE_PTE_PATH"));
```

backends/xnnpack/runtime/XNNPACKBackend.cpp
```diff
@@ -84,29 +76,32 @@ class XnnpackBackend final
...
-    // thread safe. This can heppen when multiple threads call init() on
+    // thread safe. This can happen when multiple threads call init() on
     // the same backend instance.
-#ifdef ENABLE_XNNPACK_SHARED_WORKSPACE
-    const std::lock_guard<std::mutex> lock(workspace_mutex_);
-#endif
+
+    auto program_id =
+        reinterpret_cast<uintptr_t>(context.get_runtime_allocator());
+    auto workspace = ET_UNWRAP(get_or_create_workspace(program_id));
```
digantdesai, line 85: so for a given delegate once a workspace is created, there won't be any impact of subsequent mode changes in the process, until it is delegated right?

digantdesai (approved):
Thanks @GregoryComer

backends/xnnpack/runtime/XNNPACKBackend.cpp
GregoryComer: Yeah, the workspace is effectively "locked in" at the time of the delegate instance creation. So it should remain in that mode indefinitely. That seemed like the easiest way to handle the global mode option safely, though I'm open to suggestions.

PR #11758, opened 2025-06-17
[5/N] Add get_option/set_option APIs

cccclai:
* __->__ #11758
* #11462
* #11391
* #11390
* #11389

Expose the API to set/get backend option. We can either pass in {backend_name, backend options} or {backend options map}

Differential Revision: [D76825663](https://our.internmc.facebook.com/intern/diff/D76825663/)

Differential Revision: [D76825663](https://our.internmc.facebook.com/intern/diff/D76825663)


runtime/backend/options_map.h
```diff
@@ -82,5 +84,106 @@ class BackendOptionsMap {
...
+/**
+ * Retrieves backend options for a specific backend.
+ *
+ * @param backend_name The name of the backend to get options from
+ * @param backend_options The backend option objects that will be filled with
+ * the populated values from the backend
+ * @return Error::Ok on success, Error::NotFound if backend is not found, or
+ * other error codes on failure
+ */
+ Error get_option(
```
kimishpatel, line 97:
```suggestion
 Error get_options(
```

runtime/backend/options_map.h
```diff
@@ -82,5 +84,106 @@ class BackendOptionsMap {
...
+/**
+* Retrieves backend options for multiple backends using a backend options map.
+*
+* @param backend_options_map The backend option map containing backend names
+* and their associated options, which will be filled with the populated values
+* from the backend
+* @return Error::Ok on success, or the first error encountered when processing
+* the entries
+*/
+Error get_option(
```
kimishpatel, line 125: I dont think we need this. If users want something like that, they can implement it or we wait to see if users want options like that and introduce it later.

kimishpatel (changes requested):
this looks good. My suggestion though is that not have backendOptionsMap and related functionality right now. Users can either make convenient functions for themselves or just use set_option(name, options) for the backends they care for. And maybe just wait and see if this is something users want as part of et

runtime/backend/options_map.h
cccclai: Do you mean get_option function or the get_option with backend option map?

kimishpatel: with backend options map, is what i meant

PR #11760, opened 2025-06-17
dtype selective build from model API in OSS

BujSet:
### Motivation
Given a specific model, we want to produce a binary that only includes the minimal operators and dtypes needed to run the model. This requires parsing the model to determine what kernels it launch, the operators used in those kernels, and the dtypes of the tensors in the kernels. After parsing, a header file must be generated, and the portable_kernels lib can be rebuilt to only include the operators and dtypes specified in the generated header.

### Summary
This changes completes this E2E process. A user can now specify the model they wish to optimize their binary for via the command line argument `-DEXECUTORCH_SELECT_OPS_FROM_MODEL="<file path to model pte>"`. When specified, the pte is parsed to produce a YAML file called `seleced_operators.yaml` which describes the model's operators and dtypes. From this YAML, a header file called `selected_op_variants.h` is generated that selects the described operators and dtypes. When command line argument `-DEXECUTORCH_DTYPE_SELECTIVE_BUILD=ON` is specified, the header file is linked to the `portable_kernels` lib when it's rebuilt. Only the model API is supported with dtype selective build, and using other methods such as `list` or `dict` will results in a build error. 

### Results

An example usage of this flow is included in `examples/selective_build/test_selective_build.sh:test_cmake_select_ops_in_model`. When run as `bash examples/selective_build/test_selective_build.sh cmake`, the `cmake-out/examples/selective_build/selective_build_test` binary is built. After stripping the binary, the following binary size results were seen with the following models:

#### Working models

| Model | Default Binary Size (KB) | Dtype Selected Binary Size (KB) |
| ------- | :---:| :---:|
|add | 359 | 275|
|mul | 335 | 263 | 
| add_mul| 367 | 287 |
| linear | 347 | 291 |
| softmax | 251 | 251 |
|resnet18| 643 | 515 |
|resnet50 | 643 | 515 |
|mobilebert| 707 | 415 | 
|lstm| 643 | 459|
| dl3| 607 | 539|
|edsr | 543 | 371|

#### Models that crash when run

| Model | Default Binary Size (KB) | Dtype Selected Binary Size (KB) |
| ------- | :---:| :---:|
|emformer_transcribe| 863| 555| 
|vit| 907| 687 |
|mv2| 631 | 495 |
|mv3| 843 | 535 |
|llama | 1.1M | 827|
|qwen2_5 | 1.1M | 827|

### Notes
Although there is a noted reduction in the binary size, it seems that the pte file parsing functionality from `gen_oplist.py` is incomplete. Please see the discussion on [PR #11582](https://github.com/pytorch/executorch/pull/11582) and on [issue #11762](https://github.com/pytorch/executorch/issues/11762) for more details.


tools/cmake/Codegen.cmake
```diff
@@ -104,6 +108,13 @@ function(generate_bindings_for_kernels)
   # By default selective build output is selected_operators.yaml
   set(_oplist_yaml ${_out_dir}/selected_operators.yaml)
 
+  # If dtype selective build is enable, force header file to be preserved
+  if(GEN_DTYPE_SELECTIVE_BUILD)
+    set(_opvariant_h ${_out_dir}/selected_op_variants.h)
```
jathu, line 115: Why does the location of this differ from what's used in `gen_selected_ops`?

tools/cmake/Codegen.cmake
```diff
@@ -191,29 +203,92 @@ endfunction()
...
+    if(GEN_DTYPE_SELECTIVE_BUILD)
+      if("${portable_kernels_check}" IN_LIST GEN_KERNEL_LIBS)
+        list(REMOVE_ITEM GEN_KERNEL_LIBS ${portable_kernels_check})
+
+        # Define portable kernels sources (same as in kernels/portable/CMakeLists.txt)
+        file(GLOB_RECURSE _portable_kernels_srcs
+             "${EXECUTORCH_ROOT}/kernels/portable/cpu/*.cpp"
+        )
+        list(FILTER _portable_kernels_srcs EXCLUDE REGEX "test/*.cpp")
+        list(FILTER _portable_kernels_srcs EXCLUDE REGEX "codegen")
```
jathu: Why not just set the variable in `kernels/portable/CMakeLists.txt` and reuse it here? Duplicating code like this implicitly causes a dependency and can drift

tools/cmake/Codegen.cmake
```diff
@@ -191,29 +203,92 @@ endfunction()
...
+        add_custom_target(selected_portable_kernels_header DEPENDS ${_opvariant_h})
+        add_dependencies(selected_portable_kernels selected_portable_kernels_header)
+        # Apply the compile definition for dtype selective build
+        target_compile_definitions(selected_portable_kernels PRIVATE EXECUTORCH_SELECTIVE_BUILD_DTYPE=1)
+
+        target_link_libraries(${GEN_LIB_NAME} PUBLIC selected_portable_kernels)
+      endif()
+    endif()
+
+    if(GEN_KERNEL_LIBS)
```
jathu, line 277: This is unnecessary I think — aren't we already in an `if(GEN_KERNEL_LIBS)` block from line 241?

tools/cmake/Codegen.cmake
```diff
@@ -191,29 +203,92 @@ endfunction()
...
+      if(GEN_DTYPE_SELECTIVE_BUILD)
+        message(FATAL_ERROR "Currently dtype selective build is only supported for portable_kernels but {${GEN_KERNEL_LIBS}} were provided!")
+      else()
```
jathu: What if we run this check way earlier in the function?

tools/cmake/Codegen.cmake
BujSet: It's the same location, just in `generate_bindings_for_kernels` someone created a variable for it as `_out_dir`, see line 107.

tools/cmake/Codegen.cmake
BujSet: The issue is that if `portable_kernels` is listed in `GEN_KERNEL_LIBS`, we remove it from the list (line 243). I thought it might be possible that there may be multiple options passed in here, so this check ensure that if others are specified, they still get linked. I'm not entirely sure if this use case is probable though?

tools/cmake/Codegen.cmake
jathu: It's defined as `${CMAKE_CURRENT_BINARY_DIR}/${GEN_LIB_NAME}/selected_op_variants.h` in `gen_selected_ops`. However, it's defined as `${_out_dir}/selected_op_variants.h` here. Let's be consistent?

BujSet: Line 107 has `set(_out_dir ${CMAKE_CURRENT_BINARY_DIR}/${GEN_LIB_NAME})` so that in the rest of `generate_bindings_for_kernels` can use `_out_dir` instead of specifying the full path. Seems like the same was done in `gen_custom_ops_aot_lib` and `gen_operators_lib`. I could replace it with the full path, but maybe it would make sense to define `_out_dir` in `gen_selected_ops` and use it there as well. What do you think?

jathu: ahh I didn't catch that, thats fine then

jathu (approved):

BujSet:
Tested with openai's whisper-tiny model. Dtype selection helps bring the binary size down to 523 KB, but it still crashes based on the same issues mentioned above. 

cc @lucylq

PR #11779, opened 2025-06-18
Implement python deserialize for flat_tensor

GregoryComer:
### Summary
Implement the deserialize function in the Python FlatTensorSerializer class. This allows for loading flatbuffer data from a Python environment.

### Test plan
I've added an additional test in `extension/flat_tensor/test/test_serialize.py` to cover the deserialize path.

### DIFF
--- a/extension/flat_tensor/test/test_serialize.py
+++ b/extension/flat_tensor/test/test_serialize.py
@@ -6,17 +6,19 @@
 
 # pyre-unsafe
 
+import dataclasses
 import math
 import unittest
 
 from typing import List, Optional
+
+from executorch.exir._serialize._cord import Cord
 
 from executorch.exir._serialize.data_serializer import (
     DataEntry,
     DataPayload,
     DataSerializer,
 )
-
 from executorch.exir._serialize.padding import aligned_size
 
 from executorch.exir.schema import ScalarType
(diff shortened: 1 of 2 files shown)


extension/flat_tensor/test/test_serialize.py
```diff
@@ -223,3 +225,32 @@ def test_serialize(self) -> None:
         )
 
         self.assertEqual(segments[2].offset + segments[2].size, len(segment_data))
+    
+    def test_round_trip(self) -> None:
```
GregoryComer, line 229: Note to reviewers:  I considered adding a deserialize test in the same vein as the serialize test, but it's a bit awkward to do, in that we'd need to construct a binary input without using the existing serialization logic, Could also include a pre-serialized artifact, but I think this coverage is sufficient in conjunction with the serialize test. Feel free to let me know if you want to see more coverage.

extension/flat_tensor/test/test_serialize.py
```diff
@@ -223,3 +225,39 @@ def test_serialize(self) -> None:
...
+        # properly.
+        config = FlatTensorConfig()
+        serializer: DataSerializer = FlatTensorSerializer(config)
+
+        # Round trip the data.
+        serialized_data = bytes(serializer.serialize(TEST_DATA_PAYLOAD))
+        deserialized_payload = serializer.deserialize(Cord(serialized_data))
+
+        # Validate the deserialized payload. Since alignment isn't serialized, we need to
+        # do this somewhat manually.
```
lucylq, line 240: I don't think there's currently a need to serialize the alignment(?)
It might be good to have for deserialization/general error-checking, I guess the trade-off is binary size.

extension/flat_tensor/test/test_serialize.py
```diff
@@ -223,3 +225,39 @@ def test_serialize(self) -> None:
...
+        SKIP_FIELDS = {"alignment"}  # Fields to ignore in comparison.
+        for key in TEST_DATA_PAYLOAD.named_data.keys():
+            reference = TEST_DATA_PAYLOAD.named_data[key]
+            actual = deserialized_payload.named_data[key]
+
+            for field in dataclasses.fields(reference):
+                if field.name not in SKIP_FIELDS:
+                    self.assertEqual(
+                        getattr(reference, field.name),
+                        getattr(actual, field.name),
```
lucylq, line 261: this is nice!

lucylq (approved):
LGTM, thank you for adding this!

PR #11392, opened 2025-06-05
Qualcomm AI Engine Direct - Deprecate convert_bmm_to_matmul pass

DannyYuyang-quic:
### Summary
- delete convert_bmm_to_matmul pass
- add torch.ops.aten.matmul.default in skip_decomp_table

### Test plan
General CI

PR #11782, opened 2025-06-18
Qualcomm AI Engine Direct - Delegate mutable buffer and fix the mutable buffer issue

shewu-quic:
Summary:
- Add a parameter to support mutable buffer delegation in QNN Backend
  - Set the same memory address for I/O of mutable buffer at runtime
  - Ref: https://github.com/pytorch/executorch/pull/6727
- Avoid annotating the input node because mutable buffers will be folded during the convert_pt2e process.
- Deprecated use_legacy_export in executorch llama

cc @cccclai @winskuo-quic @cbilgin


cccclai:
> Avoid annotating the input node because mutable buffers will be folded during the convert_pt2e process.

Is the input node still folded after we land https://github.com/pytorch/ao/pull/2345?

shewu-quic:
> > Avoid annotating the input node because mutable buffers will be folded during the convert_pt2e process.
> 
> Is the input node still folded after we land [pytorch/ao#2345](https://github.com/pytorch/ao/pull/2345)?

Yes, unless we apply run_decomposition after export. I think we can wait until run_decomposition becomes a pass and doesn't require re-tracing. After that we can change it back to annotate mutable buffer. What do you think?

shewu-quic:
BTW, in previous, we have submitted [a PR to deprecated convert_bmm_to_matmul pass](https://github.com/pytorch/executorch/pull/11392). It will result in multiple partitions for Meta's llama due to not using `to_edge_transform_and_lower_to_qnn`. So, I add it back and set False for activation as default value.

cccclai (approved):

cccclai:
I ran into this error for some internal use cases.
```
TypeError: LLMEdgeManager.__init__() got an unexpected keyword argument 'use_legacy_export'
```
Can we turn if off in this PR, and I will have another PR to remove this arg?

PR #11784, opened 2025-06-18
NXP backend: Add quantization of aten.view

robert-kalmar:
### Summary
Updates part of this PR: 
- Add support for quantization and conversion of the aten.view operator to Neutron Backend. 
- Fix avegage pooling, which in case of kernel (1,1) turns into Sum in Neutron IR. 

### Test plan
Unit test part of the change.
```
pytest -c /dev/null backends/nxp/tests/
```

cc @digantdesai @JakeStevens


JakeStevens:
nit: can you split this diff? It seems to be two completely separate changes, aten view quantization and the removal of avg pool/linear optimization

JakeStevens:
LGTM

JakeStevens (approved):

backends/nxp/backend/ir/tflite_optimizer/optimizer.py
```diff
@@ -164,9 +160,6 @@ def __init__(
             Optimization.COMBINE_HARD_SIGMOID_AND_MUL_INTO_HARD_SWISH: CombineHardSigmoidAndMulIntoHardSwish(
                 builder, conversion_config
             ),
-            Optimization.REPLACE_AVERAGE_POOL_BEFORE_FULLY_CONNECTED_WITH_SUM: ReplaceAveragePoolBeforeFullyConnectedWithSum(
```
digantdesai, line 167: How is this related to the view support?

backends/nxp/tests/ir/converter/node_converter/test_constant_pad_nd_converter.py
```diff
@@ -47,37 +39,6 @@ def test_constant_pad_nd_conversion__specific_constant(constant):
     convert_run_compare(edge_program, input_data)
 
 
-@pytest.mark.parametrize("constant", [0.0, 67.28, 42.0, -13.37])
-@pytest.mark.skip(reason="Neutron Converter does not fully convert for NPU")
-def test_constant_pad_nd_quant_conversion__specific_constant(mocker, constant):
```
digantdesai, line 52: Unrelated change?

digantdesai (approved):

backends/nxp/tests/ir/converter/node_converter/test_constant_pad_nd_converter.py
robert-kalmar: You are right, this should have been a separate PR to remove irrelevant test. We will pay more attention when rebasing from our development tree.

backends/nxp/backend/ir/tflite_optimizer/optimizer.py
robert-kalmar: Same here, not related to aten.view. Our fault. 
It is indeed a separate commit, and the PR descrition should have included the proper description: 
"Fix avegage pooling, which in case of kernel (1,1) turns into Sum in Neutron IR"

At least added in to the PR description.

PR #11793, opened 2025-06-18
Add last_token_pos in llama_transformer

sdwilsh:
Summary:
Add last_token_pos in the forward options.

Purpose:
* the last norm and output of lm-head can be performed with the last valid token at prefill.
* If the input sequence length is fixed when an accelerator doesn't support the dynamic shapes, selecting the last token from the input is not always guaranteed as valid.
* Thus, it needs an additional pointer to select the last valid token only to perform the last norm and output.

Differential Revision: D76440105


JacobSzwejbka (approved):
@larryliu0820 any thoughts?

sdwilsh:
I'm not the author of this code!  Closing it...

PR #11800, opened 2025-06-18
Pull c10 headers directly from PyTorch internally, not from the c10 mirror

swolchok:
* __->__ #11800

Previously, both internal and external worlds got ExecuTorch's pinned version of PyTorch. Now external is the same, but internal gets its monorepo behavior. PyTorch guarantees backward compatibility (i.e., not breaking the builds of existing PyTorch users), so our ExecuTorch code that works with pinned PyTorch should also automatically work with internal PyTorch. (bc breaks will be difficult, but that's inherent to bc breaks.)

Differential Revision: [D76832520](https://our.internmc.facebook.com/intern/diff/D76832520/)

### DIFF
--- a/runtime/core/portable_type/c10/c10/targets.bzl
+++ b/runtime/core/portable_type/c10/c10/targets.bzl
@@ -49,43 +49,6 @@ def define_common_targets():
     The directory containing this targets.bzl file should also contain both
     TARGETS and BUCK files that call this function.
     """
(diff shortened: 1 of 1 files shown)


swolchok:
test-moshi-linux was broken on trunk around the time this was sent out

janeyx99 (approved):

PR #11809, opened 2025-06-19
Ability to specify full file configs for export_llm

jackzhxng:
Hydra is a bit opinionated on how config files should be provided to the CLI, this adds some tooling so that we can specify a config file from any arbitrary location to use with export_llm.

Example:
```
python -m extension.llm.export.export_llm --config ...some-path.../example_llm_config.yaml
```


jackzhxng:
* #11836
* #11811
* #11833
* __->__ #11809

extension/llm/install_requirements.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Install requirements for LLM extension
+pip install hydra-core>=1.3.0 omegaconf>=2.3.0
```
mergennachin: I think we should just move these to https://github.com/pytorch/executorch/blob/main/requirements-dev.txt

extension/llm/export/README.md
```diff
@@ -0,0 +1,145 @@
...
+- **Phi**: Phi-3-Mini, Phi-4-Mini
+- **Stories**: Stories110M (educational model)
+- **SmolLM**: SmolLM2
+
+## Installation
+
+First, install the required dependencies:
+
+```bash
+./extension/llm/install_requirements.sh
```
mergennachin: Possibly no need

mergennachin (commented):
As part of the 0.7, one deliverable would be to swap the ci with the new API. Not in this PR though

https://github.com/pytorch/executorch/blob/496cb05668d5b29c471adab02236af709e2702b9/.ci/scripts/test_llama.sh#L254

mergennachin (approved):

PR #11810, opened 2025-06-19
Get rid of llm_config enums

jackzhxng:


jackzhxng:
* #11811
* __->__ #11810
* #11809

mergennachin:
Why is this better? It's harder to make mistakes when types are enums, as opposed to strings.

jackzhxng:
Sure, let's do it this way then - https://github.com/pytorch/executorch/pull/11833

PR #11816, opened 2025-06-19
Arm backend: Fix bug of inserting unnecessary casts for aten.where.self

YufengShi-dudu:
- In MatchWhereSelfDtypePass, target_dtype was initialized with fp32. This works when at least one of the inputs is fp32. But when both inputs are int32, the pass will incorrectly insert int32->fp32 casts. These casts are unnecessary and may introduce operand dtype mismatch issues.
- Fix it by initializing target_dtype with input_dtype.

Change-Id: Id67aed8e90f886dc2f2491946847ad01702d5aa5


YufengShi-dudu:
Unrelated fail: macos-job
@zingo

zingo (approved):

PR #11838, opened 2025-06-21
Add instructions to enable ccache

mergennachin:
cc @larryliu0820 @jathu


install_executorch.py
```diff
@@ -50,6 +50,19 @@ def clean():
...
+    try:
+        result = subprocess.run(["ccache", "--version"], capture_output=True, text=True)
+        if result.returncode == 0:
+            print("Cleaning ccache...")
+            subprocess.run(["ccache", "--clear"], check=True)
+            print("ccache cleared successfully.")
+        else:
+            print("ccache not found, skipping ccache cleanup.")
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        print("ccache not found, skipping ccache cleanup.")
```
jathu, line 63: Why do we need to do this? Also, can we do this in the cmake level so that every build benefits from it?

CMakeLists.txt
```diff
@@ -48,6 +48,16 @@
...
+# Enable ccache if available
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+  set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  message(STATUS "ccache found and enabled for faster builds")
+else()
+  message(STATUS "ccache not found, builds will not be cached")
+endif()
+
```
jathu:
```suggestion
announce_configured_options(CCACHE_PROGRAM)
```

install_executorch.py
mergennachin: This is only called for running 

install_executorch.sh --clean

CMakeLists.txt
```diff
@@ -48,6 +48,16 @@
...
+# Enable ccache if available
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+  set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  message(STATUS "ccache found and enabled for faster builds")
+else()
+  message(STATUS "ccache not found, builds will not be cached")
+endif()
+announce_configured_options(CCACHE_PROGRAM)
```
jathu: oops you need to move this below line 66 cause we import this function later. I'd suggest to move this down to line 98 after we've configured some common requirements like python and buck

jathu (approved):

PR #11842, opened 2025-06-22
Fix pip install . -e when buck2 is not installed

mergennachin:
If the installation goes through regular './install_executorch.sh' path, it sets the -DBUCK2 cmake config, but if you directly call `pip install .` it 
doesn't resolve buck dependency.

Test Plan:

Run 

- `./install_requirement.sh` and `pip install . --no-build-isolation -v`
- `./install_requirement.sh` and `pip install -e . --no-build-isolation -v`
- ./install_executorch.sh

cc @larryliu0820 @jathu


setup.py
```diff
@@ -672,6 +672,28 @@ def run(self):  # noqa C901
...
+                text=True,
+                check=True,
+            )
+            buck2_path = result.stdout.strip()
+            cmake_configuration_args.append(f"-DBUCK2={buck2_path}")
+            log.info(f"Resolved buck2 path for cmake: {buck2_path}")
+        except subprocess.CalledProcessError as e:
+            log.info(f"Failed to resolve buck2 (exit code {e.returncode}): {e.stderr}")
+            log.info(f"resolve_buck.py stdout: {e.stdout}")
+            # Continue without explicit buck2 path, let cmake handle it
```
jathu: We already do this immediately during the cmake build: 

https://github.com/pytorch/executorch/blob/608a745fd061e0f1ead2a186fca3fabc6271b7a4/CMakeLists.txt#L78-L81

jathu (approved):

PR #11847, opened 2025-06-23
Arm backend: Add missing bias in pass

oscarandersson8218:
Adds a pass which adds missing bias for convolution. Removes handling of missing bias in conv2d-visitor.


zingo (approved):

zingo:
Macos fails unrelated

PR #11848, opened 2025-06-23
Arm backend: Add sinh decomposition pass and test

tbergkvist:
Decomposes sinh into other operators/lookup table for MI/BI case.


zingo:
Nice work!
This seem to have got a merge confict in 
arm_pass_manager.py
thats needs to be adressed :(

oscarandersson8218 (approved):
Looks good to me!

PR #11852, opened 2025-06-23
Arm backend: Move Ethos-U backend to generate TOSA-1.0

oscarandersson8218:
## Summary
Move to use TOSA 1.0 for Vela compiler to consume. Vela compiler is updated to have support for TOSA 1.0.

## Test plan
Internal and external CI unit tests.


zingo:
@digantdesai and @mergennachin this one might be good to check and maybe test internally.
Vela version is updated.

zingo:
@digantdesai did you get any problem with this?

zingo:
qnn and models fails are unrelated, they fail on main alos

digantdesai:
> @digantdesai did you get any problem with this?

Ah yeah sorry forgot to reply. We get 
`RuntimeError: Compilation failed: TOSA FB Reader error : Tosa version mismatch`.

I am updating vela right now. Let's hope to unblock you guys on this ASAP.

zingo:
> I am updating vela right now. Let's hope to unblock you guys on this ASAP.

Thanks!

examples/arm/setup.sh
```diff
@@ -56,7 +56,7 @@ fi
 
 # vela
 vela_repo_url="https://gitlab.arm.com/artificial-intelligence/ethos-u/ethos-u-vela"
-vela_rev="8cac2b9a7204b57125a8718049519b091a98846c"
+vela_rev="d37febc1715edf0d236c2ff555739a8a9aadcf9a"
```
digantdesai, line 59: So IIUC this is a BC breaking change for Vela i.e. this new version can't run both v0.80 and v1.0, right since they hardcode the schema?

zingo: Thats my understanding, old version 0.80.1 only and new version TOSA 1.0 only.

zingo:
@digantdesai do you still see any problems with this?

digantdesai:
Yeah I am running into asserts because of flatbuffer issues internally. I will keep you posted.

digantdesai:
Yeah running into asserts because of flatbuffer issues internally. Still debugging.. 

For instance, like this opaque error,
```
RuntimeError: Compilation failed: TOSA FB Reader error: Failed to load TOSA model. Buffer contents inconsistent with generated schema
```
Validating that the `.tosa` and `.json` file matches between two setups. 
If they do then it has to do something to do with the Vela buck-ification I assume?

I will keep you posted.

digantdesai:
#12780 merged.

PR #11872, opened 2025-06-24
Move LlmConfig to extension/llm

jackzhxng:
Moves `LlmConfig` from `examples/models/llama/config` to `extension/llm/export/config`


digantdesai (commented):
LGTM, this is just a mechanical move? i.e. no functional change otherwise?

Also I assume moving this in extensions/llm is been through a review of sorts?

mergennachin (approved):

lucylq (approved):

PR #11873, opened 2025-06-24
Arm backend: Make per-channel quantization default

martinlsm:
Support for per-channel quantization was recently added to the Arm
backend. This patch changes the default setting to use per-channel
quantization for weights in convolutional and linear layers, instead of
per-tensor quantization, which was the previous default.

The reason for this change is that per-channel quantization offers
better numerical accuracy for models containing convolutional and/or
fully connected layers. Unless there is an explicit limitation in the
use case that prevents the use of per-channel quantization, it is
generally preferred.

The option to set quantization granularity can still be manually set
using `get_symmetric_quantization_config(is_per_channel=False)`. This
patch only changes the default.

Unit and model tests are affected by this change. Error tolerances for
those tests have not been changed, as model outputs are compared against
a reference that uses the exact same quantization strategy. That is, if
a model output is altered by this patch, the reference it is compared
against would also be altered accordingly.

To verify the impact of this change in terms of top-1 and top-5
accuracy, a manual test was run on MobileNetV2. The results show a
noticeable improvement:

- Per-tensor quantization Top-1 / Top-5 accuracy: 66.45% / 87.50%
- Per-channel quantization Top-1 / Top-5 accuracy: 70.85% / 89.50%


jackzhxng:
@martinlsm no need to add `"topic: not user facing"` anymore btw

examples/arm/ethos_u_minimal_example.ipynb
```diff
@@ -101,7 +101,7 @@
     "\n",
     "# Create and configure quantizer to use a symmetric quantization config globally on all nodes\n",
     "quantizer = EthosUQuantizer(compile_spec)\n",
-    "operator_config = get_symmetric_quantization_config(is_per_channel=False)\n",
+    "operator_config = get_symmetric_quantization_config(is_per_channel=True)\n",
```
digantdesai: Nit
```suggestion
    "operator_config = get_symmetric_quantization_config()\n",
```

digantdesai (commented):
I assume rationale is accuracy but can you please add more details? And linear and conv weights only right? perf impact? I also didn't see any ATOL/RTOL change in this diff.

examples/arm/ethos_u_minimal_example.ipynb
martinlsm: @digantdesai I have resolved your code comment and answered all your questions in the updated commit message.

digantdesai: Thank you!

digantdesai (approved):
I am glad to see we didn't have to mess with the backend code at all when quantizing the weights differently. Thanks @martinlsm.

oscarandersson8218:
> I am glad to see we didn't have to mess with the backend code at all when quantizing the weights differently. Thanks @martinlsm.

@digantdesai we did mess with the backend code but in a separate PR https://github.com/pytorch/executorch/pull/11752.

digantdesai:
> #11752 

LOL That's what I thought :D

Are we ready to merge this one then?

.github/workflows/trunk.yml
```diff
@@ -197,7 +197,7 @@ jobs:
       docker-image: executorch-ubuntu-22.04-arm-sdk
       submodules: 'recursive'
       ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 90
+      timeout: 120
```
oscarandersson8218, line 200: @digantdesai any objections to this? We were just a few minutes off the 90 min timeout with per-tensor quantization. Now it seems that we exceed that timeout with per-channel quantization by default.

digantdesai: No it should be ok.

Sebastian-Larsson (approved):
Unrelated CI failures

PR #11908, opened 2025-06-24
NXP backend: Improve quantization annotation process

robert-kalmar:
### Summary
Addresses multiple subsequent nodes with SharedSpecPattern problem. 

cc @digantdesai @JakeStevens

### DIFF
--- a/backends/nxp/quantizer/neutron_quantizer.py
+++ b/backends/nxp/quantizer/neutron_quantizer.py
@@ -10,7 +10,6 @@ import torch
 from executorch.backends.nxp.aten_passes.neutron_aten_pass_manager import (
     NeutronAtenPassManager,
 )
-
 from executorch.backends.nxp.quantizer.patterns import (
     AddmmPattern,
     AvgPoolPattern,
@@ -24,6 +23,7 @@ from executorch.backends.nxp.quantizer.patterns import (
     ReluInPlacePattern,
     ReluPattern,
     ReshapePattern,
+    SharedSpecPattern,
     SoftMaxPattern,
     ViewPattern,
 )
@@ -204,9 +204,36 @@ class NeutronQuantizer(ComposableQuantizer):
                 NeutronAtenQuantizer(ViewPattern(), static_qconfig),
             ]
         )
+        # Mapping ops defined in quantizer partition types to its quantizer
+        self.op_to_quantizer = {
+            pt: q for q in self.quantizers for pt in q.pattern.partition_types()
+        }
+        # Mapping ops to the quantizer application state
+        self.op_to_applied_quantizer = {
+            pt: False for q in self.quantizers for pt in q.pattern.partition_types()
+        }
 
     def transform_for_annotation(
         self, model: torch.fx.GraphModule
     ) -> torch.fx.GraphModule:
         pass_runner = NeutronAtenPassManager()
         return pass_runner(model).graph_module
+
+    def annotate(self, model: torch.fx.GraphModule) -> torch.fx.GraphModule:
+        nodes = list(model.graph.nodes)
+        for node in nodes:
+            if (
+                node.target not in self.op_to_quantizer
+                or self.op_to_applied_quantizer[node.target]
+            ):
+                continue
+            else:
+                quantizer = self.op_to_quantizer[node.target]
+                quantizer.annotate(model)
+                if not isinstance(quantizer.pattern, SharedSpecPattern):
+                    self.op_to_applied_quantizer[node.target] = True
+
+        return model
+
+    def validate(self, model: torch.fx.GraphModule) -> None:
+        return super().validate(model)
--- a/backends/nxp/tests/test_quantizer.py
+++ b/backends/nxp/tests/test_quantizer.py
@@ -5,6 +5,8 @@
 
 # Tests for NeutronQuantizer.
 
+from copy import deepcopy
+
 import executorch.backends.nxp.tests.models as models
 import torch
 from executorch.backends.nxp.quantizer.neutron_quantizer import NeutronQuantizer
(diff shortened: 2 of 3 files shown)


JakeStevens:
Can you provide a little more info on what this is solving?

Perhaps a test to illustrate/prove that is now correct and ensure no regression later

StrycekSimon:
Since the `SharedSpecPattern` quantization pattern relies on quantization annotation from the previous node in the graph, we need to process the nodes in order. Changes present in this PR do exactly this.
 
I will add a test that covers this case.

JakeStevens:
What is the status of this PR?

StrycekSimon:
Sorry for the wait time. I added tests that better demonstrates what is this feature trying to solve.

robert-kalmar:
@digantdesai , @JakeStevens, we added a test case demostrating and verifiing the change. The general concept is the quantization results is now independent on the order of individual annotators.

PS: please approve the workflows. The complete CI does not execute.

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -202,9 +202,34 @@ def __init__(self):
                 NeutronAtenQuantizer(AvgPoolPattern(), static_qconfig),
             ]
         )
+        self.op_to_quantizer = {
```
digantdesai, line 208: Add comments?

backends/nxp/tests/test_quantizer.py
```diff
@@ -271,3 +273,68 @@ def test_quantizer_conv2d_permute():
...
+
+    # Dry run
+    m(*example_input)
+
+    nodes = list(m.graph.nodes)
+
+    assert len(nodes) == 15
+    assert nodes[-5].name == "dequantize_per_tensor_default_3"
+    assert nodes[-4].name == "max_pool2d"
+    assert nodes[-3].name == "quantize_per_tensor_default_4"
```
digantdesai, line 305: Do we also want to check -5 and -3 has same qparams?

backends/nxp/tests/test_quantizer.py
```diff
@@ -271,3 +273,68 @@ def test_quantizer_conv2d_permute():
...
+
+    graph_module = torch.export.export_for_training(
+        model, example_input, strict=True
+    ).module()
+
+    m = prepare_pt2e(deepcopy(graph_module), quantizer)
+    m(*example_input)
+    m = convert_pt2e(m)
+
+    quantizer.quantizers = quantizer.quantizers[::-1]
```
digantdesai, line 332: curious why do you think the order matters here? IIUC this is based on the node.target dict lookup, no?

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -202,9 +202,34 @@ def __init__(self):
...
+            if (
+                node.target not in self.op_to_quantizer
+                or self.op_to_applied_quantizer[node.target]
+            ):
+                continue
+            else:
+                quantizer = self.op_to_quantizer[node.target]
+                quantizer.annotate(model)
+                if not isinstance(quantizer.pattern, SharedSpecPattern):
+                    self.op_to_applied_quantizer[node.target] = True
```
digantdesai, line 234: I am not following the logic here. Can you explain this to me? Feels like we will call annotate N times if we have N Relu nodes, no?

backends/nxp/tests/test_quantizer.py
StrycekSimon: The main point of this test is that the second node (in a row) with `SharedSpecPattern` quantizer won't even be quantized.

backends/nxp/tests/test_quantizer.py
StrycekSimon: Previously the `NeutronQuantizer` went through all the node quantizers in the `.quantizers` list. This created a problem that based on the order, some shared spec quantizers could be called before the quantization specs of the neighbor node were known - resulting in the node (the one with the shared specs) not being quantized.

backends/nxp/tests/test_quantizer.py
digantdesai: sure, but checking same q_param would also test that we did use `SharedSpecPattern`

backends/nxp/tests/test_quantizer.py
digantdesai: Not sure I follow but seems like implementation specific. Nonetheless, good to have the test :)

StrycekSimon:
@roman-janik-nxp, can you please look at the rest of the comments?

backends/nxp/quantizer/neutron_quantizer.py
roman-janik-nxp: Hi Digant, yes, your feeling is right. Before this PR, the quantizers were applied by the order defined by a list of quantizers.

This PR changes it to application of quantizers by the order of nodes in a graph. Nodes that have own quantization patterns like Conv2d or Linear do not depend on quantization of other nodes, the relevant quantizer is applied to all such nodes in a graph. On the contrary, there are nodes that use `SharedSpecPattern`, which quantization depends on other nodes. They can only be correctly quantized if the node which they depend on is quantized before. For each such node, the quantization is different due to the dependency, so the quantizer has to be applied separately. Examples are Relu, Abs, Reshape.

backends/nxp/quantizer/neutron_quantizer.py
roman-janik-nxp: This dict is a mapping from a torch operation defined in `torch.fx.node.target` to a relevant quantizer. Its created from a list of quantizers with its pattern. In pattern, there can be multiple torch operations, that is the reason for the second for loop.

Or do you mean we should add a comment to the code?

backends/nxp/tests/test_quantizer.py
StrycekSimon: Okay, I can add a check for this.

Nit: `SharedSpecPattern` ensures -6 and -5 will have same params, not -5 and -3, but I see the point you are making here.

JakeStevens:
may need to rebase to get past some errors like that java lint

backends/nxp/quantizer/neutron_quantizer.py
robert-kalmar: Yes @roman-janik-nxp , please add such a comment here for clarity.

JakeStevens (approved):
thanks for adding the tests, especially to check the qparams match like Digant suggested

PR #11909, opened 2025-06-24
feat: enable adding executorch as a module to zephyr

petriok:
### Summary
This pull request introduces the necessary structure to allow executorch to be added as a module to the Zephyr RTOS. Specifically, it adds a zephyr/ directory containing a module.yml file, which is required by Zephyr's module system.
No functional changes are introduced to the codebase—this PR is purely structural to support integration with Zephyr's build system.

## Details
Added: zephyr/module.yml
Purpose: Enables executorch to be recognized and used as a Zephyr module via west or CMake.

## Motivation
Zephyr requires a specific directory layout and metadata (module.yml) to include external projects as modules. This change ensures compatibility and simplifies integration for users who want to use executorch within Zephyr-based projects.

##Notes
This change is non-invasive and does not affect existing functionality.
Future enhancements may include additional configuration or integration logic as needed.

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -1,0 +1,4 @@
+name: executorch
+build:
+  cmake-ext: True
+  kconfig-ext: True


petriok:
Hi all,
I’ve been working on integrating Zephyr and Executorch, and I wanted to share some early progress. I’ve created a branch in my fork of the Zephyr repository:
https://github.com/petriok/zephyr/tree/poks/executorch-module-integration

This is still a work in progress, but the sample at
samples/modules/executorch/arm/hello_world
is already running correctly on hardware.

Happy to hear any feedback or suggestions on how this could align with ongoing efforts.

BujSet:
Thanks for putting this up. A few things I could use some clarification on. As I understand it, the YAML file here is needed in order to allow zephyr to build executorch as module, is that correct? If so, can you elaborate on why this change should be in the executorch repo rather than in the zephyr repo?

For example, it seems like other modules define their yaml files within the zephyr repo, e.g. tflite [sample hello world](https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/modules/tflite-micro/hello_world/sample.yaml) and [modules](https://github.com/zephyrproject-rtos/zephyr/tree/main/modules/tflite-micro). Do you think we can do something similar?

petriok:
> Thanks for putting this up. A few things I could use some clarification on. As I understand it, the YAML file here is needed in order to allow zephyr to build executorch as module, is that correct? If so, can you elaborate on why this change should be in the executorch repo rather than in the zephyr repo?
> 
> For example, it seems like other modules define their yaml files within the zephyr repo, e.g. tflite [sample hello world](https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/modules/tflite-micro/hello_world/sample.yaml) and [modules](https://github.com/zephyrproject-rtos/zephyr/tree/main/modules/tflite-micro). Do you think we can do something similar?

Yes you are absolutely right, this YAML file doesn't need to be in the executorch repo. The fork of executorch should reside in zephyrproject-rtos/executorch, and only in the fork there should be the zephyr/module.yml. This PR can be closed.

Additionally, it would probably make sense to follow the example of tflite-micro and add executorch as an optional module, in the zephyr/submanifests/optional.yaml.

PR #11917, opened 2025-06-24
Move transformer base definition to extension/llm

jackzhxng:


extension/llm/modeling/text_decoder/rope.py
```diff
@@ -0,0 +1,308 @@
+# @lint-ignore-every LICENSELINT
```
digantdesai: can we use `git mv` to preserve the history better? I saw the import error message update, perhaps that's why you didn't do it?

digantdesai (commented):
Is it mainly mechanical move? I see bunch of CI failures, also added trunk tag to make sure we don't regress silently on other jobs.

extension/llm/modeling/text_decoder/rope.py
jackzhxng: I'll try, I just forgot to do `git mv` this time

mergennachin (changes requested):
let's not make the extension/llm/modeling/text_decoder/decoder_model.py move. There are too many examples and meta specific stuff there... let's keep it in examples

mergennachin:
Why did you not move llama_transformer.py to extension inside?

PR #11922, opened 2025-06-24
Fix bug in le broadcast for single element broadcast

kimishpatel:
Summary:
Really comparison ops cannot be handled in the same way as other binary ops for broadcasting because output tensor dtype is different than input tensor dtype.

As a result we just have to fall back to portable.

Even the current vectorized impl for le, ge etc. assumes that the output type of compare is same as input type. That might actually be a bug.

A potential way to handle this maybe via vectorized compare natively supporting binary output vector

Test Plan:
Tests added which fail before and passes after

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/kernels/optimized/cpu/op_le.cpp
+++ b/kernels/optimized/cpu/op_le.cpp
@@ -30,56 +30,7 @@ Tensor& opt_le_tensor_out(
   (void)ctx;
 
   ScalarType a_type = a.scalar_type();
-  ScalarType b_type = b.scalar_type();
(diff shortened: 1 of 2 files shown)


kimishpatel:
This also fixes the garbage output issue in optimum-executorch models

guangy10 (approved):
LGTM. Accepted to unblock the work in optimum-executorch.

kimishpatel:
failures are not related

PR #11923, opened 2025-06-24
Arm Zephyr cmake Preset

BujSet:
### Summary
This change adds a cmake preset for Arm Zephyr RTOS toolchain. This requires the toolchain cmake compilation target to be added to `PATH` (see #12077 for more details). This change includes update to CI flows, modifying the CI `build_size_test` to run the test using either arm bare metal or Zephyr toolchains. Additionally, the CI for building presets is updated to target the new cmake Zephyr preset. 

### Test plan

Tested with `test/build_size_test.sh`  to determine the correct threshold for running the size test for the arm zephyr toolchain. When the CI runs the size test for the arm zephyr toolchain, the new cmake preset is used directly. The following commands are run by the CI, and can be manually invoked to confirm functionality:

```
CXXFLAGS="-fno-exceptions -fno-rtti -Wall -Werror -Wno-int-in-bool-context -DET_HAVE_PREAD=0" cmake --preset zephyr -DCMAKE_BUILD_TYPE=Release -DEXECUTORCH_OPTIMIZE_SIZE=ON -DCMAKE_INSTALL_PREFIX=cmake-out -Bcmake-out .
cmake --build cmake-out -j9 --target install --config Release
CXXFLAGS="-fno-exceptions -fno-rtti -Wall -Werror -Wno-int-in-bool-context -DET_HAVE_PREAD=0"  cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=$(realpath examples/zephyr/x86_64-linux-arm-zephyr-eabi-gcc.cmake) -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=cmake-out -Bcmake-out/test test
cmake --build cmake-out/test -j9 --config Release
arm-zephyr-eabi-strip cmake-out/test/size_test
ls -la cmake-out/test/size_test
```

Output should corroborate CI test threshold, reporting a resultant `size_test` binary size of ~125Kb (threshold set to 130KB toin CI test tolerate variability).


examples/arm/ethos-u-setup/arm-zephyr-eabi-gcc.cmake
```diff
@@ -0,0 +1,103 @@
...
+    "cortex-m55"
+    CACHE STRING "Target CPU"
+)
+string(TOLOWER ${TARGET_CPU} CMAKE_SYSTEM_PROCESSOR)
+
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_C_COMPILER "arm-zephyr-eabi-gcc")
+set(CMAKE_CXX_COMPILER "arm-zephyr-eabi-g++")
+set(CMAKE_ASM_COMPILER "arm-zephyr-eabi-gcc")
+set(CMAKE_LINKER "arm-zephyr-eabi-ld")
```
larryliu0820: I think we should error out if these are not found in PATH.

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,61 @@
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set(CMAKE_CXX_STANDARD 17)
```
jathu: We shouldn't set this here, we already do it here 

https://github.com/pytorch/executorch/blob/91c9ffad6401c2b1d546724f1df88dc77106bf0e/CMakeLists.txt#L58-L61

Similar comment for the other flags

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,61 @@
...
+set(CMAKE_SYSTEM_PROCESSOR                   cortex-m55)
```
jathu: I don't think this will work — I think this needs to be configure before even kicking off the build. Stick this in `CMakePresets.json`

CMakePresets.json
```diff
@@ -104,6 +104,17 @@
...
+    {
+        "name": "zephyr",
+        "displayName": "Build everything buildable on Zephyr RTOS",
+        "inherits": [
+            "common"
+        ],
+        "cacheVariables": {
+            "EXECUTORCH_BUILD_PRESET_FILE": "${sourceDir}/tools/cmake/preset/zephyr.cmake",
+            "CMAKE_OSX_DEPLOYMENT_TARGET": "12.0"
+        }
```
jathu: Do we want to limit that platforms can build this? See macos/ios above as an example

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,61 @@
...
+set(CMAKE_BUILD_TYPE                         Release)
+set(PYTHON_EXECUTABLE                        /home/zephyruser/executorch/.venv/bin/python3)
+set(BUCK2                                    /home/zephyruser/executorch/buck2-bin/buck2-2025-05-06-201beb86106fecdc84e30260b0f1abb5bf576988)
+set(CMAKE_CXX_COMPILER_ID                    GNU)
+set(CMAKE_TOOLCHAIN_FILE                     /home/zephyruser/executorch/examples/arm/ethos-u-setup/arm-zephyr-eabi-gcc.cmake)
+set(EXECUTORCH_ENABLE_LOGGING                ON)
```
jathu: For the executorch flags use `set_overridable_option`. Otherwise, I don't think the values will be overridable

CMakePresets.json
BujSet: I don't think we need to limit the platforms; I removed this constraint.

.github/workflows/trunk.yml
```diff
@@ -271,6 +271,62 @@ jobs:
           exit 1
         fi
 
+  test-arm-cortex-m-zephyr-size-test:
```
digantdesai: can we deduplicate and merge this job with `test-arm-cortex-m-size-test`, see `matrix` above

examples/arm/ethos-u-setup/arm-zephyr-eabi-gcc.cmake
```diff
@@ -0,0 +1,111 @@
+#
```
digantdesai: `examples/arm/ethos-u-setup/arm-none-eabi-gcc.cmake` can we overload this file, and introduce perhaps `PREFIX` or `TRIPLET` i.e. `arm-none-eabi-` vs. `arm-zephyr-eabi-`?

See `$ANDROID_NDK/build/cmake/android.toolchain.cmake` as an example.

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,34 @@
+
```
digantdesai, line 1: Nice! Does this line up with existing Arm baremetal build?

digantdesai:
@zingo - FYI

.ci/scripts/setup-arm-zephyr-tools.sh
perheld, line 1: Maybe the file should be named something that points out that its specific for a x86_64 runner?

backends/arm/scripts/build_executor_runner.sh
```diff
@@ -46,6 +46,7 @@ help() {
     echo "  --output=<FOLDER>               Output folder Default: <MODEL>/<MODEL>_<TARGET INFO>.pte"
     echo "  --et_build_root=<FOLDER>        Build output root folder to use, defaults to ${et_build_root}"
     echo "  --ethosu_tools_dir=<FOLDER>     Path to your Ethos-U tools dir if you not using default: ${ethosu_tools_dir}"
+    echo "  --toolchain=<TOOLCHAIN>   Toolchain can be specified (e.g. bare metal as arm-none-eabi-gcc or zpehyr as arm-zephyr-eabi-gcc"
```
perheld: Spelling zpehyr -> zephyr

perheld: Also a little bit of indentation. A small nit.

backends/arm/scripts/build_portable_kernels.sh
```diff
@@ -29,6 +28,7 @@ help() {
     echo "  --et_build_root=<FOLDER>   Build output root folder to use, defaults to ${et_build_root}"
     echo "  --build_type=<TYPE>        Build with Release, Debug or RelWithDebInfo, default is ${build_type}"
     echo "  --portable_kernels=<OPS>   Comma separated list of portable (non delagated) kernels to include Default: ${portable_kernels}"
+    echo "  --toolchain=<TOOLCHAIN>   Toolchain can be specified (e.g. bare metal as arm-none-eabi-gcc or zpehyr as arm-zephyr-eabi-gcc"
```
perheld: Probably copy/pasted from above. Same spelling.

examples/arm/setup.sh
```diff
@@ -200,15 +226,27 @@ function setup_fvp() {
...
-    if [[ ! -e "${toolchain_dir}.tar.xz" ]]; then
+    local selected_toolchain_url="${toolchain_url}"
+    local selected_toolchain_dir="${toolchain_dir}"
+    local selected_toolchain_md5="${toolchain_md5_checksum}"
+    local skip_md5_verify=0
+    if [[ -n "${user_toolchain_url}" ]]; then
+        selected_toolchain_url="${user_toolchain_url}"
+        selected_toolchain_dir="${user_toolchain_dir}"
+        selected_toolchain_md5=""
+        skip_md5_verify=1
```
perheld: We probably need to think about how to handle the md5sums of multiple supported compilers and versions of them. Skipping the checksum seems dangerous. But I dont think that is part of this PR.

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,34 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set_overridable_option(CMAKE_TOOLCHAIN_FILE                     "${PROJECT_SOURCE_DIR}/examples/arm/ethos-u-setup/arm-zephyr-eabi-gcc.cmake")
```
jathu: Can you verify this works? I'm pretty sure setting `CMAKE_TOOLCHAIN_FILE` within cmake itself actually doesn't work. You need to set this in the CMakePreset.json (see the iOS/Apple examples)

.github/workflows/trunk.yml
jathu: Can we simplify this? One of the motivation for the presets is so that most of the complex build steps are distilled down to:

```
$ cmake --preset zephyr
```

We want to mimic what a user would do in the CI. For example, see the simplicity of building apple frameworks:

https://github.com/pytorch/executorch/blob/ecb85cecba92f5ea56aeb9f6268e8f64f268f3ab/.github/workflows/build-presets.yml#L18-L35

(btw since we are creating a new preset, we should add a step in build-presets.yml

examples/arm/setup.sh
BujSet: I left a todo in the code for this. I think the right think to do is to have the user specify the checksum, but I'm not sure where they would get it from for arbitrary toolchains.

tools/cmake/preset/zephyr.cmake
BujSet: I believe the only difference is that the arm bare metal build requires `EXECUTORCH_BUILD_ARM_BAREMETAL` to be set to `ON`.

examples/arm/setup.sh
```diff
@@ -197,12 +200,40 @@ function setup_fvp() {
...
+   elif [[ "${ARCH}" == "aarch64" ]] || [[ "${ARCH}" == "arm64" ]]; then
+        if [[ "${OS}" == "Darwin" ]]; then
+            toolchain_url="https://armkeil.blob.core.windows.net/developer/Files/downloads/gnu/13.3.rel1/binrel/arm-gnu-toolchain-13.3.rel1-darwin-arm64-arm-none-eabi.tar.xz"
+            toolchain_dir="arm-gnu-toolchain-13.3.rel1-darwin-arm64-arm-none-eabi"
+            toolchain_md5_checksum="f1c18320bb3121fa89dca11399273f4e"
+        elif [[ "${OS}" == "Linux" ]]; then
+            toolchain_url="https://armkeil.blob.core.windows.net/developer/Files/downloads/gnu/13.3.rel1/binrel/arm-gnu-toolchain-13.3.rel1-aarch64-arm-none-eabi.tar.xz"
+            toolchain_dir="arm-gnu-toolchain-13.3.rel1-aarch64-arm-none-eabi"
+            toolchain_md5_checksum="303102d97b877ebbeb36b3158994b218"
+        fi
```
perheld: There should probably be a case here for zephyr. Atleast that prints an error or unsupported or something because now it would choose gcc on a aarch64 and you would be very confused.

examples/arm/setup.sh
```diff
@@ -197,12 +200,40 @@ function setup_fvp() {
     done
 }
 
+function select_toolchain() {
+    if [[ "${ARCH}" == "x86_64" ]]; then
+	if [[ "${target_toolchain}" == "zephyr" ]]; then
```
perheld: This should probably be inside a OS == Linux and then an else case printing something clever about unsupported OS.

.ci/scripts/setup-arm-zephyr-x86-64-tools.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# NB: This function could be used to install Arm dependencies
+# Setup arm example environment (including TOSA tools)
+git config --global user.email "github_executorch@arm.com"
+git config --global user.name "Github Executorch"
+bash examples/arm/setup.sh --i-agree-to-the-contained-eula --user-toolchain-url "https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-x86_64_arm-zephyr-eabi.tar.xz" --user-toolchain-dir "arm-zephyr-eabi"
```
perheld: --user-toolchain-url/dir and  doesnt seem implemented? I guess now the url should be removed and the --user-toolchain-dir should be --toolchain ?

perheld:
I think this is great work!

I'm just being a little bit difficult in the build shellscripts around OS and compilers since I have been trying to add support for clang with ATfE (https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded) and I hope to see a environment where multiple flavours of gcc and clang can coexist in the future.

.ci/scripts/setup-arm-zephyr-x86-64-tools.sh
BujSet: Yep, you're right. Still working on the changes for this PR. Thanks for catching this.

examples/arm/setup.sh
BujSet: Update: Added in support for checking the md5 checksums.

.github/workflows/trunk.yml
BujSet: I've updated `.github/workflows/build-presets.yml` with the changes, how can I test to see if they worked correctly (not seeing it run on the CI tests)?

examples/arm/ethos-u-setup/arm-zephyr-eabi-gcc.cmake
BujSet: Hm, I'm been trying to make changes to support this, but in doing so, the changes are becoming quite a bit more invasive. Every arm script (e.g. `examples/arm/setup.sh`, `examples/arm/run.sh`, `backends/arm/scripts/build_executorch.sh` ...) would need to take additional flags to use the `arm-none-eabi-gcc.cmake` but overloaded with a flag to specify that we shouldn't use the bare metal toolchain, but use the zephyr one instead. It is a bit of file duplication, but changes to the existing scripts are much more minimal if we retain `arm-zephyr-eabi-gcc.cmake` file separately. What do you think? I'm of the preference of keeping the `arm-zephyr-eabi-gcc.cmake` file for now, and maybe a later PR could consolidate the two?

.github/workflows/trunk.yml
BujSet: Added this to `build-presets.yml`. The job reports that this failed, but won't render the log to show me which part failed. Is there anything obvious that I could address to resolve the failure?

.ci/scripts/setup-arm-baremetal-tools.sh
```diff
@@ -8,4 +8,8 @@
 # Setup arm example environment (including TOSA tools)
 git config --global user.email "github_executorch@arm.com"
 git config --global user.name "Github Executorch"
-bash examples/arm/setup.sh --i-agree-to-the-contained-eula
+if [[ "$#" -eq 0 ]]; then
+    bash examples/arm/setup.sh --i-agree-to-the-contained-eula
+else
+    bash examples/arm/setup.sh --i-agree-to-the-contained-eula "$@"
```
digantdesai: Use bash default? 
`"${@:-}"`

zingo:
Really great work, thanks!

Just for information and not for this PR: 
Later when there is a runner that can run on FVP we can add end2end tests quite easy.

Just add a zephyr test function in backends/arm/test/test_arm_baremetal.sh  (yeah we should rename that script to test_arm_backend.sh some day)  it might work with re-reunning setup.sh to install your toolchain and then add some tests, look at  test_run_ethosu_fvp() for inspiration.

And then you just need to add then new zephyr test function name to .github/workflows/trunk.yml  in the list under test-arm-backend to make it run in it's own machine/job for tracking :) . 

We could also work the toolchain into the yml matrix/table under test-arm-backend  in .github/workflows/trunk.yml  not sure how that would turn out in the end but might be nice:er or not :)

zingo (approved):

BujSet:
> Really great work, thanks!
> 
> Just for information and not for this PR: Later when there is a runner that can run on FVP we can add end2end tests quite easy.
> 
> Just add a zephyr test function in backends/arm/test/test_arm_baremetal.sh (yeah we should rename that script to test_arm_backend.sh some day) it might work with re-reunning setup.sh to install your toolchain and then add some tests, look at test_run_ethosu_fvp() for inspiration.
> 
> And then you just need to add then new zephyr test function name to .github/workflows/trunk.yml in the list under test-arm-backend to make it run in it's own machine/job for tracking :) .
> 
> We could also work the toolchain into the yml matrix/table under test-arm-backend in .github/workflows/trunk.yml not sure how that would turn out in the end but might be nice:er or not :)

Thanks! @digantdesai have talked about refactoring a lot of the scripts here so that they aren't "bare metal" specific, and allow for generic toolchain selection (e.g. zephyr, or even anything else). From what I've seen, I think adding a matrix (like in this PR) would be a good idea for the future. I'm thinking those will be separate PRs. Thanks for all the feedback on this!

.github/workflows/trunk.yml
```diff
@@ -234,35 +238,62 @@ jobs:
...
+        ${toolchain_prefix}c++ --version
+
+        # Setup cmake target to desired toolchain
         toolchain_cmake=$(realpath ${toolchain_cmake})
 
-        # Build and test size test
-        bash test/build_size_test.sh "-DCMAKE_TOOLCHAIN_FILE=${toolchain_cmake} -DEXECUTORCH_BUILD_ARM_BAREMETAL=ON"
+        # Build and run size test
+        if [[ ${{ matrix.os}} == "bare_metal" ]]; then 
+          bash test/build_size_test.sh "-DCMAKE_TOOLCHAIN_FILE=${toolchain_cmake} -DEXECUTORCH_BUILD_ARM_BAREMETAL=ON"
```
larryliu0820, line 270: Not in this PR but some day we should have a arm baremetal preset instead of running this script.

.github/workflows/trunk.yml
```diff
@@ -234,35 +238,62 @@ jobs:
         CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
         conda activate "${CONDA_ENV}"
 
+        cxx_flags="-fno-exceptions -fno-rtti -Wall -Werror -Wno-int-in-bool-context -DET_HAVE_PREAD=0"
+        setup_script_args=""
+        if [[ ${{ matrix.os}} == "bare_metal" ]]; then 
+          toolchain_prefix=arm-none-eabi-
```
larryliu0820, line 244: We don't have to specify `--target-toolchain baremetal`?

larryliu0820 (approved):

tools/cmake/preset/zephyr.cmake
```diff
@@ -0,0 +1,33 @@
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set_overridable_option(EXECUTORCH_BUILD_COREML                  OFF)
```
jathu, line 8: nit: most, if not all, of these are off by default anyways — so I guess we don't need to manually do it here again

.github/workflows/build-presets.yml
```diff
@@ -34,6 +34,45 @@ jobs:
...
+
+        # Verify download
+        echo "93128be0235cf5cf5f1ee561aa6eac5f  ${toolchain_dir}.tar.xz" > arm-zephyr-eabi.md5
+        md5sum -c --strict arm-zephyr-eabi.md5
+        
+        # Extract and install to PATH
+        tar xf "${toolchain_dir}.tar.xz"
+        rm -f "${toolchain_dir}.tar.xz"
+        toolchain_bin_path="$(cd ${toolchain_dir}/bin && pwd)"
+        export PATH=$PATH:${toolchain_bin_path}
```
jathu, line 71: One of the goals with presets, is that it should work "out of the box" without having to manually do extra steps. If you look at the other examples in the workflow, they are all 3-4 commands. 

What can we do to achieve that here? Right now, if I fetch ET and run `cmake --preset zephyr`, it will fail

.github/workflows/build-presets.yml
```diff
@@ -34,6 +34,45 @@ jobs:
         ${CONDA_RUN} cmake --preset ${{ matrix.preset }}
         ${CONDA_RUN} cmake --build cmake-out -j$(( $(sysctl -n hw.ncpu) - 1 ))
 
+  zephyr:
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: [zephyr]
```
jathu, line 42: Am I misunderstanding this, or should this technically have two different presets for ARM builds?

```suggestion
  arm:
    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
    strategy:
      fail-fast: false
      matrix:
        preset: [zephyr, baremetal]
```

.github/workflows/build-presets.yml
larryliu0820: Well in CI we will have to download the toolchain separately right?  This is similar to `setup_macos.sh` or `setup_linux.sh`. I don't expect cmake to download this toolchain for us?

.github/workflows/trunk.yml
BujSet: In the arm scripts, the default toolchain is set to `arm-none-eabi`. The intended use is that if there is some other toolchain to use (e.g. zephyr), it can be specified. A future PR could change this so there is no default, and `baremetal` would be specified to clarify which toolchain was used. I think for now, leaving the baremetal default makes sense based on what is currently well-supported.

.github/workflows/build-presets.yml
larryliu0820: I don't think preset baremetal is a thing

.github/workflows/trunk.yml
BujSet: That makes sense, should I make an issue for it?

.github/workflows/build-presets.yml
BujSet: I think to support working out of the box, downloading the Arm toolchains would need to included as part of the preset configuration process (maybe the `curl`, `md5sum`, and `export` commands can be moved into the `tools/cmake/preset/zephyr.cmake` file directly?). I'm a little wary of this, as it seems like a platform specific configuration, [see here](https://github.com/pytorch/executorch/blob/967cfae2e2c1884c66d1f5981c8a92ad303b083d/examples/arm/setup.sh#L203-L243). Maybe this is part of the reason why there isn't an arm baremetal preset as of yet?

jathu: We [don't call setup_macos in this workflow](https://github.com/pytorch/executorch/blob/22b9e59890c7fb75784480efb100e4a68897967b/.github/workflows/build-presets.yml#L17-L35) because all the toolchains needed to build for macOS/iOS are already present in macOS.

You need to have the toolchain ready before calling cmake so I don't think we can make it a part of the cmake build process. Perhaps we can abstract this out to a script that can be used be users? 

Not a blocker for this PR, we can follow up

.github/workflows/build-presets.yml
digantdesai: we can revisit this in the future. I like what @jathu proposed but we need to think a bit more about baremetal builds I guess.

examples/zephyr/arm-x86-64-eabi-gcc.cmake
```diff
@@ -0,0 +1,106 @@
+#
```
digantdesai: I don't understand the filename, can you elaborate? The toolchain tuple naming convention - https://wiki.gentoo.org/wiki/Embedded_Handbook/Tuples#:~:text=From%20Gentoo%20Wiki,userland%20and/or%20userland%20ABI.

.ci/scripts/setup-arm-baremetal-tools.sh
```diff
@@ -8,4 +8,4 @@
 # Setup arm example environment (including TOSA tools)
 git config --global user.email "github_executorch@arm.com"
 git config --global user.name "Github Executorch"
-bash examples/arm/setup.sh --i-agree-to-the-contained-eula
+bash examples/arm/setup.sh --i-agree-to-the-contained-eula ${@:-}
```
digantdesai, line 11: :)

examples/zephyr/arm-x86-64-eabi-gcc.cmake
BujSet: Earlier, @perheld had a comment about adding some notation to the changed files to indicate that this zephyr toolchain is specific to a host x86-64 Linux system, [see here](https://github.com/pytorch/executorch/pull/11923/files#diff-9f97e331d7fea7cb9a26d40e804aaab34ff0fdc46bb4508e316a026637dbe07fL203-L244). I wasn't sure what a good name would be, maybe something of the format `hostArch-hostOS-targArch-targOS-eabi-gcc.cmake`, which in this case would be `x86_64-linux-arm-zephyr-eabi-gcc.cmake`, would be a better option?

.github/workflows/build-presets.yml
larryliu0820: Yeah it should be setup_zephyr.sh that handles the toolchain installation.

.github/workflows/trunk.yml
larryliu0820: Yeah let's create an issue

BujSet: Created the issue here: #12186, can discuss more about what this feature should look like on that thread.

digantdesai (approved):

PR #11934, opened 2025-06-24
Split neutron backend test based on executor dependency

JakeStevens:
Summary:
The current test_neutron_backend has a hard requirement on tflite due to the Executor comparing the output.

Not all systems have tflite, but we would like to have a sanity check on the backend still.

To achieve this, this PR splits the existing test into two files

test_neutron_backend contains the tests that do not use any Executor logic (ie, no end to end comparison of outputs)
test_neutron_backend_executor contains the rest of the tests.

Differential Revision: D77248751

### DIFF
--- a/backends/nxp/tests/test_neutron_backend.py
+++ b/backends/nxp/tests/test_neutron_backend.py
@@ -3,27 +3,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import numpy as np
-import torch
-
-from executorch.backends.nxp.backend.edge_program_converter import (
-    EdgeProgramToIRConverter,
-)
-from executorch.backends.nxp.backend.ir.lib.tflite.BuiltinOptions import BuiltinOptions
-from executorch.backends.nxp.backend.ir.lib.tflite.Model import Model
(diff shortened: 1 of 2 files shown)


skywall:
In general looks good to me. How do you plan to execute tests only in particular files? I believe pytest will fail to run anything if there's at least single file with problematic import. Only working solutions how to execute this are imho 1) single file's tests execution 2) predefined list of "files with supported tests".

Please rebase to main to include nxp-unittest workflow.

skywall (approved):

JakeStevens:
> In general looks good to me. How do you plan to execute tests only in particular files? I believe pytest will fail to run anything if there's at least single file with problematic import. Only working solutions how to execute this are imho 1) single file's tests execution 2) predefined list of "files with supported tests".
> 
> Please rebase to main to include nxp-unittest workflow.

This is an issue specific to our internal workflow, which does not use the OSS flow but uses BUCK entirely. So, you just only include the files you want in your BUCK target, very easy (basically a special form of your #2 suggestion).

Will rebase

robert-kalmar:
LGTM, 
executed the trunk CI to invoke the nxp-unittests too to verify the change.

robert-kalmar (approved):

digantdesai (approved):

PR #11912, opened 2025-06-24
Manual cherry-pick: Re-reapply "Add vectorized_math.h (#11204)", "Add optimized_portable_kernels test (#11205)", and "Add vectorization in elementwise_util (#9432)"

swolchok:
ghstack PR number: https://github.com/pytorch/executorch/pull/11802

Please see that original PR for details; this is a manual cherry-pick because mergebot failed.

PR #11939, opened 2025-06-24
Delete optimized sigmoid

swolchok:
We have vectorized (and parallelized) portable ops now as of #11912, so we don't need a handwritten optimized op that only does vectorization. (I spot-checked assembly to confirm the portable op is getting vectorized when we build optimized_portable_kernels.)


swolchok:
* __->__ #11939

manuelcandales (approved):
Wow, this is great!

PR #9585, opened 2025-03-25
relax tolerances for all unary float ops

swolchok:
We were requiring ourselves to compute at double-precision, but ATen actually converts non-floating-point types to `float` by default, not `double`. Use the ATen tolerances everywhere.

PR #11940, opened 2025-06-24
Reapply: "relax tolerances for all unary float ops (#9585)", "Add SupportedTensorDtypes::BOOL (#9584)", new op_mul test (#11206)

swolchok:
* (to be filled)

These were reverted because they were part of a stack with interenal test failures.

Original #9585 summary:
We were requiring ourselves to compute at double-precision, but ATen
actually converts non-floating-point types to `float` by default, not
`double`. Use the ATen tolerances everywhere.

Original #9584 summary: none

Original #11206 summary:
This tests a possibly-surprising result: int8(100) * int8(100) with
output type of long is 16 in ATen, even though the output type can hold 10000.

Differential Revision: [D76754823](https://our.internmc.facebook.com/intern/diff/D76754823/)


swolchok:
not sure why we got 3 copies of this

PR #11943, opened 2025-06-24
Reapply "Implement unary_ufunc functions using elementwise_util (#9386)"

swolchok:
* __->__ #11943
* #12121

This PR was reverted due to internal test failures, which should be fixed now (and is being sent as an exported diff to make sure).

Original summary:
One less set of independent implementations to worry about going forward
(e.g., we don't have to vectorize these separately from elementwise_util
and they get all benefits of elementwise_util).

Differential Revision: [D76754824](https://our.internmc.facebook.com/intern/diff/D76754824/)


swolchok:
noting that this one had green tests except for lintrunner prior to last update

manuelcandales (approved):

swolchok:
Per discussion wiht @manuelcandales, this does not seem like we need to focus on it before GA, so I'll just convert it to draft and we'll unshelve if/when needed or post-GA.

PR #11944, opened 2025-06-24
Fix the print

yangw-dev:
clean up the print from common.ts 

 add --debug flag to execuTorch script.


.ci/scripts/benchmark_tooling/analyze_benchmark_stability.py
```diff
@@ -81,6 +84,9 @@ def analyze_latency_stability(  # noqa: C901
     primary_datasets = {}
     documents = read_excel_with_json_header(primary_file)
 
+    if debug:
+        print(f"[Debug Mode] Printing documents: {documents}")
```
huydhn, line 91: To be honest, there are better way to do logging than print.  But then, the script has already used print, so why not

huydhn (approved):

guangy10 (approved):

guangy10 (commented):
Can I enable the log if I set verbose in the analysis script?

PR #11959, opened 2025-06-25
[Backend Tester] Add CoreML tester implementation

GregoryComer:
Add a tester implementation for Core ML. This uses the common backend tester framework to allow Core ML to plug into the test harness using pybindings.

PR #11953, opened 2025-06-25
[Backend Tester] Add FACTO operator test skeleton

GregoryComer:
Add initial skeleton for running Facto operator tests. This is only set up for XNNPACK in this commit (other delegates are later in the stack). It also relies on a manual install of Facto. I'm currently intentionally not running this in CI, due to a combination of finding a number of crashes and due to a high volume of tests. This will be addressed further up the stack. Instructions for running the tests locally are included at the top of test_facto.py.

PR #11960, opened 2025-06-25
[Backend Tester] Add backend test suite skeleton

GregoryComer:
This PR lays the foundation for the new backend test suite. It includes a skeleton of the runner (iterated on in the stack) and some basic tests for some elementwise operators.


GregoryComer:
* __->__ #11960
* #11959
* #11953

cccclai:
What will happen for the following two scenarios?

- the backend doesn't support this particular operator, but support the decomposed version instead?
- the backend supports a particular aten op, that will be decomposed later?

GregoryComer:
> What will happen for the following two scenarios?
> 
> * the backend doesn't support this particular operator, but support the decomposed version instead?
> * the backend supports a particular aten op, that will be decomposed later?

In both cases, it should be fine. These tests focus on the operators end to end essentially, in the same way they appear in the model. If they are decomposed internally, it will still validate the correctness. If they're not partitioned, it's also fine, as the tests will pass.

cccclai:
> In



> > What will happen for the following two scenarios?
> > 
> > * the backend doesn't support this particular operator, but support the decomposed version instead?
> > * the backend supports a particular aten op, that will be decomposed later?
> 
> In both cases, it should be fine. These tests focus on the operators end to end essentially, in the same way they appear in the model. If they are decomposed internally, it will still validate the correctness. If they're not partitioned, it's also fine, as the tests will pass.

How about the case that they should partition and delegated, but it was missing some cases, hence not partition and fall back to cpu? Also, how do we measure the closeness between the baseline and the backend op?

GregoryComer:
> > In
> 
> > > What will happen for the following two scenarios?
> > > 
> > > * the backend doesn't support this particular operator, but support the decomposed version instead?
> > > * the backend supports a particular aten op, that will be decomposed later?
> > 
> > 
> > In both cases, it should be fine. These tests focus on the operators end to end essentially, in the same way they appear in the model. If they are decomposed internally, it will still validate the correctness. If they're not partitioned, it's also fine, as the tests will pass.
> 
> How about the case that they should partition and delegated, but it was missing some cases, hence not partition and fall back to cpu? Also, how do we measure the closeness between the baseline and the backend op?

I might not be fully understanding the question, but this should also be fine. The purpose of these tests are twofold: (1) does the partitioner/lowering error out instead of not partitioning things it can't handle, and (2) if something is partitioned, are the outputs reasonably close to the eager value?

On numerical accuracy, I've set very loose tolerances, as my intent is mainly to catch things that are blatantly incorrect, rather than get into specific numerical deviation. It looks at the full graph output, so if something is decomposed and/or partially lowered, it will consider the net output regardless of how it's handled internally.

cccclai:
I think the following will help
1) make it configurable, meaning that backend can note what op is supported and what op is not, and being able to inject their own ops
2) check the graph (or the node) after partition, and make sure the op is lowered. This will help us knowing if the backend is doing the right thing.

backends/test/compliance_suite/README.md
```diff
@@ -0,0 +1,15 @@
+# Operator Compliance Test Suite
+
+This directory contains operator tests that all backends are expected to pass. While not every backend will implement every operator or permutation, the expectation is that backend partitioners will only partition nodes that the backend can support. The partitioner should never error out due to not supporting an input node.
```
digantdesai: say why? I.e. what makes these tests a must pass. I know it but document it.

I imagine we will build up a "qualifying criteria" for adding new things or removing things from this compliance suite. And what does it mean for a backend to be compliant, when a user reads it. Also indicate this is a purely functional and no bar on performance. Also we need to distinguish between compliant and partitioned vs. not-partitioned, when we present this, but that's for later.

backends/test/compliance_suite/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+from executorch.backends.test.harness import Tester
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+# Read enabled backends from the environment variable. Enable all if
+# not specified (signalled by None).
+def get_enabled_backends():
+    et_test_backends = os.environ.get("ET_TEST_ENABLED_BACKENDS")
```
digantdesai: why not all? Let's make a whitelist of "not_enabled" instead in the code. For testing let a user specify a single backend override, else default to testing all of them (using dirname or something dumb)

backends/test/compliance_suite/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+def is_backend_enabled(backend):
+    if _ENABLED_BACKENDS is None:
+        return True
+    else:
+        return backend in _ENABLED_BACKENDS
+
+
+ALL_TEST_FLOWS = []
+
+if is_backend_enabled("xnnpack"):
```
digantdesai: Reverse of this would be something like `is_backend_ready_for_complience_testing()` and someone has to whitelist their backend for this method to return no, and defaults to yes. And ofc you can override for testing etc.

backends/test/compliance_suite/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+    torch.int32,
+    torch.uint32,
+    torch.int64,
+    torch.uint64,
+    torch.float16,
+    torch.float32,
+    torch.float64,
+]
+
+FLOAT_DTYPES = [
```
digantdesai: missing bloat16?

I guess no complex types or fp8 cases yet.

backends/test/compliance_suite/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+    torch.float16,
+    torch.float32,
+    torch.float64,
+]
+
+
+class TestType(Enum):
+    STANDARD = 1
```
digantdesai: What does this mean?

backends/test/compliance_suite/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def _make_wrapped_test(test_func, tester_factory):
+    def wrapped_test(self):
+        test_func(self, tester_factory)
+
+    return tester_factory
+
+
+def _make_wrapped_dtype_test(test_func, dtype, tester_factory):
```
digantdesai: Nit: curious why not arg overload `_make_wrapped_test`?

backends/test/compliance_suite/operators/test_add.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def forward(self, x, y):
+        return torch.add(x, y, alpha=self.alpha)
+
+
+@operator_test
+class Add(OperatorTest):
+    @dtype_test
+    def test_add_dtype(self, dtype, tester_factory: Callable) -> None:
+        self._test_op(
```
digantdesai: I was wondering instead of calling this method here, why don't we just get an instance of "TestInputs" class which gives you logical_name, module definition, test inputs, etc. This is to disconnect the inputs from consuming those. This will allow us to repurpose these if we want to do different things with them. What do you think?

digantdesai (commented):
Too many comments, I will try to keep them structured.

- I would like to better understand and document what gets in/out of this suite but I like this idea. 
- I hope we can figure out the reporting for this.
- Another consideration is versioning, i.e. XNNPACK is complient to v1.2 vs. coreml to v1.1 etc. We can tie to to ET versions. 
- FC/BC is also a pain with this. So we have to think this through a bit more. 
- Lastly, what's the status of the suite itself, is it ready or WIP, what is the update cadence etc.

digantdesai:
> I think the following will help
> 
> 1. make it configurable, meaning that backend can note what op is supported and what op is not, and being able to inject their own ops

The whole idea is backends can't choose the test module, the suite will. Backend can use the ET APIs, if they partition it - it is expected to work correctly. If they don't partition is that is also OK.

cccclai:
> > I think the following will help
> > 
> > 1. make it configurable, meaning that backend can note what op is supported and what op is not, and being able to inject their own ops
> 
> The whole idea is backends can't choose the test module, the suite will. Backend can use the ET APIs, if they partition it - it is expected to work correctly. If they don't partition is that is also OK.

The idea is that, if I have an op A, and the backend partitioner makes a mistake and doesn't partition, and fall back to portable, the test will still pass, but it won't detect the mistakes. 

And also, say the backend supports an aten op (not portable), how do we include them here.

My goal is to make the test suit more generic.

backends/test/compliance_suite/README.md
GregoryComer: I've re-written this section to word it a bit more generally while we align on the exact expectations for backends.

backends/test/compliance_suite/__init__.py
GregoryComer: ET seems to have a bit of difficulty (outside of the backends) with bf16, so I excluded it for now. I can add a note. For example, I can't lower a graph adding 2 bf16 tensors with no delegation.

backends/test/compliance_suite/__init__.py
GregoryComer: All backends are actually enabled by default. This logic mainly exists to enable running only one backend or set of backends, since it's rare to have all backends available on a given machine and they generate a bunch of errors. Likely what we can do as a follow-up, if this works for you, is to expose the backend is_available() call through pybindings. If we can import the tester and the runtime backend is available, that we can opt in. I agree that this current solution is not ideal long-term.

backends/test/compliance_suite/__init__.py
GregoryComer: I've added some comments to clarify. I'm open to different approaches.

backends/test/compliance_suite/__init__.py
GregoryComer: I cleaned it up a little bit.

backends/test/compliance_suite/__init__.py
digantdesai: What's the issue with lowering add(bf16, bf16)? Portable op support or export or et issue?

digantdesai:
> > > I think the following will help
> > > 
> > > 1. make it configurable, meaning that backend can note what op is supported and what op is not, and being able to inject their own ops
> > 
> > 
> > The whole idea is backends can't choose the test module, the suite will. Backend can use the ET APIs, if they partition it - it is expected to work correctly. If they don't partition is that is also OK.
> 
> The idea is that, if I have an op A, and the backend partitioner makes a mistake and doesn't partition, and fall back to portable, the test will still pass, but it won't detect the mistakes.
> 
> And also, say the backend supports an aten op (not portable), how do we include them here.
> 
> My goal is to make the test suit more generic.

> it won't detect the mistakes.

Because it is not a mistake, from ET point of view this is still an OK outcome, we could generate a valid PTE. From backend point of view, there should be a report which says this backend didn't partition op A with such and such args. You or someone looking at the report from a backend partition behavior point of view should still be able to "find" it. Making a test fail requires you to know what is expected to work and that I feel is backend's job. I.e. to test a given backend partition for a "this is expected to work" is a job for individual backend's tests. I hope this helps.

> And also, say the backend supports an aten op (not portable), how do we include them here.

Gregory had a simple FACTO configuration for Linear op. In that case I would imagine if one doesn't partition we just stop the test then. I am not in favor of listing every single "do-no-decompose" op in FACTO config just a common ones. And the burden for testing all of them falls on the backend tests.

> My goal is to make the test suit more generic.

I think I understand. But I don't want this to be seen as a substitute for individual backend tests. If you say, let's add enough configurability so we can repurpose this framework to test my backend. I think that can happen, organically like XNNPACK Tester but I don't want that to be the goal for this effort esp not for GA.

backends/test/suite/operators/test_add.py
```diff
@@ -0,0 +1,82 @@
...
+        self._test_op(
+            Model(),
+            (
+                (torch.rand(2, 10) * 100).to(dtype),
+                (torch.rand(2, 10) * 100).to(dtype),
+            ),
+            tester_factory,
+        )
+
+    def test_add_f32_bcast_first(self, tester_factory: Callable) -> None:
```
digantdesai, line 44: we are not using FACTO instead because it generates too many irrelevant tests? Or this is an extra whitelist for "compliance"?

backends/test/suite/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+At a minimum, the backend will likely need to provide a custom implementation of the Partition and ToEdgeTransformAndLower stages using the appropriate backend partitioner. See backends/xnnpack/test/tester/tester.py for an example implementation.
+
+Once a tester is available, the backend flow(s) can be added in __init__.py in this directory by adding an entry to `ALL_TESTER_FLOWS`. Each flow entry consists of a name (used in the test case naming) and a function to instantiate a tester for a given model and input tuple.
+
+## Test Cases
+
+Operator test cases are defined under the operators/ directory. Tests are written in a backend-independent manner, and each test is programmatically expanded to generate a variant for each registered backend flow. The `@operator_test` decorator is applied to each test class to trigger this behavior. Tests can also be tagged with an appropriate type specifier, such as `@dtype_test`, to generate variants for each dtype. The decorators and "magic" live in __init__.py in this directory.
```
digantdesai, line 17: Do we want to highlight that "hey backend authors this is not a substitute for you individual tests, and don't use/treat it as such". I guess it is OK like XNNPACK if they use the same Tester for their internal tests. Just don't want them to add new random tests here.

backends/test/suite/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+# Function annotation for dtype tests. This instructs the test framework to run the test
+# for each supported dtype and to pass dtype as a test parameter.
+def dtype_test(func):
+    func.test_type = TestType.DTYPE
+    return func
+
+
+# Class annotation for operator tests. This triggers the test framework to register
+# the tests.
+def operator_test(cls):
```
digantdesai, line 100:
```suggestion
def operator_test_cls(cls):
```

backends/test/suite/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+    _create_tests(cls)
+    return cls
+
+
+# Generate test cases for each backend flow.
+def _create_tests(cls):
+    for key in dir(cls):
```
digantdesai, line 107: assert cls type?

backends/test/suite/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+            wrapped_test = _make_wrapped_test(test_func, dtype, tester_factory)
+            dtype_name = str(dtype)[6:]  # strip "torch."
+            test_name = f"{test_func.__name__}_{dtype_name}_{flow_name}"
+            setattr(cls, test_name, wrapped_test)
+    else:
+        raise NotImplementedError(f"Unknown test type {test_type}.")
+
+
+class OperatorTest(unittest.TestCase):
+    def _test_op(self, model, inputs, tester_factory):
```
digantdesai, line 158: static method?

backends/test/suite/README.md
```diff
@@ -0,0 +1,15 @@
+# Backend Test Suite
+
+This directory contains tests that validate correctness and coverage of backends. These tests are written such that the backend is treated as a black box. The test logic verifies that the backend is able to handle a given pattern without erroring out (not partitioning is fine) and is able to run the graphs and yield reasonable outputs. As backends may differ significantly in implementation, numerical bounds are intentionally left loose.
```
digantdesai, line 3: Add a line about why do we need this suite? 
And also for what does it mean to run this against a backend?
Lastly, how something ends up in a suite vs. not?

backends/test/suite/operators/test_add.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@operator_test
+class Add(OperatorTest):
+    @dtype_test
+    def test_add_dtype(self, dtype, tester_factory: Callable) -> None:
+        self._test_op(
+            Model(),
+            (
+                (torch.rand(2, 10) * 100).to(dtype),
```
digantdesai, line 38: can't we do `.to(dtype)` in the generated method? perhaps by passing it to `_test_op` as an arg or some other way?

digantdesai (approved):
Looks good. Left some comments.

I am glad we droped "compliance" word from this. :)

backends/test/suite/operators/test_add.py
GregoryComer: Maybe, though many stateful modules don't support non-floating weight types. If you do linear.to(torch.int32), for example, it will raise an exception. So it might need to be done on a case by case basis.

backends/test/suite/operators/test_add.py
GregoryComer: I want to explore this more. I hope to use both FACTO and manual tests. Right now, it's hard to hit all of the meaningful edge cases in a way that where they get partitioned (for example, we might not get a test that covers an important parameter set on f32 - it might be on int16 or something that doesn't get partitioned). I imagine we can experiment with adding FACTO constraints to help with this, but my initial impression is that there is value in both the FACTO tests and the standalone tests.

backends/test/suite/__init__.py
GregoryComer: Sure. I'll update this up the stack, as there's a bit of refactoring on this with the reporting diffs.

backends/test/suite/README.md
GregoryComer: I've added some additional wording here, clarifying that it's experimental and under active development. Let's chat about evolution of the test suite as a follow-up.

backends/test/suite/__init__.py
GregoryComer: Makes sense. I'll stack it on top, as a follow-up, to make it easier to refactor (since there are many op test diffs stacked).

backends/test/suite/README.md
GregoryComer: Good callout. I've updated the wording.

PR #11985, opened 2025-06-25
Prim ops replace abort with error code

Conarnar:
### Summary
We want to minimize the scenarios where ET fails fatally. If we need to check a precondition inside a kernel rather then calling ET_CHECK which internally dispatches to ABORT we should call ET_KERNEL_CHECK which sets an error state and returns.

### Test plan
Replaced TestNegScalarWithTensorDies with TestNegScalarWithTensorFails. Before, it would expect an abort when negating a tensor. The new test expects a failure state instead.

The test can be run with `test/run_oss_cpp_tests.sh`

### DIFF
--- a/kernels/prim_ops/register_prim_ops.cpp
+++ b/kernels/prim_ops/register_prim_ops.cpp
@@ -22,9 +22,16 @@ namespace function {
 
 namespace {
 
-#define __ET_PRIM_OP_ERROR_IMPL(a, b, context)                \
-  else {                                                      \
-    ET_KERNEL_CHECK(context, false, InvalidType, /* void */); \
+#define __ET_PRIM_OP_ERROR_IMPL(a, b, context) \
+  else {                                       \
+    ET_KERNEL_CHECK_MSG(                       \
+        context,                               \
+        false,                                 \
+        InvalidType,                           \
+        /* void */,                            \
+        "%zu, %zu",                            \
+        (size_t)a.tag,                         \
+        (size_t)b.tag);                        \
   }
 
 #define __NUMBER_ET_PRIM_OP_IMPL(operator, stack, context) \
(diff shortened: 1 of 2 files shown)


kernels/prim_ops/register_prim_ops.cpp
```diff
@@ -22,12 +22,11 @@ namespace function {
 
 namespace {
 
-#define __ET_PRIM_OP_ERROR_IMPL(a, b, context)                     \
-  else {                                                           \
-    ET_CHECK_MSG(false, "%zu, %zu", (size_t)a.tag, (size_t)b.tag); \
+#define __ET_PRIM_OP_ERROR_IMPL(a, b, context)                \
+  else {                                                      \
+    ET_KERNEL_CHECK(context, false, InvalidType, /* void */); \
```
lucylq: Could we use `ET_KERNEL_CHECK_MSG` to keep the format string/error message?

e.g.
```
ET_KERNEL_CHECK_MSG(context, false, InvalidType, "%zu, %zu", (size_t)a.tag, (size_t)b.tag); 
```

kernels/prim_ops/test/prim_ops_test.cpp
```diff
@@ -325,7 +325,9 @@ TEST_F(RegisterPrimOpsTest, TestNegScalarWithTensorDies) {
   }
 
   // Try to negate a tensor, which should cause a runtime error.
-  ET_EXPECT_DEATH(getOpsFn("executorch_prim::neg.Scalar")(context_, stack), "");
+  ET_EXPECT_KERNEL_FAILURE(
+      context_,
```
lucylq: Linter failure here - can you try run `lintrunner -a`?

See: https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner

lucylq (approved):
Looks good, thanks for adding this!

PR #11989, opened 2025-06-26
Qualcomm AI Engine Direct - Ensure that math invariant ops don't change scale and offset

shewu-quic:
Summary:
- After QNN 2.35, op validation will check math invariant op should not change scale and offset. So, we should replace annotate_single_in_single_out with annotate_single_in_share_out to ensure not to change scale and offset
- Fix the error for internal CI
- Fix the bug for batch norm op

cc: @haowhsu-quic, @cccclai , @winskuo-quic

### DIFF
--- a/backends/qualcomm/_passes/remove_redundancy.py
+++ b/backends/qualcomm/_passes/remove_redundancy.py
@@ -29,6 +29,7 @@ class RemoveRedundancy(ExportPass):
             # remove channel_last / contiguous _to_copy if '_skip_dim_order' is set to True
             exir_ops.edge.aten._to_copy.default: self._to_copy_op_condition,
             torch.ops.aten._assert_tensor_metadata.default: self._default_condition,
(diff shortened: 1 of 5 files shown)


cccclai:
Is there a list of math invariant op from qnn?

shewu-quic:
> Is there a list of math invariant op from qnn?

It can be found in qnn doc. Operations>Backend supplements>HTP>Quantization.
And also if you modify the quant_attr for math invariant op, you will get the error such as the below.
```
[ERROR] [Qnn ExecuTorch]:  <E> This op is math invariant, input0 scale (0.000308) and output0 scale (0.000306) should be identical but is not.
 
[ERROR] [Qnn ExecuTorch]:  <E> QnnBackend_validateOpConfig failed 3110
 
[ERROR] [Qnn ExecuTorch]:  <E> Failed to validate op aten_expand_copy_default_15 with error 0xc26
```
image

cccclai (approved):

Issue #11542, opened 2025-06-11
[Draft] CoreML GA model testing


metascroy:
### 🐛 Describe the bug

**Repro steps:**

To export a GA model, do the following:

1. Create a conda environment and install ExecuTorch
2. Check out the coreml-ga-testing branch from https://github.com/metascroy/optimum-executorch/tree/coreml-ga-testing, and install optimum-executorch.
3. cd into optimum-executorch and run the following export script:
```
python export_ga_models.py --output_dir "/path/to/directory/for/exported/models" --et_repo_dir "/path/to/executorch/repo/directory"
```

The above script will export various GA models with the ExecuTorch CoreML backend, the ExecuTorch XNNPACK backend, and the standalone CoreML flow.  There will be 1 folder per model, and within that a folder for coreml, coreml_standalone, and xnnpack.

<img width="335" alt="Image" src="https://github.com/user-attachments/assets/8a4909e7-1353-4073-8c09-76368d990ef1" />

If export fails, instead of a model file, you'll see a text file with a stack trace.

To run the GA models with pybindings, run the following:
```
python run_ga_models.py --model_dir "/path/to/directory/for/exported/models"    
```
This will run the exported models 50 times each, and report the average inference time.

Run results: [P1837452332](https://www.internalfb.com/phabricator/paste/view/P1837452332)

### Versions

NA

YifanShenSZ:
> 1. When take_over_mutable_buffer=True, smollm fails with "Buffer output getitem_1 does not point to a buffer that exist", see [smollm_coreml_executorch_exception.txt](https://github.com/user-attachments/files/20696368/smollm_coreml_executorch_exception.txt).

Looks like an executorch issue? That after we take over the mutable buffers, the new torch.export graph is not generated correctly

> 3. Model does not quantize using default iOS15, see [yolos-quantize_coreml_executorch_exception.txt](https://github.com/user-attachments/files/20696406/yolos-quantize_coreml_executorch_exception.txt)

This is expected
1. iOS 16 introduced weight-only int8 quantization
2. iOS 17 introduced int8 activation quantization
3. iOS 18 introduced weight-only int4 blockwise quantization

metascroy:
@YifanShenSZ I hopefully have a fix for issue 1 here: https://github.com/pytorch/executorch/pull/11990

On the quantization: I think people have observed that when using coremltools directly they can quantize with an iOS 15 taget, but when using CoreML via ET, they have to use iOS17.  Is that correct?

[...]

PR #11990, opened 2025-06-26
Fix mutable buffer tagging in ExecuTorch

metascroy:
This PR is intended to address https://github.com/pytorch/executorch/issues/11542.  A mutable buffer is tagged for consumption by the delegate, but the graph_signature.output_spec keeps the mutable buffer there because it is not in the state_dict (persistent=false), but instead inside of constants.


JacobSzwejbka:
@SS-JIA Is this the issue you were asking me about today?

JacobSzwejbka (approved):

metascroy:
@angelayi can you take a look before I merge?

metascroy:
> @angelayi can you take a look before I merge?

I'll merge now to make the branch cut since @JacobSzwejbka approved, but have a look when you get a chance.

PR #11996, opened 2025-06-26
Arm backend: Add VIT test

oscarandersson8218:
Adds unittest for VIT model. 

Change-Id: Icb1923a9b2cc4740ceff0d4daccba1d12ead6e5f


backends/arm/test/models/test_vit.py
```diff
@@ -0,0 +1,76 @@
...
+    EthosU55PipelineBI,
+    EthosU85PipelineBI,
+    TosaPipelineBI,
+    TosaPipelineMI,
+)
+
+from torchvision import models, transforms
+
+vit_b_16_model = models.vit_b_16(weights="IMAGENET1K_V1")
+vit = vit_b_16_model.eval()
```
digantdesai, line 22: ❤️

oscarandersson8218:
Model is not suitable for testing test/models as it's too big. Closing this.

PR #11997, opened 2025-06-26
[et] generate debug handle before opeartor decomposition

Gasoonjia:
* __->__ #11997

This diff update the debug handle generation, from each node in the edge program having a individual debug handle, to all nodes having a same ancestor in export graph sharing a same debug handle, which update the start point of tracing our node transformation from edge graph to exported graph.

Differential Revision: [D76860368](https://our.internmc.facebook.com/intern/diff/D76860368/)


exir/passes/debug_handle_generator_pass.py
```diff
@@ -4,31 +4,75 @@
...
+        def _get_greatest_ancestor_node_identifier(node: Node) -> str:
+            """Get the identifier of the greatest ancestor node of the given node.
+
+            The identifier is the concatenation of the node name and graph id of the
+            greatest ancestor node, where the graph id is the unique id for every graph
+            module in the export flow and node name is unique within the same graph module.
+            """
+
+            node_source = node.meta[FROM_NODE_KEY]
+            node_source = node_source[-1]
```
kimishpatel, line 38: this is only populated for unlifted graph when `.module` is called? if so how is it guranteed that this is present

exir/passes/debug_handle_generator_pass.py
```diff
@@ -4,31 +4,75 @@
...
+                FROM_NODE_KEY in node.meta
+            ), f"Node {node} does not have meta key {FROM_NODE_KEY}"
+
+            greatest_ancestor_node_id = _get_greatest_ancestor_node_identifier(node)
+
+            debug_handle = (
+                len(source_node_id_to_debug_handle) + 1
+                if greatest_ancestor_node_id not in source_node_id_to_debug_handle
+                else source_node_id_to_debug_handle[greatest_ancestor_node_id]
+            )
```
kimishpatel, line 65: Ok one of my concern is this: is this gonna impact any existing debug utils that rely on debug handle? earlier each node had its own debug handle, but not we are giving a bunch of nodes that might belong to the same original node (e.g. full aten graph nodes?).

Other question is how does this work for delegates that use to_edge_tranform_and_lower, if this pass is run only after to_edge?

devtools/inspector/_inspector_utils.py
```diff
@@ -303,14 +303,18 @@ def gen_graphs_from_etrecord(
     return op_graph_map
 
 
+# One debug handle should only be associated with one node. We are in the middle of migrating debug handle generation
+# from graph after to_edge to graph after torch.export, one every debug handle in exported graph may be associated with multiple nodes in to_edge
```
kimishpatel, line 307: Put before and after state

kimishpatel (commented):
For the most part looks good. I left a couple of comments, one specifically around relation of to_edge_lower_and_transform to debug handle (not related ot this diff)

exir/passes/debug_handle_generator_pass.py
Gasoonjia: no it is not only populate that `.module` is explicitly called. In our case it is guarantee generated after run_decompostition so we will have that when generating debug handle.

exir/passes/debug_handle_generator_pass.py
Gasoonjia: For first question, yes like devtool and some other infra or utils are depending on the debgu handle and in the following diffs i will do the migration. There're some testcases rely on debug handle for check and I've talked with the owners and they are ok with the update. Now ci is all clean.

For second question, not sure if I fully understand your concern, but our current debug handle generation pass also locates at the end of to_edge. Our update does not change the location. But we do have concern about whether or not the change of debug handle generation may impact the correctness of the debug handels in the downstream e.g. in delegate, though ci is all clean. And we will pay attention to it when working on the numeric descrepancy dtector on delegated model.

exir/passes/debug_handle_generator_pass.py
kimishpatel: > it is guarantee generated after run_decompostition so we will have that when generating debug handle.

There in my question around to_edge_transform_and_lower. although that specifically is unrelated to this pr

exir/passes/debug_handle_generator_pass.py
kimishpatel: Actually we should discuss this second paragraph further. THe issue is that `to_edge_tranform_and_lower` happens before we run decomposition, is that right? If so you dont have debug handles generated no? If no debug handles then no debug information. I would like to be either wrong about my assumption or we need to fix it

kimishpatel: For the first paragraph, do you forsee any disruption in existing user flows for debugging? That would be one thing that would make me block this PR, so having an answer on that would be nice

devtools/inspector/_inspector_utils.py
Gasoonjia: i've updated the comment to reflect the function behavior before, during and after the migration.

exir/passes/debug_handle_generator_pass.py
Gasoonjia: > The issue is that to_edge_tranform_and_lower happens before we run decomposition, is that right? 

No operator decomposition should be happened inside `to_edge_transform_and_lower`. `to_edge` stage happened in side `to_edge_transform_and_lower` and operator decompostition happened inside `to_edge` stage. https://github.com/pytorch/executorch/blob/main/exir/program/_program.py#L1129 

>  forsee any disruption in existing user flows for debugging

no. One heart break context is no debug tool is actually using debug handle right now. The numerical discrepancy detector will be the first one.

kimishpatel (approved):

PR #12004, opened 2025-06-26
Support while in pass base

### LINKED ISSUE #8769: Support higher-order while_loop and scan torch ops

Pytorch provides several higher-level control flow ops, including [while_loop](https://github.com/pytorch/pytorch/blob/main/torch/_higher_order_ops/while_loop.py) and [scan](https://github.com/pytorch/pytorch/blob/main/torch/_higher_order_ops/scan.py). These ops have to be used explicitly in the PyTorch model (Dynamo can't infer them from Python control flow), but they are useful in expressing more complex models. As a first step to try to handle complex patterns without needing to chop up the model and port logic to C++, we should support some of the higher-order ops in ET.

I believe while_loop, and scan should be a good starting point, as they are sufficient to express many of the data dependent patterns. This may allow for writing some generator loops in pure PyTorch, specially for non-multi-turn generation.

Currently, these ops are not supported on ET. It's likely that we will need both AOT and runtime work to do this. Exporting a simple cond model gives the following error. Simple scan and while_loop models also fail with various errors

Edit: From Tarun's comment, cond should be supported, but it seems like there are some issues to resolve, as I haven't been able to use cond when I've tried it. I will create specific tasks for cond. I've removed it from this issue for now.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @JacobSzwejbka @angelayi

angelayi:
Summary: Fixes https://github.com/pytorch/executorch/issues/8769

Differential Revision: D77382557

### DIFF
--- a/exir/pass_base.py
+++ b/exir/pass_base.py
@@ -340,6 +340,11 @@ class _ExportPassBase(PassBase):
             elif target == torch.ops.higher_order.cond:
                 pred, true_fn, false_fn, inputs = args
                 return self.callback.call_cond(pred, true_fn, false_fn, inputs, meta)
+            elif target == torch.ops.higher_order.while_loop:
+                cond, body, carried_inputs, additional_inputs = args
+                return self.callback.call_while(
+                    cond, body, carried_inputs, additional_inputs, meta
+                )
             elif target == torch.ops.higher_order.map_impl:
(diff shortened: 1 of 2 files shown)


lucylq (approved):
thanks for adding this support!

PR #12016, opened 2025-06-26
Add Pybindings for Program.h/cpp

Conarnar:
### Summary

Today our python apis in executorch.runtime are implemented off of extension/pybindings which only offers a module api. We would like to migrate to having the lower level ET api exposed to python directly and then writing the module api in python. The first step to this is adding pybindings for Program.

Bindings for the class Program and its methods num_methods and get_method_name were added.

### Test plan

Tests were added to `extension/pybindings/test/make_test.py`
1. test_program_methods_one -- verifies num_methods and get_method_name works with one method
2. test_program_methods_multi -- verifies num_methods and get_method_name works with multiple methods
3. test_program_method_index_out_of_bounds -- verifies get_method_name raises a runtime error if index is out of bounds

### DIFF
--- a/extension/pybindings/pybindings.cpp
+++ b/extension/pybindings/pybindings.cpp
@@ -965,6 +965,89 @@ struct PyModule final {
   }
 };
 
+inline std::unique_ptr<DataLoader> loader_from_buffer(
+    const void* ptr,
+    size_t ptr_len) {
+  return std::make_unique<BufferDataLoader>(ptr, ptr_len);
+}
+
+inline std::unique_ptr<DataLoader> loader_from_file(const std::string& path) {
+  Result<MmapDataLoader> res = MmapDataLoader::from(
+      path.c_str(), MmapDataLoader::MlockConfig::UseMlockIgnoreErrors);
+  THROW_IF_ERROR(
+      res.error(),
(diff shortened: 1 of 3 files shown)


extension/pybindings/pybindings.cpp
```diff
@@ -965,6 +965,99 @@ struct PyModule final {
...
+  std::string get_method_name(size_t method_index) const {
+    Result<const char*> res = program_->get_method_name(method_index);
+    THROW_IF_ERROR(
+        res.error(),
+        "Failed get method name, error: 0x:%" PRIx32,
+        static_cast<uint32_t>(res.error()));
+    return std::string(res.get());
+  }
+
+  std::string get_output_flattening_encoding(std::string method_name) const {
```
lucylq: This API is deprecated, should it still be exposed? Maybe mark it as deprecated as well. 

https://github.com/pytorch/executorch/blob/7e28a044d7acf8ae405d9d5a37d11ceb12b90899/runtime/executor/program.h#L168

cc @JacobSzwejbka

JacobSzwejbka (approved):

extension/pybindings/pybindings.cpp
Conarnar: Method removed

PR #12018, opened 2025-06-26
Show output of the `size` command in size test scripts

swolchok:
I find this more useful than raw file size, because it breaks down size by section of the binary and splits out debug info.

PR #12017, opened 2025-06-26
Unbreak build_size_test.sh (-Wvla-cxx-extension) on my local machine

swolchok:
I guess our CI compiler doesn't have this warning, but it's triggering for me locally. The warning is fundamentally correct, so I just fixed it.

PR #12019, opened 2025-06-26
Build size_test binaries with debug info

swolchok:
The CI workflow already runs strip before checking size, so this shouldn't break it (and we'll see if it does on this PR). Having debug info makes it much easier to investigate issues with bloaty.


swolchok:
* #12031
* #12023
* __->__ #12019
* #12045
* #12018
* #12017

swolchok:
looks like we need to run `strip -s`

mergennachin (approved):

swolchok:
this one looks like a can of worms and the benefit for the cost isn't there. abandoning.

PR #12037, opened 2025-06-26
Introduce sym_max and sym_min ops to executorch

### LINKED ISSUE #11988: sym_max support in ExecuTorch
### 🐛 Describe the bug

Hit this issue when I was debugging the gemma-3 error in `optimum-executorch` ([CI link](https://github.com/huggingface/optimum-executorch/actions/runs/15883131195/job/44788420294?pr=73#step:5:1730)). 

Upon checking the complained `op_overload`, the function object is actually a `sym_max` which is introduced via this line

https://github.com/huggingface/transformers/blob/1d45d90e5d1552eccb6d8cc9b7bba283ccefb808/src/transformers/cache_utils.py#L1725-L1732
```
        if self.is_sliding[layer_idx]:
            query_length = cache_position.shape[0]
            ...
            local_mask_kv_length = max(query_length, self.sliding_window)
            return local_mask_kv_length, local_mask_kv_offset
```

in the upstream `transformers` in the `HybridCache` class, which is expected as now we set the `cache_poistion` dim to be dynamic in https://github.com/huggingface/optimum-executorch/pull/73.

The export works fine to have the `sym_max` in the exported graph (`test_hybrid_cache_exportability` passed in the `transformers`), but when further lowering the graph to executorch, the call `to_executorch` failed.

On chatting with @kimishpatel , it seems like we are missing implementation for `sym_max` as an op.

### Versions

trunk

kimishpatel:
Added corresponding tests as well

Fixes https://github.com/pytorch/executorch/issues/11988

### Test plan
test added

### DIFF
--- a/kernels/prim_ops/register_prim_ops.cpp
+++ b/kernels/prim_ops/register_prim_ops.cpp
@@ -12,6 +12,7 @@
 #include <executorch/runtime/kernel/kernel_includes.h>
 #include <executorch/runtime/kernel/operator_registry.h>
 
+#include <algorithm>
 #include <cmath>
 
 using torch::executor::function::et_copy_index;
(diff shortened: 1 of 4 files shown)


kernels/prim_ops/register_prim_ops.cpp
```diff
@@ -114,6 +115,34 @@ static Kernel prim_ops[] = {
...
+        "executorch_prim::sym_max.Scalar",
+        [](KernelRuntimeContext& context, EValue** stack) {
+          (void)context;
+          EValue& a = *stack[0];
+          EValue& b = *stack[1];
+          EValue& out = *stack[2];
+          if (a.isInt() && b.isInt()) {
+            out = EValue(std::max(a.toInt(), b.toInt()));
+          } else {
+            ET_CHECK_MSG(false, "sym_max only supports int inputs, got %zu, %zu", (size_t)a.tag, (size_t)b.tag);
```
JacobSzwejbka: @Conarnar has a PR up to change all of these checks in prim ops to non fatal

kernels/prim_ops/register_prim_ops.cpp
```diff
@@ -120,6 +121,42 @@ static Kernel prim_ops[] = {
...
+        "executorch_prim::sym_max.Scalar",
+        [](KernelRuntimeContext& context, EValue** stack) {
+          (void)context;
+          EValue& a = *stack[0];
+          EValue& b = *stack[1];
+          EValue& out = *stack[2];
+          if (a.isInt() && b.isInt()) {
+            out = EValue(std::max(a.toInt(), b.toInt()));
+          } else {
+            ET_CHECK_MSG(
```
JacobSzwejbka: Please use ET_KERNEL_CHECK_MSG instead so that we dont abort the process unecessarily

JacobSzwejbka (changes requested):

JacobSzwejbka (approved):

PR #12038, opened 2025-06-26
Add gtest macros and matchers for Result<T> and Error.

hsharma35:
Summary: Adds matcher support for Result and Error types, similar to absl::Status.

Differential Revision: D77409312

### DIFF
--- a/runtime/core/testing_util/error_matchers.h
+++ b/runtime/core/testing_util/error_matchers.h
@@ -1,0 +1,303 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @file
+ * Testing utilities for working with `executorch::runtime::Result<T>` and
+ * `executorch::runtime::Error`. Provides matchers similar to `absl::StatusOr`
+ * and `absl::Status`.
+ *
+ * Defines the following utilities:
+ *
+ *   ===============
+ *   `IsOkAndHolds(m)`
+ *   ===============
+ *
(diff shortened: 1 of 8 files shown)


runtime/core/testing_util/error_matchers.h
```diff
@@ -0,0 +1,343 @@
...
+    std::move(*et_result__);           \
+  })
+
+namespace executorch {
+namespace runtime {
+namespace testing {
+namespace internal {
+namespace {
+
+constexpr const char* toString(const Error error) {
```
larryliu0820: Can you put this in Error? Would be easier for future changes to Error also change this API.

larryliu0820: Also this can be templatized

runtime/core/testing_util/test/error_matchers_test.cpp
```diff
@@ -0,0 +1,134 @@
+/*
```
larryliu0820: Naming convention is `test_error_matchers`I guess?

runtime/core/testing_util/test/targets.bzl
```diff
@@ -0,0 +1,19 @@
...
+load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime")
+
+def define_common_targets():
+    """Defines targets that should be shared between fbcode and xplat.
+
+    The directory containing this targets.bzl file should also contain both
+    TARGETS and BUCK files that call this function.
+    """
+    runtime.cxx_test(
+        name = "error_matchers_test",
```
larryliu0820:
```suggestion
        name = "test_error_matchers",
```

larryliu0820 (approved):
Can you add a test for real code (not `error_matchers_test.cpp`) that uses these matchers?

PR #12047, opened 2025-06-27
Add llm runner into apple frameworks build

larryliu0820:
Remove `EXECUTORCH_BUILD_EXTENSION_LLM` and promote `EXECUTORCH_BUILD_EXTENSION_LLM_RUNNER` to be an overridable option.

Enable it by default in llm and apple preset.


mergennachin:
Looks like ios and macos presets tests are failing?

PR #12048, opened 2025-06-27
Enable Gemma3 1B on ExecuTorch

### LINKED ISSUE #9250: Add Gemma 3

Add support for the text decoder backbone of the new [Gemma 3](https://github.com/google-deepmind/gemma) (1B / 4B for edge) model.

The model architecture should slide right into our `llama_transformer.py`, with the exception of the interspersed sliding window local attention layers as specified in the [technical report](https://storage.googleapis.com/deepmind-media/gemma/Gemma3Report.pdf), which will require some modifications to our model code. Luckily the way this is done is through [slicing the attention mask](https://github.com/huggingface/transformers/blob/1c287aecfcf94f569eb8740be5be43bda6750682/src/transformers/models/gemma3/modeling_gemma3.py#L411-L431) and not the kv cache, so we can keep using our static kv cache implementation.

Checkpoints are on HuggingFace:
- [google/gemma-3-4b-it](https://huggingface.co/google/gemma-3-4b-it)
- [google/gemma-3-1b-it](https://huggingface.co/google/gemma-3-1b-it)

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

jackzhxng:
Sample xnnpack 8da4w export:
```
python -m extension.llm.export.export_llm
--config examples/models/gemma3/config/gemma3_xnnpack_q8da4w.yaml
+base.params="examples/models/gemma3/config/1b_config.json"
+export.output_name="gemma3_1b_x_8da4w.pte"
```

### DIFF
--- a/examples/demo-apps/apple_ios/LLaMA/LLaMA/Application/Constants.swift
+++ b/examples/demo-apps/apple_ios/LLaMA/LLaMA/Application/Constants.swift
@@ -26,5 +26,14 @@ You are a helpful assistant.
     
     public static let llama3PromptTemplate = "<|begin_of_text|><|start_header_id|>user<|end_header_id|>%@<|eot_id|><|start_header_id|>assistant<|end_header_id|>"
     
-public static let phi4PromptTemplate = "<|user|>%@<|end|><|assistant|>"
+    public static let phi4PromptTemplate = "<|user|>%@<|end|><|assistant|>"
+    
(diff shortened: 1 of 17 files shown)


cccclai:
@haowhsu-quic @winskuo-quic @chunit-quic @DannyYuyang-quic @shewu-quic fyi the changes to enable Gemma3 1B on the cpu path. Maybe something can be shared across.

PR #12051, opened 2025-06-27
Add code to Bundleio to generate error stats

zingo:
Add a way to get error stats/metrics between actual and reference output.


zingo:
Hi @digantdesai and @mergennachin this PR will add a new method/API to BundleIO and maybe you want to involve the proper people about this :) Also is the metric sane? And the way to propagate it back to the runner?

My basics intentions is to be able to log/track models over time in a better way then PASS/FAIL on a set atol/rtol, as it easily miss if we could have set rtol/atol lower when improving stuff. Something like this is also useful to get a good guess of what atol/rtol could be to make it work instead of a lot of trail and error with different values.

devtools/bundled_program/bundled_program.cpp
```diff
@@ -346,6 +347,116 @@ ET_NODISCARD Error load_bundled_input(
...
+          double abs_err = std::abs(a_data[k] - e_data[k]);
+          double relative_divider =
+              std::max(std::abs(a_data[k]), std::abs(e_data[k]));
+          relative_divider = std::max(relative_divider, eps);
+          double relative_err = abs_err / relative_divider;
+
+          sum_abs += abs_err;
+          max_abs = std::max(max_abs, abs_err);
+          sum_rel += relative_err;
+          max_rel = std::max(max_rel, relative_err);
```
zingo, line 433: Is this good? I'm no ML-math-stats person so if this can be improved we should in PR or after :)

digantdesai:
I like this general direction of getting more than true/false. And since its opt in, we care a bit less about the binary size overhead. I will let @Gasoonjia weigh in. He is looking at this for AoT with multiple "distance" measures. Thanks @zingo.

devtools/bundled_program/bundled_program.cpp
Gasoonjia: I think it is good for regular cases e.g. both input and target are in float32 and good enough for bundled program. We shouldn't expect that bundled program can cover all cases like quantization

Gasoonjia (approved):
Thansk @zingo for the update and overall l love the update. It makes the error msg more meaningful. 

Also spoiler alert we will have a new api in 0.7 release (https://github.com/pytorch/executorch/blob/main/devtools/inspector/_inspector.py#L1365) for comparing intermediate output in operator-level, beyond what we have right now in bundled program for only compare the final result! Stay tune and we will have a doc for better demonstration!

digantdesai:
Thank you both. @zingo or @Gasoonjia we should use this `compute_method_output_error_stats` for other runners as well.

PR #12054, opened 2025-06-27
Qualcomm AI Engine Direct - GA Static QWEN2.5 0.5B

winskuo-quic:
### Summary
Static Qwen2.5 0.5b enablement.
Please use 16a8w for qwen as other quant configs are not yet fully supported.

Script
`python examples/qualcomm/oss_scripts/llama/llama.py -b build-android  -s $DEVICE -m SM8650 --prompt "Hello, how are you?" --temperature 0 --model_mode kv--max_seq_len 128 --ptq 16a8w --decoder_model qwen2_5`

#### Stats
SM8650
image

SM8750
image

### Test plan
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleLLMScript.test_qwen2_5 --model SM8650 --build_folder build-android/ --executorch_root . -s $DEVICE --artifact ./qwen2_5`

Author: @haowhsu-quic, @winskuo-quic


winskuo-quic:
Hi @billmguo, @cccclai,

This is a draft for qwen2.5 0.5 using static nn.Module structure.
Please notice this is just a draft as the accuracy for the model is still bad.

Thanks

cccclai:
Thanks! If it's just the accuracy is bad, @rohansjoshi has been making progress on this. Can qwen model run with this PR (don't need to be accurate)?

examples/qualcomm/oss_scripts/qwen/model/static_qwen.py
```diff
@@ -0,0 +1,577 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+from executorch.examples.models.llama.rope import precompute_freqs_cis
+from executorch.examples.qualcomm.oss_scripts.llama.model.static_llama import (
+    apply_rotary_emb_single,
+)
+from transformers.configuration_utils import PretrainedConfig
+
+
+class RMSNorm(nn.Module):
```
cccclai: I think we need to have the same static llama file (maybe with some branches) instead of starting a new static qwen model

winskuo-quic:
> Thanks! If it's just the accuracy is bad, @rohansjoshi has been making progress on this. Can qwen model run with this PR (don't need to be accurate)?

Yes. The can model can run with this PR.
I have an example script and stats shared in the summary section.

examples/qualcomm/oss_scripts/qwen/model/static_qwen.py
winskuo-quic: Thanks for reviewing the PR. We are trying to have a static decoder file that can be shared across all GA decoder models + LLama.
For this draft, we are just making sure e2e flow is working, and we will work on merging all static decoder files to 1.

rohansjoshi:
Two things you can do to fix accuracy are:
1) In `make_quantizer`, add the argument `act_observer=MinMaxObserver` (default is `MovingAverageMinMaxObserver`, which is much worse)
2) To see better outputs from the script, flush the kv cache before each token generation loop (i.e. add the line `_, atten_mask, _, k_caches, v_caches = qc_model.get_example_inputs()` before the loops)

If make both these changes in `qwen.py` you see a much better response. @haowhsu-quic, @winskuo-quic

examples/qualcomm/oss_scripts/llama/runner/runner.cpp
```diff
@@ -117,50 +139,32 @@ Error Runner::load() {
       break;
   }
 
-  auto eos_ids = std::make_unique<std::unordered_set<uint64_t>>();
-  // TODO: remove this once we could release the new tokens used for the
-  // tokenizer
-  if (tokenizer_ != nullptr) {
```
winskuo-quic, line 123: Hi @cccclai,
I am unsure about what situation will go into this `if` statement.
It is added in this PR: https://github.com/pytorch/executorch/pull/12285
Could you verify if this PR does not break the logic?

cccclai: @limintang can you take a look at the changes?

cccclai:
Oops I click merge by mistake...reverting in https://github.com/pytorch/executorch/pull/12506...I have some comments for this PR

examples/qualcomm/oss_scripts/llama/hf_converter/convert_config.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) Qualcomm Innovation Center, Inc.
+# All rights reserved
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+def convert_configs(config):
```
cccclai, line 8: Can you share where this config come from and how we can scale to more models?

examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -69,27 +75,26 @@
...
 from torchao.quantization.pt2e import MinMaxObserver
 from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e
+from transformers import AutoConfig, AutoTokenizer
 
 sys.setrecursionlimit(4096)
 FORMAT = "[%(levelname)s %(asctime)s %(filename)s:%(lineno)s] %(message)s"
 logging.basicConfig(level=logging.INFO, format=FORMAT)
 logging.getLogger().setLevel(logging.INFO)
 
+HUGGING_FACE_REPO_IDS = {"qwen2_5": "Qwen/Qwen2.5-0.5B"}
```
cccclai, line 96: Each model will have their own config and I feel like it can be shared with the cpu model https://github.com/pytorch/executorch/blob/main/extension/llm/export/config/llm_config.py#L32

It doesn't have to be this PR, and I'd like to hear your thoughts on this

examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -646,8 +674,14 @@ def permute(w, heads):
 
     if args.ptq:
         start_quantize_ts = time.time()
-        custom_annotations = (annotate_matmul_16a8w,)
-        if args.llama_model == "stories110m":
+        custom_annotations = (
+            # For qwen2.5, skip annotate_conv can improve result.
```
cccclai, line 678: Why is it the case?

cccclai:
After the comments are address, I can revert the revert again...

examples/qualcomm/oss_scripts/llama/llama.py
winskuo-quic: This skip annotate_conv here actually refers to annotating the conv as 8a4w, not skipping quantize conv2d, which means for qwen the conv2d is 16a8w. I have tested both skip and not skip annotate conv as 8a4w, and the CPU QDQ results are a lot better when I skip annotate conv2d to 8a4w.

examples/qualcomm/oss_scripts/llama/hf_converter/convert_config.py
winskuo-quic: The config can be found over here:
https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2/configuration_qwen2.py.
For now, I checked qwen and gemma, and I think most configs are following the same naming, where they all use `PretrainedConfig` as the base class. Ideally, this function should be able to support most HF decoder models, but I will need to test them 1 by 1 to confirm they can all be supported using this function.

examples/qualcomm/oss_scripts/llama/llama.py
winskuo-quic: I would like to confirm the config you are mentioning here.
Do you mean that instead of creating this dictionary, maybe we could contribute to the dict here?
https://github.com/pytorch/executorch/blob/b14cb22df6cf876c0fca4eeb1ef7bea0d49f4fa7/examples/models/llama/export_llama_lib.py#L114

winskuo-quic:
> After the comments are address, I can revert the revert again...

Also, I tried rebasing to mainline to test if tokenizer's issue is resolved (https://github.com/pytorch/executorch/pull/12333). However, after rebasing, I encountered another error:

`libc++abi: terminating due to uncaught exception of type std::runtime_error: Unsupported Normalizer type: NFC`.

PR #12062, opened 2025-06-27
Fixes in to_executorch for while

angelayi:
Summary: Fixes while loop support in executorch AOT up until emitter (https://github.com/pytorch/executorch/issues/8769#issuecomment-3012018917)

Differential Revision: D77452951

### DIFF
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -1480,10 +1480,11 @@ class _Emitter(torch.fx.Interpreter):
             # pyre-ignore
             return self._emit_free(args[0])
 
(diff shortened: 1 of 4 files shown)


JacobSzwejbka:
Memory Planning didnt require any changes? Does it work with dynamic shape?

angelayi:
> Memory Planning didnt require any changes? Does it work with dynamic shape?

My next error is in the emitter soo I guess memory planning doesn't require any changes? I didn't test for correctness though.

JacobSzwejbka (approved):
Can you make sure the xnnpack failures are not real

Issue #11447, opened 2025-06-06
Pass Dependencies when Proposing Partitions


mcr229:
### 🚀 The feature, motivation and pitch

Currently an issue with the capability based partitioner (https://github.com/pytorch/pytorch/blob/2e2ea7290a1cf2da3c3efd6e6ad4836bb94beaeb/torch/fx/passes/infra/partitioner.py#L78) is that there is no way to specify node dependencies together. For example if there is a graph that looks like:
```

input --> choose_q_params --> q --> dq --> linear --> (not_supported_op) --> bmm --> output
                                       \
                                         -------------------------------> linear --> output
```

In this case since both linear patterns are dynamically quantized and share the same QDQ Chain then then partition should be:

```
Partition 1
input --> choose_q_params --> q --> dq --> linear --> output
                                       \
                                         -------------------------------> linear --> output

Partition 2
(not supported op output) --> bmm --> output
```

However, in reality we can get a case where we see:
```
Partition 1
input --> choose_q_params --> q --> dq --> linear --> output
                                       \
                                         -------------> output

Partition 2
input --> bmm --> output

dq_output --> linear --> output
```

This is problematic because in the second linear we lose the semantics of a quantized linear from the graph. We need a way to pass in groups of nodes that must be partitioned together

digantdesai:
> This is problematic because in the second linear we lose the semantics of a quantized linear from the graph

Is this because cqp -> q are not part of the 2nd partition? Can we dup cqp->q->dq instead of just dup dq? Just an idea.

PR #12072, opened 2025-06-27
Pass Dependencies When Proposing Partitions VIA New Group-Based Partitioner

leafs1:
### Summary
The existing capability-based partitioner in PyTorch's FX module lacks the ability to specify node dependencies that must be partitioned together. This can lead to incorrect partitioning of dynamically quantized linear patterns, resulting in the loss of quantization semantics. For example, in a graph with shared QDQ (Quantize-Dequantize) chains, the partitioner may incorrectly separate nodes that should remain together, leading to incorrect execution semantics.

This PR addresses this issue by adding a new group-based partitioner that allows users to specify groups of nodes that must be partitioned together.

### Test plan
I've created test cases that replicate existing QDQ tests, as well as additional graphs with different node dependencies. These tests aim to verify that the new partitioner correctly groups nodes together based on the specified dependencies.


exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -0,0 +1,272 @@
...
+    def __init__(
+        self,
+        graph_module: GraphModule,
+        operator_support: OperatorSupportBase,
+        allows_single_node_partition: bool = False,
+        non_compute_ops: Optional[Sequence[str]] = None,
+        allowed_single_node_partition_ops: Optional[Sequence[str]] = None,
+        node_groups: List[List[Node]] = None,
+    ) -> None:
+        super().__init__(
```
digantdesai, line 76: Nit: Add a doc script

exir/backend/test/test_group_partitioner.py
```diff
@@ -0,0 +1,369 @@
...
+        def is_node_supported(self, submodules, node):
+            if node.op == "get_attr":
+                return True
+
+            if node.name in self.supported_nodes:
+                return True
+
+            return False
+
+    def find_nodes_by_names(
```
digantdesai:
```suggestion
    def _find_nodes_by_names(
```

exir/backend/test/test_group_partitioner.py
```diff
@@ -0,0 +1,369 @@
...
+            if not found:
+                not_found.append(name)
+
+        if not_found:
+            print(f"Warning: Could not find nodes matching: {not_found}")
+
+        return result
+
+    def create_model(self, model):
+        return model()
```
digantdesai:
```suggestion
        return model().eval()
```

exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -0,0 +1,272 @@
...
+            allows_single_node_partition=allows_single_node_partition,
+            non_compute_ops=non_compute_ops,
+            allowed_single_node_partition_ops=allowed_single_node_partition_ops,
+        )
+        self.dependency_viewer = _DependencyViewer(graph_module)
+        self.node_groups = (
+            [set(node_group) for node_group in node_groups] if node_groups else None
+        )
+        self.node_to_group = collections.defaultdict(int)
+        self.all_nodes_in_groups = set()
```
digantdesai, line 88: what if a node is in two groups? Should we check against that?

exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -0,0 +1,272 @@
...
+                        if downstream_node in combined_nodes:
+                            return False
+
+            return True
+
+        # Preprocess nodes to put them in same partition
+        if self.node_groups:
+            for i, group in enumerate(self.node_groups):
+                # Create a partition for each group
+                partition_id = next(new_partition_id)
```
digantdesai: What is the rationale for not allowing two node_groups in the same partition? Or can they get merged later, still reading...

exir/backend/test/test_group_partitioner.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+
+        partitions = partitioner.propose_partitions()
+
+        self.check_partition(
+            partitions, {"linear", "linear_1", "fake_quantize_per_tensor_affine"}
+        )
+        self.check_partition(
+            partitions, {"add", "linear_2", "fake_quantize_per_tensor_affine_1"}
+        )
```
digantdesai, line 384: I don't believe these tests are sufficient to cover the rather huge `propose_partitions` method. I propose, in addition to these, can we a lot more tests with `node_groups = None` and compare the partitioned output with CapabilityBasedPartitioner?

digantdesai:
>  For example, in a graph with shared QDQ (Quantize-Dequantize) chains, the partitioner may incorrectly separate nodes that should remain together

Still not fully grasped the motivation, can you explain this a bit more perhaps with an example? And why the existing greedy partitioner merging logic from the CapabilityBasedParitioner would split these nodes? Thanks.

digantdesai (commented):
Looks good, left some comments. Mainly around rationale and testing. I assume we will start using this in the XNNPACKPartitioner, if yes, that might help with testing.

leafs1:
> > For example, in a graph with shared QDQ (Quantize-Dequantize) chains, the partitioner may incorrectly separate nodes that should remain together
> 
> Still not fully grasped the motivation, can you explain this a bit more perhaps with an example? And why the existing greedy partitioner merging logic from the CapabilityBasedParitioner would split these nodes? Thanks.

Here's the initial github issue with an example https://github.com/pytorch/executorch/issues/11447. The idea is that the current greedy partitioner when dealing with a quantized graph can partition the dq/q nodes in different partitions (example partition 1 has q and partition 2 has the op and the dq). As a result, the semantics of the quantized op is lost in the graph.

mcr229 (approved):
@leafs1 thanks for all these extra tests michael. I do believe it is well tested. Just as a follow up let's test for faulty cases (enforced groups might not be valid (because of dependencies)). And let's also put up a PR wiring this up to xnnpack so we can see how this fairs. 

Not sure if this is added as well. But let's put some tests up where compatibility and group based partitioner differ in their outputted partition. Specifically in a quantization use case where a quantized node is moved out of its necessary dependency.

PR #12088, opened 2025-06-28
Add MergedDataMap to method

lucylq:
* __->__ #12088
* #12087

Differential Revision: [D77472917](https://our.internmc.facebook.com/intern/diff/D77472917/)


runtime/executor/method.cpp
```diff
@@ -400,7 +401,7 @@ Error Method::parse_external_constants(const NamedDataMap* named_data_map) {
   return Error::Ok;
 }
 
-Error Method::parse_values(const NamedDataMap* named_data_map) {
+Error Method::parse_values(const NamedDataMap* external_data_map) {
```
JacobSzwejbka, line 404: I feel like named was the better name since it possibly contains data embedded in the pte

runtime/executor/method.cpp
```diff
@@ -1680,6 +1694,10 @@ Method::~Method() {
   for (const auto i : c10::irange(n_external_constants_)) {
     external_constants_[i].buffer.~FreeableBuffer();
   }
+  // Free the MergedDataMap.
```
JacobSzwejbka, line 1697: I guess this is fine. It doesnt do anything though. The MergedMap is trivially destructable

JacobSzwejbka (approved):

runtime/executor/method.cpp
lucylq: oops, didn't see this. This data map comes from the PTD file (external), so it shouldn't contain items from the PTE  --> that's the pte_data_map

PR #12045, opened 2025-06-27
Detect release builds by comparing CMAKE_BUILD_TYPE to Debug, not Release

swolchok:
MinSizeRel and RelWithDebInfo are also release builds.

PR #12023, opened 2025-06-26
Specialize BroadcastIndexesRange for the case where there is only 1 contiguous input

swolchok:
In this case, broadcasting is not possible if I understand correctly.

NOTE TO REVIEWERS: I deleted a failing test because I think it's testing not-actually-existent-in-PyTorch functionality. Please let me know if I've made a mistake. I tried to exercise the behavior that this test implied existed like so:
```
>>> t = torch.tensor([1, 2, 3])
>>> t2 = torch.tensor(4)
>>> torch.abs(t2, out=t)
<stdin>:1: UserWarning: An output with one or more elements was resized since it had shape [3], which does not match the required output shape []. This behavior is deprecated, and in a future PyTorch release outputs will not be resized unless they have zero elements. You can explicitly reuse an out tensor t by resizing it, inplace, to zero elements with t.resize_(0). (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/native/Resize.cpp:38.)
tensor(4)
```

I think that if the test was correct, the result would have been torch.tensor([1, 2, 3]) with no message. Also, none of our operator tests seem to be failing. Have I missed anything?

PR #12089, opened 2025-06-28
Unbreak build_size_test.sh on Mac

swolchok:
By accidentally fixing the code in test/build_size_test.sh to correctly detect release builds (EQUAL is for numbers; you need STREQUAL for strings), #12045 broke that script on Mac because `--gc-sections` is not a valid linker flag there. This PR creates a general utility function for stripping dead code in linker flags and uses it everywhere we currently set `--gc-sections`.


swolchok:
* #12023
* __->__ #12089

larryliu0820 (approved):

mergennachin:
For low level build stuff, let's add `ciflow/trunk` labels so that we can catch them earlier

swolchok:
> For low level build stuff, let's add `ciflow/trunk` labels so that we can catch them earlier

I didn't realize this failure showed up on CI somewhere; I noticed because I was running the script locally. I thought that we only ran size tests on Linux.

PR #12096, opened 2025-06-28
Fix llava test target cmake

mergennachin:
Fix llava test target cmake

https://github.com/pytorch/executorch/pull/12089 had a typo


cccclai (approved):
Stamp

PR #12099, opened 2025-06-30
Parallelize optimized op_log_softmax

swolchok:
* __->__ #12099
* #12098

Straightforward application of parallel_for.

Differential Revision: [D76831122](https://our.internmc.facebook.com/intern/diff/D76831122/)


swolchok:
[pull / android / run-emulator](https://github.com/pytorch/executorch/actions/runs/15960928522/job/45019905704?pr=12099) is consistently telling us this is broken, but the bug is not obvious to me on rereading

swolchok:
after rereading the code twice, I was inclined to suspect that the kernel is somehow thread-unsafe, so I built with TSAN locally (-fsanitize=debug, CMAKE_BUILD_TYPE=RelWithDebInfo so parallel_for will actually parallelize) and didn't catch a failure. doesn't look like a pre-existing issue on HUD or anything so we shouldn't bypass, but I'm not sure what to do next. Nobody is particularly asking for parallelization AFAIK so there's no rush...

swolchok:
android / run-emulator is no longer complaining. I guess that means this is landable!

SS-JIA (approved):

swolchok:
mobilebert has been timing out on HUD, bypassing

PR #12105, opened 2025-06-30
Arm backend: Add acosh decomposition pass and test

emmakujala:
Decomposes acosh into TOSA-operations. 


oscarandersson8218 (approved):
Looks good to me!

PR #12106, opened 2025-06-30
Arm backend: Add U55 operator check for View

Sebastian-Larsson:
In order for the view operator to be compatible with the channels-last format of TosaBackend, transposes may need to be inserted before and after the view op. The transose operator comes with some constraints on U55. One such constraint is that the product of the axes should not exceed 65536. With this patch a check if implemented for the view operator to prevent such operators from being delegated.


zingo (approved):

zingo:
coreml fails unrelated

PR #12112, opened 2025-06-30
Updating documentation for cmake dtype selective build

BujSet:
### Summary
Updating documentation for cmake dtype selective build with model API. Table at `selective_build_model_doc/docs/source/kernel-library-selective-build.md` now looks like:

image

Fixes #12065

cc @mergennachin @byjlw

### DIFF
--- a/docs/source/kernel-library-selective-build.md
+++ b/docs/source/kernel-library-selective-build.md
@@ -49,6 +49,7 @@ gen_selected_ops(
 )
 ```
 
+The macro makes a call to gen_oplist.py, which requires a [distinct selection](https://github.com/BujSet/executorch/blob/main/codegen/tools/gen_oplist.py#L222-L228) of API choice. `OPS_SCHEMA_YAML`, `ROOT_OPS`, `INCLUDE_ALL_OPS`, and `OPS_FROM_MODEL` are mutually exclusive options, and should not be used in conjunction. 
 
 ### Select all ops
 
@@ -70,7 +71,7 @@ This API lets users pass in a pte file of an exported model. When used, the pte 
 
 ### Dtype Selective Build
 
-Beyond pruning the binary to remove unused operators, the binary size can furhter reduced by enforcing checks on what dtypes are used by the specific operators. For example, if your model only uses floats for the `add` operator, then including variants of the `add` operators for `doubles` and `ints` is unnecessary. The flag `DTYPE_SELECTIVE_BUILD` can be set to `ON` to support this additional optimization. Currently, dtype selective build is only support with the model API described above. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the `portable_kernels` lib. This feature is not yet supported for other libs.
+Beyond pruning the binary to remove unused operators, the binary size can further reduced by removing unused dtypes. For example, if your model only uses floats for the `add` operator, then including variants of the `add` operators for `doubles` and `ints` is unnecessary. The flag `DTYPE_SELECTIVE_BUILD` can be set to `ON` to support this additional optimization. Currently, dtype selective build is only supported with the model API described above. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the `portable_kernels` lib. This feature is only supported for the portable kernels library; it's not supported for optimized, quantized or custom kernel libraries.
 
 ## Example Walkthrough
 
--- a/docs/source/kernel-library-selective-build.md
+++ b/docs/source/kernel-library-selective-build.md
@@ -91,4 +91,3 @@ These options allow a user to tailor the cmake build process to utilize the diff
 |<code><br> cmake -D… -DSELECT_OPS_YAML=ON <br></code> | <code><br>  set(_custom_ops_yaml ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops.yaml) <br> gen_selected_ops("${_custom_ops_yaml}" "" "") <br></code> |
 |<code><br> cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out" <br></code> | <code><br> gen_selected_ops("" "" "" "${_model_path}" "") <br></code> |
 |<code><br> cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out" -DEXECUTORCH_DTYPE_SELECTIVE_BUILD=ON<br></code> | <code><br> gen_selected_ops("" "" "" "${_model_path}" "ON") <br></code> |
-
--- a/docs/source/kernel-library-selective-build.md
+++ b/docs/source/kernel-library-selective-build.md
@@ -66,13 +66,19 @@ This API lets users pass in a list of operator names. Note that this API can be 
 
 ### Select ops from model
 
-This API lets users pass in a pte file of an exported model. When enable, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes. Flag `DTYPE_SELECTIVE_BUILD` must be set to `ON` for the flow to use the generated YAML file. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the portable_kernels lib. This feature is not yet supported for other libs. 
+This API lets users pass in a pte file of an exported model. When used, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes used in the model. 
+
+### Dtype Selective Build
+
+Beyond pruning the binary to remove unused operators, the binary size can furhter reduced by enforcing checks on what dtypes are used by the specific operators. For example, if your model only uses floats for the `add` operator, then including variants of the `add` operators for `doubles` and `ints` is unnecessary. The flag `DTYPE_SELECTIVE_BUILD` can be set to `ON` to support this additional optimization. Currently, dtype selective build is only support with the model API described above. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the `portable_kernels` lib. This feature is not yet supported for other libs.
 
 ## Example Walkthrough
 
-In CMakeLists.txt we have the following logic:
+In [CMakeLists.txt](https://github.com/pytorch/executorch/blob/main/examples/selective_build/CMakeLists.txt#L89-L123) we have the following logic:
+
 ```cmake
 set(_kernel_lib)
+
 if(SELECT_ALL_OPS)
   gen_selected_ops("" "" "${SELECT_ALL_OPS}" "" "")
 elseif(SELECT_OPS_LIST)
@@ -80,9 +86,13 @@ elseif(SELECT_OPS_LIST)
 elseif(SELECT_OPS_YAML)
  set(_custom_ops_yaml ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops.yaml)
   gen_selected_ops("${_custom_ops_yaml}" "" "")
-elseif(SELECT_OPS_MODEL && DTYPE_SELECTIVE_BUILD)
+elseif(SELECT_OPS_MODEL)
  set(_model_path $(realpath model.pte))
-  gen_selected_ops("" "" "" "${_model_path}" "ON")
+  if(DTYPE_SELECTIVE_BUILD)
+    gen_selected_ops("" "" "" "${_model_path}" "ON")
+  else
+    gen_selected_ops("" "" "" "${_model_path}" "")
+  endif()
 endif()
 ```
 Then when calling CMake, we can do:
@@ -97,7 +107,14 @@ Or
 cmake -D… -DSELECT_OPS_YAML=ON
 ```
 
-To select from either an operator name list or a schema yaml from kernel library. To select operators and dtypes using the model API, we can run:
+Or 
+
+
+```
+cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out"
+```
+
+To select from either an operator name list, a schema yaml, or directly from an exported model's pte from kernel library. To further optimize the binary size for dtype selection, we can run:
 
 ```
 cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out" -DEXECUTORCH_DTYPE_SELECTIVE_BUILD=ON
--- a/docs/source/kernel-library-selective-build.md
+++ b/docs/source/kernel-library-selective-build.md
@@ -74,48 +74,20 @@ Beyond pruning the binary to remove unused operators, the binary size can furhte
 
 ## Example Walkthrough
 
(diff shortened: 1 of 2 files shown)


examples/selective_build/README.md
```diff
@@ -12,10 +12,11 @@ cd executorch
 bash examples/selective_build/test_selective_build.sh cmake
 ```
 
-Check out `CMakeLists.txt` for demo of 3 selective build APIs:
+Check out `CMakeLists.txt` for demo of 4 selective build APIs:
 1. `SELECT_ALL_OPS`: Select all ops from the dependency kernel libraries, register all of them into ExecuTorch runtime.
 2. `SELECT_OPS_LIST`: Only select operators from a list.
 3. `SELECT_OPS_YAML`: Only select operators from a yaml file.
+3. `SELECT_OPS_FROM_MODEL`: Only select operators from a from an exported model pte.
```
lucylq: Add the DTYPE_SELECTIVE_BUILD option here as well, and mention it only works with model api/portable_kernels

docs/source/kernel-library-selective-build.md
```diff
@@ -62,25 +64,31 @@ Context: each kernel library is designed to have a yaml file associated with it.
...
+elseif(SELECT_OPS_MODEL && DTYPE_SELECTIVE_BUILD)
+ set(_model_path $(realpath model.pte))
+  gen_selected_ops("" "" "" "${_model_path}" "ON")
 endif()
 ```
 Then when calling CMake, we can do:
 
 ```
-cmake -D… -DSELECT_OPS_LIST="aten::add.out,aten::mm.out”
+cmake -D… -DSELECT_OPS_LIST="aten::add.out,aten::mm.out"
```
lucylq: wow good find

docs/source/kernel-library-selective-build.md
```diff
@@ -62,25 +64,31 @@ Context: each kernel library is designed to have a yaml file associated with it.
 
 This API lets users pass in a list of operator names. Note that this API can be combined with the API above and we will create a allowlist from the union of both API inputs.
 
+### Select ops from model
+
+This API lets users pass in a pte file of an exported model. When enable, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes. Flag `DTYPE_SELECTIVE_BUILD` must be set to `ON` for the flow to use the generated YAML file. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the portable_kernels lib. This feature is not yet supported for other libs.
```
lucylq: We can de-couple the two features here, and have one section for the model API and one section for dtype selective build - a user can use the model API without dtype selective build and not really face any restrictions.

You can add the limitations on dtype selective build --> only available for portable (not optimized, quantized or custom kernel libraries) and only available for model API.

docs/source/kernel-library-selective-build.md
```diff
@@ -62,25 +64,31 @@ Context: each kernel library is designed to have a yaml file associated with it.
...
 if(SELECT_ALL_OPS)
-  gen_selected_ops("" "" "${SELECT_ALL_OPS}")
+  gen_selected_ops("" "" "${SELECT_ALL_OPS}" "" "")
 elseif(SELECT_OPS_LIST)
-  gen_selected_ops("" "${SELECT_OPS_LIST}" "")
+  gen_selected_ops("" "${SELECT_OPS_LIST}" "" "" "")
 elseif(SELECT_OPS_YAML)
  set(_custom_ops_yaml ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops.yaml)
   gen_selected_ops("${_custom_ops_yaml}" "" "")
+elseif(SELECT_OPS_MODEL && DTYPE_SELECTIVE_BUILD)
```
lucylq: update: guess we don't need this if we update the example to reflect current code

~nit: I don't think `&&` is valid cmake syntax - think it is `AND`. Also again to decouple, it might be good to indicate that we can have model api without dtype selective build:~
```
elseif (SELECT_OPS_MODEL) 
  if (DTYPE_SELECTIVE_BUILD) ...
  else ...
```

docs/source/kernel-library-selective-build.md
```diff
@@ -62,25 +64,31 @@ Context: each kernel library is designed to have a yaml file associated with it.
 
 This API lets users pass in a list of operator names. Note that this API can be combined with the API above and we will create a allowlist from the union of both API inputs.
 
+### Select ops from model
+
+This API lets users pass in a pte file of an exported model. When enable, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes. Flag `DTYPE_SELECTIVE_BUILD` must be set to `ON` for the flow to use the generated YAML file. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the portable_kernels lib. This feature is not yet supported for other libs. 
 
 ## Example Walkthrough
 
 In CMakeLists.txt we have the following logic:
```
lucylq: This looks outdated - couldn't find these in the selective_build CMakeLists.txt lol. Maybe worth updating it to what it currently is, and adding a link to the file? 

In [CMakeLists.txt](https://github.com/pytorch/executorch/blob/main/examples/selective_build/CMakeLists.txt#L89-L123) we have the following logic:
```
set(_kernel_lib)
if(EXECUTORCH_SELECT_OPS_YAML)
  set(_custom_ops_yaml
      ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops.yaml
  )
  set(kernel_sources
      ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops_1_out.cpp
      ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops_2_out.cpp
  )
  #
  # custom_kernels: C++ kernel implementations of custom ops
  #
  add_library(custom_kernels ${kernel_sources})
  target_link_libraries(custom_kernels PRIVATE executorch_core)
  target_compile_options(custom_kernels PUBLIC ${_common_compile_options})

  list(APPEND _kernel_lib custom_kernels)
else()
  list(APPEND _kernel_lib portable_kernels)
endif()

gen_selected_ops(
  LIB_NAME
  "select_build_lib"
  OPS_SCHEMA_YAML
  "${_custom_ops_yaml}"
  ROOT_OPS
  "${EXECUTORCH_SELECT_OPS_LIST}"
  INCLUDE_ALL_OPS
  "${EXECUTORCH_SELECT_ALL_OPS}"
  OPS_FROM_MODEL
  "${EXECUTORCH_SELECT_OPS_FROM_MODEL}"
  DTYPE_SELECTIVE_BUILD
  "${EXECUTORCH_DTYPE_SELECTIVE_BUILD}"
)
```

examples/selective_build/README.md
```diff
@@ -12,10 +12,11 @@ cd executorch
...
 
 Other configs:
 - `MAX_KERNEL_NUM=N`: Only allocate memory for N operators.
```
lucylq, line 23: nit: with your additions, we can delete the line below this: `We have one more API incoming: only select from an exported model file (.pte).` 

couldn't comment on the actual line for some reason..

docs/source/kernel-library-selective-build.md
BujSet: Bit unclear on what [this section](https://github.com/pytorch/executorch/pull/12112/files#diff-eaae6782e45ab51b669559d654f83a890d19d30d1098d9a4c92826566a54b8a2R80-R95) of the docs should look like. Previously, it seems like API choice was handled directly in the examples/selective_build/CMakeLists.txt file, passing different args to `gen_selected_ops` based on desired usage. This functionality is shifted to [tools/cmake/Codegen.cmake](https://github.com/pytorch/executorch/blob/main/tools/cmake/Codegen.cmake#L51-L59). Should the docs update to describe this (i.e. just remove the section about what the cmake logic and maybe put in a table to show how the args to `gen_selected_ops` change based on different input args?), or should remove this section entirely since this choice is no longer exposed to the user?

lucylq: That's a good question. As a user, I think it's useful to have something like an 'Example walkthrough' to refer to when using selective build in a custom CMakeLists.txt. From that perspective, it's probably better to have the cmake usage examples rather than the underlying logic from Codegen.cmake.

Re this PR, I think your previous changes are a clearer cmake example; and we add can something like 'refer to selective build CMakeLists.txt for more information', wdyt?

BujSet: Hm, not sure. If we want to keep some sort of information that helps explain how the flow changes with certain args, maybe we could have something like this in the Example Walkthrough section:

In [CMakeLists.txt](https://github.com/BujSet/executorch/blob/main/examples/selective_build/CMakeLists.txt#L48-L72), we have the following cmake config options:

1. EXECUTORCH_SELECT_OPS_YAML
2. EXECUTORCH_SELECT_OPS_LIST
3. EXECUTORCH_SELECT_ALL_OPS
4. EXECUTORCH_SELECT_OPS_FROM_MODEL
5. EXECUTORCH_DTYPE_SELECTIVE_BUILD

These options allow a user to tailor the cmake build process to utilize the different APIs, and results in different invocations on the `gen_selected_ops` [function](https://github.com/BujSet/executorch/blob/main/examples/selective_build/CMakeLists.txt#L110-L123). The following table describes some examples of how the invocation changes when these configs are set:

| Example cmake Call | Resultant `gen_selected_ops` Invocation |
| :----: | :---:| 
|<code><br>  cmake -D… -DSELECT_OPS_LIST="aten::add.out,aten::mm.out" <br></code> | <code><br>  gen_selected_ops("" "${SELECT_OPS_LIST}" "" "" "") <br></code> |
|<code><br> cmake -D… -DSELECT_OPS_YAML=ON <br></code> | <code><br>  set(_custom_ops_yaml ${EXECUTORCH_ROOT}/examples/portable/custom_ops/custom_ops.yaml) <br> gen_selected_ops("${_custom_ops_yaml}" "" "") <br></code> |
|<code><br> cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out" <br></code> | <code><br> gen_selected_ops("" "" "" "${_model_path}" "") <br></code> |
|<code><br> cmake -D… -DEXECUTORCH_SELECT_OPS_FROM_MODEL="model.pte.out" -DEXECUTORCH_DTYPE_SELECTIVE_BUILD=ON<br></code> | <code><br> gen_selected_ops("" "" "" "${_model_path}" "ON") <br></code> |

What do you think?

lucylq: Yeah I like it! This is nice, thanks Bujji!

BujSet:
cc @psiddh

docs/source/kernel-library-selective-build.md
```diff
@@ -36,14 +36,16 @@ The basic flow looks like this:
...
 
 ```
 gen_selected_ops(
-  LIB_NAME         # the name of the selective build operator library to be generated
-  OPS_SCHEMA_YAML  # path to a yaml file containing operators to be selected
-  ROOT_OPS         # comma separated operator names to be selected
-  INCLUDE_ALL_OPS  # boolean flag to include all operators
+  LIB_NAME              # the name of the selective build operator library to be generated
+  OPS_SCHEMA_YAML       # path to a yaml file containing operators to be selected
+  ROOT_OPS              # comma separated operator names to be selected
```
JacobSzwejbka, line 45: Root_ops, include_all_ops, and ops_from_model are intended to be exclusive right? Only 1 of those should be used at a time? Can we make that relationship clearer in the documentation.

docs/source/kernel-library-selective-build.md
```diff
@@ -62,31 +64,30 @@ Context: each kernel library is designed to have a yaml file associated with it.
 
 This API lets users pass in a list of operator names. Note that this API can be combined with the API above and we will create a allowlist from the union of both API inputs.
 
+### Select ops from model
+
+This API lets users pass in a pte file of an exported model. When used, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes used in the model. 
+
+### Dtype Selective Build
+
+Beyond pruning the binary to remove unused operators, the binary size can furhter reduced by enforcing checks on what dtypes are used by the specific operators. For example, if your model only uses floats for the `add` operator, then including variants of the `add` operators for `doubles` and `ints` is unnecessary. The flag `DTYPE_SELECTIVE_BUILD` can be set to `ON` to support this additional optimization. Currently, dtype selective build is only support with the model API described above. Once enabled, a header file that specifies only the operators and dtypes used by the model is created and linked against a rebuild of the `portable_kernels` lib. This feature is not yet supported for other libs.
```
JacobSzwejbka: furhter -> further

docs/source/kernel-library-selective-build.md
```diff
@@ -62,31 +64,30 @@ Context: each kernel library is designed to have a yaml file associated with it.
 
 This API lets users pass in a list of operator names. Note that this API can be combined with the API above and we will create a allowlist from the union of both API inputs.
 
+### Select ops from model
+
+This API lets users pass in a pte file of an exported model. When used, the pte file will be parsed to generate a yaml file that enumerates the operators and dtypes used in the model.
```
JacobSzwejbka, line 70: Is the fact that a yaml file is generated relevant to the end user? The above documentation makes it seem like the model yaml is not a user facing construct, only the pte, or the root ops list are exposed methods of passing selective information. 

Also in general I would say "root" ops is misleading since ET doesnt have transitive ops. We should just call it the op list.

docs/source/kernel-library-selective-build.md
lucylq: few nits: 

`the binary size can furhter reduced by enforcing checks on what dtypes are used by the specific operators` -->
`the binary size can be further reduced by removing unused dtypes` (feel like 'enforcing checks' can be a bit confusing - assume you are referring to the ET_CHECKs?)

`only support with the model API described above` --> support -> supported

`This feature is not yet supported for other libs.` --> 
`This feature is only supported for the portable kernels library; it's not supported for optimized, quantized or custom kernel libraries.`

docs/source/kernel-library-selective-build.md
BujSet: When using the model API, the generated YAML can be as an input to the selective build process, like when using the yaml API directly. The difference being that a user doesn't have to manually craft a yaml, but can pass in a pte.

I agree, root ops is not a great name, but it seems to be used in other places as well, e.g. https://github.com/BujSet/executorch/blob/main/codegen/tools/gen_oplist.py#L234-L243. For consistency, I think we should keep it as root ops, and a separate PR should refactor this to something more meaningful, what do you think?

lucylq (approved):

mergennachin:
@lucylq @BujSet @larryliu0820 should this part of the cherry-pick for 0.7?

lucylq:
> @lucylq @BujSet @larryliu0820 should this part of the cherry-pick for 0.7?

Yes, I think we can include it as all the code changes are there. @BujSet could you cherry-pick it following the instructions in https://github.com/pytorch/executorch/issues/11075 ?

PR #12125, opened 2025-07-01
Use safe_numerics util from PyTorch

### LINKED ISSUE #11370: Import and extend c10/util/safe_numerics.h

aaron-ang:
### Summary
Reopening from #11537
Fixes #11370

### Test plan
Use existing tests since we are doing a drop-in replacement for overflow checks.

### DIFF
--- a/runtime/core/portable_type/c10/c10/targets.bzl
+++ b/runtime/core/portable_type/c10/c10/targets.bzl
@@ -112,7 +112,9 @@ def define_common_targets():
                 "util/complex_utils.h",
                 "util/floating_point_utils.h",
                 "util/irange.h",
+                "util/llvmMathExtras.h",
                 "util/overflows.h",
+                "util/safe_numerics.h",
             ],
             exported_preprocessor_flags = [
                 "-DC10_USING_CUSTOM_GENERATED_MACROS",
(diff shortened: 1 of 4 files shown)


aaron-ang:
hi @JacobSzwejbka, I can't seem to figure out why the import `#include <executorch/runtime/core/portable_type/c10/c10/util/safe_numerics.h>` is not recognized..

Edit: I've rebased with `main` and added the file paths to `targets.bzl`. Hope this fixes the runtime errors.

swolchok (approved):
looks good as long as tests pass, just approved them to run

aaron-ang:
@swolchok can you try running the CI again?

swolchok:
CI looks good, can you rebase and fix the conflict?

aaron-ang:
@swolchok rebased

swolchok:
unittest linux failures look ignorable

PR #12126, opened 2025-07-01
add  custom annoatation for new model export

billmguo:
Summary: add  custom annoatation for new model export

Differential Revision: D77569950

### DIFF
--- a/backends/qualcomm/quantizer/custom_annotation.py
+++ b/backends/qualcomm/quantizer/custom_annotation.py
@@ -227,6 +227,40 @@ def annotate_matmul_16a8w(gm: torch.fx.GraphModule) -> None:  # noqa: C901
             _annotated=True,
         )
 
+    def annotate_single_in_share_out(
+        node: Node, quantization_config: QuantizationConfig
+    ) -> None:
+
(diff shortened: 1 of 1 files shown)


limintang (approved):

backends/qualcomm/_passes/lift_constant_scalar_operands.py
```diff
@@ -124,6 +124,10 @@ def _create_tensor_args(
     ) -> Dict[int, TensorConstant]:
         tensor_args = {}
         for i, arg in enumerate(node.args):
+            if hasattr(node.target, "_schema"):
```
winskuo-quic: Curious under what situation a node has no schema.

winskuo-quic (approved):
LGTM. Thanks

backends/qualcomm/_passes/lift_constant_scalar_operands.py
billmguo: I will remove this lines, this issue happen when I apply get_quant_embedding_transform for the model embedding layers and export model with -E "4,32"

PR #12127, opened 2025-07-01
Arm backend: Add mean.dim to CheckProperQuantization

Sebastian-Larsson:
mean.dim decomposes into an avg_pool2d, which occurs after quantization annotation. This led to the avg_pool2d node not having quantization parameters for the output. mean.dim is added to CheckProperQuantization so that if the decomposition occurs the avg_pool2d operator will not be delegated.


zingo (approved):

zingo (commented):
Got a fail on 
backends/arm/test/models/test_nn_functional.py::test_nn_functional_BI[cosine_similarity]
could it be random?

zingo (changes requested):
Try to retrigger to get rid of arm fail?

zingo:
I retriggered fail arm job.

zingo (approved):

PR #12128, opened 2025-07-01
Arm backend: Measure and show time per model during testing

zingo:
This will add time logs during testing so it will be easier to track what part of the flow take time and for what models.


oscarandersson8218 (approved):
CI failures unrelated. LGTM!

PR #12130, opened 2025-07-01
New PAL for Arm Zephyr RTOS

BujSet:
### Summary
A new runtime platform file to support applications form Arm Zephyr RTOS.


BujSet:
Turns out this is likely unnecessary as Zephyr supports POSIX.

PR #12135, opened 2025-07-01
API to Read Model and Generate Header

BujSet:
### Summary
TODO


BujSet:
This would make more sense as a PR to the Zephyr repo, not the executorch repo.

PR #12140, opened 2025-07-01
update building from source doc to make the backend support guidance clearer

Gasoonjia:
Summary: as title

Differential Revision: D77609951


larryliu0820 (approved):
Thank you for the update!

PR #12145, opened 2025-07-01
Delete opt_mul_scalar_out

swolchok:
The handwritten optimized code is similar to what we should be getting from the optimized portable op, as follows.

handwritten optimized code:
- if the input type matches the output type, then perform a vectorized loop
- otherwise, generate specific mixed-dtype kernels, which aren't vectorized.

optimized portable op:
- if the input type matches the output type, then perform a vectorized loop. (dtype_specialized_elementwise_fn_impl in elementwise_util.h)
- otherwise, generate one specific kernel per compute type. those
  kernels use non-inlined function calls to do loads and stores,
  trading off performance for a significant size reduction. (apply_elementwise_fn_generic_impl in elementwise_util.h)

Both cases in the portable op variant also use parallel_for.

I attempted to do a performance test, but I found that
`torch.mul(some_tensor, 2.0)` is exported as a call to mul.Tensor,
*not*
mul.Scalar. https://github.com/pytorch/executorch/commit/41e7ffa8b7ff09206aa5b9b5c1bbd82b9e0ff277
added the ability to pass our tests if we do emit mul.Scalar for this,
but the follow-up diff to make that happen seems not to have
landed. So, I think another reason to delete this is that (if I
understand correctly) it's not used, therefore we don't have specific
knowledge that we need it to exist and we can't just use the optimized
portable op.

### DIFF
--- a/kernels/optimized/cpu/op_mul.cpp
+++ b/kernels/optimized/cpu/op_mul.cpp
@@ -210,63 +210,6 @@ Tensor& opt_mul_out(
   return out;
 }
 
-Tensor& opt_mul_scalar_out(
-    KernelRuntimeContext& ctx,
-    const Tensor& a,
-    const Scalar& b,
-    Tensor& out) {
-  (void)ctx;
-
-  ScalarType a_type = a.scalar_type();
-  ScalarType common_type =
-      utils::promote_type_with_scalar(a_type, b, /*half_to_float*/ false);
-  ScalarType out_type = out.scalar_type();
-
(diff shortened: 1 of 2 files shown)


swolchok:
* __->__ #12145

kimishpatel (approved):
I trust that the summary holds true. I think the question of mult-threading remains in that it is not always beneficial to multithread. Our portable implementation should account for that else for some case, I wont be surprised if I saw pref regression

PR #12146, opened 2025-07-01
[not for merge] Add armv7 support in buck shim

digantdesai:


mrnhtyzgld:
Will there be a armeabi-v7a support for android executorch?

PR #12149, opened 2025-07-01
Remove _skip_type_promotion config

swolchok:
I don't think we need this any more, assuming tests pass.

It starts with an underscore, so it's private, so we can just remove it, right?


swolchok:
* __->__ #12149

larryliu0820 (approved):

PR #12151, opened 2025-07-01
Set Weight Loading to be strict

cccclai:
Summary: It often causes confusion and weird downstream error when the weight is not properly loaded. Set `load_state_dict` to be True so the error populated early.

Differential Revision: D77626023


shewu-quic (approved):
Thanks for your effort.

kimishpatel (approved):

PR #12158, opened 2025-07-02
Added Pybindings for Method.h/cpp

Conarnar:
Summary:
Continuing with migrating the pybindings API to expose the lower level ET API rather than just the `Module` level. The next step of that is to add pybindings for `Method`.

Bindings for the class Method and its methods `set_inputs`, `execute, get_outputs`, `get_attribute`, and `method_meta` were added along with `call` and `__call__` in order to easily call `set_inputs`, `execute`, and `get_outputs` all at once.

The `method_meta` method for `Program` was also added.

The `MethodMeta` binding was modified to include a shared reference to `Program` instead of `Module` if the `Module` API isn't used.

The `Program` and `Method` classes in `__init__.py` were modified to hold their pybinding equivalent rather than Module.

Differential Revision: D77565018


lucylq (approved):
Looks good, thanks for fixing the memory issues and the thorough testing!

PR #12160, opened 2025-07-02
Arm backend: Fix bug in decompose_linear_pass

oscarandersson8218:
The output of the decomposition did not get any metadata which is troublesome when you have two consecutive linear layers.


zingo (approved):

zingo:
fails unrelated, fails on main also.

PR #12162, opened 2025-07-02
Arm backend: Split executor runner into init + run

martinlsm:
Refactor the main function of arm_executor_runner.cpp by extracting code that is related to initializing and running a model into two separate helper functions (runner_init and RunnerContext::run). A new struct called RunnerContext is introduced to store data/context required to run the model.

### Test plan
The code is tested in the CI. It's a refactoring patch, so no new functionality added/removed.

### DIFF
--- a/examples/arm/executor_runner/arm_executor_runner.cpp
+++ b/examples/arm/executor_runner/arm_executor_runner.cpp
@@ -224,6 +224,65 @@ void et_pal_free(ET_UNUSED void* ptr) {}
 
 namespace {
 
+/// Lightweight heapless container that constructs and stores a T in-place.
+template <typename T>
+class Box {
+ public:
+  Box() = default;
+
+  ~Box() {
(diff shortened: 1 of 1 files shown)


zingo:
Fails are unrelated and fail on main also

zingo (approved):

zingo:
What errors/problems did you see @metascroy ?

examples/arm/executor_runner/arm_executor_runner.cpp
```diff
@@ -224,6 +224,65 @@ void et_pal_free(ET_UNUSED void* ptr) {}
 
 namespace {
 
+/// Lightweight heapless container that constructs and stores a T in-place.
```
swolchok, line 227: why not use std::optional? this appears roughly equivalent, but I read it quickly and might've missed something

PR #11357, opened 2025-06-04
Qualcomm AI Engine Direct - CI for Non-LLM GA model

winskuo-quic:
### Summary
Part1 of CI for Non LLM GA models.

PR #12166, opened 2025-07-02
Revert "Qualcomm AI Engine Direct - CI for Non-LLM GA model"

metascroy:
Reverts pytorch/executorch#11357

This job is causing CI to fail because it takes too long and preventing viable/strict from updating.  @winskuo-quic please re-land with higher timeout.


cccclai (approved):

cccclai:
Checked with @metascroy, 

> I think the operation gets canceled because it hits the timeout, which counts as a failure for the purpose of updating viabl/strict

@winskuo-quic maybe let's keep the original timeout config and see if it passes

PR #12167, opened 2025-07-02
Fixed bug for 16a4w ptq

rohansjoshi:
Summary: Currently running the script executorch/examples/models/llama/export_llama.py with the flag --ptq 16a4w, it does 16a16w quantization; this diff fixes this. This may be related to some GitHub issues

Differential Revision: D77671468


cccclai (approved):
Thanks for the fix! @haowhsu-quic @winskuo-quic @chunit-quic fyi for the fix

mergennachin (approved):

PR #12170, opened 2025-07-02
Deprecate tag qdq pass in xnnbackend

### LINKED ISSUE #11588: Deprecate tag implicit qdq pass

https://github.com/pytorch/executorch/blob/5a45132d58947b2029876d2ce9f9584e4b699af9/backends/xnnpack/_passes/tag_implicit_q_dq_pass.py

This is our pass to tag implicit q/dq. Essentially quantized patterns look like such:
```
input(fp32) --> quantize(int8) -->dequantize(f32) --> operator --> quantize(int8) -->dequantize(f32) --> output
```

this represents a quantized op because you can view it as:
```
                                                    QUANTIZED OPERATOR
                                 ___________________________________________________
                                |                                                   |
input(fp32) --> quantize(int8) -|->dequantize(f32) --> operator --> quantize(int8) -|->dequantize(f32) --> output
                                |___________________________________________________|
```

As you can see from the box above, this represents a quantized operator because the input is int8 and the output is int8. Now, the pass above helps us differentiate the quantize/dequantize nodes within the box (are implicit and only used to indicate that the operator is quantized), and the quantize/dequantize nodes outside of the box( these nodes are actually run to quantize and dequantize its inputs.

However, with our new partitioner, when determining a group of nodes is supported, we add in the implicit quantize/dequantize nodes around it so from the partition we can laready tell that they are implicit:
https://github.com/pytorch/executorch/blob/5a45132d58947b2029876d2ce9f9584e4b699af9/backends/xnnpack/partition/config/generic_node_configs.py#L41-L76

Notice when we are getting the dependencies of oeprator, we also grab the implicit quantize/dequantize nodes around it(ignore the fused section for now). Then we can actually tag the implicit nodes at this stage when partitioning so we don't have to run this pass later on.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

abhinaykukkadapu:
Summary:
This diff decentralizes the q/dq implicit node tagging to individual partition configs instead of tagging it as part of backend pass. 

Changes in this diff:
1. Deprecate tag q dq pass
2. Remove all the places where this pass is used in the backend preprocess phase.
3. Decentralize the tagging to individual configs
    a. `generic_node_configs` will handle most of the non gemm nodes
    b. `gemm_configs` will handle gemm nodes
    c. channels last pass will add (copy q dq) or (dq copy q), tag the relevant nodes.
    d. tag q dq in conv1d unsqueeze pass.
    e. Tag q dq in compose cat
4. Deprecate configs.py where all the collection of nodes is maintained

Fixes: #11588

Reviewed By: mcr229

Differential Revision: D77055623


mcr229 (approved):

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -170,15 +190,19 @@ def insert_dq_copy_q(
...
             q = self.create_call_function_node(
                 graph_module=graph_module,
                 target=exir_ops.edge.quantized_decomposed.quantize_per_tensor.default,
                 args=(copy,) + q_params,
             )
-            q.meta = copy.meta
+            q.meta = copy.meta.copy()
+
+            # Always tag q/dq as implicit
+            tag_as_implicit_q_dq(dq)
```
digantdesai, line 204: I was talking with @mcr229 about this, this can be brittle i.e. implicit tagging for q/dq in >1 places. Do we want to add a pass in the future to verify if the tagging is correct or not. Just for validation? Feel free to create an issue, don't have to add it right away.

PR #12171, opened 2025-07-02
Revert "Arm backend: Split executor runner into init + run"

metascroy:
Reverts pytorch/executorch#12162

This fails internal CI.  @digantdesai can you have a look

### DIFF
--- a/examples/arm/executor_runner/arm_executor_runner.cpp
+++ b/examples/arm/executor_runner/arm_executor_runner.cpp
@@ -224,65 +224,6 @@ void et_pal_free(ET_UNUSED void* ptr) {}
 
 namespace {
 
-/// Lightweight heapless container that constructs and stores a T in-place.
-template <typename T>
-class Box {
- public:
-  Box() = default;
-
-  ~Box() {
-    if (has_value) {
-      ptr()->~T();
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


JacobSzwejbka (approved):

guangy10 (approved):

zingo:
What errors/problems did you see @metascroy ?

martinlsm:
I have now created a new PR (#12197) with this same patch. Would be good to know what the error was. Hopefully it was just a conflict with another patch when merging(?)

digantdesai:
@metascroy I saw [D77657568](https://www.internalfb.com/diff/D77657568) and the failure doesn't seem be related to this file/PR. Do you know why this had to be reverted? Also we relanded this in #12197 - I don't think there was any change, just curious.

metascroy:
> @metascroy I saw [D77657568](https://www.internalfb.com/diff/D77657568) and the failure doesn't seem be related to this file/PR. Do you know why this had to be reverted? Also we relanded this in #12197 - I don't think there was any change, just curious.

I reverted it because the failing internal tests were blocking the difftrain / codesync.  Reverting is a preferred way to unblock the difftrain based on our oncall runbook, but in this case it looks like the failures were false positive / transient.

PR #12174, opened 2025-07-02
Supporting Zephyr with Executorch as a Module

BujSet:
### Summary
This added directory structure is recommended to allow Executorch to formally be built as a module for a Zephyr project. Please see the add `zephyr/README.md` for more details.


larryliu0820:
cc @mergennachin @digantdesai for this. I feel like this should be fine, we already have a lot top level directories anyways

larryliu0820 (approved):

digantdesai:
> cc @mergennachin @digantdesai for this. I feel like this should be fine, we already have a lot top level directories anyways

"The module repository codebase shall include a module.yml file in a zephyr/ folder at the root of the repository." Seems like we don't have a choice. If we find one I would prefer hiding this from the top level, but no hard objection TBH.

PR #11975, opened 2025-06-25
Respect selective build for dtype_specialized_elementwise_fn_impl in elementwise_util

swolchok:
This fancy fast path I added didn't respect selective build. Now it should.

PR #12183, opened 2025-07-02
Fix dtype selective build check in elementwise_util

swolchok:
I think I got this wrong in #11975 -- the compute_dtype is the one type that is NOT related to selective build.

### DIFF
--- a/kernels/portable/cpu/util/elementwise_util.h
+++ b/kernels/portable/cpu/util/elementwise_util.h
@@ -319,12 +319,13 @@ inline void apply_elementwise_fn(
   }
 
   constexpr auto compute_type = CppTypeToScalarType<CTYPE_COMPUTE>::value;
-  if constexpr (should_include_kernel_dtype(op_name, compute_type)) {
+  constexpr ScalarType out_specialized_scalar_type =
+      specialized_output_scalar_type<CTYPE_COMPUTE>(out_dtypes);
+  if constexpr (should_include_kernel_dtype(
+                    op_name, out_specialized_scalar_type)) {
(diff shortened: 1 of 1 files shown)


swolchok:
* #12219
* #12217
* __->__ #12183

swolchok:
+ Kimish and Lucy because we just discussed this in person

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -319,12 +319,13 @@ inline void apply_elementwise_fn(
   }
 
   constexpr auto compute_type = CppTypeToScalarType<CTYPE_COMPUTE>::value;
-  if constexpr (should_include_kernel_dtype(op_name, compute_type)) {
+  constexpr ScalarType out_specialized_scalar_type =
+      specialized_output_scalar_type<CTYPE_COMPUTE>(out_dtypes);
+  if constexpr (should_include_kernel_dtype(
+                    op_name, out_specialized_scalar_type)) {
```
kimishpatel, line 325: use of `should_include_kernel_dtype` is behind a macro, should we not continue with that convention?

swolchok: ET_INTERNAL_CHECK_SELECTIVE_BUILD depends implicitly on et_switch_name being defined, which doesn't make sense here, and `should_include_kernel_dtype` is a perfectly good constexpr function so I don't see why we shouldn't call it

lucylq (approved):

swolchok:
unittest-editable failure looks like a known flake. merging.

PR #12195, opened 2025-07-03
Arm backend: Introduce TOSA backend dialect

per:
### Summary
Adds a TOSA backend dialect so it's possible to convert Edge IR to specific TOSA operators in the Arm backends. This is done to enable control of types and additional arguments available for TOSA operators in comparison with the Edge IR.
The operators are registered into the exir.backend.tosa namespace and is only traceable, not executable since it's only used in the lowering step to a TOSA serialization format.

### Test plan
Existing CI tests


zingo (approved):

zingo:
Test fails unrelated and fails on main and in other PRS

PR #12196, opened 2025-07-03
[ET-VK][ez] Fix partitioner logic

SS-JIA:
Summary:
## Changes

* In partitioner, check if buffer storage can be used only after it has been determined that no valid texture layouts are available

## Context

Currently, the logic in the vulkan partitioner is incorrect.

1. First, it checks what texture layouts may be used to represent the tensors involved in the computation
2. If no texture layouts are available, it checks if buffer support is available and the tensors are small enough to be within Vulkan buffer limits
3. Then, it checks if all valid texture layouts are supported by the op.

This introduces a bug in situations where 3 fails, but 2 would pass. However, 2 is not checked due to the way the logic is structured.

The fix is to switch the order of 2 and 3.

Test Plan:
## Test Plan

Manual verification + CI

### DIFF
--- a/backends/vulkan/partitioner/vulkan_partitioner.py
+++ b/backends/vulkan/partitioner/vulkan_partitioner.py
@@ -116,9 +116,17 @@ class VulkanSupportedOperators(OperatorSupportBase):
                     arg, self.buffer_limit
                 )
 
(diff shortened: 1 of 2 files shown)


SS-JIA:
* __->__ #12196

junpi3 (approved):

PR #12217, opened 2025-07-03
Remove unused compute_fun argument to validate_elementwise_fn_inputs

swolchok:

PR #12219, opened 2025-07-03
WIP: outline mixed dtype kernel

swolchok:


swolchok:
* __->__ #12219
* #12217
* #12183

swolchok:
this is not working in local tests on my Mac, but I think the linker there might not do identical code folding, so I sent this out for a double-check with the size test jobs

swolchok:
looks like this is also regressing size on linux, giving up

PR #12220, opened 2025-07-03
Milestone2.1: Partition to_dim_order_copy op in XNN delegate

leafs1:
### Summary
This PR adds support for the `to_dim_order_copy` operation in the XNNPACK delegate partitioner, enabling direct handling of memory format conversions initiated by users via `.to(memory_format=)` calls. This enhancement significantly improves performance by producing more compressed graphs that avoid unnecessary partitioning boundaries at memory format conversion points. By delegating these operations directly to XNNPACK, we eliminate the overhead of context switching between the runtime and delegate, reducing both execution time and memory footprint. The implementation leverages XNNPACK's highly optimized memory format conversion routines, which are specifically designed for efficient tensor layout transformations on various hardware targets.

### Test plan
Confirmed expected output when having user specified dim order conversions as well as appropriate partitioning. I did this by writing individual tests for the to_copy op ensuring it changes dim order and dtype when appropriate. Also added test module to confirm that the to copy nodes are partitioned and not in another partition


backends/xnnpack/partition/config/__init__.py
```diff
@@ -102,6 +103,8 @@
     ReciprocalSquareRootConfig,
     ReLUConfig,
     TanhConfig,
+    ToDimOrderCopyConfig,
+    # SDPAConfig, TODO: D60553559: preserving SDPA for fairseq fails
```
digantdesai: why add this comment here? Bad rebase?

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -419,6 +419,11 @@ def call(self, graph_module: torch.fx.GraphModule):  # noqa: C901
                 # The node requires nchw inputs
                 for input_node in node.all_input_nodes:
                     self.input_to_nchw(graph_module, input_node, node)
+            elif node.target == exir_ops.edge.aten._to_copy.default:
```
digantdesai: so the reason we still have `to_copy` even after we partition `to_dim_order_copy` is because we revert it back to `to_copy`. So when we add a node visitor next for the `to_dim_order_copy` we should remove the revert pass.

backends/xnnpack/test/ops/test_to_copy.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: As I said earlier, using `to_copy` is OK but we can just as easily move to `to_dim_order_copy` and remove the dim_order ops revert pass.

backends/xnnpack/test/ops/test_to_copy.py
```diff
@@ -0,0 +1,85 @@
...
+class TestChannelsLastTaggedReshapePass(unittest.TestCase):
+    def setUp(self):
+        torch._dynamo.reset()
+
+    def run_tester(self, module, inputs):
+        tester = Tester(
+            module.eval(),
+            inputs,
+        )
+        tester.export().to_edge_transform_and_lower().to_executorch().serialize().run_method_and_compare_outputs()
```
digantdesai: we should add `check_not` to assert the partitioner is working OK.

backends/xnnpack/test/passes/test_channels_last_tagged_reshape.py
```diff
@@ -179,6 +179,23 @@ def test_fp32_channels_last_tagged_reshape_pass(self):
...
+        def forward(self, x):
+            y = self.linear1(x)
+            y = y.to(memory_format=torch.channels_last)
+            y = y.to(memory_format=torch.contiguous_format)
+            return self.conv1(y)
+
+    LinearConvDimSwapModule = LinearConvDimSwap()
+
+    def test_conv_linear_dim_order_swap_partitioner(self):
+        self.run_tester(self.LinearConvDimSwapModule, (torch.randn(1, 3, 6, 4),))
```
digantdesai, line 199: we should add check_not to assert the partitioner is working OK, since we are running both linear and conv in contiguous, it might just work even when you don't partition the two ops.

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
leafs1: I see, should I make those changes in a follow up pr or would it be better to keep them here?

backends/xnnpack/partition/config/generic_node_configs.py
```diff
@@ -425,6 +425,35 @@ def supported_precision_types(self) -> List[ConfigPrecisionType]:
...
+            why(
+                node,
+                reason=f"dtype conversion from {input_dtype} to {output_dtype} is not supported",
+            )
+            return False
+
+        return True
+
+    def supported_precision_types(self) -> List[ConfigPrecisionType]:
+        return [ConfigPrecisionType.FP32]
```
mcr229: add ConfigPrecisionType.STATIC_QUANT

mcr229:
lets wait for CI, but looking good!

mcr229 (approved):

PR #12234, opened 2025-07-04
Qualcomm AI Engine Direct - GA model enablement (T5)

DannyYuyang-quic:
### Summary
 - e2e script / test case for GA [T5](https://huggingface.co/google-t5/t5-small) model
    - perf: 16a8w avg encoding time: 4.09ms/inf, avg decoding time: 6ms/inf (SM8750)
    - acc: F1 Score ~= 76% in [SQuAD](https://www.kaggle.com/datasets/akashdesarda/squad-v11?select=SQuAD-v1.1.csv)
 - add QA dataset for Seq2SeqLM benchmarking

### Test plan
```
python -m examples.qualcomm.oss_scripts.t5.t5 -b build-android -m ${soc} -H ${host_id} -s ${device_id} -d ./SQuAD-v1.1.csv
```
cc: @haowhsu-quic,@cccclai


examples/qualcomm/oss_scripts/t5/t5_model.py
```diff
@@ -0,0 +1,629 @@
...
+    StaticCache,
+)
+from transformers.modeling_outputs import BaseModelOutputWithPastAndCrossAttentions
+from transformers.models.t5.modeling_t5 import T5Attention, T5Stack
+from transformers.utils import is_torchdynamo_compiling, logging
+
+logger = logging.get_logger(__name__)
+
+
+# Copy from transformers/models/t5/modeling_t5.py (transformers=4.47.1)
```
cccclai, line 25: How different it is compared with transformers/models/t5/modeling_t5.py?

cccclai:
Also can you rebase?

examples/qualcomm/oss_scripts/t5/t5_model.py
DannyYuyang-quic: There are two changes compared to `transformers/models/t5/modeling_t5.py`:
both added to move the computation of relative position out of runtime and into precomputed buffers. This is because `T5Attention._relative_position_bucket` is not QNN-friendly.

examples/qualcomm/oss_scripts/t5/t5_model.py
```diff
@@ -0,0 +1,632 @@
...
+    def __init__(
+        self,
+        config,
+        embed_tokens=None,
+        max_hidden_seq_length=4096,
+        max_cache_length=1024,
+    ):
+        super().__init__(config, embed_tokens)
+
+        # ====================Qualcomm Changed=================================
```
DannyYuyang-quic, line 36: The first part of the added is that I precompute the relative_position_bucket using `T5Attention._relative_position_bucket` and register the result as a buffer.

examples/qualcomm/oss_scripts/t5/t5_model.py
```diff
@@ -0,0 +1,632 @@
...
+        if self.is_decoder and encoder_hidden_states is not None:
+            encoder_decoder_position_bias = torch.zeros(
+                (1, self.config.num_heads, seq_length, self.max_hidden_seq_length),
+                dtype=encoder_extended_attention_mask.dtype,
+            )
+            encoder_decoder_position_bias = (
+                encoder_decoder_position_bias
+                + encoder_extended_attention_mask[:, :, :, : self.max_hidden_seq_length]
+            )
+        # ========================================================================
```
DannyYuyang-quic, line 277: The second part is in the `forward`, where I retrieve the `relative_position_bucket` by indexing into the buffer using the correct `cache position`.

cccclai: I see, thanks for optimizing the performance. Can the source model definition also lowered?

DannyYuyang-quic: No, I still need to write some source transforms or add passes. This is because the function `_relative_position_bucket` from source T5 computation has two main issues:

1. Unsupported ops for (int64/int32) datatypes:
    Source function performs operations like abs, min, and neg on int64, but these ops are not supported on int64/int32 in QNN.
2. Unsupported casting:
    There is a cast from float32 to int64 in source function, but in the 16a8w quantization case, QNN's cast op actually performs a cast from uint16 to int32, which is also unsupported.

cccclai (approved):
Thanks! The diff train is kind of stuck, will try to merge it soon

examples/qualcomm/oss_scripts/t5/t5_model.py
cccclai: Taking a step back from being performant, if user tried and it fails here, and if it's not supported in QNN, should this op fall back to cpu?

DannyYuyang-quic: Yes, if it's not supported in QNN, this should fall back to cpu.

cccclai:
diff train is recovered but seems like still need to rebase...

cccclai (approved):
Thank you!

Issue #10603, opened 2025-05-01
Update quantization overview doc page


GregoryComer:
### 📚 The doc issue

The current quantization overview page is a bit sparse: https://pytorch.org/executorch/main/quantization-overview.html. I'd like to update it as follows:

- Move under Usage/ since it's the only page under Quantization/ currently.
- Split out information intended for backend authors (info about writing a quantizer, for example). Focus on user-facing APIs.
- Document backend-invariant quantization flows (embeddings, ao kernels, etc.). Include info (and example) on composable quantizer.
- Document PT2E and quantize_ flows.
- Cover the general, high level approach to quantizing different types of models.
  - CV models
  - Transformers / language models
- Talk briefly about options for evaluating quantized model accuracy (running in eager mode vs pybindings vs on-device, for example)

PR #12240, opened 2025-07-04
Quant doc updates

metascroy:
Initial draft of quantization doc updates (https://github.com/pytorch/executorch/issues/10603)

### DIFF
--- a/docs/source/quantization-overview.md
+++ b/docs/source/quantization-overview.md
@@ -1,3 +1,37 @@
+The current quantization overview page is a bit sparse: https://pytorch.org/executorch/main/quantization-overview.html. I'd like to update it as follows:
+
+Move under Usage/ since it's the only page under Quantization/ currently.
+Split out information intended for backend authors (info about writing a quantizer, for example). Focus on user-facing APIs.
+Document backend-invariant quantization flows (embeddings, ao kernels, etc.). Include info (and example) on composable quantizer.
+Document PT2E and quantize_ flows.
+Cover the general, high level approach to quantizing different types of models.
+CV models
+Transformers / language models
+Talk briefly about options for evaluating quantized model accuracy (running in eager mode vs pybindings vs on-device, for example)
+-----
+
+# Quantizing ExecuTorch Models
+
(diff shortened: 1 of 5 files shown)


GregoryComer (approved):
Looks great. Thanks for updating this.

metascroy:
@jerryzh168 can you give the changes a look over too?

docs/source/quantization-overview.md
```diff
@@ -1,38 +1,72 @@
...
-For an example quantization flow with ``XNNPACKQuantizer``, more documentation and tutorials, please see ``Performing Quantization`` section in [ExecuTorch tutorial](https://pytorch.org/executorch/main/tutorials/export-to-executorch-tutorial).
+The PT2E quantization workflow has three main steps:
 
-## Source Quantization: Int8DynActInt4WeightQuantizer
+1. Create a backend-specific quantizer.
+2. Prepare, calibrate, convert, and evalute the quantized model in PyTorch
+3. Lower the model to the target backend
 
-In addition to export based quantization (described above), ExecuTorch wants to highlight source based quantizations, accomplished via [torchao](https://github.com/pytorch/ao). Unlike export based quantization, source based quantization directly modifies the model prior to export. One specific example is `Int8DynActInt4WeightQuantizer`.
+## 1. Create a Backend-Specific Quantizer
```
jerryzh168: nit: I think `Configure` would be more accurate

docs/source/quantization-overview.md
```diff
@@ -1,38 +1,72 @@
...
+out_quantized = quantized_model(sample)
+sqnr = compute_error(out_reference, out_quantized) # SQNR error
 ```
+
+Note that numerics on device can differ those in PyTorch even for unquantized models, and accuracy evaluation can also be done with pybindings or on device.
+
+
+## 3. Lower the model
+
+The final step is to lower the quantized_model to the desired backend, as you would an unquantized one.  See backend-specific pages for lowering information.
```
jerryzh168: nit: link to docs as well?

jerryzh168 (approved):
makes sense, wondering if we want to connect with transformer (or optimum-executorch?) as well, like what we did in https://docs.pytorch.org/ao/main/serving.html

metascroy:
> makes sense, wondering if we want to connect with transformer (or optimum-executorch?) as well, like what we did in https://docs.pytorch.org/ao/main/serving.html

The LLM doc overhaul is separate

PR #12243, opened 2025-07-07
Arm backend: Match fp32->int32 cast between pytorch and TOSA's CAST

YufengShi-dudu:
- Pytorch .to() truncates values toward zeros, while TOSA's CAST rounds values to the nearest int. They will show different behaviours on fp32->int32 casting.
- Fix the mismatched behaviour with the following transformation: 
   Before:
     output = to_copy(x, torch.int32) 
   After:
     is_non_negative = x >= 0
     floor_x = floor(x)
     ceil_x = ceil(x)
     decorated_x = where(is_non_negative, floor_x, ceil_x)
     output = to_copy(decorated_x, torch.int32)

Change-Id: I21286432eeb0a5e2f21865f3ac097051c921a9b3


oscarandersson8218 (approved):
Unrelated CI failures. LGTM!

PR #11001, opened 2025-05-20
Gate training pybind build behind flag

JacobSzwejbka:
Differential Revision: D75079436

PR #12247, opened 2025-07-07
Add TrainingModule and SGD JNI + PTE-only Training Workflow

georgehong:
### Summary
Adds JNI for SGD and TrainingModule, including a unit test that mirrors train.cpp for a simple XOR example. Also makes the following change:

- Refactor jni_layer.cpp JTensor <--> Tensor conversion to be a general TensorHybrid utility. This is useful for TrainingModule classes that move maps of Tensors around.
- Updates `android_test_setup.sh` to match the pushd-popd directory movement for consistency and flexibility. This is also used to fix errors with generating the XOR files.

Training dependencies are already enabled for Java JNI library, so we skip adding additional guard flags.

### Test plan
Updated XOR tests that check .pte only convergence workflow.

```
sh scripts/build_android_library.sh
sh executorch_android/android_test_setup.sh // Creates xor.ptd, xor.pte, and xor_full.pte files.

./gradlew :executorch_android:connectedAndroidTest // Added unit test to check toy model convergence loss < 0.01
```
For the XOR tests, the device logs will show convergence values:
```
I testTrainXOR: Step 0, Loss 0.683540, Input [1, 0], Prediction 1, Label 1
...
I testTrainXOR: Step 4500, Loss 0.000994, Input [0, 0], Prediction 0, Label 0
```


extension/android/jni/jni_layer_training.cpp
```diff
@@ -0,0 +1,350 @@
...
+      "Lorg/pytorch/executorch/Tensor;";
+
+  static facebook::jni::local_ref<TensorHybrid::javaobject>
+  newJTensorFromTensor(const executorch::aten::Tensor& tensor);
+
+  static TensorPtr newTensorFromJTensor(
+      facebook::jni::alias_ref<TensorHybrid::javaobject> jtensor);
+};
+
+class JEValue : public facebook::jni::JavaClass<JEValue> {
```
JacobSzwejbka, line 44: Hmm does this not already exist for inference?

extension/android/executorch_android/src/main/java/org/pytorch/executorch/SGD.java
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * @param namedParameters Map of parameter names to tensors to be optimized
+   * @param learningRate The learning rate for the optimizer
+   * @param momentum The momentum value
+   * @param dampening The dampening value
+   * @param weightDecay The weight decay value
+   * @param nesterov Whether to use Nesterov momentum
+   * @return new {@link org.pytorch.executorch.SGD} object
+   */
+  public static SGD create(
```
JacobSzwejbka, line 67: Doesnt have to be this diff but would it be more "java-y" to have builder classes? 

new SGDBuilder(<params>).learning_rate().buildSGD();

extension/android/executorch_android/src/main/java/org/pytorch/executorch/TrainingModule.java
```diff
@@ -0,0 +1,119 @@
...
+  public EValue[] executeForwardBackward(String methodName, EValue... inputs) {
+    if (!mHybridData.isValid()) {
+      Log.e("ExecuTorch", "Attempt to use a destroyed module");
+      return new EValue[0];
+    }
+    return executeForwardBackwardNative(methodName, inputs);
+  }
+
+  @DoNotStrip
+  private native EValue[] executeForwardBackwardNative(String methodName, EValue... inputs);
```
JacobSzwejbka, line 96: noob q. What are these "native" apis?

extension/android/jni/jni_layer_training.cpp
georgehong: Yes, these are just forward declarations referencing what exists in *jni_layer.cpp* so that it doesn't all have to be in a single file. The actual definitions are in that same file.

extension/android/executorch_android/src/main/java/org/pytorch/executorch/TrainingModule.java
georgehong: Each of these `native` methods maps to a C++ definition in the JNI _jni_layer_training.cpp_ file. 
```
  static void registerNatives() {
    registerHybrid({
        makeNativeMethod("initHybrid", ExecuTorchTrainingJni::initHybrid),
        makeNativeMethod(
            "executeForwardBackwardNative",
            ExecuTorchTrainingJni::executeForwardBackward),
        makeNativeMethod(
            "namedParametersNative", ExecuTorchTrainingJni::namedParameters),
        makeNativeMethod(
            "namedGradientsNative", ExecuTorchTrainingJni::namedGradients),
    });
  }
};
```

extension/android/executorch_android/src/main/java/org/pytorch/executorch/SGD.java
georgehong: Yes, that sounds good - having an `SGDBuilder()` sounds like a great follow-up to me.

extension/android/jni/jni_layer_training.cpp
```diff
@@ -0,0 +1,350 @@
...
+#include <executorch/extension/training/module/training_module.h>
+#include <executorch/extension/training/optimizer/sgd.h>
+#include <executorch/runtime/core/portable_type/tensor_impl.h>
+#include <executorch/runtime/platform/log.h>
+#include <cassert>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <fbjni/ByteBuffer.h>
```
JacobSzwejbka, line 22: what happens with these fbjni bindings in oss?

JacobSzwejbka: Does ET have a dep on https://github.com/facebookincubator/fbjni?

extension/android/jni/jni_layer_training.cpp
JacobSzwejbka: Could we just put them in a header?

extension/android/jni/jni_layer_training.cpp
georgehong: I think we're already using FBJNI in _jni_layer_runtime.cpp_:

https://github.com/pytorch/executorch/blob/ed9c4de583024679e027a94f368831831278b2e1/extension/android/jni/jni_layer_runtime.cpp#L9-L10

and in the Gradle build:

https://github.com/pytorch/executorch/blob/ed9c4de583024679e027a94f368831831278b2e1/extension/android/executorch_android/build.gradle#L46-L51

extension/android/jni/jni_layer_training.cpp
georgehong: Currently, all of the JNI files apart from _jni_layer_constants.h_ are in implementation files. Would this be suitable for a follow-up diff, since the refactor would further increase the size of this PR as well as the possible surfaces for updating the build files?

GregoryComer:
Would it be possible to split out the new Buck selective JNI dependencies so they're only pulled in when needed? There are some binary-size critical users currently in Meta apps.

georgehong:
Split training components into separate BUCK target and added training JNI flag to make this JNI modular to address specific binary size concerns.

georgehong:
rebase

extension/android/CMakeLists.txt
```diff
@@ -145,6 +145,12 @@ if(EXECUTORCH_JNI_CUSTOM_LIBRARY)
   )
 endif()
 
+if(EXECUTORCH_BUILD_TRAINING_JNI)
```
jathu: We shouldn't introduce a separate flag for this — technically don't we always want to do this if `EXECUTORCH_BUILD_EXTENSION_TRAINING` is on? (given Android is assumed within extension/android)

scripts/build_android_library.sh
```diff
@@ -57,6 +57,7 @@ build_android_native_library() {
     -DEXECUTORCH_BUILD_KERNELS_QUANTIZED=ON \
     -DEXECUTORCH_BUILD_KERNELS_CUSTOM="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
     -DEXECUTORCH_BUILD_LLAMA_JNI="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
+    -DEXECUTORCH_BUILD_TRAINING_JNI="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
```
jathu: As mentioned above, we shouldn't need this. (also training flag should not depend on llm extension?)

georgehong: Thanks for the catch, it should be `EXECUTORCH_BUILD_EXTENSION_TRAINING`.

extension/android/CMakeLists.txt
georgehong: Is there a historical reason that `EXECUTORCH_BUILD_EXTENSION_TRAINING` was enabled without the corresponding JNI? This was motivated by Gregory's comments on possible binary size impact of JNI, so I wanted to use the LLM/LLAMA flags as an example and keep the modularity similar to the BUCK side, especially for inference-only workflows.

jathu: `EXECUTORCH_BUILD_EXTENSION_TRAINING` was just turned on in Android to provide [some vector for continuous testing](https://github.com/pytorch/executorch/pull/11001#discussion_r2099025359). Is it fair to assume if you are building for Android + training enabled → you want the corresponding JNI bridge?

Our buck builds can move independently of cmake for the most part, so @GregoryComer's suggestion makes sense

georgehong: I see, thanks for the context! With the latest push, the flags have been consolidated so that there's not a separate JNI flag for training.

jathu (approved):

JacobSzwejbka (approved):

georgehong:
Following the second approval, imported internal diff signals are green. Will attempt to merge.

PR #12268, opened 2025-07-08
Link backend to prtable libs

neuropilot-captain:
### Summary
1. Link portable OPs and kernels
2. Remove unused cmake checks


cccclai (approved):
Ah does it mean the previous models are fully lowered?

PR #12273, opened 2025-07-08
Fix declarations of blas functions

smeenai:
Summary:
Correct the return type in the declarations to match the actual
definitions.

Differential Revision: D77942171

### DIFF
--- a/kernels/optimized/blas/CPUBlas.cpp
+++ b/kernels/optimized/blas/CPUBlas.cpp
@@ -15,10 +15,10 @@
 #include <Accelerate/Accelerate.h>
 #else
 // clang-format off
-extern "C" void dgemm_(char *transa, char *transb, int *m, int *n, int *k, double *alpha, const double *a, int *lda, const double *b, int *ldb, double *beta, double *c, int *ldc);
-extern "C" void sgemm_(char *transa, char *transb, int *m, int *n, int *k, float *alpha, const float *a, int *lda, const float *b, int *ldb, float *beta, float *c, int *ldc);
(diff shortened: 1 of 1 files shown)


swolchok (changes requested):
Two problems here:

1) I don't think this is correct in general. [MKL](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2024-1/cblas-gemm-001.html#GUID-97718E5C-6E0A-44F0-B2B1-A551F0F164B2), [Accelerate](https://developer.apple.com/documentation/accelerate/cblas_dgemm(_:_:_:_:_:_:_:_:_:_:_:_:_:_:)?language=objc), and [BLIS](https://github.com/flame/blis/blob/ef98321a7c5d927fbcc7474ff2084d09bd764cc4/frame/compat/cblas/src/cblas.h#L428) all have the return types as void. [Eigen BLAS](https://github.com/PX4/eigen/blob/7cf1c0179eb0f5499dfc1bffbd229783a7865fe1/blas/level3_impl.h#L12) appears to be an outlier with its int return type; not sure how we are supposed to be compatible with both offhand.
2) If this is correct, this file is a partial copy from PyTorch; you should also patch it [there](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/CPUBlas.cpp#L19).

smeenai:
Eigen fixed this in https://gitlab.com/libeigen/eigen/-/commit/5361dea833df76fc0af1275e708940377dfc0851, so I'm just relying on that instead.

PR #12277, opened 2025-07-08
[EE/BE] add float variant to unary_ufunc_realhbbf16_to_floathbf16

manuelcandales:
* #12279
* #12278
* __->__ #12277

Differential Revision: [D77759060](https://our.internmc.facebook.com/intern/diff/D77759060/)

### DIFF
--- a/backends/cadence/fusion_g3/operators/op_rsqrt.cpp
+++ b/backends/cadence/fusion_g3/operators/op_rsqrt.cpp
@@ -27,7 +27,8 @@ namespace native {
 
 namespace {
 
-double rsqrt(double x) {
+template <typename T>
+T rsqrt(T x) {
   return 1.0 / std::sqrt(x);
 }
 
@@ -61,7 +62,7 @@ Tensor& rsqrt_out(KernelRuntimeContext& ctx, const Tensor& in, Tensor& out) {
     return out;
   } else {
     return torch::executor::native::internal::
-        unary_ufunc_realhbbf16_to_floathbf16(rsqrt, ctx, in, out);
+        unary_ufunc_realhbbf16_to_floathbf16(rsqrt, rsqrt, ctx, in, out);
   }
 }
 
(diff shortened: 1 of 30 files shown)


backends/cadence/fusion_g3/operators/op_rsqrt.cpp
```diff
@@ -27,7 +27,8 @@ namespace native {
 
 namespace {
 
-double rsqrt(double x) {
+template <typename T>
+T rsqrt(T x) {
   return 1.0 / std::sqrt(x);
```
swolchok, line 32: think this should be T(1.0) otherwise we're forcing computation in double, right?

backends/cadence/hifi/operators/op_rsqrt.cpp
```diff
@@ -21,7 +21,8 @@ namespace HiFi {
 namespace native {
 namespace {
 
-double rsqrt(double x) {
+template <typename T>
+T rsqrt(T x) {
   return 1.0 / std::sqrt(x);
```
swolchok, line 26: ditto T(1.0)

kernels/portable/cpu/pattern/pattern.h
```diff
@@ -84,11 +84,18 @@ Tensor& unary_ufunc_realhb_to_bool(
...
     KernelRuntimeContext& ctx,
     const Tensor& in,
     Tensor& out);
 
+#define DEFINE_UNARY_UFUNC_REALHBBF16_TO_FLOATHBF16(op_name, fn)              \
+  Tensor& op_name(KernelRuntimeContext& ctx, const Tensor& in, Tensor& out) { \
+    return internal::unary_ufunc_realhbbf16_to_floathbf16(                    \
+        fn, fn, ctx, in, out);                                                \
+  }
+
```
swolchok, line 98: why not define this macro to use the full namespace (prefixed with :: so it works from anywhere) so that backends/cadence can use it too?

swolchok (approved):
accepted with minor revision, see comments

PR #12292, opened 2025-07-08
Add mtk_build_examples.sh to CI

cccclai:
Add mtk_build_examples.sh script so it's covered by CI


cccclai:
Added in a separate PR

PR #12295, opened 2025-07-08
Add deprecation message for constant buffer branch

lucylq:
Summary: ^

Differential Revision: D77965467


larryliu0820 (approved):

Sebastian-Larsson:
@lucylq This broke Arm's run.sh script. Even with a clean build we end up on this code path/branch. Any idea what we might be doing wrong?

PR #12296, opened 2025-07-09
[Backend Test] Backend test reporting skeleton

GregoryComer:
### Summary
Add the initial skeleton of reporting code for the backend tester. This PR is primarily focused on putting the hooks and runner structure in place. Follow-up work will expand the scope of collection and reporting outputs.

This PR adds the following:
- CLI runner for the test suite.
- Basic test result breakdown by success / fail and cause (failing in lowering vs output mismatch, for example).
- Refactoring of test suite logic to clean things up.

Next steps:
- Aggregate results by flow (backend).
- Add additional CLI flags to allow filtering backends and dtypes.
- Land more of the operator test suite.
- Wire up flows for quantized operators.

Note that this PR is stacked on (and thus includes) https://github.com/pytorch/executorch/pull/11960. I accidently broke my ghstack, so I'm converting this to a normal PR.

Sample output (XNNPACK):
```
Test Session Summary:

   84 Passed / 95
   11 Failed / 95
    0 Skipped / 95

[Success]
   66 Delegated
   18 Undelegated

[Failure]
    4 Lowering Fail
    0 PTE Load Fail
    0 PTE Run Fail
    6 Output Mismatch Fail
```
Reproduce with `ET_TEST_ENABLED_BACKENDS=xnnpack python -m executorch.backends.test.suite.runner.executorch.backends.test.suite`. I've temporarily commented out non-f32 dtypes to work around some crashes in XNNPACK, which are non-recoverable from Python.

### DIFF
--- a/backends/test/suite/README.md
+++ b/backends/test/suite/README.md
@@ -4,6 +4,39 @@ This directory contains tests that validate correctness and coverage of backends
 
 These tests are intended to ensure that backends are robust and provide a smooth, "out-of-box" experience for users across the full span of input patterns. They are not intended to be a replacement for backend-specific tests, as they do not attempt to validate performance or that backends delegate operators that they expect to.
 
+## Running Tests and Interpreting Output
+Tests can be run from the command line, either using the runner.py entry point or the standard Python unittest runner. When running through runner.py, the test runner will report test statistics, including the number of tests with each result type.
+
+Backends can be specified with the `ET_TEST_ENABLED_BACKENDS` environment variable. By default, all available backends are enabled. Note that backends such as Core ML or Vulkan may require specific hardware or software to be available. See the documentation for each backend for information on requirements.
+
+Example:
+```
+ET_TEST_ENABLED_BACKENDS=xnnpack python -m executorch.backends.test.suite.runner
+```
+
+```
+2465 Passed / 2494
+16 Failed
+13 Skipped
+
+[Success]
+736 Delegated
+1729 Undelegated
+
+[Failure]
+5 Lowering Fail
+3 PTE Run Fail
+8 Output Mismatch Fail
+```
+
+Outcomes can be interpreted as follows:
+ * Success (delegated): The test passed and at least one op was delegated by the backend.
+ * Success (undelegated): The test passed with no ops delegated by the backend. This is a pass, as the partitioner works as intended.
+ * Skipped: test fails in eager or export (indicative of a test or dynamo issue).
+ * Lowering fail: The test fails in to_edge_transform_and_lower.
+ * PTE run failure: The test errors out when loading or running the method.
+ * Output mismatch failure: Output delta (vs eager) exceeds the configured tolerance.
+
 ## Backend Registration
 
 To plug into the test framework, each backend should provide an implementation of the Tester class, defined in backends/test/harness/tester.py. Backends can provide implementations of each stage, or use the default implementation, as appropriate.
--- a/backends/test/suite/reporting.py
+++ b/backends/test/suite/reporting.py
@@ -1,0 +1,163 @@
+from collections import Counter
+from dataclasses import dataclass
+from enum import IntEnum, nonmember
+
+
+class TestResult(IntEnum):
+    """Represents the result of a test case run, indicating success or a specific failure reason."""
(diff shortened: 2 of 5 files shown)


backends/test/suite/runner.py
```diff
@@ -0,0 +1,155 @@
...
+    # Only run the runtime portion if something was delegated.
+    if is_delegated:
+        try:
+            tester.to_executorch().serialize()
+        except Exception as e:
+            # We could introduce a result value for this, but I'm not sure it's necessary.
+            # We can do this if we ever see to_executorch() or serialize() fail due a backend issue.
+            return build_result(TestResult.UNKNOWN_FAIL, e)
+
+        # TODO We should consider refactoring the tester slightly to return more signal on
```
digantdesai, line 79: +1

backends/test/suite/runner.py
```diff
@@ -0,0 +1,155 @@
...
+            return build_result(TestResult.PTE_RUN_FAIL, e)
+    else:
+        return build_result(TestResult.SUCCESS_UNDELEGATED)
+
+    return build_result(TestResult.SUCCESS)
+
+
+def print_summary(summary: RunSummary):
+    print()
+    print("Test Session Summary:")
```
digantdesai, line 96: Add some version number? And make sure we are parsable

backends/test/suite/reporting.py
```diff
@@ -0,0 +1,163 @@
...
+@dataclass
+class RunSummary:
+    aggregated_results: dict[TestResult, int]
+    num_test_cases: int
+    test_case_summaries: list[TestCaseSummary]
+    total_failed: int
+    total_passed: int
+    total_skipped: int
+
+    @staticmethod
```
digantdesai:
```suggestion
    @classmethod
```

backends/test/suite/runner.py
```diff
@@ -0,0 +1,155 @@
...
+    parser.add_argument("test_path", nargs="?", help="Prefix filter for tests to run.")
+    return parser.parse_args()
+
+
+def runner_main():
+    args = parse_args()
+
+    begin_test_session()
+
+    test_path = args.test_path or "executorch.backends.test.suite"
```
digantdesai: If we add tests for other logic here like runner, reporting etc.
```suggestion
    test_path = args.test_path or "executorch.backends.test.suite.operators"
```

backends/test/suite/reporting.py
```diff
@@ -0,0 +1,163 @@
+from collections import Counter
```
digantdesai, line 1: tests for this file?

backends/test/suite/README.md
```diff
@@ -0,0 +1,15 @@
+# Backend Test Suite
```
digantdesai: Do you want to add a report related line?

backends/test/suite/runner.py
digantdesai, line 96: Do we want to say which backend or backends? Also further split per backend if >1?
May be like a verbosity flag on the report?

digantdesai (approved):
Looks good for a starting point. Left some comments.

backends/test/suite/runner.py
GregoryComer: Sure, I'll add a version number. I was intending to add a machine-readable file output, such as in json format. Is that sufficient for the use cases you're envisioning, or do you think we need machine parsable text output, as well?

GregoryComer: Yep, good callout. I was planning to group by test flow (backend) in the next PR. Verbosity control would be good, as well.

backends/test/suite/README.md
GregoryComer: Added, though FYI I intend to evolve the format pretty heavily. I'll keep the README up to date with that.

backends/test/suite/runner.py
GregoryComer: Done.

backends/test/suite/reporting.py
GregoryComer: +1. I'll cover in the follow-up, if no objections. There's a bit of evolution that's happening on this structure.

PR #12297, opened 2025-07-09
Qualcomm AI Engine Direct - Runtime Option

winskuo-quic:
### Summary
Supporting following options that can be set during both AOT and runtime:

- Log Level
- Performance Mode
- Profiling Level

### Test plan

- Log Level: Check `debug` message prefix exists. 
- Performance Mode: Ensure QNN SDK prints config log for performance, and ensure burst is faster than high power saver.
- Profiling Level: Turn profiling off in compile spec and add profiling flag in runtime, ensure profiler gets expected number of events.

### DIFF
--- a/examples/qualcomm/executor_runner/qnn_executor_runner.cpp
+++ b/examples/qualcomm/executor_runner/qnn_executor_runner.cpp
@@ -17,10 +17,15 @@
  * Currently we assume that the outputs are all fp32 tensors.
  */
 
+#include <executorch/backends/qualcomm/runtime/QnnBackendOptions.h>
 #include <executorch/backends/qualcomm/runtime/QnnExecuTorch.h>
+#include <executorch/backends/qualcomm/runtime/QnnExecuTorchBackend.h>
 #include <executorch/devtools/etdump/etdump_flatcc.h>
 #include <executorch/extension/data_loader/file_data_loader.h>
 #include <executorch/extension/runner_util/inputs.h>
+#include <executorch/runtime/backend/backend_option_context.h>
+#include <executorch/runtime/backend/interface.h>
+#include <executorch/runtime/backend/options.h>
 #include <executorch/runtime/core/memory_allocator.h>
 #include <executorch/runtime/executor/method.h>
 #include <executorch/runtime/executor/program.h>
@@ -33,7 +38,6 @@
 #include <fstream>
 #include <memory>
 #include <numeric>
-
 static uint8_t method_allocator_pool[4 * 1024U * 1024U]; // 4 MB
 
 DEFINE_string(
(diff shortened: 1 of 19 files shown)


examples/qualcomm/executor_runner/qnn_executor_runner.cpp
```diff
@@ -17,10 +17,15 @@
  * Currently we assume that the outputs are all fp32 tensors.
  */
 
+#include <executorch/backends/qualcomm/runtime/QnnBackendOptions.h>
```
cccclai: Is it needed to have these headers?
```
#include <executorch/backends/qualcomm/runtime/QnnBackendOptions.h>
#include <executorch/backends/qualcomm/runtime/QnnExecuTorchBackend.h>
```

cccclai: And these two
```
#include <executorch/runtime/backend/backend_option_context.h>
#include <executorch/runtime/backend/interface.h>
```

winskuo-quic: Thanks for reviewing the PR and catching this.
Some of these headers are not required and I have removed them.
However, `QnnExecuTorchBackend.h` would still be required since the backend name variable `QNN_BACKEND` is inside  `QnnExecuTorchBackend.h`. If there are any concerns, I can move this variable to somewhere else.
Thanks

cccclai: I see, I'm just trying to understand what headers are required to be added to the runtime. What is the alternative option for the `QNN_BACKEND` variable?

winskuo-quic: I have pushed a new commit that places all these macros under `executorch/backends/qualcomm/runtime/QnnExecuTorch.h`, so we don't need to include `QnnExecuTorchBackend.h`.
However, I will still need to include `executorch/runtime/backend/interface.h` since I need to call the `set_options` api.

cccclai: I see, I might refactor to move the set_options API. But let's merge this PR for now

cccclai:
Hmm lots of internal failure, might be easier if I move the API to <options.h> now

winskuo-quic:
> Hmm lots of internal failure, might be easier if I move the API to <options.h> now

Sure! I will rebase to mainline and change the header once the API is moved to <options.h>

cccclai:
Actually can you make this one line change
```
--- a/executorch/backends/qualcomm/runtime/targets.bzl
+++ b/executorch/backends/qualcomm/runtime/targets.bzl
@@ -75,11 +75,11 @@
                 "//executorch/backends/qualcomm:schema",
                 "//executorch/backends/qualcomm/aot/ir:qcir_utils",
                 "//executorch/backends/qualcomm/aot/wrappers:wrappers",
-                "//executorch/runtime/backend:interface",
                 "//executorch/runtime/core:core",
                 "//executorch/extension/tensor:tensor",
             ],
             exported_deps = [
+                "//executorch/runtime/backend:interface",
                 "//executorch/runtime/core/exec_aten/util:scalar_type_util",
                 "//executorch/runtime/core:event_tracer",
             ],
```
It seems fix the issue

winskuo-quic:
> Actually can you make this one line change
> 
> ```
> --- a/executorch/backends/qualcomm/runtime/targets.bzl
> +++ b/executorch/backends/qualcomm/runtime/targets.bzl
> @@ -75,11 +75,11 @@
>                  "//executorch/backends/qualcomm:schema",
>                  "//executorch/backends/qualcomm/aot/ir:qcir_utils",
>                  "//executorch/backends/qualcomm/aot/wrappers:wrappers",
> -                "//executorch/runtime/backend:interface",
>                  "//executorch/runtime/core:core",
>                  "//executorch/extension/tensor:tensor",
>              ],
>              exported_deps = [
> +                "//executorch/runtime/backend:interface",
>                  "//executorch/runtime/core/exec_aten/util:scalar_type_util",
>                  "//executorch/runtime/core:event_tracer",
>              ],
> ```
> 
> It seems fix the issue

I have the 1 line change fixed and also rebased to mainline. Thanks

cccclai (commented):
Thanks for adding the feature!

cccclai (approved):

PR #12300, opened 2025-07-09
Define edge exception and preserve as global

JakeStevens:
Summary: There is a list of ops provided as arguments to the to_edge_with_preserved_ops and EdgeCompileConfig. This PR pulls these lists out of the function and hoists it above the function as a global variable, allowing it to be imported into other code to align compiler behavior between different callsites.

Differential Revision: D78008871

### DIFF
--- a/backends/cadence/aot/compiler.py
+++ b/backends/cadence/aot/compiler.py
@@ -210,6 +210,21 @@ def quantize_pt2(
     return program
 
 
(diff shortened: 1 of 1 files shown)


mcremon-meta (approved):

mcremon-meta:
LGTM with linter fix :)

PR #12302, opened 2025-07-09
Updated pre-processing runtime merging logic

Juntian777:
Summary:
This PR updates the pre-processing merging logic for runtime intermediate outputs. Previously, we assumed that the runtime debug_handle tuple was ordered by integers, so when merging, we decided which intermediate output to keep based on which had a larger debug_handle[-1]. However, this is not always the case, so we need to change the logic to decide which intermediate output to keep based on the instruction ID instead of debug_handle[-1]. Additionally, the one with a higher instruction ID debug_handle tuple should strictly appear after the one with a lower instruction ID because AOT decides which intermediate output to keep based on the runtime_debug_handle[-1], so the order needs to be correct.

As AOT debug_handle is just a tuple with one element, the pre-processing merge logic for AOT has been removed as it is no longer necessary.

Differential Revision: D77905958

### DIFF
--- a/devtools/inspector/_inspector.py
+++ b/devtools/inspector/_inspector.py
@@ -60,6 +60,7 @@ from executorch.devtools.inspector._inspector_utils import (
     is_debug_output,
     is_inference_output_equal,
     map_runtime_aot_intermediate_outputs,
+    merge_runtime_overlapping_debug_handles,
     ProgramOutput,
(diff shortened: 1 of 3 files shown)


Gasoonjia (approved):
LGTM. Please solve my last comment then we can merge.

PR #12303, opened 2025-07-09
Add example for building executor_runner in Wasm

Conarnar:
### Summary
Since we want to have ExecuTorch on web builds, we first need to be able to compile it to Wasm. An example was added that gives instructions on how to build executor_runner with Emscripten, along with some modifications to CMakeLists.txt while Emscripten is used while building.

### Test plan
The test script can be run with `bash examples/wasm/test_build_wasm.sh` from the root directory. This attempts to build executor_runner and run it on the add_mul and mv2 models.

Added a CI test-build-wasm-linux to run this script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -727,6 +727,7 @@ if(EXECUTORCH_BUILD_EXECUTOR_RUNNER)
   target_link_libraries(executor_runner ${_executor_runner_libs})
   target_compile_options(executor_runner PUBLIC ${_common_compile_options})
 
+  # Automatically set when using `emcmake cmake` for Wasm build.
   if(EMSCRIPTEN)
     # Directory of model pte files to embed in the wasm binary.
     if(NOT DEFINED WASM_MODEL_DIR)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -726,6 +726,16 @@ if(EXECUTORCH_BUILD_EXECUTOR_RUNNER)
   endif()
   target_link_libraries(executor_runner ${_executor_runner_libs})
   target_compile_options(executor_runner PUBLIC ${_common_compile_options})
+
+  if(EMSCRIPTEN)
+    # Directory of model pte files to embed in the wasm binary.
+    if(NOT DEFINED WASM_MODEL_DIR)
+      set(WASM_MODEL_DIR "${CMAKE_SOURCE_DIR}/models/")
+    endif()
+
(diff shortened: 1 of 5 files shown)


examples/wasm/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+examples/wasm
+└── README.md                         # This file
+```
+
+## Prerequisites
+
+- [emscripten](https://emscripten.org/docs/getting_started/Tutorial.html)
+- [Node.js](https://nodejs.org/en/) (Optional)
+
```
lucylq, line 20: Wdyt about adding a quick-start section to run the example that runs the CI test commands e.g.
```
bash .ci/scripts/setup-emscripten.sh
bash test_build_wasm.sh
```

CMakeLists.txt
```diff
@@ -726,6 +726,16 @@ if(EXECUTORCH_BUILD_EXECUTOR_RUNNER)
   endif()
   target_link_libraries(executor_runner ${_executor_runner_libs})
   target_compile_options(executor_runner PUBLIC ${_common_compile_options})
+
+  if(EMSCRIPTEN)
```
lucylq, line 731: Can we a comment here to say this is automatically set when we use emcmake/emmake?

examples/wasm/README.md
```diff
@@ -0,0 +1,64 @@
...
+node cmake-out-wasm/executor_runner.js --model_path mv2.pte
+```
+
+5. You can also run the model in the browser. Note that you cannot pass command line arguments to the browser version of the tool. By default, the program will load the model `model.pte` and run it. Several browsers do not support `file://` XHR requests to load the Wasm file. To get around this, you can use a local web server. For example, with Python:
+
+```bash
+python3 -m http.server --directory cmake-out-wasm
+```
+
+The page will be available at http://localhost:8000/executor_runner.html.
```
lucylq, line 76: Maybe add a section for Common Issues/FAQ. Seems like Node v22.16.0 the expected working version, and expect
```
echo $EMSDK_NODE
.../emsdk/node/22.16.0_64bit/bin/node
```

lucylq (approved):

lucylq:
Thanks for adding the CI test, and all the debugging to get it working! Great job.

CMakeLists.txt
Conarnar: Will do.

lucylq (approved):

PR #12315, opened 2025-07-09
PyMethod inputs use tensor_ptr_maker to convert attensor to etensor

Conarnar:
Summary: Previous implementations of Pybindings use TensorImpl to convert from attensor to etensors. Using TensorPtr is a cleaner method. D77759383 fixes a bug that previously prevented the CI from passing when using tensor_ptr_maker in Pybindings.

Differential Revision: D78035237


lucylq (approved):
Looks good, thanks for updating it!

PR #12320, opened 2025-07-09
Call target_link_options_shared_lib in executorch-config.cmake

swolchok:
Customers need the flags this sets (`--whole-archive` or `-force_load`), otherwise the built operator libraries won't actually register operators.

Testing: ~~I'm not sure that we have automated coverage for this file, so I'm sending a first version adding an intentional break.~~ We have automated coverage for this file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -490,7 +490,7 @@ install(
   INCLUDES
   DESTINATION ${_common_include_directories}
 )
-install(FILES tools/cmake/executorch-config.cmake
+install(FILES tools/cmake/Utils.cmake tools/cmake/executorch-config.cmake
         DESTINATION lib/cmake/ExecuTorch
 )
 
--- a/tools/cmake/executorch-config.cmake
+++ b/tools/cmake/executorch-config.cmake
@@ -24,7 +24,7 @@
 # the contract of exposing these CMake variables.
 
 cmake_minimum_required(VERSION 3.19)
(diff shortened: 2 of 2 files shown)


swolchok:
* __->__ #12320

swolchok:
size regression seems to be several KB, need to investigate

swolchok:
binary size tests are passing. looks like a lot of jobs were cancelled and we just need to retry

larryliu0820 (approved):
Thank you this makes sense. We can also remove the redundant target_link_options_shared_lib in a lot of subdirectories as well.

swolchok:
> We can also remove the redundant target_link_options_shared_lib in a lot of subdirectories as well.

1) I don't think it's redundant, sadly. This only applies to projects using ExecuTorch as an external project.
2) What we really should want is to set up CMake EXPORT and remove all the duplicate configuration in executorch-config.cmake. It is only there because we aren't using EXPORT. Last time I looked, I thought we couldn't use EXPORT because some of our dependencies don't support it, but perhaps there's a way to get around that.

larryliu0820:
> 1. I don't think it's redundant, sadly.

This error is what called redundant: https://github.com/pytorch/executorch/actions/runs/16207706949/job/45761626062

larryliu0820:
> 2\. What we really should want is to set up CMake EXPORT

Yep agreed. We should do that after we have CMake top level targets ready.

PR #12323, opened 2025-07-09
Build Android in one pass

### LINKED ISSUE #10120: Make "CMake building LLM with ET just one pass" possible

From Android issue https://github.com/pytorch/executorch/issues/6767#issuecomment-2795384036, but this is generic. 

Currently if we build LLM, it needs to be two passes, one for general ET, one for LLM. For example, https://github.com/pytorch/executorch/blob/main/scripts/build_android_library.sh#L66-L83

The reason why we can't make it one-pass is, when we add_subdirectory(), recursively, two libraries have the same name `memory`.
ET -> XNNPACK -> memory
LLM -> tokenizer -> abseil -> memory

Seems that use `ExternalProject_Add` might be an alternative.

cc @larryliu0820 

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

kirklandsign:
### Summary
Now trigger cmake once. 
Fixes https://github.com/pytorch/executorch/issues/10120 

### Test plan
CI


scripts/build_android_library.sh
```diff
@@ -55,6 +55,7 @@ build_android_native_library() {
     -DEXECUTORCH_BUILD_KERNELS_OPTIMIZED=ON \
```
larryliu0820, line 56: Not in this PR but can you add an Android preset?

larryliu0820 (approved):

PR #12324, opened 2025-07-09
Bump transformer version and fix Llava example

larryliu0820:
A lot has changed on Llava model definition between 4.47 to 4.52, this PR:

* Change the state dict key mapping to match the new Llava model definition in HF.
* Use the `processor.apply_chat_template()` API to get `input_id`s so that we can be a bit more resilient to input_id format changes.


lucylq (approved):
Thanks for updating this!

PR #12325, opened 2025-07-10
Qualcomm AI Engine Direct - Add Unittest for Saver

DannyYuyang-quic:
### Summary
- Add Unittest for Saver backend

### Test plan
```
python -m backends.qualcomm.tests.test_qnn_delegate TestQNNQuantizedUtils.test_qnn_backend_saver_backend -s ${device_id} -H ${host_id} -m ${soc} -b build-android
```


backends/qualcomm/qnn_preprocess.py
```diff
@@ -178,6 +178,11 @@ def preprocess_multimethod(
 
             if len(py_op_wrapper_list) == len(edge_programs.values()):
                 qnn_context_binary = qnn_manager.Compile(graph_name, py_op_wrapper_list)
+                if option.saver:
+                    # TODO: Currently, only the first method is saved. Update this logic if saving multiple methods becomes necessary in the future.
```
cccclai, line 182: Can you share what it means?

DannyYuyang-quic: Consider the scenario for lowering a multi-method pte:

When the saver option is enabled in QNN, the program is configured to not generate context binary. So, the condition `len(qnn_context_binary) != 0` will be False. In this case, we hope the program can exit immediately before reaching the assertion checks. As a result, only the result for `method_0` will be saved. If support for saving multi-method with `option.saver=True` is required in the future, this will need to be extended.

cccclai (approved):
Thank you!

cccclai:
Can you rebase?

PR #12329, opened 2025-07-10
Update android-release-artifacts.yml

kirklandsign:
Don't use ephemeral.linux.2xlarge


kirklandsign:
Hi @huydhn seems the workflow will otherwise stuck (https://github.com/pytorch/executorch/actions/workflows/android-release-artifacts.yml) when we use ephemeral.linux.2xlarge

huydhn (approved):
Yeah, there is no ephemeral prefix or suffix now because every Linux runners are ephemeral by default nowadays

PR #12334, opened 2025-07-10
Arm backend: Add dump_delegate_data function

wwwind:
Add dump_delegate_data function


exir/program/_program.py
```diff
@@ -1726,6 +1726,75 @@ def dump_executorch_program(
         else:
             print_program(self._emitter_output.program, out=out)
 
+    def dump_delegate_data(  # noqa: C901
+        self,
+        path: str,
```
JacobSzwejbka, line 1853: can you take a TextIO instead? Like Dump_et_program above

edit: Or just any stream really since the data isnt text

exir/program/_program.py
```diff
@@ -1726,6 +1726,75 @@ def dump_executorch_program(
...
+        extension: str,
+        delegate_id: Optional[str] = None,
+    ) -> None:
+        """
+        Dumps the delegate blob out of backend_delegate_data to <path><extension>.
+        Must have been created with extract_delegate_segments=True.
```
JacobSzwejbka, line 1859: Can you not find the blob if its embedded in the flatbuffer section?

digantdesai (changes requested):
Thanks @wwwind for the PR. 

Is the goal for extracting the blobs is to run a delegate subgraph on a TOSA reference model? If yes, then I don't think I prefer this approach of extracting the blobs and running them.

The main reason is e2e validation. You can have N blobs which will lead to N tosa_ref_sim calls and then you might have to stitch them together manually to see if the entire network worked correctly or not.

Alternatively I propose we can write a tosa runtime which works with tosa partitioner and graph breaks can be handled by portable libs seamlessly. This way you can re-use all the ET infra and have e2e working TOSA PTE :)

wwwind:
Thank you for the review @digantdesai 

Actually, this function is for cases when we don't need e2e flow. We have a graphic use case when we get these blobs as .tosa files and then we have a plugin which implements different bits around it in shaders and pass some data to run subgraphs.

digantdesai:
> Thank you for the review @digantdesai
> 
> Actually, this function is for cases when we don't need e2e flow. We have a graphic use case when we get these blobs as .tosa files and then we have a plugin which implements different bits around it in shaders and pass some data to run subgraphs.

Ok. Curious why can't we dump them after generation from `preprocess`? Extracting feels a bit like going backwards.

wwwind:
I leave this to @per to answer as I am not sure how this could be done after preprocess

per:
> Ok. Curious why can't we dump them after generation from `preprocess`? Extracting feels a bit like going backwards.

Agree, it is a bit like going backwards, _but_ with a dedicated API intstead of using different flags for the backend to dump the delegate data. For the Arm backends it will be usuable for gettning the delegate data out of the .pte for all our backends, while the primary driver is the VgfBackend and internal testing scenarios.
Gold would be to also have some dump method (akin to objdump) on the serialized .pte files (maybe a executorch.runtime.Program method) to be able to extract the delegate data for a program serialized to file.

digantdesai:
> > Ok. Curious why can't we dump them after generation from `preprocess`? Extracting feels a bit like going backwards.
> 
> Agree, it is a bit like going backwards, _but_ with a dedicated API intstead of using different flags for the backend to dump the delegate data. For the Arm backends it will be usuable for gettning the delegate data out of the .pte for all our backends, while the primary driver is the VgfBackend and internal testing scenarios. Gold would be to also have some dump method (akin to objdump) on the serialized .pte files (maybe a executorch.runtime.Program method) to be able to extract the delegate data for a program serialized to file.

Ok I don't have strong opinions, and don't want to block this. @JacobSzwejbka  would be more appropriate for the review.


That said, thinking out loud why `VGFPartitioner(dump_blobs=True)` wouldn't be a better interface? Let me think some more about your comment for delegate agnostic AoT API. 

The reason I am hesitant with the current approach is that (1) it doesn't let you avoid the AoT PTE generation flow, hence the opinion about the AoT API, (2) `pte_objdump` is being considered however that is more FC/BC tooling maintenance around the PTE as a file format, esp if we are expecting users to go through AoT flow anyway. And will get more complicated with NamedDataMap etc. features to share tensors across delegates.

digantdesai:
Any update on this?

wwwind:
@digantdesai @per What is the final decision on this API ? Do we go forward with it or should I abandon this PR ?

PR #11849, opened 2025-06-23
Arm backend: Create ethosu directory

Sebastian-Larsson:
Move EthosU related files to its own directory.

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #12338, opened 2025-07-10
Arm backend: Create ethosu directory (#11849)

digantdesai:
Summary:
Move EthosU related files to its own directory.

cc digantdesai freddan80 per zingo oscarandersson8218

Differential Revision: D77349641

Pulled By: digantdesai


backends/arm/test/TARGETS
```diff
@@ -41,7 +41,7 @@ python_library(
     deps = [
         ":common",
         "//executorch/backends/xnnpack/test/tester:tester",
-        "//executorch/backends/arm:arm_partitioner",
```
Sebastian-Larsson, line 44: Are there any users left of arm_partitioner?  What happens to tosa_backend, tosa_partitioner etc.?

We're planning to move those to separate directories just like with this change, but I'm still wondering about the implications here.

Sebastian-Larsson: @digantdesai

digantdesai: Yes, so Arm partitioner is a target for all partitions - 

https://github.com/pytorch/executorch/blob/8da2ea6ea98062d73ff5c8da1d9b1ce99ca11a18/backends/arm/TARGETS#L4

Feel free to put up a PR for your refactor I can try to do the buck change.

In the future I do want to test this buck file in GH OSS CI to allow you to move fast and test your changes in OSS.

Sebastian-Larsson: Great! The changes look good to me, do you need approval from someone else at Arm or how do we proceed here?

digantdesai: Approve stamp? :-p

AdrianLundell (approved):

PR #12340, opened 2025-07-10
Add more models as part of GA models

mergennachin:
Summary

  Added 8 new representative models to ExecutorTorch examples:

  - EfficientNet-B4: Image classification with CNN architecture
  - DETR-ResNet50: Object detection using transformer decoder
  - SegFormer-ADE: Semantic segmentation transformer
  - Swin2SR: Super-resolution with Swin transformer
  - ALBERT: Lightweight BERT for NLP tasks
  - TrOCR: Optical character recognition transformer
  - Wav2Vec2: Cross-lingual speech representation learning

  All models include XNNPACK backend support with appropriate quantization configurations and full CI integration.

  Test plan:

  - Validate model export and execution with portable backend
  - Test XNNPACK delegation and quantization (with appropriate exclusions)
  - Integrate into CI workflows for automated testing
  - Verify all models perform their intended tasks accurately

### DIFF
--- a/examples/xnnpack/__init__.py
+++ b/examples/xnnpack/__init__.py
@@ -45,6 +45,13 @@ MODEL_NAME_TO_OPTIONS = {
     "emformer_join": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
     "emformer_predict": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
     "emformer_transcribe": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "efficientnet_b4": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "detr_resnet50": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "segformer_ade": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "swin2sr_2x": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "albert": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
+    "trocr_handwritten": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "wav2vec2": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
 }
 
 
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -317,8 +317,9 @@ elif [[ "${BACKEND}" == *"xnnpack"* ]]; then
   echo "Testing ${MODEL_NAME} with xnnpack..."
   WITH_QUANTIZATION=true
   WITH_DELEGATION=true
-  if [[ "$MODEL_NAME" == "mobilebert" ]]; then
-    # TODO(T197452682)
+  if [[ "$MODEL_NAME" == "mobilebert" || "$MODEL_NAME" == "albert" ]]; then
+    # TODO(https://github.com/pytorch/executorch/issues/12341) 
+    # mobilebert, albert incompatible with XNNPACK quantization
     WITH_QUANTIZATION=false
   fi
   test_model_with_xnnpack "${WITH_QUANTIZATION}" "${WITH_DELEGATION}"
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -63,7 +63,7 @@ jobs:
       contents: read
     strategy:
       matrix:
-        model: [linear, add, add_mul, ic3, ic4, mv2, mv3, resnet18, resnet50, vit, w2l, mobilebert, emformer_join, emformer_transcribe]
+        model: [linear, add, add_mul, ic3, ic4, mv2, mv3, resnet18, resnet50, vit, w2l, mobilebert, emformer_join, emformer_transcribe, efficientnet_b4, detr_resnet50, segformer_ade, albert, wav2vec2]
         backend: [portable, xnnpack-quantization-delegation]
         runner: [linux.arm64.2xlarge]
         include:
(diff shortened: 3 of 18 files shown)


mergennachin:
Tests failures don't look related

swin2sr_2x and trocr_handwritten are not exporting yet. Disabled for now. https://github.com/pytorch/executorch/issues/12365

Need to export with strict=False so that these models are exportable. 

However, when enabling strict=False by default, https://github.com/pytorch/executorch/pull/12368, there are a few failures, namely ic3,ic4 and llama. https://github.com/pytorch/executorch/issues/12370

kimishpatel:
how were these specific models were selected?

examples/models/bilstm/model.py
```diff
@@ -0,0 +1,96 @@
...
+
+import logging
+
+import torch
+import torch.nn as nn
+
+from ..model_base import EagerModelBase
+
+
+class BidirectionalLSTM(nn.Module):
```
kimishpatel: this just seems like a sample model to contain lstm? Should we instead include something else? @jackzhxng was looking into kokoro that also has lstm. although export is a problem there

kimishpatel:
The other question I would have is, what would be the goal of adding more models? Do we just want to claim enablement or at the least provide good perf via xnnpack lowering? Because this might end up expanding the work that needs to happen, not 100% sure, but just calling it out.

Do agree on covering more vision/object detection

examples/models/bilstm/model.py
mergennachin: Yeah, I will delete this one. This one's not really useful

mergennachin:
Thanks @kimishpatel 

> how were these specific models were selected?

So, i followed this process. I looked at https://ai-benchmark.com/ranking.html and https://mlcommons.org/benchmarks/inference-mobile/ models, and see if we can enable any delta that we didn't have. However, it was difficult to find the "official" pytorch and/or huggingface implementation. Most of them are TFLite or ONNX implementations. For some, we were able to find pytorch, like EfficientNet-B4, Albert

So, I swapped with similar models that already has: 

Object detection: I swapped Yolov4, SSD-MobileNetV2 with DETR-ResNet50 
Semantic Segmentation: I swapped DeepLabV3+ with SegFormer-ADE (even though we already have deeplabv3)
Image super resolution: Swapped ESRGAN with Swin2SR
OCR: Picked TrOCR

> The other question I would have is, what would be the goal of adding more models? 

Mainly inspired by the the ai-benchmark and mlcommons. The goal would be covering specific tasks that we don't have coverage. We won't be adding many models but only representative models within each tasks. There are still a few missing important ones like depth estimation and video super resolution

For more and more models in specific tasks, we can expand to leverage optimum-executorch instead.

cccclai:
Any specific reason that ai-benchmark and mlcommons are picked as a reference for the model list? Just curious because there are lists from other source

guangy10:
@mergennachin thanks for the clarification. And I'm happy to know how we can make a joint effort regarding get more task/model coverage, something happening in optimum-et. Some models like EfficientNet, Albert Swin, were added there already.

> I looked at https://ai-benchmark.com/ranking.html and https://mlcommons.org/benchmarks/inference-mobile/ models, and see if we can enable any delta that we didn't have. However, it was difficult to find the "official" pytorch and/or huggingface implementation. Most of them are TFLite or ONNX implementations.

Looks like the main motivation is to enable ai-benchmark to use ET generated models? I recall in the meeting with them, they mentioned that we are deploying exactly same model across all devices and the expense of switching to different variant of models are high. Given it, would adding similar but not exact same models help with adoption by ai-benchmark?

 > There are still a few missing important ones like depth estimation and video super resolution

I recall there are depth estimation models enabled in HF transformers (e.g. [DepthAnything](https://github.com/huggingface/transformers/blob/70e57e4710d8a617a6f0ea73183d9bc4c91063c9/tests/models/depth_anything/test_modeling_depth_anything.py#L289-L317)). Not sure about video super resolution (HF doesn't have a task classification for image/video super resolution). 

> The goal would be covering specific tasks that we don't have coverage. We won't be adding many models but only representative models within each tasks.

What are other tasks we would like to cover for GA? Would it be a good idea if we browse top 1-2 popular models based on the task classification on hugging face hub?

examples/xnnpack/__init__.py
```diff
@@ -45,6 +45,13 @@ class XNNPACKOptions(object):
     "emformer_join": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
     "emformer_predict": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
     "emformer_transcribe": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "efficientnet_b4": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "detr_resnet50": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "segformer_ade": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "swin2sr_2x": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "albert": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
+    "trocr_handwritten": XNNPACKOptions(QuantType.STATIC_PER_CHANNEL, True),
+    "wav2vec2": XNNPACKOptions(QuantType.DYNAMIC_PER_CHANNEL, True),
```
guangy10, line 54: @mergennachin Are those models calibrated after statically quantized? And I'm curious to know how you decide how to quantize the model (static vs dynamic) and how the quality of the quantized model is validated?

mergennachin:
> What are other tasks we would like to cover for GA? Would it be a good idea if we browse top 1-2 popular models based on the task classification on hugging face hub?

@guangy10 here's what I compiled: https://github.com/pytorch/executorch/issues/12378

digantdesai:
> Any specific reason that ai-benchmark and mlcommons are picked as a reference for the model list? Just curious because there are lists from other source

Popularity? Please include other popular lists here so we can try them as well?

PR #12348, opened 2025-07-10
PAL File for Arm Zephyr Executor Runner

BujSet:
### Summary
Platform file for building Zephyr applications that use Executorch as a module. This is a simplified version of the file used in a Zephyr development branch [here](https://github.com/BujSet/zephyr/blob/executorch-module-integration/samples/modules/executorch/arm/hello_world/src/arm_zephyr_pal.hpp).

In the future, it may be prudent to have a PAL file for arm bare metal (I imagine this would require a simple splicing of he PAL functions from the `arm_executor_runner.cpp` file). Leaving that task for a separate PR. 

### Test plan
Tested against the `arm_executor_runner.cpp` that is used in the [Arm-EthosU tutorial](https://docs.pytorch.org/executorch/main/tutorial-arm-ethos-u.html). Was able to successfully build the Arm Executor Runner, and run the simple add model.

### DIFF
--- a/runtime/platform/default/arm_zephyr.hpp
+++ b/runtime/platform/default/arm_zephyr.hpp
@@ -41,7 +41,7 @@ void et_pal_emit_log_message(
       message);
 }
 
-void* et_pal_allocate(ET_UNUSED size_t size) {
+void* et_pal_allocate(size_t size) {
     return k_malloc(size);
 }
 
--- a/runtime/platform/default/arm_zephyr.hpp
+++ b/runtime/platform/default/arm_zephyr.hpp
@@ -1,0 +1,52 @@
+#ifndef _ARM_ZEPHYR_PAL_HPP
+#define _ARM_ZEPHYR_PAL_HPP
+
+#include <zephyr/kernel.h>
+#include <zephyr/sys/printk.h>
+
+void et_pal_init(void) {}
+
+ET_NORETURN void et_pal_abort(void) {
+    _exit(-1);
+}
+
+et_timestamp_t et_pal_current_ticks(void) {
+    return k_uptime_ticks();
+}
+
+et_tick_ratio_t et_pal_ticks_to_ns_multiplier(void) {
+  // Since we don't know the CPU freq for your target and just cycles in the
+  // FVP for et_pal_current_ticks() we return a conversion ratio of 1
+  return {1, 1};
+}
+
+/**
+ * Emit a log message via platform output (serial port, console, etc).
+ */
+void et_pal_emit_log_message(
+    ET_UNUSED et_timestamp_t timestamp,
+    et_pal_log_level_t level,
+    const char* filename,
+    ET_UNUSED const char* function,
+    size_t line,
+    const char* message,
+    ET_UNUSED size_t length) {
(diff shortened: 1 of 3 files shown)


larryliu0820:
Can you change the zephyr preset to use this file instead of minimal or posix?

runtime/platform/default/arm_zephyr.hpp
```diff
@@ -0,0 +1,52 @@
...
+      stderr,
+      "%c [executorch:%s:%zu %s()] %s\n",
+      level,
+      filename,
+      line,
+      function,
+      message);
+}
+
+void* et_pal_allocate(ET_UNUSED size_t size) {
```
JacobSzwejbka: its not unused anymore right? Will it not override the symbol if you remove that?

larryliu0820:
> Can you change the zephyr preset to use this file instead of minimal or posix?

For this you will have to change override the option: https://github.com/pytorch/executorch/blob/main/tools/cmake/preset/default.cmake#L38C3-L38C35

BujSet:
> Can you change the zephyr preset to use this file instead of minimal or posix?

Hm, is this what we want for the preset? The PAL requires access to the [Zephyr library](https://github.com/pytorch/executorch/pull/12348/files#diff-be98a5ce0bf91a01b297b7c0100d3bfcde4253e70d5855449c8796f94e079301R4-R5), so if I set this in the preset, the build-preset test will fail unless we build the entire test with zephyr. Similarly, if I make the change to the preset file where I set:

```
set(EXECUTORCH_PAL_DEFAULT_FILE_PATH              "${PROJECT_SOURCE_DIR}/runtime/platform/default/arm_zephyr.hpp")
```
The commands from the original zephyr cmake preset [PR](https://github.com/pytorch/executorch/pull/11923) will error out. What do you think @larryliu0820?

runtime/platform/default/arm_zephyr.hpp
BujSet: Removed it and still works, made the change here.

runtime/platform/default/arm_zephyr.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef _ARM_ZEPHYR_PAL_HPP
+#define _ARM_ZEPHYR_PAL_HPP
+
+#include <zephyr/kernel.h>
```
digantdesai: Now we are in business :)

digantdesai:
No CMake changes?

BujSet:
> No CMake changes?

@digantdesai, kind of the same comment as above in response to Mengwei's question about the preset. The problem is that this file can't be built without linking to an existing zephyr library. To avoid adding zephyr as a dependency, I made the PAL a header file instead of a cpp source file so that it never needs to be built, but can be included when Executorch is built as module of zephyr.

larryliu0820 (approved):
ok I think this is fine, but I really think we should have a CI job in ExecuTorch that:
* install zephyr and executorch as a module of it (you can check out a branch of zephyr)
* compile with arm runner and memory allocator and PAL
* run a hello world example

So maybe we can have a quick followup PR

PR #12354, opened 2025-07-10
forward fix executorch slice copy tests

pianpwk:
Summary:
T230513957

D77745298 made it so less no-op slices are emitted by export in the first place

Reviewed By: mcr229, GregoryComer

Differential Revision: D78103869


digantdesai (approved):
Thank you.

PR #12368, opened 2025-07-10
Run test_models.sh with strict=False flag

mergennachin:
This is the recommended anyway, so let's gradually start migrating to strict=False

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -91,11 +91,10 @@ test_model() {
     run_portable_executor_runner
     rm "./${MODEL_NAME}.pte"
   fi
-  STRICT="--strict"
+  STRICT="--no-strict"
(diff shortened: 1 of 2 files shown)


digantdesai (commented):
I remember (not sure how accurately) but there were some PTE size bloats reported with `strict==false` and I am not sure we have a CI to catch that. I think we should validate that and then start embracing this.

cc @GregoryComer - who I think reported this a while back for a quant model.

mcr229:
@digantdesai yea we should probably check that model size isn't bloated. We don't have any CI here that checks for this so right now we would have to do so manually.

digantdesai:
> @digantdesai yea we should probably check that model size isn't bloated. We don't have any CI here that checks for this so right now we would have to do so manually.

Can we add a CI to make sure it doesn't double or something, just a guard against regressing.

PR #12371, opened 2025-07-10
Expand CoreML CI for model testing

mergennachin:
- **Add more models to coreml testing**

We're currently testing periodically in device farm on actual iphones.

However, we should also keep testing on macrunners directly in trunk

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -18,8 +18,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-models-macos:
-    name: test-models-macos
+  test-models-macos-cpu:
+    name: test-models-macos-cpu
     uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/trunk.yml
```diff
@@ -18,8 +18,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-models-macos:
-    name: test-models-macos
+  test-models-macos-cpu:
+    name: test-models-macos-cpu
```
digantdesai, line 22: Nit: Just to more backend specific (and not confuse with XNNPACK)
```suggestion
    name: test-models-macos-portable
```

.github/workflows/trunk.yml
```diff
@@ -55,6 +55,31 @@ jobs:
...
+  test-models-macos-coreml:
+    name: test-models-macos-coreml
+    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
+    strategy:
+      matrix: 
+        model: [dl3, edsr, efficient_sam, emformer_join, emformer_transcribe, ic3, mobilebert, mv3, resnet50, vit, w2l]
+        backend: [coreml]
+      fail-fast: false
+    with:
+      runner: macos-m1-stable
```
digantdesai, line 67: IIRC this runner also had bandwidth issue and things would timeout. I am ok with this, let's see what happens.

.github/workflows/trunk.yml
```diff
@@ -55,6 +55,31 @@ jobs:
         # Build and test executorch
         PYTHON_EXECUTABLE=python ${CONDA_RUN} bash .ci/scripts/test_model.sh "${MODEL_NAME}" "${BUILD_TOOL}" "${BACKEND}"
 
+  test-models-macos-coreml:
+    name: test-models-macos-coreml
+    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
+    strategy:
+      matrix: 
+        model: [dl3, edsr, efficient_sam, emformer_join, emformer_transcribe, ic3, mobilebert, mv3, resnet50, vit, w2l]
+        backend: [coreml]
```
digantdesai: How is this different from `test-apple-model` besides number of models? Ah no mps. But so why do we have coreml there as well?

digantdesai (approved):
LGTM, left some comments.

.github/workflows/trunk.yml
mergennachin: Yeah good point. 

I removed coreml stuff from test-apple-model and left the mps only. And renamed test-models-macos-mps to be consistent

.github/workflows/trunk.yml
mergennachin: it's actually both... xnnpack and portable, lol

mergennachin:
Dropping this in lieu of https://github.com/pytorch/executorch/pull/12808

PR #12382, opened 2025-07-11
Remove EXECUTORCH_BUILD_LLAMA_JNI cmake flag

jathu:
### Summary
I think `EXECUTORCH_BUILD_LLAMA_JNI` is redundant. If we are building for Android (assumed true under `extension/android`) and with `EXECUTORCH_BUILD_EXTENSION_LLM`, then it implies we want the JNI bindings!

There is a C++ def `EXECUTORCH_BUILD_LLAMA_JNI` that does some registration:

https://github.com/pytorch/executorch/blob/31ba95966e3ef5e328b4bc9f787ad49a047c862b/extension/android/jni/jni_layer.cpp#L488-L493

However, we turn them on for both BUCK:
https://github.com/pytorch/executorch/blob/31ba95966e3ef5e328b4bc9f787ad49a047c862b/extension/android/jni/BUCK#L72-L82

And cmake:

https://github.com/pytorch/executorch/blob/b5f950ba648fab9a23fc4676b8126162a0584219/extension/android/CMakeLists.txt#L148-L151

### Test plan

CI


GregoryComer (approved):
Makes sense. Thanks for spotting and cleaning this up.

larryliu0820:
I think this is a good change. cc @kirklandsign

PR #12386, opened 2025-07-11
Try removing custom annotation

cccclai:
Differential Revision: D78142146

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/llama.py
+++ b/examples/qualcomm/oss_scripts/llama/llama.py
@@ -659,21 +659,7 @@ def compile(args, pte_filename, tokenizer):
                 tokenizer=tokenizer,
                 custom_annotations=custom_annotations,
             )
-            # If hybrid and lookahead mode, we store kv output quant_attrs and apply to prefill output quant_attrs later
-            if i == 0 and args.model_mode in ["hybrid", "lookahead"]:
-                output_indices = 0
-                for node in llama_instance.llama_graph_module.graph.nodes:
-                    if node.op == "output":
-                        for output in node.args[0]:
-                            kv_quant_attrs[output_indices] = output.args[1:]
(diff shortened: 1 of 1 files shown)


cccclai:
@shewu-quic for the graph discrepancy, do you mean changes like this? Is the quant attrs for kv cache stored somewhere else, or as part of the model?

shewu-quic:
> @shewu-quic for the graph discrepancy, do you mean changes like this? Is the quant attrs for kv cache stored somewhere else, or as part of the model?

They are as part of the model. If you would like to remove them, it is better to add a check the quant_attrs is the same both prefill model and kv model.
```
for node in llama_instance.llama_graph_module.graph.nodes:
                    if node.op == "output":
                        for output in node.args[0]:
                            kv_quant_attrs[output_indices] = output.args[1:]
                            output_indices += 1
                        break
```

cccclai:
> > @shewu-quic for the graph discrepancy, do you mean changes like this? Is the quant attrs for kv cache stored somewhere else, or as part of the model?
> 
> They are as part of the model. If you would like to remove them, it is better to add a check the quant_attrs is the same both prefill model and kv model.
> 
> ```
> for node in llama_instance.llama_graph_module.graph.nodes:
>                     if node.op == "output":
>                         for output in node.args[0]:
>                             kv_quant_attrs[output_indices] = output.args[1:]
>                             output_indices += 1
>                         break
> ```

I see, is it like compare llama_instance[0] and llama_instance[1] for the value above? The code snippets seems not comparing things

shewu-quic:
> 
> I see, is it like compare llama_instance[0] and llama_instance[1] for the value above? The code snippets seems not comparing things

Yes, we didn't compare them since we forcedly set the same value.

cccclai:
Closing for now. Can re-open if needed

PR #12390, opened 2025-07-11
Arm backend: Add function to return quant params for lowered graph

wwwind:
Summary:
Add function to return quant params for lowered graph and remove these Q/DQ from the graph. If they are needed, then the EdgeProgramManager should be copied before use of this function.

Change-Id: I09de39c603d68d5ac5de4614a35eb7e3fc9ba518

### DIFF
--- a/exir/backend/io_quant_params.py
+++ b/exir/backend/io_quant_params.py
@@ -1,0 +1,100 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, Sequence
+
+import torch.fx as fx
+from executorch.exir import EdgeProgramManager
+from executorch.exir.passes.quantize_io_pass import QuantizeInputs, QuantizeOutputs
+
+
+def extract_io_quant_params(
+    edge_prog: EdgeProgramManager,
+    *,
+    input_idxs: Sequence[int] = (0,),
+    output_idxs: Sequence[int] = (0,),
+) -> Dict[str, Dict[str, Dict[str, Any]]]:
+    """
+    Returns quantization parameters such as scale/zero_point:
+      {
+        "inputs": {
+          <placeholder_name>: {"scale": float, "zero_point": int}
+        },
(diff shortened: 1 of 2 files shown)


exir/backend/io_quant_params.py
```diff
@@ -0,0 +1,100 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, Sequence
+
+import torch.fx as fx
+from executorch.exir import EdgeProgramManager
+from executorch.exir.passes.quantize_io_pass import QuantizeInputs, QuantizeOutputs
+
+
+def extract_io_quant_params(
```
digantdesai, line 13: Perhaps move this to the quantize_io_pass.py?

exir/backend/io_quant_params.py
```diff
@@ -0,0 +1,100 @@
...
+    edge_prog: EdgeProgramManager,
+    *,
+    input_idxs: Sequence[int] = (0,),
+    output_idxs: Sequence[int] = (0,),
+) -> Dict[str, Dict[str, Dict[str, Any]]]:
+    """
+    Returns quantization parameters such as scale/zero_point:
```
digantdesai, line 20: can't we get these after `quantize_io_pass` and then the config methods it adds?

digantdesai (commented):
Perhaps I am missing something, can you help me understand the motivation please?

wwwind:
Thank you for the review! @digantdesai 

We need this function in our workflow: we have a graphic use case when we need to move out Q/DQ nodes and get scale/zero points so we prepare input data in our plugin. Then we pass it to the subgraph. Currently, we need to call these two passes and then to extract these data from config, which is not very friendly for our users. Here there is just one function call that does this job.

exir/backend/io_quant_params.py
digantdesai: nvm you are doing just that :)

digantdesai:
Sound good. Can you please move this util in the pass file? Thanks.

digantdesai (approved):

wwwind:
Replaced with https://github.com/pytorch/executorch/pull/12481

PR #12393, opened 2025-07-11
Update examples/models/llama/README.md to fix build broken

DamonFool:
Fix the following error
```
CMake Error at tools/cmake/common/preset.cmake:122 (message):
  Use of 'EXECUTORCH_BUILD_EXTENSION_MODULE' requires
  'EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR'
  tools/cmake/preset/default.cmake:300 (check_required_options_on)
  CMakeLists.txt:87 (include)
```


lucylq:
Hi @DamonFool, thanks for submitting this PR. I don't think llama build should require `flat_tensor` - is the repro command exactly from the readme/could you provide instructions for your repro?

DamonFool:
> Hi @DamonFool, thanks for submitting this PR. I don't think llama build should require `flat_tensor` - is the repro command exactly from the readme/could you provide instructions for your repro?

Thanks @lucylq for taking a look at this.
Yes, I just followed the steps in the readme and copied commands in it.

The logic is that llama requires EXECUTORCH_BUILD_EXTENSION_MODULE, which requires EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR.

The cmake error msg has showed that dependency clearly.
You can verify it by testing the commands it the readme.
Thanks.

DamonFool:
> I don't think llama build should require `flat_tensor`

Hi @lucylq , After https://github.com/pytorch/executorch/pull/11457 , `flat_tensor` is required.
Please also see https://github.com/pytorch/executorch/blob/main/tools/cmake/preset/default.cmake#L305 .
Thanks.

lucylq:
> > I don't think llama build should require `flat_tensor`
> 
> Hi @lucylq , After #11457 , `flat_tensor` is required. Please also see https://github.com/pytorch/executorch/blob/main/tools/cmake/preset/default.cmake#L305 . Thanks.

Hi @DamonFool , thank you for the PR link! I added https://github.com/pytorch/executorch/pull/12514 to remove FLAT_TENSOR when it's not needed. I think that was added in to ensure all the builds work when we were adding cmake presets; it should be removed now that things are more stable.

lucylq (approved):
Thanks for making this PR @DamonFool. 

I was mistaken, and flat_tensor is required as a dependency of extension_module.

DamonFool:
Thanks @lucylq .
The 5 cicd failures should not be related to this patch since this is only a doc change, right?

DamonFool:
Hi @lucylq , are you fine to merge this pr?
Thanks.

DamonFool:
Thanks @lucylq .

PR #12395, opened 2025-07-11
Arm backend: Keep order of TOSA arguments

per:
Keep None arguments in TOSA arguments creation as well, as they are in Node arguments.

### Test plan
Tested through normal CI flow.


digantdesai (approved):

AdrianLundell:
Looks like a genuine linting error so let's wait with merging this until @per is back from vacation.

per:
Yes, dropping this one.

PR #12398, opened 2025-07-11
Add new GA models (next set)

mergennachin:
Add next set of GA models

Summary:

On top of https://github.com/pytorch/executorch/pull/12340

Add a few more tasks:

1. Image-Text Understanding (OpenCLIP)
2. Semantic Text Search (Sentence Transformers)
3. Document Q&A (DistilBERT QA)
4. Super resolution (Real-ESRGAN)
5. Audio Classification (AST)
6. Text Sentiment Analysis (RoBERTa)
7. Depth estimation (Depth Anything 2)

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -317,8 +317,9 @@ elif [[ "${BACKEND}" == *"xnnpack"* ]]; then
   echo "Testing ${MODEL_NAME} with xnnpack..."
   WITH_QUANTIZATION=true
   WITH_DELEGATION=true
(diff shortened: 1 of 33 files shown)


digantdesai:
@mergennachin  - As I was going through this list, I realized most of them are from HF transformers. So,
(1) if we should continue adding more HF transformer based models on optimum-et, instead of here.
(2) As I looked on optimum-et[1], I saw many already there i.e. Albert, DistillBert, EfficientNet, Roberta, Swin. Even if we don't do (1) - we should deduplicate unless you have preference for keeping them in ET CI for some reason.

[1] https://github.com/huggingface/optimum-executorch/tree/main/tests/models

PR #12399, opened 2025-07-11
PAL File for Arm BareMetal Executor Runner

BujSet:
### Summary
Platform file for building Arm bare metal executorch applications.

### Test plan
Tested against the `arm_executor_runner.cpp` that is used in the [Arm-EthosU tutorial](https://docs.pytorch.org/executorch/main/tutorial-arm-ethos-u.html). Was able to successfully build the Arm Executor Runner, and run the simple add model (TODO).

### DIFF
--- a/examples/arm/executor_runner/arm_executor_runner.cpp
+++ b/examples/arm/executor_runner/arm_executor_runner.cpp
@@ -14,6 +14,7 @@
 #include <executorch/runtime/platform/log.h>
 #include <executorch/runtime/platform/platform.h>
 #include <executorch/runtime/platform/runtime.h>
+#include <executorch/runtime/platform/defualt/arm_baremetal.hpp>
 #include <stdio.h>
 #include <unistd.h>
 #include <memory>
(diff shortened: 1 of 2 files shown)


BujSet:
@digantdesai @zingo Could I get some advice on this PR? I tried making the changes to `examples/arm/executor_runner/arm_executor_runner.cpp` to use the new file `runtime/platform/default/arm_baremetal.hpp`, but when I run the `setup.sh` and `run.sh` scripts, I'm met with this error:

```
/home/zephyruser/executorch/examples/arm/executor_runner/arm_executor_runner.cpp:17:10: fatal error: executorch/runtime/platform/defualt/arm_baremetal.hpp: No such file or directory
   17 | #include <executorch/runtime/platform/defualt/arm_baremetal.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Despite the new header file being moved to the runtime dir in the ET repo, it seems like the `setup.sh` and `run.sh` scripts are unable to see this when they try to build executorch. Any ideas why?

zingo:
Lets close this and reconcire down the line. The Bare metal implemntation is kind if HW specific so might not be a good fit and better to have this in the example code so the user can tailor it more easy. But it its also nice to get example down some maybe... So lets save this for later.

PR #12409, opened 2025-07-11
Return Result to fail gracefully instead of aborting

lucylq:
Summary:
- Make TensorInfo static factory pattern to allow returning Result.
- Return Result type to fail gracefully instead of directly aborting in the runtime.

Differential Revision: D78187751

### DIFF
--- a/runtime/executor/method_meta.cpp
+++ b/runtime/executor/method_meta.cpp
@@ -52,7 +52,7 @@ Result<Tag> get_tag(
   }
 }
 
-size_t calculate_nbytes(
+Result<size_t> calculate_nbytes(
     Span<const int32_t> sizes,
     executorch::aten::ScalarType scalar_type) {
   size_t n = 1;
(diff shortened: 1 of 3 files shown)


runtime/executor/method_meta.cpp
```diff
@@ -52,7 +52,7 @@ Result<Tag> get_tag(
   }
 }
 
-size_t calculate_nbytes(
+Result<size_t> calculate_nbytes(
```
JacobSzwejbka, line 55: You probably have to leave the old one and mark it deprecated. It would be good to do it for every thing in TensorInfo that doesnt return result

lucylq: TensorInfo constructor is private (only MethodMeta can create it). Should be OK?

lucylq: I think this is the only one in TensorInfo. The rest of the functions are getters that don't error-check.

JacobSzwejbka (approved):

PR #12424, opened 2025-07-13
Add JNI utility for exporting tensor map as PTD buffer

georgehong:
### Summary
Enables `TrainingModule` workflows using _.pte_ and _.ptd_ files to export the latest weights. Also makes adjustments for serialize.h utility to be visible for JNI. 
```
// TrainingModule.java
public static ByteBuffer savePtd(Map<String, Tensor> tensorMap) 
```

### Test plan
```
sh scripts/build_android_library.sh
sh executorch_android/android_test_setup.sh
./gradlew :executorch_android:connectedAndroidTest
```


extension/android/jni/jni_layer_training.cpp
```diff
@@ -210,6 +211,37 @@ class ExecuTorchTrainingJni
     return gradients;
   }
 
+  facebook::jni::local_ref<facebook::jni::JByteBuffer> exportWeights(
+      facebook::jni::alias_ref<jstring> methodName) {
```
JacobSzwejbka: I think this should take the map not the methodName

georgehong: Addressed feedback in follow-up commit which updates `exportWeights` to be `savePtd`, accepting general tensor maps, such as those including the gradients.

extension/android/executorch_android/src/main/java/org/pytorch/executorch/TrainingModule.java
```diff
@@ -116,4 +117,17 @@ public Map<String, Tensor> namedGradients(String methodName) {
 
   @DoNotStrip
   private native Map<String, Tensor> namedGradientsNative(String methodName);
+
+  /**
+   * Exports a map of tensors as a buffer that can be saved as a PTD file.
+   *
+   * @param tensorMap map of tensors to export to buffer.
```
JacobSzwejbka: Maybe mention that the key is the tensor name

georgehong:
Address comments

PR #12425, opened 2025-07-13
Build XNNPACK as an ExternalProject

swolchok:
I believe this is necessary to support EXPORT. (We probably should not
land this unless we have a PR stacked on top of this that successfully
converts the build to EXPORT; sending this out so I can get CI going)


swolchok:
* __->__ #12425

swolchok:
Apple build is failing because it can't find libXNNPACK.a: https://github.com/pytorch/executorch/actions/runs/16251337270/job/45881437516?pr=12425#step:9:47386

android build is failing because I didn't propagate cross compilation settings to the ExternalProject: https://github.com/pytorch/executorch/actions/runs/16251337288/job/45881440178?pr=12425#step:15:16927

llama runner linux build is failing because CMake is passing -lXNNPACK instead of the path to the static lib. (this happened during debugging before and it was because the XNNPACK target wasn't defined; probably the same again?) https://github.com/pytorch/executorch/actions/runs/16251337288/job/45881439862?pr=12425

unittest failures don't *look* related but we'll want to keep an eye on them.

swolchok:
I want to see EXPORT work locally before I come back and fix broader platform support, so I'm going to do that first. Converting to draft in the meantime.

swolchok:
build_frameworks_ios error:
```
2025-07-15T00:24:13.6245540Z Merging libraries:
2025-07-15T00:24:13.6245660Z libXNNPACK.a
2025-07-15T00:24:13.6245770Z libxnnpack_backend.a
2025-07-15T00:24:13.6245920Z libxnnpack-microkernels-prod.a
2025-07-15T00:24:13.6246060Z from ios/Release
2025-07-15T00:24:13.6246360Z into library /Users/runner/work/executorch/executorch/pytorch/executorch/cmake-out/libbackend_xnnpack_ios.a
2025-07-15T00:24:13.6246770Z File ios/Release/libXNNPACK.a does not exist
```

swolchok:
test-llava-runner-linux failure looks like a flake: https://github.com/pytorch/executorch/actions/runs/16301129189/job/46035741014?pr=12425#step:15:15293 . therefore publishing for review

backends/xnnpack/CMakeLists.txt
```diff
@@ -92,9 +92,8 @@ target_include_directories(
                            ${EXECUTORCH_ROOT}/third-party/flatbuffers/include
 )
 
-set(xnnpack_third_party pthreadpool extension_threadpool cpuinfo)
-
 include(cmake/Dependencies.cmake)
+set(xnnpack_third_party XNNPACK pthreadpool extension_threadpool cpuinfo)
```
mcr229, line 96: curious what does this rearrangment do?

backends/xnnpack/cmake/Dependencies.cmake
```diff
@@ -10,71 +10,83 @@ set(THIRD_PARTY_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/third-party")
...
-include_directories(SYSTEM ${XNNPACK_INCLUDE_DIR})
-list(APPEND xnnpack_third_party XNNPACK)
-install(TARGETS xnnpack-microkernels-prod
-    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+add_dependencies(xnnpack-microkernels-prod XNNPACKExternalProject)
 
+set_target_properties(
+  XNNPACK PROPERTIES INTERFACE_LINK_LIBRARIES xnnpack-microkernels-prod
```
mcr229, line 79: oh shoot does this mean that any targets that link XNNPACK will also automatically link xnnpack-microkernels-prod as well? So we don't have to explicitly list both XNNPACK and xnnpack-microkernels-prod? I was trying to figure out how to do this for a while, but never figured it out.

mcr229 (approved):
changes in the XNNPACK cmake/dependencies.cmake look good to me

backends/xnnpack/cmake/Dependencies.cmake
```diff
@@ -10,71 +10,83 @@ set(THIRD_PARTY_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/third-party")
...
+)
+
+install(DIRECTORY ${XNNPACK_INSTALL_DIR}/ DESTINATION ${CMAKE_INSTALL_PREFIX})
 
 if(EXECUTORCH_XNNPACK_ENABLE_KLEIDI)
-    install(TARGETS kleidiai
-        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+  add_library(kleidiai SHARED IMPORTED)
```
mcr229, line 85: i think i merged something that might have conflict here. I just added a check 

if (TARGET kleidiai)

before adding the library

swolchok: should be fine I think

backends/xnnpack/cmake/Dependencies.cmake
swolchok: target_link_libraries with visibility PUBLIC or INTERFACE does this for normal libraries, but you need to do it this way for IMPORTED libraries.

swolchok:
bunch of flakes that need to be rerun to make sure but no actual problems in the 7 failures so far

swolchok:
Apparently FetchContent is preferred over ExternalProject. The thing I wanted from ExternalProject is separating our build from the dependency's build (i.e., *not* using add_subdirectory), because the XNNPACK dependency ecosystem doesn't support EXPORT. IIUC, we can get that separation using FetchContent by configuring it to support find_package (using either `FIND_PACKAGE_ARGS` when we don't require specific changes to or a bleeding-edge version of the dependency, or `OVERRIDE_FIND_PACKAGE` when we do have such a requirement; both will result in later find_package calls hitting the FetchContent version of the dep, it's just that `FIND_PACKAGE_ARGS` might find the thing installed system-wide) and then using find_package.

backends/xnnpack/CMakeLists.txt
swolchok: I just found it weird that we were mutating xnnpack_third_party in cmake/Dependencies.cmake rather than setting it all at once

PR #12427, opened 2025-07-14
Adding more explicit error code

yunquan812:
Summary: {title}, the `Internal` and `InvalidArgument` errors don't tell what is really going wrong

Differential Revision: D78233596

### DIFF
--- a/runtime/core/error.h
+++ b/runtime/core/error.h
@@ -63,6 +63,12 @@ enum class Error : error_code_t {
   /// Operator(s) missing in the operator registry.
   OperatorMissing = 0x14,
 
+  /// Registration error: Exceeding the maximum number of kernels.
+  RegistrationExceedingMaxKernels = 0x15,
+
+  /// Registration error: The kernel is already registered.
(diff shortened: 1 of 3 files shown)


runtime/core/error.h
```diff
@@ -96,6 +96,12 @@ enum class Error : error_code_t {
   /// Execute stage: The handle is invalid.
   DelegateInvalidHandle = 0x32,
 
+
+  /*
+   * Registration errors.
+   */
+  RegistrationExceedingMaxKernels = 0x40,
```
swolchok: I don't know what our convention is around numbering here but I imagine we should just use the next number. @mergennachin do you know?

JacobSzwejbka: They are loosely grouped by type with jumps to allow new types in. If anything the original OperatorMissing is in the wrong spot, but since its there already probably just make this error 0x15 @yunquan812

yunquan812: Agreed. Let me update it :)

swolchok (approved):

PR #12441, opened 2025-07-14
[ET-VK] Fix caching mechanism to account for included files

SS-JIA:
* #12444
* #12443
* #12442
* __->__ #12441

## Changes

* Update the `gen_vulkan_spv.py` script to account for changes to included files when decided whether to perform a re-compilation

## Additional Changes

Also applied some misc fixes:

* Allow the Python preprocessor to handle unescaped `\` used to extend macros to multiple lines. Fixed this by replacing these with `\` before running the preprocessor.
* Fixed an issue where build would unexpectedly pass when trying to recompile a failing build multiple time. This would cause the caching mechanism to use a old cached build artifact since no changes were detected. Fixed this by removing any existing cached artifacts on unsuccessful build.

## Context

The `gen_vulkan_spv.py` script which handles GLSL shader codegen and GLSL -> SPIR-V compilation for the ExecuTorch Vulkan backend has a caching mechanism to only recompile modified files for the purpose of developer efficiency. However, this mechanism currently doesn't consider whether included files have been changed. Therefore, if a include file is modified without first modifying the source file which uses it, the changes to the include file will not be captured.

Differential Revision: [D78275585](https://our.internmc.facebook.com/intern/diff/D78275585/)

### DIFF
--- a/backends/vulkan/runtime/gen_vulkan_spv.py
+++ b/backends/vulkan/runtime/gen_vulkan_spv.py
@@ -545,6 +545,9 @@ def escape(line: str) -> str:
 def preprocess(
     input_text: str, variables: Dict[str, Any], input_path: str = "codegen"
 ) -> str:
+    # Workaround to handle source files using \ to extend mecros to a new line
+    input_text = re.sub(r"\\$", r"\\\\", input_text, flags=re.MULTILINE)
+
     input_lines = input_text.splitlines()
     python_lines = []
 
@@ -654,8 +657,8 @@ class SPVGenerator:
         for src_path in src_dir_paths:
             # Collect glsl source files
             src_files_list = glob.glob(
-                os.path.join(src_path, "**", "*.glsl*"), recursive=True
-            )
+                os.path.join(src_path, "**", "*.[gh]lsl*"), recursive=True
+            ) + glob.glob(os.path.join(src_path, "**", "*.h"), recursive=True)
             for file in src_files_list:
                 if len(file) > 1:
                     self.src_files[extract_filename(file, keep_ext=False)] = file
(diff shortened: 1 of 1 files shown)


trviv (approved):

backends/vulkan/runtime/gen_vulkan_spv.py
```diff
@@ -545,6 +545,9 @@ def escape(line: str) -> str:
 def preprocess(
     input_text: str, variables: Dict[str, Any], input_path: str = "codegen"
 ) -> str:
+    # Workaround to handle source files using \ to extend mecros to a new line
```
digantdesai, line 548:
```suggestion
    # Workaround to handle source files using \ to extend macros to a new line
```

backends/vulkan/runtime/gen_vulkan_spv.py
```diff
@@ -654,8 +657,8 @@ def addSrcAndYamlFiles(self, src_dir_paths: List[str]) -> None:
         for src_path in src_dir_paths:
             # Collect glsl source files
             src_files_list = glob.glob(
-                os.path.join(src_path, "**", "*.glsl*"), recursive=True
-            )
+                os.path.join(src_path, "**", "*.[gh]lsl*"), recursive=True
```
digantdesai, line 660: Update the comment to include `hlsl`. Not sure if you intended to put this change in this PR.

backends/vulkan/runtime/gen_vulkan_spv.py
```diff
@@ -955,43 +1048,45 @@ def compile_spirv(shader_paths_pair):
...
-
-        # Save all source GLSL files to the cache. Only do this at the very end since
-        # multiple variants may use the same source file.
-        if cache_dir is not None:
-            for _, src_file_fullpath in self.src_files.items():
-                cached_src_file = os.path.join(
-                    cache_dir, os.path.basename(src_file_fullpath) + ".t"
-                )
-                shutil.copyfile(src_file_fullpath, cached_src_file)
+                print(spv_to_glsl_map)
```
digantdesai, line 1086:
```suggestion
```

PR #12444, opened 2025-07-14
[ET-VK][qlinear] Faster weight only quantized linear gemv kernel

SS-JIA:
* __->__ #12444

## Changes

* Introduce a new compute shader for int4 linear's gemv cases that performs much better than the existing shader. This shader is inspired from MNN's gemv_1x1_conv_buf.cl shader.

With this compute kernel, transformer models' text generation can execute much faster than before.

On Samsung Galaxy S24 for Llama 3.2 1B, generating 128 tokens:

Before: ~25 tok/s
After: ~49 tok/s

## Why this new shader is faster

The biggest reason is due to vectorized loading of the uint4 weight buffer. This new shader loads the weight buffer as a buffer/image of `uvec4`, whereas the old shader loads the weight buffer as a buffer/image of `u8vec4`. Using the Adreno Offline Compiler, I found that in the former, only one load instruction was used to load from the weight tensor, whereas in the latter 16 load instructions were used to load from the weight tensor. It appears that the data loading was not being vectorized at the assembly level. This is potentially behaviour that can be approved in the SPIR-V shader compiler.

An additional factor is better weight packing layout. The new prepacking routine results in better memory coalescing between threads in a work group.

The final major factor is the use of tree based reduction to co-operatively reduce partial results into the final output. Previously, a single thread was responsible for the final reduction.

## Future Work

* Introduce faster shader for int4 linear gemm cases
* Update QCSNW to also use these updated shaders

Differential Revision: [D78275584](https://our.internmc.facebook.com/intern/diff/D78275584/)


backends/vulkan/runtime/graph/ops/glsl/linear_qga4w_coop.glsl
```diff
@@ -13,187 +13,143 @@
...
-#define NWORKERS 8
+#define WGS ${WGS}
 
 ${define_required_extensions(DTYPE)}
-$if WEIGHT_STORAGE == "buffer":
-  ${define_required_extensions("uint8")}
+${define_required_extensions("uint8")}
 
 #extension GL_EXT_control_flow_attributes : require
+#extension GL_EXT_debug_printf : require
```
digantdesai, line 22: remove this?

backends/vulkan/runtime/graph/ops/glsl/linear_qga4w_coop.glsl
```diff
@@ -13,187 +13,143 @@
...
+void main() {
+  const uint lid = gl_LocalInvocationID.x;
+  const uint n8 = gl_GlobalInvocationID.y;
+  // The output tensor will have a shape of [n, 1, 1, 1]. Each thread computes
+  // 8 output elements, so each thread will write to 8 elements starting at the
+  // tensor index (gid.x * 8, 0, 0, 0).
+  const uint n = MUL_8(n8);
+  const uint K4 = DIV_UP_4(input_sizes.x);
+
+  const int block_num = input_sizes.x / group_size;
```
digantdesai: assert (!(input_sizes.x % group_size))?

backends/vulkan/runtime/graph/ops/glsl/linear_qga4w_coop.glsl
```diff
@@ -13,187 +13,143 @@
...
+    const uint group_idx = k / group_size;
+
+    VEC4_T scales[2];
+    VEC4_T zeros[2];
+
+    // Only update the scales/zeros if the current iteration is now working on a
+    // new quantization group.
+    if (group_idx != cur_group_idx) {
+      // The qparams tensor contains the quantization scales and zeros, with
+      // shape [2, N, K / group_size, 1].
```
digantdesai, line 84: is this accurate? Seems like 2 and 1 are flipped? i.e. for a given group, both qparams are next to each other, right?

backends/vulkan/runtime/graph/ops/glsl/pack_int4_linear_weight_transposed_block_4x8.glsl
```diff
@@ -0,0 +1,154 @@
+/*
```
digantdesai, line 1: Curious why do we pack at runtime for Vulkan?

backends/vulkan/runtime/graph/ops/impl/QuantizedLinearQGANW.cpp
```diff
@@ -155,13 +162,19 @@ void add_linear_qga4w_node(
   check_linear_qga4w_args(
       graph, mat1, mat2_data, group_size, scales_and_zeros_data, out);
 
+  bool is_gemv = should_use_coop_algorithm(&graph, mat1);
   const uint32_t group_size_val = graph.extract_scalar<uint32_t>(group_size);
 
-  ValueRef mat2 =
-      prepack_int4_linear_weight_transposed_interleaved(graph, mat2_data);
+  ValueRef mat2 = is_gemv
```
digantdesai, line 168: I would imagine we will change this condition once we introduce gemm and use block_4x8 weight packing for both shaders?

backends/vulkan/runtime/graph/ops/glsl/qlinear_utils.glslh
```diff
@@ -0,0 +1,70 @@
...
+#ifdef BUFFER_WEIGHT
+
+uvec4 load_transposed_weight_block(const uint k4, const uint n8, const uint K4) {
+  return t_qmat2[n8 * K4 + k4];
+}
+
+#else // TEXTURE_WEIGHT
+
+uvec4 load_transposed_weight_block(const uint k4, const uint n8, const uint K4) {
+  return texelFetch(t_qmat2, ivec2(k4, n8), 0);
```
digantdesai, line 28: so this load was slow and wasn't vectorizing?

backends/vulkan/runtime/graph/ops/glsl/qlinear_utils.glslh
```diff
@@ -0,0 +1,70 @@
...
+
+#endif // BUFFER_WEIGHT
+
+/***********************************
+ * Packed weight data extraction functions
+ */
+
+float extract_4bit_from_transposed_block(const uvec4 block, const uint col, const uint row) {
+  return float(int((block[row] >> (4 * (7 - col))) & 15) - 8);
+}
```
digantdesai, line 39: Any way to optimize this extracting multiple values to or something?

backends/vulkan/runtime/graph/ops/impl/QuantizedLinearQGANW.cpp
```diff
@@ -155,13 +162,19 @@ void add_linear_qga4w_node(
...
+      ? prepack_int4_linear_weight_transposed_block_4x8(graph, mat2_data)
```
digantdesai, line 169: packing shader tests?

digantdesai (approved):
LGTM. Stamping. :)

PR #12480, opened 2025-07-15
 Add initial backends/cadence/vision module scaffold with optimized softmax kernel (no iDMA)

cad-rlc:
#### Release notes: backends
1. Created backends/cadence/vision module scaffold to which we plan to add optimized kernels for Vision 130.
2. Added an optimized softmax kernel without iDMA to vision/third_party.

### Test plan
Add -DEXECUTORCH_VISION_OPT=ON to enable vision optimized kernels while building the cadence cmake. Rest of the steps remain the same.

### DIFF
--- a/backends/arm/test/test_arm_baremetal.sh
+++ b/backends/arm/test/test_arm_baremetal.sh
@@ -155,17 +155,18 @@ test_pytest_ethosu_fvp() { # Same as test_pytest but also sometime verify using 
 
 
 test_pytest_ops_vkml() { # Same as test_pytest but also sometime verify using VKML runtime
-    echo "${TEST_SUITE_NAME}: Run pytest with VKML"
+    echo "${TEST_SUITE_NAME}: Run pytest operator tests with VKML runtime"
 
     backends/arm/scripts/build_executorch.sh
     backends/arm/test/setup_testing_vkml.sh
 
-    pytest  --verbose --color=yes --numprocesses=auto --durations=10  backends/arm/test/ --ignore=backends/arm/test/models
+    pytest  --verbose --color=yes --numprocesses=auto --durations=10  backends/arm/test/ \
+            --ignore=backends/arm/test/models -k _vgf_
     echo "${TEST_SUITE_NAME}: PASS"
 }
 
 test_pytest_models_vkml() { # Same as test_pytest but also sometime verify VKML runtime
-    echo "${TEST_SUITE_NAME}: Run pytest with VKML"
+    echo "${TEST_SUITE_NAME}: Run pytest model tests with VKML runtime"
 
     backends/arm/scripts/build_executorch.sh
     backends/arm/test/setup_testing_vkml.sh
@@ -173,7 +174,7 @@ test_pytest_models_vkml() { # Same as test_pytest but also sometime verify VKML 
     # Install model dependencies for pytest
     source backends/arm/scripts/install_models_for_test.sh
 
-    pytest  --verbose --color=yes --numprocesses=auto --durations=0 backends/arm/test/models
+    pytest  --verbose --color=yes --numprocesses=auto --durations=0 backends/arm/test/models -k _vgf_
     echo "${TEST_SUITE_NAME}: PASS"
 }
 


backends/cadence/aot/functions_vision.yaml
```diff
@@ -0,0 +1,265 @@
...
+
+- func: cadence::quantized_relu.out(Tensor X, Tensor X_zero_point, int out_zero_point, Tensor out_multiplier, Tensor out_shift, *, Tensor(a!) out) -> Tensor(a!)
+  kernels:
+    - arg_meta: null
+      kernel_name: cadence::impl::vision::quantized_relu_out
+
+- func: cadence::quantized_relu.per_tensor_out(Tensor X, int X_zero_point, int out_zero_point, int out_multiplier, int out_shift, *, Tensor(a!) out) -> Tensor(a!)
+  kernels:
+    - arg_meta: null
+      kernel_name: cadence::impl::vision::quantized_relu_per_tensor_out
```
mcremon-meta: I think using
```
kernel_name: cadence::impl::reference::quantized_relu_per_tensor_out
```
should work (picked relu at random, applied to all other ops too)

backends/cadence/vision/third-party/library/api/vsoftmaxf.c
```diff
@@ -0,0 +1,241 @@
...
+        xb_vecN_2xf32 gf, zout;
+        xb_vecN_2x32v xin_i, fr, exp, t;
+        xb_vecN_2x32v y, y1, y2, c1, c2, f2;
+        xb_vecN_2x64w w;
+        xin_i = IVP_TRUNCN_2XF32(x, 24);
+        /* Multiply by 1/ln2, extract the integer and fractional (Q32)
+         * components.     */
+        /* Q54 <- Q24*Q30 */
+        w = IVP_MULN_2X32(xin_i, invln2_Q30);
+        exp = IVP_PACKVNRN_2X64W(w, 54);
```
zonglinpeng: IVP_PACKVNRN_2X64W is not included in this change

backends/cadence/vision/operators/op_softmax.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+using executorch::aten::ScalarType;
+using executorch::aten::Tensor;
+using executorch::runtime::KernelRuntimeContext;
+using torch::executor::Error;
+
+namespace cadence {
+namespace impl {
+namespace vision {
+namespace native {
```
zonglinpeng: Please make all the namespaces to be 
```
namespace cadence {
namespace impl {
namespace MV130 {
namespace native {
```

backends/cadence/vision/operators/op_full.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/kernels/portable/cpu/scalar_utils.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+namespace torch {
```
zonglinpeng: Please use the consistent namespace
```
namespace cadence {
namespace impl {
namespace MV130 {
namespace native {
```

backends/cadence/vision/third-party/include/api.h
```diff
@@ -0,0 +1,83 @@
+/* ------------------------------------------------------------------------ */
```
zonglinpeng: It's not scalable and maintainable when the third-party has subdirectories, which make it suitable to be a standalone GitHub repo like HiFi (https://github.com/foss-xtensa/nnlib-hifi4) and G3 (https://github.com/foss-xtensa/nnlib-FusionG3/). Please make the nnlib a repo before we adding much more kernels/APIs

zonglinpeng (commented):
This PR is good but requires some additional coding to fix bugs, improve on infra, and make it compatible with meta internal build systems. Please address the issues in this PR:
1. **apply changes on top of https://github.com/pytorch/executorch/pull/12864 to avoid merge conflicts.**
2. convert the namespace to MV130
3. include the missing API library(s)

Please also make sure to move the third-party nnlib to a independent GitHub repo per suggested in the comment. This can be done in a follow up PR

backends/cadence/vision/third-party/library/api/vsoftmaxf.c
cad-rlc: Hi,

IVP_PACKVNRN_2X64W is an ISA instruction on Cadence Vision 130. It should be available in xt_ivpn.h which has been included in the common.h header file. Could you please elaborate on what issues were observed at your end?

Thanks

zonglinpeng: @cad-rlc We use buck internally and the xt linker is picking up the `IVP_PACKVNRN_2X64W` as `vsoftmaxf.c:132: undefined reference to `IVP_PACKVNRN_2X64W'`. I wonder
1. If anything from `xtensa/tie` should be picked up, as `<xtensa/tie/xt_ivpn.h>` is one of them.
2. Should any `xtensa/tie` be included in this PR, if yes to 1.
3. Any compiler flags I need to feed specifically for your vision kernel, if not to 1, such as disable `COMPILER_XTENSA` flag

cad-rlc:
@zonglinpeng @mcremon-meta Latest changes are now available. Please review.

backends/cadence/vision/kernels/kernels.cpp
```diff
@@ -0,0 +1,200 @@
...
+  return temp_mem_res.ok() ? temp_mem_res.get() : nullptr;
+}
+
+// Quantize a fp32 value to an int8_t/uint8_t value
+template <typename T>
+T quantize(const float x, float scale, int32_t zero_point) {
+  constexpr float min_val = std::numeric_limits<T>::min();
+  constexpr float max_val = std::numeric_limits<T>::max();
+  float tmp = roundf(x * scale + zero_point);
+  return std::max(std::min(tmp, max_val), min_val);
```
mcremon-meta: if the kernels don't use anything specific to vision chips, we don't need that file at all and we can just import from the reference kernels

backends/cadence/vision/operators/op_softmax.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include <stdio.h>
+
+using executorch::aten::ScalarType;
+using executorch::aten::Tensor;
+using executorch::runtime::KernelRuntimeContext;
+using torch::executor::Error;
+
+namespace cadence {
+namespace impl {
+namespace vision {
```
mcremon-meta: is there a compatibility between V110 and V130 that we can use? If yes, we can use a `vision` namespace for ops that will be compatible with both. If not, let's make them `V130` since it's probably the accurate definition

mcremon-meta (commented):
overall comment is that we shouldn't be duplicating that many files IMO, virtually all of this is not vision specific and we should be able to import/compile/link from the reference/portable ops. Let's talk about that in our sync

cad-rlc:
> overall comment is that we shouldn't be duplicating that many files IMO, virtually all of this is not vision specific and we should be able to import/compile/link from the reference/portable ops. Let's talk about that in our sync

Hi @mcremon-meta,

We understand that duplicating files may not the most clean way of setting up the vision directory but we chose to proceed with said approach to allow us make incremental changes to the now existing files, rather than having to add new files for optimized/vision specific implementations. This provides us with two benifits:
    1) Allows us to use the reference code as fall back within the vision namespace, even when the optimized versions have been integrated.
    2) Prevents unexpected failures and crashes, especially while building and running the quantized models. 
We hope that this is okay with you.

Thanks

mcremon-meta (approved):

zonglinpeng:
merged in https://github.com/pytorch/executorch/pull/14398

PR #12481, opened 2025-07-15
Arm backend: Add function extract_io_quant_params

wwwind:
Summary:
Add function to return quant params for lowered graph and remove these Q/DQ from the graph. If they are needed, then the EdgeProgramManager should be copied before use of this function.

This is a replacement of PR https://github.com/pytorch/executorch/pull/12390

Change-Id: I74d7292252c6daf1486ef28bbb2dcffdd34cd7b7


Sebastian-Larsson:
@digantdesai Please take a look at this PR since it touches files outside of the arm backend

digantdesai (approved):
Assuming not much changed since https://github.com/pytorch/executorch/pull/12390

Besides the code move to the pass file?

wwwind:
@digantdesai yes, nothing changed - this is the only change

Sebastian-Larsson (approved):

PR #12483, opened 2025-07-15
Arm backend: Add atanh decomposition pass

tbergkvist:
Decomposes atanh into other operators/lookup table for MI/BI case.


Sebastian-Larsson (approved):
CI failures unrelated

PR #12487, opened 2025-07-15
Allowing an error handler from caller

yunquan812:
Summary: To give a chance to the caller to decide how to handle these errors, e.g., sending to logcat, or writing extra diagnostics events, etc.

Differential Revision: D78232763

### DIFF
--- a/runtime/kernel/operator_registry.cpp
+++ b/runtime/kernel/operator_registry.cpp
@@ -49,7 +49,7 @@ Kernel* registered_kernels = reinterpret_cast<Kernel*>(registered_kernels_data);
 size_t num_registered_kernels = 0;
 
 // Registers the kernels, but may return an error.
-Error register_kernels_internal(const Span<const Kernel> kernels) {
+Error register_kernels_internal(const Span<const Kernel> kernels, ErrorHandler errorHandler) {
   // Operator registration happens in static initialization time before or after
   // PAL init, so call it here. It is safe to call multiple times.
   ::et_pal_init();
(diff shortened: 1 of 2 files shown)


runtime/kernel/operator_registry.cpp
```diff
@@ -49,7 +49,7 @@ Kernel* registered_kernels = reinterpret_cast<Kernel*>(registered_kernels_data);
 size_t num_registered_kernels = 0;
 
 // Registers the kernels, but may return an error.
-Error register_kernels_internal(const Span<const Kernel> kernels) {
+Error register_kernels_internal(const Span<const Kernel> kernels, ErrorHandler errorHandler) {
```
JacobSzwejbka, line 52: Why is the return value not sufficient?

runtime/kernel/operator_registry.cpp
```diff
@@ -88,24 +95,33 @@ Error register_kernels_internal(const Span<const Kernel> kernels) {
...
     registered_kernels[num_registered_kernels++] = kernel;
   }
-  ET_LOG(
-      Debug,
-      "Successfully registered all kernels from shared library: %s",
-      lib_name);
 
-  return Error::Ok;
+  if (errorHandler != nullptr) {
+    err =  errorHandler(err);
```
larryliu0820, line 107: This makes me a bit nervous. Why do we want to consume an `Error` and translate it to another `Error`? In practice, are you going to add product specific `Error` like `LoggingFailure` etc?

larryliu0820: I'm not against an error handler, just not sure what are we doing with the returned error here

runtime/kernel/operator_registry.cpp
JacobSzwejbka: I guess what situation exists where you can pass in the errorHandler to operate on the error generated by this function, but you couldnt just operate on it after its returned?

runtime/kernel/operator_registry.cpp
larryliu0820: A pattern I would definitely against is to convert `Error::RegistrationAlreadyRegistered` into `Error::Ok` if that's what you are trying to do here

runtime/kernel/operator_registry.cpp
yunquan812: If all errors are returned, then we can just operate on it after it is returned, but there are 2 errors (`RegistrationExceedingMaxKernels` and `RegistrationAlreadyRegistered`) are not returned (see this [line](https://github.com/pytorch/executorch/blob/main/runtime/kernel/operator_registry.cpp#L111)). If the caller wants to handle these 2 errors, the caller needs to specify the `errorHandler` to get a chance to handle these errors.

PR #12468, opened 2025-07-14
Convert pthreadpool to ExternalProject

swolchok:
As with the previous PR #12425, I think this is needed to get EXPORT
to work.

PR #12505, opened 2025-07-15
Allow ciflow/binaries label to trigger wheel builds on PRs

swolchok:
Following PR (#12425) broke Linux manywheel and there was no way to repro it on that PR; I just happened to catch it because the PR after that (#12468) triggered the manywheel build.


swolchok:
* #8954
* #12513
* #12468
* #12425
* __->__ #12505

jathu (approved):

PR #12514, opened 2025-07-15
Disable FLAT_TENSOR when not used

lucylq:
See: https://github.com/pytorch/executorch/pull/12393

With the cmake refactor/migration to presets (see https://github.com/pytorch/executorch/pull/11457), FLAT_TENSOR was enabled wherever EXTENSION_MODULE was enabled. It's not required for most targets (should just be training and tests for now), so removing it and the rule that EXTENSION_MODULE depends on it.

Rebase on: https://github.com/pytorch/executorch/pull/12544


swolchok:
looks like lots of tests are saying TRAINING requires FLAT_TENSOR? https://github.com/pytorch/executorch/actions/runs/16303931818/job/46045288574#step:15:12001

swolchok (changes requested):
back to your queue; since it's a build change and it breaks builds I can't just stamp it

lucylq:
Rebase on: https://github.com/pytorch/executorch/pull/12544, which removes pybind preset from linux preset.

Previously seeing issues where:
- pybind preset turns FLAT_TENSOR on for training.
- llm preset turns FLAT_TENSOR off as it's not needed.

lucylq:
Abandoning this PR. flat_tensor is required by extension_module for the program-data separation features we added: https://github.com/pytorch/executorch/blob/988435dfe7a8114ab93ae7140deba9b50b97f8d4/extension/module/CMakeLists.txt#L30-L33

PR #12526, opened 2025-07-15
Remove custom annotation for kv attrs

cccclai:
Differential Revision: D78375468

Remove the custom annotation, and copy the prefill kv quant attrs to decode kv quant attrs

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/llama.py
+++ b/examples/qualcomm/oss_scripts/llama/llama.py
@@ -484,6 +484,8 @@ class SingleLlama:
     def get_quant_attrs(self):
         return self.quant_attrs
 
+def rtol(a, b):
+    return abs(a - b) / max(abs(a), abs(b))
 
 def compile(args, pte_filename, tokenizer):
     os.makedirs(args.artifact, exist_ok=True)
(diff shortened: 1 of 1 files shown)


examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -651,33 +653,37 @@ def permute(w, heads):
...
                 "get_quant_io_dtype_fn"
             ] = partial(llama_instance._tag_ios, fixed_point_type=fixed_point_type)
+        
+        # Check prefill and decode graph is the same:
+        assert len(llama_instance_list[0].llama_graph_module.graph.nodes) == len(llama_instance_list[1].llama_graph_module.graph.nodes)
+        for prefill_node, decode_node in zip(llama_instance_list[0].llama_graph_module.graph.nodes, llama_instance_list[1].llama_graph_module.graph.nodes):
+            assert prefill_node.op == decode_node.op
+            assert prefill_node.target == decode_node.target
+                
+        # Copy prefill quant_attrs to kv output quant_attrs
```
winskuo-quic, line 676: Thanks for the PR. I am just curious if you observed better accuracy when copying prefill to decoder rather than the original flow, which is copying decode to prefill?

cccclai: I think accuracy will be on par with the previous implementation, however, with this PR, we can QAT not only weight, but also activation and other ops. The accuracy will be a lot better. (Maybe we only calibrate prefill and copy the quant attrs to decode to save the time for ptq too)

PR #12532, opened 2025-07-16
CoreML partitioner improvements

metascroy:
This creates several improvements for the CoreML partitioner:

* We now log when nodes are skipped for partitioning and the reason they are skipped.
* We add new partitioner option lower_full_graph that raises Exception if the model cannot be fully delegated.  In future, when this option is enabled we plan to support additional CoreML features like enumerated shapes.
* We add take_over_constant_data option to tell CoreML delegate to not consume weight data.


metascroy:
@JacobSzwejbka  this adds take_over_constant_data to partitioner.  If you don't want CoreML to consume the weights, you can set it to False.

YifanShenSZ (approved):
Great! 🚀 These were the pain points we've been having during development, nice to see them enhanced!

cccclai (approved):

PR #12536, opened 2025-07-16
Revert "Revert "Qualcomm AI Engine Direct - CI for Non-LLM GA model" …

winskuo-quic:
…(#12166)"

This reverts commit 95548266ebe43cba0d7f333010b781c572e5cc80.

### Summary

### Test plan

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -532,7 +532,7 @@ jobs:
       docker-image: ci-image:executorch-ubuntu-22.04-qnn-sdk
       submodules: 'recursive'
       ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 90
+      timeout: 900
       script: |
         # The generic Linux job chooses to use base env, not the one setup by the image
         CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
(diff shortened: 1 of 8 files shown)


winskuo-quic:
Hi @cccclai,
Sorry for the late reland on: https://github.com/pytorch/executorch/pull/12166
I have been working on finding the root cause of timeout for certain models in this PR: https://github.com/pytorch/executorch/pull/12191.
With the same config, it took only 1 minute to lower the model on my workstation. That is why I think higher timeout won't help resolving the issue.
In this PR with newest mainline, the CI seems to be working fine again.
However, I still couldn't find the reason that is causing the previous hang.
By any chance, can you see additional logs causing the hang on your side for this PR? https://github.com/pytorch/executorch/pull/12191
I think if the root cause can be found, we can better prevent timeout from happening again in the future.
The model that is hanging previously was `deit` and `dit`.

.github/workflows/trunk.yml
```diff
@@ -501,7 +501,7 @@ jobs:
       docker-image: executorch-ubuntu-22.04-qnn-sdk
       submodules: 'recursive'
       ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 900
+      timeout: 90
```
cccclai: Hey sorry miss this PR. I thought I left comment. Is there any particular reason that we lower the timeout? I think the CI machine is not very powerful, and I'm worried that it will cause timeout again, and will block updating the strict branch. Should we remove this change and see what happens?

cccclai:
After discussion offline, we will remove the timeout model and only enable the pass models for now

.github/workflows/trunk.yml
winskuo-quic: I have reverted it back to 900.

cccclai (approved):
Thanks

PR #12540, opened 2025-07-16
Arm backend: Fix logging info in decomposition pass for asin

emmakujala:
Logging info about asin being approximated was placed before the if-statement in call_operator resulting in it being triggered for all ops, not just asin.


Sebastian-Larsson (approved):
Unrelated CI failures

PR #12541, opened 2025-07-16
Arm backend: Add sub tensor to match scalar tensors to input data types.

SaoirseARM:
### Summary

Updates existing pass to ensure that x1, x2 are of the same type.
This comes from an issue where scalars are not automatically casted to the input dtype unlike mul and add operators.


Sebastian-Larsson (approved):
Unrelated CI error

PR #12544, opened 2025-07-16
Make OS presets only build library

jathu:
### Summary

I think mixing these different motivations (i.e. pybind vs. ET lib) can causes issues in the future with conflicting flags. For example, pybind always enables `EXECUTORCH_ENABLE_LOGGING` regardless of build mode (https://github.com/pytorch/executorch/pull/12469). So, I think we should just make the platform presets build ET targeting that platform.

### Test plan

CI

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -8,7 +8,7 @@
     },
     {
       "name": "macos",
-      "displayName": "Build everything buildable on macOS",
+      "displayName": "Build ExecuTorch for macOS",
       "inherits": ["common"],
       "generator": "Xcode",
       "cacheVariables": {
@@ -25,7 +25,7 @@
     },
     {
       "name": "ios",
-      "displayName": "Build everything buildable on iOS",
+      "displayName": "Build ExecuTorch for iOS",
       "inherits": ["common"],
       "generator": "Xcode",
       "cacheVariables": {
@@ -42,7 +42,7 @@
     },
     {
       "name": "ios-simulator",
-      "displayName": "Build everything buildable on iOS simulator",
+      "displayName": "Build ExecuTorch for iOS Simulator",
       "inherits": ["common"],
       "generator": "Xcode",
       "cacheVariables": {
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -59,7 +59,7 @@
     },
     {
       "name": "linux",
-      "displayName": "Build everything buildable on Linux",
+      "displayName": "Build ExecuTorch for Linux",
```
larryliu0820, line 62: Change the Zephyr one in line 110?

tools/cmake/preset/linux.cmake
```diff
@@ -4,7 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-include(${PROJECT_SOURCE_DIR}/tools/cmake/preset/pybind.cmake)
 include(${PROJECT_SOURCE_DIR}/tools/cmake/preset/llm.cmake)
 
 set_overridable_option(EXECUTORCH_BUILD_EXECUTOR_RUNNER ON)
+set_overridable_option(EXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL ON)
```
larryliu0820, line 10: slightly tangential but can we remove this option

larryliu0820: merge it with executor_runner one

jathu: Let's do it in a follow up PR. It's used in a lot of places right now

shoumikhin (approved):

tools/cmake/preset/linux.cmake
```diff
@@ -4,7 +4,8 @@
...
+set_overridable_option(EXECUTORCH_BUILD_EXTENSION_TRAINING ON)
```
lucylq: Why is training required here? :0

jathu: oops, let me remove it

PR #12550, opened 2025-07-16
Partition to_dim_order_copy op in XNN delegate

leafs1:
### Summary
This PR adds support for the to_dim_order_copy operation in the XNNPACK delegate partitioner, enabling direct handling of memory format conversions initiated by users via .to(memory_format=) calls. This enhancement significantly improves performance by producing more compressed graphs that avoid unnecessary partitioning boundaries at memory format conversion points. By delegating these operations directly to XNNPACK, we eliminate the overhead of context switching between the runtime and delegate, reducing both execution time and memory footprint. The implementation leverages XNNPACK's highly optimized memory format conversion routines, which are specifically designed for efficient tensor layout transformations on various hardware targets.

### Test plan
Confirmed expected output when having user specified dim order conversions as well as appropriate partitioning. I did this by writing individual tests for the to_copy op ensuring it changes dim order and dtype when appropriate. Also added test module to confirm that the to copy nodes are partitioned and not in another partition


mcr229 (approved):
let's wait until the other fix lands first

PR #12554, opened 2025-07-16
Fix internal tests affected by D77980055

masnesral:
Summary: Looks like the change to how full_like is decomposed means these tests need some tweaks

Differential Revision: D77981005

### DIFF
--- a/backends/arm/test/passes/test_decompose_cosine_similarity_pass.py
+++ b/backends/arm/test/passes/test_decompose_cosine_similarity_pass.py
@@ -34,7 +34,8 @@ def test_decompose_cosine_similarity_tosa_BI(module):
         "executorch_exir_dialects_edge__ops_aten_mul_Tensor": 5,
         "executorch_exir_dialects_edge__ops_aten_sum_dim_IntList": 3,
         "executorch_exir_dialects_edge__ops_aten_pow_Tensor_Scalar": 2,
-        "executorch_exir_dialects_edge__ops_aten_full_like_default": 1,
+        # TODO(masnesral): uncomment after https://github.com/pytorch/pytorch/pull/144765
+        # "executorch_exir_dialects_edge__ops_aten_full_default": 1,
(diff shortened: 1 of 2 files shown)


exir/tests/test_memory_planning.py
```diff
@@ -868,18 +868,19 @@ def forward(self, input, label):
...
 
-        # 147 just so happens to be the index of the user_grad output arg of
+        # 149 just so happens to be the index of the user_grad output arg of
         # convolution_backward.out. This is fairly fragile.
         # Check that the None output is not memory planned.
+        self.assertEqual(len(ep.executorch_program.execution_plan[0].values), 151)
         self.assertEqual(
             ep.executorch_program.execution_plan[0]
-            .values[147]
+            .values[149]
```
digantdesai: Not sure how this will pass w/o https://github.com/pytorch/pytorch/pull/144765. But if it does, then the test may not care about it in the first place :)

backends/arm/test/passes/test_decompose_cosine_similarity_pass.py
```diff
@@ -34,7 +34,7 @@ def test_decompose_cosine_similarity_tosa_BI(module):
         "executorch_exir_dialects_edge__ops_aten_mul_Tensor": 5,
         "executorch_exir_dialects_edge__ops_aten_sum_dim_IntList": 3,
         "executorch_exir_dialects_edge__ops_aten_pow_Tensor_Scalar": 2,
-        "executorch_exir_dialects_edge__ops_aten_full_like_default": 1,
+        "executorch_exir_dialects_edge__ops_aten_full_default": 1,
```
digantdesai: same here.

digantdesai (approved):
Stamping to unblock.

masnesral:
AFAICT the test failure is entirely unrelated; it would have to be since I'm just commenting out tests here.

PR #12556, opened 2025-07-16
Bump CMake requirement to 3.29

### LINKED ISSUE #12524: Update minimum CMake version to 3.29

We need to update the minimum CMake version to 3.29, as XNNPACK implicitly requires this.

From @mergennachin:
* If we want to upgrade 3.29, we should update across the board.
https://github.com/search?q=repo%3Apytorch%2Fexecutorch%20cmake_minimum_required&type=code Can you make that codemod PR?
* We need to change in requirements-dev.txt https://github.com/pytorch/executorch/blob/main/requirements-dev.txt#L3
* We need to change in pyproject.toml in https://github.com/pytorch/executorch/blob/main/pyproject.toml#L3

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @larryliu0820 @jathu

GregoryComer:
### Summary
Update the CMake version requirement to 3.29. This is in response to XNNPACK implicitly requiring it.

Note that it should be sufficient to update the dependency only at the top-level CMake (as opposed to all of the separate locations where the version in specified).

Fixes https://github.com/pytorch/executorch/issues/12524. See the issue for more context.

### Test plan
I've built locally to verify that it builds correctly. CI will also help validate. I also attempted to build with an older CMake version (3.28), and observed the following error:

```
CMake Error at CMakeLists.txt:50 (cmake_minimum_required):
  CMake 3.29 or higher is required.  You are running version 3.28.0
```

Previously, it would error out with:
```
CMake Error at CMakeLists.txt:18 (CMAKE_POLICY):
  Policy "CMP0156" is not known to this version of CMake.
```


GregoryComer:
The failures (test-eval_llama-mmlu-linux, unittest-nxp-neutron, build-benchmark-app, and test-llama-runner-mac) are pre-existing.

mergennachin (approved):
> as opposed to all of the separate locations where the version in specified

I think we should still do a follow-up step to do proper clean-up. But we probably don't need to cherry-pick that codemod/cleanup PR.

PR #12559, opened 2025-07-16
Check for null outputs in method

GregoryComer:
Summary: When non-memory-planned outputs are unset, method execution will crash when writing to the output tensor. This manifests as a native crash with a deep stack trace that is both unrecoverable and unclear to the user. This PR adds a check in method.execute to detect null output tensor data pointers, and, if found, log a message and return an error code.

Differential Revision: D78438716

### DIFF
--- a/runtime/executor/method.cpp
+++ b/runtime/executor/method.cpp
@@ -1540,6 +1540,25 @@ Error Method::execute() {
       "Cannot execute until method has been initialized.");
   ET_LOG(Debug, "Executing method: %s.", method_meta().name());
 
+  // Validate that outputs are set.
+  for (size_t i = 0; i < method_meta().num_outputs(); i++) {
+    auto& output = mutable_value(get_output_index(i));
+    if (!output.isTensor()) {
+      continue;
+    }
+
+    auto tensor_meta = method_meta().output_tensor_meta(i);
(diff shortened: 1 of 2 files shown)


runtime/executor/method.cpp
```diff
@@ -1540,6 +1540,25 @@ Error Method::execute() {
       "Cannot execute until method has been initialized.");
   ET_LOG(Debug, "Executing method: %s.", method_meta().name());
 
+  // Validate that outputs are set.
+  for (size_t i = 0; i < method_meta().num_outputs(); i++) {
+    auto& output = mutable_value(get_output_index(i));
```
JacobSzwejbka, line 1545: nit: dont need a mutable reference here

runtime/executor/method.cpp
```diff
@@ -1540,6 +1540,25 @@ Error Method::execute() {
       "Cannot execute until method has been initialized.");
   ET_LOG(Debug, "Executing method: %s.", method_meta().name());
 
+  // Validate that outputs are set.
+  for (size_t i = 0; i < method_meta().num_outputs(); i++) {
```
JacobSzwejbka, line 1544: Can we just cache that we've checked this so we dont do this every time

JacobSzwejbka (changes requested):
just the couple nits

PR #12564, opened 2025-07-16
Add path for to_edge_transform_and_lower that avoids  EDGE_DO_NOT_DECOMP namespace

metascroy:
The current design of using EDGE_DO_NOT_DECOMP to prevent decomposition has long standing issues, and often fails lowering when certain ops are requested for preservation.  This shows up most notably in the CoreML backend, where most ops are requested for preservation.

As a band-aid, we introduced _remove_invalid_ops_for_not_decompose to cover certain kinds of ops.  But when an op is encountered that we do not have an exception for, lowering still fails.

We also recently found another bug that shows up for SDPA related to contiguous views (https://fb.workplace.com/groups/pytorch.edge.users/permalink/1796069037930048/) that we still do not fully understand the root cause of.

EDGE_DO_NOT_DECOMP is actually only used to support the "check_op_support" argument in the partitioner; ops_to_not_decompose only modifies the default composition table.

In CoreML's case, "check_op_support" is not used, and the issues with EDGE_DO_NOT_DECOMP's design causes lots of lowering issues that are hard to keep up with.  This PR enables a new path that bypasses EDGE_DO_NOT_DECOMP's when possible (_can_skip_using_EDGE_DO_NOT_DECOMP).

Long term, we need to address the buggy design of EDGE_DO_NOT_DECOMP.  There are some ideas here: https://fb.workplace.com/groups/pytorch.edge2.team/permalink/1241898747065975/

### DIFF
--- a/exir/program/_program.py
+++ b/exir/program/_program.py
@@ -1082,6 +1082,27 @@ def _remove_invalid_ops_for_not_decompose(
     return list(filter(keep, ops_to_not_decompose))
 
 
+def _can_skip_using_EDGE_DO_NOT_DECOMP(
+    partitioner: Dict[str, List[Partitioner]], aten_programs: Dict[str, ExportedProgram]
+) -> bool:
+    # THe current design of using EDGE_DO_NOT_DECOMP to prevent decomposition
+    # has long standing issues.  _remove_invalid_ops_for_not_decompose was a band-aid to
+    # fix some of the issues, but more issues are coming up over time, including a new issue with SDPA
+    # and contiguous views: https://fb.workplace.com/groups/pytorch.edge.users/permalink/1796069037930048/
+    # EDGE_DO_NOT_DECOMP is only needed by partitioners that specify check_op_support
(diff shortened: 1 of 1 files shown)


JacobSzwejbka:
@larryliu0820  @lucylq  maybe relevant for the non delegated paths we have been talking about

mcr229 (approved):
yea this makes sense. sorry it has been so buggy. I've been meaning to look into it but haven't had the bw. 

I've been trying to rethink this a little bit. Right now i think the cases `check_op_support` isn't use that often altogether(there used to be some cases, but i think they were slowly resolved out). I wonder if we should drop it altogether, but then i worry about cases where an aten op is actually sometimes supported, and for best delegation the backend should try and delegate its decomposition.

metascroy:
> check_op_support

Yeah I don't know a good solution for check_op_support.  I think @larryliu0820's idea of moving to_backend before to_edge was to let backends control this at a finer granularity.

PR #12566, opened 2025-07-16
Turn BUILD_TESTING off if EXECUTORCH_BUILD_TESTS is off

swolchok:
As the comment notes, it looks like otherwise we have third-party deps that might include(CTest) and thereby turn on testing even though our specific option is off. (Should we get rid of the specific option or more directly alias it to BUILD_TESTING?)

PR #12562, opened 2025-07-16
Remove bundled_program dep from mpsdelegate

swolchok:
It doesn't seem to be necessary; let's see if tests pass.

PR #12561, opened 2025-07-16
Build FXDiv as ExternalProject

swolchok:

PR #12560, opened 2025-07-16
format even more CMakeLists.txt files

swolchok:

PR #12578, opened 2025-07-17
Don't use iOS toolchain on macOS

swolchok:
I don't know why we were doing this, but it seems wrong. Let's see if tests pass.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -12,7 +12,6 @@
       "inherits": ["common"],
       "generator": "Xcode",
       "cacheVariables": {
-        "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/third-party/ios-cmake/ios.toolchain.cmake",
         "EXECUTORCH_BUILD_PRESET_FILE": "${sourceDir}/tools/cmake/preset/macos.cmake",
         "PLATFORM": "MAC_ARM64",
         "DEPLOYMENT_TARGET": "12.0"
@@ -47,7 +46,7 @@
       "generator": "Xcode",
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/third-party/ios-cmake/ios.toolchain.cmake",
-        "EXECUTORCH_BUILD_PRESET_FILE": "${sourceDir}/tools/cmake/preset/ios.cmake",
+        "EXECUTORCH_BUILD_PREe SET_FILE": "${sourceDir}/tools/cmake/preset/ios.cmake",
         "PLATFORM": "SIMULATORARM64",
         "DEPLOYMENT_TARGET": "17.0"
       },


swolchok:
* #8954
* __->__ #12578
* #12566
* #12562
* #12561
* #12560
* #12513
* #12468
* #12425

swolchok:
might want to pick this to release, not sure. comes from https://github.com/pytorch/executorch/pull/10768

jathu:
I think ios.toolchain.cmake is probably poor naming, but it is a general Apple cmake toolchain to build for all of their platforms (controlled via PLATFORMS): https://github.com/leetal/ios-cmake

swolchok:
> I think ios.toolchain.cmake is probably poor naming, but it is a general Apple cmake toolchain to build for all of their platforms (controlled via PLATFORMS): https://github.com/leetal/ios-cmake

using this toolchain breaks sentencepiece's build in the next PR. Why aren't CMake's default settings on macOS sufficient?

jathu:
@shoumikhin do you have context on why we use the toolchain instead of relying the on the default host configs?

CMakePresets.json
```diff
@@ -12,7 +12,6 @@
       "inherits": ["common"],
       "generator": "Xcode",
       "cacheVariables": {
-        "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/third-party/ios-cmake/ios.toolchain.cmake",
```
shoumikhin, line 15: iirc, this is not iOS-only toolchain, we use it to build mac targets too.
It contains a bunch of useful declarations to simplify our live, e.g. the very `MAC_ARM64` platform definition.

swolchok: This is not the normal way that CMake projects expect to build for Mac though. concretely, sentencepiece's build does not like CMAKE_MACOSX_BUNDLE being set on non-iOS platforms, and it wouldn't be set on macOS if not for the use of this toolchain. IMO we should keep everything standard and defaulted unless we have a reason not to. Why do we need to use this to build for Mac?

swolchok:
build-frameworks-ios error:
```
2025-07-17T01:17:59.6638640Z Merging libraries:
2025-07-17T01:17:59.6638760Z libexecutorch.a
2025-07-17T01:17:59.6638860Z libexecutorch_core.a
2025-07-17T01:17:59.6638970Z libextension_apple.a
2025-07-17T01:17:59.6639080Z libextension_data_loader.a
2025-07-17T01:17:59.6639210Z libextension_flat_tensor.a
2025-07-17T01:17:59.6639330Z libextension_module.a
2025-07-17T01:17:59.6639440Z libextension_tensor.a
2025-07-17T01:17:59.6639560Z from simulator/Release
2025-07-17T01:17:59.6639880Z into library /Users/runner/work/executorch/executorch/pytorch/executorch/cmake-out/libexecutorch_simulator.a
2025-07-17T01:17:59.6640280Z File simulator/Release/libextension_apple.a does not exist
```

so I guess this script might need CMAKE_TOOLCHAIN_FILE put back?

CMakePresets.json
shoumikhin: So we rely on that toolchain platform definitions with the extra cmake flags they add to build frameworks for iOS and macOS. So far they worked well. If you believe there's a better approach, let's explore it too.
I guess we can disabled CMAKE_MACOSX_BUNDLE for sentencepiece specifically when building for mac, that's what we do with the custom script build phase in Xcode when building extension/llm third-party deps.
Screenshot 2025-07-16 at 18 40 40

swolchok: > I guess we can disabled CMAKE_MACOSX_BUNDLE for sentencepiece specifically when building for mac, that's what we do with the custom script build phase in Xcode when building extension/llm third-party deps.

I agree that we can work around this and it's straightforward to do so. My concern is that the setting of this toolchain cost me half an hour of debugging time to track down, because it was surprising that 1) any toolchain was active for macos, since builds work fine without it when not using presets, and 2) "ios.toolchain.cmake" in particular was active. This will presumably continue to be surprising to others in the future.

What if instead we set the CMAKE_TOOLCHAIN_FILE *only* in build_apple_frameworks.sh? Do we need it anywhere else? (Long-term I think we should investigate using CMake's built-in support for producing Apple frameworks, but I don't see a particular reason to prioritize that.)

shoumikhin: I don’t have a strong preference - so long as we can build the frameworks as usual, I’m happy either way.
Just a few quick notes:
- Despite its name, this toolchain isn’t limited to iOS.
- The last time I tried building the frameworks by manually setting all the macOS-specific flags myself - it ended up costing way more time 😅
- Otherwise, this toolchain has worked well so far.

Maybe just disable that particular flag like in https://github.com/pytorch/executorch/pull/12594

swolchok: Hmm, maybe https://github.com/pytorch/executorch/pull/12594 will unblock me. If not I'll see if I can find another way to unblock export since I've found out that this is here for a reason and not just a mistake.

swolchok: Yep, I rebased and it looks like we don't need this.

PR #12584, opened 2025-07-17
NXP backend: Fix unittest-nxp-neutron introduced in 0012ffaf9

robert-kalmar:
### Test plan
```
pytest -c /dev/null backends/nxp/tests/
```

cc @digantdesai @JakeStevens


JakeStevens (approved):
not sure how this file was missed, should be better going forward with moving the nxp tests

robert-kalmar:
> not sure how this file was missed, should be better going forward with moving the nxp tests

You mean moving from trunk to pull? That is already in place, and the error get triggered on the respective PR https://github.com/pytorch/executorch/pull/12546. But merged despite that.

PR #12599, opened 2025-07-17
Rename docker image to use ci-image:name syntax

clee2000:
This changes the docker image names from `executorch/name:commitsha` to `executorch/ci-image:name-commitsha`

I changed the docker image calculation step at some point to allow tags have names and still have the .ci/docker hash.  By changing the names from `name:commitsha` to `ci-image:name-commitsha`, it is no longer necessary to create new ecr repos on AWS whenever a new `name` is added, which only a few people can do.  Instead, everything will go under `ci-image` with tags

Testing: Run CI, if any job pull the old image, its bad.  Not sure how to check tho

Pros:
* no longer need to wait for one of the people with permissions to create new ecr repo

Cons:
* xkcd about standards
* not sure how often new image names need to be created (probably not often at this point)
* dont fix what isnt broken
* testing this change is probably not super thorough

### DIFF
--- a/.ci/docker/build.sh
+++ b/.ci/docker/build.sh
@@ -7,8 +7,10 @@
 
 set -exu
 
-IMAGE_NAME="$1"
+FULL_IMAGE_NAME="$1"
 shift
+
+IMAGE_NAME=$(echo "${FULL_IMAGE_NAME}" | sed 's/ci-image://')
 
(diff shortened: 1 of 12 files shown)


huydhn (approved):
LGTM! Thank you for the fix, now we don't to manually create new entries for ExecuTorch anymore

PR #12606, opened 2025-07-17
Add executorch_llm xcframework bundling extension/llm

shoumikhin:


scripts/build_apple_frameworks.sh
```diff
@@ -34,6 +34,29 @@ libextension_module.a,\
...
+libabsl_log_initialize.a,\
+libabsl_log_internal_format.a,\
+libabsl_log_internal_log_sink_set.a,\
+libabsl_log_internal_message.a,\
+libabsl_spinlock_wait.a,\
+libabsl_str_format_internal.a,\
+libabsl_strings.a,\
+libabsl_strings_internal.a,\
+libabsl_time.a,\
+libabsl_time_zone.a,\
```
larryliu0820: wow

larryliu0820: I'm wondering how many of them are actually used. I think we only uses `libabsl_strings.a` but not sure if they are inter-dependent.

larryliu0820 (approved):

scripts/build_apple_frameworks.sh
shoumikhin: And i'm afraid that's not the full list yet :)

PR #12615, opened 2025-07-17
Ignore error when removing temp directory in coreml_preprocess

limintang:
Differential Revision: D78517331

### DIFF
--- a/backends/apple/coreml/compiler/coreml_preprocess.py
+++ b/backends/apple/coreml/compiler/coreml_preprocess.py
@@ -365,7 +365,7 @@ class CoreMLBackend(BackendDetails):
 
         match model_type:
             case CoreMLBackend.MODEL_TYPE.COMPILED_MODEL:
-                shutil.rmtree(str(model_path.resolve()))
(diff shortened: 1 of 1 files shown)


billmguo (approved):

cccclai:
Why do we need this change?

backends/apple/coreml/compiler/coreml_preprocess.py
```diff
@@ -396,7 +396,7 @@ def preprocess_model(
                 for key, value in model_debug_info.debugSymbolToHandles.items()
             }
 
-        shutil.rmtree(str(dir_path.resolve()))
+        shutil.rmtree(str(dir_path.resolve()), ignore_errors=True)
```
metascroy, line 399: I think the change on line 399 is likely harmless, but I'd still like to understand why it's needed.  It might lead to build up of old directories.

The change on line 368 has me more concerned.  If the delete doesn't complete successfully, are there conflicts with the move on line 371.

I think we should revert this until we better understand what's going on @billmguo

PR #12623, opened 2025-07-18
Add support for aten.min.dim

cccclai:
Differential Revision: D78526966

### DIFF
--- a/backends/qualcomm/builders/op_min.py
+++ b/backends/qualcomm/builders/op_min.py
@@ -16,7 +16,7 @@ from .qnn_constants import OpElementWiseMinimum, QNN_OP_PACKAGE_NAME_QTI_AISW
 
 @register_node_visitor
 class Min(NodeVisitor):
-    target = ["aten.minimum.default"]
+    target = ["aten.minimum.default", "aten.min.dim"]
 
     def __init__(self, *args) -> None:
         super().__init__(*args)
--- a/backends/qualcomm/tests/models.py
+++ b/backends/qualcomm/tests/models.py
@@ -1162,6 +1162,15 @@ class Minimum(torch.nn.Module):
         return torch.minimum(x, y)
 
 
+class MinDim(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, logits):
+        min_logits, min_indices = torch.min(logits, dim=1)
+        return min_logits, min_indices
+
+
 class Mul(torch.nn.Module):
     def __init__(self):
         super().__init__()
--- a/backends/qualcomm/tests/test_qnn_delegate.py
+++ b/backends/qualcomm/tests/test_qnn_delegate.py
@@ -884,6 +884,11 @@ class TestQNNFloatingPointOperator(TestQNN):
         sample_input = (torch.randn(1, 2, 3, 4), torch.randn(2, 3, 4))
         self.lower_module_and_test_output(module, sample_input)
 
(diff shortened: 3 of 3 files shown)


cccclai:
@haowhsu-quic @shewu-quic @DannyYuyang-quic @winskuo-quic I ran into some issues to lower an internal model to QNN and this is one of the missing ops. I think the output are two nodes, while the regular min output is one node. Do you think we should re-use the `op_min` or have a different file to handle this case?

cccclai:
Actually, seems like the min operator in HTP doesn't support returning index, we may need to decompose `min.dim`, is it correct?

haowhsu-quic:
> Actually, seems like the min operator in HTP doesn't support returning index, we may need to decompose `min.dim`, is it correct?

Yes, looks like the most similar one is ReduceMin. But it cannot return index.
Maybe we can compose it with ReduceMin + ArgMin.

DannyYuyang-quic:
> Actually, seems like the min operator in HTP doesn't support returning index, we may need to decompose `min.dim`, is it correct?

Thanks for catching that, I’ll handle the min.dim op support

cccclai:
> > Actually, seems like the min operator in HTP doesn't support returning index, we may need to decompose `min.dim`, is it correct?
> 
> Thanks for catching that, I’ll handle the min.dim op support

Thanks! That will be super helpful. I find that we need to have 2 more ops...can we have the following ops supported?
```
                "aten.min.dim",
                "aten.logical_and.default",
                "aten.max.dim",
```

DannyYuyang-quic:
> > > Actually, seems like the min operator in HTP doesn't support returning index, we may need to decompose `min.dim`, is it correct?
> > 
> > 
> > Thanks for catching that, I’ll handle the min.dim op support
> 
> Thanks! That will be super helpful. I find that we need to have 2 more ops...can we have the following ops supported?
> 
> ```
>                 "aten.min.dim",
>                 "aten.logical_and.default",
>                 "aten.max.dim",
> ```

Yes sure, could you please also share the source(testcase) for `aten.logical_and.default`~?

cccclai:
Something like this will be helpful
```
class LogicalAnd(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x, y):
        return torch.logical_and(x, y)

a = torch.tensor([0, 1, 10, 0], dtype=torch.int8)
b = torch.tensor([4, 0, 1, 0], dtype=torch.int8)
model = LogicalAnd()
example_inputs = (a, b)
model(*example_inputs)
```

DannyYuyang-quic:
> Something like this will be helpful
> 
> ```
> class LogicalAnd(torch.nn.Module):
>     def __init__(self):
>         super().__init__()
> 
>     def forward(self, x, y):
>         return torch.logical_and(x, y)
> 
> a = torch.tensor([0, 1, 10, 0], dtype=torch.int8)
> b = torch.tensor([4, 0, 1, 0], dtype=torch.int8)
> model = LogicalAnd()
> example_inputs = (a, b)
> model(*example_inputs)
> ```

`aten.min.dim` and `aten.max.dim` are almost ready, I’ll push the PR once internal CI passes.

Regarding `logical_and`, it seems that int8 inputs are not supported,  the `ElementWiseAnd` only accepts UFIXED_POINT_8, SFIXED_POINT_8, or BOOL_8. Is int8 part of your internal model in FP mode?

cccclai:
merged with a different PR

PR #12624, opened 2025-07-18
[Backend Tester] Add backend filtering, improved test discovery

GregoryComer:
Iteratively update the backend test suite to support cleaner test flow registration, backend filtering, and a bit of misc incremental cleanup.

The test suite can now be invoked with `python -m executorch.backends.test.suite.runner operators --backend xnnpack`, for example.


GregoryComer:
* #12666
* #12658
* #12625
* __->__ #12624

backends/test/suite/__init__.py
```diff
@@ -44,22 +46,20 @@ def is_backend_enabled(backend):
         return backend in _ENABLED_BACKENDS
 
 
-ALL_TEST_FLOWS = []
+_ALL_TEST_FLOWS: Sequence[TestFlow] | None = None
```
digantdesai: Nit
```suggestion
_ALL_TEST_FLOWS: Sequence[TestFlow] = []
```

backends/test/suite/flow.py
```diff
@@ -0,0 +1,63 @@
...
+    """ The name of the lowering flow. """
+
+    backend: str
+    """ The name of the target backend. """
+
+    tester_factory: Callable[[], Tester]
+    """ A factory function that returns a Tester instance for this lowering flow. """
+
+
+def create_xnnpack_flow() -> TestFlow | None:
```
digantdesai, line 29:
```suggestion
	@classmethod
	def from_name(name: str) -> TestFlow | None:
		...
```

backends/test/suite/flow.py
```diff
@@ -0,0 +1,63 @@
...
+            backend="coreml",
+            tester_factory=CoreMLTester,
+        )
+    except Exception:
+        logger.info("Skipping Core ML flow registration due to import failure.")
+        return None
+
+
+def all_flows() -> Sequence[TestFlow]:
+    flows = [
```
digantdesai, line 58: Nit: dict(name: flow)?

digantdesai (approved):

backends/test/suite/__init__.py
GregoryComer: Updated.

backends/test/suite/flow.py
GregoryComer: Updated.

backends/test/suite/flow.py
GregoryComer: I slightly prefer the approach of having individual methods that can be explicitly called per flow, but I'll defer to your judgement if you feel strongly about it.

PR #12625, opened 2025-07-18
[Backend Tester] Add test name filter

GregoryComer:
Add a --filter argument to the test runner. Minor refactoring on the discovery logic.

You can now run `python -m executorch.backends.test.suite.runner operators --backend xnnpack --filter add`, for example.


GregoryComer:
* #12736
* #12733
* #12666
* #12658
* __->__ #12625

backends/test/suite/discovery.py
```diff
@@ -32,32 +45,37 @@ def discover_tests(
...
 
-def _is_test_enabled(test_case: unittest.TestCase, backends: set[str] | None) -> bool:
+def _is_test_enabled(test_case: unittest.TestCase, test_filter: TestFilter) -> bool:
     test_method = getattr(test_case, test_case._testMethodName)
+    flow: TestFlow = getattr(test_method, "_flow")
+
+    if test_filter.backends is not None and flow.backend not in test_filter.backends:
+        return False
+
+    if test_filter.name_regex is not None and not test_filter.name_regex.search(
```
digantdesai, line 76: Feels like we are reimplementing features from unittest or pytest :p

digantdesai (approved):

backends/test/suite/discovery.py
GregoryComer: Yeah, I've tried to avoid this, but unfortunately, it doesn't look like the unittest package structure is very extensible in the way I need. There aren't a lot of hooks to control reporting / filtering / discovery without writing custom driver code. I'm open to suggestions, but this seemed to be the lowest friction path with unittest. Switching to pytest might be an option, but I'm hoping that I don't need to do much more non-differentiated work like this.

PR #12626, opened 2025-07-18
Create CMakeLists.txt for extension/llm/apple

shoumikhin:


extension/llm/apple/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Please this file formatted by running:
+# ~~~
+# cmake-format -i CMakeLists.txt
+# ~~~
+
+cmake_minimum_required(VERSION 3.19)
```
larryliu0820, line 12: This should be aligned with root level CMakeLists.txt

shoumikhin: Atm it's aligned with https://github.com/pytorch/executorch/blob/main/extension/apple/CMakeLists.txt#L12
Do you suggest to upgrade both? In this change or in another?

larryliu0820 (approved):

extension/llm/apple/CMakeLists.txt
larryliu0820: Can be in another PR.

Issue #12306, opened 2025-07-09
Promote to_edge_with_preserved_ops to non-experimental status


mcremon-meta:
### 🚀 The feature, motivation and pitch

Cadence backends do not delegate, so they cannot use `to_edge_transform_and_lower`. In some cases we do want to preserve ops and avoid decomposing them (e.g. `aten.rms_norm`) only to fuse them back later. Pattern matching is annoying, brittle and hard to maintain, so we've relied on `to_edge_with_preserved_ops`. Now that the usage is more widespread, it would be great to promote that API to non-experimental.

cc @JacobSzwejbka @JakeStevens  @larryliu0820 @cbilgin

JacobSzwejbka:
cc @SS-JIA

larryliu0820:
We think that for it to become an official API, preserved ops should not include mutation or view ops, so the verifier should raise when it sees these ops in the `preserve_ops` argument to this API. 

There's also some confusion around when to use `_core_aten_ops_exception_list` and when to use `preserve_ops`. I think we have some homework to do to decompose the ops in `_core_aten_ops_exception_list` into core ATen, so backends won't rely on this "feature" and if they need the non-core ATen op, they should put them in `preserve_ops`. Ideally we should not have any `_core_aten_ops_exception_list` in the whole compiler stack.

mcremon-meta:
> We think that for it to become an official API, preserved ops should not include mutation or view ops, so the verifier should raise when it sees these ops in the `preserve_ops` argument to this API.

Sounds reasonable, hopefully this is a small-ish change to the verifier.

> There's also some confusion around when to use `_core_aten_ops_exception_list` and when to use `preserve_ops`. I think we have some homework to do to decompose the ops in `_core_aten_ops_exception_list` into core ATen, so backends won't rely on this "feature" and if they need the non-core ATen op, they should put them in `preserve_ops`. Ideally we should not have any `_core_aten_ops_exception_list` in the whole compiler stack.

I agree, I'm only concerned about the time it would require to fully make the `_core_aten_ops_exception_list` removable (we have multiple models relying on it now) through decompositions and available implementations.

GregoryComer:
[...]

PR #12629, opened 2025-07-18
[reland] Fix coreml to edge transform and lower

metascroy:
Re-land of: https://github.com/pytorch/executorch/pull/12564 

Previous attempt had conflict with https://github.com/pytorch/executorch/issues/12306 that caused CI failure.

------

The current design of using EDGE_DO_NOT_DECOMP to prevent decomposition has long standing issues, and often fails lowering when certain ops are requested for preservation. This shows up most notably in the CoreML backend, where most ops are requested for preservation.

As a band-aid, we introduced _remove_invalid_ops_for_not_decompose to cover certain kinds of ops. But when an op is encountered that we do not have an exception for, lowering still fails.

We also recently found another bug that shows up for SDPA related to contiguous views (https://fb.workplace.com/groups/pytorch.edge.users/permalink/1796069037930048/) that we still do not fully understand the root cause of.

EDGE_DO_NOT_DECOMP is actually only used to support the "check_op_support" argument in the partitioner; ops_to_not_decompose only modifies the default composition table.

In CoreML's case, "check_op_support" is not used, and the issues with EDGE_DO_NOT_DECOMP's design causes lots of lowering issues that are hard to keep up with. This PR enables a new path that bypasses EDGE_DO_NOT_DECOMP's when possible (_can_skip_using_EDGE_DO_NOT_DECOMP).

Long term, we need to address the buggy design of EDGE_DO_NOT_DECOMP. There are some ideas here: https://fb.workplace.com/groups/pytorch.edge2.team/permalink/1241898747065975/

cc @kimishpatel @YifanShenSZ @cymbalrush

### DIFF
--- a/exir/program/_program.py
+++ b/exir/program/_program.py
@@ -1098,22 +1098,6 @@ def _can_skip_using_EDGE_DO_NOT_DECOMP(
     return can_skip_using_EDGE_DO_NOT_DECOMP
 
 
-def _replace_view_with_view_copy(program: ExportedProgram) -> ExportedProgram:
-    program = program.run_decompositions({})
-    new_gm = ReplaceViewOpsWithViewCopyOpsPass()(program.graph_module).graph_module
-    program = ExportedProgram(
-        root=new_gm,
-        graph=new_gm.graph,
-        graph_signature=_get_updated_graph_signature(program.graph_signature, new_gm),
-        state_dict=program.state_dict,
-        range_constraints=program.range_constraints,
-        module_call_graph=program.module_call_graph,
-        example_inputs=program.example_inputs,
-        constants=program.constants,
-    )
-    return program
-
-
 def _gen_edge_manager_for_partitioners(
     partitioner: Dict[str, List[Partitioner]],
     aten_programs: Dict[str, ExportedProgram],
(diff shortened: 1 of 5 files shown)


exir/program/_program.py
```diff
@@ -1112,37 +1150,55 @@ def _gen_edge_manager_for_partitioners(
...
+                    or []
+                )
 
-        _restore_transformed_ops_to_aten_ops(program)
+        if not can_skip_using_EDGE_DO_NOT_DECOMP:
+            program = program.run_decompositions(_default_decomposition_table())
+            _restore_transformed_ops_to_aten_ops(program)
 
+        # Edge will complain if there are view ops requested for preservation, so we replace them with view_copy
+        program = _replace_view_with_view_copy(program)
```
lucylq: Do backends treat these the same at to_backend?

metascroy: They do not necessarily (CoreML does for many of them), but it's already applied in  _generate_edge_program after the verifier, but before to_backend is called, so this doesn't change the behavior that exists today.

Some other options are:
1. Always filter out view ops from preservation
2. Query curr_partitioner.ops_to_not_decompose for view op's copy variant support, and only preserve the view op if the copy variant is present.

metascroy: I can also remove this line.  It shouldn't be needed after you land D78535519

billmguo (approved):
can we test the model export and make sure last Friday issue not appear again with this diff?

derekxu (approved):
So this will fix the numerical lowering issues?

metascroy:
> So this will fix the numerical lowering issues?

When combined with  https://github.com/pytorch/executorch/pull/12665, I think it should (this switches to using torchao APIs for quantization)

metascroy:
> can we test the model export and make sure last Friday issue not appear again with this diff?

I will test to_edge_transform_and_lower on the ANE static model in the examples/apple/coreml/llama folder.

larryliu0820 (approved):

billmguo (approved):

PR #12638, opened 2025-07-18
Log info instead of error

lucylq:
Fix failing jobs caused by error logging from https://github.com/pytorch/executorch/pull/12295

Test failures: https://hud.pytorch.org/pr/pytorch/executorch/12631#46239138027

Note: logging errors doesn't cause an abort in the et runtime; seems like the arm test is being extra careful and parsing the log script for errors, which is causing this failure.

https://github.com/pytorch/executorch/blob/aab802135cb9b41927893fb5870f0dc43e5719f8/backends/arm/scripts/run_fvp.sh#L111-L117

### DIFF
--- a/runtime/executor/program.cpp
+++ b/runtime/executor/program.cpp
@@ -227,7 +227,7 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
     // For support, contact the PyTorch Edge team or make an issue in:
     // https://github.com/pytorch/executorch/issues.
     ET_LOG(
-        Error,
+        Info,
         "!!DEPRECATED!! This branch is deprecated from ExecuTorch 0.7; re-export this PTE file to ensure support on newer runtimes.");
     return Program(
         loader,


runtime/executor/program.cpp
```diff
@@ -227,7 +227,7 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
     // For support, contact the PyTorch Edge team or make an issue in:
     // https://github.com/pytorch/executorch/issues.
     ET_LOG(
-        Error,
+        Info,
```
digantdesai, line 230: So is our plan to keep this functional for a while more and then delete it?
Else we shouldn't merge this, if we really want to deprecate.

digantdesai (approved):

runtime/executor/program.cpp
lucylq: Yeah - it's currently deprecated AoT (since ET beta) so we should deprecate it in the runtime. There are a bunch of tests/users with older PTEs, and they will fail if this branch is removed now.

digantdesai: But if the `older` PTEs are outside of the BC window, then we should perhaps start nudging people to update their PRs. This is one way, pinging them with a deadline is another :p

PR #12641, opened 2025-07-18
Build XNNPACK via FetchContent and reference with find_package

swolchok:
XNNPACK doesn't support CMake EXPORT, but we can work around this by using its find_package configuration instead of add_subdirectory. FetchContent expedites this.


swolchok:
* #12643
* #12642
* __->__ #12641

swolchok:
I don't expect to need to commit this; it didn't work for the purpose I wanted it for. However, we might need to do it in the future to enable ExecuTorch to be packaged for distribution by a Linux distro or other package manager and get its dependencies from the rest of the system.

PR #12642, opened 2025-07-18
Build pthreadpool via FetchContent and reference with find_package

swolchok:
pthreadpool doesn't support CMake EXPORT, but we can work around this by using its find_package configuration instead of add_subdirectory. FetchContent expedites this.


swolchok:
* #12643
* __->__ #12642
* #12641

PR #12643, opened 2025-07-18
Build cpuinfo via FetchContent and reference with find_package

swolchok:
cpuinfo doesn't support CMake EXPORT, but we can work around this by using its find_package configuration instead of add_subdirectory. FetchContent expedites this.


swolchok:
* __->__ #12643
* #12642
* #12641

PR #12646, opened 2025-07-18
Add test when there are no segments

lucylq:
Summary:
The deprecation warning should only appear when there are values inside the constant_buffer and nothing inside the constant_segment.

Even when there are no constants in the program, we expect to have one 'segment' storing the placeholder for non-constants.

Fix failing jobs caused by error logging from #12295
Test failures: https://hud.pytorch.org/pr/pytorch/executorch/12631#46239138027

Arm test is using a recently-generated PTE, however hitting the deprecation warning because the pte has no constants.

Differential Revision: D78568839

### DIFF
--- a/runtime/executor/program.cpp
+++ b/runtime/executor/program.cpp
@@ -172,7 +172,18 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
   // only offset, the constant segment is empty and does not need to be loaded.
   const auto* constant_segment = flatbuffer_program->constant_segment();
   if (constant_segment != nullptr && constant_segment->offsets() != nullptr &&
-      constant_segment->offsets()->size() > 1) {
+      constant_segment->offsets()->size() > 0) {
+    if (constant_segment->offsets()->size() == 1) {
+      // No constants; the constant segment is empty and does not
+      // need to be loaded.
+      return Program(
+          loader,
+          segment_base_offset,
+          std::move(program_data.get()),
+          flatbuffer_program,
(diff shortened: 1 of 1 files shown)


runtime/executor/program.cpp
```diff
@@ -217,18 +217,21 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
...
+    if (constant_segment->offsets()->size() == 0) {
+      // NOTE: Storing constants in the constant_buffer is deprecated from
+      // ExecuTorch 0.7 onwards. Please regenerate your PTE file to ensure newer
+      // ExecuTorch runtimes can support it. ExecuTorch deprecation policy:
+      // https://docs.pytorch.org/executorch/stable/api-life-cycle.html#deprecation-policy.
+      // For support, contact the PyTorch Edge team or make an issue in:
+      // https://github.com/pytorch/executorch/issues.
+      ET_LOG(
+          Error,
+          "!!DEPRECATED!! This branch is deprecated from ExecuTorch 0.7; re-export this PTE file to ensure support on newer runtimes.");
```
digantdesai: Nit: can we explain a bit more in the error message, and perhaps how to mitigate it? Seems like one has to read code to get around this error.

digantdesai (approved):

runtime/executor/program.cpp
lucylq: I previously had a longer message, but this caused errors in internal tests where they have a limit of 128 bytes for the log string - see [D77965467](https://www.internalfb.com/diff/D77965467?entry_point=16)

PR #12652, opened 2025-07-18
Revert "Ignore error when removing temp directory in coreml_preprocess"

metascroy:
Reverts pytorch/executorch#12615

I want to better understand what is going on


cccclai (approved):
Yeah make sense to understand better before moving forward

guangy10 (approved):

PR #12663, opened 2025-07-19
Qualcomm AI Engine Direct - fix LPBQ implementation

haowhsu-quic:
### Summary
- fix LPBQ and make test case more general

### Test plan
```python
python backends/qualcomm/tests/test_qnn_delegate.py TestQNNQuantizedOperator.test_qnn_backend_conv2d_block -b build-android -s $DEVICE -m SM8750
```

### DIFF
--- a/backends/qualcomm/builders/node_visitor.py
+++ b/backends/qualcomm/builders/node_visitor.py
@@ -163,10 +163,10 @@ class NodeVisitor:
             max_scale = scales[ch].reshape(1, -1).amax(dim=-1) / num_steps
             q_scales = torch.clamp(
                 input=scales[ch] / max_scale,
-                min=torch.iinfo(quant_scales_dtype).min,
-                max=torch.iinfo(quant_scales_dtype).max,
+                min=1,
+                max=2**bitwidth_of_scale,
             ).to(quant_scales_dtype)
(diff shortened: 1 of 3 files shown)


haowhsu-quic:
Hi @cccclai, I made a mistake when implementing LPBQ. I've tested current version against Llama3.2 1B and can get better results now. Will update the related change on llama.py in another PR.

cccclai:
> Hi @cccclai, I made a mistake when implementing LPBQ. I've tested current version against Llama3.2 1B and can get better results now. Will update the related change on llama.py in another PR.

Thanks! Mind sharing a bit more details about the mistake? Trying to follow the code but not super clear

haowhsu-quic:
Yes, currently HTP uses 4bits to store quantized scales. I didn't clip the numeric to the right range (before:  0-255 / after: 1-16). I also double check with AIMET's implementation and the mse between per-channel / per-block.

cccclai:
> Yes, currently HTP uses 4bits to store quantized scales. I didn't clip the numeric to the right range (before: 0-255 / after: 1-16). I also double check with AIMET's implementation and the mse between per-channel / per-block.

Ah that makes sense..

cccclai (approved):
Thanks for the fix

PR #12664, opened 2025-07-21
Add CoreML support for torchao quantize_

metascroy:
This is CoreML support for blockwise/channelwise quantization of linear/embedding layers using torchao APIs.

This PR is based on top of https://github.com/pytorch/executorch/pull/12648/, and will be rebased once that lands to show the changes isolated to quantization.

PR #12648, opened 2025-07-18
Add ability to add MIL ops to ET that are not yet in coremltools

metascroy:
This introduces torch_ops.py in ET's CoreML compiler.  This lets us add critical MIL ops that are not yet in ET's version of coremltools.

PR #12665, opened 2025-07-21
Switch ANE llama model to use to_edge_transform_and_lower + torchao quantization APIs

metascroy:
This switches the ANE model to use to_edge_transform_and_lower and torchao quantization APIs.

To use to_edge_transform_and_lower, we first need to land: https://github.com/pytorch/executorch/pull/12629

To use torchao quant APIs, we first need to land https://github.com/pytorch/executorch/pull/12648/ and https://github.com/pytorch/executorch/pull/12664.

This PR contains all of the changes from those PRs because it is rebased on them.  I will rebase on main once those PRs land to make this easier to review.

### DIFF
--- a/.ci/scripts/test_ane_static_llama.sh
+++ b/.ci/scripts/test_ane_static_llama.sh
@@ -28,6 +28,6 @@ pushd $EXECUTORCH_ROOT/examples/apple/coreml/llama
 # Download stories llama110m artifacts
 download_stories_model_artifacts
 
-python export.py -n model.pte -p params.json -c stories110M.pt --seq_length 32 --max_seq_length 64 --dtype fp16 --coreml-quantize c4w
+python export.py -n model.pte -p params.json -c stories110M.pt --seq_length 32 --max_seq_length 64 --dtype fp16 --coreml-quantize c4w --embedding-quantize 4,32
 
 popd
--- a/examples/apple/coreml/llama/export.py
+++ b/examples/apple/coreml/llama/export.py
@@ -148,15 +148,6 @@ def main() -> None:
             lambda m, fqn: isinstance(m, torch.nn.Embedding),
         )
 
-    # CoreML's op_linear_quantizer_config appears to have a bug where the quantization
-    # quality is subpar.  We use torchao APIs instead, which are now supported by CoreML
-    op_linear_quantizer_config = None
-    #     op_linear_quantizer_config = {
-    #         "mode": "linear_symmetric",
-    #         "dtype": "int4",
-    #         "granularity": "per_channel",
-    #     }
-
     if export_args.coreml_quantize == "b4w":
         quantize_(
             model,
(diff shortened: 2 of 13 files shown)


billmguo (approved):

.ci/scripts/test_ane_static_llama.sh
```diff
@@ -28,6 +28,6 @@ pushd $EXECUTORCH_ROOT/examples/apple/coreml/llama
 # Download stories llama110m artifacts
 download_stories_model_artifacts
 
-python export.py -n model.pte -p params.json -c stories110M.pt --seq_length 32 --max_seq_length 64 --dtype fp16 --coreml-quantize c4w
+python export.py -n model.pte -p params.json -c stories110M.pt --seq_length 32 --max_seq_length 64 --dtype fp16 --coreml-quantize c4w --embedding-quantize 4,32
```
cccclai, line 31: Is the pr for supporting embedding quantization?

examples/apple/coreml/llama/export.py
```diff
@@ -18,18 +18,19 @@
...
-    EmbeddingQuantHandler,
-)
 
 from executorch.exir.backend.utils import format_delegated_graph
 from executorch.exir.capture._config import EdgeCompileConfig, ExecutorchBackendConfig
 from executorch.exir.passes import MemoryPlanningPass
 from executorch.exir.passes.quant_fusion_pass import QuantFusionPass
 from executorch.exir.passes.sym_shape_eval_pass import ConstraintBasedSymShapeEvalPass
-from executorch.exir.program._program import to_edge
+from executorch.exir.program._program import to_edge_transform_and_lower
```
cccclai: Should we do `from executorch.exir import to_edge_transform_and_lower`

examples/apple/coreml/llama/export.py
```diff
@@ -115,19 +116,8 @@ def main() -> None:
         export_args.dtype
     ]  # dtype for model/inputs
 
-    if export_args.embedding_quantize:
```
cccclai, line 118: Oh looks like it's just to switch to a different way...

examples/apple/coreml/llama/export.py
```diff
@@ -167,15 +173,11 @@ def main() -> None:
...
-        op_linear_quantizer_config=op_linear_quantizer_config,
     )
     partitioner = CoreMLPartitioner(  # pyre-fixme[16]
         compile_specs=compile_specs,
         take_over_mutable_buffer=False,
-        skip_ops_for_coreml_delegation=[
-            "quantized_decomposed.embedding_4bit.dtype",
-            "aten.embedding.default",
-        ],
+        skip_ops_for_coreml_delegation=[],
```
cccclai, line 180: oh wait, I remember they had the error, did you switch to run cpu 4bit embedding?

metascroy: By using torchao APIs, we get around the error.

.ci/scripts/test_ane_static_llama.sh
metascroy: The PR just changes to using to_edge_transform_and_lower and torchao APIs for quantize_.  It does not add embedding quant support.

Embedding quant support existed, but wasn't being tested before, so I also enabled it in the CI test.

examples/apple/coreml/llama/export.py
cccclai: Any chance you know the reason? I just remember it errors out, but not sure the reason

metascroy: The error is a compression conflict that happens during their optimization stack.

Since we don't use their optimization stack anymore, we don't have a conflict.

PR #12671, opened 2025-07-21
Arm backend: Grouped conv per-channel quant support

AdrianLundell:
Adds support for per-channel quantization of grouped convolution.


Sebastian-Larsson (approved):
Unrelated CI errors

PR #12672, opened 2025-07-21
Arm backend: Add aten.select to "view" operator support check

Sebastian-Larsson:
Select decomposes into squeeze, which in turn becomes a view. Therefore, perform the same check on select operators as view operators.


AdrianLundell (approved):

AdrianLundell:
Failures not related

PR #12673, opened 2025-07-21
Arm backend: Divide test_unary into test_floor and test_ceil

Sebastian-Larsson:
This makes it easier to parse which operators have had their lowering implemented. Additionally, it becomes easier to add test cases that are specific to the respective operator.


AdrianLundell (approved):

AdrianLundell:
Failures unrelated

PR #12677, opened 2025-07-21
[easy] Set -x in build_size_test.sh

swolchok:
Makes it easier to repro CI jobs by printing the commands that are run.


swolchok:
* __->__ #12677

PR #12678, opened 2025-07-21
Build EXTENSION_DATA_LOADER with size_test

swolchok:
size_test.cpp uses FileDataLoader. Looks like this happens to work currently because of the missing cmake_deps.toml entry that gets fixed in https://github.com/pytorch/executorch/pull/12744


swolchok:
* #8954
* #12744
* #12743
* #12691
* __->__ #12678

extension/llm/tokenizers
```diff
@@ -1 +1 @@
-Subproject commit ea8691c6200c702cf28dd2aa2da902e9299b6bb3
+Subproject commit f09feca15849a790c05b3b7855e7c62ce26ba94b
```
digantdesai: Intentional?

tools/cmake/preset/default.cmake
```diff
@@ -249,7 +249,7 @@ check_required_options_on(
 
 check_conflicting_options_on(
   IF_ON EXECUTORCH_BUILD_ARM_BAREMETAL CONFLICTS_WITH
-  EXECUTORCH_BUILD_EXTENSION_DATA_LOADER EXECUTORCH_BUILD_PTHREADPOOL
```
digantdesai, line 252: I suspect this or `install_executorch` used by the ci scripts was the reason it was working before? 

If we want to not enable this by default then perhaps we should also update `install_executorch` (if not using the default preset already). 

And that should also make the baremetal build fail. I also realized that we don't run the baremetal size-test binary on the CI, so we may need to fix that too.

extension/llm/tokenizers
swolchok: nope, good catch, should be in the EXPORT PR or split out

tools/cmake/preset/default.cmake
swolchok: > the reason it was working before?

you may be interested in later PRs in this stack, particularly #12744

swolchok:
build-demo-ios looks like a flake, noting CI looks good before removing the accidental tokenizers bump and starting merge of the stack

test/build_size_test.sh
```diff
@@ -23,11 +23,17 @@ cmake_install_executorch_lib() {
   update_tokenizers_git_submodule
   local EXTRA_BUILD_ARGS="${@}"
 
+  if [[ "$EXTRA_BUILD_ARGS" == *"-DEXECUTORCH_BUILD_ARM_BAREMETAL=ON"* ]]; then
+      local BUILD_DATA_LOADER="OFF"
```
larryliu0820, line 27: Nit: is double quote needed?

larryliu0820 (approved):

test/build_size_test.sh
swolchok: I don't think so, but I also don't think it hurts

swolchok:
merging based on previous good CI run

PR #12683, opened 2025-07-21
Check overflow in allocation

JacobSzwejbka:
Summary: Overflow is unsafe as a user will get back much less memory then expected

Differential Revision: D78678540


lucylq (approved):
linter?

Issue #12293, opened 2025-07-08
Top-level CMake targets


GregoryComer:
### 🚀 The feature, motivation and pitch

Implement single top-level targets.

Previously discussed in build design review and https://github.com/pytorch/executorch/issues/12138.

larryliu0820:
Let's make sure we can install these top level libraries

larryliu0820:
Also 100% we should have top level selective build API

GregoryComer:
Handling a few small follow-ups and will merge by EOD tomorrow.

GregoryComer:
Core targets are landed, I've added a few small follow-up tasks:

https://github.com/pytorch/executorch/issues/12948
https://github.com/pytorch/executorch/issues/12949

GregoryComer:
Closing this as core work is done, see follow-ups above.

PR #12689, opened 2025-07-21
Add top-level CMake backends target

GregoryComer:
Add a top level backends CMake target, which includes all configured backends. I validate this change by building a simple runner using the executorch_backends target and verifying that it was able to build and run an XNNPACK-delegated binary. Once the changes land in ExecuTorch, this will be long-term validated in executorch-examples CI.

To allow the executor runner target to use the executorch_backends target, I've also moved the Vulkan defs above the runner. This causes issues with the vulkan_executor_runner target, so I'm updating the main runner to include Vulkan and remove the Vulkan-specific one (subject to @SS-JIA's signoff).


GregoryComer:
* #12807
* #12806
* #12696
* __->__ #12689

larryliu0820 (approved):

GregoryComer:
Vulkan build failure in the unittest job is real. Fixing that and then this should be good to land.

mergennachin (approved):

PR #12692, opened 2025-07-21
Add unique method ID

GregoryComer:
Summary: Add an auto-incremented identifier field to the Method class and pass into BackendInitContext. This allows backends to disambiguated between methods with the same name (two models with a "forward" method, for example).

Differential Revision: D78697798


cccclai:
Is it for weight sharing in xnnpack? Can we use the delegate blob hash instead?

PR #12695, opened 2025-07-21
migrate exir test to use torch.cond instead of the hop directly

ydwu4:
Summary: torch.cond is the use-facing api, which allows closure, have safety checks etc.

Differential Revision: D78696006


larryliu0820 (approved):

JacobSzwejbka:
do torch.map and torch.while exist as well

ydwu4:
haven't yet, they can be accessed via torch._higher_order_ops.map/while_loop now

PR #12696, opened 2025-07-21
Add top-level CMake extensions target

GregoryComer:
Add a top level extensions CMake target, which includes all configured extensions. I validate this change by building a simple runner using the executorch_extensions target and verifying that it was able to build and run using the module and tensor extension. Once the changes land in ExecuTorch, this will be long-term validated in executorch-examples CI.


GregoryComer:
* #12807
* #12806
* __->__ #12696

tools/cmake/executorch-config.cmake
```diff
@@ -73,20 +73,23 @@ set(backend_lib_list
     vulkan_backend
 )
 
-set(lib_list
-    flatccrt
-    etdump
-    bundled_program
+set(extension_lib_list
```
larryliu0820: I think we don't have to do this after EXPORT is enabled. cc @swolchok

swolchok: ~~executorch_extensions isn't a target in our internal builds~~, so we'll still need to create it as we do here because there's nothing to export.

tools/cmake/executorch-config.cmake
```diff
@@ -213,3 +216,12 @@ foreach(lib ${backend_lib_list})
     target_link_options(executorch_backends INTERFACE ${lib})
   endif()
 endforeach()
+
+# An interface target containing all available extensions.
+add_library(executorch_extensions INTERFACE)
```
swolchok: how about also `add_library(executorch::extensions ALIAS executorch_extensions)`? also fine to skip for now and we can come back with a consistent namespacing strategy later

tools/cmake/executorch-config.cmake
swolchok: > executorch_extensions isn't a target in our internal builds

Sorry, this was totally wrong, I somehow completely missed the CMakeLists.txt changes.

CMakeLists.txt
```diff
@@ -706,11 +719,15 @@ if(EXECUTORCH_BUILD_VULKAN)
...
 
 # Top-level interface targets.
-add_library(executorch_backends INTERFACE)
 
 # A target containing all configured backends.
+add_library(executorch_backends INTERFACE)
 target_link_libraries(executorch_backends INTERFACE ${_executorch_backends})
 
+# A target containing all configured extensions.
+add_library(executorch_extensions INTERFACE)
```
swolchok, line 802: can you please include executorch_backends and executorch_extensions in the install() call alongside executorch and executorch_core so that we don't have to coordinate with the EXPORT PR? that should be enough to get it to work no matter which one lands first

tools/cmake/executorch-config.cmake
GregoryComer: I've updated to revert the changes to executorch-config.cmake and install the top-level interface targets.

CMakeLists.txt
GregoryComer: I've updated to install these targets. Because of the order of declarations in the file, I can't directly add to the existing executorch / executorch_core install, as it is above many of the backend declarations. To avoid causing a bunch of merge conflicts by rearranging it, I've added an install for these new targets alone. I can add the export directive once #8954 is merged.

tools/cmake/executorch-config.cmake
GregoryComer: Updated.

larryliu0820 (approved):

CMakeLists.txt
swolchok: if you change an example to use these targets, CI will prevent us from messing it up in the future and give this diff coverage

GregoryComer: I've updated executor_runner to use the top-level backends and extensions targets. I'll update examples with the export changes, as well.

GregoryComer:
The llama runner and ARM tests are infra failures due to download failures (likely related to the ongoing github incident).

PR #12704, opened 2025-07-22
Arm backend: Fix meandim decomp bug

AdrianLundell:
Wrong indexing caused the 4th dim to be skipped when reshaping 5D tensors. Additionally add unit test which covers this case.


Sebastian-Larsson (approved):
Unrelated CI error. Approved.

PR #12706, opened 2025-07-22
Use ::std::string_view

cyyever:
### Summary
This change is to make it possible to change PyTorch to use `::std::string_view` in torchgen. Other standard types in the same file also use `::std` namespace.


cyyever:
Unrelated failures.

Issue #12494, opened 2025-07-15
[Backend Tester] Add quantized test flows


GregoryComer:
### 🚀 The feature, motivation and pitch

Register quantized tester flows for P0 delegates

GregoryComer:
Linked PR is merged for XNN + Core ML quantized test flows. Vulkan and QNN are pending.

PR #12733, opened 2025-07-22
[Backend Tester] Add quantized test flows for XNNPACK and Core ML

GregoryComer:
Add quantized test flows core static int8 quantization for XNNPACK and Core ML backends. I also ended up doing some light refactoring on the test signature to pass the TestFlow class into the individual tests. This is done to allow for passing quantization parameters into the inner test.


GregoryComer:
* #12741
* #12739
* #12738
* #12736
* __->__ #12733

backends/apple/coreml/test/tester.py
```diff
@@ -4,23 +4,64 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, List, Optional, Tuple
+from typing import Any, List, Optional, Sequence, Tuple
 
+import coremltools as ct
```
digantdesai, line 10: come on :p

```suggestion
import coremltools
```

backends/apple/coreml/test/tester.py
```diff
@@ -4,23 +4,64 @@
...
 from executorch.backends.test.harness import Tester as TesterBase
 from executorch.backends.test.harness.stages import StageType
 from executorch.exir import EdgeCompileConfig
 from executorch.exir.backend.partitioner import Partitioner
 
 
+def _get_static_int8_qconfig():
+    return ct.optimize.torch.quantization.LinearQuantizerConfig(
+        global_config=ct.optimize.torch.quantization.ModuleLinearQuantizerConfig(
+            quantization_scheme="symmetric",
```
digantdesai, line 37: is this the main int8 schema we should be testing for Linear @metascroy

backends/apple/coreml/test/tester.py
```diff
@@ -4,23 +4,64 @@
...
+from executorch.backends.apple.coreml.compiler import CoreMLBackend
 from executorch.backends.apple.coreml.partition import CoreMLPartitioner
+from executorch.backends.apple.coreml.quantizer import CoreMLQuantizer
 from executorch.backends.test.harness import Tester as TesterBase
 from executorch.backends.test.harness.stages import StageType
 from executorch.exir import EdgeCompileConfig
 from executorch.exir.backend.partitioner import Partitioner
 
 
+def _get_static_int8_qconfig():
```
digantdesai:
```suggestion
def _get_static_int8_linear_qconfig():
```

backends/apple/coreml/test/tester.py
```diff
@@ -29,9 +70,14 @@ def __init__(
         self,
         partitioners: Optional[List[Partitioner]] = None,
         edge_compile_config: Optional[EdgeCompileConfig] = None,
+        minimum_deployment_target: Optional[Any] = ct.target.iOS15,
     ):
         super().__init__(
-            default_partitioner_cls=CoreMLPartitioner,
+            default_partitioner_cls=lambda: CoreMLPartitioner(
```
digantdesai: Nit: add a static method in the `Partitioner.get_default_partitioner()` to dedup?

backends/test/harness/stages/quantize.py
```diff
@@ -31,7 +31,8 @@ def __init__(
         self.calibrate = calibrate
         self.calibration_samples = calibration_samples
 
-        self.quantizer.set_global(self.quantization_config)
+        if self.quantization_config is not None:
+            self.quantizer.set_global(self.quantization_config)
```
digantdesai, line 36: add a `set_global: Optional[bool] = True` arg to disable this when using multiple quantizers?

backends/test/suite/flows/xnnpack.py
```diff
@@ -0,0 +1,36 @@
...
+def _create_xnnpack_flow() -> TestFlow:
+    return _create_xnnpack_flow_base("xnnpack")
+
+def _create_xnnpack_static_int8_flow() -> TestFlow:
+    def create_quantize_stage() -> Quantize:
+        qparams = get_symmetric_quantization_config(is_per_channel=True) 
+        return XnnpackQuantize(
+            quantization_config=qparams,
+        )
+    return _create_xnnpack_flow_base("xnnpack_static_int8", create_quantize_stage)
```
digantdesai: Nit:
```suggestion
    return _create_xnnpack_flow_base("xnnpack_static_per_channel_int8", create_quantize_stage)
```

backends/test/suite/flow.py
```diff
@@ -21,42 +22,35 @@ class TestFlow:
...
-        return TestFlow(
-            name="xnnpack",
-            backend="xnnpack",
-            tester_factory=XnnpackTester,
-        )
-    except Exception:
-        logger.info("Skipping XNNPACK flow registration due to import failure.")
-        return None
-
+        from executorch.backends.test.suite.flows.xnnpack import XNNPACK_TEST_FLOW, XNNPACK_STATIC_INT8_TEST_FLOW
```
digantdesai: `from ... import *`?

backends/test/suite/flow.py
```diff
@@ -21,42 +22,35 @@ class TestFlow:
...
-    tester_factory: Callable[[], Tester]
+    
+    tester_factory: Callable[..., Tester]
     """ A factory function that returns a Tester instance for this lowering flow. """
 
+    quantize: bool = field(default=False)
+    """ Whether to tester should run the quantize stage on the model. """
+    
+    quantize_stage_factory: Callable[..., Quantize] | None = None
+    """ A factory function which instantiates a Quantize stage. Can be None to use the tester's default. """
```
digantdesai: why an extra flag?

backends/test/suite/reporting.py
```diff
@@ -14,23 +14,26 @@ class TestResult(IntEnum):
 
     EAGER_FAIL = 2
     """ The test failed due to the model failing to run in eager mode. """
+    
+    QUANTIZE_FAIL = 3
```
digantdesai, line 18: Do we want to distinguish export_for_trainint (old) before quantize and export after quantize?

digantdesai (approved):
you can split quantizer and the using flow into two PRs but OK too.

GregoryComer:
Closes https://github.com/pytorch/executorch/issues/12494

backends/apple/coreml/test/tester.py
GregoryComer: FYI This is pulled directly from our docs at https://docs.pytorch.org/executorch/main/backends-coreml.html#bit-quantization-using-the-pt2e-flow. Would be good to sanity check with Scott, though.

backends/test/harness/stages/quantize.py
GregoryComer: Added. It might do a slightly larger refactor as a follow-up as the current pattern is a little awkward. Maybe just get rid of the quantization_config arg to the Quantize stage, since it's only used to pass into the quantizer.

backends/test/suite/flow.py
GregoryComer: The specific reason is that if quantize_stage_factory isn't provided, it will use the default Quantize stage from the tester. I could maybe just always require the caller to provide quantize_stage_factory.

backends/test/suite/reporting.py
GregoryComer: If I'm following correctly, are you asking about catching errors in the quant export vs the quantization itself?

It would be nice to have this, though we'd need to refactor the tester a bit to split it.

GregoryComer:
Updated to address comments and lints. There are a few minor follow-ups (see specific comment chains), but I'm intending to land and take follows-up after to keep things moving.

backends/apple/coreml/test/tester.py
GregoryComer: Updated.

backends/apple/coreml/test/tester.py
GregoryComer: Updated.

backends/test/suite/flow.py
GregoryComer: I'm going to clean up flow registration slightly in the stack, so I'll take this as a follow-up there.

GregoryComer:
Unit test failures are pre-existing on main (pytest coverage error). CI was green before rebasing. Merging.

backends/test/suite/reporting.py
digantdesai: yeah finer grained. We can do this later, hence the stamp :)

PR #12736, opened 2025-07-22
[Backend Tester] Clean up operator test logic

GregoryComer:
Minor refactoring on the operator test logic - since we now have separate model tests, I've moved operator-test specific test helper logic into the operators directory and updated usages. I also updated discovery slightly to give nicer error messages when things don't import.


GregoryComer:
* #12741
* #12739
* #12738
* __->__ #12736

digantdesai (approved):
do you envision `suite/<type>/__init__.py` per suite as a "controller" for writing tests? How do you see enforcing some suite level characteristics then? like all suits do this or support that.

GregoryComer:
> do you envision `suite/<type>/__init__.py` per suite as a "controller" for writing tests? How do you see enforcing some suite level characteristics then? like all suits do this or support that.

It's a good point. As a follow-up, I'll refactor out the common logic from the model and operator tests.

GregoryComer:
Unit test failures are pre-existing on main.

PR #12740, opened 2025-07-23
Fix Apple GH workflow on release branch

shoumikhin:


huydhn (approved):
This works as a release only change for 0.7, but is too specific.  You could use contains or startWith like https://github.com/pytorch/pytorch/blob/main/.github/workflows/_linux-build.yml#L120C79-L120C124, so it works with all release branches in the future

shoumikhin:
@huydhn gonna experiment with that on main for the next release

PR #12743, opened 2025-07-23
Don't build MmapDataLoader if sys/mman.h isn't available and we're not on Windows

swolchok:
It won't work in this case.


swolchok:
* #12792
* #8954
* #12744
* __->__ #12743

cccclai (approved):
Makes sense

swolchok:
failing builds all seem to be failing in HUD as well. merging!

PR #12744, opened 2025-07-23
Fix missing cmake_deps.toml  dep from size_test to extension_data_loader

swolchok:
Missing deps like this causes executorch_srcs.cmake to just roll the sources into the would-be-depending target, which is bad.


swolchok:
* #12792
* #8954
* __->__ #12744

larryliu0820 (approved):

digantdesai:
yeah this explains why we have this compiled even when the CMake variable was OFF
https://github.com/pytorch/executorch/actions/runs/16482988738/job/46601732925?pr=12678#step:15:15962

Gasoonjia (approved):

swolchok:
https://github.com/pytorch/executorch/actions/runs/16502309674/job/46669895289?pr=12744 (test-arm-cortex-m-size-test, zephyr version) looks like a real failure; maybe I fixed it in the wrong PR or something.

swolchok:
I think we're good, jobs I was concerned about have passed. merging.

PR #12752, opened 2025-07-23
Create add-unanswered-to-project.yml

nil-is-all:
Adding a workflow automation to dynamically add unanswered PRs and issues to project view (https://github.com/orgs/pytorch/projects/136)


GregoryComer:
Is it possible to test the workflow before landing it in the repo? If so, do you have any results you can attach to the PR?

.github/workflows/add-unanswered-to-project.yml
```diff
@@ -0,0 +1,76 @@
...
+              const prs = await github.paginate(
+                github.rest.pulls.list,
+                {
+                  owner,
+                  repo,
+                  state: 'open',
+                  draft: false,
+                }
+              );
+              for (const pr of prs) {
```
GregoryComer, line 86: Can you remove the unapproved constraint? I think it might be good to track approved PRs that haven't been merged yet.

.github/workflows/add-unanswered-to-project.yml
```diff
@@ -0,0 +1,76 @@
...
+                  }
+                `);
+                console.log(`Added ${type} #${number} to project`);
+              } catch (error) {
+                console.log(`Error adding ${type} #${number}: ${error.message}`);
+              }
+            }
+
+            try {
+              // Add open issues (not PRs)
```
GregoryComer: It looks like it's only adding issues. Can we add both PRs and issues?

.github/workflows/add-unanswered-to-project.yml
```diff
@@ -0,0 +1,76 @@
+name: Add Unanswered PRs and Issues to PyTorch Org Project 136
```
GregoryComer: What do you think about renaming to "Add External PRs and Issues...", since we're not filtering by unanswered yet?

GregoryComer:
Would it be possible to add the author filter in the workflow, as well? Or is it easier to apply as a view?

.github/workflows/add-unanswered-to-project.yml
nil-is-all: Sure, considering the current queries look at just open PRs and Issues. I could change it to "Add Open External Contributor PRs and Issues to PyTorch Org Project 136"

.github/workflows/add-unanswered-to-project.yml
```diff
@@ -0,0 +1,76 @@
...
+                  filter: 'all'
+                }
+              );
+              for (const issue of issues) {
+                if (!issue.pull_request) {
+                  await addItem(issue.node_id, 'issue', issue.number);
+                }
+              }
+
+              // Add open, non-draft PRs with NO approved reviews
```
nil-is-all: @GregoryComer  this block covers PRs as well

GregoryComer (approved):

PR #12785, opened 2025-07-23
Removed print statement in XNNPACK quantize utils

leafs1:
### Summary
Removed print statement in XNNPACK quantize utils

### Test plan
Just removed the print statment.


GregoryComer:
Note that this change is landed on main as part of larger changes. This PR picks only the log removal for 0.7.

larryliu0820 (approved):

PR #12792, opened 2025-07-23
Sweep major CMake files for use of include/lib instead of CMAKE_INSTALL_INCLUDE/LIBDIR

swolchok:
Supposed to use the latter. (This is the thing that causes binaries to show up in lib64 for some weird builds, for example.)


swolchok:
* __->__ #12792

GregoryComer (approved):

swolchok:
build-demo-ios is known broken, other two failures look fine to bypsas as we are looking for cmake problems here. merging.

PR #12798, opened 2025-07-24
optimize group based partitioner to not try to merge partitions we already determined were not supposed to be merged

leafs1:
### Summary
In the group based partitioner we had a case where we would try to merge partitions we already determined couldn't be merged, which was causing a lot more checks to be done. This pr creates a set that tracks the partitions that we determined cant be merged and ensures that when considering 2 new partitions either id isnt in this set. Find below the number of checks on whether each partition can be merged before and after respectively which has a runtime of O(n^2) where n represents the number of nodes in a partition.
image

image

### Test plan
I added cnt print logs to check to see the number of times we entered the check both before and after the change.

### DIFF
--- a/exir/backend/canonical_partitioners/group_partitioner.py
+++ b/exir/backend/canonical_partitioners/group_partitioner.py
@@ -204,35 +204,37 @@ class GroupBasedPartitioner(CapabilityBasedPartitioner):
         partitions_order,
     ):
         """Merge partitions when possible."""
-        merged = True
-        while merged:
-            merged = False
-            partition_ids = list(partitions_by_id.keys())
+        # Get current partition IDs
+        partition_ids = list(partitions_by_id.keys())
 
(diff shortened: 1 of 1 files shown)


exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -204,35 +204,37 @@ def _merge_partitions(
...
-                        merged = True
-                        break
-
-                if merged:
-                    break
+                # Try to merge partitions if it doesn't create cycles
+                if self._can_merge_partitions(p1, p2, partitions_by_id):
+                    self._perform_partition_merge(
+                        p1,
+                        p2,
```
mcr229, line 228: what actually happens to p2 after it get merged into p1? is it still valid? As in can we add it to the already_merged set below? For the already_merged set should we instead be using its index in partition_ids?

mcr229 (approved):

PR #12806, opened 2025-07-24
Add top-level CMake kernels target

GregoryComer:
Add a top level kernels CMake target, which includes all configured kernel libraries. I validate this change by building a simple runner using the executorch_kernels target and verifying that it was able to build and run MobileNet v2, with and without selective build. Once the changes land in ExecuTorch, this will be long-term validated in executorch-examples CI.

There are a few follow-ups needed, which I will handle shortly. I still need to add merging logic for the YAML files to generate a single selective library that includes quantized and LLM ops, when configured to build them.


GregoryComer:
* #12807
* __->__ #12806

larryliu0820 (approved):

tools/cmake/preset/default.cmake
```diff
@@ -253,6 +275,17 @@ check_conflicting_options_on(
   EXECUTORCH_BUILD_CPUINFO
 )
 
+# Selective build specifiers are mutually exclusive.
+check_conflicting_options_on(
+  IF_ON EXECUTORCH_SELECT_OPS_YAML CONFLICTS_WITH
+  EXECUTORCH_SELECT_OPS_LIST EXECUTORCH_SELECT_OPS_MODEL
+)
```
larryliu0820, line 313: Nice

larryliu0820: I mean in the future it should be composable. I'm OK with having this for now.

CMakeLists.txt
```diff
@@ -730,8 +737,62 @@ add_library(executorch_extensions INTERFACE)
 add_library(executorch::extensions ALIAS executorch_extensions)
 target_link_libraries(executorch_extensions INTERFACE ${_executorch_extensions})
 
+# A target containing all configured kernels, with selective build, if enabled.
+add_library(executorch_kernels INTERFACE)
+add_library(executorch::kernels ALIAS executorch_kernels)
+if(NOT EXECUTORCH_SELECT_OPS_YAML STREQUAL ""
+   OR NOT EXECUTORCH_SELECT_OPS_LIST STREQUAL ""
+   OR NOT EXECUTORCH_SELECT_OPS_MODEL STREQUAL ""
+)
```
larryliu0820, line 819: Thank you for adding this!

larryliu0820:
@GregoryComer can you remove these options: https://github.com/pytorch/executorch/blob/main/examples/selective_build/CMakeLists.txt#L55 and let it use the ones you added

tools/cmake/preset/default.cmake
BujSet: @GregoryComer  Should this also check for a conflict with the `OPS_FROM_MODEL` API?

BujSet: Oh I see it there, later down.

CMakeLists.txt
```diff
@@ -730,8 +737,62 @@ add_library(executorch_extensions INTERFACE)
...
+    ""
+    DTYPE_SELECTIVE_BUILD
+    "${EXECUTORCH_ENABLE_DTYPE_SELECTIVE_BUILD}"
+  )
+
+  gen_operators_lib(
+    LIB_NAME
+    "executorch_selected_kernels"
+    KERNEL_LIBS
+    "portable_kernels"
```
lucylq, line 850: Thanks for adding this to top-level cmake!

So, `executorch_selected_kernels` is a selective build library for portable_kernels only it seems? If a user wants to selectively build on optimized or custom kernel libs, they should either:
1. Update here (for test purposes, I guess).
2. Follow the current flow and create their own selective build lib.

tools/cmake/preset/default.cmake
```diff
@@ -203,6 +203,28 @@ define_overridable_option(
...
+  "Build the executorch_kernels target with a list of selected operators."
+  STRING ""
+)
+define_overridable_option(
+  EXECUTORCH_SELECT_OPS_MODEL
+  "Build the executorch_kernels target with only operators from the given model .pte file."
+  STRING ""
+)
+define_overridable_option(
+  EXECUTORCH_ENABLE_DTYPE_SELECTIVE_BUILD
```
lucylq, line 234: Can we add a `check_required_options_on` here, that `EXECUTORCH_ENABLE_DTYPE_SELECTIVE_BUILD` relies on `EXECUTORCH_SELECT_OPS_MODEL`?

CMakeLists.txt
GregoryComer: Yeah, as a follow-up, I intend to enable optimized (and portable_optimized) kernels. To enable selective build for other libraries (quantized ops, llm ops, etc.), I should just be able to use the merge_yaml functionality to generate a combined op yaml, right?

For user's custom op libs, I think they'll probably need to define their own targets, as we can't access their targets from here (unless we re-wrote as a macro/function, but we can look at that separately).

Backend op libs are another case that needs more thinking through, as well. But hopefully this should be a good starting point.

tools/cmake/preset/default.cmake
GregoryComer: Added.

GregoryComer:
> @GregoryComer can you remove these options: https://github.com/pytorch/executorch/blob/main/examples/selective_build/CMakeLists.txt#L55 and let it use the ones you added

I've created a follow-up task to track this: https://github.com/pytorch/executorch/issues/12948. Will do this shortly.

PR #12807, opened 2025-07-24
Update docs to use top-level CMake targets

GregoryComer:
Update the "Building from Source" and "Using ExecuTorch with C++" docs to specify the top-level CMake targets. I also added some info to the selective build docs to document the CMake flags which control selective build with the `executorch_kernels` target.


GregoryComer:
* __->__ #12807

mergennachin (approved):

mergennachin:
cc @lucylq

lucylq:
Can we CP this to 0.7, do these changes only reflect code in main?

GregoryComer:
> Can we CP this to 0.7, do these changes only reflect code in main?

The actual CMake changes aren't in 0.7, so this should probably wait for 1.0.

PR #12808, opened 2025-07-24
Update CoreML model testing in CI

metascroy:
This PR updates the alpha/beta CoreML model testing in a few ways:

* All models are now testing with to_edge_lower_and_transform, the method we recommend in docs
* We run pybindings on exported models to ensure 1) they run, and 2) the output is similar to pytorch eager
* We add more model tests to CI

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -233,6 +233,7 @@ test_model_with_qnn() {
 test_model_with_coreml() {
   local should_test="$1"
   local test_with_pybindings="$2"
+  local dtype="$3"
 
   if [[ "${BUILD_TOOL}" != "cmake" ]]; then
     echo "coreml only supports cmake."
@@ -245,7 +246,7 @@ test_model_with_coreml() {
     export RUN_WITH_PYBINDINGS="--run_with_pybindings"
   fi
 
-  "${PYTHON_EXECUTABLE}" -m examples.apple.coreml.scripts.export --model_name="${MODEL_NAME}" --compute_precision float16 --use_partitioner ${RUN_WITH_PYBINDINGS}
+  "${PYTHON_EXECUTABLE}" -m examples.apple.coreml.scripts.export --model_name="${MODEL_NAME}" --compute_precision ${dtype} --use_partitioner ${RUN_WITH_PYBINDINGS}
   EXPORTED_MODEL=$(find "." -type f -name "${MODEL_NAME}*.pte" -print -quit)
 
   if [ -n "$EXPORTED_MODEL" ]; then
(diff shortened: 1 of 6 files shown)


.ci/scripts/test_model.sh
```diff
@@ -232,21 +232,23 @@ test_model_with_qnn() {
...
 test_model_with_coreml() {
   local should_test="$1"
+  local test_with_pybindings="$2"
 
   if [[ "${BUILD_TOOL}" != "cmake" ]]; then
     echo "coreml only supports cmake."
     exit 1
   fi
 
-  DTYPE=float16
```
digantdesai, line 241: we don't expect to run with other dtypes?

metascroy:
@digantdesai @cccclai can I get a stamp?

cccclai (approved):

.github/workflows/trunk.yml
```diff
@@ -590,12 +590,20 @@ jobs:
         echo "Finishing installing coreml."
 
         # Build and test coreml model
-        MODELS=(mv3 ic4 resnet50 edsr mobilebert w2l)
-        for MODEL_NAME in "${MODELS[@]}"; do
+        for MODEL_NAME in dl3 edsr emformer_join ic3 ic4 mobilebert mv2 mv3 resnet50 vit w2l; do
```
mergennachin: @metascroy have you seen this PR?

https://github.com/pytorch/executorch/pull/12371/files

What are your thoughts on consolidating? I can drop my PR but maybe you can take a look and see if there are things worth incorporating?

examples/apple/coreml/scripts/export.py
```diff
@@ -154,6 +166,37 @@ def generate_compile_specs_from_args(args):
...
+    if isinstance(eager_outputs, list | tuple):
+        eager_outputs = eager_outputs[0]
+
+    mse = ((et_outputs - eager_outputs) ** 2).mean().sqrt()
+    print(f"Mean square error: {mse}")
+    assert mse < 0.1, "Mean square error is too high."
+
+    if dtype == torch.float32:
+        assert torch.allclose(
+            et_outputs, eager_outputs, atol=1e-02, rtol=1e-02
```
digantdesai, line 194: what would these look like for fp16?

.github/workflows/trunk.yml
metascroy: @mergennachin  I missed that PR, but I can incorporate the following changes from your PR here:

* Split CoreML/MPS into two separate jobs
* Merge the model lists from the two PRs

For the runtime checks, I'll proceed with the pybind testing I'm adding here (vs. the C++ runner checking in your PR) because with pybinding we also check model accuracy.

.ci/scripts/test_model.sh
metascroy: I could add fp32 runs, as well.  But I was trying to run with the recommend config for each model, and for most models that is FP16 so things can run on the ANE.

With FP16, mobile bert returns nan output, so I'll likely run that with FP32 + quantization (in this PR, mobile bert only goes through export check)

examples/apple/coreml/scripts/export.py
metascroy: I think 1e-02 would work for many of the models, but it doesn't reliably pass so it is not something I want to add to CI.  I opted for relatively loose MSE checks instead.

.github/workflows/trunk.yml
mergennachin: Thank you @metascroy

Changes look great

.ci/scripts/test_model.sh
digantdesai: yeah such pragmatic approach with dtype for CI makes sense as a first pass.

PR #12809, opened 2025-07-24
[Backend Tester] Add tensor error statistic reporting

GregoryComer:
Report various error statistics for the test outputs, including SQNR, mean absolute error (MAE), and L2 norm. These are saved in the detail report per test case.

As an example, here is the output from Core ML running MobileNet V2 (roughly formatted from csv -> sheets -> markdown):
```
Output 0 Error Max	Output 0 Error MAE	Output 0 Error MSD	Output 0 Error L2	Output 0 SQNR
0.0005887411535		0.0001199183663		2.32E-06		0.004750485188		41.28595734
```


GregoryComer:
* #13313
* #13312
* #13311
* #13309
* #13308
* #13306
* #13258
* #13250
* #13249
* #12846
* #12838
* __->__ #12809

backends/test/harness/tester.py
```diff
@@ -302,17 +303,15 @@ def run_method_and_compare_outputs(
         atol=1e-03,
         rtol=1e-03,
         qtol=0,
+        statistics_callback: Callable[[ErrorStatistics], None] | None = None,
```
GregoryComer, line 306: I'm not completely happy with the callback approach for exposing this, but I don't really have a better idea, since the tester relies on a builder-style pattern where it returns self to allow chaining. I'm open to suggestions.

GregoryComer:
I'll add some tests for the error statistic calculation to this PR shortly.

digantdesai (approved):

backends/test/harness/tester.py
digantdesai: just update the tester method?

GregoryComer: Can you clarify what you're thinking? Are you meaning update the tester run_method_and_compare outputs to directly return the error stats and then update all of the callers to not use it in a chained fashion? Or something else?

digantdesai: Modify the existing method, keeping the outside behavior but also add `def get_comparison_stats(self)` method on that stage or something? 
Or if you want to pass a callback for flexibility that's also fine.

backends/test/harness/error_statistics.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @classmethod
+    def from_tensors(
+        cls, actual: torch.Tensor, reference: torch.Tensor
+    ) -> "ErrorStatistics":
+        """Creates an ErrorStatistics object from two tensors."""
+        actual = actual.to(torch.float64)
+        reference = reference.to(torch.float64)
+
+        if actual.shape != reference.shape:
```
digantdesai, line 76: Why is this not an error?

GregoryComer: There are some cases where in-place ops get functionalized and alter the graph outputs. This is an issue with the defunctionalization logic in ET (there's a separate issue I filed). It's not technically the backend's fault but is a real issue. I should probably just disable the affected tests for now and treat this as an error. I'll do that.

GregoryComer: Filed https://github.com/pytorch/executorch/issues/13336 as a follow-up. Will stack this change on top.

backends/test/harness/tester.py
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13337. Will stack this on top.

Issue #8700, opened 2025-02-25
Fail to allocate temp memory by exporting torch.topk


JoshuaGhost:
### 🐛 Describe the bug

Hi, all. I am following the tutorial to learn to use the executor. For my future project, I would also like to export a toy model.
In my toy model, there is a call to the `torch.topk` and I would like to export it as a portable `.pte` file. But when I run it using the executorch runtime, there is a runtime error 0x21 and the reason is that temp memory failed to allocate. Here is my code for exporting the toy module:

```python
import torch
from pathlib import Path
from torch.export import export, Dim
from executorch.exir import EdgeProgramManager, to_edge_transform_and_lower, EdgeCompileConfig
from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
from executorch.runtime import Verification, Runtime, Program, Method


class ToyModel(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()

    def forward(self, x) -> torch.Tensor:
        top_values = torch.ones(x.shape[0], 3, dtype=torch.float32)
        top_idxs = torch.ones(x.shape[0], 3, dtype=torch.long)
        torch.topk(x.contiguous(), 3, out=(top_values, top_idxs))
        return top_values, top_idxs
    
    
if __name__ == "__main__":
    model = ToyModel()
    example_input = (torch.randn(2, 256),)
    with torch.no_grad():
        aten_dialect = export(model, example_input)

        backend_dialect: EdgeProgramManager = to_edge_transform_and_lower(
            aten_dialect,
            compile_config=EdgeCompileConfig(_check_ir_validity=False),
            partitioner=[XnnpackPartitioner()],
        )
        executorch_dialect = backend_dialect.to_executorch(
        )
        executorch_dialect.save("topk.pte")
        

    et_runtime: Runtime = Runtime.get()
    program: Program = et_runtime.load_program(
        Path("topk.pte"),
        verification=Verification.Minimal,
    )
    print("Program methods:", program.method_names)
    forward: Method = program.load_method("forward")

    outputs = forward.execute(example_input)
    print(f"Ran forward({example_input[0].shape})")
    print(f"  outputs: {outputs[0].shape}")
```

By noticing the memory allocation error, I used the out variant of the `topk` (the one with the kwarg `out`). But the same error still occurs. Here is the full traceback:

```
[memory_allocator.h:88] Memory allocation failed: 4096B requested (adjusted for alignment), 0B available
[...]

PR #12810, opened 2025-07-24
Fix temp memory allocation issue in torch.topk operations

ZihangFang:
Summary: Fixes issue https://github.com/pytorch/executorch/issues/8700 

changed the temp_allocator_ from a
MemoryAllocator with null buffer to a MallocMemoryAllocator that can
 dynamically allocate memory as needed.

Test plan: To verify the fix, a new end-to-end test suite has been added (`test/end2end/test_temp_allocator_fix.py`). This suite includes tests for `torch.topk` both with and without the `out` parameter, as well as a test with a larger input to ensure the allocator can handle more significant memory requirements. These tests now pass with the implemented fix. I will now write the PR description.


cccclai (approved):
This seems good to me, though we really need to finish migrating the extension module...cc: @Gasoonjia

mergennachin:
Hi @ZihangFang can you fix the lintrunner please?

ZihangFang:
> Hi @ZihangFang can you fix the lintrunner please?

linting issue fixed

PR #12839, opened 2025-07-25
Extend `PyBundledModule` with `extension.BundledModule`

Gasoonjia:
Differential Revision: D78938344


CMakeLists.txt
```diff
@@ -561,6 +561,12 @@ install(FILES tools/cmake/executorch-config.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ExecuTorch
 )
 
+# Build devtools first if needed - some backends depend on protobuf from
+# devtools
```
lucylq: Can you specify which backends require this? :0

Can we try move this before line 426, if(EXECUTORCH_BUILD_PYBIND AND APPLE), the error in apple pybindings seems to be related to protobuf. 

https://github.com/pytorch/executorch/blob/365d4c17c702def8e682e33e6503681a408f786c/CMakeLists.txt#L426

```
CMake Error at backends/apple/coreml/CMakeLists.txt:215 (add_subdirectory):
  add_subdirectory given source
  "/Users/runner/work/executorch/executorch/pytorch/executorch/backends/apple/coreml/third-party/coremltools/deps/protobuf/cmake"
  which is not an existing directory.


CMake Error at tools/cmake/Utils.cmake:36 (target_link_options):
  Cannot specify link options for target "libprotobuf-lite" which is not
  built by this project.
  tools/cmake/Utils.cmake:54 (executorch_macos_kernel_link_options)
  backends/apple/coreml/CMakeLists.txt:219 (executorch_target_link_options_shared_lib)


CMake Error at backends/apple/coreml/CMakeLists.txt:246 (target_compile_options):
  Cannot specify compile options for target "libprotobuf-lite" which is not
  built by this project.
  ```

Gasoonjia: reasonable point! Let me try that

lucylq (approved):

PR #12844, opened 2025-07-25
[Group Partitioner] Optimize Speed

mcr229:
We do some optimizations on the group partitioner to improve its speed. Theoretically, since we've pre-grouped some partitions already, we should be faster than the capability based partitioner. For example, if we a dynamically quantized group, this could contain 
```
[ choose_q_params, quantize, dequantize, op, dequantize_weight, weight, bias, get_item_scale, get_item_zp]
```

9 nodes. In capability-based partitioner they will have to run DFS on all 9 nodes in order to group these together. Based on the hints and purpose of the group based partitioenr, we don't perform these checks and instead group all these 9 nodes, saving time by avoiding these checks. Some stats when partitioning the mobile bert model:

```
elpased time old partitioner: 65.3421
old_partitioner num partitions: 170


elpased time new partitioner: 5.1964
new_partitioner num partitions: 170
```

we see a 13x improvement in partitioning when using the group based partitioner, while still partitioning around the same number of nodes.

Differential Revision: [D79020720](https://our.internmc.facebook.com/intern/diff/D79020720)


mcr229:
* #12845
* __->__ #12844

leafs1 (approved):

mcr229:
@mcr229 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D79020720).

PR #12846, opened 2025-07-25
[Backend Tester] Report delegation statistics

GregoryComer:
Report total number of delegated and undelegated nodes and breakdown by operator count.

Example from CoreML add:
Test ID | Test Case | Backend | Delegated Nodes | Undelegated Nodes | Delegated Ops | Undelegated Ops
-- | -- | -- | -- | -- | -- | --
test_add_dtype_float32_coreml | test_add_dtype | coreml | 1 | 0 | {'aten::add.Tensor': 1} | {}
test_add_dtype_float32_coreml_static_int8 | test_add_dtype | coreml | 7 | 0 | {'aten::add.Tensor': 1, 'quantized_decomposed::dequantize_per_tensor': 3, 'quantized_decomposed::quantize_per_tensor': 3} | {}


GregoryComer:
* #13313
* #13312
* #13311
* #13309
* #13308
* #13306
* #13258
* #13250
* #13249
* __->__ #12846

GregoryComer:
For reviewers, is there a better way to present the operator counts? Are there other delegation-related stats that we should include?

backends/test/suite/reporting.py
```diff
@@ -188,6 +238,24 @@ def complete_test_session() -> RunSummary:
...
+    return sum(counter.values()) if counter is not None else None
+
+
+def _serialize_op_counts(counter: Counter | None) -> str:
+    """
+    A utility function to serialize op counts to a string, for the purpose of including
+    in the test report.
+    """
+    if counter is not None:
+        return str(dict(sorted(counter.items())))
```
digantdesai, line 254: we can make this pretty but OK for now

backends/test/suite/runner.py
```diff
@@ -98,14 +100,25 @@ def build_result(
...
-        tester.to_edge_transform_and_lower()
+        tester.to_edge_transform_and_lower(generate_etrecord=True)
         elapsed = time.perf_counter() - lower_start_time
         extra_stats["lower_time"] = timedelta(seconds=elapsed)
     except Exception as e:
         elapsed = time.perf_counter() - lower_start_time
         extra_stats["lower_time"] = timedelta(seconds=elapsed)
         return build_result(TestResult.LOWER_FAIL, e)
 
+    # Compute delegation statistics. Use the ETRecord to access the edge dialect graph between
```
digantdesai, line 111: Curious why do we need to go through ETRecord?

digantdesai (approved):

backends/test/suite/runner.py
GregoryComer: This is because we need the graph in between to_edge and to_backend when using to_edge_transform_and_lower in order to see the edge-dialect graph before partitioning. In theory, we have the info with the exported program, but it's aten dialect and pre-decompositions, so it's hard to reason about what's delegated.

With to_edge_transform_and_lower, we don't otherwise have a way to inspect the graph between the edge and lowering steps. ETRecord has effectively the same problem (they need the pre-delegation edge graph), so I'm using it here.

zingo:
As a note: This might have broken the Arm backend testing, we have a fix to be merged soon if testing goes well.
https://github.com/pytorch/executorch/pull/13370

But seem like a nice thing, would be fun to play with this!

PR #12850, opened 2025-07-25
[Backend Tester] Add permute, transpose, and masked_fill tests

GregoryComer:
Add tests for permute, transpose, and masked_fill.


GregoryComer:
* #12924
* #12856
* #12855
* #12854
* #12853
* #12852
* #12851
* __->__ #12850

backends/test/suite/operators/test_permute.py
```diff
@@ -0,0 +1,109 @@
...
+    def test_permute_dtype(self, flow: TestFlow, dtype) -> None:
+        self._test_op(
+            PermuteModel(dims=[1, 0]),
+            (torch.rand(20, 32).to(dtype),),
+            flow,
+        )
+
+    def test_permute_3d(self, flow: TestFlow) -> None:
+        self._test_op(
+            PermuteModel(dims=[2, 0, 1]),
```
digantdesai, line 42: some test generator or Facto or even a meta class to generate permutations for permute might be better?

digantdesai (approved):

backends/test/suite/operators/test_permute.py
GregoryComer: Makes sense. Should be easy to use itertools for this, I think. I'll take this as a follow-up.

PR #12851, opened 2025-07-25
[Backend Tester] Add slice and reshape tests

GregoryComer:
Add tests for a view-type ops, cat, and slice.


GregoryComer:
* #12856
* #12855
* #12854
* #12853
* #12852
* __->__ #12851

digantdesai (approved):
I am assuming you have some way to figure out "how to test" in terms of coverage and which tests to add for these ops.

GregoryComer:
> I am assuming you have some way to figure out "how to test" in terms of coverage and which tests to add for these ops.

My methodology isn't super rigorous, but I am intending to cover all relevant portions of the user-facing torch API surface, under torch.nn, torch.nn.functional, and tensor ops. I've generated the list of ops to test from inspecting the docs: https://docs.pytorch.org/docs/stable/torch.html. I'm definitely open to suggestions.

I prefer the user-facing surface over the core opset directly, in part because the wrapper modules can sometimes transform inputs in certain ways or enforce certain constraints that the backends often expect. Ideally, the backends would handle the full op-set space, but this was a common source of noise in the facto tests. The backend fails on some raw op inputs that cannot be generated through the nn.Module or similar.

PR #12860, opened 2025-07-25
[ET-VK] Add 2D Reduction to Vulkan Backend

alexdean08:
Summary: This change adds 2D reduction to the Vulkan delegate. Prior to this change, only 1D reduction was implemented. Models like MobileNetV3 and ResNet do 2D reduction, and their performance was being negatively impacted by the lack of a 2D reduction Vulkan implementation.


backends/vulkan/runtime/graph/ops/impl/Reduce.cpp
```diff
@@ -137,15 +155,89 @@ void add_reduce_node(
...
+    if (dims_list.size() == 1) {                                         \
+      const int64_t dim_val = dims_list.at(0);                           \
+      const ValueRef dim_ref = graph.get_or_add_value_for_int(dim_val);  \
+      return add_reduce_node(                                            \
+          graph, args[0], dim_ref, args[out_arg_idx], #op_name);         \
+    } else if (dims_list.size() == 2) {                                  \
+      return add_reduce2d_node(                                          \
+          graph, args[0], args[1], args[out_arg_idx], #op_name);         \
+    } else {                                                             \
+      VK_CHECK_COND(false, "Only 1 or 2 dimensions supported");          \
```
msluszniak: if and else if branches return so you can simplify your code
```suggestion
    if (dims_list.size() == 1) {                                         \
      const int64_t dim_val = dims_list.at(0);                           \
      const ValueRef dim_ref = graph.get_or_add_value_for_int(dim_val);  \
      return add_reduce_node(                                            \
          graph, args[0], dim_ref, args[out_arg_idx], #op_name);         \
    }
    if (dims_list.size() == 2) {                                  \
      return add_reduce2d_node(                                          \
          graph, args[0], args[1], args[out_arg_idx], #op_name);         \
    }
    VK_CHECK_COND(false, "Only 1 or 2 dimensions supported");          \
```

backends/vulkan/runtime/graph/ops/glsl/reduce2d.glsl
```diff
@@ -0,0 +1,128 @@
...
+// Initializing the accumulator accepts the first value in the reduction row,
+// since some reduction operations (i.e. amax, amin) prefer to initialize with
+// a data point instead of a static value.
+#define INIT_ACCUM(first_val) ${INIT_ACCUM}
+#define UPDATE_ACCUM(accum, new_val) ${UPDATE_ACCUM}
+// Useful for operators such as mean which want to perform a final calculation
+// with the accumulator.
+#define POSTPROCESS(accum) ${POSTPROCESS}
+
+void reduce_2d(const ivec2 tid, ivec3 scan_pos) {
```
SS-JIA: As I understand it, this implementation doesn't currently support reduction among the elements of a texel,.

So for example for a tensor of size `(channels = 4, height = 16, width = 16)` where reading one texel returns 4 channels, this function would only be able to handle reduction along the height and width dims, but not if channels is one of the dims being reduced - is that correct?

If so, we can leave the more generalized implementation for later, but we should add some checks to make sure we don't lower unsupported cases to Vulkan 😛 

Would also recommend renaming this function to `reduce_2d_non_texel_dim` or `reduce_2d_non_packed_dim`.

backends/vulkan/runtime/graph/ops/impl/Reduce.cpp
```diff
@@ -137,15 +155,90 @@ void add_reduce_node(
...
+  VK_CHECK_COND(dims_list.size() == 2, "reduce2d requires exactly 2 dimensions");
+  
+  int32_t reduce_dim1 = normalize(dims_list[0], ndim);
+  int32_t reduce_dim2 = normalize(dims_list[1], ndim);
+  
+  // Convert to WHCN format
+  reduce_dim1 = nchw_dim_to_whcn_dim(reduce_dim1, ndim);
+  reduce_dim2 = nchw_dim_to_whcn_dim(reduce_dim2, ndim);
+  
+  // Check that the concat dim is not one of the reduction dims
```
SS-JIA, line 190: should also add a check that `graph.packed_dim_of(in)` and `graph.packed_dim_of(out)` is not one of the reduction dims.

SS-JIA (approved):
LGTM with some small comments. Thanks for working on this!

backends/vulkan/runtime/graph/ops/impl/Reduce.cpp
alexdean08: Just added above this comment. Let me know if you're aligned

backends/vulkan/runtime/graph/ops/glsl/reduce2d.glsl
alexdean08: Yup you're right, this doesn't support reduction among the elements of a texel. Just changed to reduce_2d_non_packed_dim. Let me know if you're aligned

PR #12871, opened 2025-07-25
Change node group partitioning to be with all nodes, to keep partition ids in top sort order

leafs1:
### Summary
With the current partitioner implementation, we need to traverse the partitions in reverse top sort order for the merging algorithm, as we have some invariants when doing cycle detection and checking downstream dependencies. The current problem is that we first form the group partitions and give them incrementing ids. Since they aren't guaranteed to be in top sort order, when we assign the remaining partitions to the ungrouped nodes, their ids will start from where the grouped ids left off, breaking our invariant. This PR assigns ids in order of reverse top sort, where if we encounter a group node we create the partition for the whole group and if we just encounter ungrouped nodes we create the partition just for that node.

### Test plan
All previous tests for functionality still pass

### DIFF
--- a/exir/backend/canonical_partitioners/group_partitioner.py
+++ b/exir/backend/canonical_partitioners/group_partitioner.py
@@ -123,7 +123,7 @@ class GroupBasedPartitioner(CapabilityBasedPartitioner):
 
         return True
 
-    def _process_node_groups(
+    def _process_all_nodes(
         self,
         new_partition_id,
         partitions_by_id,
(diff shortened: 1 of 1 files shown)


exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -123,7 +123,7 @@ def _can_merge_partitions(self, p1, p2, partitions_by_id):
 
         return True
 
-    def _process_node_groups(
+    def _process_remaining_nodes(
```
mcr229: let's change this to process_all_nodes or something like that

exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -133,97 +133,60 @@ def _process_node_groups(
...
-        # TODO: this currently doesn't work with _process_remaining_nodes so
-        # if a user provides grouped nodes with operatorsupport, then this will
-        # faile
+        """Process nodes not in any predefined group."""
         for node in reversed(self.graph_module.graph.nodes):
-            if node not in self.node_to_group:
+            if node in assignment or not self._is_node_supported(node):
                 continue
 
-            if node in processed_nodes:
```
mcr229, line 154: what about this check? The purpose was that when traversing nodes in the graph module, we could get two nodes from the same group, and we could potentially add the group twice?

exir/backend/canonical_partitioners/group_partitioner.py
```diff
@@ -133,97 +133,60 @@ def _process_node_groups(
...
+                    user
+                    for node in partition.nodes
+                    for user in node.users
+                    if user not in partition.nodes
+                }
+
+                # Update partition map
+                for node in partition.nodes:
+                    for user in node.users:
+                        target_id = assignment.get(user)
```
mcr229: ``
```suggestion
                        target_id = assignment.get(user, None)
```

Just for readability

mcr229 (changes requested):
I few comments, let me know your thoughts

mcr229 (approved):

PR #12873, opened 2025-07-25
Support for Dtye Selective Build from Dictionary API in OSS (cmake)

BujSet:
### Summary
When attempting to build minimal binary size with only a select number of operators and dtypes, it's advantageous to simply specify only the needed selections at compile time. This PR allows one to specify a JSON formatted string to select operators and dtypes to be included in the final binary. An example is also provided.

### Test plan
This feature was tested with the added example in `examples/selective_build/test_selective_build.sh` which now has a function named `test_cmake_select_ops_in_dict()`. This function exemplifies the usage of the newly added dictionary API.


examples/selective_build/test_selective_build.sh
```diff
@@ -161,6 +161,32 @@ test_cmake_select_ops_in_yaml() {
...
+    echo "Exporting MobilenetV2"
+    ${PYTHON_EXECUTABLE} -m examples.portable.scripts.export --model_name="add"
+
+    local example_dir=examples/selective_build
+    local build_dir=cmake-out/${example_dir}
+    # set MAX_KERNEL_NUM=22: 19 primops, add, mul
+    rm -rf ${build_dir}
+    retry cmake -DCMAKE_BUILD_TYPE=Release \
+            -DMAX_KERNEL_NUM=22 \
+            -DEXECUTORCH_SELECT_OPS_LIST='{"aten::add": [Scalar.Float.name]}'
```
lucylq: Should this be `OPS_DICT`? or are we re-using `OPS_LIST`?

examples/selective_build/CMakeLists.txt
```diff
@@ -66,6 +66,11 @@ option(EXECUTORCH_SELECT_ALL_OPS
        "Whether to register all ops defined in portable kernel library." OFF
 )
 
+# Option to enable parsing ops and dtypes from json formatted dictionary
+option(EXECUTORCH_SELECT_OPS_FROM_DICT
+       "Enable op selection from json formattting string during build." OFF
```
lucylq, line 71: Could we add a link to the supported dtype strings, e.g. https://github.com/pytorch/executorch/blame/0211a0346455f5b1ce445c2bdf6fce89a9aa04c9/shim_et/xplat/executorch/codegen/codegen.bzl#L51

^ actually I think this list is too broad and not all supported in ET 😅

BujSet: I think there are actually only 8 dtypes that are supported and tested. I remember seeing a list of them somewhere in the code base, will note them here.

PR #12885, opened 2025-07-27
Name Correction op_mean

BujSet:
### Summary

Operator and dtype selective build requires operator names to match operators in use. This changes corrects the name for `mean` operator. Change is similar to that in #12755.


lucylq (approved):

lucylq:
thanks for fixing these!

PR #12890, opened 2025-07-28
NXP backend: Make the flow robust against input/output swapping.

jirioc:
### Summary
The NXP backend is now robust against the swapping the order of inputs/outputs in the model converter.

Release notes: NXP

### Test plan
test_neutron_backend.py tests this feature

### DIFF
--- a/backends/nxp/nxp_backend.py
+++ b/backends/nxp/nxp_backend.py
@@ -245,19 +245,23 @@ class PayloadComposer:
 
         return f"{array.size}s{self._padding_format_string_for_array(array)}"
 
-    def _create_payload_header(self, io_formats) -> np.ndarray:
+    def _create_payload_header(self, io_formats, neutron_artifacts) -> np.ndarray:
         """
         Create bytes header for returned payload. It contains information about
         input and output tensor formats. Tensors are ordered based on graph signature
         of ExportedProgram. Header schema:
 
-        +----------------------------------+-----------------------------------+
-        | Input TensorFormats length (1B)  | Output TensorFormats length (1B)  |
-        +----------------------------------+-----------------------------------+
-        | 1st input tensor format (1B)     | [nth* input tensor format (1B)]   |
-        +----------------------------------+-----------------------------------+
-        | 1st output tensor format (1B)    | [nth* output tensor format (1B)]  |
-        +----------------------------------+-----------------------------------+
+        +----------------------------+-----------------------------+------------------------+
+        | Neutron inputs length (1B) | Neutron outputs length (1B) | Input args length (1B) |
+        +----------------------------+-----------+-----------------+------------------------+
+        | 1st input tensor format (1B)           | [nth* input tensor format (1B)]          |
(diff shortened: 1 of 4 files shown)


JakeStevens:
NXP build failing

StrycekSimon:
Thanks, I noticed it earlier but I was already out of office. I will fix it.

backends/nxp/nxp_backend.py
```diff
@@ -245,39 +245,67 @@ def _format_string_for_array(self, array: np.ndarray) -> str:
...
         :return: Bytes representation of payload header.
         """
         inputs = io_formats["inputs"]
         outputs = io_formats["outputs"]
 
-        assert len(inputs) < 256, "Models with more than 255 inputs are not supported."
         assert (
-            len(outputs) < 256
+            len(neutron_artifacts.input_indices) < 256
+        ), "Models with more than 255 inputs are not supported."
```
robert-kalmar, line 274: NIT: This is actually a valid case, not a "programming error". We just use 8bit field to encode input indices. Consider using ValueError instead of assert.

backends/nxp/nxp_backend.py
```diff
@@ -245,39 +245,67 @@ def _format_string_for_array(self, array: np.ndarray) -> str:
...
+        assert (
```
robert-kalmar, line 275: NIT: This is actually a valid case, not a "programming error". We just use 8bit field to encode input indices. Consider using ValueError instead of assert.

robert-kalmar (approved):
Looks good to me. The CI failures seems irrelevant to the changes in this PR. 
Rebased to main to have more up to date test results.

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -25,37 +25,53 @@ namespace neutron {
...
+     | 1st output map (1B)                    | [nth* output map (1B)]                   |
+     +----------------------------------------+------------------------------------------+
 */
+// clang-format on
 #define ITEM_SIZE 1 // 1 Byte
 #define INPUT_TENSOR_FORMAT_LEN_POS 0
 #define OUTPUT_TENSOR_FORMAT_LEN_POS 1
-#define INPUT_TENSOR_FORMAT_ARRAY_ADDR(base) (base + 2 * ITEM_SIZE)
+#define INPUT_ARGS_LEN_POS 2
+#define INPUT_TENSOR_FORMAT_ARRAY_ADDR(base) (base + 3 * ITEM_SIZE)
```
digantdesai, line 47: This is ok for now but in the future consider FC/BC issues when updating serialization formats.

digantdesai (approved):

backends/nxp/runtime/NeutronBackend.cpp
jirioc: We hope that the format is now general enough to cover the future scenarios.

digantdesai: Says like a true developer, until a second before breaking BC :-p

PR #12891, opened 2025-07-28
Arm backend: Decrease scratch buffer for Dedicated_Sram

AdrianLundell:
For Dedicated_Sram, both ET_ARM_BAREMETAL_METHOD_ALLOCATOR_POOL and ET_ARM_BAREMETAL_SCRATCH_TEMP_ALLOCATOR_POOL shares the DDR memory area with the parameters of the network.

The DDR is 256MiB, so if we allocate 60MiB + 128MiB for buffers, that only leaves 68Mib for the parameters, severely limiting the size of network we are able to run.


Sebastian-Larsson (approved):

PR #12894, opened 2025-07-28
Extend cat op for complex dtype

pssrawat:
Summary: Need complex cat op for live translation

Differential Revision: D78934592

### DIFF
--- a/kernels/portable/cpu/op_cat.cpp
+++ b/kernels/portable/cpu/op_cat.cpp
@@ -56,27 +56,58 @@ Tensor& cat_out(
   const size_t ninputs = tensors.size();
 
   const auto out_type = out.scalar_type();
-  ET_SWITCH_REALHBBF16_TYPES(out_type, ctx, "cat.out", CTYPE_OUT, [&] {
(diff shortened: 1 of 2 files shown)


kernels/portable/cpu/op_cat.cpp
```diff
@@ -56,27 +56,57 @@ Tensor& cat_out(
...
-    CTYPE_OUT* out_ptr = out.mutable_data_ptr<CTYPE_OUT>();
-    for (size_t i = 0; i < outer; ++i) {
-      for (size_t j = 0; j < ninputs; ++j) {
-        const auto in_type = tensors[j].scalar_type();
-        ET_SWITCH_REALHBBF16_TYPES(in_type, ctx, "cat.out", CTYPE_IN, [&] {
+  const bool out_is_complex =
+      executorch::runtime::isComplexType(out.scalar_type());
+
+  if (out_is_complex) {
+    // All the input tensors and output must have same dtype
```
manuelcandales: this is not a restriction of the ATen op. i.e. you can concat complex tensor of different dtype.
please add a TODO here, to allow mixed input dtype when output is complex.

manuelcandales (approved):

PR #12908, opened 2025-07-28
Evalue** -> Span

JacobSzwejbka:
Summary: We want to add safety checks inside the kernels on the amount of args they have received. To do that we need a pointer and a length instead of just a pointer. So swapping to span fixes that.

Differential Revision: D78904269


larryliu0820 (approved):
Thank you

PR #12909, opened 2025-07-28
draff diff for pybindings task #12790

jasshone:
### Summary
Tries to add pybindings for llm runner class 

Fixes ##12790

### Test plan
functionality does not work, draff diff

### DIFF
--- a/extension/llm/runner/CMakeLists.txt
+++ b/extension/llm/runner/CMakeLists.txt
@@ -15,7 +15,7 @@
 # ~~~
 # It should also be cmake-lint clean.
 #
-
+message(STATUS "Building LLM Runner")
 if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
 endif()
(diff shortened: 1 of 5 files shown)


extension/llm/runner/CMakeLists.txt
```diff
@@ -52,9 +52,40 @@ add_subdirectory(
...
+  )
+
+  target_include_directories(runner PRIVATE ${TORCH_INCLUDE_DIRS})
+  target_compile_options(
+    runner PUBLIC -Wno-deprecated-declarations -fPIC -frtti -fexceptions
+  )
+  target_link_libraries(runner PRIVATE ${runner_dep_libs})
+
+  install(TARGETS extension_llm_runner
+          LIBRARY DESTINATION executorch/extension/training/pybindings
```
larryliu0820: Don't install it in training directory. Install it under extension/llm/runner

jasshone: oh thanks, will fix

setup.py
```diff
@@ -791,6 +792,11 @@ def run(self):  # noqa C901
             modpath="executorch.extension.training.pybindings._training_lib",
             dependent_cmake_flags=["EXECUTORCH_BUILD_EXTENSION_TRAINING"],
         ),
+        BuiltExtension(
+            src="extension/llm/runner/python_bindings*",  # @lint-ignore https://github.com/pytorch/executorch/blob/cb3eba0d7f630bc8cec0a9cc1df8ae2f17af3f7a/scripts/lint_xrefs.sh
```
larryliu0820: 1. your path is incorrect, see comment above about install location
2. your output filename is incorrect, it should be something like `runner.so` named after your pybind11 module name.

PR #12910, opened 2025-07-28
[EXIR] Add dim order aware clone operator

### LINKED ISSUE #12645: Add dim order variant clone operator

cc. @digantdesai 

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

cc @JacobSzwejbka @angelayi

keyprocedure:
### Summary
Currently, clone ops are removed during export as no-ops, causing memory layout (dim order) changes to be lost. This can cause backend failures, incorrect outputs when ops expect specific layouts, and performance degradation.

This PR adds a dim order aware clone op, `clone_dim_order`, that preserves memory layout changes by explicitly storing dim order information. It's implemented by replacing standard clone ops with this variant during export and updating the clone removal transform to preserve clones that change layout.

Fixes #12645 

### Test plan
Tests were added to verify:
- `clone_dim_order` kernel implementation
- Preservation of dim order changes
- Removal of no-op clones (dim order not changed)

All tests passed via:
`python -m unittest exir.tests.test_memory_format_ops_pass`

### DIFF
--- a/backends/transforms/remove_clone_ops.py
+++ b/backends/transforms/remove_clone_ops.py
@@ -13,11 +13,24 @@ from executorch.exir.pass_base import ExportPass, PassResult
 
 def remove_clone_ops(graph: torch.fx.Graph) -> torch.fx.Graph:
     """
-    Remove clone op nodes and replace uses with parent node.
+    Remove clone op nodes that have the same dim_order as their input and replace their uses with the input node.
     """
     clone_op = exir_ops.edge.aten.clone.default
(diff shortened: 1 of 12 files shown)


Gasoonjia:
hi @keyprocedure thanks for your quick fix!

Is there any chance you can split this PR into smaller ones? It may be better for us to review!

My mental model is:
One pr for aot updates for dim order variant clone and its test cases,  one for dim order variant clone runtime operator and its test cases, and third is `remove_clone_op_pass`.

keyprocedure:
@Gasoonjia 
No problem!

That organization makes sense, I'll split this into 3 PRs.

PR #12929, opened 2025-07-28
Re-order the model list in CI alphabetical order

cccclai:
Summary:
Re-order the model in the list in alphabetical order so it's easy to track

Rollback Plan:

Differential Revision: D79118824


digantdesai (approved):
🙏 

Leave a note saying keep this in order would keep it so :)

PR #12930, opened 2025-07-28
Enable conv_former model in CI

cccclai:
Summary:
Add conv_former to CI

Rollback Plan:

Differential Revision: D79120806


digantdesai (approved):
Thanks!

Issue #12933, opened 2025-07-28
Modularize export recipes


abhinaykukkadapu:
### 🚀 The feature, motivation and pitch

Adds support for #(7) in the RFC #12660

7. Modularization of ExportRecipe
Split current ExportRecipe into piece-meal recipes such asExportRecipe, QuantizationRecipe, LoweringRecipe and have a root recipe for example: CompileRecipe which takes all the above. This might require re-writing most of the current ExportRecipe component, so not sure if this is needed, we may just get away with doing just (5) - Supporting ExportedProgram which skips torch.export.export() and SourceTransform stages which might be just sufficient.

PR #12937, opened 2025-07-28
[Executorch][Export][2/N] Add to_edge and to_backend stages

abhinaykukkadapu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #12938
* __->__ #12937
* #12936

Address (6) in the rfc: https://github.com/pytorch/executorch/issues/12660

1. Adds stage implementations for `to_edge` and `to_backend`
2. Adds unit tests for the two stages
3. Adds these two stages in the validation pipeline.

Fixes #12932

Differential Revision: [D79120576](https://our.internmc.facebook.com/intern/diff/D79120576/)

PR #12936, opened 2025-07-28
[ExecuTorch][Export][1/N] Export API pipeline re-architecture, making it composable

abhinaykukkadapu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #12938
* #12937
* __->__ #12936

RFC: https://github.com/pytorch/executorch/issues/12660

This diff introduces composable architecture for the export pipeline.

Changes:
1. Introduces notion of `PipelineArtifact` which encompasses stage artifacts and run context necessary to execute a stage and the pipeline.
1. Move all stages to new file to better architect the component
1. Export api will now accept `pipeline_stages` and honors the sequence if provided otherwise fallback to default sequence: `source_transform -> quantize -> torch.export -> to_edge_transform_and_lower -> to_executorch`
1. Validate if pipeline provided is valid, error out if not.
1. Add unittests to test pipeline sequence and stages

With this one can execute a partial pipeline, for example, one can just do `TORCH_EXPORT -> TO_EDGE_TRANSFORM_AND_LOWER -> TO_EXECUTORCH`

Current limitation:
1. `TORCH_EXPORT` stage is mandatory to avoid someone passing incorrect input such as `ExportedProgram` instead of nn.module, this enforcement will anchor the pipeline to have expected stages to run.
   - This limitation will be removed if we choose to add the support to take `ExportedProgram` as input.

Note:
1. this diff is large because I moved stages to separate file and added bunch of unittests but core functionality that is added can be reviewed in `export.py` file. CC: @digantdesai
2. Export component was supposed to mimic stages design in tester and now it is.

Fixes: #12928

Differential Revision: [D79120574](https://our.internmc.facebook.com/intern/diff/D79120574/)

Issue #12660, opened 2025-07-19
[RFC][Short] More on ExportRecipes


abhinaykukkadapu:
### Discussed in https://github.com/pytorch/executorch/discussions/12553

<div type='discussions-op-text'>

<sup>Originally posted by **abhinaykukkadapu** July 16, 2025</sup>
This discussion outlines few proposals regarding the ExportRecipes (currently experimental). These requirements have emerged from various discussions, and I want to consolidate them here for further discussion and prioritization:

### 1. API Naming
There is potential confusion between `torch.export.export()` and `executorch.export()`, especially since the latter encompasses additional functionalities such as quantization, lowering, and serialization. A previously proposed solution is to rename `executorch.export(..) -> executorch.compile(..)`. This change could enhance clarity, but it may also be unnecessary. I would like to gather your thoughts on whether a renaming is warranted.

### 2. Support for Additional Backend Recipes
Support additional backend recipes
- XNNPack (supported)
- CoreML
- QNN

### 3. QAT Quantization support
Currently, our recipes support Post-Training Quantization (PTQ). Should we extend this support to include Quantization-Aware Training (QAT)? Additionally, which backends currently support QAT, or we should at least transparently error out if QAT based recipe is passed?

### 4. Differentiating Calibration Inputs vs Example Inputs
The proposal is to add a new parameter, calibration_inputs, to the API alongside example_inputs. This parameter would explicitly indicate to users that the data provided in calibration_inputs is used for quantization purposes, while example_inputs are solely for export.

### 5. Support for `ExportedProgram`
At present, the API requires an nn.Module and assumes that the export stage is always run within the scope of the recipe. The proposal is to decouple this to accept either an ExportedProgram or an nn.Module. This change would help use cases such as [HuggingFace Optimum](https://github.com/huggingface/optimum-executorch/blob/cf33d68045dd1dea103bc62325d18a7bb1adce42/optimum/exporters/executorch/recipes/xnnpack.py), where the ExportRecipe can be utilized for post export stages such as graph quantization, lowering. This would provide greater flexibility for models to use torch.export.export(...) in a manner that suits their needs but use recipes for just quantization and lowering.

### 6. Support to_edge -> to_backend -> to_executorch along with to_edge_transform_and_lower
Currently, the `executorch/export` supports only the `to_edge_transform_and_lower` API. Is there a need to support `to_edge() -> to_backend() -> to_executorch()`?

### 7. Modularization of ExportRecipe
[...]

PR #12938, opened 2025-07-28
[Executorch][Export][3/N] Modularize export recipes

abhinaykukkadapu:
* __->__ #12938
* #12937
* #12936

Addresses (7) in the rfc: https://github.com/pytorch/executorch/issues/12660

Changes:
1. Add data class called `LoweringRecipe`
2. Modify current xnnpack recipes to use lowering recipes

Fixes: #12933

Differential Revision: [D79120575](https://our.internmc.facebook.com/intern/diff/D79120575/)


export/export.py
```diff
@@ -189,18 +193,42 @@ def _build_default_stages(self) -> Dict[StageType, Stage]:
...
+                    transform_passes=(
+                        self._lowering_recipe.edge_transform_passes
+                        if self._lowering_recipe
+                        else None
+                    ),
+                    compile_config=(
+                        self._lowering_recipe.edge_compile_config
+                        if self._lowering_recipe
+                        else None
+                    ),
```
digantdesai: Nit: new constructor? It can be done for other stages as well.
```suggestion
stage = EdgeTransformAndLowerStage.from_recipe(self._loweing_recipe)
```

export/recipe.py
```diff
@@ -87,6 +87,26 @@ def get_quantizers(self) -> Optional[List[Quantizer]]:
         return self.quantizers
 
 
+@dataclass
+class LoweringRecipe:
```
digantdesai, line 93: Nit: name can be improved, seems a bit generic compared to `ExportRecipe`

digantdesai (approved):
Stamping to unblock. Left some comments.

export/recipe.py
abhinaykukkadapu: @digantdesai I want this to work with either to_edge -> to_backend or to_edge_transform_lower, and broadly we are lowering here, do you think `EdgeLoweringRecipe` make sense?

export/export.py
abhinaykukkadapu: @digantdesai  nice idea, will update.

larryliu0820 (approved):

PR #12941, opened 2025-07-29
Refactoring Portable Operators to Standardize op_name Format

BujSet:
### Summary
Minor refactor to standardize how operator names are printed when debugging. Especially useful when hitting a missing operator error.


lucylq (approved):
Thanks for the refactoring work!

PR #12952, opened 2025-07-29
Arm backend: Handle None arg for creating empty TosaArg

per:
### Summary
In order to get the same number and order of arguments from fx.Node.args and the list[TosaArg] used in nodevisitor, we need to allow for empty (None) args. The empty TosaArg is represented with fields set to None and empty string for name.


Sebastian-Larsson (approved):

PR #12954, opened 2025-07-29
Arm backend: Use correct name when indexing state_dict

Sebastian-Larsson:

### DIFF
--- a/backends/arm/_passes/unsqueeze_scalar_placeholders_pass.py
+++ b/backends/arm/_passes/unsqueeze_scalar_placeholders_pass.py
@@ -7,6 +7,7 @@
 
 import torch
 from executorch.exir.pass_base import ExportPass, PassResult
+from torch._export.utils import is_buffer, is_param
 
 
 class UnsqueezeScalarPlaceholdersPass(ExportPass):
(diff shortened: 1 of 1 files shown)


Sebastian-Larsson:
CI failures seem unrelated

backends/arm/_passes/unsqueeze_scalar_placeholders_pass.py
```diff
@@ -19,23 +20,27 @@ def __init__(self, exported_program):
...
-        )
-
     def call(self, graph_module: torch.fx.GraphModule):
         for node in graph_module.graph.nodes:
             if node.op != "placeholder":
                 continue
             rank = node.meta["val"].dim()
             if rank == 0:
-                if not self._is_inputs_to_buffers_or_parameters(node):
+                if is_buffer(self.exported_program, node):
```
digantdesai, line 29: nit make a util `def get_name(self, node)`?

digantdesai (approved):

backends/arm/_passes/unsqueeze_scalar_placeholders_pass.py
Sebastian-Larsson: Local to this file? Seems a little redundant to me, I suppose that would look something like this

```
def get_name(self, node):
     if is_buffer(self.exported_program, node):
         return self.exported_program.graph_signature.inputs_to_buffers[node.name]
     # Reapeat for param etc.
     ...
```

And this is done 2 times only.

I think it would belong in torch._export.utils like is_buffer etc.

AdrianLundell (approved):

backends/arm/_passes/unsqueeze_scalar_placeholders_pass.py
digantdesai: even local would make it cleaner. Feel free to put up a PR against _export.utils if you feel like others are doing this or should do it this way.

PR #12955, opened 2025-07-29
Arm backend: Move the portable kernels section to bigger memory

gggekov:
This patch moves the portable kernels from the ITCM to the BRAM so that we don't overflow on the ITCM when linking the application


digantdesai (approved):
This is OK. FWIW, we added PTE based dtype selective build, in addition to existing per op selective build to further reduce the size of these `op_*_cpp.obj` files :)

FYI @BujSet

Sebastian-Larsson (approved):
Unrelated CI errors. Approved

digantdesai:
@GeorgeARM do we need this also for CS320 for U85?

gggekov:
On the CS320, the .text section is stored in the BRAM and we have 2MB of BRAM, so we haven't seen link time failures due to building with more portable kernels. 
On the CS300, the .text section is in the ITCM and we have 512KB of ITCM, hence we saw link time errors on some models, that's why I've only changed the placement of the portable kernels for the CS300.

PR #12971, opened 2025-07-29
[EXIR] Register _clone_dim_order op and map aten.clone

### LINKED ISSUE #12645: Add dim order variant clone operator

keyprocedure:
### Summary
This is PR 2 of 3 implementing a dim order aware clone op. 

This PR registers the new `_clone_dim_order` op and maps `aten.clone` ops to `dim_order_ops._clone_dim_order` in EXIR during export to preserve memory layout changes (contiguous/channels_last). It also updates Core ML and ARM backends to handle the new clone op.

Related PRs:
- PR 1: [#12974](https://github.com/pytorch/executorch/pull/12974) - Add `_clone_dim_order` portable kernel
- PR 3: [#12976](https://github.com/pytorch/executorch/pull/12976) -  Update RemoveCloneOpsTransform to be dim order aware

Fixes #12645  

### Test plan
- Operator level tests to verify kernel behavior for layout preservation and changes.
- Graph level checks to confirm that clone mapping occurs.
- End to end tests to validate that functional clone behavior is unchanged.

All tests pass via:
`python -m unittest exir.tests.test_memory_format_ops_pass`
`python -m unittest backends.apple.coreml.test.test_torch_ops`
`pytest backends/arm/test/ops/test_clone.py`
`pytest backends/arm/test/passes/test_remove_clone_pass.py`


exir/tests/test_memory_format_ops_pass.py
```diff
@@ -389,3 +390,17 @@ def test_mobilenet_v3_xnnpack(self) -> None:
                 rtol=1e-3,
             ),
         )
+
+    def test_op_clone_dim_order_registration(self):
```
Gasoonjia: Can you follow other testcases way (e.g, `test_op_dim_order_propagation`), to reuse the memory_format_test_runner function and test at least two different scenarios: one for maintaining dim order, the other is converting the dim order?

Gasoonjia (commented):
LGTM! Left a minor feedback about test coverage!

exir/tests/test_memory_format_ops_pass.py
keyprocedure: I can definitely add more tests, the reason I didn't here was because the kernel implementation wasn't in place for this PR. I'll work on adding additional coverage.

keyprocedure:
@Gasoonjia splitting the `clone_dim_order` op implementation into 3 PRs was a great suggestion, it makes it much easier to isolate and debug issues.

The CI failures are mainly due to two issues:
1.  `clone` is now mapped to `clone_dim_order`, so tests expecting `executorch_exir_dialects_edge__ops_aten_clone_default` are failing because they now see `executorch_exir_dialects_edge__ops_dim_order_ops__clone_dim_order_default`.
2. Because of the first issue, the lack of a kernel implementation for `clone_dim_order` in this PR is also triggering failures. 

Do you have any suggestions for how we should handle the current approach of mapping `clone` to `clone_dim_order`?

Gasoonjia:
Hi @keyprocedure thanks for your update!

For places checking the `executorch_exir_dialects_edge__ops_aten_clone_default`, please update them to check `executorch_exir_dialects_edge__ops_dim_order_ops__clone_dim_order_default ` as long as the `compile_config.skip_dim_order == False` in their pipeline.

> the lack of a kernel implementation for clone_dim_order

can you share me more insights here? I think you've create clone_dim_order kernel in aot side by reusing aten clone op.

keyprocedure:
@Gasoonjia during export `_clone_dim_order` does use ATen's `clone` implementation because of the Python fallback, but at runtime the graph contains `_clone_dim_order` and since no kernel is registered for it, it fails with:
`[operator_registry.cpp:256] kernel 'dim_order_ops::_clone_dim_order.out' not found.`

This is also why I placed the tests using `memory_format_test_runner` in the kernel PR instead of this one. 

What are your thoughts?

Gasoonjia:
@keyprocedure  thanks for detailed reply!

let's swap the order of PRs. We can first finish the runtime op, then aot replacement, finally RemoveCloneOpsTransform.
That would solve the runtime operator requirement issue.

keyprocedure:
> @keyprocedure thanks for detailed reply!
> 
> let's swap the order of PRs. We can first finish the runtime op, then aot replacement, finally RemoveCloneOpsTransform. That would solve the runtime operator requirement issue.

That's a great idea! Should I move the tests from the runtime/kernel PR (since the new op won't be registered yet) to the aot replacement PR?

Gasoonjia:
> > @keyprocedure thanks for detailed reply!
> > let's swap the order of PRs. We can first finish the runtime op, then aot replacement, finally RemoveCloneOpsTransform. That would solve the runtime operator requirement issue.
> 
> That's a great idea! Should I move the tests from the runtime/kernel PR (since the new op won't be registered yet) to the aot replacement PR?

Yes please! More specific:

for the runtime pr, it should only contain:

1. the portable clone_dim_order op
2. the operator test itself

for aot pr, it should contains both graph-level check, and the end2end test using pybinding.

keyprocedure:
Hi @Gasoonjia 

I’ve updated the PRs, the [kernel PR](https://github.com/pytorch/executorch/pull/12974) is ready for review.

The CI failures from this PR that weren’t due to the missing `_clone_dim_order` kernel are all ARM related runtime failures:
- `Expected to not find "torch.ops.higher_order.executorch_call_delegate" but found it`
- `Expected to find "torch.ops.higher_order.executorch_call_delegate" but did not find it`
- `Expected out tensor to have dtype signed char, but got float instead`
- `Expected to find "executorch_exir_dialects_edge__ops_aten_clone_default" but did not find it`

It seems like these failures share the same root cause: `aten.clone` being replaced by `_clone_dim_order`, but since the failures are spread across quite a few tests, I'm not sure which strategy we should use to address this.
Does it make sense to individually try to correct each failing test or is there a way to gate the ARM backend to still use the 
`aten.clone` op? 

Any insights or guidance would be helpful

Gasoonjia:
@keyprocedure 
Thanks for sharing me the error msg!
I've left some comment for the kernel PR, please take a look.

For the issue in this PR, can you share me the code pointer to the issue? Like which line raising that?

keyprocedure:
<details>

<summary>There were a total of 49 failing tests for the initial CI run, so I listed out the test and line number of each failure:</summary>

### unittest-arm-backend-with-no-fvp (test_pytest_ops) / linux-job
19 failures
```
backends/arm/test/models/stable_diffusion/test_T5EncoderModel.py:100:
    TestT5EncoderModel.test_T5EncoderModel_tosa_BI

"AssertionError: Invalid partition, found dependency cycles"
```
```
backends/arm/test/models/stable_diffusion/test_T5EncoderModel.py:80:
    TestT5EncoderModel.test_T5EncoderModel_tosa_MI

"IndexError: list index out of range"
```
```
backends/arm/test/models/stable_diffusion/test_CLIPTextModelWithProjection.py:76:
    TestCLIPTextModelWithProjection.test_CLIPTextModelWithProjection_tosa_MI

backends/arm/test/models/stable_diffusion/test_SD3Transformer2DModel.py:128:
    TestSD3Transformer2DModel.test_SD3Transformer2DModel_tosa_BI

backends/arm/test/models/stable_diffusion/test_SD3Transformer2DModel.py:105:
    TestSD3Transformer2DModel.test_SD3Transformer2DModel_tosa_MI

backends/arm/test/models/stable_diffusion/test_vae_AutoencoderKL.py:74:
    TestAutoencoderKL.test_AutoencoderKL_tosa_BI

backends/arm/test/models/stable_diffusion/test_vae_AutoencoderKL.py:55:
    TestAutoencoderKL.test_AutoencoderKL_tosa_MI

backends/arm/test/models/test_conformer.py:60:
    test_conformer_tosa_MI

backends/arm/test/models/test_deit_tiny_arm.py:45:
    test_deit_tiny_tosa_MI

backends/arm/test/models/test_deit_tiny_arm.py:58:
    test_deit_tiny_tosa_BI

backends/arm/test/models/test_dl3_arm.py:44:
    test_dl3_tosa_MI

backends/arm/test/models/test_dl3_arm.py:57:
    test_dl3_tosa_BI

backends/arm/test/models/test_mobilenet_v2_arm.py:45:
    test_mv2_tosa_MI

backends/arm/test/models/test_mobilenet_v2_arm.py:60:
    test_mv2_tosa_BI[per_channel_quantization=true]
    test_mv2_tosa_BI[per_channel_quantization=false]

backends/arm/test/models/test_mobilenet_v2_arm.py:78:
    test_mv2_u55_BI[per_channel_quantization=true]
    test_mv2_u55_BI[per_channel_quantization=false]

backends/arm/test/models/test_mobilenet_v2_arm.py:96:
    test_mv2_u85_BI[per_channel_quantization=true]
    test_mv2_u85_BI[per_channel_quantization=false]

"RuntimeError: Expected to not find "torch.ops.higher_order.executorch_call_delegate" but found it"
```

The CI log also shows multiple TOSA partitioning rejections due to _clone_dim_order not being in BaseTOSASupportList.


***
### unittest-arm-backend-with-no-fvp (test_pytest_models) / linux-job
30 failures
```
backends/arm/test/ops/test_alias_copy.py:63:
    test_alias_tosa_BI[1d_ramp]
    test_alias_tosa_BI[2d_ones]
    test_alias_tosa_BI[3d_rand]
    test_alias_tosa_BI[4d_zeros]

"RuntimeError: Expected out tensor to have dtype signed char, but got float instead"
```

```
backends/arm/test/ops/test_clone.py:58:
    test_clone_tosa_MI[ones_1D_10]
    test_clone_tosa_MI[ones_1D_50]
    test_clone_tosa_MI[rand_1D_20]
    test_clone_tosa_MI[rand_2D_10x10]
    test_clone_tosa_MI[rand_3D_5x5x5]
    test_clone_tosa_MI[rand_4D_2x3x4x5]
    test_clone_tosa_MI[large_tensor]

backends/arm/test/ops/test_clone.py:69:
    test_clone_tosa_BI[ones_1D_10]
    test_clone_tosa_BI[ones_1D_50]
    test_clone_tosa_BI[rand_1D_20]
    test_clone_tosa_BI[rand_2D_10x10]
    test_clone_tosa_BI[rand_3D_5x5x5]
    test_clone_tosa_BI[rand_4D_2x3x4x5]
    test_clone_tosa_BI[large_tensor]

backends/arm/test/passes/test_remove_clone_pass.py:43:
    test_remove_clone_tosa_BI

"RuntimeError: Expected to find "torch.ops.higher_order.executorch_call_delegate" but did not find it"
```
```
backends/arm/test/ops/test_multihead_attention.py:48:
    test_multihead_attention_tosa_MI[rand_2d]
    test_multihead_attention_tosa_MI[randn_2d]
    test_multihead_attention_tosa_MI[randn_3d]

backends/arm/test/ops/test_multihead_attention.py:65:
    test_multihead_attention_tosa_BI[rand_2d]
    test_multihead_attention_tosa_BI[randn_2d]
    test_multihead_attention_tosa_BI[randn_3d]

backends/arm/test/ops/test_multihead_attention.py:108:
    test_multihead_attention_u85_BI[rand_2d]
    test_multihead_attention_u85_BI[randn_2d]
    test_multihead_attention_u85_BI[randn_3d]

backends/arm/test/ops/test_repeat.py:74:
    test_repeat_tosa_MI[interleave_int_3_x_1]

"RuntimeError: Expected to not find "torch.ops.higher_order.executorch_call_delegate" but found it"
```
```
backends/arm/test/ops/test_repeat.py:86:
    test_repeat_tosa_BI[interleave_int_3_x_1]

"RuntimeError: Expected out tensor to have dtype signed char, but got float instead"
```
</details>

keyprocedure:
I think we should land the kernel PR before addressing any new failures from the latest CI run in this PR since this PR is dependent on the kernel existing for `_clone_dim_order`.

Gasoonjia:
sure let's land prs one by one

exir/tests/test_memory_format_ops_pass.py
```diff
@@ -153,6 +205,29 @@ def test_op_dim_order_propagation_ambiguous(self) -> None:
         except AmbiguousDimOrderError:
             pass  # Expected error
 
+    def test_op_clone_dim_order_graph_replacement(self):
```
Gasoonjia: Thanks for adding this test! It means you have strong understanding of what we are doing in MemoryFormatPass!
Just wanna share with you that such check will be automatically happen inside `to_edge` function,  as long as the `EdgeCompileConfig._check_ir_validity == True`.   You can see https://github.com/pytorch/executorch/blob/main/exir/verification/verifier.py#L313 for more info.

Gasoonjia (commented):
LGTM thanks for your wonderful work!
Will stamp the PR after ci passed!

exir/tests/test_memory_format_ops_pass.py
keyprocedure: Oh okay, that makes sense. I'll remove this test since the other tests I added use `memory_format_test_runner`, which calls `to_edge` and `_check_ir_validity` defaults to True.

keyprocedure:
@Gasoonjia 
Could you take a look at the changes I made to fix the ARM failures from the last CI run when you have a chance? I added `_clone_dim_order` to the TOSA support check registry and updated the tests. Is TOSA available publicly? I couldn’t run the tests locally because I couldn’t source the package.

backends/apple/coreml/compiler/torch_ops.py
```diff
@@ -67,6 +68,28 @@ def _to_dim_order_copy(context, node):
...
+
+    # In CoreML, dim_order.val will be a ndarray, so we convert it to a list to check memory format.
+    dim_order = [int(d) for d in dim_order.val]
+    memory_format = get_memory_format(dim_order)
+    assert (
+        memory_format == _torch.contiguous_format
+    ), "Only contiguous memory format is supported in CoreML"
+
+    # Since CoreML only supports contiguous format, no dim_order preservation is needed. Treat this as a no-op clone.
+    noop(context, node)
```
digantdesai, line 90: Don't we have to `clone`?

backends/arm/_passes/remove_clone_pass.py
```diff
@@ -14,7 +14,11 @@ class RemoveClonePass(ExportPass):
     """Remove all clones from graph_module"""
 
     def call_operator(self, op, args, kwargs, meta):
-        if op != exir_ops.edge.aten.clone.default:
+        clone_ops = (
+            exir_ops.edge.aten.clone.default,
```
digantdesai: why do we need `aten.clone` we should be removing this when producing the edge dialect graph, no?

backends/arm/operator_support/clone_dim_order_support.py
```diff
@@ -0,0 +1,76 @@
...
+    SupportedTOSAOperatorCheck,
+)
+from executorch.backends.arm.tosa_specification import TosaSpecification
+from executorch.exir.dialects._ops import ops as exir_ops
+
+logger = logging.getLogger(__name__)
+
+
+@register_tosa_support_check
+class CloneDimOrderSupport(SupportedTOSAOperatorCheck):
```
digantdesai, line 23: Did we remove support to partition aten.clone op?

backends/apple/coreml/compiler/torch_ops.py
```diff
@@ -67,6 +68,28 @@ def _to_dim_order_copy(context, node):
     to(context, node)
 
 
+@register_torch_op(
```
digantdesai, line 71: did we remove support to partition `aten.clone` op?

digantdesai (commented):
Looks good for the most part. Left some comments.

backends/apple/coreml/compiler/torch_ops.py
keyprocedure: I didn't explicitly remove support for `aten.clone`. From my understanding this change would only add support for `_clone_dim_order`, `aten.clone` is still supported and handled by the noop handler in `coremltools`.

backends/apple/coreml/compiler/torch_ops.py
keyprocedure: `clone` is in the `torch list` for the noop handler, so I assumed that during conversion all cloning was simply returning the input. Does that seem right, or am I missing something?

backends/arm/_passes/remove_clone_pass.py
keyprocedure: Because isn't `aten.clone` only mapped to `_clone_dim_order` if `_skip_dim_order=False`?

backends/arm/operator_support/clone_dim_order_support.py
keyprocedure: We didn't, `aten.clone` is still in the `BaseTOSASupportList` and supported. `CloneDimOrderSupport` would only apply when cloning takes place with `_skip_dim_order=False`, otherwise `aten.clone` would still exist.

keyprocedure:
@digantdesai it looks like most of the CI failures were due to `_clone_dim_order` not being fully registered in `TOSA`. I had missed including `clone_dim_order_support` in `operator_support/init.py`. I've added that registration and also registered a node visitor for `_clone_dim_order`. Does it make sense to lower the identity since only `contiguous` dim orders are supported for now?

backends/arm/_passes/remove_clone_pass.py
keyprocedure: But that's a good point. Can we assume `_skip_dim_order` is always `False` in ARM? If so, we can replace all instances of `aten.clone` with `_clone_dim_order`. If not, we'd need to make the clone tests check for the correct clone op based on the flag's value. I also assumed that all clones are removed in ARM regardless of whether the dim order changes, so I didn’t add a dim_order check in the removal pass.

Gasoonjia: > we should be removing this when producing the edge dialect graph, no

aten.clone will be replaced only when `_skip_dim_order = False`. If we do not force `Arm` turning on dim order op i would love to keep @keyprocedure's work.

backends/arm/test/misc/test_partition_decomposed_quantized_ops.py
```diff
@@ -38,7 +38,7 @@
 ]
 linear_residual_exir_op: list[str] = [
     "executorch_exir_dialects_edge__ops_aten_gelu_default",
-    "executorch_exir_dialects_edge__ops_aten_clone_default",
+    "executorch_exir_dialects_edge__ops_dim_order_ops__clone_dim_order_default",
```
Gasoonjia, line 41: Removing aten.clone here because the corresponding tests is dim order only? 
@digantdesai will `Arm` use `dim order` mandatorily? If we want to continue support both `aten.clone` and `dim_order.clone`, lets keep the original test while have a new test for `dim_order` case.
Same as other tests

Gasoonjia (commented):
LGTM, though there're some Arm question maybe we need to wait for @digantdesai  for final decision.

backends/arm/test/misc/test_partition_decomposed_quantized_ops.py
keyprocedure: Since ARM's pass manager doesn't seem to pass the `_skip_dim_order` flag, it defaults to `False`, so my strategy was to replace all instances of `aten.clone` with `_clone_dim_order` in the tests. Although it might make more sense to explicitly set `_skip_dim_order=True` in the `EdgeCompileConfig` calls in [arm_tester](https://github.com/pytorch/executorch/blob/1164fe4100cd82ab3c7dbfa2a0589fe6386d27d0/backends/arm/test/tester/arm_tester.py#L160) and undo my changes to the ARM test files. Then I can either continue with adding `_clone_dim_order` support and tests (following Gasoonjia’s suggestion) in TOSA for future use or leave it out for now.

Gasoonjia: yeah my point is we should make our update in line with Arm's target in a consistent way.

That is, if we expect dim-order-only in Arm backend, there should be no `aten.clone` in our PR and we should remove `aten.clone` supports.

However if we should support dim_order equals to both on and off, we may test both case for clone operator, though I'm ok to split it into several PRs for better structure.

keyprocedure: That makes sense, thanks for breaking down the options.

digantdesai: FYI see - https://github.com/pytorch/executorch/commit/135e875f2f0c9ee9704899c1208512d5fcf697bd

Also cc @oscarandersson8218  for what if no aten.clone support.

backends/apple/coreml/compiler/torch_ops.py
digantdesai: I see https://github.com/apple/coremltools/blob/be33582e2fd62885f15d22338cac9f777ad8119f/coremltools/converters/mil/frontend/torch/ops.py#L6352

backends/arm/_passes/remove_clone_pass.py
digantdesai: who else is using `_skip_dim_order = True`? If no one then we can remove it. Arm isn't using it, but supports it I guess. 

@Gasoonjia - we should remove this flag altogether :|

backends/apple/coreml/compiler/torch_ops.py
digantdesai, line 71: FYI @metascroy

backends/arm/operators/op_clone_dim_order.py
```diff
@@ -0,0 +1,43 @@
...
+    register_node_visitor,
+)
+from executorch.backends.arm.operators.operator_validation_utils import (
+    validate_num_inputs,
+)
+from executorch.backends.arm.tosa_mapping import TosaArg
+
+
+@register_node_visitor
+class CloneDimOrderVisitor(NodeVisitor):
```
oscarandersson8218: If RemoveClonePass works as it should there should be no need for a NodeVisitor as all clones should be removed when we end up here. Also, this node_visitor is not imported in` __init__.py` and is therefore not registered.

oscarandersson8218 (commented):
@keyprocedure Nice change and thanks for updating our backend!

backends/apple/coreml/compiler/torch_ops.py
keyprocedure: Yeah, that's what I was going off of. Since `clone` was treated as a no-op, so should `_clone_dim_order`?

backends/arm/test/misc/test_partition_decomposed_quantized_ops.py
keyprocedure: @digantdesai thanks for looking into this and sharing the `enable dim_order` PR. It looks like the best strategy is to replace support for `aten.clone` with `_clone_dim_order` in ARM since the `_skip_dim_order` flag will always be `False`, following @Gasoonjia's suggestion.

backends/arm/operators/op_clone_dim_order.py
keyprocedure: Thanks for pointing this out, I'll remove this NodeVisitor. Just to confirm, if a clone changes dim_order, should the RemoveClonePass still be expected to remove it?

oscarandersson8218: I think the check in [clone_dim_order_support.py](https://github.com/pytorch/executorch/pull/12971/files#diff-5ee32c35e3e50a72fd19ca082a70f5f7e928fe5a33b810268fe236d29bcdc520R65) check will make sure that we don't partition a clone that changes dim_order so it should be ok to me.

AdrianLundell: To expand on this, the plan for dim_order handling in the arm backend is to ignore all delegated dim_order ops (no matter if it changes dim_order or not) and just make sure that the input/output of the delegated sub-graph graph is the correct, since we do our own dim_order handling in https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/annotate_channels_last_dim_order_pass.py. The dtype casting of the to_dim_order_copy is kept though. 

We will not delegate partitions containing single dim_order ops to allow the use of dim_order ops in non delegated sub-graphs, similar to this PR: https://github.com/pytorch/executorch/pull/12995

keyprocedure: I really appreciate the context and resources on how dim_order ops are handled. This also helps with diagnosing the graph delegation CI failures.

Gasoonjia:
hi @keyprocedure looks like there're still several issues on ci. Do you mind take a look?

keyprocedure:
> hi @keyprocedure looks like there're still several issues on ci. Do you mind take a look?

Hi, yeah, currently working on them. Appreciate the check in :)

keyprocedure:
Hi @gasoonjia, can we try running CI again?

I was able to configure the TOSA test setup locally and most of the test failures were due to the input dtype gating that I had in [CloneDimOrderSupport](http://github.com/pytorch/executorch/pull/12971/files#diff-5ee32c35e3e50a72fd19ca082a70f5f7e928fe5a33b810268fe236d29bcdc520R23), which caused clone ops to not be partitioned, remaining in the subgraph, and failing the [tester.check_not tests](https://github.com/pytorch/executorch/blob/3db27cdc0909e8bbf09896c5edb05342f08aceb9/backends/arm/test/tester/test_pipeline.py#L397).

Gasoonjia:
@keyprocedure  Really appreciate your work mate! Started ci and let's see hwo things work!

keyprocedure:
@Gasoonjia it looks like we're good! 

The 3 CI failures seem unrelated:
- `build doc (buck2)` failed because of an upload error
- `test-static-llama-qnn-linux` had a `Failed to download LLVM` error. This job passed last run and I only made changes to ARM since then.
- `test-llama_runner_eager-linux` failed without a log (GitHub suggests: `FLAKY - The following job failed but was likely due to flakiness present on trunk`)

backends/apple/coreml/compiler/torch_ops.py
metascroy: This change looks OK to me, but do we have a test that covers it?

backends/apple/coreml/compiler/torch_ops.py
metascroy, line 90: @keyprocedure Why not just call clone(context, node) here?  It will be equivalent to noop now, but let's use clone if that's what the op is.

Just in case coremltools updates clone to be something other than a noop in future?

backends/apple/coreml/compiler/torch_ops.py
keyprocedure: I added a test for _clone_dim_order [here](https://github.com/pytorch/executorch/pull/12971/files#diff-c597913623f7a301d96407a880a9175b9f38b745367e4f8905ca410192a83da8R224). Is this what you had in mind?

backends/apple/coreml/compiler/torch_ops.py
keyprocedure: That makes a lot of sense! But I don't see clone exposed in [coremltools](https://github.com/apple/coremltools/blob/be33582e2fd62885f15d22338cac9f777ad8119f/coremltools/converters/mil/frontend/torch/ops.py#L6340) for import, it looks like it’s only handled as an alias to noop.

backends/apple/coreml/compiler/torch_ops.py
metascroy: Looks good

backends/apple/coreml/compiler/torch_ops.py
metascroy: I guess this is OK then

Gasoonjia:
ci gogogo!

Gasoonjia:
ci passed. PR LGTM. Thanks for your contribution! @keyprocedure 

 @digantdesai  @oscarandersson8218 @metascroy please let me know if you have any concerns, or I will stamp it.

Gasoonjia (approved):

Gasoonjia:
Merged! @keyprocedure

keyprocedure:
Appreciate all the feedback and reviews, glad we got this landed!

jackzhxng:
Sorry to be the bearer of bad news, but this broke some trunk tests. Let's revert and @keyprocedure can you try to get this in again with the `ciflow/trunk` label turned on?

Gasoonjia:
@jackzhxng  stamped the revert PR. 
btw why do we need to manually set the tag for ci test, instead of make it as default? 
is there any place we can find all tags that should be tagged for enabling all cis?

keyprocedure:
Thanks for the update @jackzhxng, I'll resubmit the PR with the additional label.

digantdesai:
> btw why do we need to manually set the tag for ci test, instead of make it as default?

money 🤑

PR #12972, opened 2025-07-29
Do not decompose view ops in CoreML partitioner

metascroy:
We have some evidence that preserving view ops can hurt performance in CoreML.

In addition, ET converts view ops to copy variants before lowering, but after checking op support, so we sometimes run into an issue where a view op is preserved, but its copy variant is unsupported.  This can lead to graph breaks.


billmguo (approved):

derekxu (approved):

metascroy:
The CI failures exist on main.

PR #12974, opened 2025-07-29
Add _clone_dim_order portable kernel

### LINKED ISSUE #12645: Add dim order variant clone operator

keyprocedure:
### Summary
This is PR 1 of 3 implementing a dim order aware clone op. 

Currently, clone ops are removed during export as no-ops, causing memory layout (dim order) changes to be lost. This can cause backend failures, incorrect outputs when ops expect specific layouts, and performance degradation. This set of PRs introduces a dim order aware clone op, `_clone_dim_order`, which preserves memory layout changes by explicitly storing dim order information. This is implemented by replacing standard clone ops with this variant during export and updating the clone removal transform to preserve clones that change layout.

This PR adds the portable CPU kernel for the `_clone_dim_order` op, implementing a clone variant that preserves dim order at runtime. The portable kernel validates dtype and layout compatibility, resizes the output tensor if needed, and performs an element wise clone of the tensors.

Note: A future PR will add the ATen kernel for `_clone_dim_order`.

[Resource for Dim Order in ExecuTorch](https://github.com/pytorch/executorch/issues/8037)

Related PRs:
- PR 2: [#13735](https://github.com/pytorch/executorch/pull/13735) - Register `_clone_dim_order` op and map `aten.clone`
- PR 3: [#12976](https://github.com/pytorch/executorch/pull/12976) -  Update RemoveCloneOpsTransform to be dim_order aware

Fixes #12645 

### Test plan
Added kernel runtime tests to verify:
- Tensors of all real dtypes are cloned correctly.
- Failure when input and output tensor shapes mismatch.
- Failure with unsupported memory formats.
- Failure when `non_blocking=true` since the portable kernel only supports blocking data transfer.
- Dynamic shape outputs are cloned with correct values.
- Layout conversions are cloned correctly for `contiguous` to `channels_last`, `channels_last` to `contiguous`, and `channels_last` is preserved.

All runtime tests pass via:
`build-ninja/kernels/test/portable_kernels_test`


exir/passes/dim_order_ops_registry.py
```diff
@@ -28,6 +28,14 @@
     "_empty_dim_order.out(int[] size, *, int[]? dim_order=None, Tensor(a!) out) -> Tensor(a!)"
 )
 
+lib.define(
```
Gasoonjia: It is ok to leave here since we are gonna need it in the future, but when we talk about adding portable kernels we mainly focus on the kernels in the runtime, specificly under `executorch/kernels/portable`.

Gasoonjia (commented):
@keyprocedure 

Good start! Thanks for your great work!

Beyond than what we have now, it would be awesome if we can have a runtime test for the new runtime operator. https://github.com/pytorch/executorch/blob/main/kernels/test/op__to_dim_order_copy_test.cpp is the test for to_dim_order_copy and you can use that as an example.

Also please mark this PR as Open instead of Draft if it is ready to review. Thx!

exir/passes/dim_order_ops_registry.py
keyprocedure: That makes sense, but I needed to register the operator here otherwise the tests I added fail since there is no Python side reference to `_clone_dim_order`

keyprocedure: Oh, should all the tests for this PR have been on the kernel side and not in [test_memory_format_ops_pass.py](https://github.com/pytorch/executorch/pull/12974/files#diff-06df0587c95ca44af02799ffd147b40e324ce68def041f25e14f98744294db9cR94)?

Gasoonjia: This PR should only focus on our runtime changes, so no python side will refer to our new operator. 

> should all the tests for this PR have been on the kernel side and not in [test_memory_format_ops_pass.py](https://github.com/pytorch/executorch/pull/12974/files#diff-06df0587c95ca44af02799ffd147b40e324ce68def041f25e14f98744294db9cR94)

Yes absolutely correct! This PR is only for portable kernel and its tests. Sorry for any misleading!

keyprocedure: I’ve added the runtime test and all tests passed locally. I couldn’t run the [`DynamicShapeUnbound`](https://github.com/pytorch/executorch/pull/12974/files#diff-1f34a8df9bb81264978ee6354f330e3880eef59c21eb21f6d607c92ba4407a87R267) test since it depends on `SupportedFeatures` and `supported_features.h` doesn’t seem to be generated in OSS builds.

kernels/test/op__clone_dim_order_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+      Tensor expected = tf_out.make(test_case.sizes, data_in);
+
+      // Verifies that the returned and output tensor from _clone_dim_order both
+      // match the original input (expected).
+      EXPECT_TENSOR_EQ(ret, output);
+      EXPECT_TENSOR_EQ(ret, expected);
+    }
+  }
+
+  /* %python
```
Gasoonjia: Please remove comments unrelated with tests

kernels/test/op__clone_dim_order_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+  opt_extra_params = "non_blocking, memory_format,"
+  out_args = "out_shape, dynamism"
+  dtype = "ScalarType::Float"
+  check = "EXPECT_TENSOR_EQ" */
+
+  // Helper for testing dynamic shape outputs.
+  void test_dynamic_shape(
+      const std::vector<int32_t>& out_shape,
+      enum torch::executor::TensorShapeDynamism dynamism) {
+    /* %python
```
Gasoonjia: Same here

kernels/test/op__clone_dim_order_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+    GTEST_SKIP() << "Dynamic shape unbound not supported.";
+  }
+  test_dynamic_shape(
+      {1, 1}, torch::executor::TensorShapeDynamism::DYNAMIC_UNBOUND);
+}
+
+TEST_F(OpDimOrderCloneTest, ContiguousToChannelsLast) {
+  TensorFactory<ScalarType::Float> tf;
+
+  Tensor x = tf.make_with_dimorder(
```
Gasoonjia, line 262: now x is using contiguous dim order (0,1,2,3) as default. Plz add a comment here for clarify

kernels/test/op__clone_dim_order_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+      {0.2432, 0.8184, 0.6138, 0.9680, 0.2257, 0.5248, 0.8206, 0.1112, 0.2548,
+       0.8766, 0.5361, 0.7357, 0.2799, 0.0393, 0.2715, 0.8513, 0.9655, 0.1079,
+       0.6002, 0.1595, 0.2029, 0.4405, 0.2446, 0.4439, 0.3491, 0.7026, 0.6560,
+       0.6590, 0.5346, 0.6970, 0.6982, 0.9217, 0.3866, 0.3179, 0.8456, 0.8529,
+       0.6372, 0.7185, 0.4492, 0.2516, 0.2345, 0.8530, 0.8923, 0.3007, 0.7203,
+       0.2924, 0.8560, 0.5546, 0.3961, 0.8963, 0.7695, 0.6909, 0.6978, 0.8416,
+       0.3597, 0.0911, 0.7719, 0.8151, 0.4296, 0.5552},
+      /*dim_order=*/{0, 2, 3, 1});
+
+  Tensor expected = tf.make_with_dimorder(
```
Gasoonjia, line 311: same as here for comment

Gasoonjia (commented):
LGTM! Some minor feedback but majority looks good!

exir/passes/dim_order_ops_registry.py
keyprocedure: Please disregard my previous comment about the missing `SupportedFeatures` dependency, the issue was with my local build setup. All tests pass now.

keyprocedure:
@Gasoonjia Everything runs fine locally but CI failed because of a missing dependency in [copy_ops_util.h](https://github.com/pytorch/executorch/pull/12974/files#diff-f91ccb7a0924492d2368a528c7629c3c8dba4d208881846b991f6af5e614db04R12):
`fatal error: 'executorch/kernels/portable/cpu/util/broadcast_util.h' file not found`.

This happened after I refactored a function into copy_ops_util.h and added the import. 
I added the import to [targets.bzl](https://github.com/pytorch/executorch/pull/12974/files#diff-8267675d4c12a0c353c45f694a6bf2fc149a0688478a6410a6c34f4ed784e327R152), but CI still fails. Do you have insights into what I could be missing?

keyprocedure:
> @Gasoonjia Everything runs fine locally but CI failed because of a missing dependency in [copy_ops_util.h](https://github.com/pytorch/executorch/pull/12974/files#diff-f91ccb7a0924492d2368a528c7629c3c8dba4d208881846b991f6af5e614db04R12): `fatal error: 'executorch/kernels/portable/cpu/util/broadcast_util.h' file not found`.
> 
> This happened after I refactored a function into copy_ops_util.h and added the import. I added the import to [targets.bzl](https://github.com/pytorch/executorch/pull/12974/files#diff-8267675d4c12a0c353c45f694a6bf2fc149a0688478a6410a6c34f4ed784e327R152), but CI still fails. Do you have insights into what I could be missing?

I think I got it: `broadcast_util` wasn't being exported, so I added it to `exported_deps` for `copy_ops_util.` 
Since I was building with CMake earlier, there weren't any dependency failures, but I was able to successfully build all portable kernels locally with buck2 after the fix.

Can we try CI again?

Gasoonjia:
@keyprocedure  so glad you've fixed the issue! Sry for late review. Restart ci.

keyprocedure:
> @keyprocedure so glad you've fixed the issue! Sry for late review. Restart ci.

No worries, I appreciate all the support :)

Progress with CI: 
The 10 failing tests were all due to a link error:
`Action failed: root//examples/portable/executor_runner:executor_runner (cxx_link_executable)`
I've registered `_clone_dim_order` in [op_registration_util.bzl](https://github.com/pytorch/executorch/pull/12974/files#diff-20ab8922e8d76d51917a0a03646ae7aa25e958bc2a2c9d0e26f438a9e483d814R1332) and I can successfully build executor_runner locally. 

Do you have any recommendations on which targets I should build locally to ensure everything that relies on new ops will build successfully or is there a Docker image available to run CI locally? I've tried to build entire dirs such as `//examples/portable/...` but run into dep failures unrelated to this PR.

shim_et/xplat/executorch/kernels/portable/op_registration_util.bzl
```diff
@@ -1329,6 +1329,13 @@ ATEN_OPS = (
             "//executorch/kernels/portable/cpu/util:copy_ops_util",
```
Gasoonjia, line 1329: we can remove broadcast_util here right? Since copy_util will depend on it

Gasoonjia:
thansk for your great work! @keyprocedure 
I don't think there's one single point target you can built for running CI locally. Thanks for bring it up! I think it is a good suggestion we can work on in the future for better local ci coverage!

shim_et/xplat/executorch/kernels/portable/op_registration_util.bzl
keyprocedure: Good catch. I'll remove it here and in `op__to_dim_order_copy.cpp`, then push once CI finishes.

Gasoonjia:
BTW Have you tried this https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#testing ? @keyprocedure

keyprocedure:
> BTW Have you tried this [`main`/CONTRIBUTING.md#testing](https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md?rgh-link-date=2025-08-08T23%3A07%3A19Z#testing) ? @keyprocedure

Thanks for sharing this!

After stopping some warnings from causing the build to fail, I ran the test script and everything passes except the [PyTree EmptySpec test](https://github.com/pytorch/executorch/blob/50ffe4466b1fb6d737c7f1359c4db08c4f734ad3/extension/pytree/test/test_pytree.cpp#L202), which seems unrelated. 

But I'll run this script to validate future changes.

Gasoonjia:
ci looks good. Stamped!

Gasoonjia (approved):

shim_et/xplat/executorch/kernels/portable/op_registration_util.bzl
keyprocedure: Should I still push the changes for removing the unused `broadcast_util` dep, or save it for a follow up?

PR #12987, opened 2025-07-29
add safety check to prim kernels

JacobSzwejbka:
Summary: Just safety checking that we have as many args as we expected

Differential Revision: D79124770


GregoryComer (approved):

zingo:
This PR could have broken the cortex-m size test :(  https://github.com/pytorch/executorch/actions/runs/16613951268/job/47002649711

zingo:
Tried to trigger a trunk build but it not triggering the builds after the PR is merged (and that makes sense :) )

PR #12992, opened 2025-07-30
Arm backend: Eliminate one memory copy in executor runner

perheld:
Prior to this patch, temporary tensors/buffers were used to hold the input data which was then copied over to the actual input tensors for running the inference. This patch removes this copying by instead writing the input data directly to the input tensors.

Change-Id: I20565a67dea8d8070efe9593aff4796c54e5fe90

### DIFF
--- a/examples/arm/executor_runner/arm_executor_runner.cpp
+++ b/examples/arm/executor_runner/arm_executor_runner.cpp
@@ -289,7 +289,7 @@ class Box {
   }
 };
 
-Result<BufferCleanup> prepare_input_tensors(
+Error prepare_input_tensors(
     Method& method,
     MemoryAllocator& allocator,
     const std::vector<std::pair<char*, size_t>>& input_buffers) {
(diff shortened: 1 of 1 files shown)


perheld:
https://github.com/pytorch/executorch/actions/runs/16622225895/job/47029216900 seems the size has gone above the threshold of 109000 resulting in "Fail 109236 > 109000". I dont know where the threshold value is coming from.

perheld:
> https://github.com/pytorch/executorch/actions/runs/16622225895/job/47029216900 seems the size has gone above the threshold of 109000 resulting in "Fail 109236 > 109000". I dont know where the threshold value is coming from.

Seems the threshold was increased a bit in https://github.com/pytorch/executorch/commit/69f3da0e332 by @swolchok  just a few days ago.

zingo:
mac os fails also unrelated

zingo (approved):

digantdesai:
> I dont know where the threshold value is coming from

Arbitrary, this is to keep large, unintended size increases to sneak past without getting noticed. I can bump it again.

PR #12994, opened 2025-07-30
Arm backend: Make TOSA backend NCHW-compatible

AdrianLundell:
- Moves the needed input/output transposes into the delegated graph to run on Ethos-U rather than requiring the EthosUBackend to implement transposes on CPU.

- Renames the annotate_channels_last_dim_order_pass to to_tosa_memory_format_pass since to be more descriptive and future proof.

This changes additionally enables running multiple batches since the EthosU transpose supports that natively, whereas the CPU implementation did not.


AdrianLundell:
@digantdesai This is a quite big change, do you need to run it internally before merging?

backends/arm/_passes/to_tosa_memory_format_pass.py
```diff
@@ -70,6 +83,7 @@ def is_weight_node_for_depthwise_conv2d(self, node: torch.fx.Node):
 
         return False
 
+    @staticmethod
```
digantdesai:
```suggestion
```

backends/arm/_passes/to_tosa_memory_format_pass.py
```diff
@@ -218,25 +239,48 @@ def insert_tosa_transposes(self, graph_module: torch.fx.GraphModule):
...
+                        ToTosaMemoryFormatPass.insert_input_transpose(
+                            node, input_node, graph_module
+                        )
+
     def call(self, graph_module: torch.fx.GraphModule):
         for node in graph_module.graph.nodes:
             node_data = get_first_fake_tensor(node).data
 
-            if node_data.dim() == 4:
+            # Inputs and outputs are always in (N)NCHW format
```
digantdesai, line 245: This may not be always the case. Users can do `tensor.to(memory_format=channels_last)` to the example input.

backends/arm/test/passes/test_to_tosa_memory_format.py
```diff
@@ -0,0 +1,187 @@
...
+
+    ops_after_pass = {"torch.ops.passthrough_to_tosa._transpose.default": 2}
+    ops_not_after_pass = []
+
+    def forward(self, x):
+        x = x + x
+        return x
+
+    def get_inputs(self):
+        return (torch.rand(1, 2, 2, 2),)
```
digantdesai, line 34: We should repeat these tests with NHWC input i.e.
```
return (torch.rand(1,2,2,2).to(memory_format=torch.channels_last),)
```

backends/arm/test/passes/test_to_tosa_memory_format.py
```diff
@@ -0,0 +1,187 @@
...
+        pass_list=[RemoveGetItemPass],
+        passes_with_exported_program=[ToTosaMemoryFormatPass],
+    )
+    pipeline.run()
+
+
+@common.parametrize("module", modules)
+def test_to_tosa_memory_format_tosa_MI_functional(module):
+    # Also run the actual pass pipeline to ensure functional correctness.
+    pipeline = TosaPipelineMI[input_t](module, module.get_inputs(), [])
```
digantdesai: No FVP tests?

digantdesai (commented):
Glad to see so many xfails removed. Left some comments.

backends/arm/_passes/to_tosa_memory_format_pass.py
AdrianLundell: This is the assumption for the arm backend right now actually: https://github.com/pytorch/executorch/blob/bedce91e7f4795869158b96ef479d92317b13871/backends/arm/process_node.py#L77, which has been the case since the transpose was done on CPU where we do not have info about the graph. 

After having moved it to NPU we can remove this condition, but I think this PR is large enough as it is.

backends/arm/test/passes/test_to_tosa_memory_format.py
AdrianLundell: This will fail, see previous comment

backends/arm/test/passes/test_to_tosa_memory_format.py
AdrianLundell: My intention here was to only test the logic of the pass with as little other things that could go wrong as possible ( quantization and fvps), we do also test the pass on FVP in other places since the change affects all modules we run tests on. Sounds ok?

zingo:
This will unfortunately need a rebase after the TOSA 0.80.1 removal :(

backends/arm/_passes/to_tosa_memory_format_pass.py
digantdesai: Follow up PR is OK.

digantdesai (approved):

Sebastian-Larsson:
> @digantdesai has imported this pull request. If you are a Meta employee, you can view this in [D79720805](https://www.internalfb.com/diff/D79720805?dst_version_fbid=2001649433973967).

@digantdesai Any update on this?

digantdesai:
Hey @Sebastian-Larsson Apologies for the delay. Internal tests got broken couple of days back so they are masking this signal once I merge #13221 hopefully I we can rebase this and pull internally again. 🙏

digantdesai:
> Exception: An error occurred when running the 'ToTosaMemoryFormatPass'

Getting this on the internal CI
Haven't looked deeper yet

AdrianLundell:
> > Exception: An error occurred when running the 'ToTosaMemoryFormatPass'
> 
> Getting this on the internal CI Haven't looked deeper yet

Have you tried the latest version? I have fixed one issue from the moving of the transpose to TOSA dialect change and that passed our internal CI at least.

AdrianLundell:
The sigmoid error can be seen on other PR:s as well so it is not related to this patch.

AdrianLundell:
I see that you are getting internal errors, anything I can help with?

digantdesai:
Actually its ok. Feel free to merge.

digantdesai:
Seems like it broke `backends/arm/test/ops/test_sigmoid_16bit.py::test_sigmoid_tosa_INT_add_sigmoid[rand]`?

AdrianLundell:
Seems like it yes, I thought it was flakyness since we had similar accuracy errors in the sigmoids before but apparently it always fails, sorry about that. 

It was previously an xfail though so it is just a matter of adding it back, I can do it first thing tomorrow morning if you do not do it.

PR #12995, opened 2025-07-30
Arm backend: Do not partition noop subgraphs

AdrianLundell:
- Reject partitions that will be lowered to empty subgraphs, i.e. containing only clones/ noop expands.
- For eye/one/zeros the graph is not really empty since it contains one constant, and this works now. Simply move the previosuly xfailing tests to MI/BI. (u55/ u85 still fails because of missing CPU ops)

### DIFF
--- a/backends/arm/_passes/convert_expand_copy_to_repeat.py
+++ b/backends/arm/_passes/convert_expand_copy_to_repeat.py
@@ -8,10 +8,41 @@
 import logging
 from typing import cast
 
+import torch
+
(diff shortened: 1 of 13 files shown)


zingo:
Need a rebase, sorry :(

AdrianLundell:
I'll wait with this PR until #12994 is merged

AdrianLundell:
Failures unrelated I believe

oscarandersson8218 (approved):

PR #12996, opened 2025-07-30
Arm Backend: Add support for ELU.default operator

agrima1304:
Decomposes elu into other operators/ lookup table for MI/ BI case.


Sebastian-Larsson (approved):
The CI failure for Cortex-m is unrelated to this patch. Approved

backends/arm/_passes/convert_elu_params.py
```diff
@@ -0,0 +1,46 @@
...
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class ConvertELUParamsPass(ExportPass):
+    """
+    Pass to convert the input_scale kwarg of ELU operator from float to
+    int.
+
+    It has been set to 2 as the outputs seem to stay the same regardless of what
+    the value of input_scale is, as long as that value is not 1.
```
digantdesai, line 18: I don't understand this.

backends/arm/test/ops/test_elu.py
```diff
@@ -0,0 +1,94 @@
...
+
+from executorch.backends.arm.test import common
+from executorch.backends.arm.test.tester.test_pipeline import (
+    EthosU55PipelineBI,
+    EthosU85PipelineBI,
+    TosaPipelineBI,
+    TosaPipelineMI,
+)
+
+test_data_suite = {
```
digantdesai, line 20: nice!

backends/arm/_passes/decompose_elu_pass.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    if op in aten_elu_ops:
+        return (
+            torch.ops.aten.add.Scalar,
+            torch.ops.aten.exp.default,
+            torch.ops.aten.ge.Scalar,
+            torch.ops.aten.where.self,
+            torch.ops.aten.mul.Scalar,
+        )
```
digantdesai: Curious why do we need this?

digantdesai (approved):
Thanks!

zingo:
This will unfortunately need a rebase after the TOSA 0.80.1 removal :(

backends/arm/_passes/convert_elu_params.py
agrima1304: The default value of input_scale is 1.0, however using the default value resulted in a type error. When passing 1 as an int, it was overridden by the default value (since both values are 1 and therefore equivalent ). So input_scale had to be changed to an int that is not 1.

backends/arm/_passes/decompose_elu_pass.py
agrima1304: We don't need that, I will remove it. Thank you for raising this.

backends/arm/test/ops/test_elu.py
agrima1304: Thank you!

PR #13004, opened 2025-07-30
Move to Span<EValue*> instead of EValue** in delegate interface

JacobSzwejbka:
Summary: For better safety we want to provide a pointer and a length instead of just a pointer. This is a breaking change of the delegate interface so have to update everyone at once. Span implements operator [], so I expect this to functionally be a no-op today.

Differential Revision: D79268134

### DIFF
--- a/backends/xnnpack/runtime/XNNExecutor.h
+++ b/backends/xnnpack/runtime/XNNExecutor.h
@@ -69,7 +69,7 @@ class XNNExecutor {
    * any additional memory planning as needed
    */
   ET_NODISCARD executorch::runtime::Error prepare_args(
-      executorch::runtime::EValue** args);
+      executorch::runtime::Span<executorch::runtime::EValue*> args);
 
   /**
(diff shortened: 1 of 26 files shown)


JacobSzwejbka:
All Ive done so far is just a find and replace code mod. Im sure I have to go through and fix 1 billion individual issues. Just wanted to get this out early to give backend authors a heads up on the change.

backends/xnnpack/runtime/XNNExecutor.h
```diff
@@ -69,7 +69,7 @@ class XNNExecutor {
    * any additional memory planning as needed
    */
   ET_NODISCARD executorch::runtime::Error prepare_args(
-      executorch::runtime::EValue** args);
+      executorch::runtime::Span<EValue*> args);
```
mcr229: think you need to add the executorch::runtime::Evalue namespace for evalue still.

cccclai (approved):
Makes sense

PR #13012, opened 2025-07-30
Use temp allocator for kernel registry

JakeStevens:
Summary: When indexing to the registry to get the op, memory is allocated from the method allocator to instantiate some TensorMeta and included objects. This memory is only used for that purpose and is not needed for the entire lifetime of the Method. Thus, we can instead use temp allocator which can later be reset and free up memory as needed.

Differential Revision: D79285675

cc @larryliu0820 @JacobSzwejbka @lucylq

### DIFF
--- a/runtime/executor/method.cpp
+++ b/runtime/executor/method.cpp
@@ -670,7 +670,7 @@ Error Method::resolve_operator(
     size_t kernel_index,
     InstructionArgs args,
     size_t n_args) {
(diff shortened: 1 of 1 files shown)


larryliu0820 (approved):
Thank you for doing this.

JakeStevens:
Failing tests pre-existing. moshi from undefined libtorchaudio symbol, openvino from backend import. Landing.

PR #13019, opened 2025-07-30
Default Export for Mv2 with Real Image Input

BujSet:
### Summary
When testing executorch inference runners on real MCUs, it's helpful if the default export of the model includes a real input (rather than just filled with random values or all 1.0s). This PR does this for the mobilenet_v2 model, following the steps from the tutorial [here](https://pytorch.org/hub/pytorch_vision_mobilenet_v2/).  

### Test plan
Ran `python -m examples.portable.scripts.export --model_name="mv2"` and saw the PTE now contains values associated with the `dog.jpg` image.

### DIFF
--- a/examples/models/mobilenet_v2/model.py
+++ b/examples/models/mobilenet_v2/model.py
@@ -15,8 +15,8 @@ from ..model_base import EagerModelBase
 
 
 class MV2Model(EagerModelBase):
-    def __init__(self, useRealInput=True):
-        self.useRealInput = useRealInput
+    def __init__(self, use_real_input=True):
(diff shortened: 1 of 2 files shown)


examples/models/mobilenet_v2/model.py
```diff
@@ -15,7 +15,8 @@
 
 
 class MV2Model(EagerModelBase):
-    def __init__(self):
+    def __init__(self, useRealInput=True):
+        self.useRealInput = useRealInput
```
larryliu0820: naming convention should be `use_real_input`

larryliu0820 (approved):

Issue #12961, opened 2025-07-29
make export flow etrecord generation supportive


Gasoonjia:
https://github.com/pytorch/executorch/discussions/12925

Gasoonjia:
make to_edge_transform_and_lower support  etrecord generation: https://github.com/pytorch/executorch/pull/13034

ETrecord update: https://github.com/pytorch/executorch/pull/13020, https://github.com/pytorch/executorch/pull/13010, https://github.com/pytorch/executorch/pull/12978

Gasoonjia:
et.export supportive: https://github.com/pytorch/executorch/pull/13164
to_edge supportive: [https://github.com/pytorch/executorch/pull/13151
other class updates: https://github.com/pytorch/executorch/pull/13143, https://github.com/pytorch/executorch/pull/13148, https://github.com/pytorch/executorch/pull/13058

Gasoonjia:
ok infras for https://github.com/pytorch/executorch/discussions/12925 should be finished -- closed this issue.

PR #13034, opened 2025-07-31
make to_edge_transform_and_lower support  etrecord generation

Gasoonjia:
* #13058
* #13052
* __->__ #13034
* #13020
* #13010

Differential Revision: [D79336982](https://our.internmc.facebook.com/intern/diff/D79336982/)

umbrella issue: https://github.com/pytorch/executorch/issues/12961


jackzhxng (approved):

GregoryComer:
It might be worth adding a test that covers to_edge_transform_and_lower with a partitioner. The logic might already handle this case, but to_backend creates a new edge program manager and is only called internally when a partitioner is provided. It might need some extra logic for this case to propagate the edge manager in to_backend.

PR #13035, opened 2025-07-31
Arm backend: Add asinh decomposition pass and test

emmakujala:
Add decomposition pass and tests for asinh. 


Sebastian-Larsson (approved):
CI failure for Cortex-m is unrelated. Approved

PR #13037, opened 2025-07-31
Arm backend: support list-of-tensors args in FuseConstantArgsPass

Sebastian-Larsson:
FuseConstantArgsPass is changed so that it no longer unpacks lists or tuples in node.args and node.kwargs. Those sequences are now preserved when the operator is invoked. That means ops like aten::cat, which expect their first argument to be a `List[Tensor]`, still get exactly that. Previously, the pass would flatten the list into separate tensor arguments, leading to this runtime error:

[WARNING 2025-07-24 11:37:13,749 fuse_constant_ops_pass.py:133] Failed to fuse constant op aten_cat_default due to exception: aten::cat() Expected a value of type 'List[Tensor]' for argument 'tensors' but instead found type 'Tensor'.

### DIFF
--- a/backends/arm/_passes/fuse_constant_ops_pass.py
+++ b/backends/arm/_passes/fuse_constant_ops_pass.py
@@ -6,6 +6,7 @@
 import logging
 
 import torch._export.utils
+import torch.fx
 from executorch.backends.arm._passes.arm_pass_utils import (
     get_constant_placeholder_kind,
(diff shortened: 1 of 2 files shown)


zingo (approved):

zingo (approved):

zingo:
cortex-m size test fails is unrelated and fails independent of this PR

PR #13041, opened 2025-07-31
NXP backend: Support for Sigmoid operator conversion

StrycekSimon:
### Summary

Adds implementation of converter and quantization pattern for `aten.sigmoid` operator.

### Test plan

Unit tests that covers Sigmoid operator were added.

cc @robert-kalmar


backends/nxp/backend/ir/converter/node_converters/ops_converters/relu_converter.py
```diff
@@ -25,6 +25,8 @@ def _is_supported_in_IR(
         return True
 
     def convert(self, node: Node):
+        self.assert_convertible(node)
```
robert-kalmar, line 28: Unrelated change, but OK.

robert-kalmar (approved):

PR #13042, opened 2025-07-31
Arm backend: Add arm-baremetal CMake preset

AdrianLundell:
- Executorch may be cross compiled for arm-baremetal targets using 'cmake --preset arm-baremetal'
- Adds a new flag EXECUTORCH_BUILD_ARM_ETDUMP used in the preset to easy enabling of ET_DUMP.
- backend/arm/scripts/build_executorch.sh is updated to use the new preset with same behaviour as before.

### DIFF
--- a/tools/cmake/preset/arm_baremetal.cmake
+++ b/tools/cmake/preset/arm_baremetal.cmake
@@ -1,0 +1,25 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}")
+set_overridable_option(EXECUTORCH_BUILD_EXECUTOR_RUNNER OFF)
+set_overridable_option(EXECUTORCH_BUILD_ARM_BAREMETAL ON)
+set_overridable_option(EXECUTORCH_BUILD_KERNELS_QUANTIZED ON)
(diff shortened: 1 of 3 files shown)


tools/cmake/preset/arm_baremetal.cmake
```diff
@@ -0,0 +1,25 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
```
digantdesai, line 1: Though this PR is OK, I want to brainstorm here a bit more.

There are two, I think, orthogonal things we are mixing here, (1) Arm-ness, (2) Baremetal-ness. Can we split? For the preset its OK, I am talking about things like `EXECUTORCH_BUILD_ARM_BAREMETAL` the things it guards are not strictly related to Arm.

This will come up when we try to do Arm Cortex-m + Ethos on Zephyr. Passing ARM_BAREMETAL would be confusing at best.

digantdesai (approved):

tools/cmake/preset/arm_baremetal.cmake
AdrianLundell: Good point, changing it to something like `EXECUTORCH_BUILD_ARM_ETHOSU` seems reasonable to me.

digantdesai: I guess depends on the things we want to guard with this build flavor. We need to split this `EXEUCTORCH_BUILD_ARM_BAREMETAL` into `EXECUTORCH_BUILD_ARM_ETHOSU` and `EXECUTORCH_BUILD_BAREMETAL`. I can also see Cortex-M (no NPU) builds and not sure if the `ETHOSU` is the right suffix either.

PR #13047, opened 2025-07-31
[Android] Move training part to its own package

kirklandsign:
### Summary
Under its own namespace
### Test plan
CI


extension/android/executorch_android/src/main/java/org/pytorch/executorch/training/SGD.java
```diff
@@ -1,18 +1,22 @@
 /*
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- * All rights reserved.
  *
- * This source code is licensed under the BSD-style license found in the
- * LICENSE file in the root directory of this source tree.
+ *  * Copyright (c) Meta Platforms, Inc. and affiliates.
```
georgehong: What's the reasoning behind the extra layer of `*` for the license? This is also done for the `TrainingModule`, so just wanted to understand the distinction.

georgehong (approved):

PR #13052, opened 2025-07-31
make etrecord set representive IO

Gasoonjia:
* #13058
* __->__ #13052

representive input and reference output in etrecord will not be set during export flow. To continue supporting the two functionalities, this diff creates two class methods to customize IO in etrecord.

Differential Revision: [D79386896](https://our.internmc.facebook.com/intern/diff/D79386896/)


devtools/etrecord/_etrecord.py
```diff
@@ -345,6 +345,56 @@ def add_edge_dialect_program(
...
+                from which representative inputs will be extracted.
+        """
+        if isinstance(representative_inputs, BundledProgram):
+            self._representative_inputs = _get_representative_inputs(
+                representative_inputs
+            )
+        else:
+            self._representative_inputs = representative_inputs
+
+    def update_reference_outputs(
```
digantdesai, line 370: Nit: why not a single API? Is there a need to use these individually and not together?

JacobSzwejbka (approved):

Issue #11719, opened 2025-06-16
CoreML model with no inputs fails to load at runtime


GregoryComer:
### 🐛 Describe the bug

The following model fails to load at runtime. This is presumably due to not taking inputs? When changing the model to x + torch.randn(...), it loads successfully. This isn't a particularly high-priority case, but does appear to be a bug.

Repro:
```py
import torch

from executorch.backends.apple.coreml.partition import CoreMLPartitioner
from executorch.exir import to_edge_transform_and_lower, EdgeCompileConfig, to_edge
from executorch.extension.pybindings.portable_lib import _load_for_executorch_from_buffer

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        
    def forward(self):
        return torch.randn(5, 5)

model = Model()
inputs = (
)

print(inputs)

eager_outputs = model(*inputs)
print(f"Eager: {eager_outputs.shape} {eager_outputs}")

ep = torch.export.export(model.eval(), inputs)
print(ep)

print(f"EP: {ep.module()(*inputs)}")

lowered = to_edge_transform_and_lower(
    ep,
    partitioner=[CoreMLPartitioner()],
    compile_config=EdgeCompileConfig(_check_ir_validity=False)
).to_executorch()

print(lowered.exported_program())

et_model = _load_for_executorch_from_buffer(lowered.buffer)
et_outputs = et_model([*inputs])[0]

print(et_outputs)

et_outputs - eager_outputs
```

Outputs:
```
[ETCoreMLModelManager.mm:566] [Core ML]  Metadata is invalid or missing.
[backend_delegate.mm:288] [Core ML] Model init failed Metadata is invalid or missing.
[coreml_backend_delegate.mm:193] CoreMLBackend: Failed to init the model.
[method.cpp:113] Init failed for backend CoreMLBackend: 0x23
```

### Versions

coremltools version 8.3
executorch commit https://github.com/pytorch/executorch/commit/67b6009d6b3b67eee775c8ed2fe30eae6e0bb65c (Jun 14)

cc @kimishpatel @YifanShenSZ @cymbalrush @metascroy

metascroy:
Fix for ET here: https://github.com/pytorch/executorch/pull/13053

Issue re-filed with coremltools here: https://github.com/apple/coremltools/issues/2578

PR #13053, opened 2025-08-01
Change default minimum_deployment_target to None and support CoreML models with no inputs

metascroy:
This PR changes the default minimum_deployment_target in CoreML from the fixed iOS15 to None.  This will automatically select the minimum deployment target required to export the model.

A warning occurs after export telling the user the deploymnet target selected, and directing them to specify a specific target if that is what they need.

This is more in line with how CoreML standalone works.

In addition, this PR allows running CoreML models with no user inputs, which requires CoreML deployment target iOS18 or higher.

This addresses:
* https://github.com/pytorch/executorch/issues/11719
* https://github.com/pytorch/executorch/issues/12906

### DIFF
--- a/backends/apple/coreml/compiler/coreml_preprocess.py
+++ b/backends/apple/coreml/compiler/coreml_preprocess.py
@@ -126,15 +126,18 @@ class CoreMLBackend(BackendDetails):
 
     @staticmethod
     def generate_minimum_deployment_target_compile_spec(
-        min_deployment_target: ct.target,
+        min_deployment_target: Optional[ct.target],
     ) -> CompileSpec:
         """
         Returns the compile spec representing the minimum deployment target on which the model can run,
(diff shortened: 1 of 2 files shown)


metascroy:
@cymbalrush can you have a look?

metascroy:
@cccclai @GregoryComer let me know if you have any concerns with merging.

@cymbalrush said the changes look OK on slack

cccclai (approved):
I thought CoreML doesn't support version <iOS15, is it no longer true? Stamp anyway because Gyan seems good with it

metascroy:
> I thought CoreML doesn't support version <iOS15, is it no longer true? Stamp anyway because Gyan seems good with it

It doesn't support version < iOS15 if you use an mlpackage (which ET uses).  This PR is about choosing higher iOS versions than iOS15 if those are required for the model.  Currently it's fixed at iOS15 as the default, which could become dated.

PR #13060, opened 2025-08-01
Arm backend: Import executorch deps using find_package

AdrianLundell:
- Renames portable_ops_lib to arm_portable_ops_lib to not conflict with the op_lib from the executorch_package regestring all ops.
- Adds install target to the ethos_u_delegate to make it findable using find_package.
- Removes the --whole-archive arg for quantized_ops_lib to avoid re-registiring runtime error.


digantdesai (approved):

Sebastian-Larsson (approved):
Unrelated CI failures. Approved

PR #13062, opened 2025-08-01
Qualcomm AI Engine Direct - Fix the regression of whisper model

shewu-quic:
Summary:
- Resolve the Whisper model  accuracy issue caused by upgrading the Transformers.
  - Modify decompose_sdpa.py to support kwargs "scale"
- fixed internal CI

cc: @haowhsu-quic , @winskuo-quic

### DIFF
--- a/backends/qualcomm/tests/models.py
+++ b/backends/qualcomm/tests/models.py
@@ -1530,12 +1530,13 @@ class Rsqrt(torch.nn.Module):
 
 
 class ScaledDotProductAttention(torch.nn.Module):
-    def __init__(self):
+    def __init__(self, scale=None):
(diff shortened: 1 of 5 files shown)


shewu-quic:
Hi @cccclai,
This PR is to fixed the accuracy issue of the whisper model when upgrade transformers. The primary issue is that the original sdpa does not account for the "scale" keyword argument, which is now used by the latest Whisper model.
https://github.com/huggingface/transformers/blob/d3b8627b56caa7ca8fac113c9f28d0256db0194d/src/transformers/models/whisper/modeling_whisper.py#L358

cccclai (approved):
Thanks!

PR #13065, opened 2025-08-01
Arm backend: Add tests for TOSA and VGF for extract_io_params_tosa + fix for bug with is_U55_subset

wwwind:
Fix error with "is_U55_subset" by changing TOSAQuantizer to accept compiled spec

Change-Id: I164d68190e6fb9edd2efb9c8bd2da8c1bfc4e406


robell (approved):

zingo (approved):

zingo:
fails are unrerlated

backends/arm/quantizer/arm_quantizer.py
```diff
@@ -218,9 +218,35 @@ def not_module_type_or_name_filter(n: Node) -> bool:
 
 class TOSAQuantizer(Quantizer):
 
-    def __init__(self, tosa_spec: TosaSpecification) -> None:
+    def __init__(
+        self, compile_spec_or_tosa_spec: Union[TosaSpecification, List[CompileSpec]]
```
digantdesai, line 222: Nit: do you want to split into two constructors, instead of overloading..
```
def __init__(self, tosa_spec):
[...]
```
and
```
@classmethod
def from_compilespec(self, compilespec):
[...]
```

PR #13073, opened 2025-08-01
Fix installation that causes torch conflict

mergennachin:
When pip installs timm==1.0.7 (from requirements-examples.txt), it's pulling its dependencies from PyPI, and since the PyPI version of timm has a dependency on torch without any version constraints, pip is "upgrading" your nightly torch to the latest stable release from PyPI.

Here's what's happening:

I have torch==2.9.0.dev20250725 (nightly) installed
When pip installs timm, it sees that timm requires torch
Pip finds torch==2.7.1 on PyPI (the latest stable release)
Since 2.9.0.dev20250725 is considered a pre-release version, pip thinks 2.7.1 is "newer" in terms of stable releases
Pip uninstalls the nightly and installs the stable version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ dist/
 ethos-u-scratch/
 executorch.egg-info
 pip-out/
+constraints.txt
 
 # Any exported models and profiling outputs
 *.bin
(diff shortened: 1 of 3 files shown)


mergennachin:
I think a simpler fix is this https://github.com/pytorch/executorch/pull/13074

PR #13079, opened 2025-08-02
Remove unused sleef.h which breaks cross-compilation on windows

ykhrustalev:
### Summary

The sleef.h header contains 

```
#if (defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__) || defined(_MSC_VER)) && !defined(SLEEF_STATIC_LIBS)
#ifdef SLEEF_IMPORT_IS_EXPORT
#define SLEEF_IMPORT __declspec(dllexport)
#else // #ifdef SLEEF_IMPORT_IS_EXPORT
#define SLEEF_IMPORT __declspec(dllimport)
#if (defined(_MSC_VER))
#pragma comment(lib,"sleef.lib")
#endif // #if (defined(_MSC_VER))
#endif // #ifdef SLEEF_IMPORT_IS_EXPORT
#else // #if (defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__) || defined(_MSC_VER)) && !defined(SLEEF_STATIC_LIBS)
#define SLEEF_IMPORT
#endif // #if (defined(__MINGW32__) || defined(__MINGW64__) || defined(__CYGWIN__) || defined(_MSC_VER)) && !defined(SLEEF_STATIC_LIBS)
```
> `#pragma comment(lib,"sleef.lib")`

which forces the `sleef.lib` to be linked for the static windows lib. 
The 2 particular optimized kernels are not using any sleef functions and could be used as a subset that is statically compiled.

### Test plan

Removing the header makes the static link to pass without a need to have a sleef lib.


digantdesai:
> The 2 particular optimized kernels are not using any sleef functions and could be used as a subset that is statically compiled.

Good. yeah we use `at::native::vectorized_gelu` now for gelu for instance.
If the CI is clean perhaps we can merge. Do you mind fixing the Buck target file as well?

digantdesai:
Running CI now

ykhrustalev:
> > The 2 particular optimized kernels are not using any sleef functions and could be used as a subset that is statically compiled.
> 
> Good. yeah we use `at::native::vectorized_gelu` now for gelu for instance. If the CI is clean perhaps we can merge. Do you mind fixing the Buck target file as well?

@digantdesai the linking comes from the "header" file, not from the buck or cmake. Can you please elaborate the Buck change you mean?

digantdesai:
NVM we may not need any update to the BUCK file.

digantdesai (approved):

kimishpatel:
wait but gelu.h from aten native is all header only impl which pulls in other headers like Vectorized.h. So inclusion of sleef should likely happen there, which does include sleef under these conditions, https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cpu/vec/vec128/vec128_float_neon.h#L10-L12. So if you are avoiding use of sleef tahts fine (as long as it doesnt break other builds), but it will cost perf on windows

ykhrustalev:
> wait but gelu.h from aten native is all header only impl which pulls in other headers like Vectorized.h. So inclusion of sleef should likely happen there, which does include sleef under these conditions, https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cpu/vec/vec128/vec128_float_neon.h#L10-L12. So if you are avoiding use of sleef tahts fine (as long as it doesnt break other builds), but it will cost perf on windows

That's fine, we are using only a subset of kernels for that case, and the 2 kernels in question are not using Sleef in their implementation. Other sleef kernels are skipped via compilation flags.

swolchok (approved):

digantdesai:
Once the CI is green (after rebase) I can merge this. @kimishpatel - OK with this?

kimishpatel:
yeah ok with me

PR #13094, opened 2025-08-04
Arm backend: Move quant util functions to closer to usage

Sebastian-Larsson:
The following functions are only used in quantization_annotator and can therefore be moved from arm_quantizer_utils.py to
quantization_annotator.py:

    * is_large_scalar
    * is_non_float_tensor
    * get_node_target

Additionally, is_ok_for_quantization is removed. It combined the is_large_scalar and is_non_float_tensor checks into one, which is now done directly where is_ok_for_quantization was used.

### DIFF
--- a/backends/arm/quantizer/arm_quantizer_utils.py
+++ b/backends/arm/quantizer/arm_quantizer_utils.py
@@ -11,11 +11,9 @@
 # Utility functions for TOSAQuantizer
 #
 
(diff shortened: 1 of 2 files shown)


Sebastian-Larsson:
CI failures are unrelated

oscarandersson8218 (approved):

PR #13106, opened 2025-08-04
Cleaned up Wasm cmake

Conarnar:
### Summary
Before, if dependencies for the API were not enabled in the cmake command, you would get linking errors when trying to compile. Now, if EXECUTORCH_BUILD_WASM is turned on, an error will occur during configure stage if dependencies were not enabled.

Also changed the API to use executorch_kernels rather than hardcoding in portable_ops_lib.

For the unit tests, changed the libraries to PRIVATE instead of PUBLIC and to use `--pre-js` instead of `--post-js` as that is the recommended way of interfacing with the Embind Module.

### Test plan
```bash
bash scripts/build_wasm_tests.sh
cd cmake-out-wasm/extension/wasm/test/
npm test # after installing Jest
```


CMakeLists.txt
```diff
@@ -796,6 +796,22 @@ if(EXECUTORCH_BUILD_PYBIND)
 endif()
 
 if(EXECUTORCH_BUILD_WASM)
+  if(NOT EXECUTORCH_BUILD_EXTENSION_DATA_LOADER)
```
JacobSzwejbka: @larryliu0820  I feel like this pattern used to be really common in our cmake but now taking a look I dont see it anywhere else except pybind. How does something like Module guarantee that data loader is built now?

Conarnar:
```cmake
check_required_options_on(
  IF_ON EXECUTORCH_BUILD_EXTENSION_MODULE REQUIRES
  EXECUTORCH_BUILD_EXTENSION_DATA_LOADER EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR
)
```
Looks like if you have `-DEXECUTORCH_BUILD_EXTENSION_MODULE=ON` without its dependencies, you just get an error.

JacobSzwejbka: Did this used to be the case @GregoryComer should options be self sufficient or are we doing something else instead?

JacobSzwejbka: I guess for now @Conarnar just follow the existing pattern and shift to the requires options thing

extension/wasm/CMakeLists.txt
```diff
@@ -49,8 +48,12 @@ target_compile_options(executorch_wasm PUBLIC ${_common_compile_options})
 target_include_directories(
   executorch_wasm PUBLIC ${_common_include_directories}
 )
-target_link_libraries(executorch_wasm PUBLIC ${link_libraries})
+target_link_libraries(
+  executorch_wasm
+  PUBLIC ${link_libraries}
+  INTERFACE executorch_kernels
```
lucylq, line 54: So `executorch_kernels` is only built when one of the selective build options is enabled - maybe there should be some documentation indicating that wasm build must use the toplevel selective build API?

Conarnar: Actually, if none of the selective build options are used, executorch_kernels will default to optimized_native_cpu_ops_lib if EXECUTORCH_BUILD_KERNELS_OPTIMIZED is turned on, portable_ops_lib otherwise.
https://github.com/pytorch/executorch/blob/main/CMakeLists.txt#L885-L892

lucylq (approved):

PR #13114, opened 2025-08-04
Run Qwen script with Buck

rohansjoshi:
Summary:
Added targets to enable running script llama.py for lowering Qwen with Buck
Removed use of relative path to load Qwen config

Currently disabling SpinQuant R2 since it doesn't work with Qwen (but R1 does) -- will set use_r2 based on model later

Reviewed By: cccclai

Differential Revision: D79386256


cccclai (approved):
Thanks!

PR #13120, opened 2025-08-05
Update CoreML docs

metascroy:
This updates CoreML docs to:

* Discuss the new partitioner options
* Discuss quantize_ support
* Discuss backward compatibility guarantees

PR #13121, opened 2025-08-05
[Executorch][Recipes][Coreml] Add coreml backend recipes

### LINKED ISSUE #13100: Add CoreML backend recipes

issue to track adding CoreML recipes similar to XNNPack recipes for users to ease exporting a model for coreml backend.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

abhinaykukkadapu:
### Summary
Adds coreml recipes as discussed similar to xnnpack backend recipes to use them for export flow.

Fixes #13100

### Test plan
`python -m unittest backends/apple/coreml/test/test_coreml_recipes.py`

### DIFF
--- a/backends/apple/coreml/recipes/coreml_recipe_provider.py
+++ b/backends/apple/coreml/recipes/coreml_recipe_provider.py
@@ -1,0 +1,132 @@
+# Copyright © 2025 Apple Inc. All rights reserved.
+#
+# Please refer to the license found in the LICENSE file in the root directory of the source tree.
+
+
+from typing import Any, Optional, Sequence
+
+import coremltools as ct
+
+from executorch.backends.apple.coreml.compiler import CoreMLBackend
+from executorch.backends.apple.coreml.partition.coreml_partitioner import (
+    CoreMLPartitioner,
+)
+from executorch.backends.apple.coreml.recipes.coreml_recipe_types import (
+    COREML_BACKEND,
+    CoreMLRecipeType,
+)
+
+from executorch.exir import EdgeCompileConfig
+from executorch.export import (
+    BackendRecipeProvider,
+    ExportRecipe,
+    LoweringRecipe,
+    RecipeType,
+)
+
+
+class CoreMLRecipeProvider(BackendRecipeProvider):
+    @property
+    def backend_name(self) -> str:
+        return COREML_BACKEND
+
+    def get_supported_recipes(self) -> Sequence[RecipeType]:
+        return list(CoreMLRecipeType)
+
+    def create_recipe(
+        self, recipe_type: RecipeType, **kwargs: Any
+    ) -> Optional[ExportRecipe]:
+        """Create CoreML recipe with precision and compute unit combinations"""
+
+        if recipe_type not in self.get_supported_recipes():
+            return None
+
+        if ct is None:
+            raise ImportError(
+                "coremltools is required for CoreML recipes. "
+                "Install it with: pip install coremltools"
+            )
+
+        # Validate kwargs
+        self._validate_recipe_kwargs(recipe_type, **kwargs)
+
+        # Parse recipe type to get precision and compute unit
+        precision = None
+        if recipe_type == CoreMLRecipeType.FP32:
+            precision = ct.precision.FLOAT32
+        elif recipe_type == CoreMLRecipeType.FP16:
+            precision = ct.precision.FLOAT16
+
+        if precision is None:
+            raise ValueError(f"Unknown precision for recipe: {recipe_type.value}")
+
+        return self._build_recipe(recipe_type, precision, **kwargs)
+
+    def _validate_recipe_kwargs(self, recipe_type: RecipeType, **kwargs: Any) -> None:
+        if not kwargs:
+            return
+        expected_keys = {"minimum_deployment_target", "compute_unit"}
+        unexpected = set(kwargs.keys()) - expected_keys
+        if unexpected:
+            raise ValueError(
+                f"CoreML Recipes only accept 'minimum_deployment_target' or 'compute_unit' as parameter. "
+                f"Unexpected parameters: {list(unexpected)}"
+            )
+        if "minimum_deployment_target" in kwargs:
+            minimum_deployment_target = kwargs["minimum_deployment_target"]
+            if not isinstance(minimum_deployment_target, ct.target):
+                raise ValueError(
+                    f"Parameter 'minimum_deployment_target' must be an enum of type ct.target, got {type(minimum_deployment_target)}"
+                )
+        if "compute_unit" in kwargs:
+            compute_unit = kwargs["compute_unit"]
+            if not isinstance(compute_unit, ct.ComputeUnit):
+                raise ValueError(
+                    f"Parameter 'compute_unit' must be an enum of type ct.ComputeUnit, got {type(compute_unit)}"
+                )
+
+    def _build_recipe(
+        self,
+        recipe_type: RecipeType,
+        precision: ct.precision,
+        **kwargs: Any,
+    ) -> ExportRecipe:
+        lowering_recipe = self._get_coreml_lowering_recipe(
+            compute_precision=precision,
+            **kwargs,
+        )
+
+        return ExportRecipe(
+            name=recipe_type.value,
+            quantization_recipe=None,  # TODO - add quantization recipe
+            lowering_recipe=lowering_recipe,
+        )
+
+    def _get_coreml_lowering_recipe(
+        self,
+        compute_precision: ct.precision,
+        **kwargs: Any,
+    ) -> LoweringRecipe:
+        compile_specs = CoreMLBackend.generate_compile_specs(
+            compute_precision=compute_precision,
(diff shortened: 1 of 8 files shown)


backends/apple/coreml/recipes/coreml_recipe_types.py
```diff
@@ -0,0 +1,32 @@
...
+
+# pyre-strict
+
+from executorch.export import RecipeType
+
+
+COREML_BACKEND: str = "coreml"
+
+
+class CoreMLRecipeType(RecipeType):
```
abhinaykukkadapu, line 12: @metascroy any suggestions on what all recipes must be added for CoreML, any quantization specific ones?

metascroy: @abhinaykukkadapu I don't think precision/compute unit lead to different recipe types any more than the deployment target does (which is given as a separate arg).

The reason they're in the name in optimum-et is because they had to be strings to work with the CLI:

```
optimum-cli export model_id --recipe coreml_gpu_f32
```

But the ET recipes are a bit more general, so maybe they could be args here as well?

backends/apple/coreml/recipes/coreml_recipe_types.py
```diff
@@ -0,0 +1,32 @@
...
+    """CoreML-specific recipe types - combinations of precision and compute units"""
+
+    # FP32 precision with different compute units
+    FP32_CPU = "coreml_fp32_cpu"
+    FP32_GPU = "coreml_fp32_gpu"
+    FP32_NEURAL_ENGINE = "coreml_fp32_neural_engine"
```
metascroy: Nit: the neural engine doesn't support FP32.  Maybe call it NE instead of neaural_engine to align with coremltools naming convention?

backends/apple/coreml/test/test_coreml_recipes.py
```diff
@@ -0,0 +1,207 @@
...
+
+import unittest
+from typing import List
+
+import torch
+from executorch.backends.apple.coreml.recipes import (
+    CoreMLRecipeProvider,
+    CoreMLRecipeType,
+)
+from executorch.backends.apple.coreml.test.test_torch_ops import _TEST_RUNTIME
```
metascroy: Nit: rather than import from the other test, factor _TEST_RUNTIME out into a common test_util and use in both tests.

backends/apple/coreml/test/test_coreml_recipes.py
```diff
@@ -0,0 +1,207 @@
...
+from executorch.exir.schema import DelegateCall, Program
+from executorch.export import export, ExportRecipe, recipe_registry
+from torch import nn
+from torch.testing._internal.common_quantization import TestHelperModules
+
+try:
```
metascroy: When is coremltools not available?

It's a required dependency in ET now

backends/apple/coreml/recipes/coreml_recipe_provider.py
```diff
@@ -0,0 +1,156 @@
...
+    ) -> ExportRecipe:
+        lowering_recipe = self._get_coreml_lowering_recipe(
+            compute_precision=precision,
+            compute_unit=compute_unit,
+            minimum_deployment_target=minimum_deployment_target,
+        )
+
+        return ExportRecipe(
+            name=recipe_type.value,
+            quantization_recipe=None,  # TODO - add quantization recipe
```
metascroy, line 101: FYI: I have a PR updating CoreML docs that you might check out for quantization and available partitioner options: https://github.com/pytorch/executorch/pull/13120

backends/apple/coreml/recipes/coreml_recipe_provider.py
```diff
@@ -0,0 +1,156 @@
...
+                raise ValueError(
+                    f"Parameter 'minimum_deployment_target' must be an enum of type ct.target, got {type(minimum_deployment_target)}"
+                )
+
+    def _build_recipe(
+        self,
+        recipe_type: RecipeType,
+        precision: ct.precision,
+        compute_unit: ct.ComputeUnit,
+        minimum_deployment_target: ct.target = ct.target.iOS15,
```
metascroy: The new default minimum_deployment_target in the paritioner is None, see updated docs: https://github.com/pytorch/executorch/pull/13120

This default matches coremltools.  When deployment target is None, coremltools selects the minimum target needed to export the model.

backends/apple/coreml/recipes/coreml_recipe_provider.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+
+        partitioner = CoreMLPartitioner(
+            compile_specs=compile_specs,
+            take_over_mutable_buffer=(minimum_deployment_target >= ct.target.iOS18),
+        )
+
+        edge_compile_config = EdgeCompileConfig(
+            _check_ir_validity=False,
+            _skip_dim_order=True,
```
metascroy: CoreML now supports dim_order

backends/apple/coreml/test/test_coreml_recipes.py
abhinaykukkadapu: If it's required then we may not need it, will remove the check.

backends/apple/coreml/recipes/coreml_recipe_types.py
abhinaykukkadapu: Sure, will remove FP32_NEURAL_ENGINE and will rename FP16_NE

backends/apple/coreml/recipes/coreml_recipe_types.py
abhinaykukkadapu: @metascroy 

> I don't think precision/compute unit lead to different recipe types any more than the deployment target does 

Can you explain more, seems all of these are orthogonal?

Also, are you suggesting precision, cu, deployment target, all of these should be args, for ease of use i think these specializations would help the end user?

backends/apple/coreml/recipes/coreml_recipe_provider.py
kimishpatel: @metascroy it might make sense for you to follow up on the recipes for quant

backends/apple/coreml/recipes/coreml_recipe_types.py
kimishpatel: For canonical recipes we probably just want to go with FP16. This way you dont have combinatorial explosion either. If we want to construct a reciped then provide options to chose cu,  precision etc.

abhinaykukkadapu: @kimishpatel looking at the coreml precision and compute unit enums, these are already exhaustive combinations and are minimal so i'm thinking it makes sense to keep all valid combinations of precision and compute units and leave deployment target as default but overridable via kwargs.

Precision: https://github.com/apple/coremltools/blob/main/coremltools/converters/mil/mil/passes/defs/quantization.py#L28
CU: https://github.com/apple/coremltools/blob/main/coremltools/__init__.py#L74

backends/apple/coreml/recipes/coreml_recipe_types.py
metascroy: It's related to @kimishpatel's comment below on combinatorial explosion.

For these low-level details, maybe use the recipe provider and pass in args like you do for minimum_deployment in the unit test:

```
recipe = CoreMLRecipeProvider.create_recipe(compute_unit=ct.CPU_AND_NE, compute_precision=ct.precision.FLOAT16, minimum_deployment_target=ct.target.iOS16
)
```

The high-level recipes can be simpler and choose reasonable defaults.  Maybe there can be one FP16 and one FP32 high-level recipe, but both set compute_unit = ALL, minimum_deployment_target=None.  If a user wants to configure those, they can use the provider to create a recipe.

backends/apple/coreml/recipes/coreml_recipe_provider.py
metascroy: Sure, @abhinaykukkadapu are the quant recipes coming in a future PR?

backends/apple/coreml/recipes/coreml_recipe_types.py
abhinaykukkadapu: Sounds good, will update the in the next rev.

backends/apple/coreml/recipes/coreml_recipe_provider.py
abhinaykukkadapu: @metascroy If you want to follow up, i can leave it upto you as you have more context on quantization, or i can take it up if you don't have bandwidth.

metascroy: You can add the quantization recipe, I'm happy to review the quantize_ and PT2E recipes you add for CoreML

abhinaykukkadapu: @metascroy sure will add in a follow up diff, can we go ahead with these changes?

metascroy (approved):

PR #13123, opened 2025-08-05
Add AOTI backend skeleton code

larryliu0820:
### Summary

### Test plan


exir/lowered_backend_module.py
```diff
@@ -734,6 +734,7 @@ def create_exported_program_from_submodule(
                     ),
                 )
             ],
+            example_inputs=owning_program.example_inputs,
```
larryliu0820: This is a hack. We can't make sure the partitioned graph module has the same input as the original graph module.

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+AOTInductorModelContainerCreateWithDeviceFunc
+    AOTInductorModelContainerCreateWithDevice = nullptr;
+AOTInductorModelContainerDeleteFunc AOTInductorModelContainerDelete = nullptr;
+AOTInductorModelContainerGetNumInputsFunc
+    AOTInductorModelContainerGetNumInputs = nullptr;
+AOTInductorModelContainerGetNumOutputsFunc
+    AOTInductorModelContainerGetNumOutputs = nullptr;
+AOTInductorModelContainerRunFunc AOTInductorModelContainerRun = nullptr;
+std::unordered_map<Tensor*, std::vector<int64_t>> tensor_to_sizes;
```
JacobSzwejbka: Im assuming this is a temporary hack. I dont think we can reasonably leave these in global state like this. You probably have to do AOTITensorHandle = extension::Tensor*

larryliu0820: > You probably have to do AOTITensorHandle = extension::Tensor*

The only thing stopping us from doing so is the different size types on ATen tensor and ET tensor (int64 vs int32)

CMakeLists.txt
```diff
@@ -607,6 +609,12 @@ endif()
 
 if(EXECUTORCH_BUILD_CORTEX_M)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/backends/cortex_m)
+  list(APPEND _executorch_backends coretex_m_backend)
```
swolchok, line 591: this target doesn't seem to exist (even if I assume it's a typo for "cortex_m_backend").

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+# Source root directory for executorch.
+if(NOT EXECUTORCH_ROOT)
+  set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
+endif()
+
+# include(${EXECUTORCH_ROOT}/build/Utils.cmake)
+
+find_package(CUDAToolkit REQUIRED)
```
swolchok: presumably we need separate configuration knobs for CUDA and Metal (not blocking)

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+set(_aoti_sources runtime/AotiBackend.cpp)
+add_library(aoti_backend STATIC ${_aoti_sources})
```
swolchok: nit: just inline _aoti_sources, no need to make a variable

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+target_include_directories(
+  aoti_backend
+  PUBLIC
+    ${CUDAToolkit_INCLUDE_DIRS}
+    $<BUILD_INTERFACE:${EXECUTORCH_ROOT}>
+    $<INSTALL_INTERFACE:include>
+)
+target_compile_options(aoti_backend PUBLIC -fexceptions -frtti -fPIC)
```
swolchok: -fexceptions and -frtti are default-on, no need to specify. -fPIC should be gated by CMAKE_POSITION_INDEPENDENT_CODE; why mess with it?

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+# Ensure symbols are exported properly
+target_link_options(aoti_backend PUBLIC -Wl,--export-dynamic)
```
swolchok: 1) this is a linux-ism
2) this is for building shared libraries, but you've configured aoti_backend as a static library. Are you sure we need this?
3) if aoti_backend was a shared library, I think this would be passed by default

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  aoti_backend
+  PUBLIC
+    extension_tensor
+    CUDA::cudart
+    ${CMAKE_DL_LIBS}
+)
+# If you need other CUDA libraries, link them similarly:
+# target_link_libraries(aoti_backend PUBLIC CUDA::cublas CUDA::cufft ...)
+# If you have a custom function, keep it
+executorch_target_link_options_shared_lib(aoti_backend)
```
swolchok: this use is fine, but we really need to rename this function

backends/aoti/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+install(
+  TARGETS aoti_backend
+  EXPORT ExecuTorchTargets
+  DESTINATION lib
```
swolchok, line 53: s/lib/${CMAKE_INSTALL_LIBDIR}/

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace executorch {
+namespace backends {
+namespace aoti {
```
swolchok: nit: we have C++17, just `namespace executorch::backends::aoti {`

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+// Here is where the aoti bouncers are going to be defined.
+// I define the globals aoti generated compiled code calls
+// They can be backed by ET systems
+
+using namespace std;
+
+using executorch::aten::ScalarType;
```
swolchok: we shouldn't be hitting anything labeled aten because we're libtorch-free, right?

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+using AOTIRuntimeError = Error;
+using AOTITorchError = Error;
+
+struct AOTInductorModelContainerOpaque;
+using AOTInductorModelContainerHandle = AOTInductorModelContainerOpaque*;
+using AOTInductorStreamHandle = void*;
+using AOTIProxyExecutorHandle = void*;
+
+using AOTInductorModelContainerCreateWithDeviceFunc = AOTIRuntimeError (*)(
```
swolchok: A nicer way to get these types is to `#include "torch/csrc/inductor/aoti_runtime/interface.h"` and then use `decltype(&AOTInductorModelContainerCreateWithDevice)` etc. Does require you to *not* create a function pointer named `AOTInductorModelContainerCreateWithDevice`, though.

backends/aoti/runtime/AotiBackend.cpp
swolchok: 1) ideally this should be something analogous to PyTorch's DimVector, or even SizesAndStrides -- some sort of small_vector analogue to cut down on heap allocations
2) we should probably just assume that if we need sizes we will also need strides, again to cut down on heap allocations

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+  return 6;
+}
+
+AOTITorchError aoti_torch_delete_tensor_object(AOTITensorHandle tensor) {
+  std::cout << "Deleting " << tensor << std::endl;
+  for (auto it = tensors.begin(); it != tensors.end(); ++it) {
+    if (it->get() == tensor) {
+      tensors.erase(it);
+      break; // Exit the loop once the element is found and removed
+    }
```
swolchok: AOTITensorHandle is already Tensor*, why not just make it an owning pointer and replace this implementation with `delete Tensor`?

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+  for (int i = 0; i < tensor->dim(); i++) {
+    std::cout << "size " << i << " = " << *ret_sizes[i] << std::endl;
+  }
+  return Error::Ok;
+}
+
+AOTITorchError aoti_torch_get_storage_size(
+    AOTITensorHandle tensor,
+    int64_t* ret_size) {
+  throw std::runtime_error("Cannot get storage size on ETensor");
```
swolchok: this needs to be an error or it will cause process termination

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+    AOTITensorHandle tensor,
+    int32_t* ret_dtype) {
+  *ret_dtype = static_cast<int32_t>(tensor->scalar_type());
+  return Error::Ok;
+}
+
+AOTITorchError aoti_torch_get_sizes(
+    AOTITensorHandle tensor,
+    int64_t** ret_sizes) {
+  auto it = tensor_to_sizes.find(tensor);
```
swolchok: all these functions that might possibly throw need to do something like [AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE](https://github.com/search?q=repo%3Apytorch%2Fpytorch%20AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE&type=code)

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+            << tensor->sizes()[0] << std::endl;
+  *ret_new_tensor = tensor.get();
+  return Error::Ok;
+}
+
+void checkCudaError(cudaError_t err, const char* msg) {
+  if (err != cudaSuccess) {
+    std::cerr << "Error: " << msg << " (" << cudaGetErrorString(err) << ")"
+              << std::endl;
+    exit(EXIT_FAILURE);
```
swolchok: again process termination is pretty rude; shouldn't we be propagating AOTITorchError

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+  }
+  std::cout << "self->data_ptr(): " << self->data_ptr()
+            << " sizes: " << self->sizes().data() << std::endl;
+  std::cout << "src->data_ptr(): " << src->data_ptr()
+            << " sizes: " << src->sizes().data() << std::endl;
+  for (int i = 0; i < self->dim(); i++) {
+    if (self->sizes()[i] != src->sizes()[i]) {
+      std::cout << "self.sizes()[i] " << self->sizes()[i] << ", src.sizes()[i] "
+                << src->sizes()[i] << std::endl;
+      throw std::runtime_error("size mismatch");
```
swolchok: error code

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+  cudaPointerAttributes srcAttributes, dstAttributes;
+  cudaError_t err;
+  // Get attributes of the source pointer
+  err = cudaPointerGetAttributes(&srcAttributes, src->data_ptr());
+  checkCudaError(err, "Failed to get source pointer attributes");
+  // Get attributes of the destination pointer
+  err = cudaPointerGetAttributes(&dstAttributes, self->data_ptr());
+  checkCudaError(err, "Failed to get destination pointer attributes");
+  bool srcIsDevice = srcAttributes.type == cudaMemoryTypeDevice;
+  bool dstIsDevice = dstAttributes.type == cudaMemoryTypeDevice;
```
swolchok: const bool

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+        src->data_ptr(),
+        size,
+        cudaMemcpyHostToDevice);
+    std::cout << "Host to Device copy, src data: "
+              << ((float*)src->data_ptr())[0] << std::endl;
+    checkCudaError(err, "Failed to copy from host to device");
+  } else if (!srcIsDevice && !dstIsDevice) {
+    // Host to Host copy
+    std::cout << "Host to Host copy, src data: " << ((float*)src->data_ptr())[0]
+              << std::endl;
```
swolchok: we can even keep a reasonable amount of the debug printing if it's verbose-level ET_LOG

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+      BackendInitContext& context,
+      FreeableBuffer* processed, // This will be the buffer from aoti_backend
+      ArrayRef<CompileSpec> compile_specs // This will be my empty list
+  ) const override {
+    const char* so_path = static_cast<const char*>(processed->data());
+
+    printf("so path: %s\n", so_path);
+
+    // Load the ELF using dlopen
+    void* so_handle = dlopen(so_path, RTLD_LAZY | RTLD_LOCAL);
```
swolchok: we probably just want RTLD_NOW

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+    processed->Free();
+
+    AOTInductorModelContainerCreateWithDevice =
+        reinterpret_cast<AOTInductorModelContainerCreateWithDeviceFunc>(
+            dlsym(so_handle, "AOTInductorModelContainerCreateWithDevice"));
+    if (AOTInductorModelContainerCreateWithDevice == nullptr) {
+      perror("dlsym1");
+      return Error::AccessFailed;
+    }
```
swolchok: 1) I would recommend making this a function to avoid repetition
2) it might seem like you need a macro for the reinterpret_cast, but you can do some_global_function_ptr = reinterpret_cast<decltype(some_global_function_ptr)>(dlsym_or_error(so_handle, "NameOfTheFunction")). (you might still need/want a macro for the early return)
3) perror isn't the correct call; that one prints based on errno. you want to print dlerror plus a descriptive message mentioning the missing symbol (hence the macro)

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+  // Once per execution
+  Error execute(
+      BackendExecutionContext& context,
+      DelegateHandle* handle_,
+      Span<EValue*> args) const override {
+    AOTIDelegateHandle* handle = (AOTIDelegateHandle*)handle_;
+
+    size_t num_inputs;
+    AOTInductorModelContainerGetNumInputs(
```
swolchok: must check all error codes

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+    // not using the data_ptr from outputs.
+    for (int i = 0; i < num_outputs; i++) {
+      auto args_out = args[i + num_inputs]->toTensor();
+      aoti_torch_copy_(&args_out, outputs[i], 0);
+    }
+    return Error::Ok;
+  }
+
+  void destroy(DelegateHandle* handle_) const override {
+    AOTIDelegateHandle* handle = (AOTIDelegateHandle*)handle_;
```
swolchok: can handle_ be null?

backends/aoti/runtime/AotiBackend.cpp
```diff
@@ -0,0 +1,561 @@
...
+    dlclose(handle->so_handle);
+    AOTInductorModelContainerDelete(handle->container_handle);
+    free(handle);
+    tensor_to_sizes.clear();
+    tensor_to_strides.clear();
```
swolchok: this is incorrect in the presence of other threads that might have loaded their own DSOs; probably need to make these `thread_local`

export_and_run_aoti.sh
```diff
@@ -0,0 +1,11 @@
+./install_executorch.sh
+python $1
+./install_executorch.sh --clean
+mkdir -p cmake-out
+cd cmake-out
+cmake -DEXECUTORCH_BUILD_AOTI=ON \
+      -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
+      ..
+cd ..
+cmake --build cmake-out -j9
```
swolchok: why not let cmake choose parallelism appropriately

install_requirements.py
```diff
@@ -59,7 +59,7 @@ def python_is_compatible():
 
 
 # The pip repository that hosts nightly torch packages.
-TORCH_NIGHTLY_URL = "https://download.pytorch.org/whl/nightly/cpu"
+TORCH_NIGHTLY_URL = "https://download.pytorch.org/whl/nightly/cu126"
```
swolchok: any particular reason not to use a more recent version?

exir/lowered_backend_module.py
swolchok: sounds like a good thing to reflect in a code comment?

backends/aoti/aoti_partitioner.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
swolchok: I am wondering if we could split the partitioner (or, more generally, any other part of this PR) to help the review/land/iterate process go faster

backends/aoti/runtime/AotiBackend.cpp
JacobSzwejbka: Since ET desktop also has an interest in the shim impl can that live outside the aoti backend

PR #13124, opened 2025-08-05
Arm backend: Move QuantArgs to its own file in _passes/

Sebastian-Larsson:
The is part of an effort to reduce the number of functions/classes in utility files. This class has no reason being in a utility file when it has a clear purpose on its own.


Sebastian-Larsson:
Unrelated failing test. test-models-arm-zephyr is aborted for other jobs as well.

zingo (approved):

digantdesai (approved):

PR #13126, opened 2025-08-05
NXP backend: Return number of Neutron nodes in exception

StrycekSimon:
### Summary

Adds return number of Neutron nodes in exception.

cc @robert-kalmar @roman-janik-nxp

### DIFF
--- a/backends/nxp/neutron_node_extraction.py
+++ b/backends/nxp/neutron_node_extraction.py
@@ -45,7 +45,8 @@ def extract_artifacts_from_neutron_node(
 
     if sub_graph.OperatorsLength() == 0:
         raise RuntimeError(
-            "Model converted with neutron-converter has `0` operators instead of `1`."
+            "Model converted with neutron-converter has `0` operators instead of `1`.",
+            sub_graph.OperatorsLength(),
         )
     elif sub_graph.OperatorsLength() > 1:
         builtin_operators_map: dict[int, str] = {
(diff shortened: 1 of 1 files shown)


backends/nxp/neutron_node_extraction.py
```diff
@@ -45,7 +45,8 @@ def extract_artifacts_from_neutron_node(
 
     if sub_graph.OperatorsLength() == 0:
         raise RuntimeError(
-            "Model converted with neutron-converter has `0` operators instead of `1`."
+            "Model converted with neutron-converter has `0` operators instead of `1`.",
+            sub_graph.OperatorsLength(),
```
JakeStevens, line 49: you've already hardcoded to zero, you can remove this

JakeStevens (approved):

backends/nxp/neutron_node_extraction.py
StrycekSimon: Don't know how I missed this one...

**Edit:** Apparently, it is there just to be consistent with the other errors.

robert-kalmar: @JakeStevens , maybe I missunderstood your point, but the first check verifies there is only one `subgraph` in the Neutron IR flatbuffer (subgraph is like the pytorch's method). And this check verifies now many nodes are in the first subgraph.

robert-kalmar (approved):

backends/nxp/neutron_node_extraction.py
JakeStevens: > @JakeStevens , maybe I missunderstood your point, but the first check verifies there is only one `subgraph` in the Neutron IR flatbuffer (subgraph is like the pytorch's method). And this check verifies now many nodes are in the first subgraph.

I just mean that sub_graph.OperatorsLength() does not offer any more information in this case for the RuntimeError, since sub_graph.OperatorsLength == 0 and that is explicitly written out in the string already

PR #13128, opened 2025-08-05
NXP backend: Update to Neutron Converter SDK_25.06

StrycekSimon:
### Summary

Bumps Neutron Converter SDK version and decreases min-num-ops-per-graph to revert the original Neutron Converter functionality.

### Test plan

You can test this change by running NXP backend tests since almost all of them rely on Neutron Converter.

### DIFF
--- a/examples/nxp/aot_neutron_compile.py
+++ b/examples/nxp/aot_neutron_compile.py
@@ -167,7 +167,7 @@ if __name__ == "__main__":  # noqa C901
         "-c",
         "--neutron_converter_flavor",
         required=False,
-        default="SDK_25_03",
+        default="SDK_25_06",
         help="Flavor of installed neutron-converter module. Neutron-converter module named "
         "'neutron_converter_SDK_24_12' has flavor 'SDK_24_12'.",
     )
--- a/backends/nxp/backend/neutron_converter_manager.py
+++ b/backends/nxp/backend/neutron_converter_manager.py
@@ -50,6 +50,8 @@ class NeutronConverterManager:
 
         cctx = neutron_converter.CompilationContext()
         cctx.targetOpts = neutron_converter.getNeutronTarget(target)
+        # New switch since Neutron Converter SDK_25.06
+        cctx.compilationOpts.minNumOpsPerGraph = 1
         model_converted = neutron_converter.convertModel(list(tflite_model), cctx)
 
         return bytes(model_converted)
--- a/backends/nxp/nxp_backend.py
+++ b/backends/nxp/nxp_backend.py
@@ -64,7 +64,7 @@ class NeutronCompileSpecBuilder:
         Args:
             config: Neutron accelerator configuration, e.g. "imxrt700"
             neutron_converter_flavor: Flavor of the neutron-converter module to use. Neutron-converter module named "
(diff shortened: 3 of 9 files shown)


robert-kalmar:
The parameter introduced is for `neutron_converter_SDK_25.06` version. You must update that in the test and documentation too.

StrycekSimon:
You are right, I fixed it. 

For some reason, all of the tests passed on my machine even though I had the Neutron Converter 25.03 installed. After reinstalling it, it failed as expected. Weird...

backends/nxp/TARGETS
```diff
@@ -37,7 +37,7 @@ python_library(
     name = "neutron_sdk",
     srcs = glob(["backend/**/*.py"]),
     deps = [
-       "fbsource//third-party/pypi/neutron_convertor_SDK_25_03:neutron_convertor_SDK_25_03",
+       "fbsource//third-party/pypi/neutron_convertor_SDK_25_06:neutron_convertor_SDK_25_06",
```
JakeStevens: @robert-kalmar  following up offline about this change

JakeStevens (approved):

JakeStevens:
Hi @StrycekSimon  can you please remove the change to backends/nxp/TARGETS and then we can merge this? I will make the change separately, I need to do some things atomically on our side.

JakeStevens:
@robert-kalmar  why move to draft?

robert-kalmar:
> @robert-kalmar why move to draft?

Want to make sure we upgrade the Neutron Converter in a synchronized manner with respect to filed PR's. That is to make sure I will merge at appropriate moment.

robert-kalmar:
@JakeStevens , we are ready to merge this PR if checks successful
This PR upgrades the Neutron Converter to SDK_25.06 version. Is your approval still in place?

robert-kalmar (approved):

examples/nxp/aot_neutron_compile.py
```diff
@@ -167,7 +167,7 @@ def _get_batch_size(data):
         "-c",
         "--neutron_converter_flavor",
         required=False,
-        default="SDK_25_03",
+        default="SDK_25_06",
```
digantdesai, line 170: What are your thoughts on FC/BC? I.e. PTE generated using an older SDK running on a newer compiler (i.e. BC) and vice versa?

robert-kalmar: Neutron Converter and Neutron Runtime revisions must match. There is no BC/FC compatibility defined here. 
So PTE generated with Neutron Backend using SDK_25.03 flavor of Neutron Converter, must use the MCUXpresso SDK 25.03 containing the compatible runtime libraries (Neutron NPU driver, firmware). 

Backward/Forward compatibility of the Neutron Backend itself (i.e. its ability to work wilt wide variety of Neutron Converter flavors - SDK 25.03, 25.06, 25.09) is technically feasible, but will clutter the code with additional logic for the operator support based on Neutron Converter flavor. 

At this moment, till the `Neutron Backend` is in `Prototype Quality` - https://github.com/nxp-upstream/executorch/blob/main/backends/nxp/README.md, I prefer to not clutter the code with such a logic and consider older Neutron Converters as obsolete. 

Let's take the the BC/FC strategy offline.

PR #13131, opened 2025-08-05
Added readme for Wasm extension

Conarnar:
### Summary
Added readme containing build instructions and API documentation for Wasm extension.

### Test plan
N/A

### DIFF
--- a/extension/wasm/README.md
+++ b/extension/wasm/README.md
@@ -33,7 +33,7 @@ emcmake cmake . -DEXECUTORCH_BUILD_WASM=ON \
 cmake --build cmake-out-wasm --target executorch_wasm -j32
 ```
 
-You may also use the selective build options found in the [Kernel Library Selective Build guide](../../docs/source/kernel-library-selective-build.md) or with optimized kernels using the `EXECUTORCH_BUILD_KERNELS_OPTIMIZED` option. Portable kernels are used by default.
+To reduce the binary size, you may also use the selective build options found in the [Kernel Library Selective Build guide](../../docs/source/kernel-library-selective-build.md). You may also use optimized kernels with the `EXECUTORCH_BUILD_KERNELS_OPTIMIZED` option. Portable kernels are used by default.
 
 ### Building for Web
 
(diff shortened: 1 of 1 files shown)


extension/wasm/README.md
```diff
@@ -0,0 +1,150 @@
...
+# Configure the build with the Emscripten environment variables
+emcmake cmake . -DEXECUTORCH_BUILD_WASM=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -Bcmake-out-wasm
+
+# Build the Wasm extension
+cmake --build cmake-out-wasm --target executorch_wasm -j32
+```
+
+You may also use the selective build options found in the [Kernel Library Selective Build guide](../../docs/source/kernel-library-selective-build.md) or with optimized kernels using the `EXECUTORCH_BUILD_KERNELS_OPTIMIZED` option. Portable kernels are used by default.
```
lucylq: nit: To reduce the binary size, you may ..

extension/wasm/README.md
```diff
@@ -0,0 +1,150 @@
...
+      const et = Module; // Assign Module into et for ease of use
+      const model = et.Module.load("mv2.pte");
+      // ...
+    }
+  }
+</script>
+<script src="executorch_wasm_lib.js"></script>
+```
+
+Alternatively, you may want to have your code come after the JS library code.
```
lucylq: noob q: what's the difference between having the code come before/after the JS library code?

lucylq (approved):
Thanks for the detailed readme!

extension/wasm/README.md
lucylq: Hmn, given that the wasm build is using `executorch_kernels` in https://github.com/pytorch/executorch/pull/13106, turning on `EXECUTORCH_BUILD_KERNELS_OPTIMIZED` may not work out of box?

extension/wasm/README.md
Conarnar: Having the code before the JS library code lets you create the Module object before the library initializes and customize some behaviors. This is also how the documentation recommends you do it. [Module object documentation](https://emscripten.org/docs/api_reference/module.html)

I included the part about having the code after the JS library code because it is closer to how you would set up `onRuntimeInitialized()` when using `const et = require("./executorch_wasm_lib");`, which was how I was doing it before I ran into the problem mentioned with the File System API in the unit tests. I could probably remove those parts since they aren't exactly the recommended way of doing it.

PR #13133, opened 2025-08-05
Test xnnpack with pybindings

larryliu0820:
Make sure we run xnnpack delegated model using pybindings, in `test_models.sh`.

### DIFF
--- a/.ci/scripts/test_model.sh
+++ b/.ci/scripts/test_model.sh
@@ -134,10 +134,10 @@ test_model_with_xnnpack() {
   # Delegation and test with pybindings
   if [[ ${WITH_QUANTIZATION} == true ]]; then
     SUFFIX="q8"
-    "${PYTHON_EXECUTABLE}" -m examples.xnnpack.aot_compiler --model_name="${MODEL_NAME}" --delegate --quantize --test_with_pybindings
+    "${PYTHON_EXECUTABLE}" -m examples.xnnpack.aot_compiler --model_name="${MODEL_NAME}" --delegate --quantize --test_after_export
   else
     SUFFIX="fp32"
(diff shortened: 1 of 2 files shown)


.ci/scripts/test_model.sh
```diff
@@ -131,13 +131,13 @@ test_model_with_xnnpack() {
     return 0
   fi
 
-  # Delegation
+  # Delegation and test with pybindings
   if [[ ${WITH_QUANTIZATION} == true ]]; then
     SUFFIX="q8"
-    "${PYTHON_EXECUTABLE}" -m examples.xnnpack.aot_compiler --model_name="${MODEL_NAME}" --delegate --quantize
+    "${PYTHON_EXECUTABLE}" -m examples.xnnpack.aot_compiler --model_name="${MODEL_NAME}" --delegate --quantize --test_with_pybindings
```
digantdesai: so we always want to run first with pybind and then run with the standalone runner below?

larryliu0820: I don't see any reason not to, especially from `test_model.sh`

examples/xnnpack/aot_compiler.py
```diff
@@ -62,6 +62,14 @@
         required=False,
         help="Generate and save an ETRecord to the given file location",
     )
+    parser.add_argument(
+        "-t",
+        "--test_with_pybindings",
```
jathu: nit: we are always using pybindings, so this name isn't clear for what it's actually doing. Perhaps something like `test_after_export`?

jathu (approved):

.ci/scripts/test_model.sh
digantdesai: Sounds good.

PR #13155, opened 2025-08-06
Arm Backend: Improve VGF runtime and update MLSDK setup

YufengShi-dudu:
* Adds portability extensions to instance_info
 * Explicitly sets poolSizes in descriptor pool
 * Update Arm MLSDK manifest url to default to newly publish repo during setup.sh
 * Add new exports to setup_path.sh for mlsdk components

On behalf of Ryan O'Shea

Change-Id: I3b69108c99b7a5144898a564d519d72b8ca7f702

### DIFF
--- a/backends/arm/runtime/VGFBackend.cpp
+++ b/backends/arm/runtime/VGFBackend.cpp
@@ -264,15 +264,60 @@ VkResult vkml_allocate_basics(
       .engineVersion = 0,
       .apiVersion = VK_API_VERSION_1_3,
   };
+
+  std::vector<const char*> requested_extensions;
+  VkInstanceCreateFlags instance_flags = 0;
+
+#ifdef __APPLE__
+  instance_flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
+
+  uint32_t extension_count = 0;
+  result = vkEnumerateInstanceExtensionProperties(
+      nullptr, &extension_count, nullptr);
+
+  if (result != VK_SUCCESS) {
+    ET_LOG(Error, "Failed to enumerate instance extensions");
+    return result;
+  }
+
+  std::vector<VkExtensionProperties> extension_properties(extension_count);
+  result = vkEnumerateInstanceExtensionProperties(
+      nullptr, &extension_count, extension_properties.data());
+
+  if (result != VK_SUCCESS) {
+    ET_LOG(Error, "Failed to enumerate instance extensions");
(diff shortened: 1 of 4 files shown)


robell (approved):
Looks good - was internally reviewed.

examples/arm/setup.sh
```diff
@@ -23,9 +23,9 @@ target_toolchain=""
 enable_fvps=1
 enable_vela=1
 enable_model_converter=0   # model-converter tool for VGF output
-enable_vgf_lib=0  # vgf reader - runtime backend dependency 
+enable_vgf_lib=0  # vgf reader - runtime backend dependency
 enable_emulation_layer=0  # Vulkan layer driver - emulates Vulkan ML extensions
-mlsdk_manifest_url=""
+mlsdk_manifest_url="https://github.com/arm/ai-ml-sdk-manifest.git"
```
digantdesai, line 28: From the manifest, most of them are on github.com which is awesome but I do see one on mlplatforms.org (the tosa one). FYI in the past we have seen a lot of reliability issues on that domain, keep an eye out of that if we are enabling this on the CI with high freq like pull.yaml for example.

cc @mergennachin

backends/arm/runtime/VGFBackend.cpp
```diff
@@ -264,15 +264,60 @@ VkResult vkml_allocate_basics(
       .engineVersion = 0,
       .apiVersion = VK_API_VERSION_1_3,
   };
+
+  std::vector<const char*> requested_extensions;
+  VkInstanceCreateFlags instance_flags = 0;
+
+#ifdef __APPLE__
+  instance_flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
```
digantdesai, line 272: Curious what will this do on Apple platforms?

digantdesai (approved):

zingo:
Fails are unrelated

backends/arm/runtime/VGFBackend.cpp
YufengShi-dudu: I will leave this question to @ArmRyan

robell: it's ensuring we have a bit better vulkan functional compatibility in situations where we deploy on moltenvk or other layer implementations. See https://www.lunarg.com/wp-content/uploads/2022/04/Portability-Enumeration-Extension-APR2022.pdf for the rough overview.

PR #13161, opened 2025-08-06
Save foundation weights separately

lucylq:
* __->__ #13161

This diff:
1. Introduces SerializationConfig to llm_config. Currently, this allows user to save the foundation weights in a separate file; majorly useful for lora case.
2. Adds a pass to tag foundation (non-lora) weights. This is at the top-level (export_llama_lib). The tags are preserved through run_decomps/other passes, and do not affect functionality.
3. Tags are read when placing constants into the named_data_store.
4. Tagged weights are serialized to a separate file.

Notes
1. Adding tags to node.meta['custom']['blah'] means that they will not be discarded by run_decompositions
2. Adding tags to the lifted model (ep.graph_module) requires the EP to check is_param_node for xnnpack constants. Instead, add tags to the unlifted model (ep.module()), so we do not need to go through a re-export to get the EP.
3. Not an issue for this diff as llama doesn't have any higher order ops. Adding tags to models with higher-order ops is problematic due to nested submodules.

Differential Revision: [D79181064](https://our.internmc.facebook.com/intern/diff/D79181064/)


larryliu0820 (approved):

extension/llm/export/config/llm_config.py
```diff
@@ -227,6 +227,20 @@ def __post_init__(self):
...
 
 
+@dataclass
+class SerializationConfig:
+    """
+    Configures properties relevant to the serialization process.
+
+    Attributes:
+        foundation_weights_file: configure the foundation weights of a model
+            to be placed in a separate file, external to the PTE. Pass the
```
jackzhxng: Are you expecting a lot of other fields to be here? Could we just add this a field in `ExportConfig` instead since it has to do with how we export into the pte?

lucylq: Yeah, that sounds good, and we can add a serialize config if/when there are more options for serialization.

PR #13176, opened 2025-08-07
Arm backend: Add index_tensor/index_select to unsupported u55 operators

Sebastian-Larsson:
index_tensor and index_select both become a GATHER Tosa operators, which is not support on U55.

Add test cases for U55 for these operators. The test cases verify that the ops do not get delegated.


zingo (approved):

zingo:
test fails unrelated

PR #13178, opened 2025-08-07
Arm backend: Remove submodule serialization_lib

perheld:
Remove the git submodule serialization_lib since its pointing to an old 0.80 tag and is not used by the arm backend.

Instead this library is now cloned part of the tosa-reference-module when needed.

Change-Id: I34a3707b35cab6413e8aa5365fc41d404b4eb9fa

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "backends/arm/third-party/ethos-u-core-driver"]
 	path = backends/arm/third-party/ethos-u-core-driver
 	url = https://git.gitlab.arm.com/artificial-intelligence/ethos-u/ethos-u-core-driver.git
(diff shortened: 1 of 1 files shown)


robell (approved):

zingo:
Unit test fails unrelated

zingo:
Ooops, was maybe a bit fast on this PR  @digantdesai  maybe you wanted to review/merge this, feel free to revert if it cause problems and we can figure it out.

EDIT: I revert it so we can take this change slower as it deserves

zingo:
I revert it so we can take this change slower as it deserves, sorry for the extra work

digantdesai:
If I were to guess, this should be OK. I can pull the next, new PR internally and make sure. Thanks @zingo.

PR #13086, opened 2025-08-04
Qualcomm AI Engine Direct - GA Static Qwen3

DannyYuyang-quic:
Summary:
 - support Qwen3-0.6B
 - support Qwen3-1.7B
 - refactor HF model registration for static llama

Script
``` bash
python examples/qualcomm/oss_scripts/llama/llama.py -b build-android  -s $DEVICE -m SM8750 --prompt "I would like to learn python, could you teach me with a simple example?" --temperature 0 --model_mode hybrid --prefill_ar_len 32 --max_seq_len 128 --ptq 16a4w --decoder_model qwen3
```

Stat
ee13db20-f529-413f-95c8-b6ce4bfcb4f4

### Test plan
Note: We only run Qwen3-0.6B for CI
``` bash
python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleLLMScript.test_qwen3 --model SM8650 --build_folder build-android/ --executorch_root . -s $DEVICE --artifact ./qwen3
```

cc: @haowhsu-quic, @cccclai

PR #13179, opened 2025-08-07
Qualcomm AI Engine Direct - GA Static Phi-4-mini

DannyYuyang-quic:
### Summary
 - Support Phi-4-mini-instruct for static llama path 
 - add P-ROPE for phi-4-mini
 - add EOS tok for Phi-4-mini
 
### Test plan
```
python examples/qualcomm/oss_scripts/llama/llama.py -b build-android -s $DEVICE -m SM8750 --prompt "I would like to learn python, could you teach me with a simple example?" --temperature 0 --model_mode hybrid --prefill_ar_len 32 --max_seq_len 128 --ptq 16a8w --decoder_model phi_4_mini --num_sharding 4
```

cc: @haowhsu-quic, @shewu-quic, @winskuo-quic, @cccclai

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/llama.py
+++ b/examples/qualcomm/oss_scripts/llama/llama.py
@@ -579,7 +579,7 @@ def compile(args, pte_filename, tokenizer):
                 annotate_conv=args.ptq != "16a8w",
             ),
         )
-        if args.decoder_model == {"stories110m", "stories260k"}:
+        if args.decoder_model in {"stories110m", "stories260k"}:
             custom_annotations = custom_annotations + (
                 annotate_linear_16a8w_in_affine_layer,
             )
@@ -1175,11 +1175,16 @@ def export_llama(args) -> None:
         tokenizer = AutoTokenizer.from_pretrained(model_id)
         runtime_tokenizer_path = tokenizer.save_pretrained(args.artifact)[-1]
         tokenizer = get_tokenizer(runtime_tokenizer_path)
+    elif args.decoder_model == "phi_4_mini":
+        model_id = SUPPORTED_HF_MODELS[args.decoder_model].repo_id
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        runtime_tokenizer_path = tokenizer.save_pretrained(args.artifact)[-1]
+        tokenizer = get_tokenizer(runtime_tokenizer_path)
         with open(runtime_tokenizer_path, "r+") as file:
             data = json.load(file)
             # TODO: Encountered the following error during runtime, so switched behavior for now.
-            # Error: libc++abi: terminating due to uncaught exception of type std::runtime_error: Unsupported Normalizer type: NFC.
-            data.pop("normalizer")
+            # Error: libc++abi: terminating due to uncaught exception of type std::runtime_error: invert=true is not supported for Split PreTokenizer. Only invert=false is supported.
+            data["pre_tokenizer"]["pretokenizers"][-2]["invert"] = False
             file.seek(0)
             json.dump(data, file, indent=4)
             file.truncate()
--- a/examples/qualcomm/oss_scripts/llama/model/static_llama.py
+++ b/examples/qualcomm/oss_scripts/llama/model/static_llama.py
@@ -39,6 +39,24 @@ def apply_rotary_emb_single(
     return x_out
 
 
+def apply_partial_rotary_emb_single(
+    x: torch.Tensor, freqs_cos: torch.Tensor, freqs_sin: torch.Tensor
+) -> torch.Tensor:
+
+    if x.dim() == 4:
+        freqs_cos = freqs_cos[None, :, None, :]
+        freqs_sin = freqs_sin[None, :, None, :]
+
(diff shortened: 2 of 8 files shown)


DannyYuyang-quic:
Hi @cccclai 
This PR enables GA `Phi4-mini-instruct` for static llama path.
We will submit an optimized version ASAP. 

Thanks!!

examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -1180,6 +1182,16 @@ def export_llama(args) -> None:
...
+    elif args.decoder_model == "phi_4_mini":
+        model_id = SUPPORTED_HF_MODELS[args.decoder_model].repo_id
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        runtime_tokenizer_path = tokenizer.save_pretrained(args.artifact)[-1]
+        tokenizer = get_tokenizer(runtime_tokenizer_path)
+        with open(runtime_tokenizer_path, "r+") as file:
+            data = json.load(file)
+            # TODO: Encountered the following error during runtime, so switched behavior for now.
+            # Error: libc++abi: terminating due to uncaught exception of type std::runtime_error: invert=true is not supported for Split PreTokenizer. Only invert=false is supported.
+            data["pre_tokenizer"]["pretokenizers"][-2]["invert"] = False
```
DannyYuyang-quic: Hi @cccclai, 
And here is a temporary workaround for the tokenizer error:
`invert=true is currently not supported for Split PreTokenizer, and only invert=false is allowed.`
We would appreciate it if this could be fixed.

Thanks!

cccclai: @DannyYuyang-quic is it particularly for phi tokenizer?

examples/qualcomm/oss_scripts/llama/model/static_llama.py
```diff
@@ -39,6 +39,29 @@ def apply_rotary_emb_single(
...
+    x_rot, x_pass = x[..., :rotary_dim], x[..., rotary_dim:]
+    x_r, x_i = x_rot[..., : x_rot.shape[-1] // 2], x_rot[..., x_rot.shape[-1] // 2 :]
+    x_out_r = x_r * freqs_cos - x_i * freqs_sin
+    x_out_i = x_r * freqs_sin + x_i * freqs_cos
+    x_rotated = torch.cat([x_out_r, x_out_i], dim=-1)
+    return torch.cat([x_rotated, x_pass], dim=-1)
+
+
+APPLY_ROPE_EMBEDDING_FUNCTIONS = {
+    "phi_4_mini": apply_partial_rotary_emb_single,
```
cccclai: Can this stay in this folder? https://github.com/pytorch/executorch/tree/main/examples/models/phi_4_mini/config or is it qnn specific?

examples/qualcomm/oss_scripts/llama/llama.py
DannyYuyang-quic: Thanks for the follow-up! 
I’m not sure if tokenizers from other GA models would encounter the same issue, but based on what I’ve seen so far, models like Qwen2, Qwen3, and Gemma3 don’t seem to have the `invert` kwarg in their tokenizer.

examples/qualcomm/oss_scripts/llama/model/static_llama.py
DannyYuyang-quic: It's not qnn-specific. 
`apply_partial_rotary_emb_single`(partial ROPE embedding) is required whenever the condition `partial_rotary_factor < 1.0` is met.

I've updated the condition to explicitly check for `partial_rotary_factor < 1.0`, ensuring that partial ROPE is applied only when the condition is met.

cccclai: oh wow...you're still awake...

cccclai: Can we refer to the config in this folder https://github.com/pytorch/executorch/pull/13086 instead of having phi specific logic inside static llama?

examples/qualcomm/oss_scripts/llama/model/static_llama.py
```diff
@@ -60,6 +78,11 @@ def __init__(self, config: ModelArgs, output_new_cache_only=False):
             self.q_norm_fn = torch.nn.RMSNorm(q_norm_dim, eps=config.norm_eps)
             self.k_norm_fn = torch.nn.RMSNorm(k_norm_dim, eps=config.norm_eps)
 
+        if config.partial_rotary_factor < 1:
+            self.apply_rope_emb = apply_partial_rotary_emb_single
+        else:
+            self.apply_rope_emb = apply_rotary_emb_single
```
DannyYuyang-quic, line 84: Hi @cccclai 
> Can we refer to the config in this folder https://github.com/pytorch/executorch/pull/13086 instead of having phi specific logic inside static llama?

I've updated the condition here; static llama only depends on config file now.

cccclai:
Looks like there is conflict, can you resolve?

cccclai (approved):
Thank you for enabling phi4!

PR #13181, opened 2025-08-07
Arm backend: Add cosh decomposition pass and test

emmakujala:
Add decomposition and tests for cosh


Sebastian-Larsson (approved):
Unrelated CI failures. Approved

PR #13182, opened 2025-08-07
Arm backend: Remove submodule serialization_lib

zingo:
Remove the git submodule serialization_lib since its pointing to an old 0.80 tag and is not used by the arm backend.

Instead this library is now cloned part of the tosa-reference-module when needed.

Change-Id: I34a3707b35cab6413e8aa5365fc41d404b4eb9fa
Signed-off-by: [per.held@arm.com](mailto:per.held@arm.com)


zingo:
@digantdesai maybe you want to check this out?

mergennachin (approved):

zingo:
fails unrelated

PR #13183, opened 2025-08-07
NXP Backend: Add infrastructure for pre processing passes in edge dialect

MartinPavella:
### Summary
Add infrastructure for running pre-processing passes on edge dialect programs.
Add pre-processing pass to move `view_copy` nodes into their own QDQ clusters.

### Test plan
Unit test provided.

### DIFF
--- a/backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
+++ b/backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
@@ -1,0 +1,219 @@
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+from executorch.backends.nxp.edge_passes.neutron_edge_pass import NeutronEdgePass
+from executorch.backends.nxp.neutron_partitioner import QDQClusterRecognizer
+from executorch.exir.dialects._ops import ops as exir_ops
+from torch.fx import Node
+from torch.fx.passes.infra.pass_base import PassResult
+
+
+def insert_qdq_pair_after_node(
+    graph: torch.fx.Graph, anchor: torch.fx.Node, q_params: tuple
+):
+    # Insert a Quantize node.
+    with graph.inserting_after(anchor):
+        quantize_op = graph.create_node(
+            op="call_function",
+            target=exir_ops.edge.quantized_decomposed.quantize_per_tensor.default,
+            args=(),  # Will be added later.
+        )
+        quantize_op.meta = anchor.meta
+
+    # Insert a Dequantize node.
+    with graph.inserting_after(quantize_op):
+        dequantize_op = graph.create_node(
+            op="call_function",
+            target=exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
+            args=(quantize_op,) + q_params,
+        )
+        dequantize_op.meta = quantize_op.meta
+    anchor.replace_all_uses_with(dequantize_op)
+
+    # Add this at the end, so the `anchor.replace_all_uses_with(dequantize_op)` does not replace the first use of the
+    #  `quantize_op`.
+    quantize_op.args = (anchor,) + q_params
+
+
+def _is_dequantize(node_: Node) -> bool:
+    return (
+        node_.op == "call_function"
+        and node_.target
+        == exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default
+    )
+
+
+def _is_quantize(node_: Node) -> bool:
+    return (
+        node_.op == "call_function"
+        and node_.target
+        == exir_ops.edge.quantized_decomposed.quantize_per_tensor.default
+    )
+
+
+class MoveLeadingAuxiliaryOperatorIntoSeparateQDQClusterPass(NeutronEdgePass):
+    """
+                                                           │
+                                                     ┌─────▼──────┐
+                │                                    │ dequantize │
+          ┌─────▼──────┐                             └─────┬──────┘
+          │ dequantize │                             ┌─────▼──────┐
+          └─────┬──────┘                             │ <aux_node> │
+          ┌─────▼──────┐                             └─────┬──────┘
+          │ <aux_node> │                              ┌────▼─────┐            ┐
+          └─────┬──────┘                              │ quantize │            │
+     ┌──────────▼──────────┐       replaced with      └────┬─────┘            │
+    ⋯┤ <main_cluster_node> ├⋯     ──────────────►          │                  │ newly added nodes
+     └──────────┬──────────┘                         ┌─────▼──────┐           │
+                ▼                                    │ dequantize │           │
+                ⋮                                    └─────┬──────┘           ┘
+           ┌────▼─────┐                         ┌──────────▼──────────┐
+           │ quantize │                        ⋯┤ <main_cluster_node> ├⋯
+           └────┬─────┘                         └──────────┬──────────┘
+                ▼                                          ▼
+                                                           ⋮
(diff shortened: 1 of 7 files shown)


backends/nxp/tests/executorch_pipeline.py
```diff
@@ -49,19 +52,22 @@ def to_quantized_edge_program(
...
     )
     partitioner = NeutronPartitioner(compile_spec)
-    edge_program_manager = to_edge_transform_and_lower(
-        torch.export.export(
-            exir_program_aten__module_quant, example_input, strict=True
-        ),
-        partitioner=[partitioner],
-        compile_config=EdgeCompileConfig(_check_ir_validity=False),
-    )
+    edge_program_manager = edge_program_manager.to_backend(partitioner)
```
robert-kalmar, line 92: You have introduced back the use of `export_to_edge` and `edge_program_manager.to_edge` functions, which were replaced by the `to_edge_transform_and_lower` function. The passes we introduce (edge passes) shall run after the export to edge dialect but before partitioning. The `to_edge_transform_and_lower` have optional parameter `transform_passes` which I suppose serves this purpose.
@Pop-korn, please update the code to use this call in the `executorch_pipeline.py` and `aot_neutron_compile.py` if feasible. This shall also resolve the conflict you have with the main branch. 
Also rebase to the main. Then please mark the PR as ready to merge.

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
```diff
@@ -0,0 +1,219 @@
...
+                continue
+
+            # ---- The nodes follow the pattern described in the header. ----
+
+            q_params = dequantize_node.args[1:]
+            insert_qdq_pair_after_node(graph_module.graph, aux_node, q_params)
+
+            # The graph has now changed, and we shouldn't keep iterating through it. Return the new graph and the parent
+            #  class will call this pass again.
+            return PassResult(graph_module, True)
```
robert-kalmar, line 134: @Pop-korn, in principal this does not differ from the initial draft. You return everytime, you make a modification and the caller `EdgePassManager` starts a new iteration. Can you do all the graph modification and return then? 

@digantdesai, @Pop-korn noticed in some of the passes the code iterates over a changing graph. E.g. https://github.com/pytorch/executorch/blob/main/backends/xnnpack/_passes/fuse_batch_norm.py#L41 the `graph.nodes` gets modified in-place in the https://github.com/pytorch/executorch/blob/main/backends/xnnpack/_passes/fuse_batch_norm.py#L208, everytime a match is found. 
@Pop-korn, can you please confirm? @digantdesai Is this legit and intended?

backends/nxp/tests/executorch_pipeline.py
MartinPavella: I have resolved the conflict and re-based to main.

The `to_edge_transform_and_lower` parameter `transform_passes ` only runs a single iteration of the passes. Therefore, if a pass makes a change that would allow an earlier pass to make another change, it will not get the chance to do so. The number of iterations is limited by the [`steps` attribute of the PassManager](https://github.com/pytorch/pytorch/blob/1c2cba17eab2b09d87142883da2bdbdbcf018613/torch/fx/passes/infra/pass_manager.py#L175), which cannot be set using the current interface. The only way to use a different `steps` value (that I know of) is to modify the class attribute directly (call `PassManager.steps = 10` before calling `to_edge_transform_and_lower`), which is not god practice as it has a global effect.

I understand that the `to_edge_transform_and_lower` function is the preferred way for program lowering. So do we want to:
1. Use `to_edge_transform_and_lower` and only run the passes once.
2. Use `to_edge_transform_and_lower` and globally modify the `steps` class attribute, to run multiple iterations of passes.
3. Use the current implementation. I.e. manually export to edge, run the passes ourselves (unsing `PassManager(steps=10)`), and then explicitly export to backend (with partitioning).

Which of these solutions is preferred? Or is there a different solution I am not seeing? @digantdesai

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
MartinPavella: Responding to the first paragraph: 
We can safely to all the modifications at once, and then return. We would be modifying the list of nodes while iterating over it, but it shouldn't have negative side-effects as we are only inserting nodes. The downside is that is is not good practice. Do you think that implementation would be preferred?

The second paragraph:
As far as I can tell, the [XNNPack batch_norm fusion pass](https://github.com/pytorch/executorch/blob/main/backends/xnnpack/_passes/fuse_batch_norm.py#L41) does indeed modify the graph while iterating over it. When the pass removes a `batch_norm` node, the next iteration of the `for` loop will skip a node. If (somehow) the graph contained the sequence `convolution` -> `batch_norm` -> `batch_norm`, only the first `batch_norm` node would be fused, and the second one would be skipped.

backends/nxp/tests/executorch_pipeline.py
robert-kalmar: OK, let use the solution 3 (as we have now), but please file an issue to have a look on it with the perspective of mutliple backends.

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
robert-kalmar: Let unblock it and keep the current implementation.

robert-kalmar (approved):

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
```diff
@@ -0,0 +1,219 @@
...
+          │ dequantize │                             ┌─────▼──────┐
+          └─────┬──────┘                             │ <aux_node> │
+          ┌─────▼──────┐                             └─────┬──────┘
+          │ <aux_node> │                              ┌────▼─────┐            ┐
+          └─────┬──────┘                              │ quantize │            │
+     ┌──────────▼──────────┐       replaced with      └────┬─────┘            │
+    ⋯┤ <main_cluster_node> ├⋯     ──────────────►          │                  │ newly added nodes
+     └──────────┬──────────┘                         ┌─────▼──────┐           │
+                ▼                                    │ dequantize │           │
+                ⋮                                    └─────┬──────┘           ┘
```
digantdesai, line 73: Nit 
```suggestion
                ⋮                                     └─────┬──────┘           ┘
```

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
digantdesai: > noticed in some of the passes the code iterates over a changing graph.

This pattern is used in a lot of places across the code base. This suggests it should be OK - https://github.com/pytorch/pytorch/blob/6f0f4e0c3eacd479864319127915f869f64e1935/torch/fx/graph.py#L1076-L1088

backends/nxp/tests/executorch_pipeline.py
digantdesai: Why do we need to set steps=10?

MartinPavella: In general, let's say we have two passes. A and B, which are applied in this order. Say that pass A does not modify the graph, and pass B does. After the change by pass B, pass A would suddenly be able to make a change, but by default (steps = 1), the passes wouldn't be run again.
Therefore, we want to increase the `steps` to allow for further graph optimizations. The value `10` has empirically proven as a satisfactory iteration limit.

PR #13184, opened 2025-08-07
[ET-VK] Add mechanism to trigger command buffer re-encode only when necessary

SS-JIA:
* __->__ #13184
* #13185

## Context

Dynamic shape models currently will require the command buffer to be re-encoded every inference. However, this introduces a significant overhead when running models that require dynamic shapes.

The reality is that a command buffer  re-encode may not be needed every frame. A command buffer re-encode will only be needed when:

1. Shader dispatch parameters change; i.e. new tensor sizes require a completely different compute shader, require new local work group sizing, or require new work group grid size (i.e. global work group size / local work group size)
2. Push constants containing tensor metadata need to be updated

This diff aims to reduce the overhead of triggering tensor shape change by detecting when a command buffer re-encode is actually needed.

## Changes

`ComputeGraph`:
* Introduce `requires_reencode` flag to `ComputeGraph` to indicate when a command buffer re-encode is needed.
* Introduce a `std::set<ValueRef>` tracking which values were updated when propagating tensor sizes
  * "update" can be one of two things: 1) tensor sizes changed 2) symint value changed

`DispatchNode`:
* When propagating new tensor sizes, only execute the resize function if any of the values participating in the computation have been updated
* Mark `requries_reencode` if any push constants associated with tensor metadata need to be udpated

`DynamicDispatchNode`:
* Only recompute compute shader dispatch params if any of the values participating in the computation have been updated
* Mark `requires_reencode` if 1) a new compute shader is required, 2) local work group size changed, 3) work group grid size changed

Differential Revision: [D79813237](https://our.internmc.facebook.com/intern/diff/D79813237/)


backends/vulkan/runtime/graph/ComputeGraph.h
```diff
@@ -940,6 +956,19 @@ class ComputeGraph final {
...
+  // Check if a specific ValueRef (or ValueList) was updated, with recursive
+  // handling
+  bool was_value_updated(const ValueRef value_ref) const;
+
+  // Check if a specific ValueRef (or ValueList) was updated, with recursive
+  // handling
+  bool was_value_ref_updated(const ValueRef value_ref) const;
+
+  // Set the flag to indicate that re-encoding is required
+  inline void set_requires_reencode() {
```
msluszniak: Probably `noexcept`

backends/vulkan/runtime/graph/containers/PushConstantData.h
```diff
@@ -60,6 +65,18 @@ class PushConstantDataInfo {
...
+    return tensorUniformData != nullptr;
+  }
+
+  inline void set_value(ValueRef value) {
+    value_ = value;
+  }
+
+  inline ValueRef value() const {
+    return value_;
+  }
```
msluszniak: Probably some of these member functions might be marked with `noexcept` as well

backends/vulkan/runtime/graph/ops/ExecuteNode.h
```diff
@@ -69,10 +69,12 @@ class ExecuteNode {
     (void)graph;
   }
 
-  virtual inline void trigger_resize(ComputeGraph* graph) {
+  virtual inline bool trigger_resize(ComputeGraph* graph) {
     if (resize_fn_ != nullptr) {
```
msluszniak: Inconsistent notation: Once bool operator is used for `std::function` like for: [`pick_shader_fn_`, `pick_local_wg_fn_`](https://github.com/pytorch/executorch/pull/13184/files#diff-5dbb08d9bc7ec1ddf915dc372040d8bdaa377384238d1d4b213688eb04fdb154R109) etc. Here, is explicitly compared to `nullptr`.

SS-JIA: good points! will update

andreanicastro (approved):

PR #13207, opened 2025-08-07
Require setting all inputs before execution.

shoumikhin:
Summary: Method never cared to check all the inputs were properly set before execution and basically used some default allocated memory for unset inputs. Here we introduce a safety check to guarantee users don't forget to set all inputs explicitly.

Differential Revision: D79849134


JacobSzwejbka (approved):
Tentively approving but please watch ci carefully since this is a breaking change

PR #13209, opened 2025-08-07
Add deprecation warning for to_edge + to_backend workflow in XnnpackPartitioner

leafs1:
### Summary
This PR adds a deprecation warning in the XnnpackPartitioner to guide users away from the deprecated to_edge() + to_backend() workflow and toward the recommended to_edge_transform_and_lower() flow. We inspect the call stack in the partitioner to detect when the partitioner is called from the deprecated workflow and then print out a warning statement. This helps prevent issues that can arise from the deprecated workflow.

### Test plan
Added tests testing the deprecation warning functionality in test_xnnpack_partitioner.py. Tests verify that the warning appears when using to_edge() + to_backend() and does not appear when using to_edge_transform_and_lower().

### DIFF
--- a/backends/xnnpack/test/test_xnnpack_partitioner.py
+++ b/backends/xnnpack/test/test_xnnpack_partitioner.py
@@ -1,0 +1,84 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import io
+import logging
+import unittest
+
+import torch
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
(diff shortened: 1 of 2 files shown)


backends/xnnpack/test/test_xnnpack_partitioner.py
```diff
@@ -0,0 +1,82 @@
...
+import logging
+import io
+
+import torch
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
+from executorch.exir import to_edge, to_edge_transform_and_lower
+from torch.export import export
+
+
+class TestXnnpackPartitioner(unittest.TestCase):
```
GregoryComer, line 17: Nice tests.

GregoryComer (approved):
Looks great. Thanks for adding this.

PR #13211, opened 2025-08-08
Source File Checks and Refills for Dtype Selective Zephyr Builds

BujSet:
### Summary
When Zephyr builds the portable operator library for a data type selective build, variables created during ExecuTorch's build are not transparent to the Zephyr build process. This PR introduces a check and refill it the source file lists for the portable operators are not visible to cmake. This fix will likely be helpful for other third party collaborators to use data type selective for integrating ExecuTorch with an external build system.

### Test plan
Testing with manual changes to Zephyr CI tests. This [PR](https://github.com/pytorch/executorch/pull/13191) checks out an executorch fork that already has this change. After applying the change, the zephyr model tests use data type selective build successfully.


BujSet:
Determined this should be something handled in the zephyr application code, not in the ExecuTorch repo.

PR #13219, opened 2025-08-08
Arm backend: Remove instance checks for Tosa_1.00

Sebastian-Larsson:
With tosa 0.80 removed there's no need to check which tosa version's being used.


zingo (approved):

zingo:
test fails are unrelated

PR #13220, opened 2025-08-08
Arm backend: Generate kernel registration lib from .pte

AdrianLundell:
- Moves the arm_portable_op_lib target into the executor_runner CMakeLists.txt
- Use the OPS_FROM_MODEL arg to generate the arm_portable_op_lib if possible
- Adds possibility of setting EXECUTORCH_ENABLE_DTYPE_SELECTIVE_BUILD when building the executor_runner.
- Makes cortex_m targets findeable using find_package to avoid rebuilding.

+ run cmake-format


zingo (approved):

zingo:
Test fails unrelated

examples/arm/executor_runner/CMakeLists.txt
```diff
@@ -185,6 +165,75 @@ list(
...
+    --output_path=${CURRENT_BINARY_DIR}/temp.yaml
+  OUTPUT_VARIABLE CMD_RESULT
+)
+if(NOT CMD_RESULT MATCHES "aten::")
+  set(NO_OPS_IN_FILE "true")
+else()
+  set(NO_OPS_IN_FILE "false")
+endif()
+
+if(${SEMIHOSTING} OR ${ET_BUNDLE_IO})
```
digantdesai, line 185: curious why for bundle_io?

digantdesai:
FYI @psiddh

examples/arm/executor_runner/CMakeLists.txt
AdrianLundell: It seems that .bpte are simply not supported for the EXECUTORCH_SELECT_OPS_MODEL arg, it caused a crash when building.

PR #13226, opened 2025-08-08
[ET-VK] Allow `aten.cat.default` to handle any number of input tensors

SS-JIA:
* #13227
* __->__ #13226

## Context

Previously, I updated the implementation of `aten.cat.default` in D76305343 (https://github.com/pytorch/executorch/pull/11508) since the original implementation had a bug. The new implementation only supported up to 3 input tensors, but several models require the need for up to 6 input tensors.

This diff updates the capabilities of the `concat` op so that any arbitrary number of input tensors may be accepted.

## Changes

* Update implementation of the concat shader to be able to be called repeatedly, allowing support for any number of input tensors.

Differential Revision: [D79893084](https://our.internmc.facebook.com/intern/diff/D79893084/)


backends/vulkan/test/op_tests/cases.py
```diff
@@ -1233,66 +1233,81 @@ def get_repeat_interleave_inputs():
...
+        # Cat on Height
+        ([(M, M, 3, 5), (M, M, 0, 5)], 2),
+        ([(S1, S1, 3, 5), (S1, S1, 0, 5)], 2),
+        ([(M, M, 3, 5), (M, M, 4, 5)], 2),
+        ([(S1, S1, 3, 5), (S1, S1, 4, 5)], 2),
+        ([(M2, 3, 5), (M2, 4, 5)], 1),
+        ([(S1, 3, 5), (S1, 4, 5)], 1),
+        ([(3, 5), (4, 5)], 0),
+        ([(3, 5), (4, 5), (1, 5)], 0),
+        (
```
manuelcandales, line 1246: why is this input tuple formatted differently?

backends/vulkan/test/op_tests/cases.py
```diff
@@ -1233,66 +1233,81 @@ def get_repeat_interleave_inputs():
...
-                [
-                    (XS, 1, 2, 5),
-                    (XS, 1, 2, 5),
-                    (XS, 1, 2, 5),
-                ],
-                1,
-            ),
-        ]
-    )
+    suite_inputs = [
```
manuelcandales, line 1236: do you test somewhere else what what happens when invalid inputs are provided? i.e. can you gracefully handle wrong dimensions/incompatible shapes?

manuelcandales (approved):

PR #13228, opened 2025-08-08
Set inputs directly in Module.

shoumikhin:
Summary: .

Differential Revision: D79900450


larryliu0820 (approved):

extension/apple/ExecuTorch/Exported/ExecuTorchModule.mm
```diff
@@ -331,12 +331,21 @@ - (nullable ExecuTorchMethodMetadata *)methodMetadata:(NSString *)methodName
...
+    }
+  }];
+  if (errorCode != Error::Ok) {
+    if (error) {
+      *error = ExecuTorchErrorWithCode((ExecuTorchErrorCode)errorCode);
+    }
+    return nil;
   }
-  const auto result = _module->execute(methodName.UTF8String, inputs);
+  const auto result = _module->execute(methodNameString);
```
digantdesai, line 348: For my understanding, I can still call execute again right after this without `set_input` again right? I.e. the input_set state doesn't get reset after each execute, right?

PR #13230, opened 2025-08-08
Enable strongly typed ops for deployment

mcremon-meta:
Summary:
As titled. This allows a severe reduction in codesize by only using the bare minimum cpp code. Right now, this diff only implements that option for quantized fully connected per tensor.
This diff is introducing opt level 4 to use for deployment purposes. The idea is similar to -Os in most compilers, we just use 4 for simplicity and because for now, everything should be inclusive by construction. This decision can be re-visited later!

Differential Revision: D79867630

### DIFF
--- a/backends/cadence/hifi/operators/op_quantized_fully_connected_asym8sxasym8s_asym8s_per_tensor_out.cpp
+++ b/backends/cadence/hifi/operators/op_quantized_fully_connected_asym8sxasym8s_asym8s_per_tensor_out.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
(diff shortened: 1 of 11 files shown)


backends/cadence/hifi/operators/op_quantized_fully_connected_asym8sxasym8s_asym8s_per_tensor_out.cpp
```diff
@@ -0,0 +1,65 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/backends/cadence/hifi/kernels/kernels.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+namespace cadence {
```
msluszniak, line 12: Nested namespaces with `::` are prefered since c++17, https://clang.llvm.org/extra/clang-tidy/checks/modernize/concat-nested-namespaces.html

zonglinpeng (approved):

Issue #12898, opened 2025-07-28
[Delegate Testing] Expand to individual op level tests for core-aten, quantized and non-quantized


digantdesai:
Details TBD @GregoryComer

GregoryComer:
Main stack is merged: https://github.com/pytorch/executorch/pull/12847. Closing.

PR #13238, opened 2025-08-08
[Backend Tester] Add LSTM tests

GregoryComer:
Add tests for the LSTM module. This is done in the context of https://github.com/pytorch/executorch/issues/12898.


GregoryComer:
* #13242
* #13241
* #13240
* #13239
* __->__ #13238

backends/test/suite/operators/test_lstm.py
```diff
@@ -0,0 +1,208 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai, line 1: Not sure if its an operator :P

digantdesai (approved):

backends/test/suite/operators/test_lstm.py
GregoryComer: Yeah, this is true. The operator tests could also be considered nn.Module + torch API surface tests. Open to ideas. Some backends, like Core ML (I believe), consume LSTM as a whole module, so I wanted to test it atomically.

Issue #12962, opened 2025-07-29
Set up Windows build CI


GregoryComer:
We need CI for Windows machines.

GregoryComer:
Linked PRs will enable Windows CI for the pybind CMake preset. The build is currently failing, which I'm taking a look at. I intend to enable unit tests and and a few models next.

GregoryComer:
Confirm Windows package publishing with PyTorch Core.

Edit: Confirmed with PyTorch dev infra that Windows packages will be available.

GregoryComer:
Landed with https://github.com/pytorch/executorch/pull/13702.

PR #13243, opened 2025-08-08
Always run build-presets workflow on PRs/pushes

GregoryComer:
The build-presets workflow currently runs only when the workflow file itself is updated. As we update docs to recommend using presets, we should run this job on all PRs.

Note that the Windows preset is currently reporting success despite failing to actually build. I'm intending to resolve that shortly as a follow-up.

### DIFF
--- a/.github/workflows/build-presets.yml
+++ b/.github/workflows/build-presets.yml
@@ -6,8 +6,6 @@ on:
     branches:
       - main
       - release/*
-    paths:
-      - .github/workflows/build-presets.yml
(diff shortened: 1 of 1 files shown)


GregoryComer:
* #13260
* #13259
* #13257
* #13251
* #13247
* __->__ #13243

GregoryComer:
Tracking issue: https://github.com/pytorch/executorch/issues/12962.

jathu (approved):

PR #13248, opened 2025-08-08
0.7 release ready try 5

mergennachin:
- **[Release 0.7] Update setup-linux.sh**
- **Add torch, torchvision and torchaudio dependencies**
- **Add torch into core requirements and torchaudio and torchvision into requirements-examples.txt**
- **Remove versions and let them automatically pick**

### DIFF
--- a/.ci/scripts/wheel/pre_build_script.sh
+++ b/.ci/scripts/wheel/pre_build_script.sh
@@ -15,3 +15,5 @@ set -euxo pipefail
 # able to see the installed torch package.
 
 "${GITHUB_WORKSPACE}/${REPOSITORY}/install_requirements.sh"  --example
+
+pip list
--- a/install_requirements.py
+++ b/install_requirements.py
@@ -139,9 +139,12 @@ def install_optional_example_requirements(use_pytorch_nightly):
             "install",
             "-r",
             "requirements-examples.txt",
+            "--extra-index-url",
+            TORCH_URL,
         ],
         check=True,
     )
+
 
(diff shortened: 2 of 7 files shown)


install_requirements.py
```diff
@@ -139,23 +139,6 @@ def install_optional_example_requirements(use_pytorch_nightly):
...
-        "torchaudio==2.8.0" if use_pytorch_nightly else "torchaudio",
-    ]
-    # Then install domain libraries
-    subprocess.run(
-        [
-            sys.executable,
-            "-m",
-            "pip",
-            "install",
-            *DOMAIN_LIBRARIES,
```
larryliu0820, line 158: I think we still need to keep this? For install from source case

.ci/scripts/wheel/pre_build_script.sh
```diff
@@ -15,3 +15,5 @@ set -euxo pipefail
 # able to see the installed torch package.
 
 "${GITHUB_WORKSPACE}/${REPOSITORY}/install_requirements.sh"  --example
```
larryliu0820, line 17: Here it should install domain libraries right?

install_requirements.py
```diff
@@ -139,23 +139,6 @@ def install_optional_example_requirements(use_pytorch_nightly):
             "install",
             "-r",
             "requirements-examples.txt",
```
larryliu0820, line 141: ah nvm it will be installed here.

larryliu0820 (approved):

PR #13250, opened 2025-08-08
[Backend Tester] Add portable test flow

GregoryComer:
Add a portable tester flow to the backend tester. This is mainly intended to help validate the test suite and to get baseline numbers to compare against delegated results.


GregoryComer:
* #13313
* #13312
* #13311
* #13309
* #13308
* #13306
* #13258
* __->__ #13250

digantdesai (approved):
This is good. 

@manuelcandales is planning to add a way to run through torch.nn.Module for Portable library testing (w/ Facto). This PR wants me to up the discussion we had @GregoryComer about using Facto here. And more generally make it possible to leverage Facto for backend hardening, which may require some changes in Facto, as @manuelcandales and I also discussed.

Can we align on this? And also on the long term direction?

Stamping this PR so you are not blocked Gregory but let's sync.

PR #13247, opened 2025-08-08
Fix error reporting in Windows preset build job

GregoryComer:
Update the Windows preset build job to accurately report failure. Previously, it was silently failing.

I've updated the script to explicitly check the last exit code. I'm not entirely sure why ErrorActionPreference=Stop isn't doing this. I played around with the escaping a bit and also added PSNativeCommandUseErrorActionPreference = True, but haven't gotten it to fail on exit code. It might be possible to make this work, but I'm just adding an explicit check for now.

PR #13251, opened 2025-08-09
Fix devtools CMake build failure on Windows

GregoryComer:
Update the devtools bundled program build to work on Windows by using the non build interface string to create the directory. I'm not entirely sure why this was working on Mac/Linux, but looking at the other devtools cases where a directory was created at configuration time(?), it was using the non build interface path variable (DEVTOOLS_INCLUDE_DIR_NO_BUILD_INTERFACE), so this would appear to be the correct behavior.

Note that the Windows pybind job is still failing on this PR, but is failing later. Without this change, the CMake configuration step fails. Now, it fails in the actual build - one step closer.


GregoryComer:
* #13567
* #13564
* #13260
* #13259
* #13257
* __->__ #13251
* #13247

GregoryComer:
Note that the pybind preset failure is expected in the PR. It is resolved further up, as I incrementally fix issues. The other CI failures are pre-existing.

GregoryComer:
@swolchok Can you sanity check this change? From looking at related code and your recent changes, I believe this is correct, but wanted to double check with you.

swolchok (approved):
Not sure I understand exactly why this is needed (what's the failure?) but it seems fine to me -- BUILD_INTERFACE prevents things from showing up post-installation and that doesn't seem to be relevant here.

PR #13255, opened 2025-08-09
verify regression test

mergennachin:
- **RPATH Fix for portable_lib Python Extension**
- **Revert the rpath to verify regression test works**


mergennachin:
See if the regression test is actually working, in other words, failing before the change. https://github.com/pytorch/executorch/pull/13254

PR #13258, opened 2025-08-09
[Backend Tester] Clean up a few test issues

GregoryComer:
There are a few broken tests that need cleaning up. Some are failing due to missing portable kernels. These tests are now skipped if any unsupported portable ops remain post-delegation. I also fixed a few other small issues and bumped the element-wise tolerance to reduce false positives. SNR should hopefully catch most blatant correctness issues. The fp16 and quantized tests can generate occasional high element-wise error but still have decent SNR (~60+).


GregoryComer:
* #13360
* #13313
* #13312
* #13311
* #13308
* #13306
* __->__ #13258

backends/test/suite/reporting.py
```diff
@@ -28,35 +28,32 @@ class TestResult(IntEnum):
     SUCCESS_UNDELEGATED = 1
     """ The test succeeded without the backend delegating anything. """
 
-    EAGER_FAIL = 2
-    """ The test failed due to the model failing to run in eager mode. """
+    SKIPPED = 2
```
digantdesai, line 31: Nit:
```suggestion
    SKIPPED_NON_BACKEND_FAILURE = 2
```

backends/test/suite/runner.py
```diff
@@ -9,6 +9,14 @@
 
 import torch
 
+# Set of unsupported ops that should cause tests to be skipped
+UNSUPPORTED_PORTABLE_OPS = {
```
digantdesai, line 13: If we are adding a Portable flow, how would these show up there? Some `PTE_FAIL`?

backends/test/suite/runner.py
```diff
@@ -142,12 +159,15 @@ def build_result(
             tester.run_method_and_compare_outputs(
                 inputs=None if generate_random_test_inputs else inputs,
                 statistics_callback=lambda stats: error_statistics.append(stats),
+                atol=1e-1,
```
digantdesai, line 162: atol seems pretty high for a general default, no?

>  The fp16 and quantized tests can generate occasional high element-wise error but still have decent SNR (~60+).

Do you know how this is tested on PyTorch/PyTorch side? >60 SNR is good but outliers are not great esp for individual ops, if they are expected I would prefer if we set them per test basis which would allow us to reason about the math being done on that specific test warrenting high ATOL/RTOL.

GregoryComer: Discussed offline. We will re-evaluate this in the future. I've created https://github.com/pytorch/executorch/issues/13347 as a backlog task to track.

backends/test/suite/runner.py
GregoryComer: Fair question. They won't be reported as failures. I think that's mostly fine, since the main goal is to validate backends. That being said, we could only skip these tests for flows that are expected to delegate - though backend op libs might make it trickier. I'll make this change.

digantdesai (approved):

PR #13259, opened 2025-08-09
Fix XNNPACK build on Windows

GregoryComer:
Fix a few build-blocking issues on Windows. Enable XNNPACK on the Windows preset. This also gives CI coverage.

I've disabled 4 microkernel families that don't build currently. These likely need to be gated or fixed upstream, but I'm disabling them locally for now to unblock usage. They are relatively niche and should have little to no perf impact on the average user.


GregoryComer:
* #13567
* #13667
* #13564
* #13260
* __->__ #13259
* #13672

backends/xnnpack/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ foreach(fbs_file ${_xnnpack_schema__srcs})
   )
 endforeach()
 
-if(WIN32 AND NOT CMAKE_CROSSCOMPILING)
+if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
```
jathu, line 62:
```suggestion
if(WIN32)
```

GregoryComer: I believe this will actually not work, as WIN32 is set when the target in Windows, not the host. If we're cross compiling (for Android, for example), from a Windows host, we'll still want to use Powershell. Though feel free to correct me if I'm misunderstanding.

GregoryComer:

PR #13260, opened 2025-08-09
Fix portable library build on Windows

GregoryComer:
Enable build of the portable operator library (and executor_runner, now that we have portable) on the Windows preset. There are a couple of patterns that weren't building on Windows. Both std::nan and std::signbit complain about ambiguous overload resolution when given an integer argument.

To solve the std::isnan issue, I introduced isnan_override in math_util.h, following the naming pattern from max_override / min_override / etc. in the same file. It adds a specialized branch for integers to avoid calling std::isnan. For std::signbit, I just locally specialized.

I also re-enabled optimized and quantized kernel libraries in the Windows preset, as they now build with this change.


GregoryComer:
* #13567
* #13667
* #13564
* __->__ #13260
* #13259
* #13672

GregoryComer:
@manuelcandales Any thoughts or preference on the approach used here?

swolchok (approved):
I was skeptical, but Google (and Godbolt https://godbolt.org/z/Tq8Tr9aWz) confirms this is a well-documented problem that happens when using MSVC. Thanks!

kernels/portable/cpu/util/math_util.h
```diff
@@ -29,7 +33,8 @@ template <
     typename std::enable_if<std::is_integral<INT_T>::value, bool>::type = true>
 INT_T floor_divide(INT_T a, INT_T b) {
   const auto quot = a / b;
-  if (std::signbit(a) == std::signbit(b)) {
+  // MSVC does not like signbit on integral types.
```
JacobSzwejbka: lmao

JacobSzwejbka (approved):

GregoryComer:

PR #13264, opened 2025-08-10
Add TorchAO wrapper config to allow filter_fn for quantize_

abhinaykukkadapu:
Changes: 
1. Support filter function in quantize_ function when using torchao quantize.
2. Update unittests accordingly
3. Use `ComposableQuantizer` if there are multiple quantizers and is of type torchao, for legacy quantizers use them directly with prepare_pt2e.
4. Source transform modifies model inplace, so deep copy first to avoid modifying user provided model.


abhinaykukkadapu:
* #13265
* __->__ #13264

backends/xnnpack/test/recipes/test_xnnpack_recipes.py
```diff
@@ -38,6 +39,19 @@ def check_fully_delegated(self, program: Program) -> None:
...
+        self, session, example_inputs, atol: float
+    ) -> None:
+        """Utility to compare eager quantized model output with session output after coreml lowering"""
+        source_transform_output = session.get_stage_artifacts()[
+            StageType.SOURCE_TRANSFORM
+        ]
+        eager_quantized_model = source_transform_output.data["forward"]
+        output = session.run_method("forward", example_inputs[0])[0]
+        expected = eager_quantized_model(*example_inputs[0])
+        self.assertTrue(torch.allclose(output, expected, atol=atol))
```
digantdesai: You might want to print more stats if this fails - see https://github.com/pytorch/executorch/blob/main/backends/test/harness/tester.py#L337

backends/xnnpack/test/recipes/test_xnnpack_recipes.py
```diff
@@ -70,12 +78,8 @@ def test_int8_dynamic_quant_recipe(self) -> None:
...
                         export_recipe=export_recipe,
                     )
-                    self.assertTrue(
-                        torch.allclose(
-                            session.run_method("forward", example_inputs[0])[0],
-                            m_eager(*example_inputs[0]),
-                            atol=1e-1,
-                        )
+                    self._compare_eager_quantized_model_outputs(
+                        session, example_inputs, 1e-1
```
digantdesai, line 94: atol? Why is this so high for two linears?
```suggestion
                        session, example_inputs, atol=1e-1
```

backends/xnnpack/recipes/xnnpack_recipe_provider.py
```diff
@@ -144,14 +145,16 @@ def _build_int8da_intx_weight_recipe(
         else:
             weight_granularity = PerGroup(group_size=group_size)
 
-        config = Int8DynamicActivationIntxWeightConfig(
-            weight_dtype=weight_dtype,
-            weight_granularity=weight_granularity,
+        config = AOQuantizationConfig(
```
digantdesai, line 156: assert weight_dtype == torch.int4 because XNNPACK doesn't support anything but 4 for groupwise. For channelwise it can do both int4 and int8.

export/stages.py
```diff
@@ -331,6 +330,38 @@ def valid_predecessor_stages(self) -> List["StageType"]:
...
+            )
+
+        if torch_ao_quantizers:
+            # prepare_pt2e has backward compat with torch.ao quantizer
+            return torch_ao_quantizers[0]
+        elif torchao_pt2e_quantizers:
+            # Multiple torchao quantizers - use ComposableQuantizer
+            from torchao.quantization.pt2e.quantizer import ComposableQuantizer
+
+            return ComposableQuantizer(torchao_pt2e_quantizers)
```
digantdesai, line 365: test for this?

export/stages.py
```diff
@@ -331,6 +330,38 @@ def valid_predecessor_stages(self) -> List["StageType"]:
...
+            if isinstance(quantizer, TorchAOPT2EQuantizer):
+                torchao_pt2e_quantizers.append(quantizer)
+            else:
+                torch_ao_quantizers.append(quantizer)
+
+        if torch_ao_quantizers and torchao_pt2e_quantizers:
+            raise ValueError("Mixed quantizer types are not supported")
+        if len(torch_ao_quantizers) > 1:
+            raise ValueError(
+                "Multiple quantizers of torch.ao.quantization.quantizer not supported"
```
metascroy, line 357: Doesn't torchao already detect this and give an error if mixing?  I thought I added that

export/stages.py
```diff
@@ -331,6 +330,38 @@ def valid_predecessor_stages(self) -> List["StageType"]:
     def can_start_pipeline(self) -> bool:
         return True
 
+    def _get_quantizer_for_prepare_pt2e(self, quantizers: List[Any]):
+        torch_ao_quantizers = []
```
metascroy, line 340: I want to make sure this API does not invite people to use torch.ao quantizers.

We've migrated to torchao quantizers in ET and torch.ao will be deprecated.  The only backend that uses torch.ao in ET right now is CoreML (because it has an external depedency on coremltools adopting the torchao quantizer)

abhinaykukkadapu: I can create a follow up ticket to deprecate torch.ao support, once coreml moves to torchao. Currently we already support both torch.ao and torchao via prepare_pt2e's backward compatibility anyway.

export/stages.py
abhinaykukkadapu: May be the torchao version is different?

export/stages.py
metascroy: I guess this is a private method, so maybe it's OK.  I just want to make sure we aren't exposing this publicly

metascroy:
Overall, looks good.  Address comments before merging

metascroy (approved):

backends/xnnpack/test/recipes/test_xnnpack_recipes.py
abhinaykukkadapu: Yeah, i think 1e-2 is working on my mac, will check if linux passes on CI. Nevertheless, i'm updating the tolerance tests similar to CoreML (let me know if there is any objection) to use sqnr to compare eager model vs lowered model output.

But use tolerance checks to compare post quantized model and lowered model.

PR #13266, opened 2025-08-10
Experimental GGUF-2-PTE Converter

dillondesilva:
## Summary

**This PR is not intended for merge.** Instead, it is designed to demonstrate a potential method under which `.gguf` files can be converted to `.pte` by leveraging some of the existing `transformers` ecosystem.

The key idea is the following:

1. A GGUF model is loaded via the `transformers` library into a suitable auto class.
2. Sample tokens are generated using the model tokeniser and a dummy sentence.
3. GGUF models are then torch exported
4. The exported torch program is then lowered and exported to an Executorch format .pte

### Attached code generates a `.pte` model that is yet to be tested

The experiment in this PR converts `SmolLM2-135M-Instruct-Q8_0.gguf` into a `.pte` file. However, testing if this model works as expected within the executorch runtime is to be confirmed. This may also require some conversion of the `tokenizer` (I'm not too sure but would be interested to know, its probably in the docs somewhere)?

### Torch export errors can be...scary

Conceptually, I thought that running this experiment would hopefully not be too difficult. However some of the setup issues around getting a reliable torch export for certain models proved to be quite challenging. This could just be due to my lack of knowledge on `torch.export`, however I think it may also offer insight into the opinion of a developer who wants to focus more on having a smooth experience with converting to `.pte` files.

- **Smoothness of conversion depends largely on model being converted:** Tried converting a GGUF for LFM-2 using the same code attached (only changing `model_id` and `filename`) and its crazy how scary some of the errors became. Complex traces and rooted in various operators/potentially unsupported ops appeared in the logs. Of course, LFM-2 is quite a new model so perhaps I gave it an unfair example but nonetheless, it could be interesting to see where's the limit of models that can comfortably converted via this workflow. 

cc @lucylq

### DIFF
--- a/examples/converters/gguf2pte.py
+++ b/examples/converters/gguf2pte.py
@@ -14,9 +14,9 @@ import torch
 model_id = "bartowski/SmolLM2-135M-Instruct-GGUF" # Here we would have our HF model in GGUF form we wish to convert
 filename = "SmolLM2-135M-Instruct-Q8_0.gguf"
 
-torch_dtype = torch.float32
 tokenizer = AutoTokenizer.from_pretrained(model_id, gguf_file=filename)
 model = AutoModelForCausalLM.from_pretrained(model_id, gguf_file=filename)
+print(f"Model weights dtype: {model.dtype}")
 model.eval()
 
 # Generate some sample input for our torch export
--- a/examples/converters/gguf2pte.py
+++ b/examples/converters/gguf2pte.py
@@ -1,0 +1,38 @@
+'''
+Example to convert .gguf files into .pte format.
+
+1. Load our model using transformers/gguf
+2. Torch export
+3. Executorch lowering and export to .pte
+'''
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from executorch.exir import to_edge_transform_and_lower
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
+from torch.export import export
+import torch
+
+model_id = "bartowski/SmolLM2-135M-Instruct-GGUF" # Here we would have our HF model in GGUF form we wish to convert
+filename = "SmolLM2-135M-Instruct-Q8_0.gguf"
+
+torch_dtype = torch.float32
+tokenizer = AutoTokenizer.from_pretrained(model_id, gguf_file=filename)
+model = AutoModelForCausalLM.from_pretrained(model_id, gguf_file=filename)
+model.eval()
+
+# Generate some sample input for our torch export
+sample_inputs = tokenizer("Plants create energy through a process known as", return_tensors="pt",)
+print(sample_inputs)
+print(sample_inputs["input_ids"].shape)
+print(sample_inputs["attention_mask"].shape)
+
+sample_inputs = (sample_inputs["input_ids"], sample_inputs["attention_mask"],)
+
+# Torch export followed by ET lowering and export
+exported_program = export(model, sample_inputs)
+executorch_program = to_edge_transform_and_lower(
+    exported_program,
+    partitioner = [XnnpackPartitioner()]
+).to_executorch()
+
+with open("model.pte", "wb") as file:
+    file.write(executorch_program.buffer)
--- a/examples/converters/requirements.txt
+++ b/examples/converters/requirements.txt
@@ -1,0 +1,6 @@
+accelerate
+gguf
+setuptools
+transformers
+executorch
+torch


examples/converters/gguf2pte.py
```diff
@@ -0,0 +1,38 @@
...
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
+from torch.export import export
+import torch
+
+model_id = "bartowski/SmolLM2-135M-Instruct-GGUF" # Here we would have our HF model in GGUF form we wish to convert
+filename = "SmolLM2-135M-Instruct-Q8_0.gguf"
+
+torch_dtype = torch.float32
+tokenizer = AutoTokenizer.from_pretrained(model_id, gguf_file=filename)
+model = AutoModelForCausalLM.from_pretrained(model_id, gguf_file=filename)
```
metascroy, line 18: @dillondesilva what dtype are the weights after loading a GGUF model?  Are they dequantized to FP32?

If so, I'm not sure this is really a converter in the sense that it doesn't preserve the quantization from GGUF.

But it is a good start, especially for getting the model structure.  We just need to parse the GGUF weights and convert them to int_data/scales/zeros so we can reroute to a kernel.  We did have a rudimentary converter for GGUF in [torchchat](https://github.com/pytorch/torchchat/blob/main/torchchat/utils/gguf_loader.py) that supported Q4_0 and Q6_K, but this is no longer a popular format.

We could probably start by trying to support Q4_K_M, which requires support for  Q4_K and Q6_K.  Here is a vibe-coded version of this for Q4_K (so no guarantee that it's correct, but it looks reasonable):

```
# pip install gguf numpy
import numpy as np
import gguf

# ---- helpers ----
def _fp16le_to_f32(buf_mv):
    return np.frombuffer(buf_mv, dtype="<f2", count=1).astype(np.float32)[0]

def _unpack_q4k_scale_min_codes(bytes12: memoryview):
    """Return two arrays (8,) of 6-bit integers for sub-block scales and mins."""
    b = np.frombuffer(bytes12, dtype=np.uint8)
    # Layout per llama.cpp wiki ("Tensor Encoding Schemes"):
    #  0: EEAAAAAA   1: FFBBBBBB   2: GGCCCCCC   3: HHDDDDDD
    #  4: eeaaaaaa   5: ffbbbbbb   6: ggcccccc   7: hhdddddd
    #  8: eeeeEEEE   9: ffffFFFF  10: ggggGGGG  11: hhhhHHHH
    S0_3 =  b[0:4] & 0x3F
    S4_7 = ((b[0:4] >> 6) & 0x03) | ((b[8:12] >> 4) << 2)

    M0_3 =  b[4:8] & 0x3F
    M4_7 = ((b[4:8] >> 6) & 0x03) | ((b[8:12] & 0x0F) << 2)

    S = np.concatenate([S0_3, S4_7]).astype(np.float32)  # (8,)
    M = np.concatenate([M0_3, M4_7]).astype(np.float32)  # (8,)
    return S, M

def extract_q4k(gguf_path: str, tensor_name: str):
    """
    Returns:
      q_codes  : (n_super, 256) uint8  -- 4-bit codes per superblock (values 0..15)
      scales   : (n_super, 8)  float32 -- per-subblock scale (real units)
      mins     : (n_super, 8)  float32 -- per-subblock min/offset (real units)
      d, dmin  : (n_super,)    float32 -- super-scales used to decode the 6-bit fields
    Notes:
      - Each superblock covers 256 weights = 8 sub-blocks * 32 each.
      - Reconstruct weights for sub-block j:  w = scales[i,j] * q - mins[i,j]
      - Zero-point (affine form): z = mins / scales  (can be fractional)
    """
    r = gguf.GGUFReader(gguf_path)
    t = r.tensors_map[tensor_name]
    raw = memoryview(t.data)

    # Superblock layout (Q4_K):
    # [d fp16][dmin fp16][12B packed S/M codes][128B 4-bit codes]
    stride = 2 + 2 + 12 + 128  # 144 bytes
    n_super = len(raw) // stride
    assert len(raw) % stride == 0, "Unexpected Q4_K tensor byte length"

    d     = np.empty(n_super, dtype=np.float32)
    dmin  = np.empty(n_super, dtype=np.float32)
    S_all = np.empty((n_super, 8), dtype=np.float32)
    M_all = np.empty((n_super, 8), dtype=np.float32)
    Q_all = np.empty((n_super, 256), dtype=np.uint8)

    off = 0
    for i in range(n_super):
        # two fp16 super-scales
        d[i]    = _fp16le_to_f32(raw[off:off+2]); off += 2
        dmin[i] = _fp16le_to_f32(raw[off:off+2]); off += 2

        # packed 6-bit sub-scales / sub-mins
        s12 = raw[off:off+12]; off += 12
        S6, M6 = _unpack_q4k_scale_min_codes(s12)

        # realize to real units
        S_all[i, :] = d[i]    * S6
        M_all[i, :] = dmin[i] * M6

        # 128 bytes => 256 4-bit codes
        codes_b = np.frombuffer(raw[off:off+128], dtype=np.uint8); off += 128
        q_low   = (codes_b & 0x0F).astype(np.uint8)
        q_high  = (codes_b >> 4).astype(np.uint8)
        Q_all[i, 0::2] = q_low
        Q_all[i, 1::2] = q_high

    return Q_all, S_all, M_all, d, dmin

# ---- Example usage ----
# q, s, m, d, dmin = extract_q4k("model.gguf", "model.layers.0.self_attn.q_proj.weight")
# # Dequantize one superblock 'i', sub-block j (32 weights):
# i, j = 0, 3
# w_block = s[i, j] * q[i, j*32:(j+1)*32].astype(np.float32) - m[i, j]
# # Optional affine form zero-point:
# z_block = m[i, j] / s[i, j]
```  

Now we don't currently have any quantized kernels that will handle floating point zeros (in XNNPACK or elsewhere), but I could quickly put up a patch to support that for our [lowbit kernels](https://github.com/pytorch/executorch/blob/86a53df3fc7bd8e4f458ee94d2b7784064531a86/examples/models/llama/README.md#running-with-low-bit-kernels) in a day or two.

lucylq:
cc @swolchok on gguf-pte conversion

swolchok:
Are the models in the `transformers` library guaranteed to be exportable? I was under the impression that we generally needed to curate exportable versions of LLMs at this stage in our development, but perhaps I am out of date.

Also CC @mergennachin

examples/converters/gguf2pte.py
lucylq: Thanks for the example, the flow looks quite clean. Agree with @metascroy that we may need some custom weight conversion.

I was imagining we could export a PTE file without weights, and plug in gguf weights at runtime, but that also requires some more work on export/runtime before it's possible.

dillondesilva: Good catch about the weights being dequantized. I pushed a quick update and it does seem that the GGUF weights are dequantized to FP32 (also found it on the [docs](https://huggingface.co/docs/transformers/en/gguf)) 

As you've mentioned, it would be great to have some sort of a conversion module we route the model through once the GGUF has been loaded by HF.

What would be the best path forward for development? Do we want an RFC/some abstractions in this PR we can use to capture this process + any additional steps (e.g. dtype conversion)?

Screenshot 2025-08-12 at 8 31 14 pm

dillondesilva:
> Are the models in the `transformers` library guaranteed to be exportable? I was under the impression that we generally needed to curate exportable versions of LLMs at this stage in our development, but perhaps I am out of date.
> 
> Also CC @mergennachin

@swolchok Good point! Welp I haven't done a detailed analysis on this but I think its largely dependent on model architecture/operations within it - feel free to have a play around with changing the `model_id` + `filename` in the attached script. Last I checked, [LFM2-350M-GGUF](https://huggingface.co/LiquidAI/LFM2-350M-GGUF) did not have a smooth export experience.

Perhaps there's some commonalities between what exports well and what doesn't. If we investigate this, it could help understand the limitations of what is/isn't exportable.

swolchok:
> curate exportable versions of LLMs

I would definitely like to know how the set of exportable LLMs in `transformers` compares to the set of exportable LLMs in our examples directory. The idea of letting `transformers` be in the business of knowing how to apply GGUF hyperparameter settings to models rather than maintaining our own set has been growing on me, though we would need to figure out whether `transformers` would be OK with 1) not regresssing existing exportability 2) increasing it or accepting patches to increase it over time.

examples/converters/gguf2pte.py
swolchok: (I mentioned some of this on Discord; commenting here for more legibility.)

I think we will want to preserve the quantized weights. IMO, ideally we would have the following flow:
1) Have transformers ingest GGUF and produce model class with hyperparameters configured appropriately.
2) Export configured model and delegate to ExecuTorch. Produced PTE does not contain weights; instead, it references the original GGUF as though it was a PTD.

If we convert the weights or copy them into a PTE, we will IMO be likely to get one or more of slow inference (from lack of quantization), double (or more) peak disk space requirement, and/or double (or more) peak RAM requirement.

Regarding quantization, I think it would be easiest to decouple the rest of the work that needs to be done from quantization by starting with only LLAMA_FTYPE_ALL_F32 and fast-following with LLAMA_FTYPE_MOSTLY_F16 and LLAMA_FTYPE_MOSTLY_BF16. Then, adding Q4_K_M support can be done separately; I would propose to do that with a torchao-style model rewrite that replaces the matmuls with custom ops, which we then preserve during the actual delegation & lowering process. (If I understand correctly, this is more or less how lowbit lowering works today.)

dillondesilva:
I think I've made a few interesting findings:
1. Comparing exportable LLMs in transformers to exportable LLMs in Executorch: Inside [`src/transformers/integrations/ggml.py`](https://github.com/huggingface/transformers/blob/a2b37bfd58c55e0f0b4b3a73f2f1d177b9094437/src/transformers/integrations/ggml.py#L36), there are several GGUF config mappings specified under `GGUF_CONFIG_MAPPING. Key models that seem to be supported are: llama, mistral, qwen2, qwen2moe, qwen3, qwen3_moe, falcon, phi3, stablelm, gpt2, starcoder2, bloom, t5, mamba, nemotron, gemma2, gemma3, tokenizer

3. Building on the above, it looks like the actual mapping of GGUF to transformers models takes place in [`src/transformers/modeling_gguf_pytorch_utils.py`](https://github.com/huggingface/transformers/blob/a2b37bfd58c55e0f0b4b3a73f2f1d177b9094437/src/transformers/modeling_gguf_pytorch_utils.py). There's a lot of architecture-specific tensor conversion code but `load_gguf_checkpoint` seems to be where everything gets put together and the actual dequantization of weights is occurring. I think the `dequantize` method in use is just straight from the [Python GGML library](https://github.com/ggml-org/llama.cpp/blob/master/gguf-py/gguf/quants.py#L67) which is just calling `Tensor.view()` to cast to float32. 
4. Unfortunately, it looks like the `load_gguf_checkpoint` method does not preserve the original, quantized dtype in the returned tensors but we could possibly create an issue on this? If we could preserve the quantized dtype somehow per tensor (perhaps as metadata), then it might be handy as a reference structure we can use for bringing back the quantized model.

swolchok:
> 1. Comparing exportable LLMs in transformers to exportable LLMs in Executorch

Sorry, how do you know this set of models is torch.exportable? Did you try them all?

dillondesilva:
> > 1. Comparing exportable LLMs in transformers to exportable LLMs in Executorch
> 
> 
> 
> Sorry, how do you know this set of models is torch.exportable? Did you try them all?

I haven't tried them all. My line of thinking was more that seeing how those models torch export after loading their GGUFs via transformers might narrow the scope down a bit. But it's also possible this isn't a correct connection😅

metascroy:
> I think I've made a few interesting findings:
> 
> 1. Comparing exportable LLMs in transformers to exportable LLMs in Executorch: Inside [`src/transformers/integrations/ggml.py`](https://github.com/huggingface/transformers/blob/a2b37bfd58c55e0f0b4b3a73f2f1d177b9094437/src/transformers/integrations/ggml.py#L36), there are several GGUF config mappings specified under `GGUF_CONFIG_MAPPING. Key models that seem to be supported are: llama, mistral, qwen2, qwen2moe, qwen3, qwen3_moe, falcon, phi3, stablelm, gpt2, starcoder2, bloom, t5, mamba, nemotron, gemma2, gemma3, tokenizer
> 2. Building on the above, it looks like the actual mapping of GGUF to transformers models takes place in [`src/transformers/modeling_gguf_pytorch_utils.py`](https://github.com/huggingface/transformers/blob/a2b37bfd58c55e0f0b4b3a73f2f1d177b9094437/src/transformers/modeling_gguf_pytorch_utils.py). There's a lot of architecture-specific tensor conversion code but `load_gguf_checkpoint` seems to be where everything gets put together and the actual dequantization of weights is occurring. I think the `dequantize` method in use is just straight from the [Python GGML library](https://github.com/ggml-org/llama.cpp/blob/master/gguf-py/gguf/quants.py#L67) which is just calling `Tensor.view()` to cast to float32.
> 3. Unfortunately, it looks like the `load_gguf_checkpoint` method does not preserve the original, quantized dtype in the returned tensors but we could possibly create an issue on this? If we could preserve the quantized dtype somehow per tensor (perhaps as metadata), then it might be handy as a reference structure we can use for bringing back the quantized model.

On 3:

It sounds like you more or less have FP32 working with GGUF because it's supported via HF natively with something like this:

```
model = AutoModelForCausalLM.from_pretrained(model_id, gguf_file=filename)
```

Barring modifying transformers to support loading the quantized weights, to get quantization working AOT, I think you need to write another utility like load_quantized_state_dict.  You can then do something like this:

```
model = AutoModelForCausalLM.from_pretrained(model_id, gguf_file=filename)
state_dict = load_quantized_state_dict(gguf_file)
model.load_state_dict(state_dict, assign=True)

# Continue with export as usual
```

The method load_quantized_state_dict can use the gguf.GGUFReader from the gguf python package to parse the quantized weights.

We had a very rough working example of this for Q4_0 in torchchat: https://github.com/pytorch/torchchat/blob/main/torchchat/utils/gguf_loader.py#L664-L667 (and it includes a Q6_K parser).

To support Q4_K_M (which is a mix of Q4_K and Q6_K tensors), you just need to write a Q4_K parser (and copy the Q6_K from the above torchchat code).  I outlined a vibe-coded Q4_K parser above.

PR #13269, opened 2025-08-10
Whisper targets

rohansjoshi:
Summary: BUCK/TARGETS files to build qnn_whisper_runner and run whisper.py script

Differential Revision: D79949336


cccclai (approved):
Thanks!

PR #13271, opened 2025-08-11
Arm backend: Use bucket approach in fuse_equal_placeholder_pass

Erik-Lundell:
Verified to catch same dupes as before in lstm and mv2.

Instead of comparing all placeholders to each other, compute a hash use as key in dictionary. Equal placeholder -> equal key. If an entry in the dictionary has multiple values, we have duplicates.

This is a ~O(N) algorithm compared to earlier  O(N^2). This can be seen by measuring the speedup for lstm vs. mv2

lstm: 120 placeholders (116 dupes) 0.4s -> 0.3s

mv2: 318 placeholders (98 dupes) ~15s -> 0.5s

### DIFF
--- a/backends/arm/_passes/fuse_equal_placeholders_pass.py
+++ b/backends/arm/_passes/fuse_equal_placeholders_pass.py
@@ -3,6 +3,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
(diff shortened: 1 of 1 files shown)


Erik-Lundell:
Test failures unrelated.

Sebastian-Larsson (approved):

PR #13273, opened 2025-08-11
Arm backend: Generate ETRecord from arm_aot_compiler

zingo:
An ETRecord file is generated if ETDump or BundleIO is used. This also adds tests for --bundleio and --etdump to make sure it works.


zingo:
Test fails are unrelated

backends/arm/test/test_arm_baremetal.sh
```diff
@@ -157,17 +156,23 @@ test_run_ethosu_fvp() { # End to End model tests using run.sh
 
     # TOSA quantized
     echo "${TEST_SUITE_NAME}: Test ethos-u target TOSA"
-    examples/arm/run.sh --et_build_root=arm_test/test_run --target=${TOSA_VERSION} --model_name=add
-    examples/arm/run.sh --et_build_root=arm_test/test_run --target=${TOSA_VERSION} --model_name=mul
+    examples/arm/run.sh --et_build_root=arm_test/test_run --target=TOSA-1.0+INT --model_name=add
```
digantdesai, line 159: even better if we make `TOSA-1.0+INT` as the default target in run.sh :)

digantdesai (approved):

backends/arm/test/test_arm_baremetal.sh
zingo: Interesting, I tried to keep this script very explicit for now to be used both for testing in CI and to make it easy to just copy&paste linse if you want to run/reproduce stuff, so this line would probably still always specify them. :) But lets see. I also like haveing less to write. Not the default is some Ethos-U target.

PR #13274, opened 2025-08-11
Arm Backend: Add support for expm1.default

agrima1304:
Decompose expm1 into other operators or use the Taylor series expansion when input values are close to 0.


Sebastian-Larsson (approved):
Unrelated CI errors. Approved

PR #13278, opened 2025-08-11
Arm backend: Add example linkerscripts for U55/U85

perheld:
Add linkerscripts for exector_runner targeting U55/U85 in examples/arm instead of using the linkerscripts from core_platform and then applying patches.

The patches are deleted since they are no longer needed.

Change-Id: I379cf1cd3d9464227fee0555e39847e5b61cc392


perheld:
Lets hold off on this, I need to investigate some more.

PR #13279, opened 2025-08-11
NXP backend: Improve cifarnet speed by removing the initial pading.

jirioc:
### Summary
NXP backend: Improve cifarnet speed by removing the initial pading.

### Test plan
Update to test_remove_io_quant_ops_pass__cifarnet() is part of the diff.


examples/nxp/experimental/cifar_net/cifar_net.py
```diff
@@ -66,10 +66,7 @@ def __init__(self):
         self.softmax = nn.Softmax(1)
 
     def forward(self, x):
-
-        # Neutron Backend does not yet have passses for automated padding if number of channels does not
-        # fit to Neutron constrains (#channels == #MAC units). So define the model explicitly tailored for Neutron-C-64.
-        x = F.pad(x, (2, 2, 2, 2, 0, 5))
+        x = F.pad(x, (2, 2, 2, 2))
```
JakeStevens, line 69: Why is there remaining padding?

jirioc: The remaining padding ensures that the output of convolution has the same size as the original x before padding. It has the same effect as using padding="same" in the convolutions.

jirioc: This type of padding is fused later into convolution.

JakeStevens (approved):

backends/nxp/tests/test_integration.py
```diff
@@ -43,8 +43,8 @@ def test_cifarnet():
 
     delegation_info = get_delegation_info(exec_prog.exported_program().graph_module)
     assert delegation_info.num_delegated_subgraphs == 1
-    assert delegation_info.num_non_delegated_nodes == 17
-    assert delegation_info.num_delegated_nodes == 42
+    assert delegation_info.num_non_delegated_nodes == 11
```
digantdesai, line 46: Nice!

examples/nxp/experimental/cifar_net/cifar_net.py
digantdesai: is this part of the original model definition?

JakeStevens:
Failing tests are specified as flaky and known broken on trunk, merging

examples/nxp/experimental/cifar_net/cifar_net.py
robert-kalmar: Yes. The previous padding to 8 channels was due to Neutron NPU constrain. In the meanwhile the Neutron converter got the capability to autopad, hence it is not necessary anymore. 

For the remaining padding, it might have been used the "same" padding in the Convolution https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv2d.html. 

Filed a ticket here: https://github.com/pytorch/executorch/issues/13470

PR #13282, opened 2025-08-11
Arm backend: Update VelaIO handling

### LINKED ISSUE #13022: Arm backend: Vela internal error
### 🐛 Describe the bug

Repro 
* PR - https://github.com/pytorch/executorch/pull/13021
* cmds - 
    * 
```
# TOSA Passes
pytest --pdb --verbose --color=yes --numprocesses=auto \
backends/arm/test/models/test_nn_modules.py::test_nn_Modules_BI\[SimpleWeightReuseModel\]
    
# U55 Fails (Making it XFAIL)
pytest --pdb --verbose --color=yes --numprocesses=auto \
backends/arm/test/models/test_nn_modules.py::test_nn_Modules_U55BI\[SimpleWeightReuseModel\]
```
* error - `RuntimeError: Non-passthrough operation could not run on NPU` from Vela.
* logs
[test.log](https://github.com/user-attachments/files/21519228/test.log)
[exported_graph.log](https://github.com/user-attachments/files/21519229/exported_graph.log)

    

### Versions

main

oscarandersson8218:
VelaIO is always 6D.

- Update AOT handling of metadata from Vela.
- Adds unittest to trigger 5D cases.
- Updates EthosUBackend to read IO as 6D arrays.

Fixes #13022

### DIFF
--- a/backends/arm/arm_vela.py
+++ b/backends/arm/arm_vela.py
@@ -25,6 +25,8 @@ except ImportError:
 # per-io structs to simplify runtime use.
 def vela_bin_pack_io(prefix, data):
     vela_input_shapes = data[prefix + "_shape"]
+    # Vela input/output shape is fixed to 6D
+    vela_io_shape_dims = 6
 
     ios = struct.pack("<i", len(vela_input_shapes))
     for i in range(len(vela_input_shapes)):
@@ -32,10 +34,10 @@ def vela_bin_pack_io(prefix, data):
         io_elem_size = data[prefix + "_elem_size"][i]
         io_offset = data[prefix + "_offset"][i]
         io_region = data[prefix + "_region"][i]
-        assert len(io_shape) <= 4
-        inp_pad = io_shape.tolist() + [0] * (4 - len(io_shape))
+        assert len(io_shape) == vela_io_shape_dims
+        inp_pad = io_shape.tolist()
         io_struct = struct.pack(
-            "<iiiiiii", *inp_pad, io_elem_size, io_offset, io_region
+            "<iiiiiiiii", *inp_pad, io_elem_size, io_offset, io_region
         )
         ios += io_struct
     return ios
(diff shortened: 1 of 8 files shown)


zingo:
New Vela version that we hope fix https://github.com/pytorch/executorch/issues/1302

zingo:
@digantdesai  new Vela version that we hope fix https://github.com/pytorch/executorch/issues/13022

digantdesai:
Yeah @apullin is trying out the new vela version 🙏

zingo:
As this is not merged yet we already have an even newver vela fixing another bug, depending if we see if thei has beem brought into meta or not for testing we might just update the Vela SHA in this PR. If you want us to keep this untouch @digantdesai just let us know but in that case you will get 2 Vela update PRs to juggle merge at the same time :)

zingo (approved):

digantdesai:
We haven't had a chance to update Vela internally. Perhaps this week - CC @apullin

If I were to guess we can just move the Vela to the newer of two SHAs, and it should be OK.

zingo:
Thanks, Yes, preferable latest Vela but note that Vela needs to be steped together with this PR, the  6D support is needed to be done on Executorch side to not break things.
Latest Vela fix more things then the big, so we need it anyway independent on the bugtesting.

apullin:
Just to update here - I did try out this commit and the corresponding commit it ethos-u-vela, and it did resolve the crash in the compiler for both the minimum working example and for our network that uses that pattern internally.

There are some other ongoing issues with lowering, where all the layers get rejected, but that may be due to a totally separate issue than the observed crash that kicked off this fix.

zingo:
@oscarandersson8218 this probably need a rebase again :( now after the vela install is reworked. Sorry about that.

oscarandersson8218:
CI failures are unrelated. @digantdesai @apullin can we merge this? It's blocking quite a few of our patches.

apullin:
I don't think anything on our end precludes a merge ... ? 

I evaluated with the specific commit from this PR, and will continue to use it with that pinned version. So merging should not affect us. When this gets rolled into a release tag (non-rc), we'll probably just move on to that version.

zingo:
Merged in https://github.com/pytorch/executorch/pull/14293

digantdesai:
Thanks @zingo

PR #13283, opened 2025-08-11
Adding VGF to documentation and revising various out of date parts

robell:
Revising documentation for a number of historical changes, and to add the recently introduced VGF backend. 


robell:
Fixes are now made, the lint-urls issue is because various files have moved and a seperate PR will introduce the other one. The docs build seems to fail if it happens multiple times, is there a way to delete and start the docs again?

will add skip-url-lint for the path issues.

PR #13284, opened 2025-08-11
Fix 8w8a qat qconfig setting activations

navsud:
Summary: 8-bit activation qconfig should not use reduce_range=True which limits the range to 0,127. This diff fixes that issue.

Differential Revision: D80007226

### DIFF
--- a/backends/qualcomm/quantizer/qconfig.py
+++ b/backends/qualcomm/quantizer/qconfig.py
@@ -396,7 +396,6 @@ def get_8a8w_qnn_qat_config(
         qscheme=(
             torch.per_tensor_symmetric if act_symmetric else torch.per_tensor_affine
         ),
-        reduce_range=True,
         observer=act_observer,
     )
     act_quantization_spec = QuantizationSpec(


cccclai:
>  8-bit activation qconfig should not use reduce_range=True which limits the range to 0,127

What range should it be instead?

navsud:
> > 8-bit activation qconfig should not use reduce_range=True which limits the range to 0,127
> 
> What range should it be instead?

Without `reduce_range=True`, it would use the default of 0,255 ranges, which is the correct ranges for uint8. With this change, I was able to recover the training loss during QAT back to that of the fp32 training.

cccclai (approved):
Thanks for the fix!

Novelfor:
I found the weight use reduce_range too.. 
``` Python
    weight_fake_quant_ctr = FusedMovingAvgObsFakeQuantize.with_args(
        dtype=torch.int8,
        quant_min=torch.iinfo(torch.int8).min + 1,
        quant_max=torch.iinfo(torch.int8).max,
        qscheme=torch.per_tensor_symmetric,
        reduce_range=True,
        observer=MovingAverageMinMaxObserver,
    )
```
I think if modify it to False will improve performance too...

cccclai:
> I found the weight use reduce_range too..
> 
> ```python
>     weight_fake_quant_ctr = FusedMovingAvgObsFakeQuantize.with_args(
>         dtype=torch.int8,
>         quant_min=torch.iinfo(torch.int8).min + 1,
>         quant_max=torch.iinfo(torch.int8).max,
>         qscheme=torch.per_tensor_symmetric,
>         reduce_range=True,
>         observer=MovingAverageMinMaxObserver,
>     )
> ```
> 
> I think if modify it to False will improve performance too...


cc: @haowhsu-quic @winskuo-quic @shewu-quic @DannyYuyang-quic

navsud:
@haowhsu-quic 
I see reduce_range=True at many places in this file. Do we really need to reduce the range by 1-bit for qualcomm backends? If not, we should remove them to get back the 1-bit range, which would improve the model accuracy.

PR #13285, opened 2025-08-11
Enable QAT for static llama definition

navsud:
Summary: The model needed small modifications to be able to run QAT on GPUs.

Differential Revision: D79841467


lucylq (approved):
LGTM on the executorch changes. Could you run lintrunner?

PR #1357, opened 2023-12-05
[RFC] Scope for Custom Operator extension

DenisVieriu97:
This document proposes a solution to extend the custom operator support so that the delegates can optimize the resource sharing and adaptively commit the work based on the kernel type.

----

Note that this document is meant to gather feedback and is not an official RFC process. Currently it is following the same style as https://github.com/pytorch/executorch/pull/1262.

A draft of the custom kernel explained in this RFC is available at https://github.com/pytorch/executorch/pull/1358.

PR #13296, opened 2025-08-11
Summary: Initial CMSS-NN Add Op

psiddh:
Test Plan:
1. examples/arm/run.sh - No regressions ==> Ok
2. examples/arm/run.sh  now runs 'qadd2' in quantize only mode ==> Ok
3. python -m unittest test_replace_quant_nodes.py  ==> Ok
4. python -m unittest test_quantize_op_fusion_pass.py  ==> Ok

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/backends/cortex_m/CMakeLists.txt
+++ b/backends/cortex_m/CMakeLists.txt
@@ -5,11 +5,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Kernel library for Cortex-M operators. Please keep this file formatted by
-# running:
-# ~~~
-# cmake-format -i CMakeLists.txt
-# ~~~
 cmake_minimum_required(VERSION 3.19)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
@@ -24,28 +19,64 @@ endif()
 
 include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
 include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
+include(FetchContent)
+
+# CMSIS-NN version to download
+set(CMSIS_NN_VERSION
+    "v4.1.0"
+    CACHE STRING "CMSIS-NN version to download"
+)
+
+# Declare CMSIS-NN as a FetchContent project
+FetchContent_Declare(
+  cmsis_nn
+  GIT_REPOSITORY https://github.com/ARM-software/CMSIS-NN.git
+  GIT_TAG ${CMSIS_NN_VERSION}
+)
+
+# Download and make CMSIS-NN available
+FetchContent_MakeAvailable(cmsis_nn)
+
+# Print paths for debugging
+message(STATUS "CMSIS-NN source dir: ${cmsis_nn_SOURCE_DIR}")
+message(STATUS "CMSIS-NN binary dir: ${cmsis_nn_BINARY_DIR}")
 
 # Cortex-M ops kernel sources
 set(_cortex_m_kernels__srcs
     ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantize_per_tensor.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_dequantize_per_tensor.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantized_add.cpp
 )
 
-# Generate C++ bindings to register kernels into Executorch (for runtime). Here
-# select all ops in operators.yaml
+# Generate C++ bindings to register kernels into Executorch (for runtime)
 set(_yaml_file ${CMAKE_CURRENT_LIST_DIR}/ops/operators.yaml)
 gen_selected_ops(LIB_NAME "cortex_m_ops_lib" OPS_SCHEMA_YAML "${_yaml_file}")
 
-# Generate bindings for the kernels
 generate_bindings_for_kernels(
   LIB_NAME "cortex_m_ops_lib" CUSTOM_OPS_YAML "${_yaml_file}"
 )
 message("Generated files ${gen_command_sources}")
 
-# Build a library for _cortex_m_kernels_srcs
+# Build a library for cortex_m_kernels
 add_library(cortex_m_kernels ${_cortex_m_kernels__srcs})
-target_link_libraries(cortex_m_kernels PRIVATE executorch)
 target_compile_options(cortex_m_kernels PUBLIC ${_common_compile_options})
+
+# Include directories for cortex_m_kernels
+target_include_directories(
+  cortex_m_kernels
+  PRIVATE ${EXECUTORCH_ROOT}/..
+          ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10
+          ${cmsis_nn_SOURCE_DIR}/Include
+)
+
+# Link directly to the CMSIS-NN static library file
+target_link_libraries(
+  cortex_m_kernels PUBLIC ${cmsis_nn_BINARY_DIR}/libcmsis-nn.a executorch
+)
+
+# Add dependency to ensure CMSIS-NN builds before we try to link. Use the actual
+# CMSIS-NN target name (usually 'cmsis-nn')
+add_dependencies(cortex_m_kernels cmsis-nn)
 
 # cortex_m_ops_lib: Register Cortex-M ops kernels into Executorch runtime
 gen_operators_lib(
--- a/backends/cortex_m/ops/cortex_m_ops_common.h
+++ b/backends/cortex_m/ops/cortex_m_ops_common.h
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <executorch/kernels/portable/cpu/util/broadcast_util.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+#include <executorch/kernels/portable/cpu/scalar_utils.h>
+#include <executorch/kernels/portable/cpu/util/elementwise_util.h>
+#include <executorch/kernels/portable/cpu/util/kernel_ops_util.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+#include <executorch/runtime/platform/assert.h>
+
+// Include CMSIS-NN headers with C linkage
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+using Tensor = torch::executor::Tensor;
+using ScalarType = executorch::aten::ScalarType;
+using Scalar = torch::executor::Scalar;
+using Error = executorch::runtime::Error;
+
+// Basic tensor type / layout validation and dimension order checking
+inline void validate_cmsis_nn_tensor_requirements(
+    const Tensor& input1,
+    const Tensor& input2,
+    Tensor& output,
+    ScalarType expected_dtype = ScalarType::Char,
+    bool require_channels_last = false) {
+  // Basic dtype validation
+  ET_CHECK_MSG(
+      input1.scalar_type() == expected_dtype,
+      "Input1 dtype must be %hhd",
+      expected_dtype);
+  ET_CHECK_MSG(
+      input2.scalar_type() == expected_dtype,
+      "Input2 dtype must be %hhd",
+      expected_dtype);
+  ET_CHECK_MSG(
+      output.scalar_type() == expected_dtype,
+      "Output dtype must be %hhd",
+      expected_dtype);
+
+  // Dim order consistency
+  ET_CHECK_MSG(
+      executorch::runtime::tensors_have_same_dim_order(input1, input2, output),
+      "Tensors must have same dimension order");
+
+  // TBD: Validate memory alignment (CMSIS-NN requirement)
+}
+
+inline void validate_single_quant_params(
+    const Scalar& zero_point,
+    const Scalar& multiplier,
+    const Scalar& shift,
+    const char* param_name) {
+  int64_t zp_val = zero_point.to<int64_t>();
+  int64_t mult_val = multiplier.to<int64_t>();
+  int64_t shift_val = shift.to<int64_t>();
+
+  ET_CHECK_MSG(
+      zp_val >= std::numeric_limits<int8_t>::min() &&
+          zp_val <= std::numeric_limits<int8_t>::max(),
+      "%s zero point must be in int8 range [Value: %d]",
+      param_name,
+      zp_val);
+
+  ET_CHECK_MSG(
+      mult_val >= std::numeric_limits<int32_t>::min() &&
+          mult_val <= std::numeric_limits<int32_t>::max(),
+      "%s multiplier must be in int32 range [Value: %d]",
+      param_name,
+      mult_val);
+
+  ET_CHECK_MSG(
+      shift_val >= -31 && shift_val <= 31,
+      "%s shift must be in range [-31, 31] [Value: %d]",
+      param_name,
+      shift_val);
+}
+
+/**
+ * Validate quantization parameters for inputs and output.
+ *
+ * Checks that zero points fit in int8 range, multipliers fit in int32 range,
+ * and shifts are within a valid bit-shift range (0-31).
+ *
+ * Ensures parameters comply with Ahead-Of-Time (AOT) quantization requirements
+ * and CMSIS-NN kernel expectations.
+ *
+ * Raises errors via ET_KERNEL_CHECK if any check fails.
+ */
+inline void validate_quantization_params(
+    const Scalar& zero_point1,
+    const Scalar& multiplier1,
+    const Scalar& shift1,
+    const Scalar& zero_point2,
+    const Scalar& multiplier2,
+    const Scalar& shift2,
+    const Scalar& output_zero_point,
+    const Scalar& output_multiplier,
+    const Scalar& output_shift,
+    Tensor& output) {
+  validate_single_quant_params(
+      zero_point1, multiplier1, shift1, "Single quant Input1");
+  validate_single_quant_params(
+      zero_point2, multiplier2, shift2, "Single quant Input2");
+  validate_single_quant_params(
+      output_zero_point,
+      output_multiplier,
+      output_shift,
+      "Single quant Output");
+}
+
+inline Error resize_to_broadcast_target_size(
+    const Tensor& input1,
+    const Tensor& input2,
+    Tensor& output) {
+  static constexpr int kTensorDimensionLimit = 5;
+  Tensor::SizesType expected_output_size[kTensorDimensionLimit];
+  size_t expected_output_dim = 0;
+  auto err = torch::executor::get_broadcast_target_size(
+      input1,
+      input2,
+      expected_output_size,
+      kTensorDimensionLimit,
+      &expected_output_dim);
+
+  if (err != Error::Ok)
+    return err;
+
+  return executorch::runtime::resize_tensor(
+      output, {expected_output_size, expected_output_dim});
+}
--- a/backends/cortex_m/ops/op_quantized_add.cpp
+++ b/backends/cortex_m/ops/op_quantized_add.cpp
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "cortex_m_ops_common.h"
+
+namespace cortex_m {
+namespace native {
+using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
+
+Tensor& quantized_add_out(
+    KernelRuntimeContext& context,
+    const Tensor& input1_int8,
+    const Scalar& input1_zero_point,
+    const Scalar& input1_multiplier,
+    const Scalar& input1_shift,
+    const Tensor& input2_int8,
+    const Scalar& input2_zero_point,
+    const Scalar& input2_multiplier,
+    const Scalar& input2_shift,
+    const Scalar& output_zero_point,
+    const Scalar& output_multiplier,
+    const Scalar& output_shift,
+    Tensor& out) {
+  // Validate tensor types and dim order
+  validate_cmsis_nn_tensor_requirements(input1_int8, input2_int8, out);
+
+  // Validate quantization parameters
+  validate_quantization_params(
+      input1_zero_point,
+      input1_multiplier,
+      input1_shift,
+      input2_zero_point,
+      input2_multiplier,
+      input2_shift,
+      output_zero_point,
+      output_multiplier,
+      output_shift,
+      out);
+
+  // Broadcast if needed
+  auto result = resize_to_broadcast_target_size(input1_int8, input2_int8, out);
+  ET_CHECK_MSG(
+      (result == Error::Ok),
+      "Failed to resize output tensor. Status: [%d]",
+      result);
+
+  ET_LOG(
+      Info,
+      "quantized_add_out: input1_int8.sizes() = %zu",
+      input1_int8.sizes().size());
+
+  // FIX: Use template types that ExecutorTorch definitely provides
+  // Use to<int64_t>() and to<double>() which are commonly instantiated
+  int32_t zp1 = static_cast<int32_t>(input1_zero_point.to<int64_t>());
+  int32_t input1_mult = static_cast<int32_t>(input1_multiplier.to<int64_t>());
+  int input1_shift_val = static_cast<int>(input1_shift.to<int64_t>());
+
+  int32_t zp2 = static_cast<int32_t>(input2_zero_point.to<int64_t>());
+  int32_t input2_mult = static_cast<int32_t>(input2_multiplier.to<int64_t>());
+  int input2_shift_val = static_cast<int>(input2_shift.to<int64_t>());
+
+  int32_t out_zp = static_cast<int32_t>(output_zero_point.to<int64_t>());
+  int32_t output_mult = static_cast<int32_t>(output_multiplier.to<int64_t>());
+  int output_shift_val = static_cast<int>(output_shift.to<int64_t>());
+
+  // Left shift to maximize precision (tune as needed)
+  const int32_t left_shift = 20;
+  const int32_t activation_min = std::numeric_limits<int8_t>::min();
+  const int32_t activation_max = std::numeric_limits<int8_t>::max();
+
+  ET_LOG(
+      Info,
+      "Using AoT-computed parameters: input1[mult=%d, shift=%d], input2[mult=%d, shift=%d], output[mult=%d, shift=%d]",
+      input1_mult,
+      input1_shift_val,
+      input2_mult,
+      input2_shift_val,
+      output_mult,
+      output_shift_val);
+
+  // Call CMSIS-NN kernel with precomputed parameters
+  arm_cmsis_nn_status status = arm_elementwise_add_s8(
+      input1_int8.const_data_ptr<int8_t>(),
+      input2_int8.const_data_ptr<int8_t>(),
+      static_cast<int32_t>(zp1),
+      input1_mult,
+      input1_shift_val,
+      static_cast<int32_t>(zp2),
+      input2_mult,
+      input2_shift_val,
+      left_shift,
+      out.mutable_data_ptr<int8_t>(),
+      static_cast<int32_t>(out_zp),
+      output_mult,
+      output_shift_val,
+      static_cast<int32_t>(out.numel()),
+      activation_min,
+      activation_max);
+
+  if (status != ARM_CMSIS_NN_SUCCESS) {
+    ET_LOG(
+        Error,
+        "quantized_add_out: arm_elementwise_add_s8 failed with status [%d]",
+        status);
+
+    context.fail(Error::Internal); // Fail the execution context
+    return out;
+  }
+  ET_LOG(
+      Info,
+      "quantized_add_out: Successfully completed with AoT-computed parameters!");
+
+  return out;
+}
+
+// Stub Implementation: Non-out variant for compatibility (functional variant)
+// EXIR/ExecuTorch runs an out-variant pass that converts
+// .default operations to .out variants before memory planning.
+// In the pass we are calling quantized_add's default variant
+// but ExecuTorch's kernel dispatch mechanism will end up calling the out
+// variant. This stub is to make sure that compiler doesn't complain.
+Tensor quantized_add(
+    KernelRuntimeContext& context,
+    const Tensor& input1_int8,
+    const Scalar& input1_zero_point,
+    const Scalar& input1_multiplier,
+    const Scalar& input1_shift,
+    const Tensor& input2_int8,
+    const Scalar& input2_zero_point,
+    const Scalar& input2_multiplier,
+    const Scalar& input2_shift,
+    const Scalar& output_zero_point,
+    const Scalar& output_multiplier,
+    const Scalar& output_shift) {
+  ET_LOG(Info, "quantized_add: input1_int8.sizes() = %zu", input1_int8.sizes());
+
+  // Crash on Debug builds if invoked
+  assert(False);
+  // This is to make sure compiler doesn't complain.
+  return const_cast<Tensor&>(input1_int8);
+}
+
+} // namespace native
+} // namespace cortex_m
--- a/backends/cortex_m/ops/operators.py
+++ b/backends/cortex_m/ops/operators.py
@@ -5,10 +5,15 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
-from executorch.exir.dialects._ops import (
-    ops as exir_ops,
-)  # To provide the implementation of the operators
+from executorch.backends.cortex_m.passes.passes_utils import (
+    dequantize_per_tensor_cmsis,
+    quantize_per_tensor_cmsis,
+)
+from executorch.exir.dialects._ops import ops as exir_ops
+
+# To provide the implementation of the operators
 from torch.library import impl, Library, register_fake
+
 
 # New operator library with a custom namespace to allow fusion etc.
 lib = Library("cortex_m", "DEF")
@@ -96,3 +101,125 @@ def dequantize_per_tensor_impl(
     return exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default(
         input, scale, zero_point, quant_min, quant_max, dtype
     )
+
+
+# Define the operator schema with multipliers and shifts (11 args)
+lib.define(
+    "quantized_add("
+    "Tensor self, Scalar self_zero_point, Scalar self_multiplier, Scalar self_shift, "
+    "Tensor other, Scalar other_zero_point, Scalar other_multiplier, Scalar other_shift, "
+    "Scalar output_zero_point, Scalar output_multiplier, Scalar output_shift) -> Tensor"
+)
+
+
+@register_fake("cortex_m::quantized_add")
+def quantized_add_meta(
+    self: torch.Tensor,
+    self_zero_point: int,
+    self_multiplier: int,
+    self_shift: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    broadcasted_shape = torch.broadcast_shapes(self.shape, other.shape)
+    return torch.empty(broadcasted_shape, dtype=torch.int8, device=self.device)
+
+
+@impl(lib, "quantized_add", "CompositeExplicitAutograd")
+def quantized_add_impl(
+    self: torch.Tensor,
+    self_zero_point: int,
+    self_multiplier: int,
+    self_shift: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    self_fp = dequantize_per_tensor_cmsis(
+        self, self_zero_point, self_multiplier, self_shift
+    )
+    other_fp = dequantize_per_tensor_cmsis(
+        other, other_zero_point, other_multiplier, other_shift
+    )
+    result_fp = self_fp + other_fp
+    result_quantized = quantize_per_tensor_cmsis(
+        result_fp, output_zero_point, output_multiplier, output_shift
+    )
+    return result_quantized
+
+
+# Define the operator schema with multipliers and shifts (11 args + out tensor)
+lib.define(
+    "quantized_add.out("
+    "Tensor self, Scalar self_zero_point, Scalar self_multiplier, Scalar self_shift, "
+    "Tensor other, Scalar other_zero_point, Scalar other_multiplier, Scalar other_shift, "
+    "Scalar output_zero_point, Scalar output_multiplier, Scalar output_shift, "
+    "*, Tensor(a!) out) -> Tensor(a!)"
+)
+
+
+# Fake meta function for shape and dtype inference during compilation
+@register_fake("cortex_m::quantized_add.out")
+def quantized_add_out_meta(
+    self: torch.Tensor,
+    self_zero_point: int,
+    self_multiplier: int,
+    self_shift: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+    out: torch.Tensor,
+) -> torch.Tensor:
+    # Validate against correct broadcasted shape
+    expected_shape = torch.broadcast_shapes(self.shape, other.shape)
+    assert (
+        out.shape == expected_shape
+    ), f"Output shape {out.shape} must match broadcasted shape {expected_shape}"
+    return out
+
+
+# Actual implementation delegating to backend or custom kernel
+@impl(lib, "quantized_add.out", "CompositeExplicitAutograd")
+def quantized_add_out_impl(
+    self: torch.Tensor,
+    self_zero_point: int,
+    self_multiplier: int,
+    self_shift: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+    *,
+    out: torch.Tensor,
+) -> torch.Tensor:
+    self_fp = dequantize_per_tensor_cmsis(
+        self, self_zero_point, self_multiplier, self_shift
+    )
+    other_fp = dequantize_per_tensor_cmsis(
+        other, other_zero_point, other_multiplier, other_shift
+    )
+    result_fp = self_fp + other_fp
+    result_quantized = quantize_per_tensor_cmsis(
+        result_fp, output_zero_point, output_multiplier, output_shift
+    )
+
+    # Write into the provided output tensor
+    out.copy_(result_quantized)
+
+    return out
--- a/backends/cortex_m/ops/operators.yaml
+++ b/backends/cortex_m/ops/operators.yaml
@@ -15,3 +15,15 @@
   kernels:
     - arg_meta: null
       kernel_name: cortex_m::dequantize_per_tensor_out
+
+- func: cortex_m::quantized_add(Tensor self, Scalar self_zero_point, Scalar self_multiplier, Scalar self_shift, Tensor other, Scalar other_zero_point, Scalar other_multiplier, Scalar other_shift, Scalar output_zero_point, Scalar output_multiplier, Scalar output_shift) -> Tensor
+  variants: function
+  kernels:
+    - arg_meta: null
+      kernel_name: cortex_m::quantized_add
+
+- func: cortex_m::quantized_add.out(Tensor self, Scalar self_zero_point, Scalar self_multiplier, Scalar self_shift, Tensor other, Scalar other_zero_point, Scalar other_multiplier, Scalar other_shift, Scalar output_zero_point, Scalar output_multiplier, Scalar output_shift, *, Tensor(a!) out) -> Tensor(a!)
+  variants: function
+  kernels:
+    - arg_meta: null
+      kernel_name: cortex_m::quantized_add_out
--- a/backends/cortex_m/passes/passes_utils.py
+++ b/backends/cortex_m/passes/passes_utils.py
@@ -1,0 +1,94 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+
+import torch
+
+
+def dequantize_per_tensor_cmsis(
+    qtensor: torch.Tensor, zero_point: int, multiplier: int, shift: int
+) -> torch.Tensor:
+    """
+    Simulate CMSIS-NN fixed-point dequantization:
+    result = (qtensor - zero_point) * multiplier * 2^shift / 2^31
+    """
+    scale = multiplier * (2**shift) / (1 << 31)
+    return (qtensor.float() - zero_point) * scale
+
+
+def quantize_per_tensor_cmsis(
+    tensor: torch.Tensor,
+    zero_point: int,
+    multiplier: int,
+    shift: int,
+    qmin=-128,
+    qmax=127,
+) -> torch.Tensor:
+    """
+    Simulate CMSIS-NN fixed-point quantization:
+    result = round(tensor / scale) + zero_point, clamped to [qmin, qmax]
+    """
+    scale = multiplier * (2**shift) / (1 << 31)
+    quantized = torch.round(tensor / scale) + zero_point
+    return quantized.clamp(qmin, qmax).to(torch.int8)
+
+
+def extract_scalar_value(node_arg) -> float:
+    """
+    Extract scalar value from various PyTorch scalar representations.
+    """
+    if hasattr(node_arg, "op") and node_arg.op == "get_attr":
+        # Handle case where scalar is a graph attribute
+        return float(node_arg.target)
+    elif isinstance(node_arg, (int, float)):
+        return float(node_arg)
+    elif hasattr(node_arg, "item"):
+        return float(node_arg.item())
+    else:
+        # Try to extract from meta if available
+        if hasattr(node_arg, "meta") and "val" in node_arg.meta:
+            val = node_arg.meta["val"]
+            if hasattr(val, "item"):
+                return float(val.item())
+            return float(val)
+        raise ValueError(
+            f"Cannot extract scalar value from {type(node_arg)}: {node_arg}"
+        )
+
+
+def is_qualified_int8_node(args) -> bool:
+    try:
+        if len(args) < 6:
+            return False
+        qmin = int(args[3])
+        qmax = int(args[4])
+        dtype_str = str(args[5])
+        is_int8_range = (
+            qmin >= torch.iinfo(torch.int8).min and qmax <= torch.iinfo(torch.int8).max
+        )
+        is_int8_dtype = "int8" in dtype_str.lower()
+        return is_int8_range and is_int8_dtype
+    except (IndexError, ValueError, TypeError):
+        return False
+
+
+def quantize_multiplier_aot(scale: float) -> tuple[int, int]:
+    if scale == 0.0:
+        return 0, 0
+    mantissa, exponent = math.frexp(scale)
+    shift = -exponent
+    q_fixed = int(round(mantissa * (1 << 31)))
+    if q_fixed == (1 << 31):
+        q_fixed //= 2
+        shift -= 1
+    multiplier = max(-2147483648, min(2147483647, q_fixed))
+    return multiplier, shift
+
+
+def cleanup_erased_nodes(graph_module: torch.fx.GraphModule):
+    # Placeholder for any additional cleanup if needed
+    pass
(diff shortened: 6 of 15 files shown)


backends/cortex_m/CMakeLists.txt
```diff
@@ -23,10 +23,16 @@ endif()
 include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
 include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
 
+set(CMSIS_NN_ROOT /home/sidart/working/CMSIS-NN)
```
psiddh: TBD: Wrap this under USE_CMSIS flag and fix the path

backends/cortex_m/ops/op_add.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  // Check input data types
+  ScalarType input1_dtype = input1.scalar_type();
+  ScalarType input2_dtype = input2.scalar_type();
+  ScalarType out_dtype = out.scalar_type();
+
+  ET_LOG(Info, "Input1 dtype: %d, Input2 dtype: %d, Output dtype: %d",
+         static_cast<int>(input1_dtype), static_cast<int>(input2_dtype), static_cast<int>(out_dtype));
+
+  // Stub for now
```
digantdesai: assert just to be safe

backends/cortex_m/ops/op_add_tensor.cpp
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+extern "C" {
+#include "Include/arm_nnfunctions.h"
+}
```
digantdesai: move this to a common include for `cortex_m/ops`? So we can do guards and check in one place.

backends/cortex_m/ops/op_add_tensor.cpp
```diff
@@ -0,0 +1,45 @@
...
+using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
+
+Tensor& add_Tensor(
+    KernelRuntimeContext& ctx,
+    const Tensor& self,
+    const Tensor& other,
+    Tensor& out) {
+  ET_LOG(Info, "add_Tensor kernel called");
+
+  // Ensure input is char type
```
digantdesai: refactor these as a check_args function in the same file.

backends/cortex_m/ops/op_add_tensor.cpp
```diff
@@ -0,0 +1,45 @@
...
+      self.scalar_type() == ScalarType::Char,
+      "self.scalar_type() %" PRId8 " is not char type",
+      static_cast<int8_t>(self.scalar_type()));
+
+  ET_CHECK_MSG(
+      other.scalar_type() == ScalarType::Char,
+      "other.scalar_type() %" PRId8 " is not char type",
+      static_cast<int8_t>(other.scalar_type()));
+
+  // Stub for now
```
digantdesai: assert

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,132 @@
...
+    ET_CHECK_MSG(out.scalar_type() == ScalarType::Char, "Output must be int8");
+
+    // Check tensor shapes are compatible
+    ET_CHECK_SAME_SHAPE_AND_DTYPE2(input1_int8, input2_int8);
+
+    ET_LOG(Info, "quantized_add_out: input1_int8.sizes() = %zu", input1_int8.sizes().size());
+
+    // 🔧 FIX: Use template types that ExecutorTorch definitely provides
+    // Use to<int64_t>() and to<double>() which are commonly instantiated
+    int64_t zp1 = input1_zero_point.to<int64_t>();
```
digantdesai: why not to int32_t?

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,132 @@
...
+        static_cast<int32_t>(out.numel()),
+        activation_min,
+        activation_max
+    );
+
+    if (status != ARM_CMSIS_NN_SUCCESS) {
+        ET_LOG(Error, "quantized_add_out: arm_elementwise_add_s8 failed with status [%d]", status);
+        std::memset(out.mutable_data_ptr<int8_t>(), 0, out.nbytes());
+    } else {
+        ET_LOG(Info, "quantized_add_out: Successfully completed with AoT-computed parameters! 🎯");
```
digantdesai: Not sure about UTF8 in the logs :p

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,132 @@
...
+    const Scalar& input1_shift,
+    const Tensor& input2_int8,
+    const Scalar& input2_zero_point,
+    const Scalar& input2_multiplier,
+    const Scalar& input2_shift,
+    const Scalar& output_zero_point,
+    const Scalar& output_multiplier,
+    const Scalar& output_shift) {
+    ET_LOG(Info, "quantized_add: input1_int8.sizes() = %zu", input1_int8.sizes().size());
+    return const_cast<Tensor&>(input1_int8); // to make compiler happy
```
digantdesai: Do we need this?
Also assert?

backends/cortex_m/ops/op_add.cpp
```diff
@@ -0,0 +1,44 @@
+/*
```
digantdesai: why do we need these? Just curious.

backends/cortex_m/ops/operators.py
```diff
@@ -13,6 +13,55 @@
...
+    if out is None:
+        if self.is_quantized:
+            qparams = self.q_per_tensor()
+            out = torch._empty_affine_quantized(
+                self.size(), self.dtype, qparams.scale, qparams.zero_point
+            )
+        else:
+            out = torch.empty_like(self)
+    # Call the backend kernel that writes into 'out'
+    return exir_ops.edge.cortex_m.add.out(self, other, alpha, out)
```
digantdesai: why can't we just use regular pytorch to implement this? The c++ impl assumed to only work on Cortex-M, while this is expected to work "everywhere".

backends/cortex_m/ops/operators.py
```diff
@@ -96,3 +142,134 @@ def dequantize_per_tensor_impl(
...
+    self_shift: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    return torch.empty_like(self, dtype=torch.int8)
```
digantdesai: what about broadcasting?
See - 
* https://github.com/pytorch/executorch/blob/ec56c95baffc4c00ba47181749b7747e81a51812/kernels/portable/cpu/op_add.cpp#L41-L45
* https://docs.pytorch.org/docs/stable/notes/broadcasting.html

backends/cortex_m/ops/operators.py
```diff
@@ -96,3 +142,134 @@ def dequantize_per_tensor_impl(
...
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    # Implement your actual quantized add logic here
+    # For now, convert back to float, add, and quantize (as placeholder)
+
+    # Dequantize inputs using multiplier/shift
+    self_fp = (self.float() - self_zero_point) * (self_multiplier / (1 << (31 - self_shift)))
```
digantdesai: Nit: make small `inline`able utils for q/dq, as we will need this for other ops too.

backends/cortex_m/ops/operators.py
```diff
@@ -96,3 +142,134 @@ def dequantize_per_tensor_impl(
...
+    other_zero_point: int,
+    other_multiplier: int,
+    other_shift: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+    out: torch.Tensor,
+) -> torch.Tensor:
+    # Validate shape compatibility if needed
+    assert out.shape == self.shape, "Output shape must match input shape"
```
digantdesai: why is this a requirement?

backends/cortex_m/ops/operators.yaml
```diff
@@ -15,3 +15,26 @@
   kernels:
     - arg_meta: null
       kernel_name: cortex_m::dequantize_per_tensor_out
+- func: aten::add.Tensor(Tensor input1, Tensor input2, *, Tensor(a!) out) -> Tensor(a!)
```
digantdesai: Drop non-quantized variants?

backends/cortex_m/ops/operators.yaml
```diff
@@ -15,3 +15,26 @@
...
+    - arg_meta: null
+      kernel_name: cortex_m::add_Tensor
+
+- func: aten::add.out(Tensor self, Tensor other, Scalar alpha=1, *, Tensor(a!) out) -> Tensor(a!)
+  variants: function
+  kernels:
+    - arg_meta: null
+      kernel_name: cortex_m::add_out
+
+- func: cortex_m::quantized_add(Tensor self, Scalar self_zero_point, Scalar self_multiplier, Scalar self_shift, Tensor other, Scalar other_zero_point, Scalar other_multiplier, Scalar other_shift, Scalar output_zero_point, Scalar output_multiplier, Scalar output_shift) -> Tensor
```
digantdesai, line 19: can we reach this?

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,250 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from torch.fx.passes.infra.pass_manager import PassResult
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.dialects.edge._ops import EdgeOpOverload
+from executorch.exir.pass_base import ExportPass
+import torch
+import math
+
+class QuantizedAddFusionPass(ExportPass):
```
digantdesai: OK for now, but we can make this more generic in the future to replace other quantized ops as well.

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,250 @@
...
+                    print(f"🔍 Found quantized_decomposed.dequantize node")
+                    if self._is_qualified_int8_node(node.args):
+                        print(f"✅ Replacing dequantize node")
+                        node.target = exir_ops.edge.cortex_m.dequantize_per_tensor.default
+                        replacement_count += 1
+
+                elif node.target == exir_ops.edge.aten.add.Tensor:
+                    print(f"✅ Replacing add node")
+                    node.target = exir_ops.edge.cortex_m.add.Tensor
+                    replacement_count += 1
```
digantdesai, line 138: remove?

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,250 @@
...
+
+        print(f"🔧 Operator replacement: Replaced {replacement_count} nodes")
```
digantdesai, line 140: can we simplify this by using a `subgreph_rewriter`? I am thinking of this as a common pattern for most of these ops.
see - https://docs.pytorch.org/executorch/stable/compiler-custom-compiler-passes.html#level-2

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,250 @@
...
+
+            dequant_node1, dequant_node2 = add_node.args[:2]
+
+            if not (hasattr(dequant_node1, 'op') and dequant_node1.op == 'call_function'
+                    and dequant_node1.target == exir_ops.edge.cortex_m.dequantize_per_tensor.default
+                    and hasattr(dequant_node2, 'op') and dequant_node2.op == 'call_function'
+                    and dequant_node2.target == exir_ops.edge.cortex_m.dequantize_per_tensor.default):
+                continue
+
+            print("✅ Found complete cortex_m Q/DQ + add pattern!")
```
digantdesai: can we simplify this by using a `subgreph_rewriter`? I am thinking of this as a common pattern for most of these ops.
see - https://docs.pytorch.org/executorch/stable/compiler-custom-compiler-passes.html#level-2

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,250 @@
...
+                print(f"🧮 AoT computed parameters:")
+                print(f"   Input1: mult={input1_mult}, shift={input1_shift}")
+                print(f"   Input2: mult={input2_mult}, shift={input2_shift}")
+                print(f"   Output: mult={output_mult}, shift={output_shift}")
+
+                # Create the fused node with precomputed parameters using existing quantized_add target
+                with graph_module.graph.inserting_after(quantize_node):
+                    fused = graph_module.graph.create_node(
+                        'call_function',
+                        target=exir_ops.edge.cortex_m.quantized_add.default,  # 👈 Keep existing target
```
digantdesai: .Tensor to .default?

backends/cortex_m/test/test_quantize_add_fusion_pass.py
```diff
@@ -0,0 +1,427 @@
...
+from torchao.quantization.pt2e.quantizer.quantizer import Q_ANNOTATION_KEY
+
+
+@dataclass(eq=True, frozen=True)
+class QuantizationConfig:
+    input_activation: Optional[QuantizationSpec]
+    output_activation: Optional[QuantizationSpec]
+
+
+class AddQuantizer(Quantizer):
```
digantdesai: Do we want to use ArmQuantizer? Or share one with quant_nodes test?

backends/cortex_m/test/test_quantize_add_fusion_pass.py
```diff
@@ -0,0 +1,427 @@
...
+            exported,
+            compile_config=executorch.exir.EdgeCompileConfig(_check_ir_validity=False),
+        )
+        return edge_program
+
+    def _apply_passes(self, edge_program):
+        """Apply both ReplaceQuantNodesPass and QuantizedAddFusionPass"""
+        # Step 1: Replace quant nodes
+        replace_pass = ReplaceQuantNodesPass()
+        intermediate_program = edge_program.transform([replace_pass])
```
digantdesai: why do we need this? Ideally these passes should be independent of each other.

backends/cortex_m/test/test_quantize_add_fusion_pass.py
```diff
@@ -0,0 +1,427 @@
...
+        self.assertIn("cortex_m::quantized_add", operator_names)
+        self.assertIn("cortex_m::quantize_per_tensor", operator_names)
+        self.assertIn("cortex_m::dequantize_per_tensor", operator_names)
+        # quantize_per_tensor --> dequantize_per_tensor --> add --> quantize_per_tensor --> dequantize_per_tensor
+        # (input quant)          (dequant)               (fp32 add)       (re-quant)          (dequant)
+        #                ↓
+        # Fusion Pass detects pattern:
+        # dequantize_per_tensor --> quantized_add (Fused node) --> quantize_per_tensor
+
+
```
digantdesai: what about two inputs with different shapes to a single add?

backends/cortex_m/CMakeLists.txt
```diff
@@ -24,11 +24,36 @@ endif()
...
 include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
+include(ExternalProject)
+
+# Download and build CMSIS-NN from GitHub
+set(CMSIS_NN_VERSION "v4.1.0" CACHE STRING "CMSIS-NN version to download")
+set(CMSIS_NN_ROOT ${CMAKE_CURRENT_BINARY_DIR}/cmsis-nn)
+set(CMSIS_NN_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/cmsis-nn-build)
+set(CMSIS_NN_LIB_PATH ${CMSIS_NN_BINARY_DIR}/libcmsis-nn.a)
+
+ExternalProject_Add(
```
digantdesai: why not use `FetchContent`?

backends/cortex_m/CMakeLists.txt
```diff
@@ -24,11 +24,36 @@ endif()
...
+  cmsis_nn_external
+  GIT_REPOSITORY https://github.com/ARM-software/CMSIS-NN.git
+  GIT_TAG ${CMSIS_NN_VERSION}
+  SOURCE_DIR ${CMSIS_NN_ROOT}
+  BINARY_DIR ${CMSIS_NN_BINARY_DIR}
+  CMAKE_ARGS
+    -DCMAKE_TOOLCHAIN_FILE=${EXECUTORCH_ROOT}/examples/arm/ethos-u-setup/arm-none-eabi-gcc.cmake
+    -DTARGET_CPU=cortex-m55
```
digantdesai: why do we need to hard code this?

backends/cortex_m/CMakeLists.txt
```diff
@@ -44,9 +69,23 @@ message("Generated files ${gen_command_sources}")
...
 
+# Add dependency on CMSIS-NN external project
+add_dependencies(cortex_m_kernels cmsis_nn_external)
+
+# Set include directories - Include is directly in CMSIS-NN root
+target_include_directories(cortex_m_kernels
+  PRIVATE
+    ${EXECUTORCH_ROOT}/..
+    ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10
+    $<BUILD_INTERFACE:${CMSIS_NN_ROOT}/Include>
```
digantdesai: do we need `"include"` after this vs. `<include>`?

examples/arm/run.sh
```diff
@@ -186,12 +186,14 @@ devtools_flag=""
 bundleio_flag=""
 et_dump_flag=""
 if [ "$build_with_etdump" = true ] ; then
+    devtools_flag="--devtools --etdump"
     et_dump_flag="--etdump"
 fi
 
 if [ "$bundleio" = true ] ; then
-    devtools_flag="--devtools"
+    devtools_flag="--devtools --etdump"
```
digantdesai: accidental?

examples/arm/run.sh
```diff
@@ -204,17 +206,19 @@ if [[ -z "$model_name" ]]; then
...
     )
     model_compiler_flags=(
         ""                      # 0 softmax
         "--delegate"            # 1 add
         "--delegate"            # 2 add3
         "--delegate --quantize" # 3 qadd
         "--delegate --quantize" # 4 qadd2
-        "--delegate --quantize" # 5 qops
+        "--quantize"            # 5 qadd2
         "--delegate --quantize" # 6 mv2
```
digantdesai: s/mv2/qops?

digantdesai (commented):
LGTM :)

backends/cortex_m/CMakeLists.txt
psiddh: CMSIS-NN , it is a significant external dependency with specific toolchain requirements (e.g., ARM Cortex-M target, custom toolchain file), ExternalProject_Add is generally preferred . ExternalProject_Add  isolates the CMSIS-NN build from the main cortex build, avoiding potential conflicts or complexity.

backends/cortex_m/CMakeLists.txt
psiddh: In the latest change, I define a CMake cache variable now . So at the time of invoking the CMake file , we can pass in something like this "cmake -DTARGET_CPU=cortex-m33 ..."

backends/cortex_m/CMakeLists.txt
psiddh: Actual headers for CMSIS-NN are in ${CMSIS_NN_ROOT}/Include, so this step points to "cmsis include" directory for the build interface. Not sure if this answers your question

backends/cortex_m/ops/op_add.cpp
psiddh: Removed it, this was an artifact of my previous trials

backends/cortex_m/ops/op_add_tensor.cpp
psiddh: Removed this file too as this was an artifact of my previous trials

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: I went overboard with some of these logs :D (removed)

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Yes we need this , look at the comments starting from Ln 114 in the latest commit (above quantized_add(..) impl)

backends/cortex_m/ops/operators.py
psiddh: I remove this now, this was an artifact of some prev trials, left over code

backends/cortex_m/ops/operators.yaml
psiddh: Yes remove it, left over code from previous trials

backends/cortex_m/ops/operators.yaml
psiddh: removed add.out , add_Tensor  and other left over code from previous trials / experiments.
 In the newly added quantized add pass, it calls quantized_add's default variant but  kernel dispatch mechanism will end up calling the out variant, I think . So it expects proper registration for functional quantized_add as well (along with .out variant)

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: Agree, I am looking to replace this with subgraph_rewriter. I haven't successfully managed to do it so far. Maybe a follow up PR ,if you are ok with it

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: This the actual call that creates the fused node, quantized_add, and calls its .out variant

examples/arm/run.sh
psiddh: Oops

backends/cortex_m/ops/operators.py
psiddh: Not strictly needed, I guess. Can remove.  mostly a defensive check if the caller passes an incorrectly shaped output tensor ?

backends/cortex_m/ops/op_add.cpp
psiddh: removed this as this is an old artifcat

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Yeah ! since the CMSIS-NN API and the kernel function arm_elementwise_add_s8 expect parameters like zero points as int32_t , it is safe and reasonable to extract zero points as int32_t directly. Will udpate

backends/cortex_m/ops/op_add_tensor.cpp
psiddh: I have this helper macro : "VALIDATE_QUANTIZED_INPUTS" in cortex_m_ops_common.h

backends/cortex_m/ops/operators.py
psiddh: I think you meant in C++ impl, Added this check now as part of the macro VALIDATE_QUANTIZED_INPUTS(...)

backends/cortex_m/test/test_quantize_add_fusion_pass.py
psiddh: Yes these passes are logically separate but sequentially dependent. ReplaceQuantNodesPass prepares the graph, and QuantizedAddFusionPass performs the fusion.

backends/cortex_m/test/test_quantize_add_fusion_pass.py
psiddh: Refactored the common code into quantization_utils.py where both passes use the shared one

backends/cortex_m/test/test_quantize_add_fusion_pass.py
psiddh: Added the testcase

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <executorch/kernels/portable/cpu/util/broadcast_util.h>
+#include <executorch/runtime/kernel/kernel_includes.h>
+
+// Include CMSIS-NN headers with C linkage
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+#define VALIDATE_QUANTIZED_INPUTS(ctx, input1, input2, output)              \
```
digantdesai: Just a function?

backends/cortex_m/ops/operators.py
digantdesai: I was thinking an inline function, define is a bit brittle.

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Added assert

backends/cortex_m/ops/operators.py
psiddh: Done the inline function takes care of broadcast functionality

backends/cortex_m/ops/operators.py
psiddh: Done , added helpers

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,137 @@
...
+  int32_t output_mult = static_cast<int32_t>(output_multiplier.to<int64_t>());
+  int output_shift_val = static_cast<int>(output_shift.to<int64_t>());
+
+  // Left shift to maximize precision (tune as needed)
+  const int32_t left_shift = 20;
+  const int32_t activation_min = std::numeric_limits<int8_t>::min();
+  const int32_t activation_max = std::numeric_limits<int8_t>::max();
+
+  // Resize output tensor to match input shape
+  auto err = torch::executor::resize_tensor(out, input1_int8.sizes());
```
digantdesai: Use the broadcasting util for out resize?

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,137 @@
...
+      "Using AoT-computed parameters: input1[mult=%d, shift=%d], input2[mult=%d, shift=%d], output[mult=%d, shift=%d]",
+      input1_mult,
+      input1_shift_val,
+      input2_mult,
+      input2_shift_val,
+      output_mult,
+      output_shift_val);
+
+  // Call CMSIS-NN kernel with precomputed parameters
+  arm_cmsis_nn_status status = arm_elementwise_add_s8(
```
digantdesai, line 87: Does this handle if two inputs are of different shape i.e. [1,2,2] + [2,2,2]

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  return out;
+}
+
+// Stub Implementation: Non-out variant for compatibility (functional variant)
+// EXIR/ExecuTorch runs an out-variant pass that converts
+// .default operations to .out variants before memory planning.
+// In the pass we are calling quantized_add's default variant
+// but ExecuTorch's kernel dispatch mechanism will end up calling the out
+// variant. This stub is to make sure that compiler doesn't complain.
```
digantdesai, line 126: what's the compiler complaint?

I think you might need to write a version for `add_out` which doesn't take `context` arg IIRC.

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,43 @@
...
+inline void validate_quantized_inputs(
+    KernelRuntimeContext& context,
+    const Tensor& input1,
+    const Tensor& input2,
+    Tensor& output) {
+  ET_CHECK_MSG(input1.scalar_type() == ScalarType::Char, "Input1 must be int8");
+  ET_CHECK_MSG(input2.scalar_type() == ScalarType::Char, "Input2 must be int8");
+  ET_CHECK_MSG(output.scalar_type() == ScalarType::Char, "Output must be int8");
+  ET_CHECK_MSG(
+      input1.sizes() == input2.sizes(), "Input tensors must be the same shape");
```
digantdesai: This is too strict. Let's fix this by adding broadcasting support. See portable add impl.

backends/cortex_m/passes/passes_utils.py
```diff
@@ -0,0 +1,94 @@
...
+) -> torch.Tensor:
+    """
+    Simulate CMSIS-NN fixed-point dequantization:
+    result = (qtensor - zero_point) * multiplier * 2^shift / 2^31
+    """
+    scale = multiplier * (2**shift) / (1 << 31)
+    return (qtensor.float() - zero_point) * scale
+
+
+def quantize_tensor(
```
digantdesai:
```suggestion
def quantize_per_tensor_cmsis(
```

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def _replace_operators_with_cortex_m(
+        self, graph_module: torch.fx.GraphModule
+    ) -> int:
+        replacement_count = 0
+        for node in list(graph_module.graph.nodes):
+            if node.op == "call_function":
+                print(f"🔍 Checking node: {node.target}")
```
digantdesai: Use logger.info?

backends/cortex_m/passes/quantized_add_fusion_pass.py
digantdesai: create an issue?

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+                input2_mult, input2_shift = quantize_multiplier_aot(
+                    scale2_val / output_scale_val
+                )
+                output_mult, output_shift = quantize_multiplier_aot(
+                    1.0
+                )  # Output multiplier is 1
+                print("🧮 AoT computed parameters:")
+                print(f"   Input1: mult={input1_mult}, shift={input1_shift}")
+                print(f"   Input2: mult={input2_mult}, shift={input2_shift}")
+                print(f"   Output: mult={output_mult}, shift={output_shift}")
```
digantdesai: remove or logger.info?

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+                            output_mult,
+                            output_shift,
+                        ),
+                        kwargs={},
+                    )
+                    # Copy metadata from the quantize_node to the new fused node
+                    if hasattr(quantize_node, "meta") and quantize_node.meta:
+                        fused.meta = quantize_node.meta.copy()
+                        if "val" in quantize_node.meta:
+                            fused.meta["val"] = quantize_node.meta["val"]
```
digantdesai: Nit: you can also run your target if you prefer.

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+                print(f"🎯 Pattern fused, total so far: {fusion_count}")
+            except Exception as e:
+                print(f"❌ Error during AoT computation: {e}")
+                print("   Skipping fusion for this pattern")
+                continue
+        # Clean up erased nodes
+        for old_node in reversed(nodes_to_erase):
+            if old_node in graph_module.graph.nodes and len(old_node.users) == 0:
+                print(f"🗑️ Erasing node: {old_node}")
+                graph_module.graph.erase_node(old_node)
```
digantdesai: is remove_dead_code available or deprecated?

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+                and add_node.target == exir_ops.edge.aten.add.Tensor
+                and len(add_node.args) >= 2
+            ):
+                continue
+            dequant_node1, dequant_node2 = add_node.args[:2]
+            if not (
+                hasattr(dequant_node1, "op")
+                and dequant_node1.op == "call_function"
+                and dequant_node1.target
+                == exir_ops.edge.cortex_m.dequantize_per_tensor.default
```
digantdesai: this feels off i.e. assigning above and comparing the cortex_m targets here. I agree that it can be either order (depending on which order users ran the pass) but we can assume the cortex-M q/dq pass will run last.

backends/cortex_m/passes/quantized_add_fusion_pass.py
```diff
@@ -0,0 +1,185 @@
...
+    ExportPass that:
+    1. Replaces certain ops with cortex_m variants based on qualifiers.
+    2. Fuses the pattern dequantize_per_tensor -> add -> quantize_per_tensor
+       into cortex_m.quantized_add.default with AoT computed multipliers/shifts.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def _replace_operators_with_cortex_m(
```
digantdesai: drop this? we have another pass for this..

backends/cortex_m/test/test_helpers_passes_utils.py
```diff
@@ -0,0 +1,103 @@
...
+from torchao.quantization.pt2e.quantizer.quantizer import Q_ANNOTATION_KEY
+
+
+@dataclass(eq=True, frozen=True)
+class QuantizationConfig:
+    input_activation: Optional[QuantizationSpec]
+    output_activation: Optional[QuantizationSpec]
+
+
+class AddQuantizer(Quantizer):
```
digantdesai, line 27: Thanks for moving this to shared utils. I guess as we grow we can use either Arm or XNNPACK quantizer. We don't want to scale this..

backends/cortex_m/test/test_quantize_add_fusion_pass.py
digantdesai: AddFusion shouldn't depend on the q/dq pass, but it can assume the order though.. if you are assuming that q/dq will go first this is OK. I would have done other way around because then its less work for q/dq pass if the fusions consume most of the q/dq nodes.

backends/cortex_m/test/test_quantize_add_fusion_pass.py
```diff
@@ -0,0 +1,373 @@
...
+    def _apply_passes(self, edge_program):
+        """Apply both ReplaceQuantNodesPass and QuantizedAddFusionPass"""
+        # Step 1: Replace quant nodes
+        replace_pass = ReplaceQuantNodesPass()
+        intermediate_program = edge_program.transform([replace_pass])
+
+        # Step 2: Apply fusion pass
+        fusion_pass = QuantizedAddFusionPass()
+        final_program = intermediate_program.transform([fusion_pass])
```
digantdesai: you can give these as a list no need to call transform twice

backends/cortex_m/test/test_quantize_add_fusion_pass.py
```diff
@@ -0,0 +1,373 @@
...
+        edge_program = self._prepare_quantized_model(BroadcastAddModel)
+
+        # Apply passes
+        transformed_program = self._apply_passes(edge_program)
+        transformed_graph = transformed_program.exported_program().graph_module
+
+        # Check fusion occurred
+        check_count(transformed_graph, exir_ops.edge.cortex_m.quantized_add.default, 1)
+
+        # Verify numerical equivalence
```
digantdesai: why are these checks commented out?

backends/cortex_m/CMakeLists.txt
digantdesai: Example - https://github.com/pytorch/executorch/blob/6e520e2a8dbe128a36491fbad472dc60f087e05e/backends/arm/scripts/corstone_utils.cmake#L6

We don't have any use of ExternalProject IIRC. I think @swolchok was hinting that FetchContent might be better. But I know what you mean by the 'isolation' aspects of this.

Another, simpler and easier approach would be to add it as a submodule. I know its not always used but it is required if someone wants to build this backend.

examples/arm/aot_arm_compiler.py
```diff
@@ -755,6 +760,15 @@ def to_edge_TOSA_delegate(
...
 
+    gm = edge.exported_program().graph_module
+
+    logging.debug(">>> Lowered GraphModule code <<<")
+    logging.debug(gm.code)  # Python‐style source of the graph
+    logging.debug(">>> Lowered GraphModule nodes <<<")
+    for node in gm.graph.nodes:
+        logging.debug(f"Node: {node.target}, args={node.args}, kwargs={node.kwargs}")
+    logging.debug("==== Graph after quantization ====")
+
```
digantdesai: do we need this?

digantdesai (commented):
s

examples/arm/run.sh
```diff
@@ -202,22 +202,24 @@ backends/arm/scripts/build_executorch.sh --et_build_root="${et_build_root}" --bu
...
     )
     model_compiler_flags=(
         ""                      # 0 softmax
         "--delegate"            # 1 add
         "--delegate"            # 2 add3
         "--delegate --quantize" # 3 qadd
         "--delegate --quantize" # 4 qadd2
-        "--delegate --quantize" # 5 qops
-        "--delegate --quantize" # 6 mv2
+        "--quantize"            # 5 qadd2 (quantize only)
```
AdrianLundell: Why is this test added here?

examples/arm/aot_arm_compiler.py
```diff
@@ -790,7 +804,11 @@ def transform_for_cortex_m_backend(edge):
     # Let's make sure we are using optimized Cortex M backend
     # NB: If we can't find and replace ops those are expected to be replaced,
     # bad things will happen at runtime, like "missing operator" errors!
-    edge = edge.transform([ReplaceQuantNodesPass()])
+    # Instantiate the pass
+    replace_quant_pass = ReplaceQuantNodesPass()
+    quantized_add_fusion_pass = QuantizedAddFusionPass()
```
AdrianLundell: Maybe not in this PR but the Cortex-M backend will need a more structured way of running passes soon

backends/cortex_m/passes/quantized_add_fusion_pass.py
AdrianLundell: All CMSIS-NN ops are quantized so adding one pass per op for fusing them in this way will result in a lot of duplicate code. In the arm backend we "fold" all q/dq-ops into quantized ops like this: https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/fold_qdq_with_annotated_qparams_pass.py. 

Could we reuse that here? And then have one single pass for mapping edge ops with correct dtypes to arm ops perhaps?

psiddh: Nice! Let me look into this pass and see if I can reuse it for this purpose

psiddh: Reviewed the fold_qdq pass, I think it fundamentally serves a diff purpose, afaict, the pass only removes Q/DQ nodes and stores quantization parameters as metadata etc.. The current cortex_m pass on the other hand, replaces the operation itself with custom CMSIS-NN operations (like cortex_m::quantized_add.out) and computes CMSIS-NN specific multipliers and shifts etc..

Moroever with the refactored changes (latest version) , quantized_op_fusion_pass.py is already properly extensible - we can easily add more operations by extending the SUPPORTED_OPS_MAPPING dictionary without creating separate passes for each operation.

psiddh: @digantdesai  Here is the issue to track subgraph_rewriter : https://github.com/pytorch/executorch/issues/13627

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Yes it handles and I validated this in e2e flow

backends/cortex_m/ops/operators.py
psiddh: Resolving this as this header cortex_m_ops_common.h covers all the sanity checks + broadcast functionality

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Done , factored it in cortex_m_ops_common.h (resize_to_broadcast_target_size_quantized(...))

backends/cortex_m/ops/cortex_m_ops_common.h
psiddh: Done, fixed it , much cleaner

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: Done

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: Done

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: Pls review the recent rev (_transfer_metadata() func)

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: In the recent revision  , _normalize_to_cortex_m_targets() handles pass execution order independence. No more assumptions about pass ordering

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: Yes this was not needed at all in the pass (and was redundant) Forgot cleaning this up in the prev iteration, Took this function off now in the latest one

backends/cortex_m/test/test_quantize_add_fusion_pass.py
psiddh: Done

backends/cortex_m/test/test_quantize_add_fusion_pass.py
psiddh: Ok the test case was doing wrong thing previously ,in the sense that it was comparing f32 output with quantized int8 output. I fixed this 'numerical equivalence' in the latest iteration and it passes as expected

examples/arm/aot_arm_compiler.py
psiddh: I just left it, as it proved very useful  while debugging this pass.

examples/arm/run.sh
psiddh: This test case qadd2 (no delegation , only quantize) serves as a good e2e test case on FVP sim to validate / test the fused quantized node flow (with CMSIS-NN integration )

backends/cortex_m/CMakeLists.txt
```diff
@@ -24,29 +19,64 @@ endif()
...
+# Include directories for cortex_m_kernels
+target_include_directories(
+  cortex_m_kernels
+  PRIVATE ${EXECUTORCH_ROOT}/..
+          ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10
+          ${cmsis_nn_SOURCE_DIR}/Include
+)
+
+# Link directly to the CMSIS-NN static library file
+target_link_libraries(
```
AdrianLundell, line 73: Why is it done this way instead of directly adding the library target?

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,184 @@
...
+#include "arm_nnfunctions.h"
+}
+
+using Tensor = torch::executor::Tensor;
+using ScalarType = executorch::aten::ScalarType;
+using Scalar = torch::executor::Scalar;
+using Error = executorch::runtime::Error;
+
+// Basic tensor type / layout validation and dimension order checking
+inline void validate_quantized_tensor_types_and_dim_order(
```
AdrianLundell: Most ops will require channels-last dim_order and there are multiple dtypes available so this name might be confusing in the future.

examples/arm/run.sh
AdrianLundell: I like that we are testing this but adding them to the default list of models to run in run.sh will scale very poorly as we add more ops so I do not think it is the best place for it. Could it be added as a testsuite in https://github.com/pytorch/executorch/blob/main/backends/arm/test/test_arm_baremetal.sh perhaps (or even have a separate test_cortex_m.sh in this backend)?

examples/arm/aot_arm_compiler.py
AdrianLundell: Generally we do not keep debugging code around upstream unless it is not overly verbose and there is a really good reason for it so I would prefer to see this removed.

backends/cortex_m/passes/quantized_add_fusion_pass.py
AdrianLundell: Great improvement with the generalized pass! FWIW this is closer to what we started out with in the Arm backend and we moved to folding the Q/DQ ops separately from all other lowering logic to simplify development as the backend grew more complex. This backend is smaller though so it might not be as big of an issue, we'll see.

backends/cortex_m/ops/op_quantized_add.cpp
AdrianLundell: This could potentially be added as a pass as well: https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/broadcast_args_pass.py. But long term the ideal solution would be to add broadcast support to CMSIS-NN to get it accelerated w/o memcopies.

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,184 @@
...
+    const Scalar& zero_point1,
+    const Scalar& multiplier1,
+    const Scalar& shift1,
+    const Scalar& zero_point2,
+    const Scalar& multiplier2,
+    const Scalar& shift2,
+    const Scalar& output_zero_point,
+    const Scalar& output_multiplier,
+    const Scalar& output_shift,
+    Tensor& output) {
```
digantdesai, line 110: Make it a util to check a single quant params
```suggestion
inline void validate_quantization_params(
    const Scalar& zero_point,
    const Scalar& multiplier,
    const Scalar& shift) {
```

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,184 @@
...
+      "Shift 2 must be in range [-31, 31] [Value: %d]",
+      shift2_val);
+
+  ET_CHECK_MSG(
+      out_shift_val >= -31 && out_shift_val <= 31,
+      "Output shift must be in range [-31, 31] [Value: %d]",
+      out_shift_val);
+}
+
+inline Error resize_to_broadcast_target_size_quantized(
```
digantdesai: why quantization in the name?

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -0,0 +1,184 @@
...
+
+  for (int i = 0; i < input1.dim(); i++) {
+    inp1_shape[i + offset_inp1] = input1.size(i);
+  }
+  for (int i = 0; i < input2.dim(); i++) {
+    inp2_shape[i + offset_inp2] = input2.size(i);
+  }
+  for (int i = 0; i < output.dim(); i++) {
+    out_shape[i + offset_out] = output.size(i);
+  }
```
digantdesai: do we need all this logic if we are going to call the `get_broadcast_target_size` util?

backends/cortex_m/ops/op_quantized_add.cpp
digantdesai: It can't be added as an AoT pass and must be handled at runtime because we support dynamic shapes with these operators.

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,149 @@
...
+Tensor quantized_add(
```
digantdesai, line 127: remove this?

backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -0,0 +1,149 @@
...
+      static_cast<int32_t>(out.numel()),
+      activation_min,
+      activation_max);
+
+  if (status != ARM_CMSIS_NN_SUCCESS) {
+    ET_LOG(
+        Error,
+        "quantized_add_out: arm_elementwise_add_s8 failed with status [%d]",
+        status);
+    std::memset(out.mutable_data_ptr<int8_t>(), 0, out.nbytes());
```
digantdesai: why do we need this?
Set a failure flag with the appropriate `Error` in ctx? `ctx.fail(err);`, so runtime doesn't keep on going with wrong data

backends/cortex_m/passes/quantized_op_fusion_pass.py
```diff
@@ -0,0 +1,273 @@
...
+                from_node_list.append({
+                    "source": source_node.name, "pass": pass_name, "op": "fuse"}
+                )
+                new_node.meta["from_node"] = from_node_list
+
+
+            # Copy essential fields
+            for field in ["val", "tensor_meta", "stack_trace"]:
+                if field in source_node.meta:
+                    new_node.meta[field] = source_node.meta[field]
```
digantdesai, line 99: wouldn't `val` here be fp32?

backends/cortex_m/passes/quantized_op_fusion_pass.py
```diff
@@ -0,0 +1,273 @@
...
+        """Generic fusion for quantized binary operation patterns."""
+        fusion_count = 0
+        nodes_to_erase = []
+
+
+        for node in list(graph_module.graph.nodes):
+            if not self._is_quant_node(node):
+                continue
+
+
```
digantdesai: run code formatting?

examples/arm/aot_arm_compiler.py
digantdesai: Nit: remove?

digantdesai (approved):
Thanks @psiddh, left some nit comments.

backends/cortex_m/passes/quantized_add_fusion_pass.py
psiddh: eliminate_dead_code has been added, resolving this issue

backends/cortex_m/passes/quantized_op_fusion_pass.py
psiddh: Yes the val field in node metadata is fp32  from the original unquantized computation. This was done for debugging. In the latest iter, removed it

backends/cortex_m/ops/cortex_m_ops_common.h
psiddh: Not needed anymore, removed it

backends/cortex_m/ops/cortex_m_ops_common.h
psiddh: Removed

backends/cortex_m/CMakeLists.txt
psiddh: The CMSIS-NN library is linked by specifying the static library file path directly because CMSIS-NN is brought in via FetchContent and does not define a CMake target in the build.

backends/cortex_m/CMakeLists.txt
```diff
@@ -24,29 +19,64 @@ endif()
...
+  cortex_m_kernels PUBLIC ${cmsis_nn_BINARY_DIR}/libcmsis-nn.a executorch
+)
+
+# Add dependency to ensure CMSIS-NN builds before we try to link. Use the
+# actual CMSIS-NN target name (usually 'cmsis-nn')
+add_dependencies(cortex_m_kernels cmsis-nn)
+
```
mergennachin, line 80: There are no C++ testing? 

```
if(CORTEX_M_BUILD_TESTS)
  enable_testing()
  add_subdirectory(tests)
endif()
```

backends/cortex_m/CMakeLists.txt
```diff
@@ -24,29 +19,64 @@ endif()
...
 )
 message("Generated files ${gen_command_sources}")
 
-# Build a library for _cortex_m_kernels_srcs
+# Build a library for cortex_m_kernels
 add_library(cortex_m_kernels ${_cortex_m_kernels__srcs})
-target_link_libraries(cortex_m_kernels PRIVATE executorch)
 target_compile_options(cortex_m_kernels PUBLIC ${_common_compile_options})
 
+# Include directories for cortex_m_kernels
```
mergennachin, line 64: besides `_common_compile_options` don't you also need cortex-m specific compile options?

mergennachin (commented):
Can you convert this instructions to CI? Could be a follow-up PR.

I believe we should be able to run on on our AWS graviton instances. 

Looks like there is test_arm_baremetal.sh running, either extend that or create new tests to exercise the code you just added please.

Test Plan:

examples/arm/run.sh - No regressions ==> Ok
examples/arm/run.sh now runs 'qadd2' in quantize only mode ==> Ok
python -m unittest test_replace_quant_nodes.py ==> Ok
python -m unittest test_quantize_op_fusion_pass.py ==> Ok

backends/cortex_m/ops/op_quantized_add.cpp
psiddh: Done

backends/cortex_m/CMakeLists.txt
psiddh: @mergennachin  Created task to track dedicated test suite for cortex_m ops (that should cover Python tests, C++ unit tests & E2E tests): https://github.com/pytorch/executorch/issues/13739.   (and the initial work on this: PR : https://github.com/pytorch/executorch/pull/1357)

digantdesai: FYI we can't run cortex-m op unittests without FVP. but for c++ only utils, sure.

backends/cortex_m/CMakeLists.txt
digantdesai: comes from the toolchain file.

psiddh:
Three follow up issues identified to make further improvements:

- Optimization Improvements : https://github.com/pytorch/executorch/issues/13748
- Pass Refactor/ Simplification: https://github.com/pytorch/executorch/issues/13627 
- Add dedicated Test Suite: https://github.com/pytorch/executorch/issues/13739

PR #13304, opened 2025-08-11
Added ETDump to Wasm

Conarnar:
### Summary
Turning on the `EXECUTORCH_ENABLE_EVENT_TRACER` option will enable event tracing in the Wasm module API. The results can be obtained with the `etdump()` method.

### Test plan
Added two tests depending on whether `EXECUTORCH_ENABLE_EVENT_TRACER` is turned on or not. Added the `--enable-etdump` option to `scripts/build_wasm_tests.sh` which turns on the above option.

Added configurations to the `unittest-wasm-bindings` CI test to run with and without `--enable-etdump`.


scripts/build_wasm_tests.sh
```diff
@@ -5,6 +5,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+for arg in "$@"; do
+  if [ "$arg" == "--enable-etdump" ]; then
+    ETDUMP_OPTS="-DEXECUTORCH_ENABLE_EVENT_TRACER=ON \
+                 -DEXECUTORCH_BUILD_DEVTOOLS=ON \
+                 -DFLATCC_ALLOW_WERROR=OFF"
```
lucylq, line 12: Add a comment here on why this flag is required?

lucylq (approved):
LGTM, also cc @Gasoonjia on devtools

Gasoonjia:
LGTM. 
Is there any buck target we need to update?

lucylq:
> LGTM. Is there any buck target we need to update?

I think all of the wasm work has been in OSS, so there aren't buck targets for it and it won't break CI.

Conan and I had a quick look, and there is emscripten internally that we could use from fbsource/third-party if we wanted to buckify things cc @JacobSzwejbka

PR #13308, opened 2025-08-12
[Backend Tester] Write report progressively

GregoryComer:
Append to the report file line by line after each test, rather than all at the end. This ensures that report data is available if the test run is aborted or hit with an unrecoverable native crash.


GregoryComer:
* #13360
* #13313
* #13312
* #13311
* __->__ #13308

backends/test/suite/reporting.py
```diff
@@ -262,6 +300,15 @@ def log_test_summary(summary: TestCaseSummary):
...
         _active_session.test_case_summaries.append(summary)
 
+        if _active_session.report_path is not None:
+            file_mode = "a" if _active_session.has_written_report_header else "w"
+            with open(_active_session.report_path, file_mode) as f:
+                if not _active_session.has_written_report_header:
+                    write_csv_header(f)
+                    _active_session.has_written_report_header = True
+
+                write_csv_row(summary, f)
```
digantdesai, line 310: This implies we crash when we run into some failure? Can we try to catch and fail gracefully? Instead of assuming that we can crash anytime.

backends/test/suite/reporting.py
```diff
@@ -262,6 +300,15 @@ def log_test_summary(summary: TestCaseSummary):
     if _active_session is not None:
         _active_session.test_case_summaries.append(summary)
 
+        if _active_session.report_path is not None:
```
digantdesai, line 303: can multiple subprocesses write to this simultaneously?

backends/test/suite/reporting.py
GregoryComer: This is mainly for memory corruption, which happens in a few cases. It might be possible to catch the SIGSEGV or other native fault, but I don't know what state the process is it, so I'm not sure if it's recoverable. Open to suggestions.

backends/test/suite/reporting.py
GregoryComer: Probably not safely. What is the use case? Are you thinking we parallelize tests between processes? That seems nice to have, though I'd be inclined to deal with concurrency issues if/when we add that feature.

backends/test/suite/reporting.py
digantdesai: is this through pybinding? Does it crash the cpython process? or this is in a subprocess? if its the main process then yeah I don't know if we can try..catch but if its a subprocess then maybe.

OK with unblocking you now..

backends/test/suite/reporting.py
digantdesai: yeah all popular test frameworks do runs in parallel threads, so perhaps in the future.

digantdesai (approved):

Issue #12895, opened 2025-07-28
[Delegate Testing] Add pybinding support for Vulkan


digantdesai:
As title

cc @SS-JIA @manuelcandales @cbilgin

GregoryComer:
Stephen landed https://github.com/pytorch/executorch/pull/13327. Closing.

PR #13306, opened 2025-08-12
[Backend Tester] Clean up report output

GregoryComer:
Update the report output to be a bit cleaner. I've made the following changes:
 * Round decimal values to 3 decimal places.
 * Remove mean signed deviation and L2 norm. We have SNR, mean absolute error, and max element-wise error. I think that's sufficient.
 * Rename SQNR to SNR. Not all tests are quantized and we're using it as a general measure of tensor-wise error.
 * Split out the result column into pass/skip/fail and a detailed reason. This should be easier to parse at a glance.

Example output (not sure markdown formatting is weird here...):

Test ID | Test Case | Flow | Params | Result | Result Detail | Delegated | Quantize Time (s) | Lower Time (s) | Delegated Nodes | Undelegated Nodes | Delegated Ops | Undelegated Ops | PTE Size (Kb) | Output 0 Error Max | Output 0 Error MAE | Output 0 SNR | Output 1 Error Max | Output 1 Error MAE | Output 1 SNR
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
test_add_dtype_float32_xnnpack | test_add_dtype | xnnpack | {'dtype': torch.float32} | Pass |   | TRUE |   | 3.747 | 1 | 0 | {'aten::add.Tensor': 1} | {} | 1.6 | 0 | 0 | inf |   |   |  
test_add_dtype_float32_xnnpack_static_int8_per_channel | test_add_dtype | xnnpack_static_int8_per_channel | {'dtype': torch.float32} | Pass |   | TRUE | 0.66 | 0.818 | 7 | 0 | {'aten::add.Tensor': 1, 'quantized_decomposed::dequantize_per_tensor': 3, 'quantized_decomposed::quantize_per_tensor': 3} | {} | 2 | 0 | 0 | inf |   |   |

PR #13309, opened 2025-08-12
Link Vulkan backend with pybinding lib when built

GregoryComer:
Add the Vulkan backend as a dependency to the pybinding build, when built. I've validated that I can run a simple Vulkan model on pybindings with this change on Mac (via MoltenVK). 

Note that I'm not updating the preset to always build Vulkan yet. It requires additional dependencies and thus needs more discussion. I'm not sure if we can auto-install everything that we need. I'll get @SS-JIA's opinion on this. Also, once executorch::backends target is installable, I'll switch pybindings to use that target.

Tracking task: https://github.com/pytorch/executorch/issues/12895.


GregoryComer:
* #13313
* #13312
* #13311
* __->__ #13309
* #13308
* #13306
* #13258
* #13250

digantdesai (approved):
Thanks. Let's discuss this. @GregoryComer  do you mean other deps like - .ci/scripts/setup-vulkan-linux-deps.sh ? 

Also FYI Stephen was saying SwiftShader may not have all the Vulkan extensions we need so that might make testing a bit more complicated.

GregoryComer:
Closing as changes landed separately.

PR #13311, opened 2025-08-12
[Backend Tester] Add subtest index field

GregoryComer:
Some test functions run multiple test cases. Add a subtest index field to disambiguate these in the report.


GregoryComer:
* #13360
* #13313
* #13312
* __->__ #13311

backends/test/suite/models/__init__.py
```diff
@@ -119,6 +119,7 @@ def run_model_test(
         flow,
         context.test_name,
         context.test_base_name,
+        0, # subtest_index - currently unused for model tests
```
digantdesai: So the hierarchy is 
Test ID -> test -> subtest?

digantdesai (approved):

backends/test/suite/models/__init__.py
GregoryComer: The naming is a bit confusing. I'm open to suggestions. The test_base_name is the test name (test_add_dtype, for example). Test name adds flow and parameterization (test_add_dtype_float32_xnnpack). The subtest index corresponds to a single python unittest function that runs multiple test cases.

PR #13312, opened 2025-08-12
[Backend Tester] Reduce log verbosity / spam

GregoryComer:
When running tests, deprecation warnings for export_for_training and "internal consistency verification was requested but not available" create a large amount of spam in the CLI and drown out actual test info during the run. Neither warning is particularly useful. I've suppressed the the export_for_training warning in the backend tester only and switched the tester to not request internal consistency verification since it's not compiled into pybindings by default. This gives much cleaner CLI output.

Before:
```
test_add_dtype_float32_xnnpack_static_int8_per_channel (test_add.Add.test_add_dtype_float32_xnnpack_static_int8_per_channel) ... /home/gregory/src/executorch/src/executorch/backends/test/harness/stages/quantize.py:50: FutureWarning: `torch.export.export_for_training` is deprecated and will be removed in PyTorch 2.10. Please use `torch.export.export` instead, which is functionally equivalent.
  captured_graph = export_for_training(artifact, inputs, strict=True).module()
/home/gregory/miniconda3/envs/executorch/lib/python3.12/site-packages/torchao/quantization/pt2e/utils.py:818: FutureWarning: `torch.export.export_for_training` is deprecated and will be removed in PyTorch 2.10. Please use `torch.export.export` instead, which is functionally equivalent.
  aten_pattern = torch.export.export_for_training(
/home/gregory/miniconda3/envs/executorch/lib/python3.12/site-packages/torchao/quantization/pt2e/utils.py:818: FutureWarning: `torch.export.export_for_training` is deprecated and will be removed in PyTorch 2.10. Please use `torch.export.export` instead, which is functionally equivalent.
  aten_pattern = torch.export.export_for_training(
/home/gregory/miniconda3/envs/executorch/lib/python3.12/site-packages/torchao/quantization/pt2e/utils.py:818: FutureWarning: `torch.export.export_for_training` is deprecated and will be removed in PyTorch 2.10. Please use `torch.export.export` instead, which is functionally equivalent.
  aten_pattern = torch.export.export_for_training(
/home/gregory/miniconda3/envs/executorch/lib/python3.12/site-packages/torchao/quantization/pt2e/utils.py:818: FutureWarning: `torch.export.export_for_training` is deprecated and will be removed in PyTorch 2.10. Please use `torch.export.export` instead, which is functionally equivalent.
... (it logs like 20x per test)
```

After:
```
test_add_dtype_float32_xnnpack (test_add.Add.test_add_dtype_float32_xnnpack) ... ok
test_add_dtype_float32_xnnpack_static_int8_per_channel (test_add.Add.test_add_dtype_float32_xnnpack_static_int8_per_channel) ... ok
test_add_f32_alpha_xnnpack (test_add.Add.test_add_f32_alpha_xnnpack) ... ok
test_add_f32_alpha_xnnpack_static_int8_per_channel (test_add.Add.test_add_f32_alpha_xnnpack_static_int8_per_channel) ... ERROR
...
```


GregoryComer:
* #13360
* #13313
* __->__ #13312
* #13311

backends/test/harness/stages/serialize.py
```diff
@@ -39,7 +40,9 @@ def graph_module(self) -> None:
 
     def run_artifact(self, inputs):
         inputs_flattened, _ = tree_flatten(inputs)
-        executorch_module = _load_for_executorch_from_buffer(self.buffer)
+        executorch_module = _load_for_executorch_from_buffer(
+            self.buffer, program_verification=Verification.Minimal
```
digantdesai, line 44: what does this do? i.e. Minimal ==> "do the verification but don't print warnings"?

backends/test/suite/runner.py
```diff
@@ -249,6 +250,10 @@ def build_test_filter(args: argparse.Namespace) -> TestFilter:
 def runner_main():
     args = parse_args()
 
+    # Suppress deprecation warnings for export_for_training, as it generates a
+    # lot of log spam. We don't really need the warning here.
+    warnings.simplefilter("ignore", category=FutureWarning)
```
digantdesai, line 255: FYI PyTest has this nice feature where they count same warnings and at the end print warning summary https://docs.pytest.org/en/stable/how-to/capture-warnings.html

digantdesai (approved):

backends/test/harness/stages/serialize.py
GregoryComer: This doesn't actually cause any change in verification, because the extended verification feature isn't compiled into pybindings by default. So the runtime logs a warning because it was requested but not available. This change updates the code to not ask for extended verification.

Here's the relevant code from the runtime. The "InternalConsistency verification requested but not available " log is the one contributing to log spam. https://github.com/pytorch/executorch/blob/d7ecd871ce2472153aeedd83e87a0018f82f9bc5/runtime/executor/program.cpp#L123-L136

backends/test/suite/runner.py
GregoryComer: I want to look at using pytest or another framework once we get the core harness and suite landed. I originally went with unittest, in part, because staying closer to unittest made it easier to run within buck/Meta-internal. But if that constraint is lifted, we can probably leverage pytest for some nicer features without having to make too many code changes.

PR #13313, opened 2025-08-12
[Backend Tester] Seed based on test name

GregoryComer:
Set a manual seed for pytorch based on the test base name (test case not including flow / etc.). This makes test results stable between runs and between backends/flows. This is useful for comparing accuracy between backends, for example.

I validated this change by running convolution tests for xnnpack twice. I validated that the output accuracy statistics were identical.


GregoryComer:
* #13390
* #13360
* __->__ #13313

backends/test/suite/runner.py
```diff
@@ -40,6 +41,16 @@
 }
 
 
+def _get_test_seed(test_base_name: str) -> int:
```
digantdesai, line 46: Why not set a new, global seed every run? And print it somewhere to reproduce. Hardcoding seed ==> we will test with same random numbers every time, not sure if that's what we want.

GregoryComer: Updated to generate a random run-wide seed and allow specifying a seed from CLI.

digantdesai (approved):

PR #13315, opened 2025-08-12
Arm backend: Bugfix TosaValueError

per:
ValueError doesn't accept kwargs, so remove the usage and explicitly add the op arg instead.


per:
Unrelated failure in arm-zephyr.

Sebastian-Larsson (approved):

PR #13316, opened 2025-08-12
Arm backend: Remove function getNodeArgs

Sebastian-Larsson:
The function was only used in one place and it was only a couple of lines so the logic was moved to where the function was originally called, which is in process_node.py.


Sebastian-Larsson:
Unrelated failures

zingo (approved):

PR #13317, opened 2025-08-12
Arm backend: Bump Arm baremetal compilers to gcc 14.3

zingo:


mansnils (approved):

examples/arm/setup.sh
```diff
@@ -275,39 +275,39 @@ function setup_fvp() {
...
+                # TODO can include support for zephyr toolchain for other host platforms later
                 toolchain_url="https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-x86_64_arm-zephyr-eabi.tar.xz"
                 toolchain_dir="arm-zephyr-eabi"
                 toolchain_md5_checksum="93128be0235cf5cf5f1ee561aa6eac5f"
             else
-                toolchain_url="https://armkeil.blob.core.windows.net/developer/Files/downloads/gnu/13.3.rel1/binrel/arm-gnu-toolchain-13.3.rel1-x86_64-arm-none-eabi.tar.xz"
-                toolchain_dir="arm-gnu-toolchain-13.3.rel1-x86_64-arm-none-eabi"
-                toolchain_md5_checksum="0601a9588bc5b9c99ad2b56133b7f118"
-	    fi
+                toolchain_url="https://developer.arm.com/-/media/Files/downloads/gnu/14.3.rel1/binrel/arm-gnu-toolchain-14.3.rel1-x86_64-arm-none-eabi.tar.xz"
```
digantdesai, line 284: Nit: if the URL names are predictable, extract 14.3 out as a variable might be better to avoid missing some toolchain in future updates.

digantdesai (approved):

examples/arm/setup.sh
zingo: I kind of agree and maybe we want to do this over time. As it's part of the filename it is now at lease nice to directly be able to cut&paste and search the web for it. Lets see if we change later.

zingo:
Will be replace with a better/newer version some day hopefully. This had gone old and needs a rebase.

PR #13318, opened 2025-08-12
Arm backend: Initial int16 extension

per:
### Summary
Add TOSA extension support for tests and add initial support for int16.

### Test plan
Tested through unit tests in backends/arm.

### DIFF
--- a/backends/arm/arm_backend.py
+++ b/backends/arm/arm_backend.py
@@ -128,7 +128,7 @@ class ArmCompileSpecBuilder:
         self.compiler_flags.append("--output-format=raw")
         self.compiler_flags.append("--debug-force-regor")
 
-        base_tosa_version = "TOSA-1.0+INT"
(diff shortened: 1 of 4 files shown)


zingo (approved):

per:
Unrelated failures in test-models-arm-zephyr and build-demo-ios

backends/arm/arm_backend.py
```diff
@@ -128,7 +128,7 @@ def ethosu_compile_spec(
         self.compiler_flags.append("--output-format=raw")
         self.compiler_flags.append("--debug-force-regor")
 
-        base_tosa_version = "TOSA-1.0+INT"
+        base_tosa_version = "TOSA-1.0+INT+int16"
```
digantdesai, line 131: We have been using `ts.DType.INT16` in a few places, how did that work without this flag?

PR #13322, opened 2025-08-12
Arm backend: Run pytest model tests in parallell

Erik-Lundell:
Reduces time of pytest model jobs with ~50%. Upstream ci tested successfully 4 times, and easy to revert if it turns out to be flaky. 

### DIFF
--- a/backends/arm/test/test_arm_baremetal.sh
+++ b/backends/arm/test/test_arm_baremetal.sh
@@ -101,7 +101,7 @@ test_pytest_models() { # Test ops and other things
     source backends/arm/scripts/install_models_for_test.sh
 
     # Run arm baremetal pytest tests without FVP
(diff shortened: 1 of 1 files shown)


zingo (approved):

backends/arm/test/test_arm_baremetal.sh
```diff
@@ -101,7 +101,7 @@ test_pytest_models() { # Test ops and other things
     source backends/arm/scripts/install_models_for_test.sh
 
     # Run arm baremetal pytest tests without FVP
-    pytest  --verbose --color=yes --durations=0 backends/arm/test/models
+    pytest  --verbose --color=yes --numprocesses=auto --durations=0 backends/arm/test/models
```
digantdesai, line 104: I always assumed it would do this automatically after reading `worker` related messages. but this is good, I didn't realize `--dist=no` by default.

PR #13323, opened 2025-08-12
Arm backend: Add limited support for fish shell in setup_path.sh

zingo:
Generate both sh and fish shell versions of the setup_path script

### DIFF
--- a/examples/arm/setup.sh
+++ b/examples/arm/setup.sh
@@ -199,7 +199,7 @@ function check_options() {
 function setup_root_dir() {
     mkdir -p ${root_dir}
     root_dir=$(realpath ${root_dir})
-    setup_path_script="${root_dir}/setup_path.sh"
+    setup_path_script="${root_dir}/setup_path"
 }
 
 function check_fvp_eula () {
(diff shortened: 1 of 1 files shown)


mansnils (approved):

zingo:
This should go in after https://github.com/pytorch/executorch/pull/13369

examples/arm/setup.sh
```diff
@@ -334,46 +334,56 @@ function setup_vela() {
...
+    echo "set -Ux PATH $1" >> ${setup_path_script}.fish
+}
+
 function create_setup_path(){
     cd "${root_dir}"
 
-    echo "" > "${setup_path_script}"
+    # Clear setup_path_script
+    echo "" > "${setup_path_script}.sh"
+    echo "" > "${setup_path_script}.fish"
```
digantdesai, line 351: why no zsh? :p

examples/arm/setup.sh
```diff
@@ -461,8 +475,8 @@ if [[ $is_script_sourced -eq 0 ]]; then
         setup_vela
     fi
 
-    echo "[main] update path by doing 'source ${setup_path_script}'"
-
+    echo "[main] update path by doing 'source ${setup_path_script}.sh'"
+    hash fish 2>/dev/null && echo >&2 "[main] Or for fish shell use 'source ${setup_path_script}.fish'"
```
digantdesai, line 477: can we just generate one based on which shell user is kicking this from?

examples/arm/setup.sh
zingo: As I understand it zsh can use the sh script, Im no zsh user myself so I have not tested, but we have a few in the project an no one have complain or thought is was a pain point enough to create a PR :)

examples/arm/setup.sh
zingo: As this script is runned in a sh shell I'm not sure we can have some logic for this, I also I feel the extra logic with extra possible errors introduced in various target does not merit compared to the small size of the generated script. e.g. it's probably just faster and less error prune to create both. 

But since there are probably a lot less fish users I didn't want to pollute the logs so I added some simple logic  here to see if fish was installed.

zingo:
fails are unrelated

examples/arm/setup.sh
digantdesai: That is a good point.

examples/arm/setup.sh
digantdesai: Thanks.

PR #13300, opened 2025-08-11
[DNL] small change to trigger CI testing

SS-JIA:
Summary:
PR to check if the test added in below PR is working or not.

PR #13295, opened 2025-08-11
[ET-VK][CI] Add vulkan CI to test exporting and running models

SS-JIA:
Summary:
Title says it all!

cc @manuelcandales @cbilgin

PR #13294, opened 2025-08-11
[ET-VK][examples] Create export script for Vulkan examples

SS-JIA:
Summary:
Title says it all!

Introduce scripts to facilitate exporting and testing Vulkan delegate models. These scripts will be used in the next PR to add CI testing for Vulkan lowered models.

cc @manuelcandales @cbilgin

PR #13326, opened 2025-08-12
[ET-VK] Add optional blocklist and allowlist to vulkan partitioner to aid debugging

SS-JIA:
Summary:
## Changes

* Add `operator_allowlist` and `operator_blocklist` optional arguments to `VulkanPartitioner`
* `operator_blocklist` will prevent operators in the block list to be lowered to Vulkan
* `operator_allowlist` will only allow operators in the allow list to be lowered to Vulkan
* `operator_allowlist` takes precedence over `operator_blocklist`

## Context

When debugging models, it is useful to be able to prevent certain operators from being lowered to Vulkan, or to only allow certain operators from being lowered to Vulkan. This can help isolate which ops are causing model output to be incorrect.

Test Plan:
## Test Plan

Tested this feature locally while debugging example models.


SS-JIA:
* #13300
* #13295
* #13294
* #13327
* __->__ #13326

manuelcandales (approved):

PR #13335, opened 2025-08-12
Support for Running Arm Zephyr Toolchain on aarch64 Host Machines

BujSet:
### Summary
When testing ExecuTorch with Arm devices running Zephyr, only the toolchain download path for the x86-64 host was included. This PR adds the path for a aarch64 machine, so that that Zephyr flows can be run and validated on aarch64 machines.

### DIFF
--- a/examples/arm/setup.sh
+++ b/examples/arm/setup.sh
@@ -299,10 +299,9 @@ function select_toolchain() {
 	    fi
         elif [[ "${OS}" == "Linux" ]]; then
 	    if [[ "${target_toolchain}" == "zephyr" ]]; then
-	        # eventually, this can be support by downloading the the toolchain from 
-		# "https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-aarch64_arm-zephyr-eabi.tar.xz"
-		# but for now, we error if user tries to specify this
-                echo "[main] Error: currently target_toolchain zephyr is only support for x86-64 Linux host systems!"; exit 1;
+                toolchain_url="https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-aarch64_arm-zephyr-eabi.tar.xz"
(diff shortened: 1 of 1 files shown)


larryliu0820 (approved):

examples/arm/setup.sh
```diff
@@ -299,10 +299,9 @@ function select_toolchain() {
 	    fi
         elif [[ "${OS}" == "Linux" ]]; then
 	    if [[ "${target_toolchain}" == "zephyr" ]]; then
-	        # eventually, this can be support by downloading the the toolchain from 
-		# "https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-aarch64_arm-zephyr-eabi.tar.xz"
-		# but for now, we error if user tries to specify this
-                echo "[main] Error: currently target_toolchain zephyr is only support for x86-64 Linux host systems!"; exit 1;
+                toolchain_url="https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.17.2/toolchain_linux-aarch64_arm-zephyr-eabi.tar.xz"
```
larryliu0820, line 302: If we are recommending 0.17.2 we should probably do the same for all if branches.

larryliu0820: nvm it's the same in other branches as well

zingo: We have a PR to start using GCC 14.3 ongoing here
https://github.com/pytorch/executorch/pull/13317
This is not know benefit/downside with this more then later=better so I don't think/see any problem right now to have different version. But maybe we should wait with the 14.3 move until zephyr also make the move?

zingo (approved):

examples/arm/setup.sh
BujSet: Yea, not a bad idea to wait imo.

digantdesai: > #13317 

That's not for Zephyr though?


And FWIW 0.17.3 is out and is 2 weeks old.

PR #13341, opened 2025-08-12
Check the number of inputs in Method::set_inputs

shoumikhin:


runtime/executor/method.cpp
```diff
@@ -1178,6 +1178,10 @@ Method::set_input(const EValue& input_evalue, size_t input_idx) {
 ET_NODISCARD Error
 Method::set_inputs(const executorch::aten::ArrayRef<EValue>& input_evalues) {
   const size_t n_input = inputs_size();
+  ET_CHECK_OR_RETURN_ERROR(
+      input_evalues.size() == n_input,
+      InvalidArgument,
+      "Invalid number of inputs provided.");
```
swolchok: it would be nice to include the actual and expected number of inputs here, wouldn't it?

swolchok (approved):

PR #13342, opened 2025-08-12
Add logical and op

cccclai:
Summary:
As title, add the logical and op for an internal model

Rollback Plan:

Differential Revision: D80122607

### DIFF
--- a/backends/qualcomm/tests/test_qnn_delegate.py
+++ b/backends/qualcomm/tests/test_qnn_delegate.py
@@ -924,6 +924,13 @@ class TestQNNFloatingPointOperator(TestQNN):
         sample_input = (torch.rand([1, 2, 3, 4]),)
         self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_logical_and(self):
+        module = LogicalAnd()  # noqa: F405
+        input1 = torch.tensor([True, False, True, False])
+        input2 = torch.tensor([True, True, False, False])
+        sample_input = (input1, input2)
+        self.lower_module_and_test_output(module, sample_input)
+
     def test_qnn_backend_logical_not(self):
         module = LogicalNot()  # noqa: F405
         sample_input = (torch.rand([1, 2, 3, 4]),)
@@ -2481,6 +2488,14 @@ class TestQNNQuantizedOperator(TestQNN):
     def test_qnn_backend_log(self):
         module = Log()  # noqa: F405
         sample_input = (torch.rand([1, 2, 3, 4]),)
+        module = self.get_qdq_module(module, sample_input)
+        self.lower_module_and_test_output(module, sample_input)
+
+    def test_qnn_backend_logical_and(self):
+        module = LogicalAnd()  # noqa: F405
+        input1 = torch.tensor([0.0])
+        input2 = torch.tensor([1.0])
+        sample_input = (input1, input2)
         module = self.get_qdq_module(module, sample_input)
         self.lower_module_and_test_output(module, sample_input)
 
--- a/backends/qualcomm/_passes/layout_transform.py
+++ b/backends/qualcomm/_passes/layout_transform.py
@@ -93,6 +93,7 @@ class LayoutTransform(ExportPass):
         exir_ops.edge.aten.le.Tensor,
         exir_ops.edge.aten.linear.default,
         exir_ops.edge.aten.log.default,
+        exir_ops.edge.aten.logical_and.default,
         exir_ops.edge.aten.logical_not.default,
         exir_ops.edge.aten.lt.Scalar,
         exir_ops.edge.aten.lt.Tensor,
--- a/backends/qualcomm/builders/op_and.py
+++ b/backends/qualcomm/builders/op_and.py
@@ -16,7 +16,7 @@ from .qnn_constants import OpElementWiseAnd, QNN_OP_PACKAGE_NAME_QTI_AISW
 
 @register_node_visitor
 class OpAnd(NodeVisitor):
-    target = ["aten.bitwise_and.Tensor"]
+    target = ["aten.bitwise_and.Tensor", "aten.logical_and.default"]
 
     def __init__(self, *args) -> None:
         super().__init__(*args)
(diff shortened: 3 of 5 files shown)


backends/qualcomm/builders/qnn_constants.py
```diff
@@ -200,6 +200,11 @@ class OpElementWiseNeuron:
     param_beta: str = "beta"
 
 
+@dataclass(init=False, frozen=True)
+class OpElementWiseNot:
```
haowhsu-quic: I guess this should be `OpElementWiseAnd`.

cccclai: Oops updated

cccclai:
Can I get a review on this PR? @haowhsu-quic @shewu-quic @winskuo-quic @DannyYuyang-quic

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -856,6 +856,13 @@ def test_qnn_backend_log(self):
         sample_input = (torch.rand([1, 2, 3, 4]),)
         self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_logical_and(self):
```
haowhsu-quic, line 927: Could you also add this test case to `TestQNNQuantizedOperator`? thanks.

backends/qualcomm/builders/op_logical_and.py
```diff
@@ -0,0 +1,66 @@
...
+        output_tensor = self.get_tensor(node, node)
+        output_tensor_wrapper = self.define_tensor(
+            node,
+            node,
+            output_tensor,
+            PyQnnWrapper.Qnn_TensorType_t.QNN_TENSOR_TYPE_NATIVE,
+            nodes_to_wrappers,
+        )
+
+        logical_not_op = PyQnnWrapper.PyQnnOpWrapper(
```
haowhsu-quic: `logical_and_op` would be better I think.

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: Good catch - updated

backends/qualcomm/builders/op_logical_and.py
cccclai: Ah yes, copy paste error...thanks and updated

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -2244,6 +2252,13 @@ def test_qnn_backend_log(self):
         module = self.get_qdq_module(module, sample_input)
         self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_logical_and(self):
+        module = LogicalAnd()  # noqa: F405
+        input1 = torch.tensor([True, False, True, False])
+        input2 = torch.tensor([True, True, False, False])
+        sample_input = (input1, input2)
+        self.lower_module_and_test_output(module, sample_input)
```
haowhsu-quic, line 2500: Might need `module = self.get_qdq_module(module, sample_input)` here.

backends/qualcomm/builders/qnn_constants.py
```diff
@@ -200,6 +200,11 @@ class OpElementWiseNeuron:
     param_beta: str = "beta"
 
 
+@dataclass(init=False, frozen=True)
+class OpElementWiseAnd:
+    op_name: str = "ElementWiseAnd"
```
DannyYuyang-quic: Hi @cccclai, since we already have `OpElementWiseAnd` at line 111, I believe there's no need to add it again. Thanks!!

DannyYuyang-quic (approved):
Thanks for supporting more ops!!

backends/qualcomm/builders/op_logical_and.py
```diff
@@ -0,0 +1,66 @@
...
+import torch
+
+from .node_visitor import NodeVisitor
+from .node_visitor_manager import register_node_visitor
+from .qnn_constants import OpElementWiseAnd, QNN_OP_PACKAGE_NAME_QTI_AISW
+
+
+@register_node_visitor
+class And(NodeVisitor):
+    target = ["aten.logical_and.default"]
```
haowhsu-quic: I wonder if we could extend the `target` field in `builders/op_and.py`?

cccclai: I'm trying to do that, however I found out the logical_and op doesn't have the corresponding q/dq ops when trying `module = self.get_qdq_module(module, sample_input)`, do you know which path I should be updating?

shewu-quic: I think you need to extend annotation with new target in `quantizer/annotator.py` https://github.com/pytorch/executorch/blob/1520f9f9de6d410894fad7acbd8d9f53924c2a67/backends/qualcomm/quantizer/annotators.py#L831

cccclai: Is the latest version correct? It seems still missing the q/dq ops

winskuo-quic: I think it is because the input is bool type so it won't have qdq. Could you try either 1) FP input. 2) When calling `get_qdq_module` in UT, also set `bypass_check=True`?

shewu-quic: Oops, I believe the reason is because the input is bool (uint8) rather than float.
https://github.com/pytorch/executorch/blob/1520f9f9de6d410894fad7acbd8d9f53924c2a67/backends/qualcomm/quantizer/annotators.py#L194

cccclai: Ah thanks! Yeah it's indeed the reason

cccclai:
I have it fix, does the new version look good now? @shewu-quic @haowhsu-quic @winskuo-quic @DannyYuyang-quic

shewu-quic:
LGTM. Thanks!

shewu-quic (approved):

shoumikhin (approved):

PR #13359, opened 2025-08-13
Non-fatal error when ET_SWITCH encounters unsupported dtype

manuelcandales:
Summary: Modify `ET_INTERNAL_SWITCH` macro, so that we fail non-fatally in the default clause when we encounter an unsupported dtype.

Differential Revision: D80141272

### DIFF
--- a/kernels/portable/cpu/op_fill.cpp
+++ b/kernels/portable/cpu/op_fill.cpp
@@ -90,7 +90,7 @@ Tensor& fill_tensor_out(
   static constexpr const char op_name[] = "fill.Tensor_out";
 
   ET_SWITCH_REALHBBF16_TYPES(a_type, ctx, op_name, CTYPE_A, [&] {
-    CTYPE_A b_casted;
+    CTYPE_A b_casted{};
     ET_SWITCH_REALHBBF16_TYPES(b_type, ctx, op_name, CTYPE_B, [&] {
       CTYPE_B b_val;
       ET_EXTRACT_SCALAR_TENSOR(b, b_val);
--- a/kernels/portable/cpu/util/dtype_util.h
+++ b/kernels/portable/cpu/util/dtype_util.h
@@ -31,10 +31,11 @@ using load_to_compute_fn = CTYPE_COMPUTE (*)(const void*);
 
 template <typename CTYPE_COMPUTE, const char* op_name>
 load_to_compute_fn<CTYPE_COMPUTE> get_load_to_compute_fn_realhbbf16(
+    KernelRuntimeContext& context,
     const Tensor& t) {
   CTYPE_COMPUTE (*result)(const void*) = nullptr;
   ET_SWITCH_REALHBBF16_TYPES(
-      t.scalar_type(), unused, op_name, TENSOR_CTYPE, [&]() {
+      t.scalar_type(), context, op_name, TENSOR_CTYPE, [&]() -> void {
         result = internal::load_and_convert<CTYPE_COMPUTE, TENSOR_CTYPE>;
       });
   return result;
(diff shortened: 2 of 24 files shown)


runtime/core/exec_aten/util/scalar_type_util.h
```diff
@@ -910,20 +910,21 @@ struct promote_types {
...
-    }                                                \
+#define ET_INTERNAL_SWITCH(TYPE, CONTEXT, NAME, ...)           \
+  [&] {                                                        \
+    const auto& _st = TYPE;                                    \
+    constexpr const char* et_switch_name = NAME;               \
+    (void)et_switch_name; /* Suppress unused var */            \
+    switch (_st) {                                             \
+      __VA_ARGS__                                              \
+      default:                                                 \
+        CONTEXT.fail(torch::executor::Error::InvalidArgument); \
```
digantdesai, line 921: Nit: Leave a comment along the lines "this sets the return code which should be checked by the caller"

digantdesai (approved):

kernels/portable/cpu/op_fill.cpp
```diff
@@ -90,7 +90,7 @@ Tensor& fill_tensor_out(
   static constexpr const char op_name[] = "fill.Tensor_out";
 
   ET_SWITCH_REALHBBF16_TYPES(a_type, ctx, op_name, CTYPE_A, [&] {
-    CTYPE_A b_casted;
+    CTYPE_A b_casted{};
```
swolchok, line 93: this took me a minute -- we now return all-zeros output and signal a failure when there's an unsupported dtype instead of terminating the process. this seems fine.

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -31,77 +31,98 @@ using load_to_compute_fn = CTYPE_COMPUTE (*)(const void*);
...
+    KernelRuntimeContext& context,
+    const Tensor& t) {
+  if (t.scalar_type() != ScalarType::Bool) {
+    context.fail(torch::executor::Error::InvalidArgument);
+    ET_LOG(
+        Error,
+        "Unhandled dtype %s for %s",
+        ::executorch::runtime::toString(t.scalar_type()),
+        op_name);
+  }
```
swolchok, line 94: Sadly, `load_and_convert` has undefined behavior in this case. We should probably `return CTYPE_COMPUTE();` to be safe.

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -31,77 +31,98 @@ using load_to_compute_fn = CTYPE_COMPUTE (*)(const void*);
...
-      ::executorch::runtime::toString(common_scalar_type),
-      op_name);
+  if (t.scalar_type() != common_scalar_type) {
+    context.fail(torch::executor::Error::InvalidArgument);
+    ET_LOG(
+        Error,
+        "Unhandled dtype %s for %s",
+        ::executorch::runtime::toString(t.scalar_type()),
+        op_name);
+  }
```
swolchok, line 131: ditto undefined behavior

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -124,88 +151,114 @@ template <
...
-      ::executorch::runtime::toString(t.scalar_type()),
-      op_name);
+  if (t.scalar_type() != ScalarType::Bool) {
+    context.fail(torch::executor::Error::InvalidArgument);
+    ET_LOG(
+        Error,
+        "Unhandled dtype %s for %s",
+        ::executorch::runtime::toString(t.scalar_type()),
+        op_name);
+  }
```
swolchok, line 234: ditto undefined behavior; for the `convert_and_store` case it seems easiest to just skip the write to be safe.

kernels/portable/cpu/util/dtype_util.h
```diff
@@ -124,88 +151,114 @@ template <
...
-      ::executorch::runtime::toString(common_scalar_type),
-      op_name);
+  if (t.scalar_type() != common_scalar_type) {
+    context.fail(torch::executor::Error::InvalidArgument);
+    ET_LOG(
+        Error,
+        "Unhandled dtype %s for %s",
+        ::executorch::runtime::toString(t.scalar_type()),
+        op_name);
+  }
```
swolchok, line 273: ditto undefined behavior

kernels/quantized/cpu/embeddingxb.cpp
```diff
@@ -307,9 +309,9 @@ Tensor& quantized_embedding_xbit_out(
     int weight_nbit) {
   // TODO(larryliu): Add a context arg to the real op function and remove this
   // wrapper
-  (void)context;
-  resize_out_tensor(weight, indices, out, weight_nbit);
-  return quantized_embedding_xbit_out(
+  auto context = KernelRuntimeContext();
```
swolchok: nit: `KernelRuntimeContext context;`

PR #13360, opened 2025-08-13
[Backend Tester] Add test flow CLI arg

GregoryComer:
Add a CLI arg to filter by test flow. We have a backend filter, but this allows for selecting only a quantized flow, for example.

Verified with the following. Only the (unquantized) CoreML flow was run.
```
python -m executorch.backends.test.suite.runner operators --filter test_add --flow coreml
```


GregoryComer:
* #13398
* #13390
* __->__ #13360

digantdesai (approved):

PR #13367, opened 2025-08-13
NXP backend: Improve target support checks.

MartinPavella:
### Summary
Improve target specific checks for operator support on Neutron.

### Test plan
Almost all tests utilize the updated functionality, and some unit-tests were updated to reflect the new implementation.

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converter.py
@@ -1,11 +1,10 @@
-# Copyright 2024 NXP
+# Copyright 2024-2025 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Collection
 
 import torch
 
@@ -53,7 +52,6 @@ class NodeConverter(ABC):
     """
 
     context: ConversionContext
-    supported_targets: Collection
 
     def __init__(self, context: ConversionContext):
         self.context = context
@@ -78,25 +76,23 @@ class NodeConverter(ABC):
             Classes which implement conversion for individual operators must overwrite this method.
 
         :param node: torch.Node to check.
+        :param parameters_mapping: Dictionary mapping tensor names to their static data (if they have it).
         """
         pass
 
-    @classmethod
-    def _is_supported_on_target(cls, target: Target) -> bool:
-        """Check if the node is supported on the target platform. It uses the 'supported_platform' attribute, which is
-             a list of supported target platforms, and it must be defined by the specific `NodeConverter`.
+    @staticmethod
+    def _is_supported_on_target(
+        node: Node, target: Target, parameters_mapping: dict[str, Parameter]
+    ) -> bool:
+        """Check if the node is supported on the target platform.
+            Child classes should overwrite this method to implement specific target checks. The default implementation
+            can be used by operators with no target specific requirements.
 
+        :param node: The node (edge operator) to check.
         :param target: Value of the `Target` enum representing the target platform to check for.
+        :param parameters_mapping: Dictionary mapping tensor names to their static data (if they have it).
         """
-        if not (
-            hasattr(cls, "supported_targets")
-            and isinstance(cls.supported_targets, Collection)
-        ):
-            raise NotImplementedError(
-                f"The NodeConverter `{cls}` does not define its `supported_targets` collection."
-            )
-
-        return target == Target.IGNORE or target in cls.supported_targets
+        return target == Target.RT700
 
     @classmethod
     def is_supported(
@@ -110,7 +106,7 @@ class NodeConverter(ABC):
         """
         return cls._is_supported_in_IR(
             node, parameters_mapping
-        ) and cls._is_supported_on_target(target)
+        ) and cls._is_supported_on_target(node, target, parameters_mapping)
 
     @staticmethod
     def _has_shared_q_params_if_quantized(node: Node) -> bool:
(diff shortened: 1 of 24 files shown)


backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -78,25 +76,22 @@ def _is_supported_in_IR(
             Classes which implement conversion for individual operators must overwrite this method.
 
         :param node: torch.Node to check.
+        :param parameters_mapping: Dictionary mapping tensor names to their static data (if they have it).
         """
         pass
 
-    @classmethod
-    def _is_supported_on_target(cls, target: Target) -> bool:
```
JakeStevens, line 85: Why not leave this implementation (or similar, with new args, etc) and override in the super in only the ops with more advanced checks? This adds a lot of unnecessary code bloat for the "simple" ops that are supported fully

MartinPavella: There are a few reasons why we decided for this approach:
1. Consistency between NodeConverers.
2. We are actively improving the target support checks, which is reducing the number of fully supported operators (where the code reduction would apply).
3. We are  currently transitioning into a different system, where instead of using an `enum` for the different targets, we have a class that provides the HW specifics of the target (number of MAC units, number of pipes, ...). This reflects how the `NeutronConverter` works, and it will result in a more robust and a more maintainable implementation. But as a result, we will not be able to simply list the supported targets.

Therefore, the bloat would be avoided only temporarily.

backends/nxp/tests/test_batch_norm_fusion.py
```diff
@@ -201,8 +201,12 @@ def test_batch_norm_linear_fusing__full_pipeline(bias: bool):
...
     # Don't delegate the Linear node, because there seems to be a bug with the NeutronConverter/NeutronPartitioner.
     #  But that doesn't affect the validity of this test.
-    with OverrideSupportedTargets(AddMMConverter, new_targets=[]):
-        with OverrideSupportedTargets(MMConverter, new_targets=[]):
+    with OverrideTargetSupportCheck(
+        AddMMConverter, new_target_support_check=lambda *_: False
+    ):
+        with OverrideTargetSupportCheck(
+            MMConverter, new_target_support_check=lambda *_: False
+        ):
```
digantdesai: readability nit

```suggestion
    def unsupported_target(): return False
    
    with OverrideTargetSupportCheck(
        AddMMConverter, new_target_support_check=unsupported_target
    ):
        with OverrideTargetSupportCheck(
            MMConverter, new_target_support_check=unsupported_target
        ):
```

backends/nxp/backend/ir/converter/node_converter.py
digantdesai: I guess just to removing redundant code when possible.

JakeStevens (approved):
with Digant's nits and fixing the conflicts, LGTM

backends/nxp/tests/test_batch_norm_fusion.py
MartinPavella: The code has been updated according to your suggestion.

backends/nxp/backend/ir/converter/node_converter.py
robert-kalmar: With the transition to the system described in item 3 all the duplicity code in particular Node Converters disappears and this method will return `True` by default here in the base class. 

@Pop-korn , we can transition to this approach already and do the check for the target platform here in the base class: 
```
 @staticmethod
    def _is_supported_on_target(
        node: Node, target: Target, parameters_mapping: dict[str, Parameter]
    ) -> bool:
        match target:
            case Target.RT700:
                return True

            case _:
                return False
```
And remove the redundant code from the particular Node Converters. 

But not in favor the get back the `supported_target` attribute.

MartinPavella: I have implemented the changes. The redundant code has been removed.

backends/nxp/backend/ir/converter/node_converters/ops_converters/softmax_converter.py
```diff
@@ -13,7 +16,16 @@
 
 
 class SoftmaxConverter(NodeConverter):
-    supported_targets = []
+    @staticmethod
+    def _is_supported_on_target(
+        node: Node, target: Target, parameters_mapping: dict[str, Parameter]
+    ) -> bool:
+        match target:
+            case Target.RT700:
```
robert-kalmar, line 24: NIT: a small note, why such an unusual implementation is here (i.e. both cases returns `False`) would be beneficial: 
```
The eIQ Neutron NPU runtime software has a known issue with SoftMax operation. Till the issue is present, return False for i.MXRT700 target also. 
```

robert-kalmar (approved):

backends/nxp/backend/ir/converter/node_converters/ops_converters/softmax_converter.py
MartinPavella: I have added the comment.

robert-kalmar: ✅

robert-kalmar:
Failures seem unrelated: 
- [pull / test-binary-size-linux-gcc / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/17041449310/job/48307140913?pr=13367) - fails on main too
- [pull / unittest-arm-backend-with-no-fvp (test_pytest_ops) / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/17041449310/job/48307141648?pr=13367) - seems unrelated

PR #13368, opened 2025-08-13
NXP backend: Better handling of arg indexes in Neutron quantizer

roman-janik-nxp:
### Summary
Introduces `NodeArgsIdx` class to Neutron Quantizer for better handling of indexes to quantized node's `args` list. Allows selection of nested objects, e.g. an object in a list in node's `args` list. Simplifies `NeutronAtenQuantizer` annotation process by using `annotate_inputs()` for inputs, weights and biases.

### Test plan
All current test that use Neutron Quantizer internally.

### DIFF
--- a/backends/nxp/quantizer/patterns.py
+++ b/backends/nxp/quantizer/patterns.py
@@ -7,7 +7,6 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
-from typing import List, Optional, Tuple, Type, Union
 
 import torch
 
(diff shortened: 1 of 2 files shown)


backends/nxp/quantizer/patterns.py
```diff
@@ -22,11 +21,27 @@
...
+
+
+    Attributes:
+        idx (int): Index to Node's args paramater (list). Selects an input Node or a list of Nodes at the index.
+        inner_idx (int): If specified, index to a list pointed by 'idx' attribute. Selects an input Node at the index.
+                         Default: None.
+    """
+
+    idx: int
+    inner_idx: int = None
```
digantdesai, line 37: Doesn't seem like we are using this (i.e. arg which is a list of nodes) today, Should we add it later? If not, let's add some tests?

StrycekSimon: One of the future PRs uses this feature. You are right, it probably makes more sense to ship it together. Will add it here...

**Edit:** This might take a while. The feature I would be adding relies on changes that were not upstreamed yet. @roman-janik-nxp can you please make this PR a draft for now?

StrycekSimon: @roman-janik-nxp can you please decline/close this PR? I already opened [a new one](https://github.com/pytorch/executorch/pull/14061) for this feature.

PR #13381, opened 2025-08-13
Fix aten.amax lowering issue

cccclai:
Summary:
There was an error when lowering amax around this line `input_tensor = self.get_tensor(input_node, node)` and the issue is that we're trying to permute the tensor inside node_visitors, op_node.meta[QCOM_AXIS_ORDER] is (0, 1), however, tensor.shape is (1, 980, 49).

Rollback Plan:

Differential Revision: D80187368

### DIFF
--- a/backends/qualcomm/tests/models.py
+++ b/backends/qualcomm/tests/models.py
@@ -102,6 +102,24 @@ class AMax(torch.nn.Module):
         return torch.amax(x, dim=self.dim, keepdim=self.keepdim)
 
 
+class AMaxFollowingConv2D(torch.nn.Module):
+    def __init__(
+        self, in_channels, out_channels, kernel_size=3, dim=None, keepdim=False
+    ):
+        super().__init__()
+        self.conv = torch.nn.Conv2d(
+            in_channels, out_channels, kernel_size, padding=kernel_size // 2
+        )
+        self.dim = dim
+        self.keepdim = keepdim
+
+    def forward(self, x):
+        x = self.conv(
+            x
+        )  # Apply convolution (output shape: [batch, out_channels, H, W])
+        return torch.amax(x, dim=self.dim, keepdim=self.keepdim)
+
+
 class AMin(torch.nn.Module):
     def __init__(self, dim=None, keepdim=False):
         super().__init__()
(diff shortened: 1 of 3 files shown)


cccclai:
This PR fix the lowering issue, however, not quite sure a good unit test for it. Try to use the following code but looks like passing without this PR. Do you know what might trigger the error?

```
class ConvAMax2D(torch.nn.Module):
    def __init__(self, in_channels, out_channels, kernel_size=3, dim=None, keepdim=False):
        super().__init__()
        self.conv = torch.nn.Conv2d(in_channels, out_channels, kernel_size, padding=kernel_size//2)  
        self.dim = dim
        self.keepdim = keepdim

    def forward(self, x):
        x = self.conv(x) 
        return torch.amax(x, dim=self.dim, keepdim=self.keepdim)

sample_input = torch.randn(2, 3, 64, 64) 
module = ConvAMax2D(in_channels=3, out_channels=16, kernel_size=3, dim=1, keepdim=False)
output = module(sample_input)
```

Edit:
Actually it does trigger the error, add it in the unit test

cccclai:
Can I get a review on this PR? @haowhsu-quic @shewu-quic @winskuo-quic @DannyYuyang-quic

backends/qualcomm/tests/models.py
```diff
@@ -102,6 +102,18 @@ def forward(self, x):
         return torch.amax(x, dim=self.dim, keepdim=self.keepdim)
 
 
+class ConvAMax2D(torch.nn.Module):
```
haowhsu-quic: It would be great to put it in alphabetical order, thanks!

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -1435,6 +1442,13 @@ def test_qnn_backend_amax(self):
                 module = self.get_qdq_module(module, sample_input)
                 self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_amax_conv(self):
+        sample_input = (torch.randn(2, 3, 64, 64),)  # [batch, channels, height, width]
+        module = AMaxFollowingConv2D(  # noqa: F405
+            in_channels=3, out_channels=16, kernel_size=3, dim=-1, keepdim=False
+        )
+        self.lower_module_and_test_output(module, sample_input)
```
haowhsu-quic, line 1451: I think we should have `module = self.get_qdq_module(module, sample_input)` here.

cccclai: updated

backends/qualcomm/tests/models.py
cccclai: updated

haowhsu-quic (approved):

kirklandsign (approved):

PR #13388, opened 2025-08-13
Remove pinning to ET commit for Zephyr CI job 

BujSet:
### Summary
Currently, the CI jobs responsible for running the Zephyr tests checkout a past version of ExecuTorch, but runs install_executorch.sh with the pinned pytorch commit. Ideally, we would resolve this with resolving the cmake errors that arise from using the most up-to-date version of executorch, but for now, we can just keep the ExecuTorch and PyTorch checkouts consistent for CI. This should prevent the Zephyr CI tests from failing spuriously on unrelated PRs. 

### Test plan


psiddh (approved):

digantdesai:
FYI, it doesn't run zephyr jobs without ciflow/trunk label.

PR #13404, opened 2025-08-14
Arm backend: Add example linkerscripts for U55/U85

perheld:
Add linkerscripts for exector_runner targeting U55/U85 in examples/arm instead of using the linkerscripts from core_platform and then applying patches.

The patches are deleted since they are no longer needed.

Change-Id: I379cf1cd3d9464227fee0555e39847e5b61cc392


examples/arm/executor_runner/CMakeLists.txt
```diff
@@ -143,8 +143,47 @@ target_sources(
                               arm_memory_allocator.cpp
 )
 
-# Include the target's bare-metal linker script
-ethosu_eval_link_options(arm_executor_runner)
+# Check for "U55" in SYSTEM_CONFIG
```
digantdesai, line 146: The idea was to upstream the runner to those repositories and use it from there. Do you know if that's possible?

perheld: I was not aware of that. Let me ask around and see what the plan is.

perheld: It seems there are discussions on where this should be heading but nothing seems decided. Until then we would like a good clean example of a linkerscript (without patch files that might confuse a end user).

zingo (approved):

examples/arm/executor_runner/CMakeLists.txt
zingo: Lets clean this part up for now! Approving PR

digantdesai: Yeah OK with removing the patches. I think landing the runner and related examples in MLEK is the por.

PR #13410, opened 2025-08-14
Qualcomm AI Engine Direct - Fix mem_handel register twice issue

shewu-quic:
Summary:
- Insert registered handle in pre_registered_handles_ map to avoid register multiple times for the same data_ptr

Background:
When running llama in lookahead mode using the same AR-N model for both the prompt processor and token generator. The input and output are the same, and the kv cache is shared between both components.. This causes a "register twice" error message from QNN when a shared buffer (Smart Mask) is used.

Error message:
```
[ERROR] [Qnn ExecuTorch]:  <E> Memory Handle duplicate found, matches Handle ID 0x2

[ERROR] [Qnn ExecuTorch]:  <E> Mem handle exists already

[ERROR] [Qnn ExecuTorch]:  <E> Failed to register memHandles

[ERROR] [Qnn ExecuTorch]:  <E> Failed to register memHandles

[ERROR] [Qnn ExecuTorch]:  <E> Failed to register mem with error 0x1f42
```

cc: @sxu , @haowhsu-quic

### DIFF
--- a/backends/qualcomm/runtime/QnnManager.cpp
+++ b/backends/qualcomm/runtime/QnnManager.cpp
@@ -291,7 +291,8 @@ Error QnnManager::RegisterCustomMem(
           data_ptr,
           unaligned_custom_mem_base,
           total_custom_mem_size,
-          tensor_offset) == Error::Ok,
+          tensor_offset,
+          info) == Error::Ok,
       Internal,
       "Fail to register to shared memory.");
 
--- a/backends/qualcomm/runtime/backends/QnnMemManager.cpp
+++ b/backends/qualcomm/runtime/backends/QnnMemManager.cpp
@@ -56,13 +56,16 @@ Error QnnMemManager::RegisterIonMem(
   return Error::Ok;
 }
 
+// TODO: Find a better way to unify RegisterCustomMem and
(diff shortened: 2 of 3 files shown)


shewu-quic:
Hi @cccclai , @sxu ,

This PR is to fix register multiple times issue when using the same AR-N between both of prompt processor and token generator in lookahead mode.
Could you please help take a look?
Thanks

sxu:
@shewu-quic  Thanks for putting this up. However the situation you described (using same graph for two purposes) is not what we are doing, we have multiple graphs, and externally allocated ION buffers that we are trying to use as input for those graphs. The double registration error we are seeing comes from `RegisterIonMem`. I came up with #13421 which seems to fix the issue for us, however I don't quite understand the what the rest of the memory manager is for (the pre-register and the custom memory stuff), so I don't know if there's a better way to do it.

cccclai:
Seems like there are some ongoing discussion, what do you think regarding this particular PR?

shewu-quic:
> Seems like there are some ongoing discussion, what do you think regarding this particular PR?

I believe this PR is ready for review and merge. The PR it addresses is different from the one that @sxu encountered.

cccclai (approved):

PR #13411, opened 2025-08-14
Qualcomm AI Engine Direct - Remove input_list dependencies

chenweng-quic:
### Summary
- Current SimpleADB push logic require each script to prepare input_list, and the name must follow the logic. Remove all these dependencies.

### Test plan
All existing unit tests should already cover this change.


cccclai:
What's the difference between input and input list?

chenweng-quic:
> What's the difference between input and input list?

Hi @cccclai,
input: list of torch.Tensor that will be converted to raw data and push to device.
input_list: string of content that will be written into disk and push to device.
This PR refactors the current workflow.
Previously, users had to prepare the input_list outside the push function, resulting in duplicated preparation logic across multiple places. This PR aims to eliminate such redundancy by centralizing the input_list handling within the workflow.

cccclai (approved):

PR #13417, opened 2025-08-14
Arm backend: Remove get_output_nodes from runner_utils.

AdrianLundell:
A graph has only one output node containing a list of output tensors. Remove the use of this function to better reflect this.


digantdesai (approved):

zingo:
fails are unrelated

PR #13436, opened 2025-08-14
[ET-VK] Move rotary embedding custom op to be handled via graph pass instead of source transform

SS-JIA:
Summary:
## Motivation

Be able to test Vulkan lowering via optimum-executorch.

## Context

Currently, ET-VK implements rotary embeddings via a custom op. This op is currently inserted into Transformer models by replacing Rotary Embedding modules with a custom module that executes the custom op via a source transform.

The source transform approach makes it cumbersome to lower LLMs to Vulkan, since it requires the export logic to apply the source transform before calling `torch.export()`. This in turn makes it difficult to integrate Vulkan lowering into optimum-executorch, which tries to use a common export + lowering logic for all lowering paths.

As an alternative, leverage `SubgraphMatcher` to detect fusable patterns and fuse the rotary embedding graph pattern into the custom op as part of the Vulkan delegate's graph passes. This removes the requirement to apply a custom source transform just for Vulkan.

## Changes

* Introduce the `backends/vulkan/patterns` folder to store fusable graph patterns
* Introduce a fusable graph pattern for rotary positional embeddings
* Update partitioner logic to automatically include nodes that are part of a fusable graph pattern
* Introduce a pass to fuse known patterns into custom ops / custom op sequence

Test Plan:
Manually tested via exporting and running Llama

Differential Revision: [D80293301](https://our.internmc.facebook.com/intern/diff/D80293301)


SS-JIA:
* #13437
* __->__ #13436

SS-JIA:
@SS-JIA has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D80293301).

PR #13437, opened 2025-08-14
[ET-VK] Enable IntxWeightOnlyConfig

SS-JIA:
Summary:
## Motivation

Be able to test Vulkan lowering via optimum-executorch.

## Context

Very similar to the below PR, Int4 weight only quantization is currently enabled in Vulkan via a custom source transform quantizer that replaces linear layers with a custom linear layer that calls a custom weight only quantized linear op.

This diff aims to make it so that no Vulkan specific source transforms need to be applied by adding a fusion pattern for weight only quantized linear.

## Changes

* Introduce a fusable graph pattern for weight only quantized linear
* Add fusion logic for weight only quantized linear in the fuse patterns pass

Test Plan:
Tested via local export and run llama 3.2 1B

Differential Revision: [D80293302](https://our.internmc.facebook.com/intern/diff/D80293302)

### DIFF
--- a/backends/vulkan/_passes/fuse_patterns.py
+++ b/backends/vulkan/_passes/fuse_patterns.py
@@ -5,11 +5,15 @@
 # LICENSE file in the root directory of this source tree.
 
 import operator
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Tuple
 
 import executorch.backends.vulkan.patterns as vk_patterns
+import executorch.backends.vulkan.utils as utils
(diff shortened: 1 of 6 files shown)


SS-JIA:
* __->__ #13437
* #13436

SS-JIA:
@SS-JIA has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D80293302).

PR #13447, opened 2025-08-15
fix mismatch sub dtype

cccclai:
Differential Revision: D80312352


shewu-quic (commented):
Thank you very much. I hadn't noticed this pass. Could you let me know when it was applied?

cccclai:
> Thank you very much. I hadn't noticed this pass. Could you let me know when it was applied?

It's running during to_edge stage https://github.com/pytorch/executorch/blob/ecb639a6c22f8cac6a779c976a668f4f318c2edc/exir/program/_program.py#L684

SS-JIA (approved):

PR #13456, opened 2025-08-15
Arm backend: Add decomposition and test for fmod

emmakujala:
Add decomposition and test for fmod


backends/arm/test/ops/test_fmod.py
```diff
@@ -0,0 +1,255 @@
...
+def test_fmod_tensor_vgf_INT(test_data: Tuple):
+    module, data_x, data_y = test_data()
+    pipeline = VgfPipeline[input_t](
+        module,
+        (data_x, data_y),
+        [],
+        [],
+        tosa_version="TOSA-1.0+INT",
+    )
+    pipeline.run()
```
digantdesai, line 255: Nice tests! 👍

digantdesai (approved):

PR #13464, opened 2025-08-15
support etdump generation in executorch.runtime

Gasoonjia:
Summary: make et.runtime support etdump generation.

Differential Revision: D80373976


mergennachin (changes requested):
Add unit test

lucylq (approved):
Lgtm, looks like test failures are old (from 5 days ago), may need to re-export?

Assume the test works locally, if it's disabled because we're unable to set the etdump flag, could you create an issue to track that and add a comment to the test?

PR #13467, opened 2025-08-17
Link QNN backend to pybinding lib when built

GregoryComer:
This change makes it possible to run QNN models with pybindings, when enabled. It does still require the environment setup described in the docs.

I validated this change locally by running install_executorch.sh with CMAKE_ARGS="-DEXECUTORCH_BUILD_QNN=ON", and then lowering and running a simple add model. The CI set up further in this stack also leverage this feature and test it extensively.


GregoryComer:
* #13535
* #13534
* #13469
* __->__ #13467

digantdesai (commented):
I guess rest of the pybind code was already present?

GregoryComer:
> I guess rest of the pybind code was already present?

When all of the required libraries and environment variables are set up, it will work with only this change. I would like to further simplify, as there is a lot of setup required - including some copying around of python extension binaries. But this code here is enough to make it runnable in conjunction with the existing QNN CI scripts (which do all of the setup).

cccclai:
> as there is a lot of setup required - including some copying around of python extension binaries

Can you elaborate?

GregoryComer:
> > as there is a lot of setup required - including some copying around of python extension binaries
> 
> Can you elaborate?

Yeah, it looks like it's currently building the Python extension components separately and then copying them into the ET directory. (See https://docs.pytorch.org/executorch/main/backends-qualcomm.html#aot-ahead-of-time-components). As part of the build cleanup and pybind enablement, I'll see if I can fold that into our "main" pybind build. That will also make it just work with installation, wheels, etc.

cccclai:
> > > as there is a lot of setup required - including some copying around of python extension binaries
> > 
> > 
> > Can you elaborate?
> 
> Yeah, it looks like it's currently building the Python extension components separately and then copying them into the ET directory. (See https://docs.pytorch.org/executorch/main/backends-qualcomm.html#aot-ahead-of-time-components). As part of the build cleanup and pybind enablement, I'll see if I can fold that into our "main" pybind build. That will also make it just work with installation, wheels, etc.

They should be part of `./backends/qualcomm/scripts/build.sh`. After you run this script, does pybind work?

GregoryComer:
> > > > as there is a lot of setup required - including some copying around of python extension binaries
> > > 
> > > 
> > > Can you elaborate?
> > 
> > 
> > Yeah, it looks like it's currently building the Python extension components separately and then copying them into the ET directory. (See https://docs.pytorch.org/executorch/main/backends-qualcomm.html#aot-ahead-of-time-components). As part of the build cleanup and pybind enablement, I'll see if I can fold that into our "main" pybind build. That will also make it just work with installation, wheels, etc.
> 
> They should be part of `./backends/qualcomm/scripts/build.sh`. After you run this script, does pybind work?

Sort of. It does expect that you're running scripts from the root directory of executorch (or have an editable install), since it puts the extension binaries in the source directory. So it won't work if you're using ET from another directory, such as a dependent project. It should ideally put these binaries wherever the installed copy of the executorch python package is.

If no objection, I'm happy to take this task as part of the build workstream. Basically, running install_executorch.sh with qnn enabled will automatically do what that build.sh script does as part of the main install, in order to make it easy.

digantdesai (approved):

cccclai (approved):
Thanks for enabling pybinding + QNN

.ci/scripts/test_backend_linux.sh
```diff
@@ -25,6 +25,6 @@ source .ci/scripts/setup-vulkan-linux-deps.sh
 EXTRA_BUILD_ARGS="-DEXECUTORCH_BUILD_VULKAN=ON"
 
 # We need the runner to test the built library.
-PYTHON_EXECUTABLE=python CMAKE_ARGS="$EXTRA_BUILD_ARGS" .ci/scripts/setup-linux.sh --build-tool cmake --build-mode Release
+PYTHON_EXECUTABLE=python CMAKE_ARGS="$EXTRA_BUILD_ARGS" .ci/scripts/setup-linux.sh --build-tool cmake --build-mode Release --editable true
```
digantdesai, line 28: can you remind me why this was necessary?

PR #13460, opened 2025-08-15
Arm backend: Add missing using-declaration in VGFBackend.cpp

YufengShi-dudu:
The PR https://github.com/pytorch/executorch/pull/13004 moved `EValue` to the `Span` namespace but did not add the corresponding using-declaration in VGFBackend.cpp.

Adds the missing using-declaration to fix the build.

Change-Id: I58f1acff3fec41aee433283495ff4bbf6cb34512

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13471, opened 2025-08-18
Arm backend: Add partial vulkan runtime support for VgfPipeline

YufengShi-dudu:
- Add XfailIfNoVKMLEmulationLayer for VGF unit tests
- Introduce run_target_board() to unify FVP and VKML execution paths
- Implement run_vkml_emulation_layer() with NotImplementedError for output parsing, as the VGF runtime doesn't dump the output tensors in a usable format at the moment.

Change-Id: I3d936b737a694d90971869289b16875d6f6e55a4

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218 @SS-JIA @manuelcandales @cbilgin

### DIFF
--- a/backends/arm/test/ops/test_add.py
+++ b/backends/arm/test/ops/test_add.py
@@ -7,6 +7,7 @@
 
 from typing import Tuple
 
+import pytest
 import torch
 from executorch.backends.arm.quantizer import arm_quantizer
 from executorch.backends.arm.test import common, conftest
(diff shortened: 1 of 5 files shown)


mansnils (approved):

backends/arm/test/ops/test_add.py
```diff
@@ -187,9 +188,19 @@ def test_add_tensor_u85_INT_2(test_data: input_t2):
 
 @common.parametrize("test_data", Add.test_data)
 @common.SkipIfNoModelConverter
+@common.XfailfNoVKMLEmulationLayer
```
digantdesai, line 191: There is still value, from test coverage point of view, in running the tests on those platforms. Just we can't run the PTE.

YufengShi-dudu: Thanks @digantdesai. We have seen backend broken by upstream changes due to the missing GitHub CI https://github.com/pytorch/executorch/pull/13460
We will upstream soon to enable build and test.
cc: @mansnils

PR #13485, opened 2025-08-18
Add check_for_installed_private_headers_in_cmake_out

swolchok:
This adds a test that will run in CI as part of unittest jobs to make sure that we don't install private headers.

I fixed the existing places where we were installing private headers; reviewers, please confirm that we are OK with this technical break of source-level backward compatibility.


swolchok:
* #13588
* #13559
* #13558
* __->__ #13485

swolchok:
noting CI looks good

CMakeLists.txt
```diff
@@ -485,24 +485,29 @@ install(
...
   PATTERN "*.h"
+  PATTERN "testing_util" EXCLUDE
 )
 install(
   DIRECTORY runtime/executor/
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/executorch/runtime/executor
   FILES_MATCHING
   PATTERN "*.h"
+  PATTERN "test" EXCLUDE
+  PATTERN "platform_memory_allocator.h" EXCLUDE
```
JacobSzwejbka, line 496: I need to follow up with Greg on what this even is used for.

JacobSzwejbka (approved):
Only one that is maybe used would be testing_util but I really hope no one would be using that lol. The rest seem totally fine bc breaks.

PR #13493, opened 2025-08-18
Add Model Profiling Automation Script

leafs1:
### Summary
Added profile_model.sh to automate the ExecutorTorch model profiling workflow. The script streamlines building executor_runner with profiling enabled, running model inference with ETDump collection, and generating CSV profiling reports. It accepts model and ETDump paths as arguments with sensible defaults, consolidating what was previously a multi-step manual process into a single executable script.

### Test plan
Manually ran script on llama3.pte and confirmed csv and dump generation


devtools/scripts/profile_model.sh
```diff
@@ -0,0 +1,33 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+#!/bin/bash
+
+set -e
+
+echo "Building executor_runner with profiling enabled..."
+
+cmake --preset profiling
```
GregoryComer: Can you check to ensure it's building in release mode? It won't by default, unless there's some magic in the presets or similar. You can just add `-DCMAKE_BUILD_TYPE=Release`.

devtools/scripts/profile_model.sh
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright 2024-25 Arm Limited and/or its affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+#!/bin/bash
```
GregoryComer, line 8: Can you add a comment stating how the script is intended to be invoked, and that it expects to be run from the top-level executorch directory?

devtools/scripts/profile_model.sh
GregoryComer: Does this line need to specify the directory to build in? Like -B cmake-out? Wondering if this will place the build artifacts in the directory the script is run from.

devtools/scripts/profile_model.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -e
+
+echo "Building executor_runner with profiling enabled..."
+
+cmake --preset profiling
+cmake --build cmake-out --target executor_runner
```
GregoryComer: What do you think about changing the build directory to build-profiling or something? Reason being that this might end up reconfiguring the user's build if they're following our instructions and use cmake-out as the build directory. So it might be nice to put this build in a different directory.

GregoryComer:
Thanks for adding this! It looks great. I left a few quick comments.

devtools/scripts/generate_profiling_csv.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def generate_csv(etdump_path, output):
+
+    inspector = Inspector(etdump_path)
+    df = inspector.to_dataframe()
+    df.to_csv(output)
+
+
+def main():
```
Gasoonjia, line 26: Do you mind to add some comments in the python script for demo the usage?

GregoryComer (approved):
Looks great. This will be very helpful for GA model profiling.

PR #13495, opened 2025-08-18
Chang the tmp directory where  to-be-compiled models are stored to use the assetsDirectoryPath

metascroy:
This PR changes the temporary directory where to-be-compiled models are stored to use assetsDirectoryPath (NSCachesDirectory) instead of trashDirectory (NSTemporaryDirectory).

This is to prevent errors when the temp directory is deleted between when the model to-be-compiled is written to disk, and it is compiled.

It also cleans up the modelURL if compilation finishes.

The compiled asset is later moved to a permanent location.


metascroy:
cc @cymbalrush can you have a look?

PR #13496, opened 2025-08-18
Fix method meta error in wasm build

Conarnar:
### Summary
An error would occur when trying to get the metadata for a method containing an input or output that isn't a tensor. This occurs because there wasn't a check for the input/output tag when generating the list of tensor metadata.

Now the list of tensor metadata will have `undefined` if the input/output in that index is not a tensor.

### Test plan
Added unit tests to test the method metadata for a method that has ints in its input and outputs.
Added unit test to check if getMethod works if the module has multiple methods.

Unit tests are in the CI but can be ran with
```
bash scripts/build_wasm_tests.sh
cd cmake-out-wasm/extension/wasm/test/
npm test # after installing Jest
```

### DIFF
--- a/extension/wasm/wasm_bindings.cpp
+++ b/extension/wasm/wasm_bindings.cpp
@@ -475,16 +475,26 @@ struct ET_EXPERIMENTAL JsMethodMeta {
         val::array(),
         meta.num_instructions()};
     for (int i = 0; i < meta.num_inputs(); i++) {
-      js_array_push(new_meta.input_tags, meta.input_tag(i).get());
-      js_array_push(
-          new_meta.input_tensor_meta,
-          JsTensorInfo::from_tensor_info(meta.input_tensor_meta(i).get()));
+      Tag tag = meta.input_tag(i).get();
+      js_array_push(new_meta.input_tags, tag);
+      if (tag == Tag::Tensor) {
+        js_array_push(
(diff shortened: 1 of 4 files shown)


extension/wasm/wasm_bindings.cpp
```diff
@@ -475,16 +475,26 @@ struct ET_EXPERIMENTAL JsMethodMeta {
...
-          new_meta.output_tensor_meta,
-          JsTensorInfo::from_tensor_info(meta.output_tensor_meta(i).get()));
+      Tag tag = meta.output_tag(i).get();
+      js_array_push(new_meta.output_tags, tag);
+      if (tag == Tag::Tensor) {
+        js_array_push(
+            new_meta.output_tensor_meta,
+            JsTensorInfo::from_tensor_info(meta.output_tensor_meta(i).get()));
+      } else {
+        js_array_push(new_meta.output_tensor_meta, val::undefined());
```
lucylq, line 496: Should this return an error?

Conarnar: The entire MethodMeta object is constructed when getMethodMeta is called rather than returning only the specific information requested like in the C++ APIs. The user would get an error if they try to index it and attempt to read from it though.

JacobSzwejbka (approved):

PR #13511, opened 2025-08-19
Arm backend: Add support for QAT+per-channel combo

oscarandersson8218:
Fixes a bug where an unsupported observer was used for QAT combined with per-channel quantization.

### DIFF
--- a/backends/arm/quantizer/arm_quantizer.py
+++ b/backends/arm/quantizer/arm_quantizer.py
@@ -101,17 +101,19 @@ def get_symmetric_quantization_config(
     weight_observer_or_fake_quant_ctr: ObserverOrFakeQuantizeConstructor = (
         MinMaxObserver
     )
+
     # Determine the right observer/fake-quant constructor
     if is_qat:
-        # Set plain fake-quant with true min/max
-        weight_observer_or_fake_quant_ctr = FakeQuantize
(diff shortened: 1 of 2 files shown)


zingo (approved):

backends/arm/quantizer/arm_quantizer.py
```diff
@@ -101,18 +101,20 @@ def get_symmetric_quantization_config(
     weight_observer_or_fake_quant_ctr: ObserverOrFakeQuantizeConstructor = (
         MinMaxObserver
     )
+
     # Determine the right observer/fake-quant constructor
     if is_qat:
-        # Set plain fake-quant with true min/max
-        weight_observer_or_fake_quant_ctr = FakeQuantize
+        if is_per_channel:
+            weight_observer_or_fake_quant_ctr = PerChannelMinMaxObserver
```
digantdesai, line 108: Don't we need a FakeQuantize variant of this observer with qat?

oscarandersson8218: Yes, you're right. Will look into a fix and proper testing of QAT.

oscarandersson8218: Fixed here https://github.com/pytorch/executorch/pull/14001

PR #13513, opened 2025-08-19
Update deprecated pkg_resources api in edge._ops

Erik-Lundell:
pkg_resources is deprecated and set to be removed from setuptools as early as end of 2025. Since ET doesn't have an upper bound on the setuptools dependency this could be an issue. Additionally, this outputs a warning message everytime an executor_runner is built.

This change updates the pkg_resrouces api to the more modern importlib.resources

Test plan:
Build linux wheels using ci.
Download python 3.10 and wheel and install in new venv
run
```
from executorch.exir.dialects._ops import ops as exir_ops
from executorch.backends.apple.mps import serialization
from executorch.backends.vulkan import serialization
from executorch.backends.xnnpack import serialization
from executorch.devtools.bundled_program import serialize
from executorch.devtools.etdump import serialize
from executorch.examples.models import checkpoint
checkpoint.get_default_model_resource_dir("examples/models/llama")
from executorch.exir.serde import schema_check
from executorch.extension.flat_tensor.serialize import serialize
```
in a fresh python interpreter for every import
The imports should not crash (due to importlib), and should not generate the warning message that it currently does

### DIFF
--- a/backends/apple/mps/serialization/mps_graph_serialize.py
+++ b/backends/apple/mps/serialization/mps_graph_serialize.py
@@ -1,14 +1,16 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import importlib.resources as _resources
 import json
 import os
 import tempfile
 
-import pkg_resources
+import executorch.backends.apple.mps.serialization as serialization_package
 from executorch.backends.apple.mps.serialization.mps_graph_schema import MPSGraph
 from executorch.exir._serialize._dataclass import _DataclassEncoder
 from executorch.exir._serialize._flatbuffer import _flatc_compile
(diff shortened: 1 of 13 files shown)


zingo:
@digantdesai This one is not in Arm folders, so we let someone at Meta review it.

digantdesai:
I guess this is OK. Couple of points,
* make sure this also works with wheels (editable or otherwise)
* moving to importlib is OK, but I still see many uses of pkg_resources in, do you want to take a stab at those too? 🙏 
* And ofc make sure CI is green (as green as it can be w/ unrelated failures :()

Erik-Lundell:
> I guess this is OK. Couple of points,
> 
>     * make sure this also works with wheels (editable or otherwise)
> 
>     * moving to importlib is OK, but I still see many uses of pkg_resources in, do you want to take a stab at those too? 🙏
> 
>     * And ofc make sure CI is green (as green as it can be w/ unrelated failures :()

Working on updating all occurrences. What's the best way to test that it works with wheels?

digantdesai:
> What's the best way to test that it works with wheels?

We have a CI job https://github.com/pytorch/executorch/actions/runs/17097212094/job/48484358746 which builds wheels.

You can either do same steps locally or by manually triggering the build job on this PR and then downloading the wheel and testing locally.

Erik-Lundell:
Testing Linux wheels here
https://github.com/pytorch/executorch/actions/runs/17266659354

Erik-Lundell:
I have tested according to the testplan and it was ok, what do you think @digantdesai

digantdesai:
Thanks LGTM, make sure the ci is clean.

digantdesai (approved):

Erik-Lundell:
New failed macos job ran successfully here: https://github.com/pytorch/executorch/actions/runs/17371877730/job/49323743831.

PR #13516, opened 2025-08-19
Arm backend: Annotate ADD/SUB with indepenedent observers

### LINKED ISSUE #12959: Incorrect Observer Sharing/Derivation at Conv-ReLU+ Residual with Arm Ethos Quantizer
### 🐛 Describe the bug

There's a bug in PyTorch 2.0 quantization with ARM quantizer where observers are incorrectly getting shared/derived across different operations (add, permute, relu). Reason most likely being **add requires inputs to share observer which when merged with relu and permute observer sharing results in this bug**. Since the ReLU observer is shared and observes negative values, incorrect quantization parameters(qparams) are computed (zp!=-128). The Relu fusion is impacted [as discussed here](https://github.com/pytorch/executorch/pull/7555#issuecomment-3127997727)  which impacts model delegation to ARM U55 backend.

Compare quantized model and see qparams:
With bug triggered:

```python
class SampleModel(nn.Module):
    def __init__(self, in_channels, out_channels, kernel_size, input_dim):
        super().__init__()
        self.conv1 = nn.Conv1d(in_channels, out_channels, kernel_size)
        self.relu = torch.nn.ReLU()
        self.linear = nn.Linear(out_channels, out_channels)

    def forward(self, x):
        # Permute: (N, T, C) -> (N, C, T)
        x = x.permute(0, 2, 1)
        # Conv1D
        x = self.conv1(x)
        # ReLU
        x = self.relu(x)
        # Permute back: (N, C, T) -> (N, T, C)
        x = x.permute(0, 2, 1)
        # Residual connection
        x = x + self.linear(x)
        return x

# Input shape: (1, 10, 3)
model = SampleModel(in_channels=3, out_channels=16, kernel_size=3, input_dim=10)
# Apply post-training quantization with ARM U55 recipe
quantized_model = post_train_quantize(model, calibration_data, recipe="AR_CERES_U55_ET_RECIPE")
```

With bug not triggered:
```python
class SampleModel(nn.Module):
    def __init__(self, in_channels, out_channels, kernel_size, input_dim):
        super().__init__()
        self.conv1 = nn.Conv1d(in_channels, out_channels, kernel_size)
        self.relu = torch.nn.ReLU()
        self.linear = nn.Linear(out_channels, out_channels)

    def forward(self, x):
        # Permute: (N, T, C) -> (N, C, T)
        x = x.permute(0, 2, 1)
        # Conv1D
        x = self.conv1(x)
        # ReLU
        x = self.relu(x)
        # Permute back: (N, C, T) -> (N, T, C)
        x = x.permute(0, 2, 1)
        # Residual connection
        x = self.linear(x)
        return x

# Input shape: (1, 10, 3)
model = SampleModel(in_channels=3, out_channels=16, kernel_size=3, input_dim=10)
# Apply post-training quantization with ARM U55 recipe
quantized_model = post_train_quantize(model, calibration_data, recipe="AR_CERES_U55_ET_RECIPE")
```

Although strange that this hasn't been uncovered with other residual networks, but IMO we need to reconsider certain Shared and Derived quantization spec and optimizations to fix this bug. This is a blocker and if there's no immediate resolution, discussion on short term fixes will also be meaningful. 
cc: @digantdesai , @3l1, @Tessil 

### Versions

Python version: 3.12.11+meta (3.12:55fee9c, Jun 03 2025, 15:41:33) [Clang 19.1.2 
Python platform: Linux-5.19.0-0_fbk12_hardened_11583_g0bef9520ca2b-x86_64-with-glibc2.34
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

gggekov:
We were annotating the ADD/SUB with a shared observer resulting in the same quantisation parameters on the two inputs even if we were adding  numbers in different ranges(positive tensor to a tensor with positive and negative values). As a result,
the quantisation parameters were suboptimal. This change annotates the operator with independent observers and changes how we rescale the two inputs to bring them to the same range. Added a unit test of a resnet model. Lowered the number of channels on a few unit tests in order to keep the Total SRAM Used below 2MB for the Ethos-U55 to fit within the memory limit of the Corstone-300.

Fixes #12959


backends/arm/operators/op_add.py
```diff
@@ -85,7 +86,7 @@ def define_node(
             # Scale output back to 8 bit
             # pyre-ignore
             tqutils.insert_rescale_op_to_int8(
-                tosa_graph, add_output, scale_back, node, self.tosa_spec
+                tosa_graph, add_output, scale_back, node, False, self.tosa_spec
```
digantdesai: Readability nit 
```suggestion
                tosa_graph, add_output, scale_back, node, compute_rescale=False, self.tosa_spec
```

backends/arm/test/ops/test_conv_combos.py
```diff
@@ -47,28 +47,28 @@ def __init__(self):
         # (t, c, n, s) = (6, 96, 1, 1)
         # 1. 1x1 CONV2d + ReLU6 (Pointwise)
         self.pointwise_conv2d = torch.nn.Conv2d(
-            in_channels=32, out_channels=128, kernel_size=1, stride=1, groups=1
+            in_channels=16, out_channels=96, kernel_size=1, stride=1, groups=1
```
digantdesai, line 50: why this change?

backends/arm/test/ops/test_sub.py
```diff
@@ -48,6 +55,16 @@
 }
 fvp_sub2_xfails = {"rand_4D_2x2x4x4": "MLETORCH-517 : Multiple batches not supported"}
 
+# Sub and tan - the tan has a really steep curve just before Pi/2 and a point of discontinuity at Pi/2
```
digantdesai, line 55: ❤️

backends/arm/tosa_quant_utils.py
```diff
@@ -15,11 +15,69 @@
...
+    get_input_qparams,
+    get_output_qparams,
+)
+
 from executorch.backends.arm.tosa_mapping import TosaArg
 from torch.fx import Node
 from tosa.RoundingMode import RoundingMode  # type: ignore
 
 
+def insert_rescale_ops_to_int32_for_add_sub(
```
digantdesai: Nit why the fn name should have `_for_add_sub` suffix?

digantdesai (approved):

backends/arm/tosa_quant_utils.py
gggekov: The insert_rescale_ops_to_int32_for_add_sub function is only called for the ADD & SUB ops because only for these operators we use a common scale of 2max(scale_A,scale_B) and then multiply the original scale by 1<<20 without overflowing in a 32-bit accumualator.

backends/arm/test/ops/test_conv_combos.py
gggekov: When the ADD has different scales on the two inputs, the memory footprint for the U55 increases(only on U55, not on U85). To give you an example on a real model, on resnet18 on u55 with shared observer on the ADD, the peak tensor size is 1MB and with independent observer - 1.7MB. I've flagged this to the compiler team. The Corstone-300 has 2MB of SRAM so I lowered slightly the number of channels so that the peak intermediate tensor fits in memory.

backends/arm/tosa_quant_utils.py
digantdesai: I mean the function name shouldn't list its call sites :)

backends/arm/operators/op_add.py
gggekov: Will do, thanks. We have a long weekend in the UK, will be back on Tuesday.

backends/arm/test/ops/test_conv_combos.py
digantdesai: Thanks.

backends/arm/operators/op_add.py
gggekov: Now fixed.

backends/arm/tosa_quant_utils.py
gggekov: Fixed :)

zingo:
Rebased after a fix for some broken arm tests was merged

zingo:
Test fails are unrelated

zingo (approved):

zingo (approved):

zingo:
@digantdesai I cant merge this, is there a older version of this PR internally blocking this?

PR #13525, opened 2025-08-19
ASR JNI bindings

rohansjoshi:
### Summary

Added JNI bindings to run Qualcomm Whisper Runner in Android

### Test plan

CI tests


extension/android/CMakeLists.txt
```diff
@@ -69,8 +69,19 @@ set_target_properties(
 
 executorch_target_link_options_shared_lib(executorch)
 
-add_library(
-  executorch_jni SHARED jni/jni_layer.cpp jni/log.cpp jni/jni_layer_runtime.cpp
+add_library(executorch_jni SHARED jni/jni_layer.cpp jni/log.cpp jni/jni_layer_runtime.cpp jni/jni_layer_whisper.cpp)
```
kirklandsign: For whisper part, maybe add a flag?

extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/audio/WhisperCallback.java
```diff
@@ -0,0 +1,31 @@
...
+import org.pytorch.executorch.annotations.Experimental;
+
+/**
+ * Callback interface for Whisper model. Users can implement this interface to receive the generated
+ * tokens and statistics.
+ *
+ * <p>Warning: These APIs are experimental and subject to change without notice
+ */
+@Experimental
+public interface WhisperCallback {
```
mergennachin: Why is WhisperCallback in core module extensions/android??

mergennachin (changes requested):

extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/audio/WhisperCallback.java
cccclai: It's still wip, we're trying to get things working. I'm marking it as draft for now

GregoryComer:
I'm a little worried about maintaining model-specific APIs. Would it be possible to build it under examples or similar?

There is definitely a need to provide these sort of task-specific driver APIs, though it might be nice to have a more general AudioToTextRunner (or something similar). If we do land this as is, can we make sure it's clear that it's unstable and should not be used in production? Even if it is marked experimental in code, if it's under the Android extension, there is an implicit assumption that it will be supported.

My concern is that we'll eventually have a more general task-specific runner and will want to migrate to that without hitting BC guarantees. Having this Whisper-specific code under examples would potentially discourage production use until then.

@jackzhxng @larryliu0820 How would you expect Whisper to fit into our runner strategy in the long term?

rohansjoshi:
> I'm a little worried about maintaining model-specific APIs. Would it be possible to build it under examples or similar?
> 
> There is definitely a need to provide these sort of task-specific driver APIs, though it might be nice to have a more general AudioToTextRunner (or something similar). If we do land this as is, can we make sure it's clear that it's unstable and should not be used in production? Even if it is marked experimental in code, if it's under the Android extension, there is an implicit assumption that it will be supported.
> 
> My concern is that we'll eventually have a more general task-specific runner and will want to migrate to that without hitting BC guarantees. Having this Whisper-specific code under examples would potentially discourage production use until then.
> 
> @jackzhxng @larryliu0820 How would you expect Whisper to fit into our runner strategy in the long term?

Right now for LLMs the API is we have Java class LlmModule which wraps a C++ class which stores a pointer to a IRunner runner, and in the constructor we specify which runner to use. So one idea could be to have a ASRModule Java class wrap a C++ class which has a runner which implements some interface we could call AudioToTextRunner. Then the Whisper Runner in examples/qualcomm could implement this interface.

extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/audio/WhisperCallback.java
rohansjoshi: It is analogous to LlmCallback which is also in extension/android (also I renamed it from WhisperCallback to ASRCallback), in the Android app you can use the callback to e.g. print generated tokens to the screen one at a time

jackzhxng:
Ideally we can fit Whisper into our multimodal runner strategy but it's not super high priority at the moment. I think it's fine to have an ASR module though, but I agree we shouldn't have anything Whisper specific in there

extension/android/jni/jni_layer_asr.cpp
```diff
@@ -0,0 +1,175 @@
...
+        ::executorch::extension::cpuinfo::get_num_performant_cores() - 1;
+    if (num_performant_cores > 0) {
+      ET_LOG(Info, "Resetting threadpool to %d threads", num_performant_cores);
+      ::executorch::extension::threadpool::get_threadpool()
+          ->_unsafe_reset_threadpool(num_performant_cores);
+    }
+#endif
+
+    // create runner
+    runner_ = std::make_unique<example::WhisperRunner>(
```
jackzhxng: This should be an interface, like ASRRunner or something like that

jackzhxng (approved):

extension/android/jni/jni_layer_asr.cpp
kirklandsign: Does it mean this is not runnable without QNN?

kirklandsign (approved):

scripts/build_android_library.sh
```diff
@@ -43,6 +43,7 @@ build_android_native_library() {
     -DEXECUTORCH_BUILD_EXTENSION_LLM_RUNNER="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
     -DEXECUTORCH_BUILD_EXTENSION_TRAINING=ON \
     -DEXECUTORCH_BUILD_LLAMA_JNI="${EXECUTORCH_BUILD_EXTENSION_LLM:-ON}" \
+    -DEXECUTORCH_BUILD_ASR_JNI="${EXECUTORCH_BUILD_EXTENSION_AUDIO:-ON}" \
```
mergennachin: Can we just use EXECUTORCH_BUILD_EXTENSION_AUDIO everywhere instead of  having another flag EXECUTORCH_BUILD_ASR_JNI?

When we have iOS/Swift bindings, we would have  to do another EXECUTORCH_BUILD_ASR_IOS etc

I'd rather just use EXECUTORCH_BUILD_EXTENSION_AUDIO everywhere not have per-platform granularity

This is consistent with how we build LLM runner

mergennachin: Also in the top-level CMakeLists.txt, add something like this:


```
if(EXECUTORCH_BUILD_EXTENSION_AUDIO)
  message(STATUS "Audio/ASR extension enabled")
endif()
```

extension/llm/runner/asr_runner.h
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param seq_len Length of input sequence
+   * @param inputs A vector containing one element: a vector of bytes that
+   * encodes a float tensor in little-endian byte order
+   * @param token_callback Callback function called for each generated token
+   * @return Error::Ok if successful, an error otherwise
+   */
+  virtual runtime::Error transcribe(
+      int32_t seq_len,
+      std::vector<std::vector<char>>& inputs,
```
mergennachin: Can we use this data structure? 

https://github.com/pytorch/executorch/blob/4a4f5a034b6cda395abe7b6aa13b8066b776ffb2/extension/llm/runner/audio.h#L24

extension/llm/runner/asr_runner.h
```diff
@@ -0,0 +1,61 @@
+/*
```
mergennachin: why is this in extension/llm? 

you already have extension/audio

extension/llm/runner/asr_runner.h
```diff
@@ -0,0 +1,61 @@
...
+   */
+  virtual bool is_loaded() const = 0;
+
+  /**
+   * Load the model and prepare for inference.
+   *
+   * @return Error::Ok if successful, an error otherwise
+   */
+  virtual runtime::Error load() = 0;
+
```
mergennachin: How about reset function?

extension/llm/runner/asr_runner.h
```diff
@@ -0,0 +1,61 @@
...
+#include <executorch/runtime/core/error.h>
+
+namespace executorch {
+namespace extension {
+namespace llm {
+
+class ET_EXPERIMENTAL ASRRunner {
+ public:
+  virtual ~ASRRunner() = default;
+
```
mergennachin: Constructor for `ASRRunner` should contain configs such as temperature and other configs.

mergennachin: Also what about taking tokenizer and module as part of the constructor?

extension/android/jni/jni_layer_asr.cpp
```diff
@@ -0,0 +1,180 @@
...
+        ::executorch::extension::cpuinfo::get_num_performant_cores() - 1;
+    if (num_performant_cores > 0) {
+      ET_LOG(Info, "Resetting threadpool to %d threads", num_performant_cores);
+      ::executorch::extension::threadpool::get_threadpool()
+          ->_unsafe_reset_threadpool(num_performant_cores);
+    }
+#endif
+#if defined(EXECUTORCH_BUILD_QNN)
+    // create runner
+    runner_ = std::make_unique<example::WhisperRunner>(
```
mergennachin: top-level ASR component shouldn't depend on example Whisper

examples/qualcomm/oss_scripts/whisper/runner/runner.cpp
```diff
@@ -27,20 +27,20 @@ static constexpr auto kDecoderStartTokenId = "decoder_start_token_id";
 static constexpr auto kEosId = "get_eos_id";
 static constexpr auto kMaxContextLen = "get_max_context_len";
 } // namespace
-Runner::Runner(
+WhisperRunner::WhisperRunner(
```
mergennachin, line 30: You should refactor out general purpose ASR logic out of this file and put into asr_runner.h/cpp

mergennachin:
@kirklandsign @jackzhxng -- Why did we approve this PR? This should be reviewed carefully, as we're adding first-class citizen APIs to our extension.

Also, cc @larryliu0820 - please keep an eye on these kinds of PRs

mergennachin (changes requested):

extension/android/jni/jni_layer.cpp
```diff
@@ -508,6 +508,12 @@ class ExecuTorchJni : public facebook::jni::HybridClass<ExecuTorchJni> {
 };
 } // namespace executorch::extension
 
+#ifdef EXECUTORCH_BUILD_ASR_JNI
+extern void register_natives_for_asr();
+#else
+void register_natives_for_asr() {}
+#endif
+
```
larryliu0820: Semi related to the diff, I think we started to recognize a plugin pattern here, it would be better if we can rely on static initialization to register these natives and just iterate through them in line 535. This way we don't need all these macros, we just link the necessary library and it register itself.

extension/android/jni/jni_layer_asr.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <executorch/examples/qualcomm/oss_scripts/whisper/runner/runner.h>
+#endif
+
+#include <fbjni/ByteBuffer.h>
+#include <fbjni/fbjni.h>
+
+using ::executorch::runtime::Error;
+
+namespace {
+bool utf8_check_validity(const char* str, size_t length) {
```
larryliu0820: can this be extracted into a util?

extension/llm/runner/asr_runner.h
larryliu0820: It depends on if the implementation is managing the KV cache position. I think this interface should enforce that, ie, if we decide to manage the position inside the runner we should have a protected field for pos and a reset() API

extension/llm/runner/asr_runner.h
rohansjoshi: Most of the configs (temperature, echo, etc) which are encapsulated in the GenerationConfig struct for LLM runners aren't relevant for ASR. ASRRunner is supposed to be a general interface like IRunner so I didn't put any constructor here

kirklandsign:
Hi @rohansjoshi 

Now we have https://github.com/pytorch/executorch/pull/16979/changes
and https://github.com/pytorch/executorch/tree/main/examples/models/whisper
and https://github.com/pytorch/executorch/blob/914d5ffaa81b332948996160d8476027722997ce/extension/asr/runner/runner.h#L60

Do we need this one?

larryliu0820:
@kirklandsign I think it's safe to close this PR

kirklandsign:
Our current code should work out of box for QNN as well right? Nothing special needed for QNN

kirklandsign:
I don't see QNN specific part

PR #13526, opened 2025-08-19
Summary: Add ExecutorchRuntimeException:  Throw relevant exceptions from JNI layer in the event of errors

psiddh:
Add ExecutorchRuntimeException

Test Plan: Tested with executorch-examples and llama android demo.

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan


extension/android/jni/BUCK
```diff
@@ -28,7 +28,7 @@ non_fbcode_target(_kind = executorch_generated_lib,
 
 non_fbcode_target(_kind = fb_android_cxx_library,
     name = "executorch_jni",
-    srcs = ["jni_layer.cpp", "log.cpp", "jni_layer_runtime.cpp"],
+    srcs = ["jni_layer.cpp", "log.cpp", "jni_layer_runtime.cpp", "jni_helper.cpp"],
```
JacobSzwejbka: can you refactor these srcs into a list that can be shared across the libs

psiddh: Done

extension/android/jni/jni_layer.cpp
```diff
@@ -438,9 +451,17 @@ class ExecuTorchJni : public facebook::jni::HybridClass<ExecuTorchJni> {
...
-          "Cannot get load module");
+      // Format hex string
+      std::stringstream ss;
+      ss << "Cannot get load module [Native Error: 0x" << std::hex
+         << std::uppercase << static_cast<uint32_t>(names_result.error())
+         << "]";
+
+      jni_helper::throwExecutorchException(
+          static_cast<uint32_t>(
+              Error::InvalidArgument), // For backward compatibility
```
GregoryComer, line 462: From looking at user code, I'm a bit less worried about about strict BC here, so I think you could use the actual error code and not worry about matching the existing exception type.

GregoryComer (commented):
Overall, looks great to me. Thanks for taking this. The only thing I wanted to get your thoughts on is whether we should update the Java definition in Module.java to indicate that it throws. I'm not as familiar with some of the Java conventions, but I believe we can add a checked exception specification to the Java declarations of the JNI methods to indicate that they can throw an ExecutorchRuntimeException.

From what I've seen, opinion seems to be divided on checked exceptions. I don't have a strong opinion either way, and it would potentially require some synchronous code changes to land internally. The advantage might be that it's more clear in the signature that various methods (Module.load / execute) can throw an ExecutorchRuntimeException.

GregoryComer:
You're probably already be planning this, but could you add/modify an integration test to cover thrown exceptions?

extension/android/jni/jni_layer.cpp
psiddh: Agree, it is always better to throw the exception with native error code as-is. I wasn't sure 100% abt it, but since you checked the client code, I will make the change

GregoryComer (approved):
Talked a bit with Sid offline. Sounds like unchecked exceptions are the better option here, which is good with me.

psiddh:
> From

If the native side (JNI C++ code) throws a RuntimeException or IllegalArgumentException (which is what we are doing now), these are unchecked exceptions in Java afaik.
- Unchecked exceptions do not need to be declared in the Java method signature with throws.
- It is considered good practice to omit throws for unchecked exceptions because:
- These exceptions represent programming errors or unexpected conditions.Declaring them clutters the API without adding compile-time safety.

psiddh:
> You're probably already be planning this, but could you add/modify an integration test to cover thrown exceptions?

Yes was planning on a follow up change

PR #13528, opened 2025-08-19
Fix Olmo trunk test and skip XNNPack trunk tests on mac

jackzhxng:
Unblock trunk to advance `viable/strict`

Intermittent segmentation faults on all of the XNNPack Optimum mac tests. Described in more detail [here](https://github.com/pytorch/executorch/pull/13530) where the segmentation fault happens consistently, but interestingly enough this is before the pin bump and it is only happening intermittently.


metascroy (approved):

metascroy:
@jackzhxng it doesn't look like the XNNPACK tests are being skipped on trunk

PR #13529, opened 2025-08-19
Fix remaining instances of `for aten in (True, False):`

swolchok:
These should be using get_aten_mode_options because ATen mode is disabled in OSS.


swolchok:
* __->__ #13529

lucylq (approved):

PR #13533, opened 2025-08-19
Add 1MB temp allocator for executor

zonglinpeng:
Summary:
Method::load creates a PlatformMemoryAllocator as fallback when no temp allocator is provided. So our KernelRuntimeContext will always be PlatformMemoryAllocator since no temp mem is allocated. Overrides et_pal_allocate to allocate

code pointer: https://fburl.com/code/216qnnvt

Differential Revision: D80578578


eigen-k (approved):

jackzhxng (commented):
@zonglinpeng fix merge conflicts

PR #13538, opened 2025-08-20
Whisper audio processor

rohansjoshi:
Summary:
Running Whisper to convert audio -> text consists of two steps: 
(1) Audio preprocessor (aka Mel spectrogram feature extractor), and 
(2) Whisper model (encoder+decoder)
Currently, in examples/qualcomm/oss_scripts/whisper, we have a flow for exporting Whisper encoder+decoder (2) and running it on device. It can take in Mel spectrogram tensors as input and produce text output

**This class implements part (1), the audio processing stage, in PyTorch.** It is equivalent to HuggingFace WhisperFeatureExtractor (which computes Mel spectrograms with NumPy). It takes in an audio waveform at 16KHz (as a 1D tensor) and outputs Mel spectrograms that can be fed directly to the Whisper model (2).

The script (see test plan) compares this class agains WhisperFeatureExtractor, they have a very small numerical discrepancy (<1e-5).

Differential Revision: D80215714

### DIFF
--- a/extension/audio/TARGETS
+++ b/extension/audio/TARGETS
@@ -1,0 +1,28 @@
+load("@fbcode_macros//build_defs:python_library.bzl", "python_library")
+load("@fbcode_macros//build_defs:python_binary.bzl", "python_binary")
+load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime")
+
+oncall("executorch")
+
+python_library(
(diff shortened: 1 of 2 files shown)


cccclai (approved):
Please update the output file in the unit test and it's all good

mergennachin:
cc @msluszniak @chmjkb @mkopcins FYI the audio preprocessing part can be now just exported

PR #13542, opened 2025-08-20
Make IOManager use Module instead of Method.

shoumikhin:
Summary: Let's not expose Method from Module so that it's not getting misused beyond its owner.

Differential Revision: D80595261

### DIFF
--- a/extension/llm/runner/io_manager/io_manager.h
+++ b/extension/llm/runner/io_manager/io_manager.h
@@ -8,12 +8,8 @@
 
 #pragma once
 
-#include <vector>
-
+#include <executorch/extension/module/module.h>
 #include <executorch/extension/tensor/tensor.h>
-#include <executorch/runtime/core/exec_aten/exec_aten.h>
-#include <executorch/runtime/executor/method.h>
-#include <executorch/runtime/executor/method_meta.h>
 
 namespace executorch {
 namespace extension {
@@ -30,6 +26,13 @@ namespace llm {
 class ET_EXPERIMENTAL IOManager {
  public:
   /**
+   * @brief Construct an IOManager bound to a Module.
+   *
+   * @param module The Module used for querying method metadata and execution.
+   */
+  explicit IOManager(ET_MODULE_NAMESPACE::Module& module) : module_(module) {}
+
+  /**
    * @brief Virtual destructor to allow proper cleanup in derived classes.
    */
   virtual ~IOManager() = default;
@@ -38,18 +41,26 @@ class ET_EXPERIMENTAL IOManager {
    * @brief Load the IO manager with method metadata for prefill and
    * decode operations.
    *
-   * @param program The program prefill and decode methods are loaded from.
    * @param prefill_method The prefill method to initialize with.
    * @param decode_method The decode method to initialize with.
    */
   ET_NODISCARD virtual runtime::Error load(
-      const executorch::ET_RUNTIME_NAMESPACE::Program& program,
-      executorch::ET_RUNTIME_NAMESPACE::Method& prefill_method,
-      executorch::ET_RUNTIME_NAMESPACE::Method& decode_method) {
-    (void)program;
+      const std::string& prefill_method,
+      const std::string& decode_method) {
     (void)prefill_method;
     (void)decode_method;
     return runtime::Error::Ok;
+  }
+
+  /**
+   * @brief Load the IO manager using the default method names.
+   *
+   * Uses "forward" for both prefill and decode.
+   *
+   * @return Error code.
+   */
+  ET_NODISCARD runtime::Error load() {
+    return load("forward", "forward");
   }
 
   /**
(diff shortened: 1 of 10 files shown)


extension/llm/runner/io_manager/io_manager.h
```diff
@@ -144,13 +176,21 @@ class ET_EXPERIMENTAL IOManager {
...
     return runtime::Error::Ok;
   }
+
+  ET_NODISCARD runtime::Error update_decode(
+      const std::vector<runtime::EValue>& model_outputs) {
+    return update_decode(model_outputs, "forward");
+  }
+
+private:
+  ET_MODULE_NAMESPACE::Module& module_;
```
mergennachin, line 243: Can it be `const ET_MODULE_NAMESPACE::Module& module_;`?

extension/llm/runner/io_manager/io_manager.h
```diff
@@ -29,6 +25,10 @@ namespace llm {
  */
 class ET_EXPERIMENTAL IOManager {
  public:
+
+  explicit IOManager(ET_MODULE_NAMESPACE::Module& module)
+  : module_(module) {}
```
mergennachin: IOManager holds a reference but doesn't ensure Module outlives it.

Perhaps use shared_ptr<Module> or weak_ptr<Module> 
```
  explicit IOManager(std::shared_ptr<Module> module) : module_(module) {}
  // Or

  explicit IOManager(std::weak_ptr<Module> module) : module_(module) {
    ET_CHECK_MSG(!module.expired(), "Module must be valid");
  }
```

extension/llm/runner/io_manager/io_manager.h
```diff
@@ -38,88 +38,115 @@ class ET_EXPERIMENTAL IOManager {
...
-   * @param program The program prefill and decode methods are loaded from.
    * @param prefill_method The prefill method to initialize with.
    * @param decode_method The decode method to initialize with.
    */
   ET_NODISCARD virtual runtime::Error load(
-      const executorch::ET_RUNTIME_NAMESPACE::Program& program,
-      executorch::ET_RUNTIME_NAMESPACE::Method& prefill_method,
-      executorch::ET_RUNTIME_NAMESPACE::Method& decode_method) {
-    (void)program;
+      const std::string& prefill_method,
```
mergennachin, line 48: Check the prefill_method and decode_method methods exists in module. If there's a typo in the method name, it will be hard to catch it.

extension/llm/runner/io_manager/io_manager.h
mergennachin: Also should you test if module is loaded in the constructor?

extension/llm/runner/io_manager/io_manager.h
```diff
@@ -38,88 +38,115 @@ class ET_EXPERIMENTAL IOManager {
...
   /**
    * @brief Reset the IO manager state.
    *
    * @param prefill_method The prefill method to reset with.
    * @param decode_method The decode method to reset with.
    */
   ET_NODISCARD virtual runtime::Error reset(
-      executorch::ET_RUNTIME_NAMESPACE::Method& prefill_method,
-      executorch::ET_RUNTIME_NAMESPACE::Method& decode_method) {
+      const std::string& prefill_method,
```
mergennachin, line 73: also do validation on the method names

mergennachin (changes requested):

mergennachin:
Also fix the unit test in executorch/extension/llm/runner/test:test_text_decoder_runner

extension/llm/runner/io_manager/io_manager.h
shoumikhin: Normally Module is not supposed to be shared, thus nobody actually wraps it in a shared_ptr, and using weak_ptr is possible only of there's a shared_ptr somewhere.

Passing a reference for now is by design to convey the Module must stay valid while IOManager works with it.

If the Module is not loaded, it will be loaded lazily when IOManager works with it.

extension/llm/runner/io_manager/io_manager.h
shoumikhin: Module will return an error if IOManager tries to work with a Method that doesn't exist, so it's caught at the Module level.

extension/llm/runner/io_manager/io_manager.h
shoumikhin: IOManager may potentially call some Module methods that are not const I suspect.

mergennachin: Which methods is not const?

shoumikhin: Those that may potentially load or execute a Method

mergennachin: But within IOManager I don't see module_ being mutated.

If it compiles with const reference, then shouldn't it be good? Happy to be wrong, but I'm not convinced just yet.

shoumikhin: This particular `IOManager` is supposed to be a base class, so subclasses can totally load methods if needed. And this one already loads `Program` when querying for `Method` metadata (`method_meta()` is not const).

mergennachin (approved):

PR #13549, opened 2025-08-20
NXP backend: Add support for conversion of Conv1D operator

roman-janik-nxp:
### Summary
This PR adds delegation of `aten.conv1d` to Neutron. Fixes `input_shapes` type hint in `to_quantized_edge_program()`. Fixes `operators_not_to_delegate` assignment in partitioner.

### Test plan
Unit tests provided in backends/nxp/tests/ir/converter/node_converter/test_conv_converter.py.

### DIFF
--- a/backends/nxp/backend/ir/converter/conversion/common.py
+++ b/backends/nxp/backend/ir/converter/conversion/common.py
@@ -70,29 +70,22 @@ def try_get_input(t_op: tflite_model.Operator, idx: int) -> tflite_model.Tensor 
     return tensor
 
 
-def extend_1d_pads_to_2d(onnx_1d_pads: MutableSequence):
-    """Extend the onnx 'pads' operator attribute that represents padding for a 1D kernel to 2D, by adding '0's."""
-    if onnx_1d_pads is not None:
-        onnx_1d_pads.insert(1, 0)
-        onnx_1d_pads.append(0)
+def extend_1d_padding_to_2d(tflite_1d_padding: MutableSequence):
+    """Extend the PyTorch 'padding' operator attribute that represents padding for a 1D kernel to 2D, by adding '0's."""
+    if tflite_1d_padding is not None:
+        tflite_1d_padding.append(0)
 
 
-def extend_1d_strides_to_2d(onnx_1d_strides: MutableSequence):
-    """Extend the onnx 'strides' operator attribute that represents strides for a 1D kernel to 2D, by adding '1'."""
-    if onnx_1d_strides is not None:
-        onnx_1d_strides.append(1)
+def extend_1d_stride_to_2d(tflite_1d_stride: MutableSequence):
+    """Extend the PyTorch 'stride' operator attribute that represents stride for a 1D kernel to 2D, by adding '1'."""
+    if tflite_1d_stride is not None:
+        tflite_1d_stride.append(1)
 
 
-def extend_1d_dilations_to_2d(onnx_1d_dilations: MutableSequence):
-    """Extend the onnx 'dilations' operator attribute that represents dilations for a 1D kernel to 2D, by adding '1'."""
-    if onnx_1d_dilations is not None:
-        onnx_1d_dilations.append(1)
-
-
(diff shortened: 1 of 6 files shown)


JakeStevens (changes requested):
How does this intersect with #13576? Seems you are extending with 0s, I think this will re-introduce this bug for conv1d now.

Can use the same checks introduced in the tests in #13576 for 1D case

roman-janik-nxp:
> How does this intersect with #13576? Seems you are extending with 0s, I think this will re-introduce this bug for conv1d now.
> 
> Can use the same checks introduced in the tests in #13576 for 1D case

Yes, this is a conflict. I added fix for padding with zero-point to this PR. In our IR, Conv1D operator doesn't exist and it is emulated by changing to Conv2D and then the result is converted to Conv1D format. So this fix will change it also for Conv2D. However, the other PR is still needed as it changes Average pool.

backends/nxp/neutron_partitioner.py
```diff
@@ -297,15 +297,10 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
         qdq_cluster_recognizer = QDQClusterRecognizer()
         qdq_cluster_recognizer.tag_qdq_clusters(nodes)
         graph_module.recompile()
+        target = self.delegation_spec[1][2].value
```
robert-kalmar: This change (using `spec.key` instead of magic number indexes was introduced in https://github.com/pytorch/executorch/pull/10196#discussion_r2057342276 (difference from our code base) for better readability. See also this diff https://github.com/pytorch/executorch/compare/b102be24be4c0ee802e33e3c1467a6648910b20c..21ca220d2c3b6006e8080f765360936e9a005c78. 

Please revert back to use the `spec.key` for better readability.

robert-kalmar (changes requested):

robert-kalmar (approved):

robert-kalmar:
@JakeStevens , Roman updated based on your findings. Can you please re-review. 
The PR cannot be merged without you closing the "change request".

backends/nxp/backend/ir/converter/conversion/common.py
```diff
@@ -70,29 +70,22 @@ def try_get_input(t_op: tflite_model.Operator, idx: int) -> tflite_model.Tensor
...
 
-def extend_1d_pads_to_2d(onnx_1d_pads: MutableSequence):
-    """Extend the onnx 'pads' operator attribute that represents padding for a 1D kernel to 2D, by adding '0's."""
-    if onnx_1d_pads is not None:
-        onnx_1d_pads.insert(1, 0)
-        onnx_1d_pads.append(0)
+def extend_1d_padding_to_2d(tflite_1d_padding: MutableSequence):
+    """Extend the PyTorch 'padding' operator attribute that represents padding for a 1D kernel to 2D, by adding '0's."""
+    if tflite_1d_padding is not None:
+        tflite_1d_padding.append(0)
```
JakeStevens, line 76: This is my specific concern. We are padding with zeros not zero point, just like the 2d case previously

JakeStevens:
Sorry, I think my comment was misunderstood. I was not concerned that this doesn't *include* the changes from the 2D padding PR-- both will get into mainline eventually, this is fine and makes perfect sense.

instead, I believe this PR has the same bug; in the conversion 1d case, we are extending the existing padding by adding a zero, instead of zero point. see inline comment

backends/nxp/backend/ir/converter/conversion/common.py
roman-janik-nxp: This value express the amount of padding applied to the input, not the padding value. So zero here means that the tensor will not be padded on H dimension as this is conversion from 1D tensor NWC to 2D NHWC tensor - padding also needs to be extended/converted. The amount of padding for W dim is kept.
The padding value is the zero-point added on L352 and L390 in convolution_converter.py.

JakeStevens: Got it, thanks!

JakeStevens (approved):

PR #13550, opened 2025-08-20
Cortex_m backend: Loosen edge op check.

Erik-Lundell:
The pass checked that all ops were edge ops to detect if the pass was ran before lowering to edge. However, there are cases where aten ops survive after edge lowering, notably torch.ops.tensor_scalar. This shouldn't crash the pass. Instead, only check that q/dq ops are edge ops.


digantdesai (approved):

Erik-Lundell:
@digantdesai is the buck issue (`At least one of the pre-conditions you specified did not hold`) due to the removed import in the pass? Is it ok to merge, or if not, what do I change in the TARGET file to make it work.

digantdesai:
seems like S3 related, rerunning it..

PR #13558, opened 2025-08-20
install headers from runtime/backend

swolchok:
Looks like these are all public and we simply missed installing them.


swolchok:
* #13588
* #13559
* __->__ #13558
* #13485

CMakeLists.txt
```diff
@@ -480,6 +480,13 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/configurations)
 
 # Install `executorch` library as well as `executorch-config.cmake` under
 # ${CMAKE_INSTALL_PREFIX}/
+install(
```
JacobSzwejbka, line 483: why do we need one for basically every runtime subdir as opposed to just one runtime/?

swolchok: I wanted to keep the exclude patterns narrowly scoped.

JacobSzwejbka (approved):

PR #13559, opened 2025-08-20
Switch from print_exported_headers to CMake's installed headers in scripts/build_apple_frameworks.sh

swolchok:
This removes the need to use Buck in this script. The previous PR (#13485) adds enforcement that we don't install non-public headers from CMake.

I built the frameworks before and after. Here is a diff of all headers
present in the frameworks:
https://gist.github.com/swolchok/70647b551da7827a9cbdb083fad1e209

Generally, we now include more headers than before. The only header we
used to include that we no longer include is
schema/extended_header.h. It looks like we are not including the rest
of the headers for the PTE schema anyway, so I propose that we fix
installation of the schema headers separately.


swolchok:
* #13588
* __->__ #13559
* #13558
* #13485

shoumikhin:
There are a lot of headers for the code not shipped with the executorch [runtime framework](https://github.com/pytorch/executorch/blob/main/scripts/build_apple_frameworks.sh#L28-L34), so anyone using them by mistake will run into missing symbols error.
E.g. `extension/llm` headers technically belong to a different framework `executorch_llm` (although we currently don't expose them in favor of the native ObjC/Swift APIs).
Let's try to keep the set of headers we ship to the minimum and make sure we do have the corresponding implementation in .a we ship. Ideally, nothing should change.

swolchok:
> There are a lot of headers for the code not shipped with the executorch [runtime framework](https://github.com/pytorch/executorch/blob/main/scripts/build_apple_frameworks.sh#L28-L34), so anyone using them by mistake will run into missing symbols error. E.g. `extension/llm` headers technically belong to a different framework `executorch_llm` (although we currently don't expose them in favor of the native ObjC/Swift APIs). Let's try to keep the set of headers we ship to the minimum and make sure we do have the corresponding implementation in .a we ship. Ideally, nothing should change.

OK; we can filter the set of headers we're installing in the script. Might be a week before I can get back to this though.

PR #13560, opened 2025-08-20
[Core ML] Improve asset management

cymbalrush:
## Fix asset management issues in CoreML delegate

- Do not use the temporary directory for asset storage.
- Use the staging directory for storing temporary assets, and ensure it is cleaned up after use.
- Perform aggressive cleanup of the temporary directory to avoid leftovers.
- Guarantee proper cleanup of staging directories after operations to prevent stale files.


backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
```diff
@@ -254,6 +254,18 @@ BOOL is_asset_alive(NSMapTable<NSString *, ETCoreMLAsset *> *assets_in_use_map,
     
     return assets;
 }
+
+NSURL * _Nullable move_to_directory(NSURL *url,
```
metascroy, line 258: This looks similar to the pattern "[self moveURL:dstURL toUniqueURLInDirectory:self.trashDirectoryURL error:nil];"

Should we consolidate?

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
```diff
@@ -254,6 +254,18 @@ BOOL is_asset_alive(NSMapTable<NSString *, ETCoreMLAsset *> *assets_in_use_map,
...
+                                    NSURL *directoryURL,
+                                    NSFileManager *fileManager,
+                                    NSError * __autoreleasing *error) {
+    NSURL *dstURL = [directoryURL URLByAppendingPathComponent:[NSUUID UUID].UUIDString];
+    if (![fileManager moveItemAtURL:url toURL:dstURL error:error]) {
```
metascroy, line 273: Do we need to check if url is nil when using moveItemAtURL?

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
```diff
@@ -299,12 +311,20 @@ - (nullable instancetype)initWithDatabase:(const std::shared_ptr<Database>&)data
...
+
     NSURL *managedTrashDirectoryURL = ::create_directory_if_needed(trashDirectoryURL, @"models", fileManager, error);
     if (!managedTrashDirectoryURL) {
         return nil;
     }
-    
+
+    // Remove any existing contents in the staging directory by moving them to the trash directory,
+    // then recreate a clean staging directory for new use.
+    move_to_directory([assetsDirectoryURL URLByAppendingPathComponent:@"staging"], managedTrashDirectoryURL, fileManager, nil);
```
metascroy: What happens if we have two models?

In model2, the staging directory is being used during compilation, but then in model1 we execute this line and move the staging directory?

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
```diff
@@ -433,16 +460,25 @@ - (nullable ETCoreMLAsset *)_storeAssetAtURL:(NSURL *)srcURL
...
+            NSError *localError = nil;
+            if (![weakSelf _compact:self.maxAssetsSizeInBytes error:&localError]) {
+                ETCoreMLLogError(localError, "Failed to compact asset store.");
+            }
+        });
     }
-    
+
+    // Always clean the trash directory to ensure a minimal footprint.
+    // The `trashQueue` is serialized, so only one cleanup will run at a time.
```
metascroy, line 475: What happens if something is moving to the trashDirectory when it's being deleted by another thread?

For example, withTemporaryDirectory moves staging files to the trashDirectory after compilation, but then another model triggers  cleanupTrashDirectory on another thread.

backends/apple/coreml/runtime/delegate/ETCoreMLModelLoader.mm
```diff
@@ -62,21 +62,12 @@ + (nullable ETCoreMLModel *)loadModelWithContentsOfURL:(NSURL *)compiledModelURL
...
     }
-    
-    if (localError) {
-        ETCoreMLLogError(localError,
-                         "Failed to load model from compiled asset with identifier = %@",
-                         identifier);
-    }
-    
-    // If store failed then we will load the model from compiledURL.
-    auto backingAsset = Asset::make(compiledModelURL, identifier, assetManager.fileManager, error);
```
metascroy, line 73: Why are we removing this?

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
```diff
@@ -430,78 +435,134 @@ - (nullable NSURL *)compiledModelURLWithIdentifier:(NSString *)identifier
                                       "AOT blob is missing model file.");
         return nil;
     }
-    
-    NSURL *dstURL = [self.assetManager.trashDirectoryURL URLByAppendingPathComponent:[NSUUID UUID].UUIDString];
-    NSURL *modelURL = ::write_model_files(dstURL, self.fileManager, identifier, modelAssetType.value(), inMemoryFS, error);
+
+    // If modelURL is not provided, write model files to the destination directory (dstURL)
+    // and obtain a URL pointing to them. Otherwise, use the provided modelURL.
+    modelURL = (modelURL == nil) ? ::write_model_files(dstURL, self.fileManager, identifier, modelAssetType.value(), inMemoryFS, error) : modelURL;
```
metascroy, line 441: When would modelURL be non-null?  Is it mainly the case where the model is already compiled in the PTE?

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
```diff
@@ -430,78 +435,134 @@ - (nullable NSURL *)compiledModelURLWithIdentifier:(NSString *)identifier
...
                                                           maxWaitTimeInSeconds:(5 * 60)
                                                                          error:error];
-            
+            // Return the URL of the compiled model or nil if compilation fails.
             return compiledModelURL;
         }
     }
 }
 
-#if ET_EVENT_TRACER_ENABLED
```
metascroy, line 453: We are we removing the compile flag here?

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
```diff
@@ -430,78 +435,134 @@ - (nullable NSURL *)compiledModelURLWithIdentifier:(NSString *)identifier
...
+
+    return compiledModelAsset;
+}
+
+#if ET_EVENT_TRACER_ENABLED
+- (nullable ETCoreMLAsset *)modelAssetWithMetadata:(const ModelMetadata&)metadata
+                                        inMemoryFS:(const inmemoryfs::InMemoryFileSystem*)inMemoryFS
+                                             error:(NSError * __autoreleasing *)error {
+    NSString *identifier = @(metadata.identifier.c_str());
+    NSString *rawIdentifier = raw_model_identifier(identifier);
```
metascroy, line 505: What does raw here mean?  Is it to distinguish from the compiled asset (which gets the unqualified identified as its name)?

cymbalrush: Yes

backends/apple/coreml/runtime/delegate/ETCoreMLModelLoader.mm
cymbalrush: If store fails then it's an error there is no point recovering from it.

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
cymbalrush: That should be fine `trashQueue` is serial and we get the list of directories to delete at the start of the method

```
  NSDirectoryEnumerator *enumerator = [fileManager enumeratorAtURL:self.trashDirectoryURL
                                          includingPropertiesForKeys:@[]
                                                             options:NSDirectoryEnumerationSkipsSubdirectoryDescendants
                                                        errorHandler:errorHandler];
```

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
cymbalrush: ETCoreMLAssetManager is only created once when the backend is loaded, and this happens at`init` time.

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
cymbalrush: In the case of Analyzer we also save the raw model and we don't want to write the model files again if it already exists.

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
cymbalrush: No, moving it.

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
metascroy: ETCoreMLAssetManager is created once for a single PTE file on delegate init, but what if there are multiple PTE files?

Do they all share the same staging directory?  And can that lead to issues when one tries to delete the staging directory when another is compiling from it?

cymbalrush: No `ETCoreMLAssetManager` is only created once - https://github.com/pytorch/executorch/blob/main/backends/apple/coreml/runtime/delegate/backend_delegate.mm#L138

cymbalrush: `ETCoreMLAssetManager ` and `ETCoreMLModelManager` are only created once. You can put breakpoints to verify it.

metascroy: What if I load two ET programs on two separate processes? Won't each process create ETCoreMLModelManager?

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
cymbalrush: I added checks.

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
cymbalrush: Updated to use `move_to_directory `

backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm
cymbalrush: We need a persistent identifier, but I haven’t identified a suitable one yet. For now, I have removed the deletion step. Although we clean the directories in the staging area, if the app is terminated before the cleanup occurs, stale directories may remain. However, since these directories are located in the caches directory, the OS will automatically clean them up when storage is low.

metascroy: Sounds good.  I guess we could clean up things in staging that have last touched time stamps > 1 day during asset manager creation.

metascroy:
Overall changes look great @cymbalrush 

Running internal/external CI.  It does look like the linter is failing.  Can you run lint?

From ET repo:
```
pip install -r requirements-lintrunner.txt
```

And then run:
```
lintrunner -a
```

metascroy (approved):

metascroy:
Attemping land from phabricator

PR #13564, opened 2025-08-20
Fix data loader build on Windows, re-enable pybind job

GregoryComer:
Add mman_windows.cpp to the CMake sources. With this change, the pybind preset works and we can re-enable it in CI.


GregoryComer:
* #13567
* #13667
* __->__ #13564
* #13260
* #13259
* #13672

GregoryComer:
Note that CI failures are pre-existing.

tools/cmake/preset/pybind.cmake
```diff
@@ -21,16 +21,28 @@ set_overridable_option(EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR ON)
...
 )
-  # Windows or other OS-specific code here
+  if(NOT CMAKE_GENERATOR_TOOLSET MATCHES "ClangCL")
+    message(
+      FATAL_ERROR
+        "ExecuTorch requires the ClangCL toolset on Windows. Please configure with -T ClangCL."
+    )
+  endif()
+
+  # These XNNPACK options don't currently build on Windows with Clang.
```
JacobSzwejbka: Where do these magic incantations come from

extension/data_loader/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ if(NOT ET_HAVE_SYS_MMAN_H AND NOT WIN32)
        "extension/data_loader/mmap_data_loader.cpp"
   )
 endif()
+if(WIN32)
```
JacobSzwejbka: Why is this necessary? I thought scott already selectively included this source file in build whatever.bzl

JacobSzwejbka: Apparently not. I might be thinking of selectively removing the header from the public include if not on windows or something

JacobSzwejbka (approved):

tools/cmake/preset/pybind.cmake
GregoryComer: These are 4 optional XNNPACK features (exposed by XNNPACK's CMakeLists) that clang-cl has issues with due to the /vlen compiler argument.

These disable a small subset of the available kernels, though they are relatively niche and should have minimal perf impact. It should be fixed or gated upstream in XNNPACK itself. I intend to file an issue there and maybe just fix it myself. They have some gating logic already (https://github.com/google/XNNPACK/blob/923253023555f75c38d96511aa7fa59b9ef9c25c/CMakeLists.txt#L176)

I could also move these into the XNNPACK backend CMakeLists, rather than the presets, as they are functionally an internal detail, and it would also ensure that they are always disabled on Windows.

GregoryComer: Thinking through it more, I think I will move these into the XNNPACK backend CMakeLists, rather than the preset.

extension/data_loader/CMakeLists.txt
GregoryComer: Yeah, I chatted with Scott briefly on this. His recommendation was to manage it locally. The main build variables list has to be mechanically parsable from both CMake and Buck, so it can't include more complex conditional logic.

GregoryComer:

PR #13565, opened 2025-08-20
Create stale.yml workflow to label stale PRs

nil-is-all:
The behavior is:
- If a PR is not labeled stale, after 60 days inactivity label the PR as stale and comment about it.
- If a PR is labeled stale, after 30 days inactivity close the PR.
- `high priority` and `no-stale` PRs are exempt.

### DIFF
--- a/.github/workflows/stale
+++ b/.github/workflows/stale
@@ -93,30 +93,30 @@ jobs:
               if (labels.includes("Stale")) {
                 core.info(`[${pull.number}] Closing PR.`);
                 numAPIRequests += 1;
-                await github.rest.issues.update({
-                  owner: "pytorch",
-                  repo: "executorch",
-                  issue_number: pull.number,
-                  state: "closed",
+                  //await github.rest.issues.update({
+                  //owner: "pytorch",
+                  //repo: "executorch",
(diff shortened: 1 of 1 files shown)


GregoryComer (approved):
Thanks. Looks good to me.

GregoryComer:
@nil-is-all Actually, I have one thought. Can you change the cron schedule to run daily? That way, if it errors out, it won't spam. Once we know it's running successfully, we can bump it up to hourly.

nil-is-all:
> @nil-is-all Actually, I have one thought. Can you change the cron schedule to run daily? That way, if it errors out, it won't spam. Once we know it's running successfully, we can bump it up to hourly.

Done, changed it to hourly now.

GregoryComer:
Thanks. Thinking through it, out of an abundance of caution, can you also comment out the sections that update PRs? That will let us validate the behavior before it starts actually labeling or closing issues. I'll check the logs for the workflow run, and once it looks good, we can put up another PR to enable the actual behavior.

Specifically, lines 96-101, 107-112, and 115-120. It looks like you can either prefix each line with `//` or surround the block with `/* */`.

PR #13566, opened 2025-08-21
Added JS bindings for tokenizers library

Conarnar:
### Summary
Added JavaScript bindings for the tokenizer library so that we can use them to run LLMs in a web browser.

### Test plan
I will add end to end tests later.


CMakeLists.txt
```diff
@@ -871,6 +871,10 @@ if(EXECUTORCH_BUILD_WASM)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/extension/wasm)
 endif()
 
+if(EXECUTORCH_BUILD_TOKENIZERS_WASM)
```
JacobSzwejbka, line 874: I wonder if we have too many cmake options lmao.

extension/wasm/tokenizers/tokenizers.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+ private:
+  std::unique_ptr<T> tokenizer_;
+};
+
+} // namespace
+
+EMSCRIPTEN_BINDINGS(TokenizerModule) {
+#define JS_BIND_TOKENIZER(NAME)                                           \
+  class_<JsTokenizer<NAME>>(#NAME)                                        \
```
JacobSzwejbka, line 160: Can you measure the codesize of this lib. I'm curious how bad it is having to individually bind each template instantiation

JacobSzwejbka (commented):
I know you've tested this in the demo repo and the tokenizer unit tests are a bit of a pain to set up because they are a submodule.

mergennachin:
What are the pros-and-cons between putting here vs https://github.com/meta-pytorch/tokenizers library directly?

extension/wasm/tokenizers/tokenizers.cpp
Conarnar: 102K js, 1.4M wasm
Changing JsTokenizer to not be a template and only need to bind one class reduces the js to 101K.
Removing everything except SentencePiece leaves it with 99K js, 807K wasm so it seems like most of the codesize comes from including each of the different types of tokenizer.

Conarnar:
> What are the pros-and-cons between putting here vs https://github.com/meta-pytorch/tokenizers library directly?

Pros:
- Puts all the Wasm modules together.

Cons:
- Separated from the rest of the Tokenizers libraries, tests, resources, etc.

JacobSzwejbka:
> What are the pros-and-cons between putting here vs https://github.com/meta-pytorch/tokenizers library directly?

I have strong preference to have all the wasm code colocated right now since its in a more exploratory phase of development.

JacobSzwejbka (approved):

CMakeLists.txt
larryliu0820: All of these should belong to tokenizers repo https://github.com/meta-pytorch/tokenizers

PR #13567, opened 2025-08-21
Run Python unit test CI on Windows

GregoryComer:
I'm temporarily disabling tokenizers on Windows, as it does not currently build. I'll address this as a follow-up. It may be as simple as building with Clang. Since it's built through install_requirements.py -> pip, I need to figure out how to pass the toolset variable into the inner CMake build. May by a simple change in the tokenizers repo setup.py.


GregoryComer:
* __->__ #13567
* #13667
* #13564
* #13260
* #13259
* #13672

PR #13572, opened 2025-08-21
Summary: Add MCU model verification to CI

psiddh:
Test Plan: OSS CI

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -940,3 +940,42 @@ jobs:
       build-mode: Release
       build-tool: cmake
       docker-image: ci-image:executorch-ubuntu-22.04-clang12
+
+  test-mcu-models:
+    name: test-mcu-models
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    strategy:
+      matrix:
+        include:
+          - build-tool: cmake
+      fail-fast: false
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: ci-image:executorch-ubuntu-22.04-arm-sdk
+      submodules: 'recursive'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
+      script: |
+        BUILD_TOOL=${{ matrix.build-tool }}
+
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+
+        # Try to mirror these as closely as possible
+        source .ci/scripts/utils.sh
+        install_executorch "--use-pt-pinned-commit"
+
+        .ci/scripts/setup-arm-baremetal-tools.sh
+        source examples/arm/ethos-u-scratch/setup_path.sh
+
+        # Run selective Build
+        chmod +x examples/selective_build/test_selective_build.sh
+        examples/selective_build/test_selective_build.sh "${BUILD_TOOL}"
+
+        # Run MCU models
+        chmod +x examples/arm/run_mcu_models_fvp.sh
+        examples/arm/run_mcu_models_fvp.sh --target=cortex-m55
--- a/examples/arm/run_mcu_models_fvp.sh
+++ b/examples/arm/run_mcu_models_fvp.sh
@@ -19,23 +19,16 @@ set -u
 
 # Valid targets for MCU model validation
 VALID_TARGETS=(
-    "ethos-u55-32"
-    "ethos-u55-64"
-    "ethos-u55-128"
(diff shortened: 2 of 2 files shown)


.github/workflows/trunk.yml
```diff
@@ -940,3 +940,40 @@ jobs:
...
+      id-token: write
+      contents: read
+    with:
+      runner: linux.2xlarge
+      docker-image: ci-image:executorch-ubuntu-22.04-arm-sdk
+      submodules: 'recursive'
+      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      timeout: 90
+      script: |
+        BUILD_TOOL=${{ matrix.build-tool }}
```
digantdesai, line 962: Do we support anything else?

digantdesai:
Looks good. Kicked them off on the CI. Let's see how long the job takes else we can reduce the number of models running serially on the job or somehow run them in parallel.

psiddh:
> Looks good. Kicked them off on the CI. Let's see how long the job takes else we can reduce the number of models running serially on the job or somehow run them in parallel.

Looks like job passed : https://github.com/pytorch/executorch/actions/runs/17146863588/job/48644753234?pr=13572 

════════════════════════════════════════════════════════════════
🏁 MCU MODEL VALIDATION SUMMARY - TARGET: ethos-u85-128
════════════════════════════════════════════════════════════════

lstm         : ✅ Passed
mv2          : ✅ Passed
mv3          : ✅ Passed
resnet18     : ✅ Passed

.github/workflows/trunk.yml
```diff
@@ -940,3 +940,42 @@ jobs:
...
+
+        # The generic Linux job chooses to use base env, not the one setup by the image
+        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
+        conda activate "${CONDA_ENV}"
+
+        # Try to mirror these as closely as possible
+        source .ci/scripts/utils.sh
+        install_executorch "--use-pt-pinned-commit"
```
mergennachin, line 970: Do you need "--use-pt-pinned-commit"?

psiddh: > Do you need "--use-pt-pinned-commit"?

No I don't think this is needed
[EDIT] Looks like this is needed, When I removed it, the CI job was failing

.github/workflows/trunk.yml
```diff
@@ -940,3 +940,42 @@ jobs:
...
+        .ci/scripts/setup-arm-baremetal-tools.sh
+        source examples/arm/ethos-u-scratch/setup_path.sh
+
+        # Run selective Build
+        chmod +x examples/selective_build/test_selective_build.sh
+        examples/selective_build/test_selective_build.sh "${BUILD_TOOL}"
+
+        # Run MCU models
+        chmod +x examples/arm/run_mcu_models_fvp.sh
+        examples/arm/run_mcu_models_fvp.sh --target=ethos-u85-128
```
digantdesai: why do we need to have ethos targets for this script? just take cortex-m55, cortex-m85?

digantdesai (approved):
please resolve comments before merging.

.github/workflows/trunk.yml
psiddh: Yes this involves changing build_executor_runner.sh script (which the newly added 'run_mcu_models_fvp.sh' relies on). In the build_executor_runner.sh, it has something like this

if [[ ${target} == *"ethos-u55"*  ]]; then
    target_cpu=cortex-m55
else
    target_cpu=cortex-m85
fi

.github/workflows/trunk.yml
psiddh: I dont think we do, I was copy pasting this from other jobs to keep it consistent with other jobs

.github/workflows/trunk.yml
digantdesai: let's do
```
if [[ ${target} == "cortex-m55" ]]; then
   target_for_build_et_runner = "ethos-u55-128"
else ...
```

.github/workflows/trunk.yml
digantdesai: why was it failing? and how? Link?

examples/arm/run_mcu_models_fvp.sh
```diff
@@ -71,6 +64,22 @@ validate_models() {
     return 0
 }
 
+cpu_to_ethos_target() {
+  local cpu=$1
+  case $cpu in
+    cortex-m55)
```
digantdesai, line 70: ah perfect!! ❤️

psiddh:
Latest run passed : https://github.com/pytorch/executorch/actions/runs/17278005534/job/49039591722?pr=13572

As a follow up ,check why removing "--use-pt-pinned-commit" is causing assertion failures while building portable kernels

PR #13574, opened 2025-08-21
NXP backend: Remove IR optimization to remove dead branches.

MartinPavella:
### Summary
This PR replaces an IR optimization that removes dead code from the model, by an equivalent executorch call.

### Test plan
Unit test provided in `backends/nxp/tests/test_removing_dead_code.py`.


JakeStevens (approved):

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -242,8 +243,13 @@ def __init__(self):
     def transform_for_annotation(
         self, model: torch.fx.GraphModule
     ) -> torch.fx.GraphModule:
-        pass_runner = NeutronAtenPassManager()
-        return pass_runner(model).graph_module
+        model.graph.eliminate_dead_code()  # Remove dead code to simplify the graph for the passes.
```
digantdesai, line 246: why do we need before?

MartinPavella: In case the model already contains dead code. We want to simplify the graph as much as possible for the passes.

PR #13575, opened 2025-08-21
Arm backend: Dont try to fuse const for TOSA ops

per:
### Summary
The TOSA ops doesn't have any backed reference implementation, so avoid trying to fuse constant ops and leave for the backend compiler to do the work.

### Test plan
Tested through unit tests.


oscarandersson8218 (approved):

per:
Unrelated failure.

PR #13576, opened 2025-08-21
NXP backend: Use zero point for quantized padding.

MartinPavella:
### Summary
This PR fixes cases where padding with the value `0` was used for quantized operators. Now, zero point is used instead.

### Test plan
Unit tests provided.

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/avg_pool_2d_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/avg_pool_2d_converter.py
@@ -3,11 +3,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import numpy as np
+
 from executorch.backends.nxp.backend.ir.converter.conversion import (
     aten_translator,
     common,
 )
 from executorch.backends.nxp.backend.ir.converter.conversion.common import OpsList
+from executorch.backends.nxp.backend.ir.converter.conversion.translator import (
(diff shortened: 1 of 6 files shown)


backends/nxp/backend/ir/converter/node_converters/ops_converters/avg_pool_2d_converter.py
```diff
@@ -57,9 +62,20 @@ def _convert_2d_avg_pool(
         )
 
         if explicit_padding is not None:
-            # Need to prepend a 'Pad' operator, which adds 0s. But these will be included in the computation!
+            # Need to prepend a 'Pad' operator, which adds 0s (or `zero_point` for the quantized case). But these will
+            #  be included in the computation!
+            input_quantization = t_op.tmp_inputs[0].quantization
+            pad_value = (
+                None
```
JakeStevens, line 74: Why None instead of 0 in non quantized?

JakeStevens (approved):

backends/nxp/backend/ir/converter/node_converters/ops_converters/avg_pool_2d_converter.py
MartinPavella: `None` is the default value of the `builder.create_pad_operator_before()` method's `constant_value` parameter. This way, the actual default padding value (`0`) is only defined in 1 place.

But it's hard to image that the default padding value would ever be changed, and using `0` here would make the code more understandable. I have no problem using `0` instead of `None` if you prefer.

PR #13588, opened 2025-08-21
Delete tools/cmake/{buck_util,resolve_buck}.py

swolchok:
We don't need resolve_buck.py in check_for_installed_private_headers_in_cmake_out, so we can just delete it.


swolchok:
* __->__ #13588
* #13559
* #13558
* #13485

PR #13606, opened 2025-08-22
Move undelegated constants

lucylq:
Currently, the xnnpack + lora + program-data separation flow moves all delegated constants to the .ptd file. This change moves the undelegated constants as well, reducing the pte file size.

Run export_lora.sh from https://github.com/meta-pytorch/executorch-examples/pull/54

```
-rw-r--r-- 1 lfq users 5994013600 Aug 22 13:32 foundation.ptd
-rw-r--r-- 1 lfq users   27628928 Aug 22 13:32 llama_3_2_1B_lora.pte
-rw-r--r-- 1 lfq users 5994013600 Aug 22 13:30 llama_3_2_1B.ptd
-rw-r--r-- 1 lfq users     317248 Aug 22 13:30 llama_3_2_1B.pte
```

fp32 Weights: ~6GB
Program: ~300KB
Lora adapter weights: ~27MB

### DIFF
--- a/exir/program/_program.py
+++ b/exir/program/_program.py
@@ -808,11 +808,13 @@ def edge_to_executorch_passes(
     Get the pre memory planning passes based on the method name, if the pass is not in the dict, use the default pass.
     """
     passes: List[PassType] = [
-        *config.passes,
         SpecPropPass(),
         # ExecuTorch backend ops are unable to handle unbacked symints. So after
         # this pass, passes cannot be Interpreter-based, because it will fail if
         # there exists an unbacked symint operation.
+        *config.passes,
+        # config.passes may contain external_constants_pass. This pass has to
+        # run after SpecPropPass, which populates tensor names.
         EdgeToBackendOpsPass(),
         RemoveGraphAssertsPass(),
     ] + pre_memory_planning_passes(config, name)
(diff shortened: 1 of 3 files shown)


exir/program/_program.py
```diff
@@ -808,13 +808,13 @@ def edge_to_executorch_passes(
...
     """
     passes: List[PassType] = [
-        *config.passes,
         SpecPropPass(),
         # ExecuTorch backend ops are unable to handle unbacked symints. So after
         # this pass, passes cannot be Interpreter-based, because it will fail if
         # there exists an unbacked symint operation.
         EdgeToBackendOpsPass(),
         RemoveGraphAssertsPass(),
+        *config.passes,
```
lucylq: external_constants_pass has to run after SpecPropPass, which populates the tensor names. See if CI passes, not sure if this will break anything.

JacobSzwejbka: Can you do the more minimal change then and just move it after specProp rather then all of them then?

lucylq: @JacobSzwejbka, updated. Do EdgeToBackendOpsPass/RemoveGraphAssertsPass have to be relatively late in the passes?

mergennachin (approved):

PR #13614, opened 2025-08-22
Update coremltools to 9b1

metascroy:
This updates main to use coremltools 9.0b1.

We will pin the ExecuTorch GA release to use the official coremltools 9.0 release when it comes out.

### DIFF
--- a/backends/apple/coreml/compiler/torch_ops.py
+++ b/backends/apple/coreml/compiler/torch_ops.py
@@ -12,15 +12,13 @@ import numpy as np
 import torch as _torch
 from coremltools import _logger
 from coremltools.converters.mil.frontend import _utils
-from coremltools.converters.mil.frontend.torch.dom_order_ops import (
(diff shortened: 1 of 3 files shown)


backends/apple/coreml/compiler/torch_ops.py
```diff
@@ -47,26 +44,6 @@ def split_copy(context, node):
     split(context, node)
 
 
-@register_torch_op(
```
metascroy, line 50: This op was added to coremltools 9.0, so we no longer need to register it in ET

metascroy: Actually, the failing unit tests are being the version of this op in coremltools 9.0 is not being lowered correctly, so we might need to ET version still.  Will investigate further.

JacobSzwejbka (approved):

YifanShenSZ (approved):

PR #13630, opened 2025-08-25
Arm backend: Use cmake for building in Ethos-U jupyter example

AdrianLundell:
Building the example application using cmake is now straight forward enough to not need any helper scripts.

Additionally simplify the the example + path setup in the executor runner cmake script and make the default path match the example.

### DIFF
--- a/examples/arm/ethos_u_minimal_example.ipynb
+++ b/examples/arm/ethos_u_minimal_example.ipynb
@@ -18,13 +18,12 @@
    "source": [
     "# Ethos-U delegate flow example\n",
     "\n",
-    "This guide demonstrates the full flow for running a module on Arm Ethos-U using ExecuTorch. \n",
+    "This guide demonstrates the full flow for running a module on Arm Ethos-U55 using ExecuTorch.\n",
     "Tested on Linux x86_64 and macOS aarch64. If something is not working for you, please raise a GitHub issue and tag Arm.\n",
     "\n",
     "Before you begin:\n",
     "1. (In a clean virtual environment with a compatible Python version) Install executorch using `./install_executorch.sh`\n",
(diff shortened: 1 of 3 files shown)


oscarandersson8218 (approved):

oscarandersson8218:
Unrelated CI failures

examples/arm/ethos_u_minimal_example.ipynb
```diff
@@ -193,22 +184,37 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%bash \n",
+    "source ethos-u-scratch/setup_path.sh\n",
+    "\n",
+    "# Build example executor runner application to examples/arm/ethos_u_minimal_example\n",
```
digantdesai, line 207: I think we run this notebook in CI but just to avoid code here, can we use some script like build_executor.sh which is covered by CI?

AdrianLundell: We wanted to avoid the "bespoke inconsistent scripts" mentioned in https://github.com/pytorch/executorch/discussions/10661 and show the build process as true to reality as possible here. We want the cmake to be documented somewhere imo, so adding an additional script only forces us to document and maintain both that script and the real flow. Why do you want to avoid the code here?

The notebook should be tested here: https://github.com/pytorch/executorch/blob/deaf37f4f7e2a4205ad09b75f04427568ff66ff2/.github/workflows/trunk.yml#L432

digantdesai: I think I remember the test_arm_ootb.sh script. I was highlight to just call existing build scripts instead of spelling out the cmake command here.

PR #13636, opened 2025-08-25
ET windows build flags

ivayloen:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13638
* #13637
* __->__ #13636

Add compiler flags for windows msvc builds

Differential Revision: [D78854132](https://our.internmc.facebook.com/intern/diff/D78854132/)

PR #13638, opened 2025-08-25
ET use compiler macros in ops

ivayloen:
* __->__ #13638
* #13637
* #13636

Use new compiler macros to fix operator library builds.
Additional issues,
isnan and signbit supported types are only on float, double and long double. Apply static cast to avoid msvc compiler error.

Differential Revision: [D78854136](https://our.internmc.facebook.com/intern/diff/D78854136/)


hsharma35 (approved):

GregoryComer:
Overall, these changes look reasonable to me. The main concern that I want to address prior to approving is ensuring that we have GitHub CI coverage for all platforms we're supporting internally.

As @mergennachin mentioned, if you can use clang-cl internally, that would align with what we're supporting via CMake. We could add CI coverage for the Windows build via buck to guard this.

PR #13645, opened 2025-08-25
Allow none and string input types for Method

shoumikhin:


mergennachin:
Add unit test please

shoumikhin:
@mergennachin this is to fix periodic jobs, specificaly emform_predict model that's currently failing, so it's gonna be the test essentially

kirklandsign (approved):

runtime/executor/method.cpp
```diff
@@ -1088,7 +1088,9 @@ Method::set_input(const EValue& input_evalue, size_t input_idx) {
     return Error::InvalidArgument;
   }
 
-  if (e.isTensor()) {
+ if (e.isNone()) {
```
JacobSzwejbka: What are you doing here?

JacobSzwejbka: Oh you are avoiding the error in the else.

Can you 1. Check this after tensor since tensor will be the common case. 2. Use else if

JacobSzwejbka (changes requested):

runtime/executor/method.cpp
shoumikhin: Had to use "else if" indeed

JacobSzwejbka (approved):

PR #13649, opened 2025-08-25
[ET][Optimized] Fallback to portable op utils (add/div/mul/sub/le)

manuelcandales:
* #13650
* __->__ #13649
* #13648

Optimized implementations of `add`, `div`, `mul`, `sub` & `le` fallback to portable implementations (via portable op utils) for the generic case.

Differential Revision: [D80963874](https://our.internmc.facebook.com/intern/diff/D80963874/)


swolchok (approved):
this seems great!

PR #13651, opened 2025-08-25
Resurface low level runtime API page

mergennachin:
See https://github.com/pytorch/executorch/issues/13631


JacobSzwejbka:
It was hidden because its stale I believe. I was going to clean it up after branch cut

JacobSzwejbka (approved):

GregoryComer (approved):
Looks good to me. I probably haven't covered this in enough detail in the new docs flow. I'd like to do a pass on these docs before 1.0 to clean up and homogenize a bit - I'll create a task to track it.

PR #13656, opened 2025-08-25
Fully enable the stale PR workflow

nil-is-all:
Changed lines 96-101, 107-112, and 115-120 which update the PRs based on stale condition


GregoryComer:
For additional context, we landed the original workflow in a "dry run" mode by commenting out the actual API calls to label and close issues. I've spot checked the run logs (see https://github.com/pytorch/executorch/actions/workflows/stale.yml) and the PRs it identifies as stale look sane. I'm good to enable the full workflow now.

GregoryComer (approved):

PR #13653, opened 2025-08-25
Add 16A8W support and test for add operation

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #13653

Add 16A8W quantization support and test for the add operation in ExecutorTorch ARM backend.

This follows the pattern established for linear operations, extending int16 support to add operations.

Changes:
- Add INT16 dtype validation support in op_add.py
- Add test_add_tensor_16a8w_tosa_INT test function
- Enable test_add.py in test targets configuration

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency.

Differential Revision: [D80510463](https://our.internmc.facebook.com/intern/diff/D80510463/)

PR #13641, opened 2025-08-25
Add 16A8W quantization configuration utility for ARM backend

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13653
* #13658
* __->__ #13641

This diff implements a 16A8W (16-bit activations, 8-bit weights) quantization configuration utility for the ExecutorTorch ARM backend, following the feedback from D79746479.

## Key Changes

**1. New Quantization Configuration Function**
- Add `get_16a8w_quantization_config()` in `fbcode/executorch/backends/arm/quantizer/arm_quantizer.py`
- Provides 16-bit activations with HistogramObserver (better precision than 8A8W)
- Maintains 8-bit weights with MinMaxObserver/PerChannelMinMaxObserver (memory efficient)
- **Technically supported by TOSA through [EXT-INT16 extension/profile](https://www.mlplatform.org/tosa/tosa_spec.html#_conv2d)**

## Benefits
- **Better Precision**: 16-bit activations provide higher precision than 8-bit. Useful for carrying precision for recurring neural nets.

Differential Revision: [D79763381](https://our.internmc.facebook.com/intern/diff/D79763381/)

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13448, opened 2025-08-15
Add 16A8W linear ops support and test

Ninja91:
Summary:
- Adds linear ops test using the 16A8W config in INT16 profile.
- Adds support in view ops validation for INT16 Dtype.
- Validated with TOSA pipeline test.

Note: Not verified with tosa reference model run.

Differential Revision: D80308822

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13658, opened 2025-08-25
Add 16A8W linear ops support and test

Ninja91:
* #13653
* __->__ #13658
* #13641

- Adds linear ops test using the 16A8W config in INT16 profile.
- Adds support in view ops validation for INT16 Dtype.
- Validated with TOSA pipeline test.
- Checked earlier marked flaky tests no longer flaky and remove markers.

Note: Not verified with tosa reference model run.

Differential Revision: [D80308822](https://our.internmc.facebook.com/intern/diff/D80308822/)


Ninja91:
Updated version of https://github.com/pytorch/executorch/pull/13448. Closing the former.

Ninja91:
@per @digantdesai some of CI is failing but it doesn't appear to be related to code change here.

backends/arm/operators/op_transpose.py
```diff
@@ -47,7 +47,7 @@ def define_node(
         validate_valid_dtype(
             self.target,
             [inputs[0], output],
-            [ts.DType.INT8, ts.DType.INT32, ts.DType.FP32],
+            [ts.DType.INT8, ts.DType.INT16, ts.DType.INT32, ts.DType.FP32],
```
per: This could be split up into:
```
valid_dtypes = [ts.DType.bool]
if self.tosa_spec.support_integer():
     valid.dtypes.extend([ts.DType.INT8, ts.DType.INT16])
if self.tosa_spec.support_float():
    valid.dtypes.extend([ts.DType.FP16, ts.DType.FP32])

```

backends/arm/operators/op_view.py
```diff
@@ -44,7 +44,13 @@ def define_node(
         validate_valid_dtype(
             self.target,
             [inputs[0], output],
-            [ts.DType.INT8, ts.DType.INT32, ts.DType.FP32, ts.DType.BOOL],
```
per, line 47: Same as for op_transpose.py

per (approved):
Nice! Two small suggestions, otherwise LGTM.

backends/arm/test/ops/test_linear.py
```diff
@@ -258,3 +262,64 @@ def test_linear_vgf_INT(test_data: torch.Tensor):
...
+        get_symmetric_a16w8_quantization_config(
+            is_per_channel=per_channel_quantization
+        ),
+    )
+
+
+@common.parametrize("test_data", test_data_rank1_INT | test_data_rank4_INT)
+@pytest.mark.xfail(
+    reason="missing int16 linear ops support; fails at TOSA reference model run with Invalid TOSA graph"
+)
```
digantdesai, line 282: Seems like bias and output dtype mismatch was the issue, did we look into that? Did you try without bias?

digantdesai:
`backends/arm/test/ops/test_sigmoid_16bit.py::test_sigmoid_tosa_INT_add_sigmoid[rand]` failure seems unrelated.

backends/arm/test/ops/test_linear.py
Ninja91: Will look into that as a follow up. Based on test you shared: https://git.mlplatform.org/tosa/conformance_tests.git/tree/operators/tensor/fully_connected/fully_connected_6x44_i16xi8_acci48

It looks like the output dtype and bias dtype should be INT48. How do we go about this? 
In our current quantization setup we don't add any quant spec for bias and it gets quantized to INT32. Do you recommend adding BIAS spec to quantize to INT16 as well and try? Or do you want me to change both output and bias dtype to int32/int64 as there's no int48 in Pytorch AFAIK.

backends/arm/operators/op_transpose.py
Ninja91: Will update.

backends/arm/test/ops/test_linear.py
Ninja91: Created task here: https://github.com/pytorch/executorch/issues/13729

per: I have an internal fix for that to handle it in the backend.

per:
@digantdesai Were was this merged? Can't see it on main... Getting a bit confused by the ghstack stuff 😕

digantdesai:
@Ninja91 ?

Ninja91:
> @digantdesai Were was this merged? Can't see it on main... Getting a bit confused by the ghstack stuff 😕

This diff was merged and new PR was created for main branch here: https://github.com/pytorch/executorch/pull/13754/

I see that there are some tests failing. Looks like the change to use support_integer() causes FP tests fail as some ops are still INT.
`
if self.tosa_spec.support_integer():
     valid.dtypes.extend([ts.DType.INT8, ts.DType.INT16])
if self.tosa_spec.support_float():
    valid.dtypes.extend([ts.DType.FP16, ts.DType.FP32])
`

What would be your suggestion?

digantdesai:
this didn't merge with main, can you push this to main manually?

PR #13660, opened 2025-08-25
Force -O3 for executorch op_div.cpp in "clang 17" also

akrieger:
Summary:
Some platforms use clang 19 under the hood but are setting the clang 17
constraint for bad reasons.

Differential Revision: D80980948


swolchok (approved):
as long as it passes OSS CI it seems fine

PR #13672, opened 2025-08-25
Fix build on Windows, re-enable Windows build CI job

GregoryComer:
CMake doesn't seem to like empty exclude patterns.


GregoryComer:
* __->__ #13672

CMakeLists.txt
```diff
@@ -595,13 +595,22 @@ if(EXECUTORCH_BUILD_EXTENSION_DATA_LOADER)
...
     set(data_loader_exclude_pattern "*mman_windows.h")
   endif()
-  install(
-    DIRECTORY extension/data_loader/
-    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/executorch/extension/data_loader
-    FILES_MATCHING
-    PATTERN "*.h"
-    PATTERN ${data_loader_exclude_pattern} EXCLUDE
-  )
+  if (DEFINED data_loader_exclude_pattern)
```
GregoryComer: @swolchok It looks like CMake complains when there's an exclude with an empty pattern (see https://github.com/pytorch/executorch/actions/runs/17222786697/job/48861521892, line 818). This is a bit of a pain as there doesn't seem to be an elegant solution to this. Do you have any preferences on this approach?

I suppose we could default the exclude pattern to some string that matches nothing, though that seems a little hacky. It looks like we could also maybe build a list of install args, including the "magic" keywords, though I haven't tried it.

GregoryComer:
Note: When I attempted to re-order my stack, ghstack seems to have closed a bunch of the PRs and merged them into this one. I'm not sure why the re-ordering triggered this, but I'm going to raise this as a new stack.

GregoryComer:
@digantdesai Can you take a look at the XNNPACK changes? Should be straightforward, just want an approval from someone with context.

PR #13673, opened 2025-08-26
Install NXP and OpenVINO backends, top-level CMake targets

GregoryComer:
Make the top-level backends, extensions, and kernels targets installable. See https://github.com/pytorch/executorch/issues/12949. Add CMake export directives for NXP and OpenVINO, which are needed to support the exportable top-level targets.

I've also removed the apple and llm/apple extensions from the executorch_extensions target. From what I understand, these won't be consumed this way, as they are build via apple frameworks. There is some installation complexity to make it work with install, so I'm inclined to skip it if it won't be used. Will check with Anthony in the comments.


GregoryComer:
* __->__ #13673

GregoryComer:
Note that the failures on the macos jobs (related to the private header check) are pre-existing. I verified that the failure matches the latest jobs from the HUD.

GregoryComer:
@shoumikhin Are the extension/apple and extension/apple/llm targets ever used via add_subdirectory or find_package, or are they only intended to be used via the compiled framework?

I noticed that they aren't currently installed with CMake, and am wondering if they need to be or if we can assume that the CMake targets will only be used to build the frameworks.

GregoryComer:
I talked with @shoumikhin offline. The extension/apple and extension/llm/apple targets are only used for building the frameworks and have their own install mechanism - so they don't need to be installed with CMake or included in the interface targets.

GregoryComer:
@mergennachin Do you have any concerns with approving this PR? The export changes for NXP and OpenVINO should be low/no-risk, they were just not done as part of the main export changes.

mergennachin (approved):

PR #13674, opened 2025-08-26
Enable per-row/per-col grouping in CoreML LUT ops

metascroy:
This PR enables per-row grouping in CoreML LUT ops.


YifanShenSZ (approved):
LGTM! Thanks for adding more LUT support!

One quick question though: In [CoreML LUT](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS18.compression.constexpr_lut_to_dense) we allow arbitrary vector axis, is there a reason in torch we support only row or col?

metascroy:
> LGTM! Thanks for adding more LUT support!
> 
> One quick question though: In [CoreML LUT](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS18.compression.constexpr_lut_to_dense) we allow arbitrary vector axis, is there a reason in torch we support only row or col?

This is still scalar-valued LUT (vector axis is for vector-valued LUT, which is still not supported).

The row/col here refer to the granularity of the LUTs as discussed here: https://apple.github.io/coremltools/docs-guides/source/opt-palettization-overview.html

The diagram for "per-grouped channel lookup table" corresponds to per-row grouping.

If the original tensor has size (n, k), we can support 1 LUT per col_group_size columns (can be specified with block_size = [-1, col_group_size], where -1 is interpreted as n), or we can support 1 LUT per row_group_size rows (can be specified as block_size = [row_group_size, -1], where -1 is interpreted as k).

metascroy:
@cccclai @digantdesai can I get a stamp for ET to merge?

billmguo (approved):

PR #13676, opened 2025-08-26
Qualcomm AI Engine Direct - Fixed the accuracy bug for conv with bias in block quant

shewu-quic:
As title.

cc: @haowhsu-quic

### DIFF
--- a/backends/qualcomm/quantizer/observers/per_block_param_observer.py
+++ b/backends/qualcomm/quantizer/observers/per_block_param_observer.py
@@ -35,6 +35,9 @@ class PerBlockParamObserver(AffineQuantizedMinMaxObserver):
             **kwargs,
         )
         self.block_size = block_size
+        # TODO: expand this when QNN starts to support more configurations
+        self.bitwidth_of_scale = 4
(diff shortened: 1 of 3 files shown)


shewu-quic:
Hi @cccclai,

This PR is to improve the accuracy of Qwen 2,5 in block quant. We have observed a bug when conv op with bias in block quant,
Could you please take a look? Thanks.

backends/qualcomm/quantizer/observers/per_block_param_observer.py
```diff
@@ -35,6 +35,9 @@ def __init__(
             **kwargs,
         )
         self.block_size = block_size
+        # TODO: expand this when QNN starts to support more configurations
+        self.bitwidth_of_scale = 4
```
cccclai, line 39: Is it the only bit width scale supported currently?

cccclai (approved):

backends/qualcomm/quantizer/observers/per_block_param_observer.py
shewu-quic: Yes, at the moment, it is the only supported bit-width scale.

PR #13682, opened 2025-08-26
Arm backend: Remove unused xfail dict

martinlsm:


zingo:
rebasing to see if sigmoind pass after the other sigmoind fix

martinlsm:
Change is already merged with another patch

PR #13687, opened 2025-08-26
Build an AAR with QNN

kirklandsign:


.github/workflows/android-release-artifacts.yml
```diff
@@ -89,11 +90,22 @@ jobs:
...
         FLAVOR="${{ inputs.flavor }}"
         if [[ "$FLAVOR" == "vulkan+xnnpack" || -z "$FLAVOR" ]]; then
           export EXECUTORCH_BUILD_VULKAN=ON
         fi
 
+        if [[ "$FLAVOR" == "qnn" ]]; then
+          PYTHON_EXECUTABLE=python bash .ci/scripts/setup-qnn-deps.sh
+          PYTHON_EXECUTABLE=python bash .ci/scripts/build-qnn-sdk.sh
+          export EXECUTORCH_BUILD_QNN=ON
+          export QNN_SDK_ROOT=/tmp/qnn/2.28.0.241029
```
cccclai: Can we try to reduce the hardcode version and refer to this https://github.com/pytorch/executorch/blob/main/backends/qualcomm/scripts/qnn_config.sh ? If it needs to refactor a bit that's fine

kirklandsign: Nice

kirklandsign: Oh... This is part of docker image anyways..

cccclai: Ah yeah I just meant this part `/tmp/qnn/2.28.0.241029`, the `2.28.0.241029` can be different when we bump the qnn version

kimishpatel (commented):
Thanks Hansong. This looks like the right approach. But I was trying to validate something similar and it doesnt quite seem to work yet. One thing is that libQnnModelDlc.so is not distributed in the maven package. I am gonna ask qualcomm to publish one

kimishpatel (approved):

.github/workflows/android-release-artifacts.yml
```diff
@@ -83,17 +84,29 @@ jobs:
...
         fi
 
+        if [[ "$FLAVOR" == "qnn" ]]; then
+          PYTHON_EXECUTABLE=python bash .ci/scripts/setup-qnn-deps.sh
+          PYTHON_EXECUTABLE=python bash .ci/scripts/build-qnn-sdk.sh
+          export EXECUTORCH_BUILD_QNN=ON
+          source backends/qualcomm/scripts/qnn_config.sh
+          export QNN_SDK_ROOT="/tmp/qnn/${QNN_VERSION}"
+          export ANDROID_ABIS=arm64-v8a
+          GRADLE_ARGS+=" -DqnnVersion=2.28.0"
```
mergennachin, line 107: cc @cccclai make sure to update this when you update the QNN version

PR #13695, opened 2025-08-26
Fix build_android_library.sh

kimishpatel:
### Summary
Qnn backend requires libQnnDlc. Package that in the aar file

### Test plan
CI
Plus local testing for llamademo app revealed that without this change it throws error

### DIFF
--- a/scripts/build_android_library.sh
+++ b/scripts/build_android_library.sh
@@ -67,6 +67,15 @@ build_android_native_library() {
   # Copy QNN related so library
   if [ -n "$QNN_SDK_ROOT" ] && [ "$ANDROID_ABI" == "arm64-v8a" ]; then
     cp "${CMAKE_OUT}"/lib/libqnn_executorch_backend.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnModelDlc.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnHtp.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnSystem.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnHtpV69Stub.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnHtpV73Stub.so ${SO_STAGE_DIR}
+    cp "${QNN_SDK_ROOT}"/lib/aarch64-android/libQnnHtpV75Stub.so ${SO_STAGE_DIR}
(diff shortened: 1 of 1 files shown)


mergennachin:
@kirklandsign @kimishpatel @cccclai why we didn't our CI catch this error? Is it something we can add regression test in our CI?

kimishpatel:
actually this one is probably not needed. I erroneously thought this lib is needed. Issue is the qnn delegate tries to load this library in "some" cases. Those cases are not valid for us, so I am not sure why that loading path exist to begin with. Maybe @haowhsu-quic knows

kimishpatel:
> actually this one is probably not needed. I erroneously thought this lib is needed. Issue is the qnn delegate tries to load this library in "some" cases. Those cases are not valid for us, so I am not sure why that loading path exist to begin with. Maybe @haowhsu-quic knows

actually this one, https://github.com/pytorch/executorch/blob/main/examples/qualcomm/utils.py#L146, is doing exactly this thing

cccclai:
I'm still not sure why it's needed...when I just compile the binary and use adb, it seems fine without this .so. It's good to have for jit use cases anyway (most users don't use this flow)

cccclai (approved):

JacobSzwejbka:
seems like the branch history is super messed up

shoumikhin (approved):

kimishpatel:
> seems like the branch history is super messed up

wow ok will fix

kimishpatel:
@kirklandsign so now this pr adds the lines that you removed earlier. I think for aar build for maven we dont need to include these in executorch.aar, but anyone relying on building aar locally needs these. what do you wanna do?

kirklandsign (approved):

kirklandsign:
> @kirklandsign so now this pr adds the lines that you removed earlier. I think for aar build for maven we dont need to include these in executorch.aar, but anyone relying on building aar locally needs these. what do you wanna do?

Thank you. I have https://github.com/pytorch/executorch/pull/13912/files to update the tutorial.

kirklandsign:
> @kirklandsign @kimishpatel @cccclai why we didn't our CI catch this error? Is it something we can add regression test in our CI?

We are building and testing the LlamaDemo app with XNNPACK AAR package only. We didn't build QNN AAR CI. Currently we are using emulator to run CI and QNN won't run.

kirklandsign (commented):
If we copy these files to AAR, and the user continues to use com.qualcomm.qti:qnn-runtime, it will cause duplicated .so right?

kirklandsign (changes requested):
So for local build, we need to add com.qualcomm.qti:qnn-runtime manually, per use case. See https://github.com/pytorch/executorch/pull/13912/files

kimishpatel:
> So for local build, we need to add com.qualcomm.qti:qnn-runtime manually, per use case. See https://github.com/pytorch/executorch/pull/13912/files

yeah thats better. abandoning this one

PR #13696, opened 2025-08-26
Fix batch norm partitioning with Conv3d

GregoryComer:
Summary:
Models with a batch norm following a conv3d cause an internal error during lowering. This diff fixes it by updating the partitioning logic to only rely on fusion with 1d and 2d convs.

This is because XNNPACK doesn't currently support standalone batch norms and only partitions norms that can be fused. We can't fuse with Conv3d, because XNNPACK doesn't have an implementation. The partitioner constraint was missing logic to exclude Conv3d.

Differential Revision: D81069236


digantdesai (approved):
LGTM. Stamping to unblock.

GregoryComer:
CI failures are infra related or broken on trunk.

PR #13669, opened 2025-08-25
Temporarily disable windows preset build in CI

GregoryComer:
Job is failing on trunk. Temporarily disabling while I resolve it.

PR #13699, opened 2025-08-26
Re-enable Windows build CI

GregoryComer:
The Windows CI build job was disabled in https://github.com/pytorch/executorch/pull/13669 due to a new Windows build failure in https://github.com/pytorch/executorch/pull/13485. Since that diff was reverted in https://github.com/pytorch/executorch/pull/13685, we can re-enable the job.

I started fixing the issue in https://github.com/pytorch/executorch/pull/13672, but dropped it due to the revert. It's a simple change, so it can likely be bundled when the affected PR is re-landed.

Test Plan:
The Build Preset / Windows job is passing on this PR.


GregoryComer:
* __->__ #13699

jackzhxng (approved):

PR #13700, opened 2025-08-26
Fix portable library build on Windows

GregoryComer:
**Note: This is a cherry-pick / identical copy of the existing approved PR (https://github.com/pytorch/executorch/pull/13260). I inadvertently broke the ghstack when attempting to re-order the stack, leading to the old PR being merged by ghstack into the ghstack-created branch (not main).**

Enable build of the portable operator library (and executor_runner, now that we have portable) on the Windows preset. There are a couple of patterns that weren't building on Windows. Both std::nan and std::signbit complain about ambiguous overload resolution when given an integer argument.

To solve the std::isnan issue, I introduced isnan_override in math_util.h, following the naming pattern from max_override / min_override / etc. in the same file. It adds a specialized branch for integers to avoid calling std::isnan. For std::signbit, I just locally specialized.

I also re-enabled optimized and quantized kernel libraries in the Windows preset, as they now build with this change.

### DIFF
--- a/kernels/portable/cpu/op_amax.cpp
+++ b/kernels/portable/cpu/op_amax.cpp
@@ -9,6 +9,7 @@
 #include <c10/util/irange.h>
 #include <cmath>
 
+#include <executorch/kernels/portable/cpu/util/math_util.h>
 #include <executorch/kernels/portable/cpu/util/reduce_util.h>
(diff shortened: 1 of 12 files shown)


GregoryComer:
* #13716
* #13702
* __->__ #13700

mergennachin (approved):

PR #13716, opened 2025-08-26
[Windows] Run Python unit test CI on Windows

GregoryComer:
Run unit test CI on Windows. I've disabled a few jobs (mainly pending tokenizers being available on Windows), but most everything is working out of box.

I did also make a few additional fixes:
 * Shorten names for a few external project targets in CMake, in order to work around max path length limitations.
 * Update load code for kernel libraries in python. The binary names do not start with lib by default on Windows. This could be solved in the build, but it seems easier to just extend the glob.
 * Fix an issue blocking iterative builds for XNNPACK - we need to cause the schema move command to overwrite the existing artifact in the build directory.
 * Disable tokenizer install on Windows for now. Pending enablement in the tokenizers repo.
 * Fix paths in setup.py for Windows builds.


GregoryComer:
* #13837
* #13836
* __->__ #13716

third-party/CMakeLists.txt
```diff
@@ -79,8 +79,8 @@ endif()
 # Similar to flatbuffers, we want to build flatcc for the host. See inline comments
 # in the flatbuffers ExternalProject_Add for more details.
 ExternalProject_Add(
-  flatcc_external_project
```
GregoryComer, line 82: This is to resolve path length issues.

.ci/scripts/unittest-windows.ps1
```diff
@@ -0,0 +1,36 @@
...
+    install_executorch.bat
+}
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "Installation was unsuccessful. Exit code: $LASTEXITCODE."
+    exit $LASTEXITCODE
+}
+
+# Run pytest with coverage
+# pytest -n auto --cov=./ --cov-report=xml
+pytest --continue-on-collection-errors -v --full-trace -c pytest-windows.ini -n auto
```
mergennachin: why --continue-on-collection-errors?

third-party/CMakeLists.txt
```diff
@@ -28,8 +28,8 @@ endif()
 # We use ExternalProject to build flatc from source to force it target the host.
 # Otherwise, flatc will target the project's toolchain (i.e. iOS, or Android).
 ExternalProject_Add(
-  flatbuffers_external_project
-  PREFIX ${CMAKE_CURRENT_BINARY_DIR}/flatbuffers_external_project
+  flatbuffers_ep
+  PREFIX ${CMAKE_CURRENT_BINARY_DIR}/flatc_proj
```
mergennachin, line 32: isn't it flatbuffers_ep?

pytest-windows.ini
```diff
@@ -0,0 +1,115 @@
+# NOTE: This file is a copy of pytest.ini, but with additional tests disabled for Windows. This
+# is intended to be a short-term solution to allow for incrementally enabling tests on Windows.
+# This file is intended to be deleted once the enablement is complete.
```
mergennachin, line 3: create an issue to track deletion of this file please

mergennachin (commented):
See inline

pytest-windows.ini
GregoryComer: Filed as https://github.com/pytorch/executorch/issues/13883.

.ci/scripts/unittest-windows.ps1
GregoryComer: Good catch. This is a holdover from progressive enablement and should not be necessary any more. I've removed it.

GregoryComer:
After rebasing, I do see one test failure on the Windows job that looks like a flake. I'm going to investigate.

```
To see why this happens see 'Known limitations' in documentation for pytest-xdist
=========================== short test summary info ===========================
ERROR devtools/etrecord/tests/etrecord_test.py - torch._dynamo.exc.InternalTorchDynamoError: PermissionError: [WinError 32] The process cannot access the file because it is being used by another process
```

third-party/CMakeLists.txt
GregoryComer: This was intentional, though I can rename it if desired. Flatbuffers builds flatc (distinct from flatcc). The path limitations are pretty tight in CI so I squeezed out a few more characters.

mergennachin (approved):
Thanks, looks good!

GregoryComer:
There appears to be a race condition with the inductor code cache when performing strict mode export from multiple processes on Windows. It's causing occasional test failures on ET, so I'm disabling parallel execution on the Windows unit test jobs for now. I'll see if I can put together a nicer repro and file a bug against PyTorch.

If not having parallel execution is prohibitive, we might be able work around it by not exporting on import in the devtools path, which is what consistently repros the issue during test collection. The race condition will still exist, but the rate at which we hit it might be acceptably low.

jackzhxng (commented):
Btw - can you add a `release notes: windows` and add this to it?

GregoryComer:
I've verified that the remaining CI failures are pre-existing (arm tests, core ML model tests, and NXP tests).

kernels/quantized/__init__.py
```diff
@@ -7,7 +7,7 @@
 try:
     from pathlib import Path
 
-    libs = list(Path(__file__).parent.resolve().glob("**/libquantized_ops_aot_lib.*"))
+    libs = list(Path(__file__).parent.resolve().glob("**/*quantized_ops_aot_lib.*"))
```
kimishpatel, line 10: why this change? is lib name something different on windows?

GregoryComer: Yeah, it's named quantized_ops_aot_lib.dll (no lib prefix). We could maybe force the dll to have the lib prefix in the filename in cmake, but it seemed more idiomatic to do this.

PR #13727, opened 2025-08-27
Arm backend: Added TOSA workflow minimal example

Michiel-Olieslagers:
Added a python notebook containing a minimalistic example of how to best use the TOSA workflow within the ARM backend. This is intended for validating and experimenting with model lowering to TOSA, and is aimed at contributors and developers, rather than production deployment.


zingo:
Arm tests fail should be fixed now. I triggered a rebase to test.

zingo (approved):

PR #13733, opened 2025-08-27
Add qnn backend to pip package

cccclai:
Summary:
When generating the wheel packages
- download the qnn sdk such that we can build the qnn backend
- remove the downloaded sdk files so it won't be part of wheel packages

Inside the qnn packages, it includes the AoT part and the the qnn backend.so files (the build from executorch/backends/qualcomm)

For users:
```
# This step will install qnn backend
pip install executorch-0.8.0a0+e5f94da-cp310-cp310-linux_x86_64.whl


# The qnn sdk will be downloaded when users first , if the QNN_SDK_ROOT is not set. This step will only run once 
from executorch.backend.qualcomm import any module
```

Rollback Plan:

Differential Revision: D80992066

### DIFF
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+set -e  # Exit on any error
+
+echo "=== Building Wheel Package ==="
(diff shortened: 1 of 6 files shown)


cccclai:
Use https://github.com/pytorch/executorch/pull/13794 instead

PR #13736, opened 2025-08-27
Enable long audio for WhisperAudioProcessor

rohansjoshi:
Summary: Enabled WhisperAudioProcessor for audio which is longer than one chunk (i.e. > 30 sec)

Reviewed By: jackzhxng

Differential Revision: D81093558


jackzhxng (approved):

jackzhxng (changes requested):
If you unsqueeze(0) then batch size will always be 1?

jackzhxng (approved):
Will handle the batching by chunk outside of this module

PR #13741, opened 2025-08-27
Update selective build example + CI for top-level targets

GregoryComer:
### Summary
Update the selective build example and CI to use the new top-level targets, where appropriate. I've positioned the example as showcasing two flows - a standard way, using top-level targets, and appropriate for most use cases, and an advanced flow where the user creates a custom kernel target. This is mainly useful when integrating custom ops, which we can't easily integrate with top-level targets due to the inverted dependency on the user build.

I've updated the CMake to take an additional arg to specify whether or not to define a custom target and refactored to use the standard top-level target when not. I've also unified the CMake option naming with the top-level target. Finally, I've updated the CI job to cover both flows.

For additional context, I'd like to take a larger look at simplifying the selective build flows and CMake APIs. This will likely happen after 1.0, as it will be a larger effort.

Tracking task: https://github.com/pytorch/executorch/issues/12948.

### Test plan
Changes are covered by the test selective build job.

### DIFF
--- a/examples/selective_build/basic/CMakeLists.txt
+++ b/examples/selective_build/basic/CMakeLists.txt
@@ -1,0 +1,77 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+#
+# Simple CMake build system for selective build demo.
+#
+# ### Editing this file ###
+#
+# This file should be formatted with
+# ~~~
+# cmake-format -i CMakeLists.txt
+# ~~~
+# It should also be cmake-lint clean.
+#
+cmake_minimum_required(VERSION 3.19)
+project(selective_build_example)
+
+set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+set(TORCH_ROOT ${EXECUTORCH_ROOT}/third-party/pytorch)
+
+include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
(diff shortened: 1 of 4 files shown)


GregoryComer:
@psiddh What are your thoughts on the updated example and README?

psiddh:
> @psiddh What are your thoughts on the updated example and README?

The README is much better now and clear. 
- Can you also add a note on which libs to explicitly link to ?

GregoryComer:
> > @psiddh What are your thoughts on the updated example and README?
> 
> The README is much better now and clear.
> 
> * Can you also add a note on which libs to explicitly link to ?

Thanks. I did include this on lines 67 and 135, though it might not be discoverable enough. Do you have any suggestions for how to expose this info?

psiddh (approved):

examples/selective_build/basic/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
+include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
+
+if(NOT PYTHON_EXECUTABLE)
+  resolve_python_executable()
+endif()
+
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+  # Can't set to 11 due to executor_runner.cpp make_unique
```
swolchok, line 33: executorch in general requires 17 anyway, doesn't it?

GregoryComer: Yeah, we can remove this, as it's a holdover from an older version of the example. I'll clean it up.

PR #13755, opened 2025-08-28
Qualcomm AI Engine Direct - Static LLM Refactor & Qwen3 1.7B Improvement

winskuo-quic:
### Summary

- Refactor llama.py. The current script has some limitations when it gets to customizing configs, especially quantization configs. As there are more models enabled, the script is a little messy, consisting of multiple `if`-`else` statement deciding what model should go into specific optimization. We want to move it all the model specs under `__init__.py`.
- Hiding scale/offset into model's metadata, so args.quant_attrs_path is no longer required when evaluating ppl score.
- Enable Qwen3 1.7B with 16a4w_block quant. Before is using 16a8w, which is much slower. Targeting maximizing token rate while ensuring ppl remains within a 20% margin compared to the FP CPU baseline

#### Stats
token rate = 37tok/sec
ppl = 14.79

### Test plan
Tested all scripts ensuring no regression

cc: @haowhsu-quic


winskuo-quic:
Hi @cccclai,
This is the PR @haowhsu-quic was mentioning earlier in the email.

1. We noticed the codebase is getting a little messy as each model has it's own config.
We have moved all these to `__init__.py` to make it easier for everyone to maintain the best config.


2. We have also tuned Qwen3 1.7B.
Before: Config=16a8w all layers, ppl=14.1, token_rate=27tok/sec 
After: Config=16a4w_block16+8bit_kv_io, ppl=14.79, token_rate=37tok/sec
Although PPL increased by 4.89% but token_rate imrpoved 37%. Also using 8bit kv io instead of 16bit kv io, which should reduce some memory usage.
We have also verified output with prompt. It seems reasonable and able to hit EOT.

Please have a look. Thanks

rohansjoshi:
`apply_spinquant` from torchao can't be used to apply SpinQuant R4 (only R1/R2), because of our custom LlamaModel definition: `use_r4` should always be false. To implement R4 (which is a good idea nevertheless) we would need to add it to the model definition in `static_llama.py`, as was done with R3.

winskuo-quic:
> `apply_spinquant` from torchao can't be used to apply SpinQuant R4 (only R1/R2), because of our custom LlamaModel definition: `use_r4` should always be false. To implement R4 (which is a good idea nevertheless) we would need to add it to the model definition in `static_llama.py`, as was done with R3.

Hi @rohansjoshi,
Thanks for reviewing the PR. I have removed r4 config for now and set r4=False when applying spin quant. We can add r4 back in the future if it helps improvement in accuracy.

winskuo-quic:
Rebased to resolve conflict.

cccclai (approved):

PR #13756, opened 2025-08-28
NXP backend: Disable training mode and replace deprecated call

MartinPavella:
### Summary
This PR ensures that input models are always in evaluation mode. Additionally, a deprecated export call was replaced by an up-to-date one.

### Test plan
Correct function is tested by most of the existing tests. 

### DIFF
--- a/examples/nxp/aot_neutron_compile.py
+++ b/examples/nxp/aot_neutron_compile.py
@@ -15,7 +15,6 @@ import executorch.extension.pybindings.portable_lib
 import executorch.kernels.quantized  # noqa F401
 
 import torch
-
 from executorch.backends.nxp.backend.ir.edge_passes.remove_io_quant_ops_pass import (
     RemoveIOQuantOpsPass,
 )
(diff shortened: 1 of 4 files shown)


examples/nxp/aot_neutron_compile.py
```diff
@@ -11,11 +11,9 @@
 from collections import defaultdict
 from typing import Iterator
 
-import executorch.extension.pybindings.portable_lib
```
robert-kalmar: This import must stay in order the next one to pass. The import loads symbols from `_portable_lib*.so`,  e.g. from `lib/python3.10/site-packages/executorch/extension/pybindings/_portable_lib.cpython-310d-x86_64-linux-gnu.so` which is a dependency of the next import, which loads the `libquantized_ops_aot_lib.so` 
I have not yet figured out why the loader does not finds the `portable_lib*.so` automatically. 

Keep it and the tests will pass.

robert-kalmar (changes requested):

examples/nxp/aot_neutron_compile.py
MartinPavella: Thanks for the tip. I have reverted this change.

robert-kalmar (approved):

shoumikhin:
@MartinPavella please check if this CI failure is related to your change
https://github.com/pytorch/executorch/actions/runs/17398933240/job/49387315586

robert-kalmar:
> @MartinPavella please check if this CI failure is related to your change https://github.com/pytorch/executorch/actions/runs/17398933240/job/49387315586

It is, Martin already works on the follow up fix.

PR #13757, opened 2025-08-28
NXP backend: Add preprocessing pass to split multilayer `GRU`.

MartinPavella:
### Summary
This PR introduces a pre-processing pass on the aten dialect level, which splits `gru` nodes with `num_layers > 1` into an equivalent sequence of single layer `gru` nodes.

### Test plan
Unit tests provided in `backends/nxp/tests/test_gru_splitting.py`.

cc  @robert-kalmar @roman-janik-nxp @StrycekSimon @jirioc

### DIFF
--- a/backends/nxp/aten_passes/neutron_aten_pass_manager.py
+++ b/backends/nxp/aten_passes/neutron_aten_pass_manager.py
@@ -16,6 +16,9 @@ from executorch.backends.nxp.aten_passes.fuse_batch_norm_with_linear_pass import
 from executorch.backends.nxp.aten_passes.split_group_convolution import (
     SplitGroupConvolution,
 )
+from executorch.backends.nxp.aten_passes.split_gru_based_on_num_layers import (
+    SplitGRUBasedOnNumLayers,
+)
 from executorch.exir.pass_manager import PassManager
 from torch import nn
 from torch.fx.passes.infra.pass_base import PassResult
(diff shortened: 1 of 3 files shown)


robert-kalmar (approved):

backends/nxp/aten_passes/neutron_aten_pass_manager.py
```diff
@@ -26,6 +29,7 @@ def __init__(self, passes: list[PassType] = None):
         passes: list[PassType] = passes or [
             FuseBatchNormWithConvPass(),
             FuseBatchNormWithLinearPass(),
+            SplitGRUBasedOnNumLayers(),
```
digantdesai, line 36: we do it here in aten is because we want to support quantization?

MartinPavella: We do it in the aten dialect because it allows the pass to focus only of the transformation, and not worry about the quantization. This way, the implementation is simpler, and if the quantization requirements of GRU (or the other ops) ever change, we only need to update the quantizer (otherwise we would also have to update this pass).

robert-kalmar: We do it here for 2 reasons: 
- Neutron NPU supports GRU as primitive operation Neutron IR ==>  for Neutron NPU we do not need and **want** to decompose it to primitive ops by the ExecuTorch. 
- Neutron NPU supports GRU, but only single layer. Therefore we transform the multilayer GRU to a sequence of single layer GRU, perform the quantization and obtain a graph what can be represented in Neutron IR, including the proper quantization parameters for inputs and outputs of individual GRU layers. 
Note: we will preserve the GRU op in to_edge tranformation - coming change

PR #13759, opened 2025-08-28
NXP Backend: Use --index-url for eiq.nxp.com/repository

robert-kalmar:
### Summary
Avoid attack on PyPI index. If `--extra-index-url` is used there is an attack vector when adversary publish a malicious package with same name on pypi.org. And pip might pick the malicious package instead of one on eiq.nxp.com/repository, as "there is no priority in the locations that are searched" [https://pip.pypa.io/en/stable/cli/pip_install/]
Using --index-url eliminates this attack. 

### Test plan
N/A

cc @digantdesai @JakeStevens

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -855,7 +855,8 @@ jobs:
         .ci/scripts/setup-linux.sh --build-tool "cmake"
 
         # Install test requirements
-        pip install -r backends/nxp/requirements-tests.txt
+        pip install -r backends/nxp/requirements-tests-pypi.txt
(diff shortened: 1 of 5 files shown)


digantdesai:
Let's also reserve the package name on pypi?

digantdesai (approved):

robert-kalmar:
> Let's also reserve the package name on pypi?

That is also in the plan. Contacted pypi.org at the beggining of the week already.

digantdesai:
Thanks @robert-kalmar  for a quick turn around.

PR #13760, opened 2025-08-28
NXP backend: Fix incorrect `linear` test.

MartinPavella:
### Summary
This PR fixes an incorrect `linear` test without a bias.

### Test plan
Unit test provided.

cc @robert-kalmar @jirioc @JakeStevens

### DIFF
--- a/backends/nxp/tests/ir/converter/node_converter/test_linear_converter.py
+++ b/backends/nxp/tests/ir/converter/node_converter/test_linear_converter.py
@@ -10,6 +10,7 @@ import torch
 from executorch.backends.nxp.tests.executorch_pipeline import to_edge_program
 from executorch.backends.nxp.tests.executors import convert_run_compare
 from executorch.backends.nxp.tests.models import LinearModule
+from executorch.exir.dialects._ops import ops as exir_ops
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


MartinPavella:
@JakeStevens thank you for pointing out the incorrect test. However, I cannot replicate your issue:
> I believe bias=False causes a crash, or at least does for me in the converter

The test case passes without issues for me.

robert-kalmar:
> @JakeStevens thank you for pointing out the incorrect test. However, I cannot replicate your issue:
> 
> > I believe bias=False causes a crash, or at least does for me in the converter
> 
> The test case passes without issues for me.

@JakeStevens , @MartinPavella , what neutron converter revision you used for testing? The tests here uses still the NeutronConverter_SDK_**25.03**. 
Jake the crash in your case was with 25.06?

backends/nxp/tests/ir/converter/node_converter/test_linear_converter.py
```diff
@@ -26,15 +26,15 @@ def test_linear_conversion__with_bias():
...
 
-    convert_run_compare(edge_program, input_data=input_data, atol=1.0e-6)
+    convert_run_compare(edge_program, input_data=input_data)
 
 
 def test_linear_conversion__without_bias():
     input_shape = (10, 32)
     edge_program = to_edge_program(
-        LinearModule(bias=True), input_shape
+        LinearModule(bias=False), input_shape
```
digantdesai, line 40: curious how did we not test this?

MartinPavella:
> > @JakeStevens thank you for pointing out the incorrect test. However, I cannot replicate your issue:
> > > I believe bias=False causes a crash, or at least does for me in the converter
> > 
> > 
> > The test case passes without issues for me.
> 
> @JakeStevens , @MartinPavella , what neutron converter revision you used for testing? The tests here uses still the NeutronConverter_SDK_**25.03**. Jake the crash in your case was with 25.06?

I use `25.03`.

backends/nxp/tests/ir/converter/node_converter/test_linear_converter.py
MartinPavella: We must have missed it in the rapid development during the early stages.
I have updated the tests again, to make sure the bias is (or is not) present.

robert-kalmar (approved):

PR #13768, opened 2025-08-28
Fix qnn backend error message

kimishpatel:
### Summary

### Test plan

### DIFF
--- a/backends/qualcomm/runtime/backends/QnnImplementation.cpp
+++ b/backends/qualcomm/runtime/backends/QnnImplementation.cpp
@@ -7,12 +7,31 @@
  */
 #include <executorch/backends/qualcomm/runtime/backends/QnnImplementation.h>
 
+#include <cstdlib>
+#include <cstring>
 #include "QnnInterface.h"
 namespace executorch {
 namespace backends {
 namespace qnn {
 
 using executorch::runtime::Error;
+
+namespace {
+// Helper function to check if the error message contains libc++ related issues
+// and provide helpful guidance for conda users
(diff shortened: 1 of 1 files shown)


backends/qualcomm/runtime/backends/QnnImplementation.cpp
```diff
@@ -7,13 +7,32 @@
...
+void CheckLibcppError(const char* error_msg) {
+  if (error_msg == nullptr) {
+    return;
+  }
+
+  // Check if error contains "libc++"
+  if (strstr(error_msg, "libc++") != nullptr) {
+    QNN_EXECUTORCH_LOG_ERROR(
+        "Detected libc++ library loading issue. "
+        "If you are using conda, try: conda install anaconda::libcxx");
```
cccclai, line 31: Maybe let's ask users to run the install_qnn_sdk command instead. I'm not sure the source of `conda install anaconda::libcxx` and it's corresponding version

kimishpatel: We are talking about when users are using pip install executorch. they are not cloning from source.

cccclai: If that's the case, the better solution is we package libcxx for them

kimishpatel: we cannot package libcxx. thats not the piece we can or should own. Users may have their own variant and we dont want to be in the business of doing that.

kimishpatel:
@cccclai can we land this?

Issue #13732, opened 2025-08-27
[RFC] Multi-backend recipes for easy target focused model deployment


abhinaykukkadapu:
## Goal

Reduce cognitive friction to learn about Executorch (and its backends) to lower and execute a model on a specific target. With multi-backend recipes, given a target, we provide an optimized set of configurations that work out of the box, enabling users to easily lower and execute their models with just a few lines of code.

### Before
Without `executorch.export()` and `ExportRecipe`, one typically has to follow these steps for successful model export and execution on a target device:

Let's discuss with an example model that one want to execute with CoreML backend and fallback to XNNPACK for CPU execution:

```python
# BEFORE: Manual lowering and API orchestration

import torch
from executorch.exir.program import to_edge
from executorch.exir.backend.backend_api import validation_disabled
from executorch.exir import EdgeCompileConfig
from torchao.quantization.pt2e.quantize_pt2e import prepare_pt2e, convert_pt2e
from torchao.quantization.pt2e.quantizer import ComposableQuantizer

# CoreML imports
import coremltools as ct
from executorch.backends.apple.coreml.compiler import CoreMLBackend
from executorch.backends.apple.coreml.partition.coreml_partitioner import CoreMLPartitioner
from executorch.backends.apple.coreml.quantizer import CoreMLQuantizer

# XNNPACK imports
from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
from executorch.backends.xnnpack.partition.config.xnnpack_config import ConfigPrecisionType
from executorch.backends.xnnpack.utils.configs import get_xnnpack_executorch_backend_config
from executorch.backends.xnnpack.quantizer.xnnpack_quantizer import (
    XNNPACKQuantizer, get_symmetric_quantization_config
)

# Create CoreML quantizer (optional)
config = ct.optimize.torch.quantization.LinearQuantizerConfig(
    global_config=ct.optimize.torch.quantization.ModuleLinearQuantizerConfig(
        quantization_scheme="symmetric",
        activation_dtype=torch.quint8,
        weight_dtype=torch.qint8,
        weight_per_channel=True,
    )
)
coreml_quantizer = CoreMLQuantizer(config)

# create coreml partitioner
compile_specs = CoreMLBackend.generate_compile_specs(
    compute_precision=ct.precision.FLOAT16,
    compute_unit=ct.ComputeUnit.CPU_AND_NE,
    minimum_deployment_target=ct.target.iOS16,
)

[...]

PR #13791, opened 2025-08-29
[Executorch][target recipes] Add target based recipes for lowering models to a target device

abhinaykukkadapu:
* __->__ #13791

This diff introduces multi backend/ target based recipes to lower a model with very less code. Target recipes provide pre-configured backend recipes to use them and retarget if needed.

See RFC: https://github.com/pytorch/executorch/issues/13732

## Usage

```
from executorch.export import export
from executorch.export.target_recipes import get_ios_recipe

# CoreML + XNNPACK (FP32)
recipe = get_ios_recipe() # default = "ios-arm64-coreml-fp16"
session.save_pte_file("model.pte")
```

## Advanced usage to combine specific recipes of one or two or backends
one can directly use `ExportRecipe.combine_recipes()` to combine specific backend recipes.
```
recipe1 = ExportRecipe.get_recipe(CoreMLRecipeType.FP32)
recipe2 = ExportRecipe.get_recipe(XNNPackRecipeType.FP32)
combined_recipe = ExportRecipe.combine(
                     [recipe1, recipe2],
                     recipe_name="multi_backend_coreml_xnnpack_fp32"
                  )
session = export(model, combined_recipe, example_inputs)
```
Additional changes:
1. Relaxed kwarg validation in the backend providers to just ignore them instead of erroring out.

Fixes: #13732

Differential Revision: [D81297451](https://our.internmc.facebook.com/intern/diff/D81297451/)


kimishpatel:
I left a comment on the original issue. Not fully convinced why we need multibackend recipe as a separate concept

abhinaykukkadapu:
> I left a comment on the original issue. Not fully convinced why we need multibackend recipe as a separate concept

@kimishpatel  Removed the multi backend provider and went with simple funciton calls.

backends/apple/coreml/recipes/coreml_recipe_provider.py
```diff
@@ -111,8 +112,9 @@ def _validate_recipe_kwargs(self, recipe_type: RecipeType, **kwargs: Any) -> Non
 
         unexpected = set(kwargs.keys()) - expected_keys
         if unexpected:
-            raise ValueError(
-                f"Recipe '{recipe_type.value}' received unexpected parameters: {list(unexpected)}"
+            logging.warning(
```
kimishpatel, line 115: Why this turned into warning now? to combine multiple recipes? and fallback to the other backend when it fails?

abhinaykukkadapu: Correct, to combine multiple recipes, for example, `get_coreml_recipe(**kwargs)` this will have to forward them to both coreml and xnnpack. Hard failure like above would restrict the creation of multi backend recipes.

export/recipe.py
```diff
@@ -184,3 +185,126 @@ def get_recipe(cls, recipe: "RecipeType", **kwargs) -> "ExportRecipe":
...
+            if recipe.lowering_recipe and recipe.lowering_recipe.edge_transform_passes:
+                all_transform_passes.extend(
+                    recipe.lowering_recipe.edge_transform_passes
+                )
+
+            # Collect for quantize stage
+            if quantization_recipe := recipe.quantization_recipe:
+                # Collect PT2E quantizers
+                if quantization_recipe.quantizers:
+                    all_quantizers.extend(quantization_recipe.quantizers)
```
kimishpatel, line 260: use composable quantizers here. Generally the concern with quantization is when you combine two quantizers how do you ensure that the latter is not messing with the first one. Composable quantizer in pt2e is for that. I dont know what is equivalent of that for ao quant recipes

export/recipe.py
```diff
@@ -184,3 +185,126 @@ def get_recipe(cls, recipe: "RecipeType", **kwargs) -> "ExportRecipe":
...
+    @classmethod
+    def combine(
+        cls, recipes: List["ExportRecipe"], recipe_name: Optional[str] = None
+    ) -> "ExportRecipe":
+        """
+        Combine multiple ExportRecipe objects into a single recipe.
+
+        Args:
+            recipes: List of ExportRecipe objects to combine
+            recipe_name: Optional name for the combined recipe
```
kimishpatel, line 198: why not just combine it from, the name of the underlying recipes

kimishpatel (commented):
looks ok to me. some nits plus a comment on quantizaiton.

you probably want to add recipes plus test for android

export/recipe.py
abhinaykukkadapu: This is just to compose it as a list for ExportRecipe object, in the actual execution, we do use composable quantizer: [main/export/stages.py#L365](https://github.com/pytorch/executorch/blob/main/export/stages.py#L365)

abhinaykukkadapu:
> looks ok to me. some nits plus a comment on quantizaiton.
> 
> you probably want to add recipes plus test for android

Thanks for the review, will add android specific things after adding QNN recipes, which i'm working on right now.

export/recipe.py
kimishpatel: ok thats good. about ao quantizer, how do we gaurantee the same thing

abhinaykukkadapu: Good point, i don't see any variants of composable quantizer for torchao/ source quantizers: https://github.com/search?q=org%3Apytorch+%22class+Composable%22+language%3APython&type=code

Including **no** code patterns of applying multiple ao quantizers, a noob question, does it even makes sense to apply these sequentially at all? Since the configs can quantize weights and activations differently including a filter function that can be applied to qualify a module.

kimishpatel: Yeah so thats the problem with ao quantizer. For PT2E it is fine because they cannot override previous quantizer. I think for that you probably want to just disallow that pattern for ao quantization for now. Else it will be a debug nightmare

abhinaykukkadapu: Gotcha, thanks for pointing it, will disable in next version.

kimishpatel (approved):

PR #13792, opened 2025-08-29
Arm backend: Fix compilation errors in executor runner

martinlsm:
When building arm_executor_runner.cpp with the two macros ET_EVENT_TRACER_ENABLED and SEMIHOSTING set, compilation errors occur. This patch corrects the code to fix this problem.


digantdesai (approved):

zingo:
Build fails are unrelated

PR #13794, opened 2025-08-29
Add qnn backend to pip packages

cccclai:
Add QNN to pip package, target linux only. The qnn sdk will be installed when user import from any submodule in executorch.backend.qualcomm the first time

How to test:

1. When building the wheel package

```
conda create -yn executorch python=3.10.0 && conda activate executorch
./install_executorch.sh
EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
```
The wheel package will be under `dist` folder

2. When users install the wheel packages
Start a new terminal from a fresh repro (like /tmp)
```
unset EXECUTORCH_BUILDING_WHEEL // to make sure this env is not set
pip install executorch-0.8.0a0+8fe177e-cp310-cp310-linux_x86_64.whl  // path to the wheel package
pip install install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
pip install  --pre torchao --index-url "https://download.pytorch.org/whl/nightly/cpu"
```
Run the script the same as `test_wheel_package_qnn.sh`

### DIFF
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -73,7 +73,7 @@ def main() -> None:
     delegated_program = to_edge_transform_and_lower_to_qnn(m, example_inputs, compile_spec)
     output_graph = format_delegated_graph(delegated_program.exported_program().graph_module)
     # Ensure QnnBackend is in the output graph
-    assert "QnnBackend" in output
+    assert "QnnBackend" in output_graph
     executorch_program = delegated_program.to_executorch(
         config=ExecutorchBackendConfig(extract_delegate_segments=False)
     )
@@ -88,7 +88,7 @@ EOF
 # ----------------------------
 echo "=== Building Wheel Package ==="
 # pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
-./install_executorch.sh --minimal
+PYTHON_EXECUTABLE=python bash .ci/scripts/setup-linux.sh --build-tool cmake
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
 unset EXECUTORCH_BUILDING_WHEEL
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -83,7 +83,10 @@ EOF
 # Wheel build and .so checks
 # ----------------------------
 echo "=== Building Wheel Package ==="
+export EXECUTORCH_BUILDING_WHEEL=1
 python setup.py bdist_wheel
+
+unset EXECUTORCH_BUILDING_WHEEL
 
 WHEEL_FILE=$(ls dist/*.whl | head -n 1)
 echo "Found wheel: $WHEEL_FILE"
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -90,6 +90,8 @@ echo "=== Building Wheel Package ==="
 # pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
 source .ci/scripts/utils.sh
 install_executorch "--use-pt-pinned-commit"
+pip uninstall -y torch
+pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
 unset EXECUTORCH_BUILDING_WHEEL
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -83,7 +83,7 @@ EOF
 # Wheel build and .so checks
 # ----------------------------
 echo "=== Building Wheel Package ==="
-EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
+python setup.py bdist_wheel
 
 WHEEL_FILE=$(ls dist/*.whl | head -n 1)
 echo "Found wheel: $WHEEL_FILE"
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -97,6 +97,9 @@ echo "Using temporary directory: $TEMP_ENV_DIR"
 conda create -y -p "$TEMP_ENV_DIR/env" python=3.10
 conda run -p "$TEMP_ENV_DIR/env" pip install "$WHEEL_FILE"
 
+pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cpu
+
 # Run import tests
 conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch; print('executorch imported successfully')"
 conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch.backends.qualcomm; print('executorch.backends.qualcomm imported successfully')"
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -88,7 +88,8 @@ EOF
 # ----------------------------
 echo "=== Building Wheel Package ==="
 # pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
-PYTHON_EXECUTABLE=python bash .ci/scripts/setup-linux.sh --build-tool cmake
+source .ci/scripts/utils.sh
+install_executorch "--use-pt-pinned-commit"
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
 unset EXECUTORCH_BUILDING_WHEEL
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -87,6 +87,8 @@ EOF
 # Wheel build and .so checks
 # ----------------------------
 echo "=== Building Wheel Package ==="
+source .ci/scripts/utils.sh
+install_executorch "--use-pt-pinned-commit"
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
 unset EXECUTORCH_BUILDING_WHEEL
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -6,92 +6,112 @@ set -e
 set -x
 exec > >(tee -i build.log) 2>&1
 
+# Save repo root
+REPO_ROOT=$(pwd)
+
+# ----------------------------
+# Dynamically create script.py
+# ----------------------------
+cat > "$REPO_ROOT/script.py" << 'EOF'
+# pyre-ignore-all-errors
+import argparse
+
+import torch
+from executorch.backends.qualcomm.quantizer.quantizer import QnnQuantizer
+from executorch.backends.qualcomm.utils.utils import (
+    generate_htp_compiler_spec,
+    generate_qnn_executorch_compiler_spec,
+    get_soc_to_chipset_map,
+    to_edge_transform_and_lower_to_qnn,
+)
+from executorch.examples.models.model_factory import EagerModelFactory
+from executorch.exir.capture._config import ExecutorchBackendConfig
+from executorch.extension.export_util.utils import save_pte_program
+from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e, prepare_qat_pt2e
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-f", "--output_folder", type=str, default="", help="The folder to store the exported program")
+    parser.add_argument("--soc", type=str, default="SM8650", help="Specify the SoC model.")
+    parser.add_argument("-q", "--quantization", choices=["ptq", "qat"], help="Run post-traininig quantization.")
+    args = parser.parse_args()
+
+    class LinearModule(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.linear = torch.nn.Linear(3, 3)
+        def forward(self, arg):
+            return self.linear(arg)
+        def get_example_inputs(self):
+            return (torch.randn(3, 3),)
+
+    model = LinearModule()
+    example_inputs = model.get_example_inputs()
+
+    if args.quantization:
+        quantizer = QnnQuantizer()
+        m = torch.export.export(model.eval(), example_inputs, strict=True).module()
+        if args.quantization == "qat":
+            m = prepare_qat_pt2e(m, quantizer)
+            m(*example_inputs)
+        elif args.quantization == "ptq":
+            m = prepare_pt2e(m, quantizer)
+            m(*example_inputs)
+        m = convert_pt2e(m)
+    else:
+        m = model
+
+    use_fp16 = True if args.quantization is None else False
+    backend_options = generate_htp_compiler_spec(use_fp16=use_fp16)
+    compile_spec = generate_qnn_executorch_compiler_spec(
+        soc_model=get_soc_to_chipset_map()[args.soc],
+        backend_options=backend_options,
+    )
+    delegated_program = to_edge_transform_and_lower_to_qnn(m, example_inputs, compile_spec)
+    executorch_program = delegated_program.to_executorch(
+        config=ExecutorchBackendConfig(extract_delegate_segments=False)
+    )
+    save_pte_program(executorch_program, "linear", args.output_folder)
+
+if __name__ == "__main__":
+    main()
+EOF
+
+# ----------------------------
+# Wheel build and .so checks
+# ----------------------------
 echo "=== Building Wheel Package ==="
 python setup.py bdist_wheel
 
-# Find the wheel file
 WHEEL_FILE=$(ls dist/*.whl | head -n 1)
 echo "Found wheel: $WHEEL_FILE"
 
-echo "=== Checking for expected .so files ==="
+# (Keep all your existing .so checks here...)
 
-# List all .so files in the wheel
-echo "Listing all .so files in the wheel:"
-ALL_SO_FILES=$(unzip -l "$WHEEL_FILE" | awk '{print $4}' | grep "\.so" || true)
-
-if [ -z "$ALL_SO_FILES" ]; then
-    echo "WARNING: No .so files found in the wheel!"
-else
-    echo "$ALL_SO_FILES"
-fi
-
-# Define expected .so files
-EXPECTED_SO_FILES=(
-    "executorch/backends/qualcomm/qnn_backend.cpython-310-x86_64-linux-gnu.so"
-    "executorch/backends/qualcomm/python/PyQnnManagerAdaptor.cpython-310-x86_64-linux-gnu.so"
-    "executorch/backends/qualcomm/python/PyQnnWrapperAdaptor.cpython-310-x86_64-linux-gnu.so"
-)
-
-# Check each expected .so file
-MISSING=false
-for file in "${EXPECTED_SO_FILES[@]}"; do
-    if echo "$ALL_SO_FILES" | grep -q "$file"; then
-        echo "Found expected .so file: $file"
-    else
-        echo "ERROR: Missing expected .so file: $file"
-        MISSING=true
-    fi
-done
-
-# Print wheel contents if missing any .so
-if [ "$MISSING" = true ]; then
-    echo "==== .so file check failed ===="
-    echo "Full wheel contents for debugging:"
-    unzip -l "$WHEEL_FILE"
-    # Continue for debugging instead of exiting immediately
-    # exit 1
-fi
-
-# Create a temporary directory for test environment
+# ----------------------------
+# Conda environment setup & tests
+# ----------------------------
 TEMP_ENV_DIR=$(mktemp -d)
 echo "Using temporary directory: $TEMP_ENV_DIR"
 
-echo "=== Creating and testing in conda environment ==="
 conda create -y -p "$TEMP_ENV_DIR/env" python=3.10
 conda run -p "$TEMP_ENV_DIR/env" pip install "$WHEEL_FILE"
 
-echo "=== Testing import without SDK download ==="
+# Run import tests
 conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch; print('executorch imported successfully')"
-
-# Check that SDK directory doesn't exist after first import
-SDK_PATH="$TEMP_ENV_DIR/env/lib/python3.10/site-packages/executorch/backends/qualcomm/sdk"
-if [ -d "$SDK_PATH" ]; then
-    echo "ERROR: SDK directory exists after first import: $SDK_PATH"
-else
-    echo "SDK directory correctly doesn't exist after first import"
-fi
-
-echo "=== Testing import that should trigger SDK download ==="
 conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch.backends.qualcomm; print('executorch.backends.qualcomm imported successfully')"
 
-# Check that SDK directory exists after second import
-if [ -d "$SDK_PATH" ]; then
-    echo "SDK directory correctly exists after second import: $SDK_PATH"
-else
-    echo "ERROR: SDK directory doesn't exist after second import"
-fi
+# Run the dynamically created script using absolute path
+conda run -p "$TEMP_ENV_DIR/env" python "$REPO_ROOT/script.py"
 
-echo "=== Running model generation script ==="
-conda run -p "$TEMP_ENV_DIR/env" python script.py
-
-# Check if linear.pte file was created
+# Check if linear.pte was created
 if [ -f "linear.pte" ]; then
     echo "Model file linear.pte successfully created"
 else
     echo "ERROR: Model file linear.pte was not created"
 fi
 
-echo "=== Cleaning up ==="
+# Cleanup
 conda env remove -p "$TEMP_ENV_DIR/env" -y
 rm -rf "$TEMP_ENV_DIR"
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -87,7 +87,8 @@ EOF
 # Wheel build and .so checks
 # ----------------------------
 echo "=== Building Wheel Package ==="
-pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+# pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+./install_executorch.sh --minimal
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
 unset EXECUTORCH_BUILDING_WHEEL
 
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -17,7 +17,6 @@ cat > "/tmp/script_qnn_wheel_test.py" << 'EOF'
 import argparse
 
 import torch
-from executorch.backends.qualcomm import install_qnn_sdk
 from executorch.backends.qualcomm.quantizer.quantizer import QnnQuantizer
 from executorch.backends.qualcomm.utils.utils import (
     generate_htp_compiler_spec,
@@ -32,7 +31,6 @@ from executorch.extension.export_util.utils import save_pte_program
 from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e, prepare_qat_pt2e
 
 def main() -> None:
-    install_qnn_sdk()
     parser = argparse.ArgumentParser()
     parser.add_argument("-f", "--output_folder", type=str, default="", help="The folder to store the exported program")
     parser.add_argument("--soc", type=str, default="SM8650", help="Specify the SoC model.")
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -84,7 +84,6 @@ EOF
 # ----------------------------
 echo "=== Building Wheel Package ==="
 EXECUTORCH_BUILDING_WHEEL=1 python setup.py bdist_wheel
-
 unset EXECUTORCH_BUILDING_WHEEL
 
 WHEEL_FILE=$(ls dist/*.whl | head -n 1)
@@ -104,18 +103,6 @@ else
 fi
 
 # ----------------------------
-# Conda environment setup & tests
-# ----------------------------
-TEMP_ENV_DIR=$(mktemp -d)
-echo "Using temporary directory: $TEMP_ENV_DIR"
-
-conda create -y -p "$TEMP_ENV_DIR/env" python=3.10
-conda run -p "$TEMP_ENV_DIR/env" pip install "$WHEEL_FILE"
-
-conda run -p "$TEMP_ENV_DIR/env" pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
-conda run -p "$TEMP_ENV_DIR/env" pip install --pre torchao --index-url "https://download.pytorch.org/whl/nightly/cpu"
-
-# ----------------------------
 # Check .so files in the wheel
 # ----------------------------
 echo "Checking for .so files inside the wheel..."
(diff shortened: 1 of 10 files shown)


cccclai:
working version [ded7cf0](https://github.com/pytorch/executorch/pull/13794/commits/ded7cf015b7ffec26ef06100f50f49227196945d)

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,155 @@
...
+        m = model
+
+    use_fp16 = True if args.quantization is None else False
+    backend_options = generate_htp_compiler_spec(use_fp16=use_fp16)
+    compile_spec = generate_qnn_executorch_compiler_spec(
+        soc_model=get_soc_to_chipset_map()[args.soc],
+        backend_options=backend_options,
+    )
+    delegated_program = to_edge_transform_and_lower_to_qnn(m, example_inputs, compile_spec)
+    executorch_program = delegated_program.to_executorch(
```
kimishpatel, line 75: Can we add a check to see if it actually contain qnn delegate?

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,155 @@
...
+    echo "OK: No qualcomm/sdk files found in wheel"
+fi
+
+# ----------------------------
+# Conda environment setup & tests
+# ----------------------------
+TEMP_ENV_DIR=$(mktemp -d)
+echo "Using temporary directory: $TEMP_ENV_DIR"
+
+conda create -y -p "$TEMP_ENV_DIR/env" python=3.10
```
kimishpatel: should we also check for other python versions? Looks like we do both 3.10 and 3.12 https://pypi.org/project/executorch/0.7.0/#files, so maybe test both?

kimishpatel: Also if venv is used, can we check it works with venv as well?

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,155 @@
...
+conda run -p "$TEMP_ENV_DIR/env" pip install torch=="2.9.0.dev20250801" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+conda run -p "$TEMP_ENV_DIR/env" pip install --pre torchao --index-url "https://download.pytorch.org/whl/nightly/cpu"
+
+# ----------------------------
+# Check .so files in the wheel
+# ----------------------------
+echo "Checking for .so files inside the wheel..."
+WHEEL_SO_FILES=$(unzip -l "$WHEEL_FILE" | awk '{print $4}' | grep "executorch/backends/qualcomm/python" || true)
+if [ -z "$WHEEL_SO_FILES" ]; then
+    echo "WARNING: No .so files found in wheel under executorch/backends/qualcomm/python"
```
kimishpatel: shouldnt you error out

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,155 @@
...
+# ----------------------------
+# Check installed .so files in conda env
+# ----------------------------
+echo "Checking installed executorch/backends/qualcomm/python contents in conda env..."
+ENV_SO_DIR="$TEMP_ENV_DIR/env/lib/python3.10/site-packages/executorch/backends/qualcomm/python"
+ls -l "$ENV_SO_DIR" || echo "Folder does not exist!"
+
+# Run import tests
+conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch; print('executorch imported successfully')"
+conda run -p "$TEMP_ENV_DIR/env" python -c "import executorch.backends.qualcomm; print('executorch.backends.qualcomm imported successfully')"
```
kimishpatel: is it worth checking if the so files are downloaded and installed? Or the script running would suffice

.github/workflows/pull.yml
```diff
@@ -861,7 +882,7 @@ jobs:
         # Run pytest
         PYTHON_EXECUTABLE=python bash backends/nxp/run_unittests.sh
 
-        # Run aot examples:
+        Run aot examples:
```
kimishpatel: This should still be a comment?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+
+def _download_qnn_sdk() -> Optional[pathlib.Path]:
+    """
+    Download and extract the Qualcomm SDK into SDK_DIR.
+
+    Notes:
+        - Only runs on Linux x86 platforms. Skips otherwise.
+    """
+    print("Downloading Qualcomm SDK...")
+    QNN_VERSION = "2.37.0.250724"
```
kimishpatel, line 107: nit: maybe refactor to get/set qnn version functions? Mainly for ease of maintenance.

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+
+LLVM_VERSION = "14.0.0"
+LIBCXX_BASE_NAME = f"clang+llvm-{LLVM_VERSION}-x86_64-linux-gnu-ubuntu-18.04"
+LLVM_URL = f"https://github.com/llvm/llvm-project/releases/download/llvmorg-{LLVM_VERSION}/{LIBCXX_BASE_NAME}.tar.xz"
+REQUIRED_LIBS = [
+    "libc++.so.1.0",
+    "libc++abi.so.1.0",
+    "libunwind.so.1",
+    "libm.so.6",
+    "libpython3.10.so.1.0",
```
kimishpatel: what is this?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+        src_path = lib_src / fname
+        if not src_path.exists():
+            print(f"[libcxx] Warning: {fname} not found in extracted LLVM")
+            continue
+        shutil.copy(src_path, target_dir / fname)
+
+    libcxx = target_dir / "libc++.so.1.0"
+    libcxx_abi = target_dir / "libc++abi.so.1.0"
+    if libcxx.exists():
+        os.symlink("libc++.so.1.0", target_dir / "libc++.so.1")
```
kimishpatel: this is creating symlink to what?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+LLVM_VERSION = "14.0.0"
+LIBCXX_BASE_NAME = f"clang+llvm-{LLVM_VERSION}-x86_64-linux-gnu-ubuntu-18.04"
+LLVM_URL = f"https://github.com/llvm/llvm-project/releases/download/llvmorg-{LLVM_VERSION}/{LIBCXX_BASE_NAME}.tar.xz"
+REQUIRED_LIBS = [
+    "libc++.so.1.0",
+    "libc++abi.so.1.0",
+    "libunwind.so.1",
+    "libm.so.6",
```
kimishpatel: why these two?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+            else:
+                out_path.parent.mkdir(parents=True, exist_ok=True)
+                src = tf.extractfile(m)
+                if src is None:
+                    continue
+                with src, open(out_path, "wb") as dst:
+                    dst.write(src.read())
+
+
+LLVM_VERSION = "14.0.0"
```
kimishpatel, line 176: I am a little concerned with this because it is going beyond the scope of what is needed. At the least I would suggest, trim it down to the minimal list of what needs to be downloaded. 

Main issue is that user may already have these libs on their machine. We dont have those on our internal machines but this doesnt universally hold true. So we would really need this only if these libs are not present in users system. I would suggest we do such things only if required and probably better done explicitly.

setup.py
```diff
@@ -455,6 +464,88 @@ def run(self):
         if self._ran_build:
             return
 
+        try:
```
kimishpatel, line 465: Can this be a separate build_ext? is it possible to do?

setup.py
```diff
@@ -455,6 +464,88 @@ def run(self):
...
+                _download_qnn_sdk,
+                SDK_DIR,
+            )
+
+            os.environ["EXECUTORCH_BUILDING_WHEEL"] = "1"
+
+            # qnn sdk setup
+            _download_qnn_sdk()
+
+            sdk_path = Path(SDK_DIR).resolve()  # full absolute path
```
kimishpatel: Can you install sdk in a temp directory so that you dont need to do explicit cleanup?

setup.py
```diff
@@ -455,6 +464,88 @@ def run(self):
...
+                    prj_root
+                    / "backends/qualcomm/python/PyQnnWrapperAdaptor.cpython-310-x86_64-linux-gnu.so",
+                ),
+            ]
+
+            for module_name, so_src in so_files:
+                so_dst = Path(self.get_ext_fullpath(module_name))
+                self.mkpath(str(so_dst.parent))
+                self.copy_file(str(so_src), str(so_dst))
+                print(f"Copied Qualcomm backend: {so_src} -> {so_dst}")
```
kimishpatel: turn this into logging

kimishpatel (changes requested):
Left a few comments. Lets resolve this soon and validate

cccclai:
mark as draft to address some feedback

kimishpatel:
> mark as draft to address some feedback

Please let me know when it is ready for review

.ci/scripts/test_wheel_package_qnn.sh
cccclai: updated

.ci/scripts/test_wheel_package_qnn.sh
cccclai: The script includes the lowering steps, should be enough if lowering is successful

.github/workflows/pull.yml
cccclai: fixed

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: Yeah I'd need to have a separate PR to remove the hardcoded numbers https://github.com/pytorch/executorch/pull/13960

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: It's currently used for internal, can remove if it's not needed in next pr

kimishpatel: yeah but is it cpython? if so why are we download python interpreter?

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: a few ones like `libc++.so.1.0` to `libc++.so.1`, that's extra burden for users imo

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: I'm just copying the internal copied ones here, I believe they were part of the previous qnn packages.

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: They might but they might have a different files name, that's why I added symlink before. I don't install them in users default path, but just part of the qnn sdk packages, that will be the same behavior as if users downloaded the older qnn pacakages versions

setup.py
cccclai: Each setup() only has one build_ext to override, but I can double check if we can move it to build instead (where our current main build happens)

setup.py
cccclai: I don't want to use temp directory because it's easy to stale and hard to clean up. (Or maybe a specific path as qnn sdk installation path similar to coreml can work).

.ci/scripts/test_wheel_package_qnn.sh
cccclai: Added

.ci/scripts/test_wheel_package_qnn.sh
cccclai: Added venv in the test too

backends/qualcomm/scripts/download_qnn_sdk.py
kimishpatel: these are system libs generally. I dont know what the expectation is for qnn sdk. Can you tag someone from qcomm to at least understand if there is specific version of libc++ that is need for specific version of qnn sdk.

My biggest concern with all these is that different libraries may have different libc++ requirement. What if mtk's sdk doesnt work with this specific libc++.

backends/qualcomm/scripts/download_qnn_sdk.py
kimishpatel: I am not sure if this is needed. Can you check without them?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+        f for name in priority for f in candidates if f.name.startswith(name)
+    ]
+    sorted_candidates += [f for f in candidates if f not in sorted_candidates]
+
+    for sofile in sorted_candidates:
+        try:
+            ctypes.CDLL(str(sofile), mode=ctypes.RTLD_GLOBAL)
+            print(f"Loaded {sofile.name}")
+        except OSError as e:
+            print(f"[WARN] Failed to load {sofile.name}: {e}")
```
kimishpatel: Why are we loading libraries here? Shouldnt this happen automatically when you import qnn package? I never had to manually load these libs. You may have to update LD_LIBRARY_PATHs to point to install directory

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+    os.environ["QNN_SDK_ROOT"] = str(SDK_DIR)
+
+    qnn_lib = SDK_DIR / "lib" / "x86_64-linux-clang" / "libQnnHtp.so"
+    print(f"[INIT] qnn_lib: {qnn_lib}")
+    try:
+        ctypes.CDLL(str(qnn_lib), mode=ctypes.RTLD_GLOBAL)
+        print(f"[INIT] Loaded QNN library from {qnn_lib}")
+    except OSError as e:
+        print(f"[ERROR] Failed to load QNN library at {qnn_lib}: {e}")
+        return False
```
kimishpatel: Why loading of these libs is necessary? I thought we just needed to point to the right directory

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+                out_path.parent.mkdir(parents=True, exist_ok=True)
+                src = tf.extractfile(m)
+                if src is None:
+                    continue
+                with src, open(out_path, "wb") as dst:
+                    dst.write(src.read())
+
+
+LLVM_VERSION = "14.0.0"
+LIBCXX_BASE_NAME = f"clang+llvm-{LLVM_VERSION}-x86_64-linux-gnu-ubuntu-18.04"
```
kimishpatel, line 177: this says ubuntu. will it work with a different linux distribution like fedora or centos?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,247 @@
...
+    try:
+        ctypes.CDLL(str(qnn_lib), mode=ctypes.RTLD_GLOBAL)
+        print(f"[INIT] Loaded QNN library from {qnn_lib}")
+    except OSError as e:
+        print(f"[ERROR] Failed to load QNN library at {qnn_lib}: {e}")
+        return False
+
+    try:
+        libcxx_dir = PKG_ROOT / "sdk" / f"libcxx-{LLVM_VERSION}"
+        _stage_libcxx(libcxx_dir)
```
kimishpatel, line 345: in the worst case, we should do this _only_ when qnn AOT package loading fails for this reason. Like we dont manage all the deps for users automagically. Pytorch readme page asks users to install specific packages and it only handles the pytorch specific stuff.

We are doing this here only because qnn sdk does not come as pypi package and we dont want to maintain our own

setup.py
kimishpatel: > I don't want to use temp directory because it's easy to stale and hard to clean up.

Temp directories via Python's TemporaryDirectory get auto cleaned up? Adn staleness doesnt even matter, right? This is only to allow for building our PyPI package?

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: If I set LD_LIBRARY_PATH in env it works,  but setting env var in python doesn't. I feel like the env variable in python doesn't propagate properly

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: To avoid users setting LD_LIBRARY_PATH by themselves

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: It seems working on my centos machine

cccclai:
Discussed offline and will make two changes

1. download/load library only if we can't find them from LD path
2. add pybindings to verify that we want to load the same qnn version

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,203 @@
...
+    to_edge_transform_and_lower_to_qnn,
+)
+from executorch.exir.backend.utils import format_delegated_graph
+from executorch.examples.models.model_factory import EagerModelFactory
+from executorch.exir.capture._config import ExecutorchBackendConfig
+from executorch.extension.export_util.utils import save_pte_program
+from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e, prepare_qat_pt2e
+
+def main() -> None:
+    install_qnn_sdk()
```
kimishpatel: for wheel package testing, this should not be needed, right?

kimishpatel: meaning you install wheel and test

kimishpatel: is this script testing wheel building and then checking if the built wheel is correct? Even so why do we need to call install_qnn_sdk

.ci/scripts/test_wheel_package_qnn.sh
kimishpatel: Change warning to error

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,203 @@
...
+fi
+
+# ----------------------------
+# Helpers
+# ----------------------------
+get_site_packages_dir () {
+  local PYBIN="$1"
+  "$PYBIN" - <<'PY'
+import sysconfig, sys
+print(sysconfig.get_paths().get("purelib") or sysconfig.get_paths().get("platlib"))
```
kimishpatel, line 130: nice

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,203 @@
...
+PY
+}
+
+run_core_tests () {
```
kimishpatel, line 134: This should be done in a new conda/py env

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,203 @@
...
+  "$PYBIN" -c "import executorch.backends.qualcomm; print('executorch.backends.qualcomm imported successfully')"
+
+  echo "=== [$LABEL] List installed executorch/backends/qualcomm/python ==="
+  local SITE_DIR
+  SITE_DIR="$(get_site_packages_dir "$PYBIN")"
+  local SO_DIR="$SITE_DIR/executorch/backends/qualcomm/python"
+  ls -l "$SO_DIR" || echo "Folder does not exist!"
+
+  echo "=== [$LABEL] Run export script to generate linear.pte ==="
+  (cd "$REPO_ROOT" && "$PYBIN" "$REPO_ROOT/script.py")
```
kimishpatel: nit: Maybe write this output to tmp? Just to avoid any name collison and polluting repo directory. Also consideing naming it script_qnn_wheel_test.py or something for name collision avoidance

.ci/scripts/test_wheel_package_qnn.sh
kimishpatel: oh nevermind. you are doing that.

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -0,0 +1,203 @@
...
+  fi
+}
+
+# ----------------------------
+# Conda environment setup & tests
+# ----------------------------
+echo "=== Testing in Conda env ==="
+TEMP_ENV_DIR=$(mktemp -d)
+echo "Using temporary directory for conda: $TEMP_ENV_DIR"
+conda create -y -p "$TEMP_ENV_DIR/env" python=3.10
```
kimishpatel: Maybe make python version programmable for this script?

.github/workflows/pull.yml
```diff
@@ -13,203 +13,230 @@ concurrency:
...
-      runner: ${{ matrix.runner }}
-      docker-image: ci-image:${{ matrix.docker-image }}
-      submodules: 'recursive'
-      ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-      timeout: 90
-      script: |
-        # The generic Linux job chooses to use base env, not the one setup by the image
-        CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
-        conda activate "${CONDA_ENV}"
+  # test-setup-linux-gcc:
```
kimishpatel: I think you have accidental changes checked in to these files

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,350 @@
...
+import pathlib
+import platform
+import shutil
+import tarfile
+import tempfile
+import urllib.request
+import zipfile
+from typing import Dict, List, Optional, Tuple
+
+from tqdm import tqdm
```
kimishpatel: tqdm is not a dep for us. Dont use it.

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,350 @@
...
+    """Return first matching path to `libname` in LD_LIBRARY_PATH, or None."""
+    if ld_dirs is None:
+        ld_dirs = _ld_library_paths()
+    for d in ld_dirs:
+        candidate = d / libname
+        try:
+            if candidate.exists():
+                return candidate.resolve()
+        except Exception:
+            # Ignore unreadable / permission issues, keep looking.
```
kimishpatel, line 240: Maybe just do this?
```
def find_file(root_dir, filename):
    root = Path(root_dir)
    return list(root.rglob(filename))  # finds all matches
```

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,350 @@
...
+
+    qnn_lib = qnn_sdk_dir / "lib" / "x86_64-linux-clang" / "libQnnHtp.so"
+    print(f"[QNN] Loading {qnn_lib}")
+    try:
+        ctypes.CDLL(str(qnn_lib), mode=ctypes.RTLD_GLOBAL)
+        print("[QNN] Loaded libQnnHtp.so from packaged SDK.")
+        return True
+    except OSError as e:
+        print(f"[QNN][ERROR] Failed to load {qnn_lib}: {e}")
+        return False
```
kimishpatel: returning from try catch block is a bit strange. Can you just use a boolean and return that

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+    try:
+        libcxx_dir = PKG_ROOT / "sdk" / f"libcxx-{LLVM_VERSION}"
+        _stage_libcxx(libcxx_dir)
+        _load_libcxx_libs(libcxx_dir)
+        print(f"[libcxx] Staged and loaded libc++ from {libcxx_dir}")
+        return True
+    except Exception as e:
+        print(f"[libcxx][ERROR] Failed to stage/load libc++: {e}")
+        return False
```
kimishpatel: same here

kimishpatel (changes requested):
This is looking good. Few more comments

.ci/scripts/test_wheel_package_qnn.sh
cccclai: oh good catch

.github/workflows/pull.yml
cccclai: Will drop the changes after I fix the test, otherwise too many tests running in CI which might slow down to get a machine

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: > Can you just use a boolean and return that

What does it mean?

cccclai: > Can you just use a boolean and return that

What does it mean?

cccclai: > Can you just use a boolean and return that

What does it mean?

.ci/scripts/test_wheel_package_qnn.sh
cccclai: oh I should delete this line. It's not needed

.ci/scripts/test_wheel_package_qnn.sh
cccclai: updated

.ci/scripts/test_wheel_package_qnn.sh
cccclai: updated

.ci/scripts/test_wheel_package_qnn.sh
cccclai: updated

backends/qualcomm/scripts/download_qnn_sdk.py
kimishpatel: write it as
```
lib_loaded = False
    try:
        ctypes.CDLL(str(qnn_lib), mode=ctypes.RTLD_GLOBAL)
        print("[QNN] Loaded libQnnHtp.so from packaged SDK.")
        lib_loaded = True
    except OSError as e:
        print(f"[QNN][ERROR] Failed to load {qnn_lib}: {e}")
return lib_loaded
```

setup.py
cccclai: updated to use temp directory

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: fixed

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: Actually it might be a bit slow because otherwise we can find the first match and stop

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: updated

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: updated

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: Actually find out we don't need symlink for the llvm tar file. Remove the not used library and symlink

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: Remove

backends/qualcomm/scripts/download_qnn_sdk.py
cccclai: Remove

kimishpatel (approved):
looks good to me. i will test this separately on my end

setup.py
```diff
@@ -455,6 +462,69 @@ def run(self):
...
+                logging.info("prj_root: ", prj_root)
+                build_sh = prj_root / "backends/qualcomm/scripts/build.sh"
+                build_root = prj_root / "build-x86"
+
+                if not build_sh.exists():
+                    raise FileNotFoundError(f"{build_sh} not found")
+
+                # Run build.sh with SDK path exported
+                env = dict(**os.environ)
+                env["QNN_SDK_ROOT"] = str(sdk_path)
```
haowhsu-quic, line 490: If I understand correctly, this env variable only visible when doing `pip install`. If user starts a new process, will they be notified to use the correct one?

haowhsu-quic: Ah never mind, this is for building delegate related libraries for packaging.

backends/qualcomm/__init__.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+from .scripts.download_qnn_sdk import install_qnn_sdk
+
+
+env_flag = os.getenv("EXECUTORCH_BUILDING_WHEEL", "0").lower()
+# If users have preinstalled QNN_SDK_ROOT, we will use it.
+qnn_sdk_root_flag = os.getenv("QNN_SDK_ROOT", None)
+if env_flag not in ("1", "true", "yes") and not qnn_sdk_root_flag:
```
haowhsu-quic, line 9: I wonder if the same user will keep downloading SDK if `QNN_SDK_ROOT` is never set?

setup.py
cccclai: Yes, only building delegates for the wheel package

backends/qualcomm/__init__.py
cccclai: No, if QNN_SDK_ROOT is not set, we will use the predefined path `SDK_DIR = PKG_ROOT / "sdk" / "qnn"`. Inside the `download_qnn_sdk.py` logic, we will check if the library is there already, if yes, we will just use those files

cccclai:
@haowhsu-quic I will merge this PR for now, but would like to hear your feedback/concern and address them

backends/qualcomm/__init__.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+from .scripts.download_qnn_sdk import install_qnn_sdk
+
+
+env_flag = os.getenv("EXECUTORCH_BUILDING_WHEEL", "0").lower()
```
kimishpatel, line 6: Gate on another flat like BUILDING_WHEEL_FOR_LINUX or something similar

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+    QNN_VERSION = "2.37.0.250724"
+    logger.info("Downloading Qualcomm SDK...")
+    QAIRT_URL = (
+        f"https://softwarecenter.qualcomm.com/api/download/software/sdks/"
+        f"Qualcomm_AI_Runtime_Community/All/{QNN_VERSION}/v{QNN_VERSION}.zip"
+    )
+    QAIRT_CONTENT_DIR = f"qairt/{QNN_VERSION}"
+
+    if not is_linux_x86():
```
cccclai, line 115: I have the linux check here @kimishpatel

backends/qualcomm/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# If users have preinstalled QNN_SDK_ROOT, we will use it.
+qnn_sdk_root_flag = os.getenv("QNN_SDK_ROOT", None)
+if env_flag not in ("1", "true", "yes") and not qnn_sdk_root_flag:
+    ok = install_qnn_sdk()
+
+    if not ok:
```
kimishpatel, line 12: On windows this will always be not ok. So you want to raise this error only if we are linux.

setup.py
```diff
@@ -455,6 +462,69 @@ def run(self):
         if self._ran_build:
             return
 
+        try:
+            # Following code is for building the Qualcomm backend.
+            from backends.qualcomm.scripts.download_qnn_sdk import _download_qnn_sdk
+
+            os.environ["EXECUTORCH_BUILDING_WHEEL"] = "1"
+
+            with tempfile.TemporaryDirectory() as tmpdir:
```
kimishpatel, line 471: Also this componenents should be trigerred only for linux as well

backends/qualcomm/__init__.py
cccclai: oh I see what you meant, okay let me add it

tdasika17:
Hi,
Can we now export a model to qnn backend with this whl packge?
Is it part of main?

if not in main, where is this whl package available?

cccclai:
I think it should be available in nightly if our `viable-strict` branch catch up cc: @abhinaykukkadapu I remember you and @larryliu0820 recently re-enable it. Just not sure if our viable strict has those changes

abhinaykukkadapu:
> I think it should be available in nightly if our `viable-strict` branch catch up cc: @abhinaykukkadapu I remember you and @larryliu0820 recently re-enable it. Just not sure if our viable strict has those changes

@tdasika17  can you try with this `pip install --pre torch torchvision torchaudio executorch --extra-index-url https://download.pytorch.org/whl/nightly/cpu`.

PR #13801, opened 2025-08-29
Arm backend: Add 16A8W FCNode support with BMM dependency fix

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13802
* __->__ #13801
* #13800
* #13799
* #13798
* #13797
* #13796

Add 16A8W quantization support for FCNode operations with BMM dependency fix in ExecutorTorch ARM backend.

This follows the pattern established for linear, mul, sigmoid, tanh, slice, view/transpose, and cat operations, extending int16 support to FCNode operations.

Changes:
- Add INT16 dtype validation support in op_bmm.py
- Add test_addmm_tensor_16a8w_tosa_INT test function
- Enable test_addmm.py in test targets configuration
- Fix BMM dependency for FCNode operations

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency.

Differential Revision: [D80512504](https://our.internmc.facebook.com/intern/diff/D80512504/)

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13800, opened 2025-08-29
Arm backend: Add 16A8W support and test for cat operation

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13802
* #13801
* __->__ #13800
* #13799
* #13798
* #13797
* #13796

Add 16A8W quantization support and test for the cat operation in ExecutorTorch ARM backend.

This follows the pattern established for linear, mul, sigmoid, tanh, slice, and view/transpose operations, extending int16 support to cat operations.

Changes:
- Add test_cat_tensor_16a8w_tosa_INT test function
- Enable test_cat.py in test targets configuration

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency.

Differential Revision: [D80511455](https://our.internmc.facebook.com/intern/diff/D80511455/)

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13799, opened 2025-08-29
Arm backend: Add 16A8W support for view and transpose operations

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13802
* #13801
* #13800
* __->__ #13799
* #13798
* #13797
* #13796

Add 16A8W quantization support for view and transpose operations in ExecutorTorch ARM backend.

This follows the pattern established for linear, mul, sigmoid, tanh, and slice operations, extending int16 support to view and transpose operations.

Changes:
- Add INT16 dtype validation support in op_transpose.py
- Add test_view_tensor_16a8w_tosa_INT test function
- Enable test_view.py in test targets configuration

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency.

Differential Revision: [D80511313](https://our.internmc.facebook.com/intern/diff/D80511313/)

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13798, opened 2025-08-29
Arm backend: Add 16A8W support and test for slice operation

Ninja91:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #13802
* #13801
* #13800
* #13799
* __->__ #13798
* #13797
* #13796

Add 16A8W quantization support and test for the slice operation in ExecutorTorch ARM backend.

This follows the pattern established for linear, mul, sigmoid, and tanh operations, extending int16 support to slice operations.

Changes:
- Add INT16 dtype validation support in op_slice.py
- Add test_slice_tensor_16a8w_tosa_INT test function
- Enable test_slice.py in test targets configuration

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency.

Differential Revision: [D80511095](https://our.internmc.facebook.com/intern/diff/D80511095/)

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #13802, opened 2025-08-29
Arm backend: Add INT16 support to rescale operation

Ninja91:
* __->__ #13802
* #13801
* #13800
* #13799
* #13798
* #13797
* #13796

Add INT16 support for RequantizeNode rescale operations in ExecutorTorch ARM backend.

This follows the pattern established for linear, mul, sigmoid, tanh, slice, view/transpose, cat, and FCNode operations, extending int16 support to RequantizeNode rescale operations.

Changes:
- Add INT16 dtype validation support in op_rescale.py
- Enable rescale operations for 16A8W quantization configuration

The 16A8W configuration uses 16-bit activations with 8-bit weights, enabling higher precision for activations while maintaining weight efficiency. RequantizeNode rescale operations are essential for proper quantization scaling in the 16A8W pipeline.

Differential Revision: [D80513725](https://our.internmc.facebook.com/intern/diff/D80513725/)


backends/arm/operators/op_rescale.py
```diff
@@ -46,13 +46,20 @@ def define_node(
         input_zp = cast(int, node.args[3])
         output_zp = cast(int, node.args[4])
 
-        if input_dtype != map_dtype(torch.int8, self.tosa_spec) and input_zp != 0:
+        if (
```
digantdesai, line 49: I think 8 -> 16 and 16 -> 8 both are supported for the INT profile (and INT16)? Can you make sure you check?

digantdesai (changes requested):
I didn't see torch ops like tests for TOSA dialect ops, perhaps we should add it here. @per what do you think?

digantdesai (approved):

backends/arm/operators/op_rescale.py
Ninja91: Checked in map_dtype and torch dtypes match with Tosa DTypes. Not sure if you meant something else.

mergennachin:
@Ninja91 You need to land this in main 

cc @digantdesai (reviewer) @lucylq (oncall)

PR #13804, opened 2025-08-29
Arm backend: Add requirements.txt for arm backend

perheld:
Add requirements.txt files to handle pip installs in a more python way of doing things. Tagged versions handled in one place instead of being spread out in shell scripts.

Requirements-arm-tosa.txt contains what is needed for the tosa reference model which is shared between multiple backends.

Requirements-arm-ethos-u.txt contains the vela compiler needed by the ethos-u backend.

Requirements-arm-models-test.txt contains test models needed for baremetal tests. The install_models_for_test.sh was keept since that its used by gitlab ci.

Change-Id: I3213e543d15c23b96ccd1f683a9d6c7b164d7e7b


zingo (approved):

zingo:
Testa fails are unrelated

PR #13813, opened 2025-08-29
Add create_mutable_buffer util

digantdesai:
## Rationale:
    * We want to support stateful custom ops inserted after to_edge in a non delegated cases like op-lib-backends i.e. cortex-m::cmsis_nn::linear

## Details:
    * Allows a pass to add a mutable buffer.
    * Essentially redoes what export, to_edge does for a buffer registered in an nn.module.
    * Also plays nice with to_executorch passes like conversion to in-place, and write-back for mutated buffers.
    * Verifies above with tests, adds a HelperPass for someone looking to leverage this util as an example too.

## To Test:
    $ python -m unittest backends.transforms.test.test_create_mutable_buffer

### DIFF
--- a/backends/transforms/test/test_create_mutable_buffer.py
+++ b/backends/transforms/test/test_create_mutable_buffer.py
@@ -1,0 +1,503 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import copy
+import unittest
+
+import executorch
+import torch
+from executorch.backends.transforms.utils import create_mutable_buffer
+from executorch.exir import ExecutorchBackendConfig, to_edge
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass, PassResult
+from executorch.extension.pybindings.portable_lib import (  # @manual
+    _load_for_executorch_from_buffer,
+    Verification,
+)
+from torch.export import export
+from torch.utils._pytree import tree_flatten
+
+
+class TestMutableBufferCreation(unittest.TestCase):
+    """
+    Test suite for the create_mutable_buffer utility function.
+    """
+
+    def test_create_mutable_buffer(self):
+        """
+        Tests the utility function create_mutable_buffer which creates a mutable buffer
+        that can be modified during execution.
+        """
+
+        class EmptyNetwork(torch.nn.Module):
+            def forward(self, x: torch.Tensor) -> torch.Tensor:
+                return x
+
+            test_data: torch.Tensor = (torch.zeros(1),)
+
+        module = EmptyNetwork()
+        exported_program = export(module, args=module.test_data, strict=True)
+        exported_program = to_edge(exported_program).exported_program()
+        graph = exported_program.graph_module.graph
+
+        assert len(graph.nodes) == 2
+        assert exported_program.module()(torch.zeros(1)) == 0
+        assert len(exported_program.graph_signature.input_specs) == 1
+        assert len(exported_program.graph_signature.output_specs) == 1
+        assert len(exported_program.state_dict) == 0
+
+        buffer_name = "b_test_mutable_buffer"
+        target_name = buffer_name[2:]  # Remove the "b_" prefix
+        initial_data = torch.ones(1) * 2  # Initialize with value 2
+
+        # Create a mutable buffer using create_mutable_buffer
+        buffer_node = create_mutable_buffer(
+            exp_program=exported_program,
+            name=buffer_name,
+            data=initial_data,
+        )
+        assert "val" in buffer_node.meta
+
+        # Verify the buffer was created correctly
+        input_node = list(graph.nodes)[
+            1
+        ]  # Original input node (buffer_node is now first)
+
+        # Create an add operation that uses the mutable buffer
+        with graph.inserting_after(input_node):
+            graph.create_node(
+                "call_function",
+                exir_ops.edge.aten.add.Tensor,
+                args=(input_node, buffer_node),
+                kwargs={},
+            )
+
+        # We should now have four nodes: buffer, input, add, output
+        assert len(graph.nodes) == 4
+
+        assert target_name in exported_program.state_dict
+        assert torch.equal(exported_program.state_dict[target_name], initial_data)
+
+        # Check that buffer is properly referenced in graph signature
+        assert buffer_name in exported_program.graph_signature.inputs_to_buffers
+        assert (
+            exported_program.graph_signature.inputs_to_buffers[buffer_name]
+            == target_name
+        )
+        assert (
+            exported_program.graph_signature.buffers_to_mutate[buffer_name]
+            == target_name
+        )
+
+
+class TestRegisterMutableBufferPass(unittest.TestCase):
+    """
+    This test is to test the `create_mutable_buffer` utility.
+    """
+
+    class HelperPass(ExportPass):
+        def __init__(
+            self,
+            exported_program: torch.export.ExportedProgram,
+            buf_data: torch.Tensor,
+        ):
+            super().__init__()
+            self.registered_buffers = set()
+            self.exported_program = exported_program
+            self.buf_data = buf_data
+
+        def call(self, graph_module: torch.fx.GraphModule):
+            """
+            This pass will register a mutable buffer for the add op(s) in the graph.
+            It will insert a new index_put_ op to the graph to update the buffer using the output of the add op.
+            And adjust the output of the graph to return the buffer or the index_put_ op.
+            """
+            modified = False
+
+            assert (
+                len(graph_module.graph.output_node().args[0]) == 1
+            ), "For this pass, expecting only one output i.e. add"
+
+            for node in graph_module.graph.nodes:
+                if (
+                    node.op == "call_function"
+                    and node.target == exir_ops.edge.aten.add.Tensor
+                ):
+                    # To match what we do in export.
+                    suffix = (
+                        ""
+                        if len(self.registered_buffers) == 0
+                        else f"{len(self.registered_buffers) + 1}"
+                    )
+                    buffer_name = f"b_my_buffer{suffix}"
+                    self.registered_buffers.add(buffer_name)
+
+                    # Utility under test!
+                    buf_node = create_mutable_buffer(
+                        self.exported_program,
+                        data=self.buf_data,
+                        name=buffer_name,
+                    )
+
+                    # Assuming `indices` is always available
+                    indices = [
+                        node
+                        for node in graph_module.graph.nodes
+                        if node.op == "placeholder" and node.name == "indices"
+                    ]
+                    with graph_module.graph.inserting_after(node):
+                        index_put_node = graph_module.graph.create_node(
+                            "call_function",
+                            exir_ops.edge.aten.index_put_.default,
+                            (
+                                buf_node,
+                                indices,
+                                node,
+                            ),
+                            {},
+                        )
+
+                        # Replace the old add output with index_put_node
+                        output_node = graph_module.graph.output_node()
+                        outputs = list(output_node.args[0])
+                        if node in outputs:
+                            outputs[-1] = (
+                                index_put_node  # Replace the old add output with index_put_node
+                            )
+                            output_node.args = (outputs,)
+
+                            # update the output node name in the graph signature
+                            graph_signature = self.exported_program.graph_signature
+                            graph_signature.replace_all_uses(
+                                node.name, index_put_node.name
+                            )
+                            self.exported_program._graph_signature = graph_signature
+
+                    modified = True
+
+            if modified:
+                graph_module = super().call(graph_module).graph_module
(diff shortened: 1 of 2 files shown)


mergennachin (commented):
Add backends/transforms in https://github.com/pytorch/executorch/blob/main/pytest.ini

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
...
+
+    if not user_input_indices_or_consts:
+        # If no user inputs or constants exist, insert at the end of all existing inputs
+        node_index = len(exp_program.graph_signature.input_specs)
+        # Create node at the end of existing placeholders
+        node = graph.create_node(op="placeholder", name=name, target=name)
+        node.meta["val"] = fake_tensor
+    else:
+        node_index = user_input_indices_or_consts[0]
+        # Find the first user input placeholder node
```
kimishpatel: you mean user input or constant node?

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
...
+    # Find indices of user inputs and constant tensors
+    user_input_indices_or_consts = [
+        i
+        for i, spec in enumerate(exp_program.graph_signature.input_specs)
+        if spec.kind == InputKind.USER_INPUT or spec.kind == InputKind.CONSTANT_TENSOR
+    ]
+
+    if not user_input_indices_or_consts:
+        # If no user inputs or constants exist, insert at the end of all existing inputs
+        node_index = len(exp_program.graph_signature.input_specs)
```
kimishpatel: unused in this if block

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
...
+
+    # Create input spec for the mutable buffer
+    input_specs = exp_program.graph_signature.input_specs
+    mutable_input_spec = InputSpec(
+        InputKind.BUFFER, TensorArgument(name), target, persistent_buffer
+    )
+    input_specs.insert(node_index, mutable_input_spec)
+
+    # Create output spec for the mutable buffer
+    # We don't have to explicitly update the output.args[0] since it will be updated by the user based on the graph.
```
kimishpatel: I dont follow this. If you have to return input placeholder node, that you created, because it is a mutation, doesnt the output node's args have to be updated?

backends/transforms/test/test_create_mutable_buffer.py
```diff
@@ -0,0 +1,485 @@
...
+                            {},
+                        )
+                        # Update the output of the graph. This is a bit hacky, but it is done through `to_edge`.
+                        # We have to mimic it here for testing against eager.
+                        output_node = graph_module.graph.output_node()
+                        outputs = list(output_node.args[0])
+                        if self.return_buffer:
+                            outputs.insert(0, buf_node)
+                        else:  # return the index_put node
+                            outputs.insert(0, index_put_node)
```
kimishpatel: Yeah so why is this not done with create_mutable_buffer_node. Feels like this is an atomic thing that should be done together

backends/transforms/test/test_create_mutable_buffer.py
```diff
@@ -0,0 +1,485 @@
...
+
+        # Explicitly passing inplace_pass to make sure it works with our manually inserted buffer and index_put_ node(s).
+        executorch_program_manager = edge_program.to_executorch(
+            ExecutorchBackendConfig(
+                emit_mutable_buffer_names=True, run_reinplace_pass=True
+            )
+        )
+        return executorch_program_manager
+
+    def _test_eager(self, model, example_inputs, num_lifted_args=1):
```
kimishpatel, line 224: dont follow what this is doing for eager?

backends/transforms/test/test_create_mutable_buffer.py
```diff
@@ -0,0 +1,485 @@
...
+        with torch.no_grad():
+            graph_model = CustomModuleGraph().eval()
+            et_1 = self._test_edge_pass(graph_model, example_inputs)
+
+            src_model = CustomModuleSrc().eval()
+            et_2 = self._test_eager(src_model, example_inputs)
+
+        self.compare(et_1, et_2, example_inputs)
+
+    def test_basic_with_param(self):
```
kimishpatel, line 375: are you testing when you dont have any inputs in the graph? there is a path for that in your pass.

kimishpatel (approved):
looks pretty good. left some comments

backends/transforms/test/test_create_mutable_buffer.py
```diff
@@ -0,0 +1,485 @@
...
+                                node,
+                            ),
+                            {},
+                        )
+                        # Update the output of the graph. This is a bit hacky, but it is done through `to_edge`.
+                        # We have to mimic it here for testing against eager.
+                        output_node = graph_module.graph.output_node()
+                        outputs = list(output_node.args[0])
+                        if self.return_buffer:
+                            outputs.insert(0, buf_node)
```
JacobSzwejbka: Why does this not happen in the mutable buffer pass? Surely if you add a buffer you say is mutated it needs to be returned

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
 
     # Remove node from graph
     node.graph.erase_node(node)
+
+
+def create_mutable_buffer(
+    exp_program: ExportedProgram,
+    graph: torch.fx.Graph,
```
JacobSzwejbka: theres a 1:1 mapping of EP to root graph why bother passing as separate arg?

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
...
+    if name.startswith("b_"):
+        target = name[2:]
+    else:
+        target = name
+
+    # Check if target already exists
+    if target in exp_program.state_dict:
+        raise RuntimeError(f"Buffer target '{target}' already exists in state_dict")
+
+    persistent_buffer = True
```
JacobSzwejbka, line 306: hmmm my gut feeling is that this should not be in the state dict. Any idea where qparams go? If they go in the state dict I guess its fine

backends/transforms/utils.py
```diff
@@ -187,3 +230,128 @@ def delete_constant_placeholder(exp_program: ExportedProgram, node: torch.fx.Nod
...
+        )
+
+    for node, input_spec in zip(placeholders, exp_program.graph_signature.input_specs):
+        if node.name != input_spec.arg.name:
+            raise RuntimeError(
+                f"Input node {node.name} does not match input spec {input_spec.arg.name}"
+            )
+
+    # Find indices of user inputs and constant tensors
+    user_input_indices_or_consts = [
```
JacobSzwejbka: I feel like a loop would be a cleaner way of writing this through line 324. 

Since you have validated the Spec.kind and the nodes earlier you could just loop over the input_specs and index into graph.nodes

backends/transforms/utils.py
JacobSzwejbka: iiuc your end idea @digantdesai  is to add the buffer and then consume and mutate it with another pass that adds a custom op, so you could just do the work of setting up the outputs when you actually mutate the buffer. 

I agree with Kimish though that I think its a lot easier to compose this pass if it leaves the graph in a valid state rather then implicitly assuming the following pass will fix it up.

JacobSzwejbka:
@angelayi @avikchaudhuri @tugsbayasgalan 

I think in general it would be a good value add from compiler if these sorts of primitive graph operations lived in core. 

Things like node.is_mutable_buffer(), ep.add_buffer() ep.add_mutable_buffer() etc

JacobSzwejbka (changes requested):
.

backends/transforms/utils.py
digantdesai: it is used outside the block

backends/transforms/utils.py
digantdesai: I was thinking of this as a util (to be called from a pass) which returns a node and the pass calling this does actual graph manipulations. 
IIUC, you are saying after each line in the pass the graph should be in a valid state, but I also see the "logically belongs together" argument let me rework..

backends/transforms/test/test_create_mutable_buffer.py
digantdesai: Not sure I follow. `x` is the normal input, and I am adding a param (beyond the `test_basic` module) to test if the signature update is correct.

backends/transforms/test/test_create_mutable_buffer.py
digantdesai: `eager` might be a bad name, but the idea is to compare the ExecuTorch graph from two different origins for a mutable buffer (1) written in torch.nn.Module by the end user as in `eager` mode, and (2) inserted via this util + pass, and both ET graphs should be indistinguishable in every way.

backends/transforms/utils.py
JacobSzwejbka: >IIUC, you are saying after each line in the pass the graph should be in a valid state

I think public utils should probably leave it in a valid state. The graph can obviously have invalid state in the middle of a pass, but if I was joe schmo who didnt really understand the nuances of what this util did and just had the high level goal of "add a buffer to it" I would expect the graph to be sound after that operation. 

Like to me its hard to reason about even where in the outputs arg a user is expected to add the actual node return since it needs to match exactly with the spot you have already put it in the graph signature. The disconnect between prepopulating the signature but not populating the output node is what makes it weird.

backends/transforms/utils.py
digantdesai: Fixed

backends/transforms/utils.py
digantdesai: Fixed in the latest push. Now the util will create placeholder node, update signature and outputs it.

backends/transforms/test/test_create_mutable_buffer.py
digantdesai: moved the output update logic in the util fn.

backends/transforms/utils.py
digantdesai: it is in the exported_program.state_dict for the "eager" case. Added a state_dict comparison in the test.

backends/transforms/utils.py
digantdesai: refactored but still not as clean as I like.

backends/transforms/test/test_create_mutable_buffer.py
digantdesai: updated in the follow up.

backends/transforms/utils.py
digantdesai: fixed in the follow up.

mergennachin:
@digantdesai see my previous comment https://github.com/pytorch/executorch/pull/13813#pullrequestreview-3180893134

digantdesai:
> @digantdesai see my previous comment [#13813 (review)](https://github.com/pytorch/executorch/pull/13813#pullrequestreview-3180893134)

separate PR?

backends/transforms/test/test_create_mutable_buffer.py
```diff
@@ -0,0 +1,503 @@
...
+                                node.name, index_put_node.name
+                            )
+                            self.exported_program._graph_signature = graph_signature
+
+                    modified = True
+
+            if modified:
+                graph_module = super().call(graph_module).graph_module
+                graph_module.graph.lint()
+                graph_module.graph.eliminate_dead_code()
```
JacobSzwejbka, line 185: this fn is technically unsafe to call if the graph is not functional.

backends/transforms/utils.py
```diff
@@ -22,10 +22,43 @@
...
+    """
+    nodes = list(graph.nodes)
+    if not nodes:
+        raise RuntimeError(
+            "Cannot create fake tensor: graph has no nodes to extract fake_mode from"
+        )
+
+    example_node = nodes[0]
+    if isinstance(
+        example_node.meta["val"], (tuple, torch.fx.immutable_collections.immutable_list)
```
JacobSzwejbka, line 53: seems like a bit of a strange check. Is this robust?

backends/transforms/utils.py
JacobSzwejbka: >it is in the exported_program.state_dict for the "eager" case. Added a state_dict comparison in the test.

Only when persistent is set to true right? Which is user configureable.

JacobSzwejbka (approved):

backends/transforms/utils.py
digantdesai: Well, I am not sure.

For the buffers inserted by this util fn it is always with `persistent_buffer=True`. FWIW, the reference eager case I am not doing anything special besides defining them in `__init__()`.

backends/transforms/utils.py
digantdesai: Hmm just refactored in this PR.. but what else it can be when number of tensors > 1?

digantdesai (commented):
Thanks @JacobSzwejbka

PR #13816, opened 2025-08-29
[ET-VK] Quantized Int8 Linear

SS-JIA:
* #14020
* #14019
* #13879
* #13818
* #13817
* __->__ #13816

Title says it all!

This PR adds implementations for int8 linear layers. Convolution is implemented in a later step, computing convolution as matrix multiplication via the im2col procedure.

For both linear and convolution, two versions are implemented:

1. `q8ta_q8csw` variant which quantized the input tensor and then performs integer accumulation via the int8 dot product extension
2. `q8csw` variant which dequantized the weight tensor in-shader and performs floating point accumulation.

The second one is needed to provide an alternative path for executing quantized models if the target GPU does not support int8 dot product extension.

These new ops are tested via the custom op testing + benchmarking framework introduced in the previous diff.

Differential Revision: [D81323424](https://our.internmc.facebook.com/intern/diff/D81323424/)


backends/vulkan/test/custom_ops/quantized_linear.cpp
```diff
@@ -0,0 +1,352 @@
...
+  // Calculate number of output elements
+  int64_t num_output_elements = batch_size * out_features;
+
+  auto& ref_data = output_spec.get_ref_float_data();
+  ref_data.resize(num_output_elements);
+
+  // Perform quantized linear transformation (matrix multiplication)
+  for (int64_t b = 0; b < batch_size; ++b) {
+    for (int64_t out_f = 0; out_f < out_features; ++out_f) {
+      float sum = 0.0f;
```
digantdesai: No int32 accumulation in the reference? Why not use q->dq->linear->q->dq canonical pytorch pattern then?

backends/vulkan/runtime/graph/ops/impl/QuantizedLinear.cpp
```diff
@@ -0,0 +1,548 @@
...
+ *
+ * The reason for this split is twofold:
+ * - Some devices may not support accelerated int8 dot product. In that case,
+ *   there is no benefit to quantizing the input tensor. In that case
+ * linear_q8csw is required.
+ * - For LLMs, which switch between GEMM and GEMV input conditions when going
+ *   from prefill to decode. GEMM is typically a compute bound operation, which
+ *   will benefit from accelerated int8 accumulation. On the other hand, GEMV
+ *   is usually memory bound, which means it may actually suffer from the extra
+ *   cost of having to quantize and pack the input tensor. Therefore,
```
digantdesai: Did we benchmark this?

backends/vulkan/runtime/graph/ops/glsl/linear_fp_bias_load.glslh
```diff
@@ -0,0 +1,30 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#ifndef LINEAR_FP_BIAS_LOAD_GLSLH
+#define LINEAR_FP_BIAS_LOAD_GLSLH
+
+#include "linear_fp_per_out_channel_params.glslh"
+
+VEC4_T load_bias_x4(const int n4) {
+  return t_bias[n4];
```
digantdesai, line 15: is `t_bias` global?

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile.glslh
```diff
@@ -0,0 +1,43 @@
...
+ * Settings:
+ * - TILE_M: number of rows in the tile
+ * - TILE_K4: number of (groups of 4) columns in the tile
+ */
+
+struct FPInputTile {
+  VEC4_T data[TILE_M][TILE_K4];
+};
+
+#ifdef DEBUG_MODE
```
digantdesai, line 27: not defined otherwise?

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile_load.glslh
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+#ifndef LINEAR_FP_INPUT_TILE_LOAD_GLSLH
+#define LINEAR_FP_INPUT_TILE_LOAD_GLSLH
+
+#extension GL_EXT_control_flow_attributes : require
+
+#include "linear_fp_input_tile.glslh"
+
+#ifdef INPUT_BUFFER
```
digantdesai, line 27: Nit: why is this not inside `load_input_x4`?

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile_load.glslh
```diff
@@ -0,0 +1,91 @@
...
+#endif // INPUT_BUFFER
+
+// To be used if (M - m_start >= TILE_M) || (K4 - k4_start >= TILE_K4)
+void load_input_tile_no_checks(
+    out FPInputTile in_tile,
+    const int k4_start,
+    const int m_start,
+    const int K4,
+    const int M) {
+#if TILE_K4 == 1
```
digantdesai, line 48: why do we need this specialization? Does compiler not do this for you?

backends/vulkan/runtime/graph/ops/glsl/linear_q8csw_tiled.glsl
```diff
@@ -0,0 +1,114 @@
...
+void main() {
+  // Each thread writes out a 4 wide x 4 high tile of output values
+  const int out_tile_x = int(gl_GlobalInvocationID.x);
+  const int out_tile_y = int(gl_GlobalInvocationID.y);
+
+  const int n = out_tile_x * TILE_N;
+  const int m = out_tile_y * TILE_M;
+
+  const int n4 = div_4(n);
+  const int m4 = div_4(m);
```
digantdesai, line 64: `m4` seems redundant, remove?

backends/vulkan/runtime/graph/ops/glsl/linear_q8csw_tiled.glsl
```diff
@@ -0,0 +1,114 @@
...
+#include "linear_fp_output_tile_fp_compute.glslh"
+#include "linear_fp_output_tile_fp_int8_compute.glslh"
+#include "linear_fp_output_tile_store.glslh"
+#include "linear_fp_scales_load.glslh"
+#include "linear_fp_bias_load.glslh"
+
+void main() {
+  // Each thread writes out a 4 wide x 4 high tile of output values
+  const int out_tile_x = int(gl_GlobalInvocationID.x);
+  const int out_tile_y = int(gl_GlobalInvocationID.y);
```
digantdesai, line 58: I am still trying to process this and figuring out what the right way to do this, but just as a hint, in absence of shared memory based memory reuse, at a high level, it feels like we want each local group to generate a tile of output tiles to ensure memory reuse on both weight tiles and input tiles. 

IIUC this produces a row or a column of output tiles, but not a tile of tiles in 2D.

manuelcandales (approved):

digantdesai (commented):
still looking...

backends/vulkan/test/custom_ops/quantized_linear.cpp
SS-JIA: I updated the test file to use int accumulation for the 8bit activation quantized path, as you suggested!

backends/vulkan/runtime/graph/ops/impl/QuantizedLinear.cpp
SS-JIA: Haven't had the time to validate - but decided to just go this path for now because weight only quant is needed anyway to support devices that don't have the int8 dot product extension + the fact that ML drift also utilizes this approach. I plan to do some more experimentation to confirm post GA

backends/vulkan/runtime/graph/ops/glsl/linear_fp_bias_load.glslh
SS-JIA: yes, many of these header files assume that certain variables are defined in the shader layout.

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile.glslh
SS-JIA: The usage of this is to be able to
```
#define DEBUG_MODE
```

in order to access debugging functions in the shader template.

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile_load.glslh
SS-JIA: that's a great point. I will include this + other ifdef cleanups in a follow up diff.

backends/vulkan/runtime/graph/ops/glsl/linear_fp_input_tile_load.glslh
SS-JIA: Yeah, I plan to simplify these ifdefs in my most recent diff.

backends/vulkan/runtime/graph/ops/glsl/linear_q8csw_tiled.glsl
SS-JIA: Currently each thread produces a 4x4 output block. The header files have been configured to make the tile size configurable in groups of 4.

PR #13819, opened 2025-08-29
Arm backend: Add NGP unit test scripts

mansnils:
Prepare scripts for CI.


backends/arm/scripts/build_executor_runner_vkml.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "Building with extra flags: ${build_with_etdump_flags} ${extra_build_flags}"
+cmake \
+    -DCMAKE_BUILD_TYPE=${build_type}            \
+    -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
+    -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
+    -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
+    -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
+    -DEXECUTORCH_BUILD_XNNPACK=OFF \
+    -DEXECUTORCH_BUILD_VULKAN=ON \
```
digantdesai, line 73: this will build and link ET Vulkan backend, ok for now but why do we need this?

digantdesai (approved):
Couple of high level questions,
(1) how do you envision evolution of these setup, build, run script with ethos? 
(2) why not add the scripts introduced in the PR on the CI? I might be missing something in terms of runtime dependency but I am asking as a way to test these code you are introducing. If not run, let's at least build?

backends/arm/scripts/build_executor_runner_vkml.sh
mansnils: We use the platform and runtime environment provided by the Vulkan delegate

mansnils:
> Couple of high level questions, (1) how do you envision evolution of these setup, build, run script with ethos? (2) why not add the scripts introduced in the PR on the CI? I might be missing something in terms of runtime dependency but I am asking as a way to test these code you are introducing. If not run, let's at least build?

Thanks for your comments.
1) We plan to refactor this, specially the setup part
2) The patch has potential for merge conflicts, in fact one already happened. I think the fastest way is to merge it then enable CI as a second patch

mansnils:
Ethos-U failures unrelated to this PR

Issue #13773, opened 2025-08-28
Some CI jobs are failing on Apple


shoumikhin:
### 🐛 Describe the bug

See https://hud.pytorch.org/hud/pytorch/executorch/main
The failures started yesterday afternoon
<img width="899" height="403" alt="Image" src="https://github.com/user-attachments/assets/8d1a0cab-2bab-41b5-ad59-3950f15209be" />

By looking at the logs I can say that those previous logs that succeeded mentioned `macos-14-arm64` runners, and the failed ones are using `macos-15-arm64`.

There's a common error fetching a prebuilt torch wheel:

<img width="421" height="103" alt="Image" src="https://github.com/user-attachments/assets/29a8efb5-3bc9-421e-bb20-c57d636ebce0" />

### Versions

Current CI

atalman:
Looks like issue is related to torchaudio:
```
MD -MT src/libtorchaudio/sox/CMakeFiles/libtorchaudio_sox.dir/types.cpp.o -MF src/libtorchaudio/sox/CMakeFiles/libtorchaudio_sox.dir/types.cpp.o.d -o src/libtorchaudio/sox/CMakeFiles/libtorchaudio_sox.dir/types.cpp.o -c /private/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/pip-req-build-x_jy67c1/src/libtorchaudio/sox/types.cpp
      ninja: build stopped: subcommand failed.
      
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 35, in <module>
        File "/private/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/pip-req-build-x_jy67c1/setup.py", line 144, in <module>
          _main()
        File "/private/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/pip-req-build-x_jy67c1/setup.py", line 99, in _main
          setup(
        File "/Users/runner/work/_temp/conda_environment_17279168142/lib/python3.11/site-packages/setuptools/__init__.py", line 117, in setup
          return distutils.core.setup(**attrs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/runner/work/_temp/conda_environment_17279168142/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 186, in setup
          return run_commands(dist)
                 ^^^^^^^^^^^^^^^^^^
[...]

PR #13821, opened 2025-08-29
Use macos-14-xlarge instead of macos-latest-xlarge which has switched…

shoumikhin:
https://github.com/pytorch/executorch/issues/13773


jackzhxng (commented):
Hmm seems like some different errors now, cc @huydhn

shoumikhin:
@jackzhxng GitHub looks unreliable atm, I'll keep retrying

jackzhxng (approved):

PR #13823, opened 2025-08-29
Support Half/BFloat16 in native_group_norm

manuelcandales:
Differential Revision: D81351792


digantdesai (approved):
Stamping to unblock!

PR #13825, opened 2025-08-29
Jni layer use multimodal runner

kirklandsign:


jackzhxng (commented):
Nice, how did you test?

kirklandsign:
> Nice, how did you test?

Honestly not sure how to test multimodal. I can just check for llama regression for now.

Maybe llava? Do you have idea?

jackzhxng:
Yeah, I think just checking Llama for regression would be good for now

kirklandsign:
> Yeah, I think just checking Llama for regression would be good for now

Validated

jackzhxng (approved):

mergennachin:
@kirklandsign Is there a test for jni layer?

kirklandsign:
> @kirklandsign Is there a test for jni layer?

No. We use E2E CI.

However, I have trouble with validating this. Multimodal runner requires method `get_max_seq_len` but when I tried to export llava, seems this method is missing.

kirklandsign:
I have some difficulty validating this. Getting error 
```
Attempted to resize a static tensor. Expected shape (1), but received (4).
Error resizing tensor at input 0
```
https://github.com/pytorch/executorch/blob/41c299f29f4cffa6303a00c9211651cc0d8a3cc8/extension/llm/runner/multimodal_prefiller.cpp#L110-L111

jackzhxng:
@kirklandsign is this for text prefill or image prefill? If it's text prefill, then you should check if the encoder embeddings has a batch dimension

kirklandsign:
> @kirklandsign is this for text prefill or image prefill? If it's text prefill, then you should check if the encoder embeddings has a batch dimension

text.

kirklandsign:
Now the part I'm not sure is https://github.com/pytorch/executorch/blob/41c299f29f4cffa6303a00c9211651cc0d8a3cc8/extension/llm/runner/multimodal_prefiller.cpp#L108

which is different from llava runner
https://github.com/pytorch/executorch/blob/19d22ab8deb7dfecb03e08342656052f03ea1224/examples/models/llava/runner/llava_text_decoder_runner.h#L34

Otherwise we can merge this PR

PR #13833, opened 2025-08-30
Fix reduction over dim list for empty input

manuelcandales:
Summary:
Allow empty input in `MapReduceOverDimListPlan`
Avoid integer division by zero in `parallel_for_each_reduce_over_dim_list_output_index`
Add empty input tests for ops: any, mean, sum & var

Differential Revision: D81383049


swolchok (changes requested):

swolchok (approved):

PR #13836, opened 2025-08-31
[Windows] Add model run CI

GregoryComer:
Add Windows CI jobs to test models. From looking at the existing CI, it looks like our model test jobs are pretty minimal and don't verify much other than it runs without crashing. I'd like to re-visit this in the near future and integrate with the new test harness, but for the purpose this PR, I'm just matching the Windows jobs to what the Linux jobs do.


GregoryComer:
* #13837
* __->__ #13836

.ci/scripts/test_model.ps1
```diff
@@ -0,0 +1,81 @@
...
+    python -m examples.xnnpack.aot_compiler --model_name="${MODEL_NAME}" --delegate | Write-Host
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Model export failed. Exit code: $LASTEXITCODE."
+        exit $LASTEXITCODE
+    }
+
+    "$($modelName)_xnnpack_fp32.pte"
+}
+
+.ci/scripts/setup-windows.ps1
```
mergennachin: can you do it outside of this script and call directly in trunk.yml

that way, we have a clear separation of roles

.ci/scripts/test_model.ps1
```diff
@@ -0,0 +1,81 @@
...
+}
+$executorBinaryPath = Join-Path -Path $buildDir -ChildPath "Release\executor_runner.exe"
+Pop-Location
+
+# Export the model
+switch ($backend) {
+    "portable" {
+        $model_path = ExportModel-Portable -model_name $modelName -strict $strict
+    }
+    "xnnpack" {
```
mergennachin: also support `xnnpack-quantization-delegation` which calls examples.xnnpack.aot_compiler with --quantize flag

mergennachin (approved):

.ci/scripts/test_model.ps1
GregoryComer: Done.

.ci/scripts/test_model.ps1
GregoryComer: Done.

PR #13839, opened 2025-09-01
Arm backend: Change W2L "input type" for Ethos-U55 test

martinlsm:
The W2L model includes a conv1d layer with an unsupported stride length (160) when the input_type argument to the model's constructor is set to the default value ("waveform"). Change the input_type parameter to "power_spectrum" as a workaround to not include this unsupported layer.

Still the test case will not pass because there is lack of memory (SRAM). Change the pytest.mark.skip decorator to the U55 test to reflect the new expected reason for failure.


zingo:
Test fails are unrelated

zingo (approved):

PR #13843, opened 2025-09-01
Arm backend: Fix qdq_fusion flag

Sebastian-Larsson:
The flag was incorrectly handled leading to broken CI. Instead, make the flag behave as a boolean. Leaving it out behaves the same way as setting it to False previously should have.


zingo (approved):

zingo:
Fails unrelated

PR #13856, opened 2025-09-02
let backends set info on the lowered_backend.meta from preprocess

JacobSzwejbka:
Summary: We want some delegates to be able to test experimental features that interact with the core runtime. Typically when doing this there is an expectation of tight coupling in the full lowering process so we arent super concerned about UX. The backend api is pretty hard to change because the code is mostly owned by 3rd parties, so we want to be really sure when we change it that its correct. By exposing this module.meta hookup we can test things in a controled maner without changing the blessed api before we are ready.

Differential Revision: D81466391

### DIFF
--- a/exir/backend/backend_details.py
+++ b/exir/backend/backend_details.py
@@ -7,7 +7,7 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from executorch.exir._serialize._named_data_store import NamedDataStoreOutput
 
@@ -31,6 +31,11 @@ class PreprocessResult:
     # Named Data store contains all the named data that is stored in the PTE file,
     # but retrieveable by delegates via the NamedDataMap at runtime.
     data_store_output: Optional[NamedDataStoreOutput] = None
+
+    # Optional delegate-specific information that will be added to the
+    # lowered_module.meta field in the graph, but not directly serialized
+    # into the PTE file.
+    _delegate_info_meta: Optional[Any] = None
 
 
 """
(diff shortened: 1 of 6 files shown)


hsharma35 (approved):

exir/backend/backend_details.py
```diff
@@ -32,6 +32,9 @@ class PreprocessResult:
     # but retrieveable by delegates via the NamedDataMap at runtime.
     data_store_output: Optional[NamedDataStoreOutput] = None
 
+    # Optional delegate-specific information that will be added to the lowered_module's metadata.
+    delegate_info: Optional[Any] = None
```
digantdesai: Seems a little unstructured, but OK I guess. How about `delegate_info_meta` just to discourage people to put flatbuffers in here :p

cccclai: How is it different than having it as part of the delegated blob? From the logic it looks like we just have this information as part of the preprocess result (the meaningful part if just the blob) and put it as part of the lowered module (which again still just the blob)

hsharma35: Doing anything with delegated blob means we need to disassemble - modify - assemble the blob again. This is not expected to be trivial for backends.

JacobSzwejbka: > Seems a little unstructured, but OK I guess.

Intentionally so. Even ignoring the current motivation around shared memory planning, I think it is generically useful to let us have a mechanism to experiment with tightly coupled passes before promoting things to the backend api which we expect to be composable.

JacobSzwejbka: >How is it different than having it as part of the delegated blob?  

Some delegate blobs are not trivial to unpack is the main motivation. AOTI for instance would just be compiled code iiuc.

mergennachin: @JacobSzwejbka 

This is a meta information that won't be serialized as part of the PTE file, right? 

If that's the case, can you document it here? 

Also I'd like to rename delegate_info to be something like `_delegate_info_meta` or `_delegate_info_internal` or just `_meta`?

JacobSzwejbka: yes it just gets added to the fxGraph. Updated the name and comment

mergennachin (approved):

Issue #7345, opened 2024-12-17
insert_write_back_for_buffers_pass should inject copy_ nodes at the earliest possible spot.


JacobSzwejbka:
### 🐛 Describe the bug

Currently the pass injects all the 'copy_' nodes at the end of the graph. It should instead insert them immediately following the last usage of the thing being copied into the original mutable buffer. By waiting till the end we are arbitrarily extending the lifetime of those tensors potentially wasting a non trivial amount of space in the memory plan. 

### Versions

.

JacobSzwejbka:
> Looks interesting as a first issue, can I give this one a try?

Yeah feel free to take a crack at it and I'll review any PR you put up. If you need some more guidance let me know. 

the pass is called in _program.py.

metascroy:
@prmaynard are you still working on this?  If not, we'd like to return it to the pool.

RihaanBH-1810:
@JacobSzwejbka  @metascroy  can I please take this up as my first issue, since there hasn't been much activity by the previous assignee.

prmaynard:
@RihaanBH-1810 please feel welcome to take a crack at this! Here is where I was able to get (forgive me it has been a while since I looked at this, life got busy)

```
def _insert_copy(
    gm: torch.fx.GraphModule,
    mutated_outputs: List[Optional[str]],
    input_name_to_node: Dict[str, torch.fx.Node],
):

    if all(name is None for name in mutated_outputs): # if all mutated outputs are none, do not execute
        return []    

    output_node = None
    for node in gm.graph.nodes:
        if node.op == "output":
            output_node = node
            break
    assert output_node is not None

    outputs = pytree.tree_flatten(output_node.args)[0]
    assert len(outputs) == len(mutated_outputs)

    user_returns = []
    buffer_copies = []

    for return_node, mutated_node_name in zip(outputs, mutated_outputs):

        if mutated_node_name is None:
            user_returns.append(return_node)
            continue

        # Mutable buffer grab the node
        if mutated_node_name in input_name_to_node:
            mutated_node = input_name_to_node[mutated_node_name]
        else:
            raise RuntimeError(
                f"Could not find {mutated_node_name} in either buffer or input nodes"
            )
    
        # needed to rebuild node index pairs after graph modification
        nodes = list(gm.graph.nodes)
        node_index = {node: i for i, node in enumerate(nodes)}

        # last reader
        last_read_idx = node_index[mutated_node]

        for user in mutated_node.users:
[...]

PR #13857, opened 2025-09-02
[Draft PR] Bugfix improve the logic for inserting copy nodes after the last use of tensor in buffer pass.

RihaanBH-1810:
This is a draft PR to fix issue #7345


exir/passes/insert_write_back_for_buffers_pass.py
```diff
@@ -26,21 +26,27 @@ def _insert_copy(
...
+
     for return_node, mutated_node_name in zip(outputs, mutated_outputs):
-        # User output, leave alone
+
         if mutated_node_name is None:
-            user_output_nodes.append(return_node)
+            user_returns.append(return_node)
             continue
 
         # Mutable buffer grab the node
```
JacobSzwejbka, line 52: This is a bit of a stupid question since I think I wrote this code, but looking back on it. Are mutated node and return node not the exact same node?

exir/passes/insert_write_back_for_buffers_pass.py
```diff
@@ -26,21 +26,27 @@ def _insert_copy(
...
-    output_node = gm.graph.output_node()
+
+    if all(name is None for name in mutated_outputs): # if all mutated outputs are none, do not execute
+        return []    
+
+    output_node = None
+    for node in gm.graph.nodes:
+        if node.op == "output":
+            output_node = node
+            break
```
JacobSzwejbka, line 37: nit: go back to gm.graph.output_node()

PR #13859, opened 2025-09-02
Arm backend: Move is_consumer_node_depthwise_conv2d to the pass using it

Sebastian-Larsson:
The function is_consumer_node_depthwise_conv2d is only used by annotate_channels_last_dim_order_pass and can therefore be moved closer to where it is used.


Sebastian-Larsson:
Unrelated failures

zingo (approved):

PR #13865, opened 2025-09-02
Arm backend: Refactor setup.sh to break out fvp installation

perheld:
Broken out the fvp installation into a separate script handling only that part. No functional change.

Change-Id: I8accf263ae8617dc5679a6b4e21e9331dffdda34


zingo (approved):

zingo:
Test fails unrelated

PR #13866, opened 2025-09-02
                  Arm backend: Skip int operation tests as for FP profile

per:
### Summary
Switching to TOSA 1.0 the FP profile doesn't support int (as the 0.80 MI profile did). The test was not updated to reflect that, so skip them for op argument validation patches to land.

### Test plan
Tested through unit tests.


oscarandersson8218 (approved):
Looks good to me.

PR #13875, opened 2025-09-02
Del redundant-static-def in executorch/schema/extended_header.cpp +2

r-barnes:
Summary:
LLVM has a warning `-Wdeprecated-redundant-constexpr-static-def` which raises the warning:

> warning: out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated

Since we are now on C++20, we can remove the out-of-line definition of constexpr static data members. This diff does so.

 - If you approve of this diff, please use the "Accept & Ship" button :-)

Differential Revision: D81359733

### DIFF
--- a/schema/extended_header.cpp
+++ b/schema/extended_header.cpp
@@ -91,8 +91,5 @@ uint64_t GetUInt64LE(const uint8_t* data) {
   };
 }
 
-// Define storage for the static.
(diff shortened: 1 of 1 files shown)


JacobSzwejbka (changes requested):

JacobSzwejbka (approved):
Misunderstood the comment on c++20

PR #13876, opened 2025-09-02
Update Neutron Converter

JakeStevens:
Summary:
allow-large-files

Newest update released, bumping version.

Note, for some reason NXP has the full qualifier like

neutron_converter_SDK_25_06.neutron_converter

so we have to update the test.

Differential Revision: D80261684


shoumikhin (approved):

digantdesai:
nxp jobs are on trunk, we should always run trunk on nxp PRs.

PR #13885, opened 2025-09-02
Fix prefill_inference

rohansjoshi:
Summary: Fixes bugs in prefill_inference function

Differential Revision: D81532886


cccclai (approved):
Thanks for the fix! cc: @haowhsu-quic

PR #13886, opened 2025-09-02
Add program-data separation to pybindings

lucylq:
Summary: Add support for optional data path for pybindings.

Differential Revision: D76353209

### DIFF
--- a/extension/pybindings/test/test_pybindings.py
+++ b/extension/pybindings/test/test_pybindings.py
@@ -22,6 +22,7 @@ from executorch.extension.pybindings.test.make_test import (
     ModuleAddWithAttributes,
     ModuleChannelsLast,
     ModuleChannelsLastInDefaultOut,
+    ModuleLinear,
     ModuleMulti,
 )
 from torch.export import export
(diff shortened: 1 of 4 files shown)


extension/pybindings/test/test_pybindings.py
```diff
@@ -600,3 +601,32 @@ def test_method_method_meta(self) -> None:
...
+            pte_file = os.path.join(tmpdir, "linear.pte")
+            with open(pte_file, "wb") as f:
+                f.write(exec_program.buffer)
+            
+            ptd_file = os.path.join(tmpdir, "linear.ptd")
+            with open(ptd_file, "wb") as ptd:
+                tensor_data = bytes(exec_program._tensor_data.pop("_default_external_constant"))
+                ptd.write(tensor_data)
+            
+            executorch_program = self.runtime._load_for_executorch(pte_file, ptd_file)
```
JacobSzwejbka, line 654: probably need to update the pyi

lucylq: Thanks, updated!

JacobSzwejbka (approved):

extension/pybindings/test/test_pybindings.py
```diff
@@ -600,3 +601,35 @@ def test_method_method_meta(self) -> None:
...
+                f.write(exec_program.buffer)
+
+            ptd_file = os.path.join(tmpdir, "linear.ptd")
+            with open(ptd_file, "wb") as ptd:
+                tensor_data = bytes(
+                    exec_program._tensor_data.pop("_default_external_constant")
+                )
+                ptd.write(tensor_data)
+
+            executorch_program = self.runtime._load_for_executorch(pte_file, ptd_file)
```
mergennachin, line 654: It's fine to update the legacy API (methods that start with _) but we have modern pybindings APIs that we want users to use.

https://colab.research.google.com/drive/1qpxrXC3YdJQzly3mRg-4ayYiOjC6rue3?usp=sharing#scrollTo=Kp-MxVVHkirE

Don't we need to update runtime.load_program(pte, ptd) too?

mergennachin (commented):
https://github.com/pytorch/executorch/blob/cac1a71a7f4224f074150abe25ab91340fb037f4/runtime/__init__.py#L199-L204

extension/pybindings/test/test_pybindings.py
lucylq: Thanks Mergen, sorry for the late response. I think runtime.load_program is only intended to take a PTE file --> it understands and deserializes the PTE file format (not PTD). The C++ Program object doesn't hold a PTD file.

We can potentially support this in load_method, which takes a NamedDataMap (from PTD) as an optional argument. We'd need to add pybindings on extension/flat_tensor to load the PTD file for this.

Having it in Module may be OK at the moment, let me know what you think (I can add pybindings for extension/flat_tensor in a separate PR).

PR #13887, opened 2025-09-02
XNNPACK: Kleidi QP8 and SME2

digantdesai:
### Summary
* enable qp8 dqlinear paths at runtime
* enable xnnpack arm sme2 isa build flag (XNNPACK_ENABLE_ARM_SME is already ON)

### Benchmark and SME2 data

* Not so rigorous benchmarking on an M4 using 
`executor_runner --model_path vit_xnnpack_q8.pte --num_executions 100` 
with and without SME2 flag shows, and w/ SME2 it is consistently faster.
* Verified kleidi SME2 kernel is used through Instrumentation on M4
* LLM packing perf is still needs to be validated with QP8.


mergennachin (approved):
Godspeed!

PR #13892, opened 2025-09-03
Qualcomm AI Engine Direct - Strengthen Unit Test Robustness

shewu-quic:
Summary:
  - Fixed the seed for E2E model scripts in unit test
  - Resolved the bug during dump optrace
  - The per-channel quant config of the bias for conv op should be derived by activation and weight.
  - Resolve the issue caused by [the PR](https://github.com/pytorch/executorch/pull/13606) that prevented the spec from being correctly updated to the quantized type.

### DIFF
--- a/backends/qualcomm/_passes/build_quant_io.py
+++ b/backends/qualcomm/_passes/build_quant_io.py
@@ -5,9 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 import torch
 from executorch.backends.qualcomm.utils.constants import QCOM_QUANTIZED_IO
+from executorch.exir.delegate import executorch_call_delegate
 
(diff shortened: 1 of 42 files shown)


shewu-quic:
Hi @cccclai ,
This PR is to resolve some errors in our internal unit test with ExecuTorch mainline and qnn 2.37.
Could you please take a look?
Thanks

cccclai:
It seems like static_llama_qnn test is failing, can you take a look

shewu-quic:
> It seems like static_llama_qnn test is failing, can you take a look

Thanks for reminder.
It seems like network issue to download llvm. It's ok after I rebased.

cccclai (approved):

PR #13896, opened 2025-09-03
Fix torchao quantization for deployment targets under iOS18

metascroy:
_construct_constexpr_dequant_op should use axis = None (default value), rather than -1.  This will infer the axis from the shape of scales.

On iOS18, this was not an issue because the axis is not used.  For iOS16, the axis is used and the value -1 usually caused an issue during export.

### DIFF
--- a/backends/apple/coreml/compiler/torch_ops.py
+++ b/backends/apple/coreml/compiler/torch_ops.py
@@ -175,7 +175,6 @@ def dequantize_affine(context, node):
         int_data.astype(quantized_np_dtype),
         zero_point,
         scale,
-        axis=-1,
         name=node.name,
     )
     context.add(output, node.name)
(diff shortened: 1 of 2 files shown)


metascroy:
Fixes https://github.com/pytorch/executorch/issues/13122

metascroy:
FYI to @abhinaykukkadapu.  Per-channel 8-bit TorchAO quantization is now supported on iOS16+ with this PR. We can follow up with recipe update.

jackzhxng:
Passing reminder to use `release notes: quantization` if you think we should call out this bug fix on the release notes

YifanShenSZ (approved):

metascroy:
@cccclai @digantdesai @abhinaykukkadapu @jackzhxng I need a stamp from someone in PyTorch to merge.  The Apple engineer has already approved.

abhinaykukkadapu (approved):

PR #13903, opened 2025-09-03
Arm backend: Refactor setup.sh to break out fvp installation

perheld:
Broken out the fvp installation into a separate script handling only that part. No functional change.

Change-Id: I8accf263ae8617dc5679a6b4e21e9331dffdda34


perheld:
Ooops, pushed wrong branch. :face_holding_back_tears:

PR #13908, opened 2025-09-03
Arm backend: Enable capture of debug information during TOSA serialization

tom-arm:
* Add _serialize_operator() function to NodeVisitor
* Enable passing of DebugHook to NodeVisitor if enabled
* Add option dump_debug_info to ArmCompileSpecBuilder

Change-Id: I9dc8da50a7ef16053b778ab33a03a990035de5ad

### DIFF
--- a/backends/arm/operators/node_visitor.py
+++ b/backends/arm/operators/node_visitor.py
@@ -5,10 +5,11 @@
 
 # pyre-unsafe
 
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import torch
 
+from executorch.backends.arm.debug.schema import DebugHook
 from executorch.backends.arm.tosa_mapping import TosaArg
 from executorch.backends.arm.tosa_specification import TosaSpecification
 from torch.export import ExportedProgram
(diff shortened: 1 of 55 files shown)


zingo (approved):

backends/arm/operators/node_visitor.py
```diff
@@ -29,9 +30,38 @@ class NodeVisitor:
...
+        self,
+        exported_program: ExportedProgram,
+        tosa_spec: TosaSpecification,
+        debug_hook: Optional[DebugHook] = None,
+    ):
         self._exported_program = exported_program
         self.tosa_spec = tosa_spec
+        self.debug_hook = debug_hook
+
+    def _serialize_operator(
```
digantdesai, line 43: Can we extend `tosa_graph` and not overload node visitor with this method?
The `tosa_graph_torch` (let's say) constructor takes debug_hook or other args and prepares a "graph", through `addOperator` like methods, which can be observed independently just like an FX graph.

Just an idea, don't want to block this.

tom-arm: Thanks for the suggestion Digant. I like the idea, I think it would look a bit cleaner

I can look into this, there will be one/two more commits to finalise this feature. So I think there is scope to incorporate this.

PR #13913, opened 2025-09-03
Deprecate method retrieval in module

shoumikhin:


extension/module/module.h
```diff
@@ -215,7 +217,7 @@ class Module {
    * @returns A Result object containing either a pointer to the requested
    *          method or an error to indicate failure.
    */
-  ET_NODISCARD runtime::Result<Method*> method(const std::string& method_name);
+  ET_DEPRECATED ET_NODISCARD runtime::Result<Method*> method(const std::string& method_name);
```
mergennachin: Why now deprecating? This method is not part of 0.7 release, so technically we can delete right away https://www.internalfb.com/diff/D80595270?transaction_fbid=669692512088982

shoumikhin: I see people trying to use it, so want to prevent further propagation while I'm cleaning out the last use case under examples/.

lucylq (approved):
lgtm, maybe make an issue to remove this for 1.0 after removing the last usages?

see if @mergennachin has any other comments?

mergennachin (approved):

PR #13917, opened 2025-09-03
Stale workflow: Include issues and changed it to weekly instead of daily

nil-is-all:


GregoryComer:
@nil-is-all I don't think we should state issues at this point, pending more discussion as they may still be relevant. What are your thoughts? At a a minimum, I'd like to get team buy in on whatever solution we go with.

PR #13923, opened 2025-09-03
[Windows] Run native unit tests in CI

GregoryComer:
### Summary
Build and run C++ tests in the unit test CI job on Windows. Fix a few lingering issues on Windows that surfaced.

Changes:
 * Add CMake build + CTest execution the Windows unit test job.
 * Gate Core ML export recipe tests (and coremltools imports) to non-Windows platforms.
 * Add some explicit cases for size_t 0 literals to make MSVC overload resolution happy.
 * Remove  2> /dev/null from some CMake PTE export custom commands. I'm not entirely sure why this is there in the first place (keep the log a little cleaner?), but it doesn't work on Windows. Easiest to just remove it.
 * Update the temp file utility code (in extension/testing_util/temp_file.h) to use standard library file IO (as opposed to posix).
 * **(Critical)**: Replace long with int64_t in a few portable kernels that deal with indices. It should be int64_t, as this is the correct type for long tensors. On Windows, for example, long is actually 32-bit for historical reasons, and it causes memory corruption / crashes when trying to read/write ScalarType::Long tensors with long* pointers.
 * Update a bunch of custom commands to use platform-specific syntax or replace with a cross-platform approach.
 * Use non-strict export for test models. We can support strict export on Windows, but it take a bit of env setup. This is covered in CI in places, but non-strict is the default so it makes it a bit easier to build tests this way.

I've also excluded method_test and tensor_parser_test for now, due to the env issues with some of the PTE test models. The tests pass locally - but I need to debug the build-time PTE creation for these tests. I'm disabling to get the remaining 98% of the tests up first, and will fix these shortly. Tracking in https://github.com/pytorch/executorch/issues/13883.

### Test plan
CI - the tests are covered by the unittest jobs (debug, release, and editable).


kernels/portable/cpu/op_argmax.cpp
```diff
@@ -49,21 +49,21 @@ Tensor& argmax_out(
   static constexpr const char op_name[] = "argmax.out";
 
   ET_SWITCH_REALHBF16_TYPES(in.scalar_type(), ctx, op_name, CTYPE, [&] {
-    long* out_data = out.mutable_data_ptr<long>();
+    int64_t* out_data = out.mutable_data_ptr<int64_t>();
```
JacobSzwejbka, line 52: Nice. We should have a lint for this cc @manuelcandales

test/models/export_delegated_program.py
```diff
@@ -178,7 +178,7 @@ def forward(self, *args, **kwargs):
                 module=tagged_module,
                 gen_tag_fn=lambda x: module_class.__name__,
             )
-            exported_program = export(tagged_module, args=inputs, strict=True)
+            exported_program = export(tagged_module, args=inputs)
```
JacobSzwejbka, line 181: does it break with strict=True?

runtime/kernel/test/CMakeLists.txt
```diff
@@ -39,12 +39,6 @@ add_test(kernel_runtime_context_test kernel_runtime_context_test)
 add_executable(
   operator_registry_max_kernel_num_test
   operator_registry_max_kernel_num_test.cpp
-  ../operator_registry.cpp
```
JacobSzwejbka, line 42: Why are all these sources not needed anymore?

runtime/core/exec_aten/testing_util/tensor_util.cpp
```diff
@@ -52,25 +52,52 @@ bool data_is_close(
...
+          return false;
+        }
       }
+
     } else {
-      auto allowed_error = atol + std::abs(rtol * bi);
-      auto actual_error = std::abs(ai - bi);
-      if (!std::isfinite(actual_error) || actual_error > allowed_error) {
-        return false;
+      const auto ai = a[i];
```
JacobSzwejbka: I guess youd have to template to avoid this big copy paste?

extension/evalue_util/test/print_evalue_test.cpp
```diff
@@ -267,7 +267,7 @@ TEST(PrintEvalueTest, UnelidedBoolLists) {
   // case; the other scalar types use the same underlying code, so they don't
   // need to test this again.
   {
-    EValue value(ArrayRef<bool>(list.data(), 0ul));
+    EValue value(ArrayRef<bool>(list.data(), (size_t)0ul));
```
JacobSzwejbka: nit: c cast

test/models/export_delegated_program.py
GregoryComer: It works, it just requires all of the MSVC environment as it triggers compilation during export, which is a bit of a pain during build time as I was hitting some conflicts with clang-cl's setup. This path is covered in CI in a few places, so I figured it's safe (and nicer) to just use non-strict, which is the default. Likely the reason it's using strict is the codemod from when they switched the default.

runtime/kernel/test/CMakeLists.txt
GregoryComer: They're included in the executorch_core target and it causes duplicate symbol errors. Not sure why these are here to begin with.

export/tests/test_target_recipes.py
```diff
@@ -26,12 +31,14 @@ def setUp(self) -> None:
...
-        # pyre-ignore
-        recipe_registry.register_backend_recipe_provider(CoreMLRecipeProvider())
+        if sys.platform != "win32":
+            # pyre-ignore
+            recipe_registry.register_backend_recipe_provider(CoreMLRecipeProvider())
 
     def tearDown(self) -> None:
         super().tearDown()
 
+    @unittest.skipIf(sys.platform == "win32", "Core ML is not available on Windows.")
```
JacobSzwejbka, line 41: This seems like evidence that we shouldnt have a mono recipe test with real backends.

export/target_recipes.py
```diff
@@ -94,6 +100,9 @@ def get_ios_recipe(
         recipe = get_ios_recipe('ios-arm64-coreml-int8')
         session = export(model, recipe, example_inputs)
     """
+    if sys.platform == "win32":
```
JacobSzwejbka: Do we expect lowering to work in general on win32? I dont think pytorch core promises this.

GregoryComer: I did ask the compiler team a little while back, and their answer was that export should work and is intended to be supported on Windows, although it's less tested.

export/tests/test_target_recipes.py
GregoryComer: Yeah, might make sense to refactor it as a follow up. CC @abhinaykukkadapu Do you have any preferences on handling of platform-specific recipes?

runtime/core/exec_aten/testing_util/tensor_util.cpp
GregoryComer: Yeah, it could be a little more concise. I'm happy to take a follow-up to shrink it a little more, if desired.

export/target_recipes.py
abhinaykukkadapu: Can someone create a lowered pte file targeting coreml backend on a windows platform and subsequently use that on a platform where coreml is available, if that is the case, we don't need this check, if we never tested windows path, i'm fine with this.

export/tests/test_target_recipes.py
abhinaykukkadapu: I was under the impression that lowering works across all platforms, although i haven't thought about windows tbf.

> This seems like evidence that we shouldnt have a mono recipe test with real backends.

these tests only call forward if those specific et backends are available, check here: https://github.com/pytorch/executorch/blob/main/export/tests/test_target_recipes.py#L103, otherwise these tests just lowers the model.

@metascroy do you know if coreml tools work on a windows platform?

GregoryComer: Thanks. From what I read, coremltools AOT flow is not available on Windows.

export/target_recipes.py
GregoryComer: Thanks. From https://github.com/apple/coremltools/issues/1281, coremltools doesn't seem to offer Windows support at all, so I think we will need to gate it out.

runtime/core/exec_aten/testing_util/tensor_util.cpp
GregoryComer: I've taken another pass at this and consolidated it. I'm open to ideas if there's a better way.

extension/evalue_util/test/print_evalue_test.cpp
GregoryComer: Resolved.

GregoryComer:
@JacobSzwejbka I've made a few small updates to address comments. Do you have any blocking concerns?

export/target_recipes.py
abhinaykukkadapu: Thanks Gregory, what about QNN, if you rebase the PR, you would also see Android configs, also this check is not needed here as the `IOS_CONFIGS` themselves are filtered for windows above, and we can centralize the management by guarding the `IOS_CONFIGS` map or `ANDROID_CONFIGS` like you did above if any limitations comes up in future, wdyt?

Edit: Android configs, haven't landed yet, so we are good.

GregoryComer: Sure, I'll update it to remove this check. It might be nice to ensure that there is a clear error message if the user tries to do something unsupported, but I'll defer to your judgement on the specifics.

GregoryComer: I've updated the PR to remove this check. Do you have any other concerns with the changes I've made?

abhinaykukkadapu: LGTM, @GregoryComer Just curious, how are you gating entire coreml backend, are you not building that folder at all?

GregoryComer: For the python code, it does all get put into the wheel. The files are there on Windows, they just can't be imported due to coremltools import failing. I imagine we should either throw a nice error message on Windows if someone tries to import it or maybe exclude it from the wheel. Probably not GA-critical, though.

GregoryComer:
After rebase and minor cleanup, I'm seeing one (flaky) test failure relating to etdump generation due to being unable to access a temp file because there's another open handle to it. Tracking this in https://github.com/pytorch/executorch/issues/13883. I've updated and am re-running CI now. If no additional objections, I hope to land this shortly, pending final approval from reviewers.

JacobSzwejbka (approved):

mergennachin (approved):

PR #13924, opened 2025-09-04
Delete copy of quantized SDPA in torchao/experimental

metascroy:
Summary:
This deletes the SDPA/matmul kernels in torchao/experimental.

These kernels were already migrated to torchao/csrc/cpu/torch_free_kernels in the first diff, but we needed to keep them around in torchao/experimental to do a pin bump in ExecuTorch OSS after the first diff lands.

Differential Revision: D81640227


kimishpatel (approved):

kimishpatel (approved):

PR #13928, opened 2025-09-04
Android audio llm

kirklandsign:
### Summary
Need to re-think about API. Get rid of prefill stuff?

### Test plan
CI


kirklandsign:
Use https://github.com/pytorch/executorch/pull/14136 instead

PR #13935, opened 2025-09-04
Allow zero-element tensors for shared inputs.

hsharma35:
Summary: Adds support for the case where method inputs are zero-element. Similar to https://github.com/pytorch/executorch/pull/13623, but adds support for shared input instead.

Differential Revision: D81653487

### DIFF
--- a/runtime/core/exec_aten/util/tensor_util_portable.cpp
+++ b/runtime/core/exec_aten/util/tensor_util_portable.cpp
@@ -147,12 +147,17 @@ Error share_tensor_data(
       t_dst.nbytes(),
       t_src.nbytes());
 
+  // Either the t_src is empty or contains valid data.
   ET_CHECK_OR_RETURN_ERROR(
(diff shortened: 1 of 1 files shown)


runtime/core/exec_aten/util/tensor_util_portable.cpp
```diff
@@ -147,6 +147,11 @@ Error share_tensor_data(
       t_dst.nbytes(),
       t_src.nbytes());
 
+  if (t_src.numel() == 0) {
```
JacobSzwejbka: Wouldnt you still want to alias the pointers? You can have numel == 0 but not have a null data pointer in the world of dynamic shape.

JacobSzwejbka: Oh its the reverse case thats the motivation? You want to be able to call this on 2 empty tensors?

JacobSzwejbka (approved):

PR #13938, opened 2025-09-04
Arm backend: Move tosa files to tosa/ directory

Sebastian-Larsson:


zingo:
Fails are unrelated

zingo (approved):

digantdesai:
will cause problems internally, let's see. I hope I can fwd fix.

shoumikhin:
@zingo @digantdesai need to address https://github.com/pytorch/executorch/blob/b02db12227eb3425f926fc52cfd42859f04d672b/backends/arm/TARGETS#L40-L41 at least

zingo:
Sorry we cant build or see the error but we tried to fix it with this  https://github.com/pytorch/executorch/pull/14006

PR #13942, opened 2025-09-04
NXP backend: Resolve limitations of uncertain tensor formats

MartinPavella:
### Summary
This PR resolves format related issues by inferring the format (NCHW/NHWC) for all nodes before partitioning. These formats are then used by the NeutronPartitioner to accurately determine which nodes are supported on Neutron.

### Test plan
Unit tests provided, and correct function is tested by nearly every test in the nxp backend.

### DIFF
--- a/backends/nxp/_passes/remove_getitem_pass.py
+++ b/backends/nxp/_passes/remove_getitem_pass.py
@@ -1,0 +1,103 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright 2025 NXP
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+from executorch.backends.nxp.backend.node_format_inference import (
+    NodeFormat,
+    NXP_NODE_FORMAT,
+)
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class RemoveGetItemPass(ExportPass):
+    """
+    This remove item is used to remove getitem operator for max_pool2d_with_indices.default operator, and replace it with a single operator,
+    that extracts the first output. More specifically, we are only getting the first output from aten::maxpool2d operator.
+    Before Pass:
+        MaxPool2d ---> GetItem[max_values, max_indexes]
+    After Pass:
+        MaxPool2d -> max_values
+    """
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        module = graph_module
+        for node in module.graph.nodes:
+            if node.op == "call_function":
+                if (
+                    node.target.__name__ == "aten.max_pool2d_with_indices.default"
+                    or node.target.__name__ == "aten.max.dim"
+                ):
+                    users = list(node.users.keys())
+
+                    if len(users) != 1:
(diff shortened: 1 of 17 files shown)


roman-janik-nxp (approved):

backends/nxp/backend/node_format_inference.py
```diff
@@ -77,7 +76,15 @@ def identify_node_formats(self) -> dict[Node, NodeFormat]:
...
-        return self._node_format_mapping
+        for node in self._nodes:
+            if self._get_node_op_type(node) is None:
+                continue
+            if not hasattr(node, "meta"):
+                logging.warning(f"Node `{node}` does not have the `meta` attribute.")
+                node.meta = {}
+            if NXP_NODE_FORMAT not in node.meta:
+                logging.warning(f"Node `{node}` does not have inferred format.")
+                node.meta[NXP_NODE_FORMAT] = NodeFormat.NONE
```
robert-kalmar, line 83: As we now perform the node format inference during partition (that is on the whole Edge Program), it is likely that some nodes wont have the format determined, as the NodeFormat inference algorithm does not know them. Right? 

We should make sure we stop the `channel_first` tag propagation on unknown operator, as we cannot determine if it propagates the `channel_first` or stops it. As example, the Reshape stops the propagation of channel first tag. But if we would not know the `Reshape`, op we were incorrectly propagate the `channel_first` tag behind it in the compute path. 
So we must defensively stop the propagation at every unknown node. Is my thought process correct?

backends/nxp/_passes/remove_getitem_pass.py
```diff
@@ -0,0 +1,103 @@
...
+        MaxPool2d -> max_values
+    """
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        module = graph_module
+        for node in module.graph.nodes:
+            if node.op == "call_function":
+                if (
+                    node.target.__name__ == "aten.max_pool2d_with_indices.default"
+                    or node.target.__name__ == "aten.max.dim"
```
robert-kalmar, line 34: Why do we handle the aten.max.dim too? Is it a loftover from original pass?

MartinPavella: Yes, it was in the original file. I wanted to make as few changes as possible, as it is not the main focus of this PR.

backends/nxp/backend/node_format_inference.py
MartinPavella: Yes, you are correct, with a few caveats.
The fact that we currently propagate the format through unknown nodes, should never cause crashes,  as our format handling system is quite robust. It can, however, result in unnecessary transpositions. As the "unknown" operators will inevitably not be delegated, they will split the graph, resulting in multiple delegated partitions. It is possible (and likely) that one of these partitions requires NHWC, which is propagated to the second partition (via the "unknown" node), but the second partition doesn't require NHWC. If we keep our format inference as is, unnecessary transpositions would have to be done at the inputs and outputs of the second partition.

I will update the code to not propagate the format through "unknown" operators.

MartinPavella:
@robert-kalmar I have added the changes to not infer tensor format for unknown operations.

robert-kalmar (approved):

MartinPavella:
The failing tests seem unrelated. Merging the PR.

PR #13943, opened 2025-09-04
Arm backend: Auto generate cycle table if ETDump is used

zingo:
### Summary
If you use **run.sh** with **--etdump** e.g. something like this:

`examples/arm/run.sh --et_build_root=arm_test/test_run --target=ethos-u55-256 --model_name=mv2 --etdump`

The script will dig out the ETDump file from the FVP logs and use it to produce a cycle table automatically.

Please remember the the CorstoneFVP is not a cycle accurate simulator and the CPU cycle numbers can not be trusted, please run on real hardware or FPGA to get valid number. Running it on a FVP is mostly useful to check that the tools/flow work and not the actual CPU cycle numbers.

### Test plan
This is tested by the test scripts


oscarandersson8218 (approved):
Looks good to me. CI failures seem unrelated.

PR #13959, opened 2025-09-04
Disable xnnpack jobs in macos

metascroy:
This PR disables the flakey job: qwen3-1.7b|xnnpack|--quantize

It also renames test-huggingface-transformers-coreml to the more accurate name test-huggingface-transformers-macos.

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -823,10 +823,10 @@ jobs:
           --tsv_path ${TSV_PATH}
         echo "::endgroup::"
 
-  test-huggingface-transformers-coreml:
+  test-huggingface-transformers-macos:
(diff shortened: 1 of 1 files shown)


.github/workflows/trunk.yml
```diff
@@ -844,10 +844,10 @@ jobs:
           # phi4-mini|xnnpack|--quantize,
           # smollm2-135m|xnnpack|--quantize,
           # smollm3-3b|xnnpack|--quantize,
+          # qwen3-1.7b|xnnpack|--quantize,
```
digantdesai, line 847: This list is growing, so sooner or later we have to debug :)

digantdesai (approved):

PR #13965, opened 2025-09-04
Fix qwen2_5_xnnpack_q8da4w.yaml

jackzhxng:


examples/models/qwen2_5/config/qwen2_5_xnnpack_q8da4w.yaml
```diff
@@ -8,4 +8,7 @@ model:
 
 backend:
   xnnpack:
-    enabled: True
\ No newline at end of file
+    enabled: True
+
+quantization:
```
cccclai, line 13: For llama, we use
```
quantization:
  qmode: 8da4w
  group_size: 128
  embedding_quantize: 4,32
```
Is 8da4w good enough for accuracy?

jackzhxng: Yeah we can do that as well. Caveat though that I remember the 0.5B version not reacting well to embedding quantization

cccclai: Ah yeah we don't want embedding to quantize for sure. But I think we need group wise quantization. If we just do `qmode: 8da4w`, is it using group wise quantization with a default group size?

jackzhxng: Yeah, 128 would be the default group size

cccclai (approved):
Makes sense. Thanks!

PR #13967, opened 2025-09-04
Update OSS flow to call the right APIs

mcremon-meta:
Summary: Reorganize the calls so that we don't run into issues calling `eval()`

Reviewed By: zonglinpeng

Differential Revision: D81698817

### DIFF
--- a/backends/cadence/aot/compiler.py
+++ b/backends/cadence/aot/compiler.py
@@ -388,9 +388,6 @@ def quantize_and_export_to_cadence(
     )
 
 
-# Export the model and lower it to an EdgeProgramManager (in edge IR), and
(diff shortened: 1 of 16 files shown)


zonglinpeng (approved):

mergennachin:
@mcremon-meta Perhaps we can start thinking about having an OSS CI test?

mergennachin:
Looks like current tests will run internally but not in OSS. Since you're starting to have OSS contributors, setting up OSS test suite would be good too so that you're all on the same page.

mcremon-meta:
> @mcremon-meta Perhaps we can start thinking about having an OSS CI test?

agreed! We have definitely thought about it :) and it's planned to start this half. Hoping to get some early coverage by EoY and keep improving from there next year.

PR #13982, opened 2025-09-05
[Backend Tester] Don't run fp64 model tests

GregoryComer:
They add to a lot of failures and double precision isn't commonly used. We can look at re-enabling these in the future. Ideally all backends would just not delegate if they can't run it.


GregoryComer:
* #13992
* #13991
* #13990
* #13989
* __->__ #13982

mergennachin (approved):

PR #13985, opened 2025-09-05
Arm backend: Replace asserts with report_reject in operator_support

Sebastian-Larsson:
Remove redundant check `node.target in self.targets`, as well as replacing asserts with proper report_reject. This way the graph won't be stopped from lowering, but the operators will instead end up on CPU.


zingo (approved):

zingo:
Fails unrelated

PR #13989, opened 2025-09-05
[Backend Tester] Skip in-place activation tests due to lack of support in ET

GregoryComer:
These tests are failing for all backends due to lack of defunctionalization support for in-place activations. This PR disables them to reduce noise. They can be re-enabled in the future when supported. Tracking in https://github.com/pytorch/executorch/issues/11700.

PR #13990, opened 2025-09-05
[Backend Tester] Skip inf and nan operator tests

GregoryComer:
Skip tests that validate inf and nan edge cases. In hindsight, it's not reasonable to validate these on backends, with varying quantization and precision. I've marked them as skip to allow for manually running as they might have some conceptual value to backend authors.


GregoryComer:
* #14200
* #14190
* #14185
* #14137
* #13992
* #13991
* __->__ #13990
* #13989

digantdesai (approved):

PR #13991, opened 2025-09-05
[Backend Tester] Add error message field to CSV report

GregoryComer:
Include a (truncated) error message in the CSV. This should make it easier to parse at a glance.

PR #13992, opened 2025-09-05
[Backend Tester] Add flow-specific skipped tests, skip argmin/max on Core ML due to native crashes, float16/64 on Vulkan

GregoryComer:
Argmin and argmax are crashing on Core ML. Native crashes are hard to recover from. In the medium term, I'd like to look at using pytest-xdist, which, by nature of process isolation, can recover. But for now, I'm adding a way to skip tests on specific backend flows to allow running the remaining tests. I've also skipped Vulkan float16 and float64 tests, as they crash when using swiftshader (used in the CI for software execution of the Vulkan compute shaders). This should be re-visited in the future.

There is also a little technical debt in the code duplication between operator and model tests. This will also be cleaned up shortly.


GregoryComer:
* #14200
* #14190
* #14185
* #14137
* __->__ #13992
* #13991

backends/test/suite/flows/coreml.py
```diff
@@ -19,6 +19,7 @@ def _create_coreml_flow(
             CoreMLTester, minimum_deployment_target=minimum_deployment_target
         ),
         quantize=quantize,
+        skip_patterns=["test_argmin", "test_argmax"],
```
digantdesai, line 22: How does this play when doing a cross delegate % - pass like comparison?

backends/test/suite/operators/__init__.py
```diff
@@ -97,6 +97,9 @@ def _make_wrapped_test(
 ):
     def wrapped_test(self):
         with TestContext(test_name, test_base_name, flow.name, params):
+            if flow.should_skip_test(test_name):
+                raise unittest.SkipTest(f"Skipping test due to matching flow {flow.name} skip patterns")
```
digantdesai: will this be counted and reported as SKIPPED in the report?

digantdesai (approved):
stamping to unblock.

backends/test/suite/operators/__init__.py
GregoryComer: Yeah, it's reported as skipped.

backends/test/suite/flows/coreml.py
GregoryComer: It's not directly reported as a fail. It's not ideal, but native crashes are unrecoverable without process isolation, so I've been accounting for this in the reported numbers. My plan is to switch to pytest with parallel execution to provide process isolation. But, for now, hiding those 2 faiing tests lets the run at least complete in CI.

PR #14006, opened 2025-09-05
Arm backend: Fix buck2 files after moving TOSA partitioner

zingo:
This try to fix broke buck2 builds

### DIFF
--- a/backends/arm/TARGETS
+++ b/backends/arm/TARGETS
@@ -1,17 +1,11 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+#
 # @noautodeps
 load("@fbcode_macros//build_defs:python_library.bzl", "python_library")
 
-python_library(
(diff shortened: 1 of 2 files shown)


zingo:
We have unfortunally no way of knowing if this is correct as the github builds are not buck2 based, but we have tried to dry-code what we think will work for you.

shoumikhin:
@digantdesai can you please verify if this change fixes the internal issues?

zingo:
@shoumikhin @digantdesai as we soon are moving out of the local working window :)  feel free to just merge this directly when you approve it and not wait for me, as I assume you want to fix this faster them my response time might be, Ill try to keep an eye on it anyway if you don't but it might take a bit of time :)

zingo:
Fails are unrelated

digantdesai:
Thanks @zingo for the bold attempt, while flying blind. Let me test it.
I think we should enable buck2 AoT unit tests (at least for TOSA) in OSS. Let me do it next week.

zingo:
A better fix for this was merged https://github.com/pytorch/executorch/pull/14031

PR #14017, opened 2025-09-05
Update default executor runner with new optional options

mansnils:
* By default not all outputs are printed. Adds option for printing all output. 
* Adds option to save output to file.
* In the executor runner all inputs get hard coded to ones. Adds optional input option, in which case tensor inputs will be written from supplied binary input files.

For Arm backend update the Arm VKML unit test runner as a user with real inputs and file output. Enables add/acos_unit tests to run on Vulkan runtime that depend on this.
 

zingo:
Hi @digantdesai this is touching stuff outside Arm backend and we need help reviewing it.

backends/arm/test/ops/test_acos.py
```diff
@@ -102,8 +103,12 @@ def test_acos_vgf_FP(test_data: Tuple):
         [],
         [],
         tosa_version="TOSA-1.0+FP",
+        run_on_vulkan_runtime=True,
     )
-    pipeline.run()
+    try:
```
digantdesai, line 108: Nit: can we do it like this - https://github.com/pytorch/executorch/blob/3c533aae0ca7f28c48b4b298397e5ce5aa6af999/backends/arm/test/tester/test_pipeline.py#L285

Else I imagine we have to write this for every single test..

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -49,6 +50,16 @@ DEFINE_string(
     model_path,
     "model.pte",
     "Model serialized in flatbuffer format.");
+DEFINE_string(inputs, "", "Comma-separated list of input files");
```
digantdesai, line 53: JFYI you can also use bundled-io in the PTE.

digantdesai (approved):
LGTM, seems like BC changes for the runner and the extn/utils, make sure the CI is green.

backends/arm/test/ops/test_acos.py
mansnils: Yes indeed we shall not do a try/except in every unit test. Just wanted to enable a test that tested this. We will either do a central try/except or add xfails for remaining tests in coming patches.

examples/portable/executor_runner/executor_runner.cpp
mansnils: Yes, perhaps add that in a separate patch

backends/arm/test/ops/test_acos.py
zingo: Yes lets do keep that separately.

examples/portable/executor_runner/executor_runner.cpp
zingo: Outputs can also be saved with ETDump  but that is maybe a bigger/separate patch this file should probably be re-synced with examples/devtools/example_runner/example_runner.cpp  someday.

Or @digantdesai do you want to avoid adding "inputs" as an argument and only handle it via BundleIO? 

E.g. is this something you prefere us to fix now in this PR and later after 1.0/GA?

zingo: Just spoted your "LGTM" so that answers my question :)

zingo:
Test fails are unrelated.

zingo:
We are pausing the merge a short bit, we want to make sure the testing also work if VGF tools/VulkanML is installed, and it currently isn't yet in the GitHub testing. But  will be in a separate PR.

UPDATE: should be merged at the same time / together with https://github.com/pytorch/executorch/pull/14116

PR #14021, opened 2025-09-05
Fix some CI tests - don't raise RuntimeError on ISSRuntimeFailure + split ISS...Failure into Expected and Unexpected versions

ThomasJannaud:
Summary:
- For some tests, we except a ISSRuntimeFailure. However we re-raise this as a Runtime exception in some cases, which fails some tests
- We would raise ISSRuntimeFailure both in expected and unexpected cases (and the meaning was that it was expected)

Reviewed By: hsharma35

Differential Revision: D81754432


ThomasJannaud:
abandoned

PR #14022, opened 2025-09-05
OpenVINO Export Llama Support

cavusmustafa:
### Summary

This PR includes the changes below:
-  Initial export llama support with OpenVINO backend. Please find the instructions on how to export and execute inference in OpenVINO backend examples documentation: https://github.com/cavusmustafa/executorch/blob/openvino_llama_support/examples/openvino/README.md#export-llama-with-openvino-backend
- Improvements for NNCF quantizer for exporting quantized models using export llama.
- Build script improvements for better user experience. (`backends/openvnio/scripts/openvino_build.sh`)
- Documentation updates related to the changes provided in this PR.

Release notes: backends

### Test plan
The features provided in this PR are tested using the instructions provided in the readme files below. Export llama functionality with OpenVINO backend is not part of the backend test suite for the time being due to model file dependencies.
OpenVINO Backend Setup:
https://github.com/cavusmustafa/executorch/blob/openvino_llama_support/backends/openvino/README.md
OpenVINO Backend Example for Llama:
https://github.com/cavusmustafa/executorch/blob/openvino_llama_support/examples/openvino/llama/README.md

### Performance Results
Model: [meta-llama/Llama-3.2-1B-Instruct](meta-llama/Llama-3.2-1B-Instruct)

|   | tokens per second | ms per token |
| -- | -- | -- |
| XNNPACK CPU FP32 | 14.1 | 68 |
| OpenVINO CPU FP32 | 16.6 | 60.3 |
| OpenVINO CPU INT4 | 63.6 | 15.7 |
| OpenVINO NPU INT4 | 42.9 | 23.3 |
| OpenVINO NPU INT4 Channel Wise Quant | 50.4 | 19.8 |

System Config:
Hardware: Intel(R) Core(TM) Ultra 7 258V (Intel Lunar Lake)
Architecture: x86_64, 8 cores, 4.8 GHz max
Memory: 32GB RAM
OS: Ubuntu 24.04

CC: @ynimmaga @suryasidd @anzr299 @daniil-lyakhov @MaximProshin @kimishpatel @cbilgin


kimishpatel:
Thanks for the PR. will do a review. I do see some CI failures. 

Also why NPU int4 is actually slower? is there a config where NPU is actually faster than CPU

kimishpatel:
also can you measure xnnpack cpu int4 perf on the same platform?

suryasidd:
> Thanks for the PR. will do a review. I do see some CI failures.
> 
> Also why NPU int4 is actually slower? is there a config where NPU is actually faster than CPU
 
Hey Kimish, for NPU we have updated the numbers. We are seeing much better performance than previously reported with different config options. With channel wise quantization we are getting about 50 tok/s. We're currently working on additional optimizations to further improve NPU performance to get it closer to CPU.

suryasidd:
> also can you measure xnnpack cpu int4 perf on the same platform?

Xnnpack int4 is currently not working on my side. I have added more details in discord.

kimishpatel:
> > Thanks for the PR. will do a review. I do see some CI failures.
> > Also why NPU int4 is actually slower? is there a config where NPU is actually faster than CPU
> 
> Hey Kimish, for NPU we have updated the numbers. We are seeing much better performance than previously reported with different config options. With channel wise quantization we are getting about 50 tok/s. We're currently working on additional optimizations to further improve NPU performance to get it closer to CPU.

Ok let me review the diff. However, I would expect NPU to significantly outperform CPU

swolchok:
> However, I would expect NPU to significantly outperform CPU

for token generation, it depends on how the NPU accesses memory and thus the memory bandwidth, doesn't? Wouldn't the expected gain be in prompt processing? I only see one set of performance numbers, so I assume the numbers given are token generation rather than processing time for a longer prompt.

kimishpatel:
> > However, I would expect NPU to significantly outperform CPU
> 
> for token generation, it depends on how the NPU accesses memory and thus the memory bandwidth, doesn't? Wouldn't the expected gain be in prompt processing? I only see one set of performance numbers, so I assume the numbers given are token generation rather than processing time for a longer prompt.

Yes Scott. Thats right. But besides the issue you mentioned about not having reported prefill numbers, I also see taht NPU is slower than the CPU counterpart.

backends/openvino/README.md
```diff
@@ -59,44 +76,45 @@ cd <your_preferred_install_location>
...
    ```bash
+   openvino_build.sh
+   ```
+- Optionally, `openvino_build.sh` script can be used to build python package or C++ libraries/binaries seperately.
+
+   **Build OpenVINO Backend Python Package with Pybindings**: To build and install the OpenVINO backend Python package with Python bindings, run the `openvino_build.sh` script with the `--enable_python` argument as shown in the below command. This will compile and install the ExecuTorch Python package with the OpenVINO backend into your Python environment. This option will also enable python bindings required to execute OpenVINO backend tests and `aot_optimize_and_infer.py` script inside `executorch/examples/openvino` folder.
+     ```bash
    ./openvino_build.sh --enable_python
    ```
+   **Build C++ Runtime Libraries for OpenVINO Backend**: Run the `openvino_build.sh` script with the `--cpp_runtime` flag to build the C++ runtime libraries as shown in the below command. The compiled libraries files and binaries can be found in the `<executorch_root>/cmake-out` directory. The binary located at `<executorch_root>/cmake-out/backends/openvino/openvino_executor_runner` can be used to run inference with vision models.
```
kimishpatel: why do we need separate `openvino_executor_runner` runner btw?

backends/openvino/README.md
```diff
@@ -59,44 +76,45 @@ cd <your_preferred_install_location>
...
+     ```bash
+   ./openvino_build.sh --cpp_runtime
+   ```
+   **Build C++ Llama Runner**: First, ensure the C++ runtime libraries are built by following the earlier instructions. Then, run the `openvino_build.sh` script with the `--llama_runner flag` to compile the LlaMA runner as shown the below command, which enables executing inference with models exported using export_llama. The resulting binary is located at: `<executorch_root>/cmake-out/examples/models/llama/llama_main`
```
kimishpatel, line 112: same here. why do we need a separate cpp runner for llama models or llms in general

backends/openvino/quantizer/observers.py
```diff
@@ -0,0 +1,186 @@
...
+)
+from nncf.quantization.algorithms.weight_compression.weight_lowering import (  # type: ignore[import-untyped]
+    do_integer_quantization,
+)
+from nncf.tensor.tensor import Tensor as NNCFTensor  # type: ignore[import-untyped]
+from nncf.torch.graph.transformations.commands import (  # type: ignore[import-untyped]
+    PTTargetPoint,
+    TargetType,
+)
+from nncf.torch.quantization.layers import (  # type: ignore[import-untyped]
```
kimishpatel, line 33: are these custom observers necessary? Why can we not use the standard PT2E or torchao's quantization APIs?

backends/openvino/scripts/openvino_build.sh
```diff
@@ -7,55 +7,105 @@ set -e
...
+    # Note: Add any additional configuration options you need here
+    cmake -DCMAKE_INSTALL_PREFIX="${build_dir}" \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DEXECUTORCH_BUILD_OPENVINO=ON \
+          -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
+          -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
+          -DEXECUTORCH_BUILD_EXTENSION_RUNNER_UTIL=ON \
+          -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
+          -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
+          -DEXECUTORCH_BUILD_OPENVINO_EXECUTOR_RUNNER=ON \
```
kimishpatel: if we remove the needs for separate runners it will also be easier since yoiu wont have to maintain your own build scripts. You should really only be requiring to install the deps

examples/models/llama/export_llama_lib.py
```diff
@@ -471,6 +475,14 @@ def build_args_parser() -> argparse.ArgumentParser:
...
         help="Delegate llama2 to qnn backend (Qualcomm), please use it --kv_cahce=True",
     )
+    parser.add_argument("--openvino", action="store_true")
+    parser.add_argument(
+        "--openvino_device",
+        type=str,
+        default="CPU",
+        choices=["CPU", "GPU", "NPU"],
+        help="Specify the device for Openvino (CPU, GPU or NPU).",
+    )
```
kimishpatel, line 485: cc @jackzhxng for export_llama related changes

examples/models/llama/export_llama_lib.py
```diff
@@ -887,6 +907,34 @@ def _to_edge_and_lower_llama_xnnpack(
     return builder.to_executorch(passes=additional_passes)
 
 
+def _to_edge_and_lower_llama_openvino(
```
kimishpatel, line 910: I dont like this as a separate function. We should aim to just leverage partitioners and use `openvino_executor_runner`

kimishpatel (changes requested):
Overall I have two major comments
1. quantization: can we leverage existing quantization API instead of doing somethign custom
2. For runner, why do we need a separate runner. Can we not use existing llm runner?

cc: @jackzhxng

backends/openvino/quantizer/observers.py
anzr299: There are a few reasons for this:
1. The OpenVINO runtime expects a specific decompression subgraph pattern to be placed after quantized weights to correctly optimize it. For this, I have used the `convert` method to define an alternative flow for the convert_pt2e call.
2. We intenally use a slightly different scale calculation formula which is essential to recreate here. This is to make sure models quantized using our path and the OpenVINO quantizer perform the same.
3. The same applies to the formula used for calculating the values of quantized weights.

kimishpatel: Ok makes sense.

kimishpatel:
@cavusmustafa please request review again once the comments are addressed

cavusmustafa:
> @cavusmustafa please request review again once the comments are addressed

@kimishpatel, thank you for taking the time to review and share your feedback! We’ll revise and update based on your suggestions and let you know once the updates are ready.

CC: @suryasidd

jackzhxng (commented):
Everything under `examples/` looks good to me

backends/openvino/README.md
suryasidd: Changed this to use the default runner. I think we created this so we can add more options to the openvino executor. For now we can just use the default runner. Can we also change the default runner to add more information to the output. Currently it just prints the output tensor. Some other stats like first inference latency time, avg inf latency etc might be good to add so users can try it out quickly. We can add these as debug logs as well.

backends/openvino/README.md
suryasidd: We are not using a different binary for llama runner. The same llama_main executable that gets created is used

backends/openvino/scripts/openvino_build.sh
suryasidd: Yes in openvino_build.sh we are just adding the normal executor now. This build script is just to make it easier for users to build the stack with openvino instead of manually using the cmake commands. This is the feedback we got from other users as well when they were trying out the build instructions for openvino.

examples/models/llama/export_llama_lib.py
cavusmustafa: The reason we created a separate function is to be able to use "to_edge_transform_and_lower" as we need the control over disabling decomposition of certain operators: https://github.com/cavusmustafa/executorch/blob/24f67b6bd98bbb70142367200903df9c0bdecc3d/backends/openvino/partitioner.py#L136

Is there any way to pass custom edge_config into the builder? (https://github.com/cavusmustafa/executorch/blob/24f67b6bd98bbb70142367200903df9c0bdecc3d/extension/llm/export/builder.py#L406) If so, we can pass an opt list (preserved ops) as part of the config. But I am not sure if this is possible currently.

cavusmustafa:
> @cavusmustafa please request review again once the comments are addressed

@kimishpatel, we have updated the runner code, and it is ready for further review. Please let us know if any additional changes are needed. In the meantime, we are running further experiments with NPU and will share our findings soon. Thank you.

cc: @suryasidd

examples/models/llama/export_llama_lib.py
kimishpatel: Ok I think I Understand. But we need to fix `_to_edge_and_lower_llama` to leverage to_edge_and_transform. cc @larryliu0820

suryasidd:
> > @cavusmustafa please request review again once the comments are addressed
> 
> @kimishpatel, we have updated the runner code, and it is ready for further review. Please let us know if any additional changes are needed. In the meantime, we are running further experiments with NPU and will share our findings soon. Thank you.
> 
> cc: @suryasidd

Added one more commit with very minor change. Changing the quantization scheme for 4 bit to Symmetric to extract more performance for Llama models on NPU and ratio to 1 to compress all the nodes.

kimishpatel (approved):

mergennachin:
@cavusmustafa don't merge just yet, let me import and do some sanity check on our internal unit tests

cavusmustafa:
> @cavusmustafa don't merge just yet, let me import and do some sanity check on our internal unit tests

Sure @kimishpatel, thank you for reviewing the PR. Let us know of any issues

PR #14036, opened 2025-09-06
Utility function for numerical correctness of edge dialect graphs and reference implementations

DrJessop:
Summary:
Created two utility functions
1. Converts an edge dialect graph into one where custom cadence op nodes are replaced with python references
2. Validates the outputs (and optionally intermediates) of the graphs

Updated two tests in test_replace_ops_passes to utilize these utility functions.

Differential Revision: D81843001


digantdesai:
close this?

PR #14060, opened 2025-09-08
CI: Enable trunk jobs on PRs on main/release

digantdesai:
This is a temporary change to ensure stability for the main branch before 1.0 cut. This will be reverted next week.


.github/workflows/trunk.yml
```diff
@@ -11,6 +11,9 @@ on:
     paths:
       - .ci/docker/ci_commit_pins/pytorch.txt
       - .ci/scripts/**
+    branch:
+      - main
+      - release/*
```
mergennachin: don't you need to remove the paths segment too?

digantdesai: I thought its `paths` or `branches` triggers but I don't see it in the CI but it may trigger after landing?

mergennachin (approved):
Sounds good

PR #14064, opened 2025-09-08
Arm backend: Align dim_order ops handling

AdrianLundell:
- Removes all use of the _to_copy/clone operator. These ops are lowered to _to_dim_order_copy/ clone_dim_order in the to_edge step when skip_dim_order=False, which we expect.

- Remove all dim_order kwargs in the to_tosa_memory_format_pass since we set our own dim_order.

- Add a pass for storing the initial output dim_order to verify that it has not changed unexpectedly.

- Replace RemoveClonePass with more general RemoveNoopPass which also handles to_dim_order ops casting to the same dtype. Move to after the to_tosa_memory_pass to tidy up after it.

- Change the delegation of to_copy_dim_order_support to pick up casts of the same dtype, as they are removed as noops anyways.

Minor fixes:
- Renames to_copy tests to be picked up by the op name parser.
- Moves cast_int64 pass to after compute_constant_ops_aot to assert the buffer exists in .state_dict before accessing it.
- Replace fatal xfail for test_CLIPTextModelWithProjection with numerical diff xfail


AdrianLundell:
Fails are unrelated

zingo (approved):

PR #14065, opened 2025-09-08
Qualcomm AI Engine Direct - GA Static Olmo-1b

DannyYuyang-quic:
Summary:
- e2e script for GA Static OLMo-1b
 - perf 16a4w block quant token rate in kv mode: ~= 63 tokens/sec(SM8750)
 - acc: PPL ~= (fp: 8.735 -> htp: 9.945) in wikitext dataset
- add model params file & model weight converter
- add workaround pass for LayerNorm without weight & unitest
- fix layernorm op builder & fix layernorm quant annotator

### Test plan
``` bash
python examples/qualcomm/oss_scripts/llama/llama.py -b build-android -s ${SERIAL_NUM} -m SM8750 --decoder_model olmo-1b --model_mode kv --max_seq_len 1024 --prompt "Simply put, the theory of relativity states that" --tasks wikitext --limit 1 --eval_perplexity
```
cc: @haowhsu-quic

### DIFF
--- a/backends/qualcomm/_passes/__init__.py
+++ b/backends/qualcomm/_passes/__init__.py
@@ -28,6 +28,7 @@ from .fold_qdq import FoldQDQ
 from .fuse_consecutive_cast import FuseConsecutiveCast
 from .fuse_consecutive_transpose import FuseConsecutiveTranspose
 from .i64_to_i32 import I64toI32
+from .insert_frozen_layer_norm_weight import InsertFrozenLayerNormWeight
 from .insert_io_qdq import InsertIOQDQ
 from .insert_requantize import InsertRequantize
 from .layout_transform import LayoutTransform
@@ -67,6 +68,7 @@ __all__ = [
     FuseConsecutiveCast,
     FuseConsecutiveTranspose,
     I64toI32,
+    InsertFrozenLayerNormWeight,
     InsertIOQDQ,
     InsertRequantize,
     LayoutTransform,
--- a/backends/qualcomm/_passes/insert_frozen_layer_norm_weight.py
+++ b/backends/qualcomm/_passes/insert_frozen_layer_norm_weight.py
@@ -1,0 +1,64 @@
+# Copyright (c) Qualcomm Innovation Center, Inc.
+# All rights reserved
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import torch
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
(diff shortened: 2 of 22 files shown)


DannyYuyang-quic:
Hi @cccclai,

This PR enables [OLMo-1B](https://huggingface.co/allenai/OLMo-1B-hf) from the [GA list](https://github.com/huggingface/optimum-executorch?tab=readme-ov-file) in the static version. 
The perplexity looks promising, but there's a concern regarding its behavior as a base model(not instructed or fine-tuned). Specifically, it does well for the first ~20 tokens, but after that, it tends to repeat the prompt

``` python
model_id = "allenai/OLMo-1B-hf"
from transformers import AutoTokenizer, AutoModelForCausalLM
model = AutoModelForCausalLM.from_pretrained(model_id).eval()
hf_tokenizer = AutoTokenizer.from_pretrained(model_id)
inputs = hf_tokenizer(prompt, return_tensors="pt")
output_ids = model.generate(
      **inputs,
      max_length=1024,
      eos_token_id=hf_tokenizer.eos_token_id,
      do_sample=False,
)
print(hf_tokenizer.decode(output_ids[0], skip_special_tokens=True))
```
Output:
``` bash
Simply put, the theory of relativity states that the speed of light is the same in all directions.
The speed of light is the same in all directions.
The speed of light is the same in all directions. The speed of light is the same in all directions.
The speed of light is the same in all directions. 
....
....
```
Due to this repetitive behavior, I’ve created this as a draft PR. 
I’d like to check: **do you think it’s appropriate to enable this kind of model?**

cccclai:
I think it is worth enabling it first and improving the accuracy later. If the accuracy isn't great, maybe let's have the default recipe to be 8w weight for now, just so users can get reasonable results from it.

cccclai:
Also did you apply spinquant and seqmse with this model?

cccclai:
cc: @Rohan if you have bandwidth on this

haowhsu-quic:
Hi @cccclai, I think the FP model itself already keeps repeating. Could we switch to `olmo2-1B-instruct`? maybe it would perform better.

cccclai:
> Hi @cccclai, I think the FP model itself already keeps repeating. Could we switch to `olmo2-1B-instruct`? maybe it would perform better.

I see, does this model work better? https://huggingface.co/allenai/OLMo-2-0425-1B

haowhsu-quic:
> > Hi @cccclai, I think the FP model itself already keeps repeating. Could we switch to `olmo2-1B-instruct`? maybe it would perform better.
> 
> I see, does this model work better? https://huggingface.co/allenai/OLMo-2-0425-1B

We'll try both and see which one is more feasible. I think instruct version is empirically better than vanilla one.

PR #14069, opened 2025-09-08
Fix missing kernel header in stack util

jathu:
Summary:
Given `stack_util.h` is exported and it includes `kernel_includes.h`, we need `kernel_includes` in the build dependency.

One solution is to expect the clients who use `slice_util` to also add `kernel_includes`. However, "doing it for them" via exported_deps seems like a nicer solution.

Reviewed By: silverguo

Differential Revision: D81930473

### DIFF
--- a/kernels/portable/cpu/util/targets.bzl
+++ b/kernels/portable/cpu/util/targets.bzl
@@ -302,6 +302,8 @@ def define_common_targets():
         exported_headers = ["stack_util.h"],
         deps = [
             "//executorch/kernels/portable/cpu/util:copy_ops_util",
+        ],
+        exported_deps = [
             "//executorch/runtime/kernel:kernel_includes",
         ],
         visibility = ["//executorch/kernels/portable/cpu/...", "@EXECUTORCH_CLIENTS"],


kernels/portable/cpu/util/targets.bzl
```diff
@@ -302,6 +302,8 @@ def define_common_targets():
         exported_headers = ["stack_util.h"],
         deps = [
             "//executorch/kernels/portable/cpu/util:copy_ops_util",
+        ],
+        exported_deps = [
```
manuelcandales, line 306: why exported_deps rather than adding it to the list of deps?

manuelcandales (approved):

kernels/portable/cpu/util/targets.bzl
jathu: I think this is due to GitHub's poor diff UI, but it currently exists in `deps`. I am moving it to `external_deps`

https://github.com/pytorch/executorch/blob/ea37202c0e8adeaeb06bca95a390d245b8451c5f/kernels/portable/cpu/util/targets.bzl#L299-L308

PR #14070, opened 2025-09-08
custom fbcode serialize stage to run FVP internally on arm ops tests

3l1:
Summary:
update the codebase to allow a system whereby:

- the test inventory in https://www.internalfb.com/code/fbsource/fbcode/executorch/backends/arm/test/ops/, which until now runs on open source CI can all be exercised internally on our internal diff-time sandcastle and CI runs
- those tests can **transparently** switch between running an internal version of FVP using Meta's FVP runtimes, build configs and artifacts when run in fbsrouce, but then use the open-source FVp runtime when executed in github opens-ource CI

Differential Revision: D81159036

### DIFF
--- a/backends/arm/test/ops/test_tanh.py
+++ b/backends/arm/test/ops/test_tanh.py
@@ -70,25 +70,27 @@ def test_tanh_tosa_INT(test_data: Tuple):
 
 
 @common.parametrize("test_data", test_data_suite)
+@common.XfailIfNoCorstone300
 def test_tanh_u55_INT(test_data: Tuple):
     pipeline = EthosU55PipelineINT[input_t1](
         Tanh(),
         (test_data(),),
         aten_op,
         exir_ops=[],
-        run_on_fvp=False,
+        run_on_fvp=True,
     )
     pipeline.run()
 
 
 @common.parametrize("test_data", test_data_suite)
+@common.XfailIfNoCorstone320
 def test_tanh_u85_INT(test_data: Tuple):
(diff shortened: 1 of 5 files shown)


backends/arm/test/targets.bzl
```diff
@@ -36,7 +37,7 @@ def define_arm_tests():
     for test_file in test_files:
         test_file_name = paths.basename(test_file)
         test_name = test_file_name.replace("test_", "").replace(".py", "")
-
+
```
digantdesai: nit: whitespace

backends/arm/test/ops/test_tanh.py
```diff
@@ -76,7 +76,7 @@ def test_tanh_u55_INT(test_data: Tuple):
         (test_data(),),
         aten_op,
         exir_ops=[],
-        run_on_fvp=False,
+        run_on_fvp=True,
```
digantdesai, line 80: was there a reason it was disabled in OSS in the first place? If CI is clean I am ok with this PR enabling this.

digantdesai: also add `@common.XfailIfNoCorstone300` as a decorator?

digantdesai (approved):
**Thank @3l1!**
LGTM, stamping to unblock but need some small fixes. And make sure the CI is clean.

mergennachin (approved):

mergennachin:
macos jobs are already fixed in main, merging

PR #14090, opened 2025-09-08
Add EXECUTORCH_THREADPOOL_SIZE options, default to using only performance cores

GregoryComer:
### Summary
Allow build-time configuration of the thread pool size and default to a performance heuristic.

There are 2 modes that we want to support:
 * Heuristic-based. Choose the number of threads according to a performance heuristic. Use threads equal to the number of detected performance cores, but we can continue to iterate on this by adding fine-grained heuristics for specific chipsets in the future.
 * All cores (threads=cores). This is the current behavior. We need to maintain this as an option for some use cases.

With this PR, the default (for OSS) is to use performance cores. From testing with CV models on ~10 representative devices across the performance spectrum, this gives anywhere from parity with the existing perf to up to a 13x speedup (measured on Pixel 6). Many common devices (S20, S22, iPhone 15 Pro) show a 2-4x speedup.

#### Specifying Threadpool Size

To specify the threadpool size, I've added two preprocessor options (and corresponding CMake options):
* `EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES`- Use threads = detected perf cores.
* `EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES` - Use threads = logical cores.

### Test plan
I've verified that logic functions correctly in OSS by building the executor_runner on M1 Mac and observing the existing logging in cpuinfo_utils.

Measuring MobileNet V3 (exported from examples) on XNNPACK, time to run 100 iterations drops from ~450ms to ~230ms on M1 Pro with this change.

### DIFF
--- a/extension/threadpool/CMakeLists.txt
+++ b/extension/threadpool/CMakeLists.txt
@@ -20,6 +20,16 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
+# Threadpool size specifiers. Mutual exclusion is checking in default.cmake.
+# Default to using performance cores if
+# EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES isn't set.
+set(_threadpool_size_flag)
+if(EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES)
+  set(_threadpool_size_flag "EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES")
+else()
+  set(_threadpool_size_flag "EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES")
+endif()
+
 add_library(
   extension_threadpool threadpool.cpp threadpool_guard.cpp thread_parallel.cpp
                        cpuinfo_utils.cpp
@@ -36,7 +46,9 @@ target_include_directories(
     $<BUILD_INTERFACE:${EXECUTORCH_ROOT}/backends/xnnpack/third-party/cpuinfo/include>
     $<BUILD_INTERFACE:${EXECUTORCH_ROOT}/backends/xnnpack/third-party/pthreadpool/include>
 )
-target_compile_definitions(extension_threadpool PUBLIC ET_USE_THREADPOOL)
+target_compile_definitions(
+  extension_threadpool PUBLIC ET_USE_THREADPOOL ${_threadpool_size_flag}
+)
 target_compile_options(extension_threadpool PUBLIC ${_common_compile_options})
 
 # Install libraries
--- a/extension/threadpool/targets.bzl
+++ b/extension/threadpool/targets.bzl
@@ -22,6 +22,7 @@ def define_common_targets():
         name = "threadpool_lib",
         srcs = _THREADPOOL_SRCS,
         deps = [
+            ":cpuinfo_utils",
             "//executorch/runtime/core:core",
             "//executorch/runtime/core/portable_type/c10/c10:c10",
         ],
--- a/extension/threadpool/threadpool.cpp
+++ b/extension/threadpool/threadpool.cpp
@@ -6,6 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
 #include <executorch/extension/threadpool/threadpool.h>
 
 #include <algorithm>
@@ -14,8 +15,25 @@
 
 #include <executorch/extension/threadpool/threadpool_guard.h>
 #include <executorch/runtime/platform/assert.h>
+#include <executorch/runtime/platform/runtime.h>
 
 #include <cpuinfo.h>
+
+// At most one mode should be set.
+#if (                                                       \
+    defined(EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES) && \
+    defined(EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES))
+#error Multiple \
+            threadpool size specifiers are set.At most one of                \
+    EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES,                             \
+    and EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES may be defined.
+#endif
+
+// Default to EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES if no mode is set.
+#if !defined(EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES) && \
+    !defined(EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES)
+#define EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES 1
+#endif
 
 namespace executorch::extension::threadpool {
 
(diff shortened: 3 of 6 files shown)


extension/threadpool/threadpool.h
```diff
@@ -14,6 +14,32 @@
...
+ *          for CMake.
+ *   - -1: Set the thread count equal to the number of cores on the
+ *          active hardware.
+ *
+ *   Any other positive value will be interpreted as a thread count.
+ *   For example, setting EXECUTORCH_THREADPOOL_SIZE=4 will default
+ *   the threadpool to use 4 threads.
+ */
+
+#ifndef EXECUTORCH_THREADPOOL_SIZE
```
kimishpatel: why is this a compile time option? How do you expect users to use it?

kimishpatel (changes requested):
left a comment

swolchok:
I like the 3 mutually exclusive options better. I don't see why we are vague about the "size heuristic" given that we are intentionally making the user select a sizing algorithm; I would just `s/EXECUTORCH_THREADPOOL_USE_SIZE_HEURISTIC/EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES/`. Also, it is unclear to me whether `EXECUTORCH_THREADPOOL_USE_ALL_CORES` will count hyperthreading as creating extra cores; might want to call it `EXECUTORCH_THREADPOOL_USE_ALL_PHYSICAL_CORES` or `EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES` depending on which is correct. (FWIW, not counting hyperthreading would IMO be the correct choice for memory-bandwidth-bound applications.)

Agree with Kimish that it is unclear how users are supposed to correctly choose a fixed size for the thread pool at compile time.

extension/threadpool/threadpool.h
GregoryComer: I've piped it through CMake options now. In theory, I'd prefer to have a runtime API, but I don't really see a good way to do this given that it needs to be set prior to threadpool creation, which is implicit. I'm open to ideas. My expectation is that 99.9% of users will use the default perf core option, so needing to override at build time is an acceptable compromise.

GregoryComer:
> I like the 3 mutually exclusive options better. I don't see why we are vague about the "size heuristic" given that we are intentionally making the user select a sizing algorithm; I would just `s/EXECUTORCH_THREADPOOL_USE_SIZE_HEURISTIC/EXECUTORCH_THREADPOOL_USE_PERFORMANCE_CORES/`. Also, it is unclear to me whether `EXECUTORCH_THREADPOOL_USE_ALL_CORES` will count hyperthreading as creating extra cores; might want to call it `EXECUTORCH_THREADPOOL_USE_ALL_PHYSICAL_CORES` or `EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES` depending on which is correct. (FWIW, not counting hyperthreading would IMO be the correct choice for memory-bandwidth-bound applications.)
> 
> Agree with Kimish that it is unclear how users are supposed to correctly choose a fixed size for the thread pool at compile time.

I've updated to use the discrete options. Regarding physical or logical cores, from a look at cpuinfo, it initially looks inconsistent - Windows counts logical cores but reading the code, it looks like it's counting physical cores on Linux?. I'm going to double check the behavior on a few systems. The x86 system I have easy access to at this exact moment don't report multiple threads per core.

The main intent of this option is to be backwards compatible with the existing behavior. I'd imagine that very few OSS users would want this behavior, as it's slow and there's no OSS way to use fewer threads.

Regarding the user-facing interface, I've added CMake options. I responded to Kimish's comment above - I'd prefer a runtime API in theory but it's awkward to do with the current design. I'm open to ideas.

extension/threadpool/threadpool.h
kimishpatel: Yeah but compile option is just not useful. You compile once deploy on many platforms, you cannot possibly know what the right size is

kimishpatel:
> whether `EXECUTORCH_THREADPOOL_USE_ALL_CORES` will count hyperthreading as creating extra cores

normally it would

kimishpatel:
> > whether `EXECUTORCH_THREADPOOL_USE_ALL_CORES` will count hyperthreading as creating extra cores
> 
> normally it would

As in it would reported as separate cores. you will have to explicitly ask, I think, for actual core count

extension/threadpool/CMakeLists.txt
```diff
@@ -20,6 +20,18 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
+# Threadpool size specifiers. Mutual exclusion is checking in default.cmake.
+set(_threadpool_size_flag)
```
kimishpatel, line 26: where is this used?

kimishpatel (commented):
I am not convinced that using `EXECUTORCH_THREADPOOL_SIZE` as build time option is a good. For oss users that are trying to use this option, it feels you are better off with using the unsafe option.

The other possibility is that set_num_thread option will create a new threadpool and use that as a current threadpool without destorying the existing one. 

In any case, I think the other two compile time options look good to me and I am not sure if we have to add `EXECUTORCH_THREADPOOL_SIZE` right now. So the immediate need should be met by the other two, no?

GregoryComer:
> I am not convinced that using `EXECUTORCH_THREADPOOL_SIZE` as build time option is a good. For oss users that are trying to use this option, it feels you are better off with using the unsafe option.
> 
> The other possibility is that set_num_thread option will create a new threadpool and use that as a current threadpool without destorying the existing one.
> 
> In any case, I think the other two compile time options look good to me and I am not sure if we have to add `EXECUTORCH_THREADPOOL_SIZE` right now. So the immediate need should be met by the other two, no?

Sure. I don't necessarily see a strong use case for `EXECUTORCH_THREADPOOL_SIZE`, so I'm happy to drop it. The perf core option as default is what I really care about. I'll update the PR.

extension/threadpool/targets.bzl
```diff
@@ -34,7 +35,10 @@ def define_common_targets():
         ],
         exported_preprocessor_flags = [
             "-DET_USE_THREADPOOL",
-        ],
+        ] + (
+            # Default to perf heuristic (0) in OSS, all cores (-1) otherwise.
```
mergennachin: update this inline comment

extension/threadpool/CMakeLists.txt
GregoryComer: This is a default value to make sure the CMake variable is defined in case it's not otherwise set below by one of the specific options. It's used on line 48 as part of target_compile_definitions.

extension/threadpool/targets.bzl
GregoryComer: Updated.

GregoryComer:
@kimishpatel I've updated to remove the explicit thread count option.

extension/threadpool/threadpool.cpp
```diff
@@ -101,7 +118,18 @@ ThreadPool* get_threadpool() {
...
+  // options. See the description in threadpool.h.
+
+#if defined(EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES)
+  // Use threads=cores.
+  static int num_threads = cpuinfo_get_processors_count();
+#else
+  // Set threads equal to the number of performance cores.
+  static int num_threads =
+      ::executorch::extension::cpuinfo::get_num_performant_cores();
+#endif
```
kimishpatel, line 134: default behavior than seems get num performance cores? I thought you would want this the other way around. That is by default you have logical cores and in oss cmake we can make performant core as default build option.

Issue is that for internal uses now you only have performant cores

kimishpatel (changes requested):
Left one comment around preserving the default behavior for internal uses

extension/threadpool/threadpool.cpp
GregoryComer: In extension/threadpool/targets.bzl, I changed it to define EXECUTORCH_THREADPOOL_USE_ALL_LOGICAL_CORES when not in OSS, so that should cover this case. If there's a better way to ensure this, I'm definitely open to it. I could add an API to retrieve the threadpool size and add an internal test to verify the behavior, if you'd like.

GregoryComer:
@kimishpatel I believe I've addressed all the comments. Do you have any additional concerns?

kimishpatel (approved):
synced offline for making default behavior chose all core and add cmake flag to chose perf core for oss

GregoryComer:
Fighting some out of sync issues with the imported diff, maybe due to diff train issues. Finally got it resolved now. I haven't touched this PR's contents really at all since earlier today and CI was green. Going to merge as this is a critical feature.

PR #14105, opened 2025-09-09
change docker image for using android ndk

Jiseong-oh:
### Summary
Test CI for changing docker image


.ci/scripts/setup-samsung-linux-deps.sh
```diff
@@ -11,7 +11,7 @@ set -ex
 
 download_ai_lite_core() {
   API_BASE="https://soc-developer.semiconductor.samsung.com/api/v1/resource/ai-litecore/download"
-  API_KEY="kn10SoSY3hkC-9Qny5TqD2mnqVrlupv3krnjLeBt5cY"
```
mergennachin, line 14: hopefully you'll deactivate this key and using a new secret key

once the code is in public domain in github, key will be always public

Jiseong-oh: Yes, once the behavior is confirmed in the CI, I will change the Key and request it from the Stephen.
This pr is for checking CI. it will be deleted after verification is done. plz, refer #14106

PR #14106, opened 2025-09-09
Support more ops and verify more models

Jiseong-oh:
Expands more ops and Verify more models

### Summary
  - Added for more OPs currently supported by Exynos
  - Added functionality for models for Showcase
  - Additional Module for Optimization
  - Fixed ci issue

### Test plan
models="ic3, ic4, resnet18, resnet50, mv3, edsr, dl3, w2l, vit, mobilebert
python -m executorch.examples.samsung.aot_compiler --model_name=$model -c E9955


examples/samsung/aot_compiler.py
```diff
@@ -81,6 +74,8 @@ def save_tensors(tensors, prefix, artifact_dir):
     model = model.eval()
     outputs = model(*example_inputs)
 
+    print("start start ...")
```
SS-JIA: remove?

examples/samsung/scripts/mobilebert_finetune.py
```diff
@@ -0,0 +1,261 @@
...
+from executorch.examples.samsung.utils import save_tensors
+from executorch.exir import ExecutorchBackendConfig
+from executorch.extension.export_util.utils import save_pte_program
+from transformers import AutoTokenizer, MobileBertForSequenceClassification
+
+
+# Output from pretrained model exceeds the representation scale of half-float.
+# Finetune bert model on specific task and make output more reasonable for hardware.
+# Here is an example.
+class MobileBertFinetune:
```
SS-JIA, line 28: very cool/useful to have this script as an example, by the way

SS-JIA (approved):
LGTM. Please remove the print statement and I will merge.

examples/samsung/aot_compiler.py
Jiseong-oh: it's removed.

backends/samsung/builders/op_constant_pad_nd.py
```diff
@@ -0,0 +1,56 @@
...
+from executorch.backends.samsung.serialization.enn_graph_schema import EnnGraph
+from executorch.backends.transforms import get_shape
+
+
+@register_node_visitor
+class ConstantPadNDVisitor(NodeVisitor):
+    target = "aten.constant_pad_nd.default"
+
+    def __init__(self, *args) -> None:
+        super().__init__(*args)
```
swolchok, line 25: I think `__init__` definitions like this can just be omitted

backends/samsung/builders/op_bmm.py
```diff
@@ -0,0 +1,40 @@
...
+        input_id_1 = self.define_tensor(input1, enn_graph, vals_to_ids)
+        input2 = node.args[1]
+        input_id_2 = self.define_tensor(input2, enn_graph, vals_to_ids)
+
+        # output
+        output_id = self.define_tensor(node, enn_graph, vals_to_ids)
+
+        enn_graph.define_op(
+            node.name, "BATCH_MATMUL", [input_id_1, input_id_2], [output_id]
+        )
```
swolchok, line 40: it seems like there's a lot of boilerplate in these visitor definitions. You could package up a few helper subclasses like UnaryOpVisitor, BinaryOpVisitor, etc. that get the operator name ("BATCH_MATMUL" etc.) from a class property similar to the existing `target` property, and then also accommodate the ones with params by having the helper subclass call self.get_params() (default implementation that returns None on the helper subclass) and pass the result to define_op if it isn't None.

backends/samsung/test/ops/test_log_softmax.py
```diff
@@ -0,0 +1,45 @@
...
+from executorch.backends.samsung.test.tester import SamsungTester
+
+
+class LogSoftmax(torch.nn.Module):
+    def __init__(self, dim) -> None:
+        super().__init__()
+        self.module = torch.nn.LogSoftmax(dim=dim)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.module(x)
```
swolchok, line 25: this looks like it could be deleted and replaced with using torch.nn.LogSoftmax directly

PR #14111, opened 2025-09-09
Arm backend: Refactor compile spec handling

Erik-Lundell:
Instead of factories to create list[CompileSpec], use a dataclass that contains all specs directly. This cleans up the API for creating CompileSpecs and aligns it with how the partitioner and quantizer is designed. Keep compile spec in easily handled ArmCompileSpec format as long as possible, only go to list[CompileSpec] format when neccessary for the backend API. ArmCompileSpec is subclassed for each backend to minimize risk of using incorrect compile spec for a specific target.


zingo:
test fails are unrelated

zingo (approved):

zingo:
@Erik-Lundell this seem to need a rebase :( Sorry about that.

backends/arm/arm_backend.py
```diff
@@ -1,245 +0,0 @@
-# Copyright 2023-2025 Arm Limited and/or its affiliates.
```
swolchok, line 1: the buck files need to be updated to reflect this deletion.

swolchok: looks like there are also references in docs/

PR #14116, opened 2025-09-09
Arm backend: Remove _vgf_ op pytest xfails

mansnils:
Use VK_LAYER_PATH instead of VK_ADD_LAYER_PATH when setting up mlsdk deps since looks like some loaders ignore VK_ADD_LAYER_PATH.


zingo (approved):
Should be merged at the same time / together with https://github.com/pytorch/executorch/pull/14017 when tests has passed OK.

PR #14118, opened 2025-09-09
Arm backend: Add pass to handle int64 placeholders

YufengShi-dudu:
- Add InsertInt32CastsAfterInt64PlaceholdersPass to insert an int64->int32 cast node after each int64 placeholder.
- Deprecate the use of InsertCastForOpsWithInt64InputPass and in favor of InsertInt32CastsAfterInt64PlaceholdersPass; replace its usage

Change-Id: Ic092f57b56d1bab0e82205e5dd129c49e50862c0


zingo:
MacOs fails are unrelated

zingo (approved):

PR #14122, opened 2025-09-09
Migrate to pytorch_sphinx_theme2

svekars:


mergennachin:
@svekars

How ready is pytorch_sphinx_theme2? 

We cut our 1.0 branch about two weeks ago. Do you recommend cherry-picking to 1.0?

svekars:
> How ready is pytorch_sphinx_theme2?

We use it in pytorch in the stable branch: https://docs.pytorch.org/docs/stable/index.html and a few other repositories already.
The theme you have right now has an outdated top navigation bar which will not be updated.

GregoryComer:
Regarding the 1.0 release, my opinion is that we should merge these changes soon/now and and pick onto the release branch rather than waiting for after 1.0. As mentioned, if we land on main, we'll want to make sure we pick to make sure doc picks don't conflict. These changes would be good to have for 1.0, as well as for the TOC / doc overhaul that @psiddh is driving.

mergennachin:
Thanks @svekars for the context. Glad that it is already stable.

Okay @GregoryComer @psiddh - please review this and land ASAP in both main and release branches, and start cherry-picking docs changes on top of this PR.

GregoryComer (approved):

GregoryComer:
@svekars We're good to merge the changes. I saw you just rebased the PR. Are there any additional changes you're making, or is this just checking against the latest?

svekars:
@GregoryComer I think this should be good to go.

PR #14123, opened 2025-09-09
AWQ for Qualcomm LlamaModels

rohansjoshi:
Differential Revision: D81645608


cccclai:
@haowhsu-quic here is the awq that @rohansjoshi implemented that also helps with improving the accuracy, what is your take on this?

PR #14128, opened 2025-09-09
Add manifest extension AoT

JacobSzwejbka:
Summary:
Add some infra for us to optionally add some key structured data to the end of a pte. This diff is around enabling users to easily tag their model with a cryptographic signature. Has room to expand later.

A key design motivation is it would be ideal if this is transparent to the rest of the extensions we have today. Im claiming the prime footer real estate for this which is unused today by anything in tree. This should let it be composable with other formats like bundledProgram too.

Differential Revision: D82052721

### DIFF
--- a/extension/manifest/_manifest.py
+++ b/extension/manifest/_manifest.py
@@ -1,0 +1,206 @@
+from dataclasses import dataclass
+from typing import ClassVar, Literal
+
+from executorch.exir._serialize.padding import padding_required
+from executorch.exir._warnings import experimental
+
+# Byte order of numbers written to the manifest. Always little-endian
+# regardless of the host system, since all commonly-used modern CPUs are little
+# endian.
+_MANIFEST_BYTEORDER: Literal["little"] = "little"
+
+
+@dataclass
+class _ManifestLayout:
+    """Python class mirroring the binary layout of the manifest.
+    separate from the Manifest class, which is the user facing
+    representation.
+    """
+
+    EXPECTED_MAGIC: ClassVar[bytes] = b"em00"
+
+    MAX_SIGNATURE_SIZE: ClassVar[int] = 512
+
+    EXPECTED_MIN_LENGTH: ClassVar[int] = (
+        # Header magic
+        4
+        # Header length
+        + 4
+        # program offset
+        + 8
+        # Padding
+        + 4
+        # signature size
+        + 4
+    )
+
+    EXPECTED_MAX_LENGTH: ClassVar[int] = EXPECTED_MIN_LENGTH + MAX_SIGNATURE_SIZE
+
+    signature: bytes
+
+    # The actual size of the signature
+    signature_size: int = 0
+
+    # The size of any padding required
+    padding_size: int = 0
+
+    # Size in bytes between the top of the manifest and the start of the data it was appended to.
+    program_offset: int = 0
+
+    # The manifest length, in bytes, read from or to be written to the binary
+    # footer.
+    length: int = 0
+
+    # The magic bytes read from or to be written to the binary footer.
+    magic: bytes = EXPECTED_MAGIC
+
+    def __post_init__(self):
+        """Post init hook to validate the manifest."""
+        if self.signature_size == 0:
+            self.signature_size = len(self.signature)
+        if self.length == 0:
+            self.length = _ManifestLayout.EXPECTED_MIN_LENGTH + self.signature_size
+
+        # Not using self.is_valid() here to deliver better error messages.
+        if len(self.signature) > _ManifestLayout.MAX_SIGNATURE_SIZE:
+            raise ValueError(
+                f"Signature is too large. {self.signature_size}. Manifest only supports signatures up to {_ManifestLayout.MAX_SIGNATURE_SIZE} bytes."
+            )
+        if self.magic != _ManifestLayout.EXPECTED_MAGIC:
+            raise ValueError(
+                f"Invalid magic. Expected {_ManifestLayout.EXPECTED_MAGIC}. Got {self.magic}"
+            )
+        if self.length < _ManifestLayout.EXPECTED_MIN_LENGTH:
+            raise ValueError(
+                f"Invalid length. Expected at least {_ManifestLayout.EXPECTED_MIN_LENGTH}. Got {self.length}"
+            )
+        if self.length > _ManifestLayout.EXPECTED_MAX_LENGTH:
+            raise ValueError(
+                f"Invalid length. Expected at most {_ManifestLayout.EXPECTED_MAX_LENGTH}. Got {self.length}"
+            )
+        if self.signature_size != len(self.signature):
+            raise ValueError(
+                f"Invalid signature size must match len(self.signature). Expected {len(self.signature)}. Got {self.signature_size}"
+            )
+
+    def is_valid(self) -> bool:
+        """Returns true if the manifest appears to be well-formed."""
+        return (
+            self.magic == _ManifestLayout.EXPECTED_MAGIC
+            and self.length >= _ManifestLayout.EXPECTED_MIN_LENGTH
+            and self.length <= _ManifestLayout.EXPECTED_MAX_LENGTH
+            and self.signature_size >= 0
+            and self.signature_size <= _ManifestLayout.MAX_SIGNATURE_SIZE
+            and self.program_offset >= 0
+            and len(self.signature) == self.signature_size
+        )
+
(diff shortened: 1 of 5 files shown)


extension/manifest/_manifest.py
```diff
@@ -0,0 +1,154 @@
...
+ def is_valid(self) -> bool:
+     """Returns true if the manifest appears to be well-formed."""
+     return (
+         self.magic == _ManifestLayout.EXPECTED_MAGIC
+         and self.length >= _ManifestLayout.EXPECTED_LENGTH
+     )
+
+ def to_bytes(self) -> bytes:
+     """Returns the binary representation of the Manifest. Written
+     bottom up.
```
lucylq: nit, explain why it's written bottom up

"Returns the binary representation of the Manifest. Written bottom up to allow for BC considerations. The compatibility-preserving way to make changes is to increase the header's length field and add new fields at the top. This means we can always check the last n bytes for the magic and size, and then load the full footer."

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,154 @@
...
+         
+         # uint64_t: Signature unique ID for the data the manifest was appended to.
+         self.signature.to_bytes(8, byteorder=_MANIFEST_BYTEORDER)
+         # uint64_t: Size in bytes between the manifest and the data it was appended to.
+         +self.program_offset.to_bytes(8, byteorder=_MANIFEST_BYTEORDER)
+         # uint32_t: Size in bytes between the manifest and the data it was appended to.
+         +self.padding_size.to_bytes(4, byteorder=_MANIFEST_BYTEORDER)
+         # uint32_t: Size of this manifest. This makes it easier to add new
+         # fields to this header in the future. Always use the proper size
+         # (i.e., ignore self.length) since there's no reason to create an
```
lucylq: What is self.length used for?

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,154 @@
...
+   + 8
+ )
+
+ signature: int
+
+ # Size in bytes between the top of the manifest and the end of the data it was appended to
+ padding_size: int = 0
+
+ # Size in bytes between the top of the manifest and the start of the data it was appended to.
+ program_offset: int = 0
```
lucylq: Maybe 'pte_offset'? program could refer to pte without segments.

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,154 @@
...
+   + 4 
+   # program offset
+   + 8
+   # signature
+   + 8
+ )
+
+ signature: int
+
+ # Size in bytes between the top of the manifest and the end of the data it was appended to
```
lucylq: nit: add "The PTE data is aligned to 16 by default."

JacobSzwejbka:
Oh the padding isnt needed since the alignment doesnt matter since we reconstruct byte by byte anyway. Ill remove

GregoryComer:
Is there a strong reason to not include this in the core AOT code (as opposed to a dedicated extension?). I don't have a super strong opinion on this, but I do worry about the growing number of fine-grained extensions being confusing to users and compromising UX.

JacobSzwejbka:
>Is there a strong reason to not include this in the core AOT code (as opposed to a dedicated extension?). I don't have a super strong opinion on this 

I could put the aot stuff in core and put the runtime reader in extension? The modularity is a feature for embedded. We shouldnt expose so many options in mobile builds which is why we have the presets.

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,206 @@
...
+
+from executorch.exir._serialize.padding import padding_required
+from executorch.exir._warnings import experimental
+
+# Byte order of numbers written to the manifest. Always little-endian
+# regardless of the host system, since all commonly-used modern CPUs are little
+# endian.
+_MANIFEST_BYTEORDER: Literal["little"] = "little"
+
+
```
mergennachin, line 12: Write a docblock about an example usage of manifest file for higher layer consumers. Also mention that manifest is a mechanism, not a security policy.  And explicitly say that consumers implements appropriate security for their threat model

```
# 1. Generate PTE file
pte_data = serialize_pte_binary(program)

# 2. Create cryptographic signature of PTE data
signature = sign_with_private_key(pte_data, private_key)  # e.g., RSA, ECDSA

# 3. Append manifest with signature
manifest = Manifest(signature=signature)
pte_with_manifest = append_manifest(pte_data, manifest)

Verification Process

# 1. Extract manifest from end of file
manifest = Manifest.from_bytes(file_data)

# 2. Extract PTE data (using program_offset)
pte_data = file_data[:-(manifest_length + padding)]

# 3. Verify signature with public key
is_valid = verify_signature(pte_data, manifest.signature, public_key)

```

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,206 @@
...
+@dataclass
+class Manifest:
+    """A manifest that can be appended to a binary blob. The manifest contains
+    meta information about the binary blob. You must know who created the manifest
+    to be able to interpret the data in the manifest."""
+
+    # Unique ID for the data the manifest was appended to. Often this might contain
+    # a crytographic signature for the data.
+    signature: bytes
+
```
mergennachin, line 181: Add version

mergennachin: I can imagine people can use this for other use-cases besides security, such as saving arbitrary serializable metadata.

For instance, saving tokenizer.json file location etc.

mergennachin, line 181: Should you add timestamp field too?

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,206 @@
...
+@dataclass
+class _ManifestLayout:
+    """Python class mirroring the binary layout of the manifest.
+    separate from the Manifest class, which is the user facing
+    representation.
+    """
+
+    EXPECTED_MAGIC: ClassVar[bytes] = b"em00"
+
+    MAX_SIGNATURE_SIZE: ClassVar[int] = 512
```
mergennachin, line 22: why is this fixed?

extension/manifest/_manifest.py
```diff
@@ -0,0 +1,206 @@
...
+            # Manifest magic. This lets consumers detect whether the
+            # manifest was inserted or not. Always use the proper magic value
+            # (i.e., ignore self.magic) since there's no reason to create an
+            # invalid manifest.
+            + self.EXPECTED_MAGIC
+        )
+        return data
+
+    @staticmethod
+    def from_bytes(data: bytes) -> "_ManifestLayout":
```
mergennachin, line 128: For large files you have to read the whole thing?

extension/manifest/_manifest.py
mergennachin, line 181: Something like this?

```
  @dataclass
  class Manifest:
      type: str  # "signature", "checksum", "metadata", etc.
      version: int
      payload: bytes 
      timestamp: Optional[int]
      attributes: Dict[str, str] 
```

extension/manifest/_manifest.py
JacobSzwejbka: So we can just do one load at runtime. Instead of 2 loads or a stream.

extension/manifest/_manifest.py
JacobSzwejbka: Could add more methods like from file that just load the last MAX_SIZE bytes. Or stream.

extension/manifest/_manifest.py
JacobSzwejbka: is version the version of the manifest struct or user specified?

If a user wanted to have multiple things then would you expect them to daisy chain manifests?

What is attributes?

JacobSzwejbka: > I can imagine people can use this for other use-cases besides security, such as saving arbitrary serializable metadata.

It wasnt really the intent. I chose this impl here because I wanted a really light weight way to attach security information or other core metadata about the pte.

If we want it to store arbitrary user defined things like a json then I dont really think appending to the .pte is the correct solution, just shove it all in a zip would be my opinion.

extension/manifest/_manifest.py
JacobSzwejbka: 512 should also cover the vast majority of cryptographic signature algorithms I saw.

extension/manifest/_manifest.py
JacobSzwejbka: I dont really expect people to be verifying the signature in python though. Its mostly just there for testing.

PR #14129, opened 2025-09-09
Update torch pin to Sep 06, 2025

mergennachin:
Summary:

Branch cut for PT/PT is Sep 08, so let's bring it forward just a few days before the branch cut.

Once we cut the 1.0 branch cut in ET, we will pin to 2.9 which is basically equivalent of Sep 08


digantdesai (approved):
🥇

swolchok:
Looks like this may have broken test-moshi-linux. Hard to tell because the previous PR had an infra flake. https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=moshi-linux

PR #14131, opened 2025-09-09
Extend data file support for other pybindings

LeeOHzzZ:
Summary:
This diff extends the support for external data_file to other executorch runtime pybindings.

Differential Revision: D82046648


lucylq:
Creating new PR after commandeering diff.

PR #14142, opened 2025-09-10
NXP backend: Update aot_neutron_compile.py example pipeline

StrycekSimon:
### Summary
Updates example in `aot_neutron_compile.py` to reflect current state of NXP backend.

### Test plan
AOT examples should run automatically in CI. You can manually test it using `backends/nxp/run_aot_example.sh`

cc @digantdesai @JakeStevens @robert-kalmar @MartinPavella @roman-janik-nxp

### DIFF
--- a/backends/nxp/tests/executorch_pipeline.py
+++ b/backends/nxp/tests/executorch_pipeline.py
@@ -15,9 +15,6 @@ from executorch.backends.nxp.backend.custom_delegation_options import (
 from executorch.backends.nxp.edge_passes.neutron_edge_pass_manager import (
     NeutronEdgePassManager,
 )
-from executorch.backends.nxp.edge_passes.remove_io_quant_ops_pass import (
-    RemoveIOQuantOpsPass,
-)
 from executorch.backends.nxp.neutron_partitioner import NeutronPartitioner
 from executorch.backends.nxp.nxp_backend import generate_neutron_compile_spec
 from executorch.backends.nxp.quantizer.neutron_quantizer import NeutronQuantizer
@@ -115,12 +112,9 @@ def to_quantized_edge_program(
         edge_compile_config=edge_compile_config,
     )
 
-    edge_program_manager = NeutronEdgePassManager()(edge_program_manager)
-
-    if remove_quant_io_ops:
-        edge_program_manager = edge_program_manager.transform(
-            [RemoveIOQuantOpsPass(edge_program_manager=edge_program_manager)]
-        )
+    edge_program_manager = NeutronEdgePassManager(
+        remove_io_quant_ops=remove_quant_io_ops
+    )(edge_program_manager)
 
     compile_spec = generate_neutron_compile_spec(
         target,
--- a/backends/nxp/tests/executorch_pipeline.py
+++ b/backends/nxp/tests/executorch_pipeline.py
@@ -117,6 +117,11 @@ def to_quantized_edge_program(
 
     edge_program_manager = NeutronEdgePassManager()(edge_program_manager)
 
+    if remove_quant_io_ops:
+        edge_program_manager = edge_program_manager.transform(
+            [RemoveIOQuantOpsPass(edge_program_manager=edge_program_manager)]
+        )
+
     compile_spec = generate_neutron_compile_spec(
         target,
         operators_not_to_delegate=operators_not_to_delegate,
@@ -124,11 +129,6 @@ def to_quantized_edge_program(
     )
     partitioner = NeutronPartitioner(compile_spec, custom_delegation_options)
     edge_program_manager = edge_program_manager.to_backend(partitioner)
-
-    if remove_quant_io_ops:
-        edge_program_manager = edge_program_manager.transform(
-            [RemoveIOQuantOpsPass(edge_program_manager=edge_program_manager)]
-        )
 
     return edge_program_manager
 
--- a/examples/nxp/aot_neutron_compile.py
+++ b/examples/nxp/aot_neutron_compile.py
@@ -23,9 +23,13 @@ from executorch.backends.nxp.nxp_backend import generate_neutron_compile_spec
 from executorch.backends.nxp.quantizer.neutron_quantizer import NeutronQuantizer
 from executorch.examples.models import MODEL_NAME_TO_MODEL
 from executorch.examples.models.model_factory import EagerModelFactory
-from executorch.exir import EdgeCompileConfig, ExecutorchBackendConfig
+from executorch.exir import (
+    EdgeCompileConfig,
+    ExecutorchBackendConfig,
+    to_edge_transform_and_lower,
+)
 from executorch.extension.export_util import save_pte_program
-from executorch.extension.export_util.utils import export_to_edge
+from torch.export import export
 from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e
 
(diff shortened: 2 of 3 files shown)


examples/nxp/aot_neutron_compile.py
```diff
@@ -254,39 +252,42 @@ def _get_batch_size(data):
...
+        module,
+        example_inputs,
+        edge_compile_config=edge_compile_config,
+    )
+
+    logging.debug(f"Exported graph:\n{edge_program_manager.exported_program().graph}")
+
+    edge_program_manager = NeutronEdgePassManager()(edge_program_manager)
+
+    if args.remove_quant_io_ops:
```
roman-janik-nxp: This step is done before delegation, but in `executorch_pipeline.py` it is done after delegation. In our local repo it is placed before delegation. Why are there differences?

examples/nxp/aot_neutron_compile.py
```diff
@@ -254,39 +252,42 @@ def _get_batch_size(data):
...
+        edge_program_manager = edge_program_manager.transform(
+            [RemoveIOQuantOpsPass(edge_program_manager=edge_program_manager)]
+        )
+
+    # 5. Delegate to Neutron
+    if args.delegate:
+        logging.info("Executing Neutron Partitioner and Delegate")
+        edge_program_manager = edge_program_manager.to_backend(
             NeutronPartitioner(
```
roman-janik-nxp: Please unify `edge_program_manager.to_backend()`, `compile_spec` and `partitioner` creation with version in executorch_pipeline.py`.

examples/nxp/aot_neutron_compile.py
StrycekSimon: I think it should be placed before the delegation. @MartinPavella, can you confirm it since you are the author of the updated `executorch_pipeline.py`?

MartinPavella: In `executorch_pipeline`, the removal of I/O quantization **after** partitioning was up-streamed by you, with @skywall as the original author (https://github.com/pytorch/executorch/commit/cf2f170d585a71dae342880aed087d67f1697d20). 
I did not add the IO removal, so I'm not sure which order is preferred. Perhaps Lukas can comment. (But I am leaning towards keeping it consistent with `main-nxp`.)

StrycekSimon: I agree, I will change it.

**Edit: Done ✅**

roman-janik-nxp (changes requested):

examples/nxp/aot_neutron_compile.py
StrycekSimon: Done ✅

roman-janik-nxp (approved):

backends/nxp/tests/executorch_pipeline.py
```diff
@@ -119,14 +119,15 @@ def to_quantized_edge_program(
         operators_not_to_delegate=operators_not_to_delegate,
         neutron_converter_flavor=neutron_converter_flavor,
     )
-    partitioner = NeutronPartitioner(compile_spec, custom_delegation_options)
-    edge_program_manager = edge_program_manager.to_backend(partitioner)
 
     if remove_quant_io_ops:
         edge_program_manager = edge_program_manager.transform(
             [RemoveIOQuantOpsPass(edge_program_manager=edge_program_manager)]
```
digantdesai: unrelated to this PR but why is this not part of the NeutronEdgePassManager?

examples/nxp/aot_neutron_compile.py
```diff
@@ -254,39 +252,42 @@ def _get_batch_size(data):
...
-    logging.debug(f"Exported graph:\n{edge_program.exported_program().graph}")
+
+    logging.debug(f"Exported graph:\n{edge_program_manager.exported_program().graph}")
+
+    edge_program_manager = NeutronEdgePassManager()(edge_program_manager)
 
     if args.remove_quant_io_ops:
-        edge_program = edge_program.transform(
-            [RemoveIOQuantOpsPass(edge_program_manager=edge_program)]
+        edge_program_manager = edge_program_manager.transform(
```
digantdesai: as I said before can we move the pass inside the passmanager and add a `remove_quant_io` as an arg to it?

digantdesai (approved):

backends/nxp/tests/executorch_pipeline.py
StrycekSimon: Good point, I probably did put the `RemoveIOQuantOpsPass` call on a wrong place. I will double check it.

robert-kalmar: The `RemoveIOQuantOps` is intentionally not part of the `NeutronPassManager` default passes, as this changes the model API:
Normally when the quantization is performed and consequently the delegation to Neutron backend the model input and output type is untouched - e.g. FP32. There is a Quantize node doing the input quantization for the rest of the (optionally) Neutron compute. 
This pass removes the Quantize and Dequantize nodes from the I/O tensors, and **changes the datatype to int8** . 

It can be applied by the `NeutronPassManager`, but with user choice option. Not by default.

digantdesai: As I said, we can add an optional arg to the PassManager for this pass. It feels like someone has to go out of their way to do this.

StrycekSimon: I moved the application to the PassManager itself. Now it is an opt-in boolean.

StrycekSimon:
I refactored the `aot_neutron_compile.py` file to use `to_edge_transform_and_lower` for simpicity of the example. The `executorch_pipeline.py` still requires converting to edge and than to backend separately for testing purposes.

examples/nxp/aot_neutron_compile.py
```diff
@@ -141,14 +145,6 @@ def _get_batch_size(data):
         required=True,
         help=f"Provide model name. Valid ones: {set(models.keys())}",
     )
-    parser.add_argument(
```
roman-janik-nxp, line 149: Why we are removing this argument?

roman-janik-nxp (changes requested):

examples/nxp/aot_neutron_compile.py
StrycekSimon: Since we replaced `to_edge` with `to_edge_transform_and_lower`, the delegation is no longer optional.

robert-kalmar: It is. If the partitioner_list is empty, no delegation happens - https://github.com/pytorch/executorch/pull/14142/files#diff-757cb2218dabf56479448db90ab097f2b4a15eed5b7f3840f50b52bd8348bdc5L260

StrycekSimon: Ok, my bad...

StrycekSimon:
Sorry everyone for the tag, I somehow managed to mess up a rebase.

StrycekSimon:
The fail with Samsung's tests is unrelated and appears in other PRs. The second red signal is some problem in the runner setup, and I don't see any relation to changes made in this PR.

robert-kalmar (approved):

robert-kalmar:
@roman-janik-nxp please re-review if you agree with the implementation of the requested change.

roman-janik-nxp (approved):

PR #14143, opened 2025-09-10
Arm backend: Add support for ETDump of outputs

zingo:
This also pass the generated output to the inspector if run.sh/runFVP.sh is used together with the generated ETRecord file.

This also try to sort the order of the c++/cmake flags to keep them in similar order in different files.

### Test plan
This is runned by the run.sh testing.


zingo:
@digantdesai merge this for GA or should we wait, it's nice to get it as it clean up some comments in the runner but as the same time it can perfectly wait until after GA if we suspect it to be bad.

digantdesai (approved):

PR #14148, opened 2025-09-10
Arm backend: Add pass order validation to ArmPassManager

martinlsm:
Introduce a mechanism to enforce required ordering of passes in ArmPassManager. Each ArmPass must now declare which passes are required to run after it, ensuring ordering constraints are always upheld.

This prevents accidental breakage when modifying pass ordering in the manager.

Ordering constraints are verified by the new method ArmPass.validate_constraints_mandatory. We considered reusing torch.fx.passes.infra.pass_manager.PassManager.validate_constraints, but that utility only checks pairwise ordering and cannot enforce that a pass is actually run, which did not meet our needs.

This patch only implements the mechanism and tests for it. Defining the actual pass orderings are done in a later patch.

### Test plan
The change comes with added unit tests in backends/arm/test/misc/test_pass_required_order.py

### DIFF
--- a/backends/arm/_passes/arm_pass.py
+++ b/backends/arm/_passes/arm_pass.py
@@ -6,7 +6,8 @@
 # pyre-unsafe
 
 import traceback
-from typing import Optional
+from abc import abstractmethod
+from typing import List, Optional, Set, Type
 
 import torch
 from executorch.exir.pass_base import ExportPass, NodeMetadata
@@ -18,6 +19,36 @@ class ArmPass(ExportPass):
     def __init__(self, exported_program: Optional[torch.export.ExportedProgram] = None):
         super(ArmPass, self).__init__()
         self.exported_program = exported_program
+
+    @property
+    @abstractmethod
+    def _passes_required_after(self) -> Set[Type[ExportPass]]:
+        """The subclass defines passes that must run after it"""
+        pass
+
+    @staticmethod
+    def get_required_passes(pass_) -> List[str]:
+        """
+        Returns the list of passes that must be run after this pass, sorted by name.
+        """
+        if hasattr(pass_, "_passes_required_after"):
+            return sorted([ArmPass.get_name(p) for p in pass_._passes_required_after])
+        else:
+            return []
+
+    @staticmethod
+    def get_name(pass_) -> str:
+        """
+        Returns the name of the pass.
+        """
+        if isinstance(pass_, ExportPass):
+            return pass_.__class__.__name__
+        elif hasattr(pass_, "__name__"):
+            return pass_.__name__
+        else:
+            raise ValueError(
+                f"Cannot get name for pass: {pass_}. It must be an instance of ExportPass or have a __name__ attribute."
+            )
 
     def call_operator(self, op, args, kwargs, meta, updated: Optional[bool] = False):
         if not updated:
--- a/backends/arm/_passes/add_bias_pass.py
+++ b/backends/arm/_passes/add_bias_pass.py
@@ -3,13 +3,15 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from typing import Set, Type
+
 import torch
 from executorch.backends.arm._passes import ArmPass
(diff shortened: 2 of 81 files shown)


backends/arm/_passes/arm_pass.py
```diff
@@ -19,6 +20,36 @@ def __init__(self, exported_program: Optional[torch.export.ExportedProgram] = No
...
+        """The subclass defines passes that must run after it"""
+        pass
+
+    @staticmethod
+    def get_required_passes(pass_) -> List[str]:
+        """
+        Returns the list of passes that must be run after this pass, sorted by name.
+        """
+        if hasattr(pass_, "_passes_required_after"):
+            return sorted([ArmPass.get_name(p) for p in pass_._passes_required_after])
```
digantdesai, line 35: One issue I can see with this approach is transitive dependency and running into weird errors.

At a high level, do we see many cases where we run same pass multiple times in certain order for a given backend, or do something different, in terms of pass ordering, based on some flag or a model architecture?

martinlsm: I'm not fully understanding the case of transitive dependencies. With this implementation, if we got the following scenario:
```
PassA:
    _passes_required_after = {PassB}

PassB:
    _passes_required_after = {PassC}

PassC:
    _passes_required_after = set()
```

Then `PassA` will transitively depend on `PassC`.

Then if create a pipeline where we remove `PassB` like this:
```
[
    PassA,
    PassC,
]
```
Then we will get the following error:
```
The following constraints for passes are not met:
    PassB must run after PassA
```

Is that not an reasonable way to handle it? I cannot think of a case where a missing transitive dependency will slip through here and not get caught by `validate_constraints_mandatory`.

Then regarding your second point: We have a long-term plan to change the passes such that they will always be added in the same order, but let themselves check whether they should run or not. This means that the order will always remain the same no matter which state/hardware profile we are in. Does that answer your question or maybe I'm not fully following?

digantdesai: > Is that not an reasonable way to handle it?

Yeah I guess. Thanks. I was also thinking like,
```
PassA --(needs)--> Pass B
PassB --(needs)--> Pass C
PassC --(needs)--> Pass A
```

More broadly,

> let themselves check whether they should run or not

I am failing to see the gain from this added complexity over a `list[Pass]` for a given backend. Mainly because this is not a user facing surface.

digantdesai: If you are saying we don't have to maintain "per backend + config list of passe" perhaps that is a good enough reason to embrace this complexity.. But I doubt if we can get rid of backend "specialized" logic from the PassManager but happy to be proven wrong :)

martinlsm: Thanks @digantdesai for the thorough review and feedback!

Regarding the cycles, yes, this implementation would not be able to handle that. But we also want to avoid those kind of loops, so having circular dependencies such as that is something we want to avoid. I cannot say it's 100% future proof but it's unlikely that we would ever need cyclic dependencies I'm thinking.

Regarding the added complexity, perhaps it adds a bit yes; the idea is to uphold intended ordering of the passes in cases where a developer wants to change the ordering. It's not obvious what should come before/after what when there's so many of them. One could say that the unit tests should protect against improper reordering, but this feature gives us quicker and clearer feedback in those cases.

I see your argument about special cases might work against this design. In the worst case we would have to revisit this feature and try to extend/change it somehow. But we do hope to eliminate special cases and always list the passes in the same order at least.

digantdesai: `_passes_required_after` being empty for almost all existing passes implies that a simple (ordered) list is sufficient :)
I don't want to block you here, and since this is inside Arm dir, I will let you make the call. Thanks for the discussion.

digantdesai (approved):
stamping to unblock from merging if/when ready. Thanks @martinlsm

backends/arm/_passes/arm_pass.py
martinlsm: Thanks! I see your point, but the actual dependencies are added in a later patch that is on its way.

zingo:
Test fails seems unrelated

PR #14158, opened 2025-09-10
Module support for multiple ptd files

lucylq:
* __->__ #14158

Support multiple PTD files in Module. This change updates the following private variables in Module:

```
std::string data_path --> std::unordered_set<std::string> data_files_
std::unique_ptr<DataLoader> data_map_loader --> std::vectror<std::unique_ptr<DataLoader>> data_map_loaders_
std::unique_ptr<NamedDataMap> data_map --> std::vector<std::unique_ptr<NamedDataMap> named_data_maps_
```
And introduces a new private variable. When we have multiple NamedDataMaps, they need to be merged into one, for use in method, etc. This is not implemented yet.
```
std::unique_ptr<NamedDataMap> merged_data_map_
```

The process of using a PTD file is:
```
std::string file --> wrapped in DataLoader --> wrapped in NamedDataMap.
```
At each stage we can have multiple.

This diff also introduces a new Module constructor that takes in `std::unordered_set<std::string> named_data_map_paths_`

Differential Revision: [D82059808](https://our.internmc.facebook.com/intern/diff/D82059808/)


kimishpatel (approved):

PR #14161, opened 2025-09-10
Deprecate debug buffer in Module

lucylq:
* __->__ #14161

It's in release 0.7 so can't remove it, following deprecation schedule.

debug_buffer is part of the event tracer, which is passed into Module. There isn't a public API on event tracer to set the debug_buffer, so the one held by Module is always empty.

Added in D71135352 for pybindings migration, however the debug buffer is held by PyModule in that case.

Differential Revision: [D82125911](https://our.internmc.facebook.com/intern/diff/D82125911/)


cccclai (approved):
Thanks!

PR #14169, opened 2025-09-10
Fix XNNPACK handling of negative permute dims

GregoryComer:
### Summary
Fix handling of negative dims in permute ops on XNNPACK. They currently fail to lower. This bug was surfaced by the backend operator test suite.

### Test plan
I've added a test to cover negative dims in permute.


metascroy (approved):

backends/xnnpack/operators/op_permute.py
```diff
@@ -44,14 +44,21 @@ def define_node(
...
         # output
         output_id = vals_to_ids[node]
 
         # permutation
+        input_rank = input_node.meta["val"].dim()
         permute_order = cast(List[int], node.args[1])
 
+        # Handle negative dimensions by converting them to positive indices
+        permute_order = [
+            (dim + input_rank) if dim < 0 else dim for dim in permute_order
```
swolchok, line 59: is it possible that dim + input_rank is still negative?

GregoryComer: Good question. I just tried in eager mode, and it seems to enforce that dims are in the range of [-rank, rank-1] inclusive.

```python
>>> temp = torch.randn(1,2,3,4,5)
>>> temp.permute(-9, 2, 3, 4, 0)
  File "<stdin>", line 1, in <module>
IndexError: Dimension out of range (expected to be in range of [-5, 4], but got -9)
```

PR #14171, opened 2025-09-10
Bump torchao pin and use v2 torchao tensors

metascroy:
This PR bumps the torchao pin in ExecuTorch, and adjusts the code in ExecuTorch to rely less on deprecated features.  In particular,

* torchao/experimental folder is being deprecated, so we switch embedding / tied embedding quantizers to their new home
* v1 tensors based on AffineQuantizedTensor + QDQLayout are being deprecated.  This switches ExecuTorch to use v2 tensors.  See https://github.com/pytorch/ao/issues/2967.

### DIFF
--- a/backends/apple/coreml/test/test_coreml_recipes.py
+++ b/backends/apple/coreml/test/test_coreml_recipes.py
@@ -3,6 +3,7 @@
 # Please refer to the license found in the LICENSE file in the root directory of the source tree.
 
 
+import copy
 import unittest
 
 import coremltools as ct
@@ -152,8 +153,9 @@ class TestCoreMLRecipes(unittest.TestCase):
         # Test with different group sizes
         for group_size in [8, 16, 32]:
             with self.subTest(group_size=group_size):
+                model_to_export = copy.deepcopy(model)
                 session = export(
-                    model=model,
+                    model=model_to_export,
                     example_inputs=example_inputs,
                     export_recipe=ExportRecipe.get_recipe(
                         CoreMLRecipeType.TORCHAO_INT4_WEIGHT_ONLY_PER_GROUP,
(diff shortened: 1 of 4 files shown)


jerryzh168 (approved):

backends/apple/coreml/test/test_coreml_recipes.py
```diff
@@ -152,8 +153,9 @@ def forward(self, x):
         # Test with different group sizes
         for group_size in [8, 16, 32]:
             with self.subTest(group_size=group_size):
+                model_to_export = copy.deepcopy(model)
```
metascroy, line 156: FYI @abhinaykukkadapu

It looks like previously we were repeated quantizing the model on each iteration of the for loop.  Not sure if you had to set the tolerance really low because of them

abhinaykukkadapu: I think you are right, there is a bug in the Source Quantize stage of et.export. Will put a PR for it, i think the test still uses 1e-3 and sqnr=20, do you expect it to be better than this with the fix?

metascroy: I didn't look too carefully, but I caught it because the the same tensor can no longer be quantized twice after my PR lands, and that hit an error in your test.

abhinaykukkadapu: Fix in this pr: #14186, i had to reduce the SQNR to 15. Also you don't have to copy it here after my pr lands, it is copied before the source transform stage.

metascroy: Wait, after the fix you're reducing the SQNR?

Does that make sense?

abhinaykukkadapu: Continuing discussion in #14186

abhinaykukkadapu, line 156: @metascroy can you try without a copy here, as the PR i landed should take care of this within the source transform stage.

PR #14176, opened 2025-09-10
Re-enable eval script

rohansjoshi:
Summary:
Revised diff to enable eval_llama_qnn script which does evaluations with prefill model

Will add unit test in a later diff

Differential Revision: D82143474


cccclai (approved):
Thanks! Please add a unit test to guard this flow

PR #14180, opened 2025-09-10
[Windows] Re-enable tokenizers

GregoryComer:
### Summary
After https://github.com/meta-pytorch/tokenizers/pull/121 and https://github.com/pytorch/executorch/pull/13888, we should be able to enable tokenizers on Windows.

### Test plan
CI. Also verified that install functions on a local Windows machine.

cc @larryliu0820


mergennachin (approved):

GregoryComer:
I'm seeing some path length issues when building tokenizers in-place. I'm going to see if I can set up a symlink in the windows setup CI script.

GregoryComer:
Rebasing to test with https://github.com/pytorch/executorch/pull/13923 to be safe, but will land today if everything looks good.

GregoryComer:
Removing 1.0 milestone from this PR in favor of using the pip package, when available. I'd need to make some more aggressive infra changes to the wheel build to make this work when building from source.

PR #14185, opened 2025-09-10
[Backend Tester] Skip tests with undelegated conv3d ops

GregoryComer:
We are missing a portable kernel implementation for Conv3d, so any tests that contain one will fail. Since conv1d, 2d, and 3d all share the same op, we need some logic to handle this constraint and skip the test. This fixes a small number of (incorrect) test failures on the backend suite when decomposed conv3d patterns are partially delegated.


GregoryComer:
* #14200
* #14190
* __->__ #14185

digantdesai (approved):

PR #14186, opened 2025-09-10
Fix bug where source transform passes untransformed model to next stage

abhinaykukkadapu:
Summary: `quantize_()` modifies model in place, we need to make a copy to avoid making changes to user passed model. Fix a bug as discussed in https://github.com/pytorch/executorch/pull/14171#discussion_r2338013170

Differential Revision: D82167495

### DIFF
--- a/backends/apple/coreml/test/test_coreml_recipes.py
+++ b/backends/apple/coreml/test/test_coreml_recipes.py
@@ -166,7 +166,7 @@ class TestCoreMLRecipes(unittest.TestCase):
                     session, example_inputs, atol=1e-3
                 )
                 self._compare_eager_unquantized_model_outputs(
-                    session, model, example_inputs
+                    session, model, example_inputs, sqnr_threshold=15
                 )
 
     def test_int4_weight_only_per_group_validation(self):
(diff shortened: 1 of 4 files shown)


backends/apple/coreml/test/test_coreml_recipes.py
```diff
@@ -166,7 +166,7 @@ def forward(self, x):
                     session, example_inputs, atol=1e-3
                 )
                 self._compare_eager_unquantized_model_outputs(
-                    session, model, example_inputs
+                    session, model, example_inputs, sqnr_threshold=15
```
metascroy, line 169: What was the existing value?

metascroy:
LGTM

metascroy (approved):

backends/apple/coreml/test/test_coreml_recipes.py
abhinaykukkadapu: It was 20, it was passing because the model was modified in place, and we were comparing with the same modified one.

metascroy: And it no longer passes with 20?

What is this comparing?  The quantized PTE to the unquantized eager model?

abhinaykukkadapu: Correct, Quantized (including lowering) vs Eager model, it was passing previously because the model was modified in place, so effectively we were comparing that to itself.

abhinaykukkadapu: Just checked what would be the error between eager vs just quantized (without lowering) and it is still the same, which makes sense because it is toy model.

PR #14187, opened 2025-09-11
[ET-VK] Optimize conv2d s1p0

alexdean08:
This change improves the execution of the pointwise conv2d s1p0 shader. It does through more of a GEMM-like implementation and employing more explicit loop unrolling.


backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
```diff
@@ -50,119 +47,89 @@ ${layout_declare_spec_const(C, "int", "ngroups", "1")}
...
-      //  for each output position.
-      for (int j = 0; j < 4; ++j) {
-        sum[i * 4 + j] = tex_values[0] * kernel_values[0 + j] + sum[i * 4 + j];
-        sum[i * 4 + j] = tex_values[1] * kernel_values[4 + j] + sum[i * 4 + j];
-        sum[i * 4 + j] = tex_values[2] * kernel_values[8 + j] + sum[i * 4 + j];
-        sum[i * 4 + j] = tex_values[3] * kernel_values[12 + j] + sum[i * 4 + j];
-      }
-    }
-  }
+  if (threadHW < inputAndOutputWidth * inputAndOutputHeight && gid1 < outputChannel) {
```
SS-JIA: nit: can we modify this to be in the style of

```
if (threadHW >= inputAndOutputWidth * inputAndOutputHeight && gid1 >= outputChannel) {
  return;
}
```

(i.e. return early if out of bounds pattern) instead? This is more in line with our other shaders.

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
```diff
@@ -50,119 +47,89 @@ ${layout_declare_spec_const(C, "int", "ngroups", "1")}
...
-  float sum[TILE_SIZE_X * TILE_SIZE_Y * 4];
+  // Divided by 4 because the input channels are packed
+  int inputChannel = in_group_size/4;
 
-  // Initialize the output array with the bias value
-  for (int i = 0; i < TILE_SIZE_X * TILE_SIZE_Y * 4; i++) {
-    sum[i] = 0;
-  }
+  int threadHW = int(gl_GlobalInvocationID.x);
+  int gid1 = int(gl_GlobalInvocationID.y);
```
SS-JIA: nit: would suggest renaming to something more descriptive, like `threadOutChannel`

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
```diff
@@ -50,119 +47,89 @@ ${layout_declare_spec_const(C, "int", "ngroups", "1")}
...
+      weight3OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 2, gid1), 0);
+      weight4OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 3, gid1), 0);
+
+      outputTexel[0] += dot(inputVec, vec4(weight1OutputChannelPacked[0], weight2OutputChannelPacked[0], weight3OutputChannelPacked[0], weight4OutputChannelPacked[0]));
+      outputTexel[1] += dot(inputVec, vec4(weight1OutputChannelPacked[1], weight2OutputChannelPacked[1], weight3OutputChannelPacked[1], weight4OutputChannelPacked[1]));
+      outputTexel[2] += dot(inputVec, vec4(weight1OutputChannelPacked[2], weight2OutputChannelPacked[2], weight3OutputChannelPacked[2], weight4OutputChannelPacked[2]));
+      outputTexel[3] += dot(inputVec, vec4(weight1OutputChannelPacked[3], weight2OutputChannelPacked[3], weight3OutputChannelPacked[3], weight4OutputChannelPacked[3]));
     }
+
+    imageStore(t_out, ivec3(xIdx, yIdx, gid1), op(outputTexel, out_min, out_max));
```
SS-JIA: iiuc, previously the shader calculated 4 output texels but the new one only calculates one. Have you experimented with computing a bigger output tile? Might be something that can get us a further boost. Note that I am also ok with landing the shader in its current form though, given the perf improvement.

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
```diff
@@ -50,119 +47,89 @@ ${layout_declare_spec_const(C, "int", "ngroups", "1")}
...
+
+      inputVec = texelFetch(t_in, ivec3(xIdx, yIdx, inputC), 0);
+
+      weight1OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 0, gid1), 0);
+      weight2OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 1, gid1), 0);
+      weight3OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 2, gid1), 0);
+      weight4OutputChannelPacked = texelFetch(t_kernel, ivec2(inputC * 4 + 3, gid1), 0);
 
-  const vec4 bias = texelFetch(t_bias, ivec2(out_pos_z, 0), 0);
+      outputTexel[0] += dot(inputVec, vec4(weight1OutputChannelPacked[0], weight2OutputChannelPacked[0], weight3OutputChannelPacked[0], weight4OutputChannelPacked[0]));
```
SS-JIA: In my experience, computing matmul-like operations with `dot` was not as fast as `fma` . Did you try computing with `fma` as well? Curious to know if you have had a different experience. Btw you can take a look at the infographic in the old shader's comments to see how fma can be used instead of dot.

SS-JIA (changes requested):
Overall LGTM. Just some questions about things tried and some small requests for stylistic changes. Thanks for working on this!

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
alexdean08: I tried a bigger output tile, and on my end it's slightly slower for MobileNet's case. I believe this is due to the excessive vector registers that are used when we increase the tile size.

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
alexdean08: Just tried it on my end. For me it's essentially the exact same for me to use ```fma``` instead of ```dot```. Might be some compiler magic happening that makes it be the same operations.

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
alexdean08: Done

backends/vulkan/runtime/graph/ops/glsl/conv2d_pw_s1p0.glsl
alexdean08: Done

andreanicastro (approved):

SS-JIA (approved):

PR #14190, opened 2025-09-11
[Backend Tester] Add ARM TOSA flow

GregoryComer:
Wire up the Arm tester with the backend test suites. This is an initial implementation, which is missing a few things. Most notably, output accuracy stats aren't available. I intend to take this as a follow-up.


GregoryComer:
* #14200
* __->__ #14190
* #14185

backends/test/suite/runner.py
```diff
@@ -191,8 +191,11 @@ def build_result(
     # Only run the runtime portion if something was delegated (or the flow doesn't delegate)
     if is_delegated or not flow.is_delegated:
         try:
-            tester.to_executorch().serialize()
-            extra_stats["pte_size_bytes"] = len(tester.get_artifact())
+            tester.to_executorch()
+
+            if flow.supports_serialize:
```
GregoryComer, line 196: ARM tester requires a compile spec giving a location to dump the artifact. I'm just skipping this for initial integration, but we can wire this up as a follow-up. I'm not sure how meaningful the binary size for the TOSA flow, as well. Is that something we should report?

digantdesai: yeah we can ignore binary size for TOSA-only flows

backends/test/suite/flows/arm.py
```diff
@@ -0,0 +1,24 @@
...
+
+    return ArmTester(
+        *args,
+        **kwargs,
+    )
+
+
+def _create_tosa_flow() -> TestFlow:
+    return TestFlow(
+        "arm_tosa",
```
digantdesai, line 17:
```suggestion
        "arm_tosa_fp",
```

digantdesai (approved):

digantdesai:
cc @zingo , @per  as FYI

zingo:
This will be nice to see and play with! 
It might be good to look into Ethos-U/VGF FLOWs down the line and this takes us quite long way to fullfill that, thanks!

backends/test/suite/flows/arm.py
GregoryComer: Updated.

PR #14191, opened 2025-09-11
Arm backend: Refactor compile spec handling (Try 2)

mergennachin:
Summary: Redoing https://github.com/pytorch/executorch/pull/14111 with additional fixes

Differential Revision: D82171193


digantdesai (approved):
Approved!!

backends/arm/TARGETS
```diff
@@ -64,11 +42,55 @@ runtime.python_library(
...
+    ],
+)
+
+runtime.python_library(
+    name = "tosa_compile_spec",
+    srcs = [
+        "tosa/__init__.py",
+        "tosa/compile_spec.py",
+    ],
+    deps = [
```
swolchok: I recommend moving this into tosa/TARGETS so you can reference these files

backends/arm/TARGETS
```diff
@@ -64,11 +42,55 @@ runtime.python_library(
...
+    ],
+)
+
+runtime.python_library(
+    name = "vgf",
+    srcs = [
+        "vgf/__init__.py",
+        "vgf/backend.py",
+        "vgf/compile_spec.py",
+        "vgf/partitioner.py",
```
swolchok, line 73: this is fine as long as vgf/TARGETS doesn't exist

PR #14192, opened 2025-09-11
Fix get_max_seq_len metadata method not found

jackzhxng:
### Summary
Fixes Llava not exporting with `get_max_seq_len` metadata method in the pte, which was happening because `__post_init__` only works with `@dataclass`.

### Testing
Exported a Llava pte and ran with the multimodal runner

### DIFF
--- a/extension/llm/export/builder.py
+++ b/extension/llm/export/builder.py
@@ -93,7 +93,7 @@ class LLMEdgeManager:
         calibration_data: Optional[str] = None,
         tokenizer_path: Optional[str] = None,
         verbose: bool = False,
-        metadata: dict = {},
+        metadata: Optional[dict] = None,
         dynamic_shapes: Optional[Any] = None,
         save_exported_program: bool = False,
         generate_etrecord: bool = False,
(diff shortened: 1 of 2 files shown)


extension/llm/export/builder.py
```diff
@@ -93,7 +93,7 @@ def __init__(
         calibration_data: Optional[str] = None,
         tokenizer_path: Optional[str] = None,
         verbose: bool = False,
-        metadata: Optional[dict] = None,
+        metadata: dict = {},
```
mergennachin: Keep it `metadata: Optional[dict] = None`

mergennachin: And in the constructor do

```
self.metadata = metadata if metadata is not None else {}
```

mergennachin:
```
def __init__(self, metadata: dict = {}):
```

is not great practice because {} is created once when the function is defined, not each time it's called. All instances of LLMEdgeManager the same dictionary object

mergennachin (approved):

extension/llm/export/builder.py
```diff
@@ -132,18 +133,20 @@ def __init__(
...
+                self.dynamic_shapes = (
+                    {1: torch.export.Dim("token_dim", max=self.max_seq_len - 1)},
+                )
+            else:
+                # Two input arguments: tokens and input_pos but input_pos is static shape
+                self.dynamic_shapes = (
+                    {1: torch.export.Dim("token_dim", max=self.max_seq_len)},
+                    {"input_pos": {0: 1}},
                 )
-            self.metadata["get_max_seq_len"] = token_dim.max
```
mergennachin, line 146: @jackzhxng removing this line has semantic difference? is it intentional?

jackzhxng: It's moved up to line 118, plus previously this line was never actually called because it was in `__post_init__` that wasn't ever running

PR #14209, opened 2025-09-11
Arm backend: make compile_spec API backwards compatible

Erik-Lundell:


backends/arm/TARGETS
Erik-Lundell, line 1: This is just a guess

zingo:
### NOTE: Only merge if @mergennachin  wants it #14191 will probably solve this and this might not be needed.

mergennachin:
Closing

PR #13946, opened 2025-09-04
Add GitHub Actions workflow for building CPU Aarch64 Linux wheels

chizkiyahu:
### Summary

Fixes #13044

Release notes: build

PR #14211, opened 2025-09-11
Remove torchcodec from install_requirements.py

mergennachin:
Summary:

torchcodec is used when only torchaudio users use torch.save and torch.load. It is an optional requirement.

Only moshi test uses torch.save and torch.load, and we shouldn't make this to install across the board.

Currently, it is blocking linux aarch64 wheel building (https://github.com/pytorch/executorch/pull/13946) because during smoke testing it is trying to install torchcodec on linux aarch64 machine and it is failing

### DIFF
--- a/examples/models/moshi/mimi/install_requirements.sh
+++ b/examples/models/moshi/mimi/install_requirements.sh
@@ -8,6 +8,7 @@
 set -x
 
 conda install "ffmpeg<8"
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):
This makes sense to me, I agree you probably don't want torchcodec to be a hard install-time dependency of executorch

digantdesai (approved):

PR #14230, opened 2025-09-11
Add option in memory planning to put shared state on same location across entry points

JacobSzwejbka:
Summary: API that lets you place the same state tensor on the same id and offset across entry points. Lets you have get and set state more natively in the runtime if the underlying arenas are the same.

Differential Revision: D82250153

### DIFF
--- a/exir/passes/memory_planning_pass.py
+++ b/exir/passes/memory_planning_pass.py
@@ -4,10 +4,12 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import itertools
 import logging
 import warnings
+from dataclasses import dataclass, field
 from functools import partial
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 
 import torch
 from executorch.exir._warnings import deprecated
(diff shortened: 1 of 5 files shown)


GregoryComer (approved):
I don't have full context on the memory planning logic, but this looks reasonable to me. This will be a very nice feature for encoder-decoder models. As an aside, do you think it will be feasible to default share_mutable_buffers to true in the future?

exir/passes/memory_planning_pass.py
```diff
@@ -36,6 +44,84 @@ def _callable_name(any_callable: Callable[..., Any]) -> str:
...
+                    return True, fqn
+    return False, None
+
+def _get_spec_from_node(node: fx.Node) -> TensorSpec:
+    specs = get_node_tensor_specs(node)
+    assert(len(specs) == 1)
+    return specs[0]
+
+def _insert_mutable_buffer_specs(state: "_MemoryPlanningState", gm: torch.fx.GraphModule, gs: ExportGraphSignature):
+    for node in gm.graph.nodes:
```
larryliu0820, line 86: Would be good to have some docstring here

exir/passes/memory_planning_pass.py
```diff
@@ -164,3 +265,31 @@ def run(
             # I dont know if that is a valid thing but if it is we should adjust verify_storage_reuse function
             verifier.verify_storage_reuse()
         return PassResult(graph_module, True)
+
+    def run_multimethod(self):
+        "Resolve any memory planning done across entry points"
```
larryliu0820, line 290: This should be a docstring right?
```suggestion
        """Resolve any memory planning done across entry points"""
```

exir/passes/memory_planning_pass.py
```diff
@@ -36,6 +44,84 @@ def _callable_name(any_callable: Callable[..., Any]) -> str:
...
+            state.mutable_buffers[fqn].add(spec)
+            continue
+        is_buffer, fqn = _is_buffer(node, gs)
+        # If it is not a mutable buffer it might just appear to be a buffer in this entry point. Think model.get_state()
+        # So cache it and later double check that this buffer never appears mutable
+        if is_buffer:
+            assert(fqn)
+            spec = _get_spec_from_node(node)
+            if getattr(spec, 'mem_id', None) is not None or getattr(spec, 'mem_offset', None) is not None:
+                raise ValueError("Cannot share mutable buffers if they already have a mem_id or mem_offset assigned")
```
larryliu0820: Should this be an exception or just a warning?

JacobSzwejbka: I think Id rather start with an exception and relax it later if we have a use case that needs it. If a mem_id or mem_offset is already present that means someone is running a custom memory plan before this. This already probably doesn't compose well in that scenario because we just place the buffers on mem_id=2 and assert everything else is on 1 or not planned.

PR #14238, opened 2025-09-11
Swap Llava export arg order

jackzhxng:
### Summary
Swaps Llava export arg order, such that forward takes `embeddings, cache_position`, instead of `cache_position, embeddings`.

### Test plan
Existing Llava tests

### DIFF
--- a/examples/models/llava/model.py
+++ b/examples/models/llava/model.py
@@ -405,5 +405,5 @@ class LlavaModel(EagerModelBase):
 
     def _get_prompt_dynamic_shapes(self):
         dim = torch.export.Dim("token_dim", min=2, max=self.max_seq_len)
-        text_model_dynamic_shapes = ({0: 1}, {1: dim})
+        text_model_dynamic_shapes = ({1: dim}, {0: 1})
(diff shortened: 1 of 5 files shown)


kirklandsign (approved):

examples/models/llava/model.py
```diff
@@ -405,5 +405,5 @@ def _get_image_dynamic_shapes(self):
 
     def _get_prompt_dynamic_shapes(self):
         dim = torch.export.Dim("token_dim", min=2, max=self.max_seq_len)
-        text_model_dynamic_shapes = ({0: 1}, {1: dim})
+        text_model_dynamic_shapes = ({1: dim}, {0: 1})
```
JacobSzwejbka, line 408: nit: 1 and 0 are the dims of a single tensor. So swapping doesnt do anything

JacobSzwejbka: also do you even have to specify that the batch is static 1?

jackzhxng: {0:1} is the dynamic shapes for the first arg, {1:dim} is for the second arg, so I swap them here

PR #14240, opened 2025-09-11
Improve lintrunner job runtime

swolchok:
lintrunner is required to land, but it takes 10-15 minutes. This seems to improve on that by:
- split out lintrunner-mypy job
- copy pytorch/pytorch's changed files tracking
- don't run `lintrunner init` unnecessarily

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -111,6 +111,7 @@ jobs:
           LINTRUNNER_FILES="--all-files"
         else
           LINTRUNNER_FILES="${CHANGED_FILES}"
+        fi
         if ! lintrunner --force-color ${LINTRUNNER_FILES} --skip MYPY --tee-json=lint.json 2> /dev/null; then
           echo ""
           echo -e "\e[1m\e[36mYou can reproduce these results locally by using \`lintrunner --skip MYPY\`. (If you don't get the same results, run \'lintrunner init\' to update your local linter)\e[0m"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -45,9 +45,6 @@ jobs:
           cp -r "${CACHE_DIRECTORY}" . || true
         fi
 
-        # This has already been cached in the docker image
-        lintrunner init
-
         RC=0
         # Run lintrunner on all files
         if ! lintrunner --force-color --all-files --take MYPY --tee-json=lint.json 2> /dev/null; then
@@ -93,9 +90,6 @@ jobs:
           # again if they do not exist
           cp -r "${CACHE_DIRECTORY}" . || true
         fi
-
-        # This has already been cached in the docker image
-        lintrunner init
 
         RC=0
         # Run lintrunner on all files
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lintrunner:
+  lintrunner-mypy:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
     permissions:
       id-token: write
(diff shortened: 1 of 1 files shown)


swolchok:
* __->__ #14240

.github/workflows/lint.yml
```diff
@@ -50,9 +50,53 @@ jobs:
 
         RC=0
         # Run lintrunner on all files
-        if ! lintrunner --force-color --all-files --tee-json=lint.json 2> /dev/null; then
+        if ! lintrunner --force-color --all-files --take MYPY --tee-json=lint.json 2> /dev/null; then
```
digantdesai, line 63: so we will run non-MYPY stuff twice?

swolchok: no, --take restricts the set of linters to run

.github/workflows/lint.yml
```diff
@@ -50,9 +50,53 @@ jobs:
...
+          # again if they do not exist
+          cp -r "${CACHE_DIRECTORY}" . || true
+        fi
+
+        # This has already been cached in the docker image
+        lintrunner init
+
+        RC=0
+        # Run lintrunner on all files
+        if ! lintrunner --force-color --all-files --skip MYPY --tee-json=lint.json 2> /dev/null; then
```
digantdesai: I was thinking what if instead of `--all-files` can we run this on "changed" files in this PR as a required, not sure if that helps with MYPY performance or not..

swolchok: the problem isn't MYPY performance, it's the time to build and install executorch. check out some recent lintrunner runs on HUD: https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=lintrunner

digantdesai (approved):
Stamping to unblock you from landing once ready

.github/workflows/lint.yml
swolchok: > instead of --all-files can we run this on "changed" files in this PR

good idea, PyTorch's lint.yml does something like this, but it's out of scope for this PR

swolchok:
ugh, this only knocks 3 minutes off the time for the lintrunner job. I think we're better off just not running `lintrunner init`, I'll try that. Also note that these jobs are mysteriously doing nothing for 5 minutes -- there's a huge gap in the log timestamps.

.github/workflows/lint.yml
swolchok: on second thought, this is a good idea and I'm open to just increasing scope of this PR so that we can be confident that we actually get a speed improvement.

swolchok:
@digantdesai can you please rereview? this has had several major changes

.github/workflows/_get-changed-files.yml
```diff
@@ -0,0 +1,43 @@
...
+        run: |
+          # Check if we're in a pull request context
+          if [ "${{ github.event_name }}" = "pull_request" ] || [ "${{ github.event_name }}" = "pull_request_target" ]; then
+            echo "Running in PR context"
+
+            # Get the PR number from the github context
+            PR_NUMBER="${{ github.event.number }}"
+
+            # Use gh CLI to get changed files in the PR with explicit repo
+            CHANGED_FILES=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER/files --paginate --jq '.[] | select(.status != "removed") | .filename' | tr '\n' ' ' | sed 's/ $//')
```
larryliu0820, line 30: Will —paginate give all the files at once or gives a page and ask you to go to next page?

larryliu0820 (approved):

.github/workflows/lint.yml
```diff
@@ -15,15 +15,27 @@ concurrency:
...
+      github.repository_owner == 'pytorch' && (
+        needs.get-changed-files.outputs.changed-files == '*' ||
+        contains(needs.get-changed-files.outputs.changed-files, '.py') ||
+        contains(needs.get-changed-files.outputs.changed-files, '.pyi')
+      )
     with:
       runner: linux.2xlarge
       docker-image: ci-image:executorch-ubuntu-22.04-linter
-      submodules: 'recursive'
+      submodules: true
```
digantdesai, line 38: is this an intentional change?

.github/workflows/_get-changed-files.yml
swolchok: beats me. this is an exact copy from pytorch/pytorch

.github/workflows/lint.yml
```diff
@@ -45,14 +57,64 @@ jobs:
           cp -r "${CACHE_DIRECTORY}" . || true
         fi
 
-        # This has already been cached in the docker image
-        lintrunner init
```
digantdesai, line 49: I guess this is the prime mover?

.github/workflows/lint.yml
swolchok: nope, fixed in a follow-up

.github/workflows/lint.yml
swolchok: surprisingly no, it was the set of optimizations together. https://github.com/pytorch/executorch/pull/14263 wasn't very impressive on its own

swolchok: I suspect changed files tracking might've been the biggest contributor. The jobs on main don't have changed files tracking and are still taking close to 10 minutes, though that's better than the 13 minutes they were taking before.

PR #14243, opened 2025-09-12
fix openvino

cccclai:
Try fixing the test


kimishpatel:
what does this fix?

cccclai:
> what does this fix?

The error message seems failing to find backend option, trying to set it properly

cccclai:
fix in https://github.com/pytorch/executorch/pull/14246

PR #14246, opened 2025-09-12
try fix

cccclai:
try fix openvino test

gcc 9.5 does not allow for constructing the struct with initializer list (https://godbolt.org/z/E51z63KWW) and somehow for openvino we trigger wheel building with gcc 9.5

### DIFF
--- a/runtime/backend/options.h
+++ b/runtime/backend/options.h
@@ -32,6 +32,12 @@ struct BackendOption {
   char key[kMaxOptionKeyLength]{};
   // value is the value of the backend option, like 4, true, etc
   OptionValue value;
+
+  BackendOption(const char* k, OptionValue v) {
+    strncpy(key, k, kMaxOptionKeyLength);
+    key[kMaxOptionKeyLength - 1] = '\0'; // ensure null-termination
+    value = v;
+  }
 };
 
 /**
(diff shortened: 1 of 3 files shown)


runtime/backend/options.h
```diff
@@ -32,6 +32,12 @@ struct BackendOption {
   char key[kMaxOptionKeyLength]{};
   // value is the value of the backend option, like 4, true, etc
   OptionValue value;
+
+  BackendOption(const char* k, OptionValue v) {
+    strncpy(key, k, kMaxOptionKeyLength);
+    key[kMaxOptionKeyLength - 1] = '\0'; // ensure null-termination
```
kimishpatel: this is not right. you need the length of the k. else you will have garbage values being compared

consider doing this
```
    BackendOption option;
    strncpy(option.key, QNN_RUNTIME_LOG_LEVEL, strlen(QNN_RUNTIME_LOG_LEVEL) - 1);
    option.key[strlen(QNN_RUNTIME_LOG_LEVEL)] = '\0';
    option.value = -1;
```

kimishpatel: do this in QnnBackendOptions.cpp

cccclai: updated and remove the lines in backend option

kimishpatel (approved):
I think this should work. Stamping for now. Highlight that gcc 9.5 does not allow for constructing the struct with initializer list

kimishpatel:
you need `identifier 'kMaxOptionKeyLength'; did you mean 'runtime::kMaxOptionKeyLength`. else it fails

cccclai:
> gcc 9.5

where did you find it's due to gcc 9.5

kimishpatel:
updated summary

PR #14249, opened 2025-09-12
fix hanging lint xref

cccclai:
@lint-ignore doesn't ignore the line properly and cause the script to hang. Here is the fix


cccclai:
not needed

PR #14225, opened 2025-09-11
Update cache position population and arg order for multimodal runner

kirklandsign:
### Summary
For voxtral and phi-3, we construct the cache_position_tensor like before; for llava, it will construct underneath so we pass in size 1.
### Test plan
CI

PR #14250, opened 2025-09-12
Run Llava with MultimodalRunner

rohansjoshi:
### Summary
Run Llava model with MultimodalRunner instead of LlavaRunner

### DIFF
--- a/.ci/scripts/test_llava.sh
+++ b/.ci/scripts/test_llava.sh
@@ -149,7 +149,7 @@ run_and_verify() {
 
     # verify result.txt
     RESULT=$(cat result.txt)
-    EXPECTED_PREFIX="ASSISTANT: image captures a basketball game in progress, with"
+    EXPECTED_PREFIX="ASSISTANT: The image captures a basketball game in progress, with"
 
     if [[ "${RESULT}" == *"${EXPECTED_PREFIX}"* ]]; then
         echo "Expected result prefix: ${EXPECTED_PREFIX}"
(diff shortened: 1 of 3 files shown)


examples/models/llava/export_llava.py
```diff
@@ -77,18 +77,20 @@ def __init__(self, llava):
             super().__init__()
             self.text_model = llava.text_model
 
-        def forward(self, input_pos, embeddings):
-            return self.text_model(None, {"input_pos": input_pos}, embeddings)
+        def forward(self, cache_positions, embeddings):
+            return self.text_model(None, {"input_pos": cache_positions[:1]}, embeddings)
```
larryliu0820: Why is this necessary?

rohansjoshi: E.g. say embeddings is length 5 and start pos is 2, MultimodalRunner text decoder passes input like ([2,3,4,5,6], embeddings), OTOH for LlavaRunner it would be ([2], embeddings). 
This makes the first argument need dynamic shape. Alternatively, the other route would be to modify MultimodalRunner to handle both cases

jackzhxng: We are modifying the multimodal runner to handle both cases

jackzhxng: https://github.com/pytorch/executorch/pull/14225

kirklandsign (approved):

mergennachin:
@kirklandsign @larryliu0820 cherry-pick?

Issue #13902, opened 2025-09-03
(CMSIS-NN Integration) Testing strategy


AdrianLundell:
Let's use this issue to align on the testing for the Cortex-M backend. I suggest the following tests, but I am happy to hear other opinions:

## Kernel tests
All kernels should have GTests similar to the quantization ops to separate testing kernel logic from the full executorch framework. These tests should be easily be built and run on an FVP through some helper script. This should simplify debugging of kernels significantly.

## Operator integration tests
We should write a cortex_m implementation of https://github.com/pytorch/executorch/blob/main/backends/test/harness/tester.py to get the kernel integration tests for free, stealing some code from the arm backend to run pte:s on FVP using semihosting.

## Python code tests
We should have some level of unittesting of the python code. We can reuse the tester for easily counting the expected number of ops before/ after passes. For testing the dialect ops we should write some helper functions for asserting that they produce expected shapes and results, if they do not exist already?

## Network tests 
In principle we can reuse the tester framework also for running full network tests instead of having to keep integrating cortex_m support into the run.sh script/ arm_aot_compiler. We should still have some proper full flow test which is also useful as an example for users, I would suggest something close to the https://github.com/pytorch/executorch/blob/main/examples/arm/ethos_u_minimal_example.ipynb but for cortex_m in that case.

Thoughts?

@digantdesai @psiddh @per @zingo

PR #14252, opened 2025-09-12
Summary: Add Stateful FC Cortex-m linearOps

psiddh:
Integrate with CMSIS-NN with per-channel quantization support

Test Plan:
With local changes :Run e2e test on FVP simulator
./examples/arm/run_mcu_models_fvp.sh --target=cortex-m55 --models=qlinear

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/backends/cortex_m/CMakeLists.txt
+++ b/backends/cortex_m/CMakeLists.txt
@@ -12,7 +12,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
-# Source root directory for executorch.
+# Source root directory for executorch
 if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
@@ -21,62 +21,82 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
 include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
 include(FetchContent)
 
-# CMSIS-NN version to download
+# CMSIS-NN configuration with dynamic path detection
 set(CMSIS_NN_VERSION
-    "v4.1.0"
+    "v7.0.0"
     CACHE STRING "CMSIS-NN version to download"
 )
-
-# Declare CMSIS-NN as a FetchContent project
-FetchContent_Declare(
-  cmsis_nn
-  GIT_REPOSITORY https://github.com/ARM-software/CMSIS-NN.git
-  GIT_TAG ${CMSIS_NN_VERSION}
+set(CMSIS_NN_LOCAL_PATH
+    ""
+    CACHE PATH "Path to existing local CMSIS-NN installation"
 )
 
-# Download and make CMSIS-NN available
-FetchContent_MakeAvailable(cmsis_nn)
+# Try to find existing / local CMSIS-NN installation. This is useful for
+# debugging and testing with local changes. This is not common, as the CMSIS-NN
+# library is downloaded via FetchContent in the default/regular case.
+if(CMSIS_NN_LOCAL_PATH AND EXISTS "${CMSIS_NN_LOCAL_PATH}")
+  message(STATUS "Using CMSIS-NN from specified path: ${CMSIS_NN_LOCAL_PATH}")
+  add_subdirectory(${CMSIS_NN_LOCAL_PATH} cmsis_nn_build)
+else()
+  # Use FetchContent with automatic fallback
+  message(STATUS "Using CMSIS-NN via FetchContent")
 
-# Print paths for debugging
-message(STATUS "CMSIS-NN source dir: ${cmsis_nn_SOURCE_DIR}")
-message(STATUS "CMSIS-NN binary dir: ${cmsis_nn_BINARY_DIR}")
+  FetchContent_Declare(
+    cmsis_nn
+    GIT_REPOSITORY https://github.com/ARM-software/CMSIS-NN.git
+    GIT_TAG ${CMSIS_NN_VERSION}
+    GIT_SHALLOW TRUE
+  )
+
+  FetchContent_GetProperties(cmsis_nn)
+  if(NOT cmsis_nn_POPULATED)
+    FetchContent_Populate(cmsis_nn)
+    add_subdirectory(${cmsis_nn_SOURCE_DIR} ${cmsis_nn_BINARY_DIR})
+  endif()
+endif()
+
+# Add MVEI define to cmsis-nn target
+if(TARGET cmsis-nn)
+  target_compile_definitions(cmsis-nn PUBLIC ARM_MATH_MVEI=1)
+  get_target_property(CMSIS_NN_INCLUDES cmsis-nn INTERFACE_INCLUDE_DIRECTORIES)
+  message(STATUS "CMSIS-NN include dirs: ${CMSIS_NN_INCLUDES}")
+else()
+  message(
+    FATAL_ERROR
+      "CMSIS-NN target not found. Check your CMSIS_NN_LOCAL_PATH or network connection."
+  )
+endif()
 
 # Cortex-M ops kernel sources
 set(_cortex_m_kernels__srcs
     ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantize_per_tensor.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_dequantize_per_tensor.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantized_add.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantized_linear.cpp
 )
 
-# Generate C++ bindings to register kernels into Executorch (for runtime)
+# Generate C++ bindings to register kernels into Executorch
 set(_yaml_file ${CMAKE_CURRENT_LIST_DIR}/ops/operators.yaml)
 gen_selected_ops(LIB_NAME "cortex_m_ops_lib" OPS_SCHEMA_YAML "${_yaml_file}")
-
 generate_bindings_for_kernels(
   LIB_NAME "cortex_m_ops_lib" CUSTOM_OPS_YAML "${_yaml_file}"
 )
-message("Generated files ${gen_command_sources}")
 
-# Build a library for cortex_m_kernels
+# Build library for cortex_m_kernels
 add_library(cortex_m_kernels ${_cortex_m_kernels__srcs})
-target_compile_options(cortex_m_kernels PUBLIC ${_common_compile_options})
+
+# Use PRIVATE for implementation dependencies to avoid INTERFACE pollution
+target_link_libraries(
+  cortex_m_kernels
+  PRIVATE cmsis-nn
+  PRIVATE executorch
+)
 
 # Include directories for cortex_m_kernels
 target_include_directories(
-  cortex_m_kernels
-  PRIVATE ${EXECUTORCH_ROOT}/..
-          ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10
-          ${cmsis_nn_SOURCE_DIR}/Include
+  cortex_m_kernels PRIVATE ${EXECUTORCH_ROOT}/..
+                           ${EXECUTORCH_ROOT}/runtime/core/portable_type/c10
 )
-
-# Link directly to the CMSIS-NN static library file
-target_link_libraries(
-  cortex_m_kernels PUBLIC ${cmsis_nn_BINARY_DIR}/libcmsis-nn.a executorch
-)
-
-# Add dependency to ensure CMSIS-NN builds before we try to link. Use the actual
-# CMSIS-NN target name (usually 'cmsis-nn')
-add_dependencies(cortex_m_kernels cmsis-nn)
 
 # cortex_m_ops_lib: Register Cortex-M ops kernels into Executorch runtime
 gen_operators_lib(
@@ -84,7 +104,7 @@ gen_operators_lib(
 )
 
 install(
-  TARGETS cortex_m_kernels cortex_m_ops_lib
+  TARGETS cortex_m_kernels cortex_m_ops_lib cmsis-nn
   EXPORT ExecuTorchTargets
   DESTINATION lib
   PUBLIC_HEADER DESTINATION include/executorch/backends/cortex_m/ops/
--- a/backends/cortex_m/ops/cmsis_scratch_buffer_context.h
+++ b/backends/cortex_m/ops/cmsis_scratch_buffer_context.h
@@ -1,0 +1,187 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+#pragma once
+
+#include "cortex_m_ops_common.h"
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+namespace cortex_m {
+namespace native {
+
+// During AOT phase, quantized_linear_fusion_pass allocates total buffer
+// and passes in as 'Tensor'. (Total buffer = 8-byte header + x bytes)
+// ┌─────────────────┬─────────────────────────────────────┐
+// │ KernelSum Header│        CMSIS Workspace              │
+// │    (8 bytes)    │         (x bytes)                   │
+// └─────────────────┴─────────────────────────────────────┘
+//          │                           │
+//          │                           └─> Passed to CMSIS API
+//          │
+//          └─> State for kernel sum
+
+// C++ Runtime:
+// ┌─────────────────┬─────────────────────────────────────┐
+// │ KernelSum Header│        CMSIS Workspace              │
+// │    (8 bytes)    │         (x bytes)                   │
+// └─────────────────┴─────────────────────────────────────┘
+// ^                 ^
+// │                 │
+// scratch_ptr       cmsis_workspace_ptr
+// │                 │
+// ▼                 ▼
+//             arm_vector_sum_s8() writes kernel sums (with bias if avail):
+//             [sum₀+bias₀][sum₁+bias₁][sum₂+bias₂]...[sum_{n-1}+bias_{n-1}]
+//             (n * 4-byte int32_t values = x bytes)
+//
+// - n = out_features (number of output features)
+// - x = n * 4 bytes (total CMSIS buffer size)
+// - Total buffer = 8 + x bytes
+
+class CMSISScratchBufferContext final {
+ public:
+  CMSISScratchBufferContext(
+      Tensor& scratch_buffer,
+      const Tensor& weights,
+      const Tensor& weight_zero_point,
+      const torch::executor::optional<Tensor>& bias)
+      : scratch_ptr_(scratch_buffer.mutable_data_ptr<int8_t>()),
+        total_size_(scratch_buffer.size(0)),
+        base_ptr_(reinterpret_cast<uint8_t*>(scratch_ptr_)),
+        in_features_(weights.size(1)),
+        out_features_(weights.size(0)),
+        is_per_channel_(weight_zero_point.numel() > 1),
+        weight_data_offset_(calculate_offset(weights.const_data_ptr<int8_t>())),
+        weight_zp_data_offset_(
+            calculate_offset(weight_zero_point.const_data_ptr<int32_t>())),
+        bias_data_offset_(
+            bias.has_value()
+                ? calculate_offset(bias.value().const_data_ptr<int32_t>())
+                : 0),
+        header_(reinterpret_cast<KernelSumHeader*>(scratch_ptr_)),
+        cmsis_workspace_ptr_(scratch_ptr_ + KERNEL_SUM_HEADER_SIZE) {
+    cmsis_nn_dims filter_dims = {in_features_, 1, 1, out_features_};
+    validate_size(filter_dims);
+  }
+
+  cmsis_nn_context get_cmsis_ctx() const {
+    cmsis_nn_context ctx;
+    ET_CHECK_MSG(
+        reinterpret_cast<uintptr_t>(cmsis_workspace_ptr_) % 4 == 0,
+        "CMSIS workspace not 4-byte aligned");
+    ctx.buf = cmsis_workspace_ptr_;
+    ctx.size = get_cmsis_workspace_size();
+    return ctx;
+  }
+
+  bool is_kernel_sum_updated() const {
+    return header_->updated;
+  }
+
+  void compute_kernel_sums_if_needed() {
+    if (!header_->updated) {
+      arm_vector_sum_s8(
+          reinterpret_cast<int32_t*>(cmsis_workspace_ptr_),
+          in_features_,
+          out_features_,
+          get_weight_data(),
+          get_weight_zp_data()[0],
+          0,
+          get_bias_data());
+      header_->updated = true;
+      ET_LOG(
+          Info,
+          "Computed kernel sums. [required_bytes : %d]",
+          header_->required_size);
+    }
+  }
+
+  const int8_t* get_weight_data() const {
+    return reinterpret_cast<const int8_t*>(base_ptr_ + weight_data_offset_);
+  }
+
+  const int32_t* get_weight_zp_data() const {
+    return reinterpret_cast<const int32_t*>(base_ptr_ + weight_zp_data_offset_);
+  }
+
+  const int32_t* get_bias_data() const {
+    return bias_data_offset_ == 0
+        ? nullptr
+        : reinterpret_cast<const int32_t*>(base_ptr_ + bias_data_offset_);
+  }
+
+  bool is_per_channel_quant() const {
+    return is_per_channel_;
+  }
+  int32_t get_in_features() const {
+    return in_features_;
+  }
+  int32_t get_out_features() const {
+    return out_features_;
+  }
+
+ private:
+  static constexpr size_t KERNEL_SUM_HEADER_SIZE = 8;
+
+  // Header for kernel sum computation state only
+  struct KernelSumHeader {
+    bool updated = false;
+    int32_t required_size = 0;
+  };
+  static_assert(
+      sizeof(KernelSumHeader) == KERNEL_SUM_HEADER_SIZE,
+      "KernelSumHeader must be exactly 8 bytes");
+
+  int8_t* scratch_ptr_;
+  size_t total_size_;
+  uint8_t* base_ptr_;
+
+  // Context members
+  const int32_t in_features_;
+  const int32_t out_features_;
+  const bool is_per_channel_;
+  const uint32_t weight_data_offset_;
+  const uint32_t weight_zp_data_offset_;
+  const uint32_t bias_data_offset_;
+
+  KernelSumHeader* header_;
+  int8_t* cmsis_workspace_ptr_;
+
+  uint32_t calculate_offset(const void* ptr) const {
+    if (ptr == nullptr)
+      return 0;
+
+    const uint8_t* ptr_bytes = reinterpret_cast<const uint8_t*>(ptr);
+    ET_CHECK_MSG(ptr_bytes >= base_ptr_, "Pointer is before base address");
+
+    const std::ptrdiff_t offset = ptr_bytes - base_ptr_;
+    ET_CHECK_MSG(
+        offset >= 0 && offset <= UINT32_MAX, "Offset out of valid range");
+    return static_cast<uint32_t>(offset);
+  }
+
+  size_t get_cmsis_workspace_size() const {
+    return total_size_ - KERNEL_SUM_HEADER_SIZE;
+  }
+
+  void validate_size(const cmsis_nn_dims& filter_dims) const {
+    header_->required_size =
+        arm_fully_connected_s8_get_buffer_size(&filter_dims);
+
+    ET_CHECK_MSG(
+        get_cmsis_workspace_size() >=
+            static_cast<size_t>(header_->required_size),
+        "Scratch buffer size %zu insufficient for required size %d",
+        get_cmsis_workspace_size(),
+        header_->required_size);
+  }
+};
+
+} // namespace native
+} // namespace cortex_m
--- a/backends/cortex_m/ops/cortex_m_ops_common.h
+++ b/backends/cortex_m/ops/cortex_m_ops_common.h
@@ -22,6 +22,10 @@ using ScalarType = executorch::aten::ScalarType;
 using Scalar = torch::executor::Scalar;
 using Error = executorch::runtime::Error;
 
+// From arm_nn_math_types.h
+#define ARM_NN_Q31_MAX ((int32_t)(0x7FFFFFFFL))
+#define ARM_NN_Q31_MIN ((int32_t)(0x80000000L))
+
 // Basic tensor type / layout validation and dimension order checking
 inline void validate_cmsis_nn_tensor_requirements(
     const Tensor& input1,
@@ -32,16 +36,19 @@ inline void validate_cmsis_nn_tensor_requirements(
   // Basic dtype validation
   ET_CHECK_MSG(
       input1.scalar_type() == expected_dtype,
-      "Input1 dtype must be %hhd",
-      expected_dtype);
+      "Input1 dtype must be %hhd, got %hhd",
+      expected_dtype,
+      input1.scalar_type());
   ET_CHECK_MSG(
       input2.scalar_type() == expected_dtype,
-      "Input2 dtype must be %hhd",
-      expected_dtype);
+      "Input2 dtype must be %hhd, got %hhd",
+      expected_dtype,
+      input2.scalar_type());
   ET_CHECK_MSG(
       output.scalar_type() == expected_dtype,
-      "Output dtype must be %hhd",
-      expected_dtype);
+      "Output dtype must be %hhd, got %hhd",
+      expected_dtype,
+      output.scalar_type());
 
   // Dim order consistency
   ET_CHECK_MSG(
@@ -112,6 +119,33 @@ inline void validate_quantization_params(
       output_multiplier,
       output_shift,
       "Single quant Output");
+}
+
+// Refer to CMSIS-NN 'arm_nn_requantize' implementation for details:
+// https://github.com/ARM-software/CMSIS-NN/blob/main/Include/arm_nnsupportfunctions.h#L1625
+// multiplier: Range {ARM_NN_Q31_MIN + 1, Q32_MAX}
+// shift     : Range {-31, 30}
+inline bool validate_per_channel_quant_params(
+    const int32_t* multipliers,
+    const int32_t* shifts,
+    int num_channels) {
+  for (int i = 0; i < num_channels; ++i) {
+    // Multiplier: {ARM_NN_Q31_MIN + 1, ARM_NN_Q31_MAX}
+    if (multipliers[i] <= ARM_NN_Q31_MIN || multipliers[i] > ARM_NN_Q31_MAX) {
+      ET_LOG(
+          Error,
+          "weight_multiplier[%d] out of CMSIS-NN range: %d",
+          i,
+          multipliers[i]);
+      return false;
+    }
+    // Shift: {-31, 30} for arm_nn_requantize
+    if (shifts[i] < -31 || shifts[i] > 30) {
+      ET_LOG(Error, "weight_shift[%d] out of range: %d", i, shifts[i]);
+      return false;
+    }
+  }
+  return true;
 }
 
 inline Error resize_to_broadcast_target_size(
--- a/backends/cortex_m/ops/op_quantized_linear.cpp
+++ b/backends/cortex_m/ops/op_quantized_linear.cpp
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "cmsis_scratch_buffer_context.h"
+#include "cortex_m_ops_common.h"
+
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+namespace cortex_m {
+namespace native {
+using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
+
+Tensor& quantized_linear_out(
+    KernelRuntimeContext& context,
+    const Tensor& input,
+    const Scalar& input_zero_point,
+    const Scalar& input_multiplier,
+    const Scalar& input_shift,
+    const Tensor& weights,
+    const Tensor& weight_zero_point,
+    const Tensor& weight_multiplier,
+    const Tensor& weight_shift,
+    const torch::executor::optional<Tensor>& bias,
+    const Tensor& bias_multiplier,
+    const Tensor& bias_shift,
+    const Tensor& scratch_buffer,
+    const Scalar& output_zero_point,
+    const Scalar& in_features,
+    const Scalar& out_features,
+    Tensor& out) {
+  ET_LOG(Info, "quantized_linear_out: called");
+  validate_cmsis_nn_tensor_requirements(input, weights, out);
+
+  ET_CHECK_MSG(
+      scratch_buffer.scalar_type() == ScalarType::Char,
+      "Scratch buffer must be int8");
+
+  const int32_t batch_size = input.size(0);
+  const int32_t in_feat = static_cast<int32_t>(in_features.to<int64_t>());
+  const int32_t out_feat = static_cast<int32_t>(out_features.to<int64_t>());
+  const int32_t input_zp = static_cast<int32_t>(input_zero_point.to<int64_t>());
+  const int32_t output_zp =
+      static_cast<int32_t>(output_zero_point.to<int64_t>());
+  const bool is_per_channel = (weight_zero_point.numel() > 1);
+
+  const int8_t* input_data = input.const_data_ptr<int8_t>();
+  const int8_t* weight_data = weights.const_data_ptr<int8_t>();
+  const int32_t* bias_data =
+      bias.has_value() ? bias.value().const_data_ptr<int32_t>() : nullptr;
+  int8_t* output_data = out.mutable_data_ptr<int8_t>();
+  const int32_t* weight_zp_data = weight_zero_point.const_data_ptr<int32_t>();
+  const int32_t* weight_mult_data = weight_multiplier.const_data_ptr<int32_t>();
+  const int32_t* weight_shift_data = weight_shift.const_data_ptr<int32_t>();
+
+  if (!validate_per_channel_quant_params(
+          weight_mult_data, weight_shift_data, out_feat)) {
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  // Initialize scratch buffer context (validates early)
+  CMSISScratchBufferContext scratch_ctx(
+      const_cast<Tensor&>(scratch_buffer), weights, weight_zero_point, bias);
+
+  scratch_ctx.compute_kernel_sums_if_needed();
+  cmsis_nn_context ctx = scratch_ctx.get_cmsis_ctx();
+
+  // Setup CMSIS-NN parameters
+  cmsis_nn_fc_params fc_params;
+  fc_params.input_offset = -input_zp;
+  fc_params.output_offset = output_zp;
+  fc_params.activation.min = std::numeric_limits<int8_t>::min();
+  fc_params.activation.max = std::numeric_limits<int8_t>::max();
+
+  cmsis_nn_dims input_dims = {1, 1, 1, in_feat};
+  cmsis_nn_dims filter_dims = {in_feat, 1, 1, out_feat};
+  cmsis_nn_dims bias_dims = {1, 1, 1, out_feat};
+  cmsis_nn_dims output_dims = {1, 1, 1, out_feat};
+
+  arm_cmsis_nn_status status;
+  for (int32_t b = 0; b < batch_size; b++) {
+    const int8_t* batch_input = input_data + b * in_feat;
+    int8_t* batch_output = output_data + b * out_feat;
+
+    ET_CHECK_MSG(
+        batch_input != nullptr && weight_data != nullptr,
+        "Null input pointers");
+    ET_CHECK_MSG(in_feat > 0 && out_feat > 0, "Invalid dimensions");
+
+    if (is_per_channel) {
+      cmsis_nn_per_channel_quant_params per_channel_quant_params;
+      per_channel_quant_params.multiplier =
+          const_cast<int32_t*>(weight_mult_data);
+      per_channel_quant_params.shift = const_cast<int32_t*>(weight_shift_data);
+
+      status = arm_fully_connected_per_channel_s8(
+          &ctx,
+          &fc_params,
+          &per_channel_quant_params,
+          &input_dims,
+          batch_input,
+          &filter_dims,
+          weight_data,
+          &bias_dims,
+          bias_data,
+          &output_dims,
+          batch_output);
+    } else {
+      fc_params.filter_offset = -weight_zp_data[0];
+      cmsis_nn_per_tensor_quant_params per_tensor_quant_params;
+      per_tensor_quant_params.multiplier = weight_mult_data[0];
+      per_tensor_quant_params.shift = weight_shift_data[0];
+
+      status = arm_fully_connected_s8(
+          &ctx,
+          &fc_params,
+          &per_tensor_quant_params,
+          &input_dims,
+          batch_input,
+          &filter_dims,
+          weight_data,
+          &bias_dims,
+          bias_data,
+          &output_dims,
+          batch_output);
+    }
+
+    if (status != ARM_CMSIS_NN_SUCCESS) {
+      ET_LOG(
+          Error,
+          "quantized_linear_out: CMSIS-NN failed with status [%d]",
+          status);
+      context.fail(Error::Internal);
+      return out;
+    }
+  }
+  return out;
+}
+
+// Functional variant (stub, not used at runtime)
+Tensor quantized_linear(
+    KernelRuntimeContext& context,
+    const Tensor& input,
+    const Scalar& input_zero_point,
+    const Scalar& input_multiplier,
+    const Scalar& input_shift,
+    const Tensor& weights,
+    const Tensor& weight_zero_point,
+    const Tensor& weight_multiplier,
+    const Tensor& weight_shift,
+    const torch::executor::optional<Tensor>& bias,
+    const Tensor& bias_multiplier,
+    const Tensor& bias_shift,
+    const Tensor& scratch_buffer,
+    const Scalar& output_zero_point,
+    const Scalar& in_features,
+    const Scalar& out_features) {
+  ET_LOG(Info, "quantized_linear: called");
+  assert(false);
+  return const_cast<Tensor&>(input);
+}
+
+} // namespace native
+} // namespace cortex_m
--- a/backends/cortex_m/ops/operators.py
+++ b/backends/cortex_m/ops/operators.py
@@ -223,3 +223,216 @@ def quantized_add_out_impl(
     out.copy_(result_quantized)
 
     return out
+
+
+# ===================================================================
+# QUANTIZED LINEAR OPERATION DEFINITION
+# ===================================================================
+
+
+def _check_per_tensor_or_per_channel(param: torch.Tensor, out_channels: int, name: str):
+    assert param.numel() in [
+        1,
+        out_channels,
+    ], f"{name} must be per-tensor (1) or per-channel ({out_channels}), got {param.numel()}"
+
+
+lib.define(
+    "quantized_linear.out("
+    "Tensor input, Scalar input_zero_point, Scalar input_multiplier, Scalar input_shift, "
+    "Tensor weights, "
+    "Tensor weight_zero_point, Tensor weight_multiplier, Tensor weight_shift, "
+    "Tensor? bias, Tensor bias_multiplier, Tensor bias_shift, "
+    "Tensor scratch_buffer, Scalar output_zero_point, Scalar in_features, Scalar out_features, "
+    "*, Tensor(a!) out) -> Tensor(a!)"
+)
+
+# Define functional variant (non-out version)
+lib.define(
+    "quantized_linear("
+    "Tensor input, Scalar input_zero_point, Scalar input_multiplier, Scalar input_shift, "
+    "Tensor weights, "
+    "Tensor weight_zero_point, Tensor weight_multiplier, Tensor weight_shift, "
+    "Tensor? bias, Tensor bias_multiplier, Tensor bias_shift, "
+    "Tensor scratch_buffer, Scalar output_zero_point, Scalar in_features, Scalar out_features"
+    ") -> Tensor"
+)
+
+
+# Fake meta function for shape inference (out variant)
+@register_fake("cortex_m::quantized_linear.out")
+def quantized_linear_out_meta(
+    input: torch.Tensor,
+    input_zero_point: int,
+    input_multiplier: int,
+    input_shift: int,
+    weights: torch.Tensor,
+    weight_zero_point: torch.Tensor,
+    weight_multiplier: torch.Tensor,
+    weight_shift: torch.Tensor,
+    bias: torch.Tensor,
+    bias_multiplier: torch.Tensor,
+    bias_shift: torch.Tensor,
+    scratch_buffer: torch.Tensor,
+    output_zero_point: int,
+    in_features: int,
+    out_features: int,
+    out: torch.Tensor,
+) -> torch.Tensor:
+    # Validate dimensions
+    batch_size = input.shape[0]
+    out_channels = weights.shape[0]
+
+    # Validate weight quantization parameters dimensions
+    _check_per_tensor_or_per_channel(
+        weight_zero_point, out_channels, "weight_zero_point"
+    )
+    _check_per_tensor_or_per_channel(
+        weight_multiplier, out_channels, "weight_multiplier"
+    )
+    _check_per_tensor_or_per_channel(weight_shift, out_channels, "weight_shift")
+
+    # Validate output shape
+    expected_shape = (batch_size, out_channels)
+    assert (
+        out.shape == expected_shape
+    ), f"Output shape {out.shape} must be {expected_shape}"
+
+    return out
+
+
+# Fake meta function for shape inference (functional variant)
+@register_fake("cortex_m::quantized_linear")
+def quantized_linear_meta(
+    input: torch.Tensor,
+    input_zero_point: int,
+    input_multiplier: int,
+    input_shift: int,
+    weights: torch.Tensor,
+    weight_zero_point: torch.Tensor,
+    weight_multiplier: torch.Tensor,
+    weight_shift: torch.Tensor,
+    bias: torch.Tensor,
+    bias_multiplier: torch.Tensor,
+    bias_shift: torch.Tensor,
+    scratch_buffer: torch.Tensor,
+    output_zero_point: int,
+    in_features: int,
+    out_features: int,
+) -> torch.Tensor:
+    # Validate dimensions (same as out variant)
+    batch_size = input.shape[0]
+    out_channels = weights.shape[0]
+
+    # Validate weight quantization parameters dimensions
+    _check_per_tensor_or_per_channel(
+        weight_zero_point, out_channels, "weight_zero_point"
+    )
+    _check_per_tensor_or_per_channel(
+        weight_multiplier, out_channels, "weight_multiplier"
+    )
+    _check_per_tensor_or_per_channel(weight_shift, out_channels, "weight_shift")
+
+    # Calculate output shape for functional variant
+    output_shape = (batch_size, out_channels)
+    return torch.empty(output_shape, dtype=input.dtype, device=input.device)
+
+
+@impl(lib, "quantized_linear.out", "CompositeExplicitAutograd")
+def quantized_linear_out_impl(
+    input: torch.Tensor,
+    input_zero_point: int,
+    input_multiplier: int,
+    input_shift: int,
+    weights: torch.Tensor,
+    weight_zero_point: torch.Tensor,
+    weight_multiplier: torch.Tensor,
+    weight_shift: torch.Tensor,
+    bias: torch.Tensor,
+    bias_multiplier: torch.Tensor,
+    bias_shift: torch.Tensor,
+    scratch_buffer: torch.Tensor,
+    output_zero_point: int,
+    in_features: int,
+    out_features: int,
+    *,
+    out: torch.Tensor,
+) -> torch.Tensor:
+    """
+    Fallback implementation for meta/testing
+    Note: This won't be called at runtime, only during compilation
+    """
+
+    # Per-channel dequantization
+    input_scale = input_multiplier * (2.0 ** (-input_shift))
+    input_fp = (input.float() - input_zero_point) * input_scale
+    if weight_zero_point.numel() == 1:
+        # Per-tensor
+        weight_scale = weight_multiplier.item() * (2.0 ** (-weight_shift.item()))
+        weights_fp = (weights.float() - weight_zero_point.item()) * weight_scale
+    else:
+        # Per-channel
+        weight_scales = weight_multiplier.float() * (2.0 ** (-weight_shift.float()))
+        weights_fp = (
+            weights.float() - weight_zero_point.float().unsqueeze(1)
+        ) * weight_scales.unsqueeze(1)
+    bias_fp = None
+    if bias is not None:
+        bias_scales = bias_multiplier.float() * (2.0 ** (-bias_shift.float()))
+        bias_fp = bias.float() * bias_scales
+
+        result_fp = torch.nn.functional.linear(input_fp, weights_fp, bias_fp)
+    else:
+        result_fp = torch.nn.functional.linear(input_fp, weights_fp)
+    result_quantized = torch.clamp(
+        torch.round(result_fp + output_zero_point), -128, 127
+    ).to(torch.int8)
+    out.copy_(result_quantized)
+    return out
+
+
+# Functional variant implementation
+@impl(lib, "quantized_linear", "CompositeExplicitAutograd")
+def quantized_linear_impl(
+    input: torch.Tensor,
+    input_zero_point: int,
+    input_multiplier: int,
+    input_shift: int,
+    weights: torch.Tensor,
+    weight_zero_point: torch.Tensor,
+    weight_multiplier: torch.Tensor,
+    weight_shift: torch.Tensor,
+    bias: torch.Tensor,
+    bias_multiplier: torch.Tensor,
+    bias_shift: torch.Tensor,
+    scratch_buffer: torch.Tensor,
+    output_zero_point: int,
+    in_features: int,
+    out_features: int,
+) -> torch.Tensor:
+    """
+    Functional variant - creates output tensor and calls out variant
+    """
+    # Create output tensor
+    batch_size = input.shape[0]
+    output = torch.empty(
+        (batch_size, out_features), dtype=torch.int8, device=input.device
+    )
+    return quantized_linear_out_impl(
+        input,
+        input_zero_point,
+        input_multiplier,
+        input_shift,
+        weights,
+        weight_zero_point,
+        weight_multiplier,
+        weight_shift,
+        bias,
+        bias_multiplier,
+        bias_shift,
+        scratch_buffer,
+        output_zero_point,
+        in_features,
+        out_features,
+        out=output,
+    )
--- a/backends/cortex_m/passes/passes_utils.py
+++ b/backends/cortex_m/passes/passes_utils.py
@@ -8,6 +8,10 @@ import math
 
 import torch
 
+from executorch.exir.dialects._ops import ops as exir_ops
+
+from torch.fx import Node
+
 
 def dequantize_per_tensor_cmsis(
     qtensor: torch.Tensor, zero_point: int, multiplier: int, shift: int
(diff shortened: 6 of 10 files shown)


backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -114,6 +117,26 @@ inline void validate_quantization_params(
...
+    const int32_t* shifts,
+    int num_channels) {
+  for (int i = 0; i < num_channels; ++i) {
+    if (multipliers[i] < (1LL << 30) || multipliers[i] > ((1LL << 31) - 1)) {
+      ET_LOG(
+          Error,
+          "weight_multiplier[%d] out of CMSIS-NN range: %d",
+          i,
+          multipliers[i]);
+      return;
```
digantdesai: return false, and check the return code to return Error to ET

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,218 @@ def quantized_add_out_impl(
...
+)
+
+
+# Fake meta function for shape inference (out variant)
+@register_fake("cortex_m::quantized_linear.out")
+def quantized_linear_out_meta(
+    input: torch.Tensor,
+    input_zero_point: int,
+    input_multiplier: int,
+    input_shift: int,
```
digantdesai, line 268: Do we want to just take Tensor in (even for a single element)? Rationale is to support per-token like quant if we want like per-tensor today.

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,218 @@ def quantized_add_out_impl(
...
+        out_channels,
+    ], f"weight_zero_point must be per-tensor (1) or per-channel ({out_channels}), got {weight_zero_point.numel()}"
+    assert weight_multiplier.numel() in [
+        1,
+        out_channels,
+    ], f"weight_multiplier must be per-tensor (1) or per-channel ({out_channels}), got {weight_multiplier.numel()}"
+    assert weight_shift.numel() in [
+        1,
+        out_channels,
+    ], f"weight_shift must be per-tensor (1) or per-channel ({out_channels}), got {weight_shift.numel()}"
```
digantdesai: deduplicate these checks

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,218 @@ def quantized_add_out_impl(
...
+    input_shift: int,
+    weights: torch.Tensor,
+    weight_zero_point: torch.Tensor,
+    weight_multiplier: torch.Tensor,
+    weight_shift: torch.Tensor,
+    bias: torch.Tensor,
+    bias_multiplier: torch.Tensor,
+    bias_shift: torch.Tensor,
+    scratch_buffer: torch.Tensor,
+    output_zero_point: int,
```
digantdesai, line 277: And this one too..

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,218 @@ def quantized_add_out_impl(
...
+    if bias is not None:
+        if bias_multiplier.numel() == 1:
+            bias_scale = bias_multiplier.item() * (2.0 ** (-bias_shift.item()))
+            bias_fp = bias.float() * bias_scale
+        else:
+            bias_scales = bias_multiplier.float() * (2.0 ** (-bias_shift.float()))
+            bias_fp = bias.float() * bias_scales
+        result_fp = torch.nn.functional.linear(input_fp, weights_fp, bias_fp)
+    else:
+        result_fp = torch.nn.functional.linear(input_fp, weights_fp)
```
digantdesai: refactor

```suggestion

    bias_fp = None
    if bias is not None:
        if bias_multiplier.numel() == 1:
            bias_scale = bias_multiplier.item() * (2.0 ** (-bias_shift.item()))
            bias_fp = bias.float() * bias_scale
        else:
            bias_scales = bias_multiplier.float() * (2.0 ** (-bias_shift.float()))
            bias_fp = bias.float() * bias_scales

     result_fp = torch.nn.functional.linear(input_fp, weights_fp, bias_fp)
```

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,218 @@ def quantized_add_out_impl(
...
+        weights_fp = (weights.float() - weight_zero_point.item()) * weight_scale
+    else:
+        # Per-channel
+        weight_scales = weight_multiplier.float() * (2.0 ** (-weight_shift.float()))
+        weights_fp = (
+            weights.float() - weight_zero_point.float().unsqueeze(1)
+        ) * weight_scales.unsqueeze(1)
+    if bias is not None:
+        if bias_multiplier.numel() == 1:
+            bias_scale = bias_multiplier.item() * (2.0 ** (-bias_shift.item()))
```
digantdesai: why do we need `.item()` specialization for numel == 1?

backends/cortex_m/passes/passes_utils.py
```diff
@@ -92,3 +97,77 @@ def quantize_multiplier_aot(scale: float) -> tuple[int, int]:
...
+                new_node.meta[field] = source_node.meta[field]
+
+
+def find_insertion_point(graph) -> Node:
+    """Find appropriate insertion point in graph."""
+    for node in graph.nodes:
+        if node.op == "placeholder":
+            if not any(node.name.startswith(prefix) for prefix in ["p_", "b_"]):
+                return node
+    return next(iter(graph.nodes))
```
digantdesai: This depends on what do you want to insert. I.e. if you are inserting another placeholder node (and of which type) or some other node. Typically you do this by finding the relevant node of interest and use insert_before/after like utils.

digantdesai: Also not used? remove?

backends/cortex_m/passes/passes_utils.py
```diff
@@ -92,3 +97,77 @@ def quantize_multiplier_aot(scale: float) -> tuple[int, int]:
...
+    """Trace through transformations to find dequantize node."""
+    current_node = node
+    depth = 0
+    while current_node and depth < max_depth:
+        if is_dequant_node(current_node):
+            return current_node
+        if current_node.op == "call_function" and current_node.target in {
+            exir_ops.edge.aten.permute_copy.default,
+            exir_ops.edge.aten.view_copy.default,
+        }:
```
digantdesai: this makes is somewhat specific to linear, do you want to move this back to the pass as opposed to be used as a general util?

backends/cortex_m/passes/passes_utils.py
```diff
@@ -92,3 +97,77 @@ def quantize_multiplier_aot(scale: float) -> tuple[int, int]:
...
+
+
+def cleanup_nodes(nodes_to_erase, graph):
+    """Clean up marked nodes from graph."""
+    for node in reversed(nodes_to_erase):
+        if node in graph.nodes and len(node.users) == 0:
+            try:
+                graph.erase_node(node)
+            except Exception as e:
+                print(f"Warning: Failed to erase node {node}: {e}")
```
digantdesai, line 146: Do we want to continue the loop after this?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+    SUPPORTED_OPS_MAPPING = {
+        exir_ops.edge.aten.linear.default: exir_ops.edge.cortex_m.quantized_linear.default,
+        exir_ops.edge.aten.addmm.default: exir_ops.edge.cortex_m.quantized_linear.default,
+        # Future FC-like ops can be added here
+    }
+
+    requires_exported_program = True
+
+    def __init__(self, exported_program: ExportedProgram):
+        logger.info("C'tor QuantizedLinearFusionPass")
```
digantdesai: seems redundant , remove?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        super().__init__()
+        self._exported_program = exported_program
+        self.nodes_to_erase = []
+
+    @property
+    def exported_program(self) -> ExportedProgram:
+        return self._exported_program
```
digantdesai: who uses this for this pass instance?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        logger.info("Starting QuantizedLinearFusionPass")
+        try:
+            fusion_count = self._fuse_quantized_linear_patterns(graph_module)
+            for node in graph_module.graph.nodes:
+                logger.debug(
+                    "Post-fusion Node: %s, op: %s, target: %s",
+                    node.name,
+                    node.op,
+                    getattr(node, "target", None),
+                )
```
digantdesai: remove this? We can just print the graph when debugging..

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+            if fusion_count > 0:
+                graph_module.graph.eliminate_dead_code()
+                graph_module.graph.lint()
+                graph_module.recompile()
+            logger.info(f"Linear fusion completed: {fusion_count} patterns fused")
+            return PassResult(graph_module, fusion_count > 0)
+        except Exception as e:
+            logger.error(f"Error in QuantizedLinearFusionPass: {e}")
+            raise
```
digantdesai: raise e

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+    """
+    Cortex-M backend pass that fuses quantized linear-like patterns.
+    Fuses: dequantize -> [linear/addmm/fc_ops] -> quantize
+    Into: cortex_m.quantized_linear.default with direct parameters.
+    """
+
+    # Extensible operation mapping for FC-like operations
+    SUPPORTED_OPS_MAPPING = {
+        exir_ops.edge.aten.linear.default: exir_ops.edge.cortex_m.quantized_linear.default,
+        exir_ops.edge.aten.addmm.default: exir_ops.edge.cortex_m.quantized_linear.default,
```
digantdesai, line 40: so ideally (once we add the Partitioner synthetic sugar for the op-lib-backends) we don't need to support addmm. We can just [ask](https://github.com/pytorch/executorch/blob/c4fc948261af4930744a1bafffca5c9fae735256/exir/program/_program.py#L928) export to not decompose via partitioner. Adding the synthetic sugar is on me, perhaps post GA.

The pattern matching to map addmm or mm to linear is fragile and can easily break with decomposition. Cadence has (not so clean) [way](https://github.com/pytorch/executorch/blob/c4fc948261af4930744a1bafffca5c9fae735256/backends/cadence/aot/compiler.py#L71) to stop the linear decomp. 

If we want to support it without `do-no-decompose-linear` for now, we should atleast split it out of this pass and use a standalone passs. [[1](https://github.com/pytorch/executorch/blob/main/backends/transforms/addmm_mm_to_linear.py)][[2](https://github.com/pytorch/executorch/blob/c4fc948261af4930744a1bafffca5c9fae735256/backends/xnnpack/_passes/convert_to_linear.py)]. And this pass just focuses on (quantized) aten.linear -> cortex_m.quantized_linear

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+            logger.error(f"Error in QuantizedLinearFusionPass: {e}")
+            raise
+
+    def _is_supported_fc_op(self, node: Node) -> bool:
+        """Check if node is a supported FC-like operation."""
+        return node.op == "call_function" and node.target in self.SUPPORTED_OPS_MAPPING
+
+    def _get_quantized_target(self, fc_op_target):
+        """Get the corresponding quantized target for a FC operation."""
+        return self.SUPPORTED_OPS_MAPPING.get(fc_op_target)
```
digantdesai: just inline this

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+                graph_module.recompile()
+            logger.info(f"Linear fusion completed: {fusion_count} patterns fused")
+            return PassResult(graph_module, fusion_count > 0)
+        except Exception as e:
+            logger.error(f"Error in QuantizedLinearFusionPass: {e}")
+            raise
+
+    def _is_supported_fc_op(self, node: Node) -> bool:
+        """Check if node is a supported FC-like operation."""
+        return node.op == "call_function" and node.target in self.SUPPORTED_OPS_MAPPING
```
digantdesai: inline this

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        return (
+            quantize_node,
+            fc_node,
+            input_dq_node,
+            weight_dq_node,
+            bias_dq_node,
+            op_name,
+        )
+
+    def _extract_weight_bias_from_fc_op(self, fc_node: Node):
```
digantdesai, line 101: this is fragile is what I was referring to.

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        elif any(op in str(fc_node.target) for op in ["linear", "mm", "bmm", "addmv"]):
+            if len(fc_node.args) >= 2:
+                weight_arg = fc_node.args[1]
+                bias_arg = fc_node.args[2] if len(fc_node.args) > 2 else None
+                weight_dq_node = trace_to_dequantize(weight_arg)
+                bias_dq_node = trace_to_dequantize(bias_arg) if bias_arg else None
+                return weight_dq_node, bias_dq_node
+        return None, None
+
+    def _extract_param_value(self, node_or_value):
```
digantdesai: Is this not available already somewhere else? if not move this to utils?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+                    input_shift,
+                    weight_tensor,
+                    weight_zp_node,
+                    weight_mult_node,
+                    weight_shift_node,
+                    bias_tensor,
+                    bias_multiplier,
+                    bias_shift,
+                    scratch_buffer,
+                    output_zp,
```
digantdesai, line 515: No output scale?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        # If it's a tensor, get its scalar value if possible
+        if isinstance(node_or_value, torch.Tensor):
+            return node_or_value.item() if node_or_value.numel() == 1 else node_or_value
+        # If it's a Node, use get_param_tensor
+        if hasattr(node_or_value, "op"):
+            tensor = get_param_tensor(self._exported_program, node_or_value)
+            return tensor.item() if tensor.numel() == 1 else tensor
+        raise TypeError(f"Unsupported parameter type: {type(node_or_value)}")
+
+    def _extract_quantization_parameters(
```
digantdesai: split output node? and just keep this for input?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+                "input_shift": input_shift,
+                "output_scale": output_scale,
+                "output_zero_point": output_zero_point,
+                "input_tensor": input_quantize_node,
+            }
+        except Exception as e:
+            logger.error(f"Failed to extract quantization parameters: {e}")
+            return None
+
+    def _create_parameter_buffer(
```
digantdesai:
```suggestion
    def _create_constant_parameter_buffer(
```

digantdesai: you can also try using `create_constant_placeholder` from utils.py but `get_attr` is also OK for constants like `placeholder`.

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        except Exception as e:
+            logger.error(f"Failed to extract bias parameters: {e}")
+            return None
+
+    def _create_scratch_buffer(
+        self, graph, quantize_node: Node, batch_size: int, out_features: int
+    ):
+        struct_size = 2060  # 2048 + sizeof(kernel_sum_state defined in kernel impl)
+        cmsis_scratch = 1024  # CMSIS scratch buffer size
+        total_size = struct_size + cmsis_scratch  # Should be ~3084
```
digantdesai: don't hardcode these values.. reverse engineer, implement, and call `cmsis_get_scratch_size(weights)`

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+
+        print(f"Allocating scratch buffer size: {total_size}")
+        # Creates mutable runtime buffer
+        scratch_buffer = create_mutable_buffer(
+            self._exported_program,
+            name=f"b_scratch_{id(quantize_node)}",
```
digantdesai:
```suggestion
            name=f"b_cmsis_linear_scratch_{id(quantize_node)}",
```

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+            graph, quantize_node, weight_params["weight_shift_data"], "weight_shift"
+        )
+        # Get dimensions
+        try:
+            weight_shape = get_param_tensor(self._exported_program, weight_tensor).shape
+            in_features = weight_shape[1]
+            out_features = weight_shape[0]
+        except Exception:
+            in_features = 1
+            out_features = 1
```
digantdesai: why use try..except here? just assert ndim == 2 and use the shape values?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+            # If already a tensor, leave as is
+            if isinstance(bias_shift, int):
+                bias_shift = torch.tensor([bias_shift], dtype=torch.int32)
+            elif isinstance(bias_shift, list):
+                bias_shift = torch.tensor(bias_shift, dtype=torch.int32)
+            # If already a tensor, leave as is
+        else:
+            # No bias: pass zero tensors of correct shape
+            bias_multiplier = torch.zeros([out_features], dtype=torch.int32)
+            bias_shift = torch.zeros([out_features], dtype=torch.int32)
```
digantdesai: move it in `_extract_bias_parameters`?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,529 @@
...
+        self.nodes_to_erase = []
+
+    @property
+    def exported_program(self) -> ExportedProgram:
+        return self._exported_program
+
+    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
+        logger.info("Starting QuantizedLinearFusionPass")
+        try:
+            fusion_count = self._fuse_quantized_linear_patterns(graph_module)
```
digantdesai, line 58: do we want to 
```
assert(id(self._exported_program.graph_module.graph) == id(graph_module.graph)), "this pass doesn't work otherwise?"

File: backends/cortex_m/CMakeLists.txt [hunk 27]
```diff
@@ -21,31 +21,59 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
...
-
-# Download and make CMSIS-NN available
-FetchContent_MakeAvailable(cmsis_nn)
-
-# Print paths for debugging
-message(STATUS "CMSIS-NN source dir: ${cmsis_nn_SOURCE_DIR}")
-message(STATUS "CMSIS-NN binary dir: ${cmsis_nn_BINARY_DIR}")
+set(USE_LOCAL_CMSIS_NN OFF)
+
+if(USE_LOCAL_CMSIS_NN)
```
digantdesai: why do we need this?

examples/arm/aot_arm_compiler.py
```diff
@@ -797,22 +819,33 @@ def to_edge_no_delegate(exported_program, args, model: torch.nn.Module, example_
...
+        # Check by class name or use specific attribute
+        if pass_cls.__name__ == "QuantizedLinearFusionPass":
+            transform_pass = pass_cls(current_edge.exported_program())
+        elif issubclass(pass_cls, ExportPass):
+            transform_pass = pass_cls()
+        else:
+            raise RuntimeError(
+                f"Expecting ExportPass, but got pass: {pass_cls} with type: {type(pass_cls)}"
+            )
+        current_edge = current_edge.transform([transform_pass])
```
digantdesai, line 831: if you want you can fold transform call inside if/else :)

examples/arm/aot_arm_compiler.py
```diff
@@ -297,6 +303,20 @@ def forward(self, x: torch.Tensor, y: torch.Tensor):
...
+        # Define a simple linear layer
+        self.linear = torch.nn.Linear(64, 32)
+
+    def forward(self, x):
+        return self.linear(x)
+
+    # Example input dimensions that will test your op
+    example_input = (torch.randn([8, 64], dtype=torch.float32),)
+    can_delegate = True
+
```
digantdesai: You need more variants like
```
combos_to_test = combinations(batch_size, input_features, output_features, has_bias)
```

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+    const Scalar& input_zero_point,
+    const Scalar& input_multiplier,
+    const Scalar& input_shift,
+    const Tensor& weights,
+    const Tensor& weight_zero_point,
+    const Tensor& weight_multiplier,
+    const Tensor& weight_shift,
+    const torch::executor::optional<Tensor>& bias,
+    const Tensor& bias_multiplier, // IGNORE - not used
+    const Tensor& bias_shift, // IGNORE - not used
```
digantdesai: just needed for python?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+  cmsis_nn_fc_params fc_params;
+  fc_params.input_offset = -input_zp;
+  fc_params.output_offset = output_zp;
+  fc_params.activation.min = std::numeric_limits<int8_t>::min();
+  fc_params.activation.max = std::numeric_limits<int8_t>::max();
+  cmsis_nn_dims input_dims = {1, 1, 1, in_feat};
+  cmsis_nn_dims filter_dims = {out_feat, 1, 1, in_feat};
+  cmsis_nn_dims bias_dims = {1, 1, 1, out_feat};
+  cmsis_nn_dims output_dims = {1, 1, 1, out_feat};
```
digantdesai: comment what and why are these 1s..

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "arm_nnfunctions.h"
+}
+
+namespace cortex_m {
+namespace native {
+using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
+
+struct kernel_sum_state {
+  bool updated = false;
+  char buffer[2048] = {0};
```
digantdesai: why hardcoded size here? In my example I was just using 1024 for illustration. We don't want any static allocation here..

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+  // Buffer pointers:
+  // ^                        ^                                 ^
+  // scratch_ptr(start)   scratch_ptr + cmsis_scratch   scratch_ptr + total_size
+  //
+  // - CMSIS-NN workspace: used by CMSIS-NN kernels for temporary data
+  // - Always give CMSIS-NN the start of the buffer for alignment
+  // - Place kernel_sum_state structs at the end to avoid breaking alignment
+  cmsis_nn_context ctx;
+  kernel_sum_state* state = reinterpret_cast<kernel_sum_state*>(
+      scratch_ptr + scratch_buffer.size(0) - sizeof(kernel_sum_state));
```
digantdesai: seems convoluted, the cleaner approach would be to hide all this in a constructor(&Tensor scratch_buffer, filter_state) and a `kernel_sum_state->get_scrtach_pointer()` method on it. Also you can put the `validate_size()` inside it which callas arm_fully_connected_s8_get_buffer_size(filter_dims)`

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+  if (!state->updated) {
+    int required_bytes = arm_fully_connected_s8_get_buffer_size(&filter_dims);
+    ET_CHECK_MSG(
+        (scratch_buffer.size(0) - sizeof(kernel_sum_state) >= required_bytes),
+        "Scratch buffer size %zu is not enough for kernel sum buffer size %d",
+        sizeof(state->buffer),
+        required_bytes);
+
+    // Compute kernel sums once
+    arm_vector_sum_s8(
```
digantdesai: Perfect!

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+        (int32_t*)scratch_ptr,
+        in_feat,
+        out_feat,
+        weight_data,
+        weight_zp_data[0],
+        0, // rhs_offset (int32_t)
+        nullptr // bias (const int32_t*)
```
digantdesai: even if we have bias?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+    state->updated = true;
+    ET_LOG(
+        Info,
+        "Computed kernel sums, stored in state->buffer [required_bytes : %d]",
+        required_bytes);
+  }
+
+  // start of cmsis buffer
+  ctx.buf = scratch_ptr;
+  ctx.size = scratch_buffer.size(0) - sizeof(kernel_sum_state);
```
digantdesai: this can also be in the `kernel_sum_state->size_for_cmsis()` or something..

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+    );
+    state->updated = true;
+    ET_LOG(
+        Info,
+        "Computed kernel sums, stored in state->buffer [required_bytes : %d]",
+        required_bytes);
+  }
+
+  // start of cmsis buffer
+  ctx.buf = scratch_ptr;
```
digantdesai: `kernel_sum_state->get_scrtach_ptr()`

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,206 @@
...
+  ctx.size = scratch_buffer.size(0) - sizeof(kernel_sum_state);
+
+  for (int32_t b = 0; b < batch_size; b++) {
```
digantdesai, line 88: did we verify this as the right way to call this API>?

examples/arm/aot_arm_compiler.py
digantdesai: Also per_channel and per_tensor quantization for weights

digantdesai (changes requested):
Great starting point. Thanks @psiddh for all the back and forth.
Left some comments.

digantdesai:
@AdrianLundell  - Just FYI. We need more work to do here but want to give you a heads up.

backends/cortex_m/ops/cortex_m_ops_common.h
psiddh: Done

backends/cortex_m/passes/passes_utils.py
psiddh: Removed

backends/cortex_m/passes/passes_utils.py
psiddh: Done

backends/cortex_m/passes/passes_utils.py
psiddh: Done

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Removed the log stmt

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Removed

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: ah one of the remaining logs, took it off

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Done

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Got it, I just left the first one entry

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Yes afaict it is correct because bias addition happens in arm_fully_connected_s8() / per channel api, not during kernel sum pre-computation.

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Refactored

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: I encapsulated everything in the simple helper class now

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Verify as-in , I get the success result from cmsis api call.
Again reading the impl the code,, we must loop over the batch dimension and call the function for each input vector. I think fully connected (FC) functions like arm_fully_connected_s8 and arm_fully_connected_per_channel_s8 process a single input vector at a time.

backends/cortex_m/ops/operators.py
psiddh: .item() is needed for single-element tensors to extract a Python scalar for correct math .. isn't ?

backends/cortex_m/CMakeLists.txt
psiddh: This flag can be turned on if you are trying to link to local CMSIS-nn lib instead of relying on FetchContent

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Yes

backends/cortex_m/ops/operators.py
digantdesai: float() can work with numel() == 1 and then you can get float when you do consume it or pass it down as as tensor..

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+// scratch_ptr(start)   scratch_ptr + cmsis_scratch   scratch_ptr + total_size
+//
+// - CMSIS-NN workspace: used by CMSIS-NN kernels for temporary data
+// - Always give CMSIS-NN the start of the buffer for alignment
+// - Place kernel_sum_state structs at the end to avoid breaking alignment
+class CMSISScratchBufferContext final {
+ public:
+  CMSISScratchBufferContext(
+      Tensor& scratch_buffer,
+      const cmsis_nn_dims& filter_dims)
```
digantdesai: Nit just take weight tensor ref as an arg?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+// Include CMSIS-NN headers with C linkage
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+namespace cortex_m {
+namespace native {
+using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
+
+// Pass allocates a flat scratch buffer:
```
digantdesai: Nit move this comment inside the class?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+class CMSISScratchBufferContext final {
+ public:
+  CMSISScratchBufferContext(
+      Tensor& scratch_buffer,
+      const cmsis_nn_dims& filter_dims)
+      : scratch_ptr_(scratch_buffer.mutable_data_ptr<int8_t>()),
+        total_size_(scratch_buffer.size(0)),
+        state_(reinterpret_cast<kernel_sum_state*>(
+            scratch_ptr_ + total_size_ - sizeof(kernel_sum_state))) {
+    validate_size(filter_dims);
```
digantdesai: +1

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+// - Place kernel_sum_state structs at the end to avoid breaking alignment
+class CMSISScratchBufferContext final {
+ public:
+  CMSISScratchBufferContext(
+      Tensor& scratch_buffer,
+      const cmsis_nn_dims& filter_dims)
+      : scratch_ptr_(scratch_buffer.mutable_data_ptr<int8_t>()),
+        total_size_(scratch_buffer.size(0)),
+        state_(reinterpret_cast<kernel_sum_state*>(
+            scratch_ptr_ + total_size_ - sizeof(kernel_sum_state))) {
```
digantdesai: why is this at the end? instead of the beginning..

state* = scratch_buffer;
scratch_ptr* = scratch_buffer + state_size; # doesn't have to be in the state

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+      int32_t weight_zp) {
+    if (!state_->updated) {
+      arm_vector_sum_s8(
+          reinterpret_cast<int32_t*>(scratch_ptr_),
+          in_feat,
+          out_feat,
+          weight_data,
+          weight_zp,
+          0, // rhs_offset
+          nullptr // bias
```
digantdesai: does this depend on if we have bias or not?

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: we may have some std unused attribute to make all compilers happy. Worst case just do `(void)bias_multiplier;` in the body..

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: OK for now but let's put a TODO(ISSUE) to confirm this..

backends/cortex_m/passes/passes_utils.py
digantdesai: No I was asking if you want to raise or continue LOL

backends/cortex_m/CMakeLists.txt
```diff
@@ -21,31 +21,59 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
...
+  # Add MVEI define to cmsis-nn target
+  target_compile_definitions(cmsis-nn PUBLIC ARM_MATH_MVEI=1)
+  # Get the correct source and binary dirs
+  FetchContent_GetProperties(
+    cmsis_nn
+    SOURCE_DIR CMSIS_NN_SOURCE_DIR
+    BINARY_DIR CMSIS_NN_BINARY_DIR
+  )
+  set(CMSIS_NN_INCLUDE_DIR "${CMSIS_NN_SOURCE_DIR}/Include")
+  set(CMSIS_NN_LIB "${CMSIS_NN_BINARY_DIR}/libcmsis-nn.a")
```
AdrianLundell: This is an absolute path to the binary tree which causes portability issues when installing. Could you take another look at this and try again to only use the cmsis-nn target when building?

backends/cortex_m/CMakeLists.txt
AdrianLundell: What do you think about giving the possibility to define a custom CMSIS-NN path, and if the lib does not already exist at that location it downloads it? We might want to make sure that it installs new version as well though to make thing more complicated.

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -114,6 +117,27 @@ inline void validate_quantization_params(
       "Single quant Output");
 }
 
+inline bool validate_per_channel_quant_params(
```
AdrianLundell, line 128: Would be nice to document where these constraints come from.

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+// Pass allocates a flat scratch buffer:
+// [------------------- scratch_buffer -----------------------]
+// |<- CMSIS-NN workspace ->|<--- kernel_sum_state struct --->|
+//
+// Buffer pointers:
+// ^                        ^                                 ^
+// scratch_ptr(start)   scratch_ptr + cmsis_scratch   scratch_ptr + total_size
+//
+// - CMSIS-NN workspace: used by CMSIS-NN kernels for temporary data
```
AdrianLundell: Nice comment, though I feel like it is missing a description of the kernel_sum_state struct?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+          "quantized_linear_out: CMSIS-NN failed with status [%d]",
+          status);
+      context.fail(Error::Internal);
+      return out;
+    }
+  }
+  return out;
+}
+
+// Functional variant (stub, not used at runtime)
```
AdrianLundell, line 147: I feel like we should work toward removing the need for these stub functions

examples/arm/aot_arm_compiler.py
```diff
@@ -297,6 +303,20 @@ def forward(self, x: torch.Tensor, y: torch.Tensor):
     can_delegate = True
 
 
+class QuantLinearTest(torch.nn.Module):
```
AdrianLundell: We are trying to move away from adding testing logic to the aot_arm_compliler, please see https://github.com/pytorch/executorch/issues/13902 for my suggestions on the cortex_m testing strategy

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,251 @@
...
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  cmsis_nn_fc_params fc_params;
+  fc_params.input_offset = -input_zp;
+  fc_params.output_offset = output_zp;
+  fc_params.activation.min = std::numeric_limits<int8_t>::min();
+  fc_params.activation.max = std::numeric_limits<int8_t>::max();
+  cmsis_nn_dims input_dims = {1, 1, 1, in_feat};
```
AdrianLundell, line 82: This is channels last where as pytorch per default is channels first, how are you handling that?

backends/cortex_m/ops/operators.py
```diff
@@ -223,3 +223,220 @@ def quantized_add_out_impl(
     out.copy_(result_quantized)
 
     return out
+
```
AdrianLundell, line 226: Should we start splitting these definitions into separate files?

backends/cortex_m/passes/quantized_linear_fusion_pass.py
```diff
@@ -0,0 +1,527 @@
...
+                logger.info(f"✅ Successfully fused {op_name} operation {fusion_count}")
+            except Exception as e:
+                logger.error(
+                    f"Failed to fuse {op_name} pattern for {fc_node.name}: {e}"
+                )
+                continue
+        self._cleanup_nodes(graph)
+        return fusion_count
+
+    def _find_original_input_placeholder(self, dq_node: Node) -> Node:
```
AdrianLundell: not used?

examples/arm/aot_arm_compiler.py
digantdesai: I think this can be reworked in the coming PRs, don't want to block this one. WDYT @AdrianLundell ?

AdrianLundell: One way would be that you test this way internally for now but don't upstream it, that way we don't have to undo it later?

backends/cortex_m/CMakeLists.txt
psiddh: Done

backends/cortex_m/CMakeLists.txt
psiddh: Done

backends/cortex_m/ops/cortex_m_ops_common.h
psiddh: Added doc , and explicitlt stated where these constraints are coming from

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Added comments

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: We then have to pass in/out feat along with weight tensor if we have to not pass 'cmsis_nn_dims' arg

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: I first tried with the struct at the beginning , this led to misalignment issues.

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: No, it doesn't depend on bias. Kernel sums compute weight, independent of bias

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: I think there is no spatial dims for linear, {1, 1, 1, in_feat} sets H=W=1 because linear ops operate on flattened vectors .
Looking [here](https://github.com/ARM-software/CMSIS-NN/blob/88f1982a69c00ed13dd633a63da1009c48abbb4d/Include/arm_nn_types.h#L89) :  it appears to NHWC.
Do we further need to handle  NCHW↔NHWC for linear ops ?

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Agree! I will create a follow up issue to clean this up further..

backends/cortex_m/ops/operators.py
psiddh: I will do it in the follow up task to clean this up, if that is ok with you .

backends/cortex_m/ops/operators.py
psiddh: Will do this in follow up change  (if you are ok )

backends/cortex_m/ops/operators.py
psiddh: Done

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Done

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Done

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Moved it to utils for now

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Good point, split it to _extract_input_quantization_parameters & _extract_output_quantization_parameters

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Oh forgot to update this in my last iter.. updated it now, Added comments , with references to the logic in CMSIS-NN impl

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: yeah fail-fast is better here as there's no reasonable fallback for missing dimensions

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: refactored to _prepare_bias_tensors

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: The output scale is already baked into the weight multipliers during quantization parameter calculation

backends/cortex_m/passes/quantized_linear_fusion_pass.py
psiddh: Removed!

examples/arm/aot_arm_compiler.py
psiddh: Added comprehensive tests    (but to aot_arm_compiler.py)  I will not add them to this PR as we are still discussing the CMSIS strategy, but i am using the following test coverage to validate my ongoing code changes
   "qlinear": QuantLinearTest,
    "qlinear_small_nobias_pt": QuantLinearTest_Small_NoBias_PerTensor,
    "qlinear_small_bias_pt": QuantLinearTest_Small_WithBias_PerTensor,
    "qlinear_medium_nobias_pc": QuantLinearTest_Medium_NoBias_PerChannel,
    "qlinear_large_nobias_pc": QuantLinearTest_Large_NoBias_PerChannel,
    "qlinear_large_nobias_pt": QuantLinearTest_Large_NoBias_PerTensor,
    "qlinear_large_bias_pc": QuantLinearTest_Large_WithBias_PerChannel,

examples/arm/aot_arm_compiler.py
psiddh: refactored this function now

psiddh:
Created this followup task for further cleanup of this PR based on the comments / feedback
https://github.com/pytorch/executorch/issues/14331

@digantdesai  @AdrianLundell

backends/cortex_m/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
digantdesai: We already have Meta copyright on the top

backends/cortex_m/CMakeLists.txt
```diff
@@ -21,70 +27,81 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
...
-)
-
-# Download and make CMSIS-NN available
-FetchContent_MakeAvailable(cmsis_nn)
+# CMSIS-NN configuration with dynamic path detection
+set(CMSIS_NN_VERSION "v7.0.0" CACHE STRING "CMSIS-NN version to download")
+set(CMSIS_NN_PATH "" CACHE PATH "Path to existing CMSIS-NN installation")
+
+# Try to find existing CMSIS-NN installation first
+if(CMSIS_NN_PATH AND EXISTS "${CMSIS_NN_PATH}")
```
digantdesai: remove this? Or add `LOCAL_` or similar prefix which makes it clear that this is not meant to be set "normally"

backends/cortex_m/CMakeLists.txt
```diff
@@ -21,70 +27,81 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
...
+    add_subdirectory(${cmsis_nn_SOURCE_DIR} ${cmsis_nn_BINARY_DIR})
+  endif()
+endif()
 
-# Print paths for debugging
-message(STATUS "CMSIS-NN source dir: ${cmsis_nn_SOURCE_DIR}")
-message(STATUS "CMSIS-NN binary dir: ${cmsis_nn_BINARY_DIR}")
+# Add MVEI define to cmsis-nn target
+if(TARGET cmsis-nn)
+  target_compile_definitions(cmsis-nn PUBLIC ARM_MATH_MVEI=1)
```
digantdesai, line 60: +1

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -22,6 +22,10 @@ using ScalarType = executorch::aten::ScalarType;
 using Scalar = torch::executor::Scalar;
 using Error = executorch::runtime::Error;
 
+// From arm_nn_math_types.h
+#define NN_Q31_MAX ((int32_t)(0x7FFFFFFFL))
+#define NN_Q31_MIN ((int32_t)(0x80000000L))
```
digantdesai: nit: if they are the same in arm_nn_math_types.h then prefix them with `ARM_`

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -114,6 +121,32 @@ inline void validate_quantization_params(
       "Single quant Output");
 }
 
+// From CMSIS-NN implementation, See arm_nn_requantize for details,
```
digantdesai: Nit Add perm-link?

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: move it inside the class? else it might "float away" :p

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,268 @@
...
+// Pass allocates a flat scratch buffer:
+// [------------------- scratch_buffer -----------------------]
+// |<- CMSIS-NN workspace ->|<--- kernel_sum_state struct --->|
+//
+// Buffer pointers:
+// ^                        ^                                 ^
+// scratch_ptr(start)   scratch_ptr + cmsis_scratch   scratch_ptr + total_size
+//
+// - CMSIS-NN workspace: used by CMSIS-NN kernels for temporary data
+// - Always give CMSIS-NN the start of the buffer for alignment
```
digantdesai: can we have a fat-header i.e. 32B to solve alignment problem? The reason I Don't like trailing header is because OOB read (if not also write) to match SIMD width is an acceptable practice.

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: see comment above re fat header and oob read/writes

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: Not sure I follow but OK for now.

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: https://github.com/ARM-software/CMSIS-NN/blob/88f1982a69c00ed13dd633a63da1009c48abbb4d/Source/FullyConnectedFunctions/arm_vector_sum_s8.c#L59

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,268 @@
...
+  bool is_kernel_sum_updated() const {
+    return state_->updated;
+  }
+
+  // Helper to compute kernel sums if not already done
+  void compute_kernel_sums_if_needed(
+      int32_t in_feat,
+      int32_t out_feat,
+      const int8_t* weight_data,
+      int32_t weight_zp) {
```
digantdesai: I see in/out feat comment, what I was suggesting was instead of taking `cmsis_filter_dims` at ctor and weight data/zp here - what if we take `weight_tensor` at ctor and hide the rest inside the class?

backends/cortex_m/ops/op_quantized_linear.cpp
digantdesai: I think you are right Sid. Make sure to add tests though if you haven't already.

backends/cortex_m/ops/operators.py
digantdesai: File an issue?

backends/cortex_m/ops/operators.py
digantdesai: File an issue?

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Refactored, resolving this comment

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Not sure how I missed it, it was right there :) . Now I pass the bias data too

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Good point, so I refactored this one more time to fit the FAT header at the beginning with 32 byte alignment.. also updated the diagram

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Done

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Ok, I think this time I factored in your ask

backends/cortex_m/ops/operators.py
psiddh: Yes already tracking as part of this issue: https://github.com/pytorch/executorch/issues/14331

backends/cortex_m/CMakeLists.txt
psiddh: fixed

backends/cortex_m/CMakeLists.txt
```diff
@@ -21,70 +21,90 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
 include(${EXECUTORCH_ROOT}/tools/cmake/Codegen.cmake)
 include(FetchContent)
 
-# CMSIS-NN version to download
+# CMSIS-NN configuration with dynamic path detection
 set(CMSIS_NN_VERSION
-    "v4.1.0"
+    "v7.0.0"
```
digantdesai, line 26: Nice!

backends/cortex_m/CMakeLists.txt
```diff
@@ -21,70 +21,90 @@ include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
...
-message("Generated files ${gen_command_sources}")
 
-# Build a library for cortex_m_kernels
+# Build library for cortex_m_kernels
 add_library(cortex_m_kernels ${_cortex_m_kernels__srcs})
-target_compile_options(cortex_m_kernels PUBLIC ${_common_compile_options})
 
-# Include directories for cortex_m_kernels
-target_include_directories(
+# Use PRIVATE for implementation dependencies to avoid INTERFACE pollution
```
digantdesai, line 88: +1

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -114,6 +121,33 @@ inline void validate_quantization_params(
       "Single quant Output");
 }
 
+// Refer to CMSIS-NN 'arm_nn_requantize' implementation for details:
+// https://fburl.com/afvegf0m
```
digantdesai: just regular github perm link? not sure if this is accessible in oss?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+  // └─────────────────┴─────────────────────────────────────┘
+  //          │                           │
+  //          │                           └─> Passed to CMSIS API
+  //          │
+  //          └─> Used by this C++ custom kernel only
+
+  // C++ Runtime:
+  // ┌─────────────────┬─────────────────────────────────────┐
+  // │   Fat Header    │        CMSIS Workspace              │
+  // │    (32 bytes)   │         (X bytes)                   │
```
digantdesai: Nit
```suggestion
  // │    (32 bytes)   │         (x bytes)                   │
```

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+  // └─────────────────┴─────────────────────────────────────┘
+  // ^                 ^
+  // │                 │
+  // scratch_ptr       cmsis_workspace_ptr
+  // │                 │
+  // ▼                 ▼
+  //             arm_vector_sum_s8() writes kernel sums (with bias if avail):
+  //             [sum₀+bias₀][sum₁+bias₁][sum₂+bias₂]...[sum_{n-1}+bias_{n-1}]
```
digantdesai: Nice subscript, sad can't do for `{n-1}` :-p
```suggestion
  //             [sum₀+bias₀][sum₁+bias₁][sum₂+bias₂]...[sum_{n-1}+bias_{n-1}]
```

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+  //             (n × 4-byte int32_t values = X bytes)
+  //
+  // - n = out_features (number of output features)
+  // - X = n × 4 bytes (total CMSIS buffer size)
```
digantdesai:
```suggestion
  // - x = n * 4 bytes (total CMSIS buffer size)
```

digantdesai: 4 because sizeof(int32)?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+        cmsis_workspace_ptr_(scratch_ptr_ + CMSIS_FAT_HEADER_SIZE),
+        base_ptr_(reinterpret_cast<uint8_t*>(scratch_ptr_)) {
+    // Extract dimensions from weight tensor
+    // weights shape: [out_features, in_features]
+    const int32_t in_feat = weights.size(1);
+    const int32_t out_feat = weights.size(0);
+
+    // Initialize fat header with offsets
+    header_->kernel_sum_state.updated = false;
+    header_->kernel_sum_state.required_size = 0;
```
digantdesai: why do we need this?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+        calculate_offset(weights.const_data_ptr<int8_t>());
+    header_->weight_zp_data_offset =
+        calculate_offset(weight_zero_point.const_data_ptr<int32_t>());
+    header_->bias_data_offset = bias.has_value()
+        ? calculate_offset(bias.value().const_data_ptr<int32_t>())
+        : 0; // 0 = no bias
+
+    // Create filter dims for validation
+    cmsis_nn_dims filter_dims = {in_feat, 1, 1, out_feat};
+    validate_size(filter_dims);
```
digantdesai: Do we want to assert first before using the args?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+    int32_t in_features;
+    int32_t out_features;
+
+    // Store offsets instead of raw pointers
+    uint32_t weight_data_offset;
+    uint32_t weight_zp_data_offset;
+    uint32_t bias_data_offset; // (0 if no bias)
+
+    bool is_per_channel;
+    uint8_t padding[3]; // 3 bytes explicit padding for 32-byte alignment
```
digantdesai: I am not sure why do these have to be in this struct as opposed to in the `CMSISScratchBufferContext`

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+  int8_t* cmsis_workspace_ptr_; // Points to 32-byte aligned CMSIS workspace
+  uint8_t* base_ptr_; // Base pointer for offset calculations
+
+  // Helper to calculate offset from base pointer
+  uint32_t calculate_offset(const void* ptr) const {
+    if (ptr == nullptr)
+      return 0;
+    const uint8_t* ptr_bytes = reinterpret_cast<const uint8_t*>(ptr);
+    ET_CHECK_MSG(ptr_bytes >= base_ptr_, "Pointer is before base address");
+    uint64_t offset = ptr_bytes - base_ptr_;
```
digantdesai:
```suggestion
    const std::ptrdiff_t offset = ptr_bytes - base_ptr_;
```

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  // Initialize scratch buffer context
+  CMSISScratchBufferContext scratch_buf_ctx_mgr(
+      const_cast<Tensor&>(scratch_buffer), weights, weight_zero_point, bias);
+  cmsis_nn_context ctx = scratch_buf_ctx_mgr.get_cmsis_ctx();
+  // Compute kernel sums if needed
+  scratch_buf_ctx_mgr.compute_kernel_sums_if_needed();
```
digantdesai:
```suggestion
  // Compute kernel sums if needed
  scratch_buf_ctx_mgr.compute_kernel_sums_if_needed();
  
  cmsis_nn_context ctx = scratch_buf_ctx_mgr.get_cmsis_ctx();
```

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+  //             arm_vector_sum_s8() writes kernel sums (with bias if avail):
+  //             [sum₀+bias₀][sum₁+bias₁][sum₂+bias₂]...[sum_{n-1}+bias_{n-1}]
+  //             (n × 4-byte int32_t values = X bytes)
+  //
+  // - n = out_features (number of output features)
+  // - X = n × 4 bytes (total CMSIS buffer size)
+  // - Total buffer = 32 + x bytes
+
+ public:
+  CMSISScratchBufferContext(
```
digantdesai: Let's move this out of linear since other ops can also use this.

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -0,0 +1,369 @@
...
+    return header_->out_features;
+  }
+
+ private:
+  static constexpr size_t CMSIS_FAT_HEADER_SIZE = 32;
+
+  // Fixed-size metadata header using offsets for architecture independence.
+  // Stores cached weight information and kernel sum computation state.
+  // Ensures 32-byte alignment for the CMSIS workspace that follows.
+  struct CMSISFatHeader {
```
digantdesai: Nit: ```suggestion
  struct CMSISHeader {
```

Comment (digantdesai) [inline backends/cortex_m/ops/op_quantized_linear.cpp]: The only purpose of introducing a larger (=fat though we don't want to call it that :-p) header was to fix alignment for the kernel_sum buffer. We don't need to have that as a concept at runtime, header size of 32 at aot, and an aligned tensor from ET runtime is sufficient.

Ok for now but let's refactor and scrub it in the next PR. Also move it out of linear.cpp

Comment (digantdesai) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: File a task? 🙏

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 67 above)
Comment (digantdesai) [inline backends/cortex_m/ops/op_quantized_linear.cpp]: You don't need any padding here, as long as this assert is OK,

`cmsis_workspace_ptr_ % 32 == 0`

File: backends/cortex_m/passes/quantized_linear_fusion_pass.py (hunk 14 above)
Comment (digantdesai) [reply backends/cortex_m/passes/quantized_linear_fusion_pass.py:40]: TBH we need the opposite but strictly speaking this is more accurate. We can move to `aten.linear` and drop support for `aten.addmm` and `aten.mm` later. Please file a task?

**Review by digantdesai — APPROVED** [review]
We need some more work but let's move forward. Make a list of tasks and we can fix these next..

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 67 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Yep

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 71 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: I moved the struct into another header already.

Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: renamed to KernelSumHeader

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 70 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Done

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 68 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Done

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 69 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Done

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 67 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Done

File: backends/cortex_m/ops/op_quantized_linear.cpp (hunk 66 above)
Comment (psiddh) [reply backends/cortex_m/ops/op_quantized_linear.cpp]: Done

File: backends/cortex_m/ops/cmsis_scratch_buffer_context.h [hunk 72]
```diff
@@ -0,0 +1,165 @@
...
+  int32_t get_out_features() const { return out_features_; }
+
+ private:
+  static constexpr size_t KERNEL_SUM_HEADER_SIZE = 32;
+
+  // Header for kernel sum computation state only
+  struct KernelSumHeader {
+    bool updated = false;
+    int32_t required_size = 0;
+    uint8_t padding[24]; // Pad to 32 bytes for alignment
```
digantdesai: why do we need both of these fields

backends/cortex_m/ops/cmsis_scratch_buffer_context.h
```diff
@@ -0,0 +1,165 @@
...
+        header_(reinterpret_cast<KernelSumHeader*>(scratch_ptr_)),
+        cmsis_workspace_ptr_(scratch_ptr_ + KERNEL_SUM_HEADER_SIZE) {
+    
+    cmsis_nn_dims filter_dims = {in_features_, 1, 1, out_features_};
+    validate_size(filter_dims);
+  }
+
+  cmsis_nn_context get_cmsis_ctx() const {
+    cmsis_nn_context ctx;
+    ctx.buf = cmsis_workspace_ptr_;
```
digantdesai, line 78: Do you want to put an alignment check on this pointer before passing it down to CMSIS?

backends/cortex_m/ops/op_quantized_linear.cpp
psiddh: Ys

backends/cortex_m/ops/cmsis_scratch_buffer_context.h
psiddh: required_size is used for assertion/validation along with logging . It is used to compare the size passed by aot vs c++ runtime required size by cmsis api , looking at cmsis impl code, standard 4 byte alignment is enough.so removed extra padding field

backends/cortex_m/ops/cmsis_scratch_buffer_context.h
psiddh: like this you meant ?
ET_CHECK_MSG(
        reinterpret_cast<uintptr_t>(cmsis_workspace_ptr_) % 4 == 0,
        "CMSIS workspace not 4-byte aligned");

AdrianLundell:
Hi, our builds are failing after this was merged, we should make sure to test the arm backend after changes in the cortex_m backend since they are so intertwined. Can you have a look?

reproducer:
```
 backends/arm/scripts/build_executorch.sh
```
error:
```
In file included from /home/svc_mlsystest/executorch/backends/cortex_m/ops/op_quantized_linear.cpp:9:
11:42:08  /home/svc_mlsystest/executorch/backends/cortex_m/ops/cmsis_scratch_buffer_context.h: In member function 'void cortex_m::native::CMSISScratchBufferContext::compute_kernel_sums_if_needed()':
11:42:08  /home/svc_mlsystest/executorch/backends/cortex_m/ops/cmsis_scratch_buffer_context.h:89:7: error: 'arm_vector_sum_s8' was not declared in this scope
11:42:08     89 |       arm_vector_sum_s8(
11:42:08        |       ^~~~~~~~~~~~~~~~~
11:42:08  /home/svc_mlsystest/executorch/backends/cortex_m/ops/op_quantized_linear.cpp: In function 'Tensor& cortex_m::native::quantized_linear_out(KernelRuntimeContext&, const Tensor&, const Scalar&, const Scalar&, const Scalar&, const Tensor&, const Tensor&, const Tensor&, const Tensor&, const std::optional<executorch::runtime::etensor::Tensor>&, const Tensor&, const Tensor&, const Tensor&, const Scalar&, const Scalar&, const Scalar&, Tensor&)':
11:42:08  /home/svc_mlsystest/executorch/backends/cortex_m/ops/op_quantized_linear.cpp:103:16: error: 'arm_fully_connected_per_channel_s8' was not declared in this scope; did you mean 'arm_fully_connected_s8'?
11:42:08    103 |       status = arm_fully_connected_per_channel_s8(
11:42:08        |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
11:42:08        |                arm_fully_connected_s8
11:42:08  gmake[2]: *** [backends/cortex_m/CMakeFiles/cortex_m_kernels.dir/build.make:121: backends/cortex_m/CMakeFiles/cortex_m_kernels.dir/ops/op_quantized_linear.cpp.obj] Error 1
11:42:08  gmake[2]: *** Waiting for unfinished jobs....
11:42:08  gmake[1]: *** [CMakeFiles/Makefile2:1110: backends/cortex_m/CMakeFiles/cortex_m_kernels.dir/all] Error 2
11:42:08  gmake[1]: *** Waiting for unfinished jobs....
```

AdrianLundell:
After some more investigation it builds for a clean environment so this works fine, would be nice long term to fix the build scripts so that they are able to do incremental rebuilds though instead of having to do a full rebuild.

examples/arm/aot_arm_compiler.py
```diff
@@ -806,22 +812,24 @@ def to_edge_no_delegate(exported_program, args, model: torch.nn.Module, example_
...
-    # Conditionally add the QuantizedOpFusionPass
+    passes = [ReplaceQuantNodesPass]
     if args.enable_qdq_fusion_pass:
-        passes.append(QuantizedOpFusionPass())
-
-    # Apply the passes
-    edge = edge.transform(passes)
-
-    return edge
+        passes += [QuantizedLinearFusionPass, QuantizedOpFusionPass]
```
digantdesai, line 823: ok so this is enabled and we don't have any tests. can you either disable or land at least some tests?

PR #14258, opened 2025-09-12
Arm backend: Int16 linear support

### LINKED ISSUE #13729: [Arm] Support Linear INT16 TOSA reference model run

With this PR: https://github.com/pytorch/executorch/pull/13658/files we add support and test to 16a8w_quantize and lower Linear op. This fails at tosa reference model run and been marked as xfail. Enable run on TOSA ref model and enable this test.
cc: @gggekov , @digantdesai , @3l1 

### Alternatives

_No response_

### Additional context

https://github.com/pytorch/executorch/issues/13635

### RFC (Optional)

_No response_

per:
### Summary
Adds support for a16w8 for linear when targeting a backend with +int16 extension.

Fixes #13729

### Test plan
Tested through unit tests.

### DIFF
--- a/backends/arm/_passes/add_bias_pass.py
+++ b/backends/arm/_passes/add_bias_pass.py
@@ -8,6 +8,7 @@ from typing import Set, Type
 import torch
 from executorch.backends.arm._passes import ArmPass
 from executorch.backends.arm._passes.arm_pass_utils import get_first_fake_tensor
+from executorch.backends.arm.tosa.mapping import TosaSpecialDtype
 from executorch.backends.transforms.utils import create_constant_placeholder
 
 from executorch.exir.dialects._ops import ops as exir_ops
@@ -60,7 +61,9 @@ class AddBiasPass(ArmPass):
                         name=f"{node.name}_bias",
                     )
                     if node.args[0].meta["val"].dtype == torch.int16:
-                        bias_node.meta["tosa_dtype_48bit"] = True
+                        bias_node.meta[TosaSpecialDtype.meta_key()] = (
+                            TosaSpecialDtype.INT48
+                        )
                 node.update_arg(2, bias_node)
 
         if modified:
--- a/backends/arm/_passes/add_bias_pass.py
+++ b/backends/arm/_passes/add_bias_pass.py
@@ -59,6 +59,8 @@ class AddBiasPass(ArmPass):
                         persistent_buffer=True,
                         name=f"{node.name}_bias",
                     )
+                    if node.args[0].meta["val"].dtype == torch.int16:
+                        bias_node.meta["tosa_dtype_48bit"] = True
                 node.update_arg(2, bias_node)
 
         if modified:
--- a/backends/arm/_passes/arm_pass_manager.py
+++ b/backends/arm/_passes/arm_pass_manager.py
@@ -42,6 +42,7 @@ from executorch.backends.arm._passes import (
     DecomposeAtanPass,
     DecomposeAvgPool2d,
     DecomposeBatchNormNoStatsPass,
+    DecomposeConv2dWithInt16ActivationPass,
     DecomposeCoshPass,
     DecomposeCosineSimilarityPass,
     DecomposeCumsumPass,
@@ -183,6 +184,7 @@ class ArmPassManager(PassManager):
         self.add_pass(ComputeConstantOpsAOT(exported_program))
 
         self.add_pass(DecomposeGroupedConv())
+
         self.add_pass(ConvertExpandCopyToRepeatPass())
         self.add_pass(UnsqueezeBeforeRepeatPass())
         self.add_pass(CastInt64BuffersToInt32Pass(exported_program))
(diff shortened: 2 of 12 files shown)


per:
Unrelated failures.

backends/arm/_passes/add_bias_pass.py
```diff
@@ -55,6 +55,8 @@ def call(self, graph_module):
                         persistent_buffer=True,
                         name=f"{node.name}_bias",
                     )
+                    if node.args[0].meta["val"].dtype == torch.int16:
+                        bias_node.meta["tosa_dtype_48bit"] = True
```
digantdesai: Nit: something like..
```suggestion
                        bias_node.meta["tosa_dtype"] = tosa_dtype.INT48
```

backends/arm/_passes/arm_pass_manager.py
```diff
@@ -167,9 +169,14 @@ def _tosa_INT_pipeline(self, exported_program: ExportedProgram) -> GraphModule:
 
         self.add_pass(FuseViewCopyTransform())
         self.add_pass(FuseConstantArgsPass(exported_program))
+        self.add_pass(InsertTableOpsPass(exported_program))
+        # If we have a conv2d with int16 activation split up into a convolution
+        # and an addition, to work-around the lack of support for int48 in torch
```
digantdesai, line 203: > # and an addition, to work-around the lack of support for int48 in torch
Or can it be done by using torch.dtype.int64 instead and then detecting and lowering it as int48 downstream?

digantdesai (approved):
This is awesome, No conv tests for 16a8w, just curious.

backends/arm/_passes/add_bias_pass.py
per: 👍

backends/arm/_passes/arm_pass_manager.py
per: I was starting of in that direction, but it interfere a bit with the int64->int32 handling, so rather keep it separate.

digantdesai: yeah given int64 is treated as radioactive :P

backends/arm/test/ops/test_linear.py
```diff
@@ -276,10 +275,19 @@ def get_symmetric_a16w8_linear_quantizer(
...
+test_data_all_16a8w.pop(
+    "model_linear_rank4_negative_large_rand,per_channel_quant=False"
+)
+test_data_all_16a8w.pop("model_linear_rank1_negative_large_rand,per_channel_quant=True")
+test_data_all_16a8w.pop(
+    "model_linear_rank1_negative_large_rand,per_channel_quant=False"
 )
+
+
+@common.parametrize("test_data", test_data_all_16a8w)
```
digantdesai, line 287: @per - Does it work with U55 and U85?
@Ninja91 - IIRC you already have tests seems like haven't merged?

zingo:
I see this testfail :( 
[test-arm-backend (test_pytest_ops_ethosu_fvp) / linux-job https://github.com/pytorch/executorch/actions/runs/17768877335/job/50517971935?pr=14258#logs

FAILED backends/arm/test/ops/test_linear.py::test_linear_16a8w_tosa_INT[model_linear_rank1_large_randn,per_channel_quant=True] - AssertionError: Output 0 does not match reference output.
	Given atol: 0.004528127912431955, rtol: 0.001.
	Output tensor shape: torch.Size([20]), dtype: torch.float32
	Difference: max: 0.1481781005859375, abs: 0.1481781005859375, mean abs error: 0.008467483520507812.
	-- Model vs. Reference --
	 Numel: 20, 20
	Median: 14.398289680480957, 14.398289680480957
	  Mean: 2.563008487224579, 2.555246722698212
	   Max: 69.8498764038086, 69.84635162353516
	   Min: -115.46151733398438, -115.60969543457031

backends/arm/test/ops/test_linear.py
per: No, not yet. Test will be added for Ethos-U with a Vela update when it's in place.

zingo:
I removed milestone 1.0 for now Vela does not support this anyway so lets take this a bit calmer.

zingo:
@digantdesai , this has come out of sync with Meta internal version and I'm not allowed merge it, would you like to assist?

digantdesai:
yeah let me try to merge this.

digantdesai:
if it doesn't work I can give you a patch..

zingo:
Fails are unrelated
Note: Fail in trunk / test-arm-ootb-linux / linux-job  is unrelated and handled by another PR

zingo:
@digantdesai  we had to rebase as a file was changed and as we changed it again we got back to the same problem and need your help again. Sorry about that.

backends/arm/test/ops/test_linear.py
Ninja91: I have added tests for U55 and U85

PR #14263, opened 2025-09-12
Do not run `lintrunner init` in lintrunner CI job

swolchok:
Two reasons:
1) It will end up spamming pip install -r requirements-dev.txt once for each linter because of the way we have init commands set up. This is fixable, but:
2) setup_linux sets up lintrunner for us already anyway via install_requirements.


swolchok:
* __->__ #14263

swolchok:
hmm, this is not much if any faster, guess the split jobs were a good idea and so is not running on all files

PR #14272, opened 2025-09-12
Update version.txt to 1.0

GregoryComer:
### Summary
Update version.txt to 1.0.0 in preparation for the release branch cut.


version.txt
```diff
@@ -1 +1 @@
-0.8.0a0
+1.0.0a0
```
mergennachin, line 1: oh yeah, should've been 1.0.0 after 0.7.0 to begin with.

mergennachin (approved):

version.txt
abhinaykukkadapu: @mergennachin do we need `a0` at the end for 1.0 release as it is not treated as alpha release?

PR #14274, opened 2025-09-12
Add CI for checking file size

cccclai:
As title. xrefs will hang if the files change too much, however it's a bit hard to tell. Make file size check explicit. The rule is following
1. For all files, the file size can't be larger than 1MB
2. For images/vidoes, the files size can't be larger than 7MB
3. There is an exception list defined in the script if it's really needed

### DIFF
--- a/.github/workflows/_link_check.yml
+++ b/.github/workflows/_link_check.yml
@@ -55,3 +55,27 @@ jobs:
           echo "Or add \`@lint-ignore\` somewhere on the same line as the reference you want to skip checking."
           exit 1
         }
+
+  lint-file-size:
+    if: ${{ github.event_name == 'pull_request' }}
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      runner: linux.2xlarge
+      docker-image: ci-image:executorch-ubuntu-22.04-linter
+      submodules: false
+      fetch-depth: 0
+      ref: ${{ inputs.ref }}
+      timeout: 30
+      script: |
+        ./scripts/lint_file_size.sh $(
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}"
+          else
+            echo "${{ github.event.before }}" "${{ github.sha }}"
+          fi
+        ) || {
+          echo
+          echo "File size lint failed: some files exceed the 1 MB limit."
+          echo "If you really need large files, consider using Git LFS or storing them elsewhere."
+          exit 1
+        }
--- a/.github/workflows/_link_check.yml
+++ b/.github/workflows/_link_check.yml
@@ -67,7 +67,7 @@ jobs:
       ref: ${{ inputs.ref }}
       timeout: 30
       script: |
-        sleep 900
+        chmod +x ./scripts/lint_file_size.sh
         ./scripts/lint_file_size.sh $(
           if [ "${{ github.event_name }}" = "pull_request" ]; then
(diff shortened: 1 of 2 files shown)


zingo:
Might be nice to have this check in lintrunner also down the line.

cccclai:
> Might be nice to have this check in lintrunner also down the line.

I think this check each file already. What does it mean by down the line?

scripts/lint_file_size.sh
```diff
@@ -0,0 +1,73 @@
...
+
+set -euo pipefail
+
+status=0
+
+green='\e[1;32m'; red='\e[1;31m'; cyan='\e[1;36m'; reset='\e[0m'
+
+# Following is the rules for the file size linting:
+# 1. For all files, the file size can't be larger than 1MB
+# 2. For images/vidoes, the files size can't be larger than 7MB
```
swolchok: nit: videos, the file size

scripts/lint_file_size.sh
```diff
@@ -0,0 +1,73 @@
...
+
+for file in $files; do
+  if is_exception "$file"; then
+    echo -e "${cyan}SKIP${reset}  $file (in exception list)"
+    continue
+  fi
+  if [ -f "$file" ]; then
+    # Set size limit depending on extension
+    if [[ "$file" =~ \.(png|jpg|jpeg|gif|svg|mp3|mp4)$ ]]; then
+      MAX_SIZE=$((8 * 1024 * 1024))  # 5 MB for pictures
```
swolchok: 5 or 8? also the description above says 7, and mp3 isn't an image or a video

.github/workflows/_link_check.yml
```diff
@@ -55,3 +55,29 @@ jobs:
...
         }
+
+  lint-file-size:
+    if: ${{ github.event_name == 'pull_request' }}
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      runner: linux.2xlarge
+      docker-image: ci-image:executorch-ubuntu-22.04-linter
+      submodules: false
+      fetch-depth: 0
```
swolchok, line 66: IIRC 0 is unlimited. couldn't we make this some large (because it gates the maximum depth of a stack) but finite number, like say 50?

.github/workflows/_link_check.yml
```diff
@@ -55,3 +55,29 @@ jobs:
           echo "Or add \`@lint-ignore\` somewhere on the same line as the reference you want to skip checking."
           exit 1
         }
+
+  lint-file-size:
+    if: ${{ github.event_name == 'pull_request' }}
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      runner: linux.2xlarge
```
swolchok, line 63: do we have smaller runners we can use? this isn't particularly resource-intensive

swolchok (approved):
minor quibbles but directionally this is fine

scripts/lint_file_size.sh
cccclai: Oops sorry, I was trying to find out the files exceed the limit and forget to update...

.github/workflows/_link_check.yml
cccclai: Tried linux.large and docker image pulling failed. Bump it to 2x

.github/workflows/_link_check.yml
cccclai: I tried look up `fetch-depth` in both executorch and pytorch, and looks like all are defined as 0 or 1...I can't find definition for this variable so using the same as other jobs to be safe

PR #14277, opened 2025-09-12
Remove generate_from_pos since there's no user of it

larryliu0820:
As titled, rely on CI jobs

### DIFF
--- a/extension/llm/runner/text_llm_runner.cpp
+++ b/extension/llm/runner/text_llm_runner.cpp
@@ -43,6 +43,7 @@ TextLLMRunner::TextLLMRunner(
       io_manager_(std::move(io_manager)),
       text_token_generator_(std::move(text_token_generator)),
       stats_(std::move(stats)),
+      pos_(0),
       temperature_(temperature) {
   // Note: This constructor assumes that text_prefiller and text_token_generator
   // already have references to the Module and TextDecoderRunner they need
@@ -70,9 +71,8 @@ Error TextLLMRunner::load() {
     ET_LOG(Info, format, __VA_ARGS__);     \
   }
 
-Error TextLLMRunner::generate_from_pos(
+Error TextLLMRunner::generate(
     const std::string& prompt,
-    ET_UNUSED int64_t start_pos,
     const GenerationConfig& config,
     std::function<void(const std::string&)> token_callback,
     std::function<void(const Stats&)> stats_callback) {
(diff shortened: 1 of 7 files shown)


mergennachin:
Doesn't this API adhere to the lifecycle policy? Stable, experimental, deprecated... ?

kirklandsign (approved):

mergennachin:
We need at least two releases to deprecate an API since beta.

larryliu0820:
> We need at least two releases to deprecate an API since beta.

It's marked as experimental though

extension/llm/runner/text_llm_runner.cpp
```diff
@@ -217,15 +217,6 @@ Error TextLLMRunner::generate_from_pos(
   return Error::Ok;
 }
 
-Error TextLLMRunner::generate(
-    const std::string& prompt,
-    const GenerationConfig& config,
-    std::function<void(const std::string&)> token_callback,
-    std::function<void(const Stats&)> stats_callback) {
-  pos_ = 0;
```
lucylq, line 225: I think this behaves slightly differently to setting it once in the constructor

lucylq: When we set pos_=0 inside generate:
```
I 00:00:22.490268 executorch:text_llm_runner.cpp:167] LFQ: pos before prefill_ is 0
...
I 00:00:27.365336 executorch:text_llm_runner.cpp:167] LFQ: pos before prefill_ is 0
```
When we set pos_0 only inside the constructor:
```
I 00:00:22.584236 executorch:text_llm_runner.cpp:167] LFQ: pos before prefill_ is 0
...
I 00:00:26.869992 executorch:text_llm_runner.cpp:167] LFQ: pos before prefill_ is 9
```

Sounds like when pos_= 0, it sets a new whitespace at the start??

extension/llm/runner/text_llm_runner.cpp
```diff
@@ -43,6 +43,7 @@ TextLLMRunner::TextLLMRunner(
       io_manager_(std::move(io_manager)),
       text_token_generator_(std::move(text_token_generator)),
       stats_(std::move(stats)),
+      pos_(0),
```
lucylq, line 46: vs this happens once in the constructor, and not on each generate call

extension/llm/runner/text_llm_runner.cpp
```diff
@@ -70,9 +71,8 @@ Error TextLLMRunner::load() {
     ET_LOG(Info, format, __VA_ARGS__);     \
   }
 
-Error TextLLMRunner::generate_from_pos(
+Error TextLLMRunner::generate(
     const std::string& prompt,
-    ET_UNUSED int64_t start_pos,
     const GenerationConfig& config,
     std::function<void(const std::string&)> token_callback,
     std::function<void(const Stats&)> stats_callback) {
```
lucylq, line 78: add pos_ = 0 here fixes the lora failure for me

lucylq (approved):

PR #14290, opened 2025-09-12
Add android target recipes and extensive model tests using ios and android recipes

abhinaykukkadapu:
Summary:
Changes in this diff:
1. Adds android target recipes using QNN backend.
2. Add extensive model lowering and accuracy checks using torchvision models.

```
from executorch.export import export
qnn_android_recipe = get_android_recipe()

# lower a model to pte

out = session.run_method("forward", *example_inputs)
```

Differential Revision: D82284987

### DIFF
--- a/backends/qualcomm/_passes/TARGETS
+++ b/backends/qualcomm/_passes/TARGETS
@@ -15,5 +15,6 @@ runtime.python_library(
         "//executorch/backends/transforms:decompose_sdpa",
         "//executorch/exir/backend:backend_details",
         "//executorch/exir/backend:compile_spec_schema",
+        "//executorch/backends/qualcomm/quantizer:quantizer",
     ],
 )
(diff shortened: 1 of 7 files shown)


backends/qualcomm/_passes/TARGETS
```diff
@@ -15,5 +15,6 @@ runtime.python_library(
         "//executorch/backends/transforms:decompose_sdpa",
         "//executorch/exir/backend:backend_details",
         "//executorch/exir/backend:compile_spec_schema",
+        "//executorch/backends/qualcomm/quantizer:quantizer",
```
cccclai, line 18: Is it needed?

abhinaykukkadapu: Yeah, weirdly one of the pass requires this import via buck.

cccclai: which pass is it?

abhinaykukkadapu: It is `seq_mse`, here is a stripped down version:

```
../executorch/backends/qualcomm/_passes/seq_mse.py", line 11, in <module>
    from executorch.backends.qualcomm.quantizer.observers.per_block_param_observer import (
ModuleNotFoundError: No module named 'executorch.backends.qualcomm.quantizer'
```

cccclai: oh.....I see yeah that makes sense

cccclai (approved):

PR #14308, opened 2025-09-15
solved serde issue when serializing op using list of faketensor as output

Gasoonjia:
Summary: as title.

Differential Revision: D82473504

### DIFF
--- a/exir/serde/export_serialize.py
+++ b/exir/serde/export_serialize.py
@@ -1211,7 +1211,7 @@ class GraphModuleSerializer:
         """
         meta_val = node.meta["val"]
 
-        if isinstance(meta_val, tuple):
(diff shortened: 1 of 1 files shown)


exir/serde/export_serialize.py
```diff
@@ -1211,7 +1211,7 @@ def serialize_hoo_outputs(self, node: torch.fx.Node) -> List[Argument]:
         """
         meta_val = node.meta["val"]
 
-        if isinstance(meta_val, tuple):
+        if isinstance(meta_val, (tuple, list)):
```
JacobSzwejbka, line 1214: why not just check sequence then

JacobSzwejbka (approved):

Issue #14314, opened 2025-09-15
Re-enable OpenVINO CI


GregoryComer:
The OpenVINO CI job is currently timing out. Example run: https://github.com/pytorch/executorch/actions/runs/17743994437/job/50424658111. We're going to temporarily disable the job to allow viable/strict to move and avoid false signal, but we need to re-enable it as soon as possible to get CI coverage.

GregoryComer:
@kimishpatel Can you help find an owner for this?

kimishpatel:
Assigned to @ynimmaga

PR #14315, opened 2025-09-15
Temporarily disable OpenVINO CI job

GregoryComer:
Temporarily disable the OpenVINO CI job due to consistent timeouts. See https://github.com/pytorch/executorch/issues/14314 for the tracking task, which has more details. It is intended that this job is fixed and re-enabled quickly.


mergennachin (approved):

kimishpatel:
cc: @ynimmaga can you take a look?

PR #14320, opened 2025-09-15
Add minimum_length to extended header for BC

lucylq:
Summary:
Error happening when we have older PTE files with extended header size 24. 

When we call 'from_bytes', we expect header size 32 after adding segment_data_size field. 

This is BC on C++ side because we have a minimum length.
Add minimum length to python to make the change BC.

Differential Revision: D82492169

### DIFF
--- a/exir/_serialize/_program.py
+++ b/exir/_serialize/_program.py
@@ -136,8 +136,7 @@ class _ExtendedHeader:
 
     # The magic bytes that should be at the beginning of the header.
     EXPECTED_MAGIC: ClassVar[bytes] = b"eh00"
-    # The length of the header in bytes.
-    EXPECTED_LENGTH: ClassVar[int] = (
+    MINIMUM_LENGTH: ClassVar[int] = (
         # Header magic
         4
         # Header length
@@ -146,9 +145,18 @@ class _ExtendedHeader:
         + 8
         # Segment base offset
         + 8
+    )
+    # The length of the header in bytes.
+    EXPECTED_LENGTH: ClassVar[int] = (
+        MINIMUM_LENGTH
         # Segment data size
         + 8
     )
+
+    # To find the header, callers should provide at least this many bytes of
+    # the head of the serialized Program data. Keep this in sync with
+    # kNumHeadBytes in //executorch/schema/extended_header.cpp
+    NUM_HEAD_BYTES: ClassVar[int] = 64
 
     # Instance attributes. @dataclass will turn these into ctor args.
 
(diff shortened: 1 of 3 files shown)


exir/_serialize/test/test_program.py
```diff
@@ -1009,14 +1009,30 @@ def test_named_data_segments(self) -> None:
 EXAMPLE_HEADER_DATA: bytes = (
     # Magic bytes
     b"eh00"
-    # uint32_t header size (little endian)
+    # uint32_t header size (little endian) --> 32
```
JacobSzwejbka: ? whats this

JacobSzwejbka: oh its just the number 32 in little endian in hex

lucylq: The header size (x20) is 32 bytes. This is the new one with segment_data_size.

Below is the old one (x18), 24 bytes.

exir/_serialize/_program.py
```diff
@@ -146,10 +145,18 @@ class _ExtendedHeader:
...
+    # The length of the header in bytes.
+    EXPECTED_LENGTH: ClassVar[int] = (
+        MINIMUM_LENGTH
         # Segment data size
         + 8
     )
 
+    # To find the header, callers should provide at least this many bytes of
+    # the head of the serialized Program data.
+    NUM_HEAD_BYTES: ClassVar[int] = 64
```
JacobSzwejbka, line 159: Is 64 just a random number with some leeway for future expansion?

JacobSzwejbka (approved):

exir/_serialize/_program.py
lucylq: This is in sync with the C++ kNumHeadBytes, I think the value 64 is arbitrary leeway.

mergennachin:
Hi @lucylq -- this is concerning.

Is the extended header a BC breaking change for the C++ runtime? We shouldn't make do any BC breaking. In other words, older .PTE files should be runnable with new runtime. 

This PR seems like only do mitigation for python side -- but it doesn't make sure older PTE has still be loadable.

lucylq:
> Hi @lucylq -- this is concerning.
> 
> Is the extended header a BC breaking change for the C++ runtime? We shouldn't make do any BC breaking. In other words, older .PTE files should be runnable with new runtime.
> 
> This PR seems like only do mitigation for python side -- but it doesn't make sure older PTE has still be loadable.

Hi @mnachin the PRs landed last week are not BC breaking for C++ side; we check for segment_data_size if it exists and not otherwise. 

There isn't an explicit test in executorch given we generate our PTE files on the fly (so they have segment_data_size), however there are tests from internal et users that cover this. I can check in an older PTE file if we want to explicitly test this in C++ as well?

It was a breaking change for python (which is my bad), which this PR changes.

mergennachin:
> I can check in an older PTE file if we want to explicitly test this in C++ as well?

@lucylq @JacobSzwejbka 

Yeah, going back to release/0.6, generate an old .PTE file and save it something like "model_v_0_6.pte" and then load this in the new runtime.

We should also look into automating this soon.

mergennachin:
Actually we should start with release/0.4 -- that's when released beta

mergennachin (approved):
The python side looks good neverthess.

PR #14332, opened 2025-09-16
Add selective build support for prim ops

kimishpatel:
Summary:
This diff implements selective build functionality for primitive operations (prim ops) in ExecutorTorch, allowing users to include only specific prim ops in their builds to reduce binary size and compilation time.

## Key Changes:

1. **Conditional compilation in register_prim_ops.cpp**: Wrapped each of the prim op registrations with conditional compilation macros that check both selective build enablement (`EXECUTORCH_ENABLE_PRIM_OPS_SELECTIVE_BUILD`) and individual op selection (e.g., `INCLUDE_EXECUTORCH_PRIM_ET_VIEW_DEFAULT`).

2. **Code generation tool**: Added `gen_selected_prim_ops.py` that takes comma-separated prim op names and generates a header file (`selected_prim_ops.h`) containing appropriate `#define` statements for selected ops. The tool normalizes op names to macro-safe format (e.g., `executorch_prim::et_view.default` → `INCLUDE_EXECUTORCH_PRIM_ET_VIEW_DEFAULT`).

3. **Build system integration**:
  In order to make et_operator_library also handle prim of selective build we
  make a few changes.
  1. Extract prim ops in et_operator_library
  2. Similar to gen_op_list, we invoke script that geneates selected_prim_ops.h
  file per et_operator_library target. Thus et_operator_library now generates
  selected_operators.yaml and selected_prim_ops.h.
     Note that in order to make these work we have to allow et_operator_libray to
     handle the following cases.
     1. All ops are aten ops
     2. All ops are prim ops
     3. Mix
     To do this we must make sure that the genrule continues to produce the file
     it says it will produce. In the case of 1 we have to produce empty
     selected_prim_opsh. and in case 2 we have to produce emtpy
     selected_operators.yaml
  3. In gen_all_oplist we allow for empty selected_operators.yaml and skip the
  file.
  4. Similar to gen_all_oplist we introduce another binary that combines all
  selected_prim_ops.h.
  5. Then in executorch_generated_lib we query targets from 4 that have
  selected_prim_ops and use those to compile register_prim_ops.cpp.
  In executorch_generate_lib we introduce include_all_prim_ops which by default
  is True. Hence if one wants to enable selective build for prim ops one must
  turn off that flag

## Usage:

Users can now specify prim ops like:
```
et_operator_library(name="my_aten_prim_ops", ops=["aten::mul.out", "executorch_prim::et_view.default", "aten::sym_size.int"])
executorch_generated_lib(name="my_lib", deps=[":my_aten_prim_ops"] +
other_deps, include_all_prim_ops=False)
```

Reviewed By: ivayloen, larryliu0820

Differential Revision: D81648030


larryliu0820 (approved):

PR #14345, opened 2025-09-16
Fix buck target for LLM runner util

kirklandsign:
### Summary
util.h needs constants.h and module.h

### Test plan
CI

### DIFF
--- a/extension/llm/runner/targets.bzl
+++ b/extension/llm/runner/targets.bzl
@@ -12,17 +12,6 @@ def define_common_targets():
     )
 
     runtime.cxx_library(
-        name = "stats",
-        exported_headers = [
-            "stats.h",
-            "util.h",
-        ],
-        visibility = [
-            "@EXECUTORCH_CLIENTS",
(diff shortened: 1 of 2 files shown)


lucylq (approved):

extension/llm/runner/targets.bzl
```diff
@@ -35,6 +24,21 @@ def define_common_targets():
...
+            exported_headers = [
+                "stats.h",
+                "util.h",
+            ],
+            visibility = [
+                "@EXECUTORCH_CLIENTS",
+            ],
+            exported_deps = [
+                ":constants",
+                 "//executorch/extension/module:module" + aten_suffix,
```
larryliu0820, line 38: Why does stats need module

larryliu0820: oh this target name should be renamed to `util`

larryliu0820: maybe in another PR

PR #14348, opened 2025-09-16
[Windows] Reduce trunk model test count

GregoryComer:
Reduce the number of models we test on Windows CI in order to reduce runner utilization. I've tried to pick a reasonably representative set. I don't think we're getting much incremental value from the ones we're cutting and will save on CI costs. We can re-evaluate this when we have shared/cached build for Windows CI runs.


GregoryComer:
* #14350
* __->__ #14348

mergennachin (approved):

PR #14350, opened 2025-09-16
[Windows] Reduce Windows unit test CI usage

GregoryComer:
Trim down the Windows unit test CI. We can skip the native tests in the editable job, since they don't change at all with the python install. We can also skip the python tests in the release mode job, since it currently only builds the native tests in release mode (we always build the python extension in release mode).


GregoryComer:
* __->__ #14350
* #14348

PR #14355, opened 2025-09-16
Use cuda version PT when build with CUDA delegate

Gasoonjia:
This PR enables CUDA version PT installation when building with CUDA delegate enabled from source. 

More specific:
1. ET will keep depending on cpu PT as long as CUDA delegate is not enabled;
2. We will choose the CUDA PT exactly match user's cuda version: if user don't have CUDA, or have CUDA but not exactly match the versions PT supported, the installation script will raise error.


install_requirements.py
```diff
@@ -71,7 +79,137 @@ def python_is_compatible():
...
+NIGHTLY_VERSION = "dev20250915"
+
+
+def _check_cuda_enabled():
+    """Check if CUDA delegate is enabled via CMAKE_ARGS environment variable."""
+    cmake_args = os.environ.get("CMAKE_ARGS", "")
+    return "-DEXECUTORCH_BUILD_CUDA=ON" in cmake_args
+
+
+def _cuda_version_to_pytorch_suffix(major, minor):
```
larryliu0820: Can we pull all cuda related functions into `util`?

install_requirements.py
```diff
@@ -71,7 +79,137 @@ def python_is_compatible():
...
+
+    return f"{TORCH_NIGHTLY_URL_BASE}/{cuda_suffix}"
+
+
+# url for the PyTorch ExecuTorch depending on, which will be set by _determine_torch_url().
+# please do not directly rely on it, but use _determine_torch_url() instead.
+_torch_url = ""
+
+
+def _determine_torch_url():
```
larryliu0820: This seems like it belongs to util as well

install_utils.py
```diff
@@ -0,0 +1,201 @@
...
+    cuda_suffix = _cuda_version_to_pytorch_suffix(major, minor)
+
+    return f"{torch_nightly_url_base}/{cuda_suffix}"
+
+
+# Global variable for caching torch URL
+_torch_url_cache = ""
+
+
+def determine_torch_url(torch_nightly_url_base, supported_cuda_versions):
```
JacobSzwejbka, line 106: this function is only called twice is it really necessary to cache?

Gasoonjia: main reason to cache is try not print too much noise output.

JacobSzwejbka (approved):

install_utils.py
larryliu0820: Use `@functools.lru_cache`

larryliu0820 (approved):

PR #14369, opened 2025-09-17
Arm backend: Convert remaining asserts to exceptions in tosa/

Sebastian-Larsson:
In `tosa/quant_utils.py`, add message to assert.
In `tosa/backend.py` and `tosa/mapping.py` convert asserts to exceptions.

### DIFF
--- a/backends/arm/tosa/quant_utils.py
+++ b/backends/arm/tosa/quant_utils.py
@@ -245,7 +245,9 @@ def compute_multiplier_and_shift(
         const_2_power_15_or_31 = 1 << offset
         shifted_mantissa = round(mantissa * const_2_power_15_or_31)
 
-        assert shifted_mantissa <= const_2_power_15_or_31
+        assert (
+            shifted_mantissa <= const_2_power_15_or_31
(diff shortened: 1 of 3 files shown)


backends/arm/tosa/quant_utils.py
```diff
@@ -245,7 +245,9 @@ def compute_multiplier_and_shift(
         const_2_power_15_or_31 = 1 << offset
         shifted_mantissa = round(mantissa * const_2_power_15_or_31)
 
-        assert shifted_mantissa <= const_2_power_15_or_31
+        assert (
```
digantdesai, line 248: Not raising here?

digantdesai (approved):

backends/arm/tosa/quant_utils.py
Sebastian-Larsson: No, the reason for converting to exceptions was that asserts might get optimized out. But since this error "should never happen" there's no need to change this one into an exception.

zingo:
Needs rebase :(

Sebastian-Larsson:
Unrelated CI failures

PR #14370, opened 2025-09-17
Arm backend: Remove sin_cos_support.py

Sebastian-Larsson:
The operator_support check always approved sin and cos operators, which means they can be moved to the supported operator lists in tosa_profile_supported_op_lists.py.


digantdesai (approved):

zingo:
File removed but it not present in the buck2 files

PR #14371, opened 2025-09-17
Arm backend: Replace asserts/raises with reporter rejects

Sebastian-Larsson:
- embedding_support: replace input-count assert with reporter.report_reject + return False
- index_tensor_support: add explicit rejects for None in indices, rank >= 4 indexing tensors, and int32 overflow of value tensor; previously returned False without explanation
- minmax_support: add reject when min/max.dim’s argmax output is used
- ethos_u55_support: replace IndexError raises in view/select checks (invalid dim/index) with reporter.report_reject + return False
- Improves partition diagnostics and avoids hard crashes

### DIFF
--- a/backends/arm/operator_support/embedding_support.py
+++ b/backends/arm/operator_support/embedding_support.py
@@ -27,11 +27,16 @@ class EmbeddingSupported(SupportedTOSAOperatorCheck):
     def is_node_tosa_supported(
         self, node: fx.Node, tosa_spec: TosaSpecification
     ) -> bool:  # type: ignore[override, misc]
-        # Note aten.embedding.default requires int64 indices and TOSA does not support it.
-        # Int32 indices here for aten.embedding.default is ok since it will be decomposed into ops that can handle it.
-        assert (
-            len(node.all_input_nodes) == 2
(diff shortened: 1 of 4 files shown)


backends/arm/operator_support/embedding_support.py
```diff
@@ -27,11 +27,16 @@ class EmbeddingSupported(SupportedTOSAOperatorCheck):
...
-        # Int32 indices here for aten.embedding.default is ok since it will be decomposed into ops that can handle it.
-        assert (
-            len(node.all_input_nodes) == 2
-        ), "Number of inputs to aten.embedding is not 2"
+        # Note aten.embedding.default requires int64 indices and TOSA does not
+        # support it. Int32 indices here for aten.embedding.default is ok since
+        # it will be decomposed into ops that can handle it.
+
+        if len(node.all_input_nodes) != 2:
+            self.reporter.report_reject(
```
digantdesai, line 35: this is cool!

digantdesai (approved):

backends/arm/operator_support/embedding_support.py
Sebastian-Larsson: :D

PR #14373, opened 2025-09-17
NXP backend: Add infrastructure for context dependant partitioning

MartinPavella:
### Summary
This PR adds the option to specify delegation conditions which depend on the partition a particular node ends up in. This infrastructure is applied to the `view_copy` node.

### Test plan
Unit tests provided.

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
@@ -14,6 +14,7 @@ from executorch.backends.nxp.backend.ir.converter import quantization_utils
 from executorch.backends.nxp.backend.ir.converter.conversion.common import OpsList
 from executorch.backends.nxp.backend.ir.converter.node_converter import (
     CustomDelegationOptions,
+    is_not_qdq_node,
     NodeConverter,
 )
 from executorch.backends.nxp.backend.ir.converter.node_converters.shared.reshape_transposition import (
@@ -23,6 +24,7 @@ from executorch.backends.nxp.backend.ir.tflite_generator.builtin_options import 
     reshape_options,
 )
 from torch.fx import Node
+from torch.fx.passes.infra.partitioner import Partition
 from torch.nn import Parameter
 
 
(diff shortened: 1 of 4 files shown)


jirioc (approved):
Now that the unit tests restore the original state of supports_partitioning_result, the PR looks OK.

MartinPavella:
The failing test `pull / test-samsung-models-linux / linux-job (pull_request)` fails on the command `mkdir -p /tmp/exynos_ai_lite_core`, which seems unrelated to this PR.
All other failing tests seem unrelated to this PR, and are also failing on `main`.

robert-kalmar (approved):

backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
```diff
@@ -45,6 +47,27 @@ def _is_supported_in_IR(
...
+    ):
+        view_copy_partitions = [
+            partition for partition in partition_list if node in partition.nodes
+        ]
+        assert len(view_copy_partitions) == 1
+        non_q_dq_partition_nodes = list(
+            filter(is_not_qdq_node, view_copy_partitions[0].nodes)
+        )
+
+        if len(non_q_dq_partition_nodes) == 1:
```
robert-kalmar, line 65: For any practical model the contion is enough, but in theory not right. 
Thinking loudly: 
- The view_copy (==> Reshape in Neutron IR) cannot be alone in the partition because on Neutron it is a NOOP. Regardless of the count (2 view_copies is also not allowed) 
- Another operator is clone, as we convert it to an "identity" tensor, what will be NOOP in neutron too. 
So I would say, the correct condition is there must be at least 1 compute operator, that is the partition must not contain only view_copy and clone ops (any number).

MartinPavella: Good catch.
I will create a ticket for this, as it is not trivial to determine which nodes will end up being "noops". And it would require [another PR](https://github.com/pytorch/executorch/pull/14576) to be merged first.

PR #14377, opened 2025-09-17
Add support for non tensor inputs to portable executor runner

mansnils:
Additionally:
* Fix issue in input file handling where vector reallocations could cause input_buffers pointers to point to garbage.
* Enable pytest_sum_vgf pytest in Arm backend that need this support.


examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -175,21 +175,33 @@ int main(int argc, char** argv) {
...
-  std::string token;
+  std::string path;
+
+  // First reserve memory for number of vector elements to avoid vector
+  // reallocations when emplacing back.
+  std::vector<std::string> file_paths;
+  while (std::getline(list_of_input_files, path, ',')) {
+    file_paths.push_back(std::move(path));
+  }
+  inputs_storage.reserve(file_paths.size());
```
digantdesai, line 186: +1

extension/runner_util/inputs.cpp
```diff
@@ -78,7 +78,40 @@ Result<BufferCleanup> prepare_input_tensors(
...
     if (tag.get() != Tag::Tensor) {
-      ET_LOG(Debug, "Skipping non-tensor input %zu", i);
+      if (!hard_code_inputs_to_ones) {
+        Error err = Error::Ok;
+        auto [buffer, buffer_size] = input_buffers.at(i);
+
+        ET_LOG(
+            Debug, "Verifying and setting input for non-tensor input %zu", i);
+
+        if (tag.get() == Tag::Int) {
```
digantdesai, line 88: nits:
why not switch-case?
Does it make sense to cover all possible tag values - see `EXECUTORCH_FORALL_TAGS`?

extension/runner_util/inputs.cpp
```diff
@@ -78,7 +78,40 @@ Result<BufferCleanup> prepare_input_tensors(
       continue;
     }
     if (tag.get() != Tag::Tensor) {
-      ET_LOG(Debug, "Skipping non-tensor input %zu", i);
+      if (!hard_code_inputs_to_ones) {
+        Error err = Error::Ok;
```
digantdesai, line 82: do you want to add tests for this? Perhaps here - extension/runner_util/test/inputs_test.cpp

backends/arm/test/ops/test_acos.py
```diff
@@ -105,10 +104,7 @@ def test_acos_vgf_FP(test_data: Tuple):
         tosa_version="TOSA-1.0+FP",
         run_on_vulkan_runtime=True,
     )
-    try:
-        pipeline.run()
-    except FileNotFoundError as e:
-        pytest.skip(f"VKML executor_runner not found - not built - skip {e}")
+    pipeline.run()
```
digantdesai, line 107: enabled by this PR?

digantdesai (approved):

backends/arm/test/ops/test_acos.py
mansnils: no this was just moving the try/except to a central location

mansnils:
@digantdesai thanks for your comments - agree that all of them would be nice to have/improvements. In any case I prefer that in a separate patch(es)

digantdesai:
> In any case I prefer that in a separate patch(es)

Sounds good, your call. Feel free to create an issue - some oss contributors can also do it for you :)

PR #14379, opened 2025-09-17
Fix eval_llama_qnn script

rohansjoshi:
Fix eval_llama_qnn.py after recent changes and use eval utils


cccclai (approved):
Thanks for the fix!

PR #14394, opened 2025-09-17
Fix message truncating logic to respect UTF8 encoding.

shoumikhin:
Summary: .

Differential Revision: D82681904

### DIFF
--- a/runtime/platform/log.cpp
+++ b/runtime/platform/log.cpp
@@ -59,6 +59,38 @@ static_assert(
     kLevelToPal[size_t(LogLevel::Fatal)] == et_pal_log_level_t::kFatal,
     "");
 
+#if ET_LOG_ENABLED
+static size_t get_valid_utf8_prefix_length(const char* bytes, size_t length) {
+  if (!bytes || length == 0) {
+    return 0;
+  }
+  const auto* data = reinterpret_cast<const unsigned char*>(bytes);
+  size_t i = length;
+  while (i > 0 && (data[i - 1] & 0xC0) == 0x80) {
+    --i;
+  }
+  if (i == 0) {
+    return 0;
+  }
+  const size_t lead_pos = i - 1;
+  const unsigned char lead = data[lead_pos];
+  size_t need = 0;
+
+  if (lead < 0x80) {
+    need = 1;
+  } else if ((lead & 0xE0) == 0xC0) {
+    need = 2;
+  } else if ((lead & 0xF0) == 0xE0) {
+    need = 3;
+  } else if ((lead & 0xF8) == 0xF0) {
+    need = 4;
+  } else {
+    return lead_pos;
+  }
+  return length - lead_pos == need ? length : lead_pos;
+}
+#endif // ET_LOG_ENABLED
+
 /**
  * Log a string message.
  *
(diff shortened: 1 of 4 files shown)


runtime/platform/log.cpp
```diff
@@ -84,20 +120,24 @@ void vlogf(
...
+  char buffer[kMaxLogMessageLength];
+
+  const auto write_count =
+      vsnprintf(buffer, kMaxLogMessageLength, format, args);
+  const size_t used_length = (write_count < 0)
+      ? 0
+      : (write_count >= static_cast<int>(kMaxLogMessageLength)
+             ? kMaxLogMessageLength - 1
+             : static_cast<size_t>(write_count));
+  const auto valid_length = get_valid_utf8_prefix_length(buffer, used_length);
```
swolchok, line 128: Supposing that the buffer was valid UTF-8 except for possible truncation, you don't have to scan it from the beginning to detect truncation in the middle of a UTF-8 code point. Looking at https://en.wikipedia.org/wiki/UTF-8#Description :

- If the new last byte doesn't have the high bit set, it's valid.
- Otherwise, scan backward from the end until you find a first byte of a code point (highest and second highest bit set). Look its high bits (to determine how many continuation bytes should be there) and at how many bytes are in the buffer after it; either you have all of its continuation bytes (nothing to drop) or you don't (drop the partial code point).

swolchok (changes requested):

runtime/platform/log.cpp
```diff
@@ -59,6 +59,40 @@ static_assert(
     kLevelToPal[size_t(LogLevel::Fatal)] == et_pal_log_level_t::kFatal,
     "");
 
+#if ET_LOG_ENABLED
+static inline size_t get_valid_utf8_prefix_length(
```
swolchok: why does this need to be `inline`?

runtime/platform/test/logging_test.cpp
```diff
@@ -10,24 +10,62 @@
...
+  InterceptWith iw(spy);
+
   ET_LOG(Info, "Sample log with integer: %u", 100);
+  EXPECT_EQ(spy.last_log_message_args.message, "Sample log with integer: 100");
+}
+
+TEST_F(LoggingTest, Utf8Truncation) {
+  PalSpy spy;
+  InterceptWith iw(spy);
+
```
swolchok, line 62: What about cutting off the second byte of a codepoint with a two-byte UTF-8 encoding, like https://codepoints.net/U+00C3 ?

swolchok: might also be prudent to test with the prefix filled with multi-byte codepoints (say, 0-2 copies of A followed by 63 instances of \xF0\x9F\x91\x8D) to make sure the truncation logic isn't in any way confused by that. For example:

```
std::string get_prefix(std::size_t length, bool use_multibyte) {
  if (!use_multibyte) {
    return std::string(length, 'A');
  }
  std::ostringstream result;
  result << std::string(length % 4, 'A');
  std::size_t rest = length - length % 4;
  while (rest > 0) {
    result << "\xF0\x9F\x91\x8D";
    rest -= 4;
  }
  return result.str();
}
```

and then just loop your tests:

```
for (const auto& prefix : {get_prefix(253, false), get_prefix(253, true)}) { ... }
```

you could even have the tests loop over your three codepoints to truncate in the middle, calculate the proper prefix length, etc. so you don't have to repeat code.

swolchok (approved):
looks reasonable, please at least test truncating in the middle of a 2-byte codepoint

PR #14396, opened 2025-09-17
Fix regnet on XNNPACK

GregoryComer:
### Summary
The regnet_y_32gf model from torchvision does not lower on XNNPACK. This is because a "-" in a generated weight name (from conv + bn fusion) causes things to blow up. I've updated the fusion logic to properly sanitize the generated weight name and added a test for regnet to validate the fix.

Here's the part of the eager mode model responsible for this:
```
(trunk_output): Sequential(
    (block1): AnyStage(
      (block1-0): ResBottleneckBlock(
        (proj): Conv2dNormActivation(
          (0): Conv2d(32, 232, kernel_size=(1, 1), stride=(2, 2), bias=False)
          (1): BatchNorm2d(232, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
        )
```

cc @digantdesai @mcr229 @cbilgin

### DIFF
--- a/backends/xnnpack/_passes/fuse_batch_norm.py
+++ b/backends/xnnpack/_passes/fuse_batch_norm.py
@@ -18,6 +18,7 @@ from executorch.backends.xnnpack.utils.utils import (
     get_param_tensor,
     get_tensor_name,
     is_param_node,
+    sanitize_node_name,
 )
 from executorch.exir import ExportedProgram
 from executorch.exir.dialects._ops import ops as exir_ops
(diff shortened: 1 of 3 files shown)


backends/xnnpack/utils/utils.py
```diff
@@ -219,3 +219,15 @@ def is_depthwise_conv(
...
+def sanitize_node_name(name: str) -> str:
+    """
+    Modify a (generated) node name to replace invalid characters (. and -) with underscores.
+    """
+    INVALID_CHARS = [".", "-"]
+
+    sanitized = name
+    for c in INVALID_CHARS:
+        sanitized = sanitized.replace(c, "_")
+    return sanitized
```
mergennachin, line 233: This is more performant

```
return name.translate(str.maketrans('.-', '__'))
```

mergennachin (approved):

digantdesai:
> causes things to blow up

Can you put the error log? I am trying to figure out what's the exact issue..

backends/xnnpack/_passes/fuse_batch_norm.py
```diff
@@ -208,13 +209,13 @@ def _fuse_ops(
...
+        fused_weight_name = sanitize_node_name(input_node_weight_name + "_fused_bn")
         if input_node_bias_name == "":
-            fused_bias_name = (input_node_weight_name + "_bias_fused_bn").replace(
-                ".", "_"
+            fused_bias_name = sanitize_node_name(
+                input_node_weight_name + "_bias_fused_bn"
             )
         else:
-            fused_bias_name = (input_node_bias_name + "_fused_bn").replace(".", "_")
+            fused_bias_name = sanitize_node_name(input_node_bias_name + "_fused_bn")
```
digantdesai, line 218: Seems like we are sensitive to names, we should put assert somewhere downstream where we check not INVALID_CHARS sneaks in..

PR #14401, opened 2025-09-18
Arm Backend: Expose PMU trace output from FVP run

jmahbs:
Exposes PMU trace output from an FVP. This lays part of the foundation to enable us to use this output to as a data overlay in Model Explorer visualisations.

The end goal here is to be able to visualise some profiling data in Model Explorer using our Tosa Flatbuffer adapter. To enable this we need to implement a few changes:

1. Expose PMU trace output from a FVP. This gives us performance data from an FVP run. (This PR)
2. Expose Vela's debug database. This gives us generic information on operators in a our model, and can be combined with the trace output to provide more detailed profiling analysis
3. Write a script to combine the trace output and the debug database so we can visualise it in Model Explorer in Executorch. 

Here's a snippet of the PMU trace output: 
```
{
            "name": "axi_enabled_cycles",
            "ph": "X",
            "ts": "1029",
            "pid": "DMA",
            "tid": "axi_enabled_cycles",
            "dur": "1014"
        }
```

### DIFF
--- a/backends/arm/scripts/run_fvp.sh
+++ b/backends/arm/scripts/run_fvp.sh
@@ -22,6 +22,7 @@ data_file=""
 target="ethos-u55-128"
 timeout="600"
 etrecord_file=""
+trace_file=""
 
 help() {
     echo "Usage: $(basename $0) [options]"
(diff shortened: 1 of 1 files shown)


digantdesai:
Hi @jmahbs thanks for the PR. Can you add more details (perhaps a snippet of output) on what's in this trace file and how this will be consumed downstream? Thanks.

digantdesai (approved):

digantdesai:
FYI https://docs.pytorch.org/executorch/main/etdump.html

This is going to be fun! :)

PR #14415, opened 2025-09-18
[ET-VK] Add Fusing for Conv/Binary Ops, Clamp/Binary Ops, and Clamp/Clamp

alexdean08:
With the motivation of improving performance, this change adds the functionality for fusing the following ops:
- conv2d PW s1p0 and binary ops (add, sub, mul, div)
- clamp and binary ops (add, sub, mul, div)
- clamp and clamp


backends/transforms/fuse_clamp_with_binary_op.py
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
```
SS-JIA, line 7: @alexdean08 would you mind taking a stab at implementing the fusions via the new `patterns/` folder I added? This folder is meant to abstract matching fusable operator patterns and fusing them into a single op.

Here is an example for quantized convolution: https://github.com/pytorch/executorch/blob/main/backends/vulkan/patterns/quantized_convolution.py

Basically, you will need to implement 3 things in a file called, say `patterns/fused_convolution.py`

```python
# patterns/fused_convolution.py

FUSEABLE_OPS = [
    exir_ops.edge.aten.relu.default,
    exir_ops.edge.aten.hardtanh.default,
    exir_ops.edge.aten.clamp.default,
]
FUSEABLE_BINARY_OPS = [
    exir_ops.edge.aten.add.Tensor,
    exir_ops.edge.aten.sub.Tensor,
    exir_ops.edge.aten.mul.Tensor,
    exir_ops.edge.aten.div.Tensor,
]

# Implement a Match class to represent the patterns involved in a pattern match
class FusedConvolutionMatch(PatternMatch):
    def __init__(self, conv_node: torch.fx.Node) -> None:
        self.anchor_node = conv_node
        # Store all relevant args here
        ...

        first_user = list(conv_node.users)[0]

        if first_user not in FUSABLE_OPS:
            return
        
        self.output_node = first_user
        # Store other relevant args here
        ...


        self.match_found = True


convolution_anchor_nodes = {
    exir_ops.edge.aten.conv2d.default,
    exir_ops.edge.aten.convolution.default,
}


@register_pattern_detector("fused_convolution")
def find_fused_convolution_patterns(
    node: torch.fx.Node,
) -> Optional[FusedConvolutionMatch]:
    if node.target not in convolution_anchor_nodes:
        return None

    matched_pattern = FusedConvolutionMatch(node)
    if matched_pattern.match_found:
        return matched_pattern

    return None


##
## Pattern Replacement
##

FUSED_NODE_TARGET_TO_CUSTOM_OP_TARGET = {
  exir_ops.edge.aten.add.Tensor: exir_ops.edge.et_vk.conv_with_binary_add.default
  ...
}

@register_pattern_replacement("fused_convolution")
def make_fused_conv2d_custom_op(
    ep: ExportedProgram,
    graph_module: torch.fx.GraphModule,
    match: QuantizedConvolutionMatch,
):
    with graph_module.graph.inserting_before(match.output_node):
        op_target = FUSED_NODE_TARGET_TO_CUSTOM_OP_TARGET[match.output_node.target]

        custom_op_node = graph_module.graph.create_node(
            "call_function",
            op_target,
            args=(
                ...
            ),
        )

    qconv_node.meta["val"] = match.output_node.meta["val"]
    match.output_node.replace_all_uses_with(custom_op_node)
```

This should simplify the logic a bit and allow us to reduce the amount of transforms we need to maintain. Although I think `fuse_clamps.py` still needs to be its own pass.

Once you've added the pattern file, remember to import it here: https://github.com/pytorch/executorch/blob/main/backends/vulkan/patterns/__init__.py so that registrations are triggered when importing the module.

SS-JIA: Also for fusing clamps with binaries, how impactful is it? If the impact is not super significant, I think I would prefer to avoid fusing for the sake of simplifying our codebase.

SS-JIA (changes requested):
Runtime side LGTM. Just comments about how op fusions are implemented. Thanks for working on this!

SS-JIA (approved):
LGTM. All tests are passing.

PR #14429, opened 2025-09-19
Add prefill API to MultimodalRunner

rohansjoshi:
Add prefill_inputs function to MultimodalRunner, this is useful for example to prefill chat history


extension/llm/runner/multimodal_runner.cpp
```diff
@@ -62,6 +62,16 @@ Error MultimodalRunner::load() {
     ET_LOG(Info, format, __VA_ARGS__);     \
   }
 
+Error MultimodalRunner::prefill_inputs(std::vector<MultimodalInput>& inputs) {
```
larryliu0820: Just `prefill` is enough

larryliu0820 (approved):

PR #14435, opened 2025-09-19
Make _get_program_from_buffer work for bundled programs

Erik-Lundell:
Add some cmake to only do this if executorch is built with bundleio.

codegen/tools subdirectory include needs to be moved in top-level CmakeLists.txt
to have access to the bundled_program target.

Follow-up patch to enable the fix in arm backend.


Gasoonjia (approved):
LGTM!

PR #14438, opened 2025-09-19
Aoti cuda export support

Gasoonjia:
This PR introduces the export support for cuda delegate using aoti library. Also create ci test for verification.


.ci/scripts/test_cuda_export_aoti.py
```diff
@@ -0,0 +1,228 @@
...
+            cwd=cwd,
+            capture_output=True,
+            text=True,
+            timeout=timeout,
+            check=False,  # We'll handle the return code ourselves
+        )
+
+        if result.stdout:
+            print("STDOUT:")
+            print(result.stdout)
```
larryliu0820, line 41: Please use `logging` it provides different levels of logging

backends/cuda/cuda_partitioner.py
```diff
@@ -0,0 +1,61 @@
...
+    DelegationSpec,
+    Partitioner,
+    PartitionResult,
+)
+from executorch.exir.backend.utils import tag_constant_data
+from torch.export.exported_program import ExportedProgram
+
+
+@final
+class CudaPartitioner(Partitioner):
```
JacobSzwejbka, line 24: so this is basically just skeleton code? We could skip having an initial partitioner implementation entirely I think and just only allow the other to_backend api for now? Is that how it works?

exir/backend/backend_api.py
```diff
@@ -720,6 +726,8 @@ def to_backend(
             fake_edge_program = copy.deepcopy(edge_program)
         partitioner_result = partitioner_instance(fake_edge_program)
         tagged_exported_program = partitioner_result.tagged_exported_program
+        tagged_exported_program.example_inputs = edge_program.example_inputs
```
JacobSzwejbka, line 729: Are we serializing the example inputs?

.ci/scripts/test_cuda_export_aoti.py
```diff
@@ -0,0 +1,228 @@
...
+    for pattern in cleanup_patterns:
+        try:
+            import glob
+
+            files = glob.glob(pattern)
+            for file in files:
+                if os.path.isfile(file):
+                    os.remove(file)
+                    print(f"Removed file: {file}")
+        except Exception as e:
```
JacobSzwejbka, line 116: these should be fatal exceptions right? The test should fail?

.gitignore
```diff
@@ -34,6 +34,14 @@ tokenizer.json
 !test_bpe_tokenizer.bin
 !test_tiktoken_tokenizer.model
 
+# AOTI temporary files
+*.cubin
+*kernel_metadata.json
+*kernel.cpp
+*wrapper_metadata.json
+*wrapper.cpp
```
JacobSzwejbka, line 42: does wrapper .cpp stick around? AOTI compilation doesnt clean it up after generating the .so and .cubin?

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,116 @@
...
+    ExportedProgram,
+    PreprocessResult,
+)
+from executorch.exir.backend.compile_spec_schema import CompileSpec
+from torch._inductor.codegen.cpp_wrapper_cpu import CppWrapperCpu
+from torch.export.passes import move_to_device_pass
+
+
+# exist fallback operators in et namespace;
+supported_fallback_kernels: Dict[str, Any] = {}
```
JacobSzwejbka, line 28: @larryliu0820 I feel like I keep hearing conflicting information. Is AOTI falling back to ET or is it a graph break. Graph break sounds more natural in the ET ecosystem to me

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Move the edge_program from CPU to CUDA for aoti compile
+        cuda_edge_program = move_to_device_pass(edge_program, "cuda")
+
+        edge_program_module = cuda_edge_program.module()
+        args, kwargs = cuda_edge_program.example_inputs
+
+        output_path = os.path.join(os.getcwd(), "aoti.so")
+
+        options: dict[str, typing.Any] = {
```
JacobSzwejbka, line 86: Can you add some documentation on what these options are or where they are defined? 

debug_compile
embed_kernel_binary

are the two non obvious to me

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,116 @@
...
+            "aot_inductor.embed_kernel_binary": True,
+            "aot_inductor.link_libtorch": False,
+            "aot_inductor.package_constants_in_so": True,
+            "aot_inductor.output_path": output_path,
+            "aot_inductor.debug_compile": True,
+            "aot_inductor.force_mmap_weights": False,
+            "max_autotune": True,
```
JacobSzwejbka, line 93: How are we autotuning? We dont know what gpu we are running on?

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,116 @@
...
+                formatted_kernels = "\n  - ".join(sorted(missing_fallback_kernels))
+                raise RuntimeError(
+                    f"Missing fallback kernels ({len(missing_fallback_kernels)} total):\n  - {formatted_kernels}\n"
+                    "Please add them to the AOTI backend."
+                )
+
+        with open(so_path, "rb") as f:
+            so_data = f.read()
+
+        named_data_store.add_named_data("so_blob", so_data, 1, "aoti_cuda_blob")
```
JacobSzwejbka, line 110: where do you put the cubin?

.gitignore
Gasoonjia: right now it will generating lots of files, including cubin, wrapper.cpp,etc, and will be auto cleaned up.

backends/cuda/cuda_backend.py
Gasoonjia: Here we will leverage AOTI fallback and don't break graph, for working on missing operators, which is more natural for aoti and convient for us.

backends/cuda/cuda_backend.py
Gasoonjia: Maybe @yushangdi  can answer this question better. My understanding is it will autotune for the GPU we have during aoti compile. it can get the info automatically.

backends/cuda/cuda_backend.py
Gasoonjia: no cubin is explicitly needed. All stuffs has been in .so

exir/backend/backend_api.py
Gasoonjia: serializing here you mean exir.save, or to pte?

backends/cuda/cuda_partitioner.py
Gasoonjia: We should keep it; we need this partitioner to skip et operator decomposition for all ops.

exir/lowered_backend_module.py
```diff
@@ -720,6 +721,11 @@ def create_exported_program_from_submodule(
     in_spec = pytree.tree_flatten((tuple(subgraph_signature.user_inputs), {}))[1]
```
Gasoonjia, line 721: todo: add a check for examing the input signature of first partition and original edge program

backends/cuda/cuda_backend.py
yushangdi: yeah that's correct.

backends/cuda/cuda_backend.py
yushangdi: yeah `"aot_inductor.embed_kernel_binary": True,` puts the kernel in .so

backends/cuda/cuda_backend.py
larryliu0820: > Graph break sounds more natural in the ET ecosystem to me

Yes I'm thinking we want to have some cuda kernels and let it graph break. We can also reuse that kernel in AOTI fallback.

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
larryliu0820, line 1: Let's add some unit tests under `backends/cuda/test`

exir/lowered_backend_module.py
```diff
@@ -720,6 +721,11 @@ def create_exported_program_from_submodule(
...
     out_spec = pytree.tree_flatten(subgraph_signature.user_outputs)[1]
 
+    # only the example inputs of first parition equals to the example inputs of the owning program
+    submodule_exmaple_inputs = (
+        owning_program.example_inputs if is_first_partition else None
+    )
```
larryliu0820, line 727: This is not enough, we need to make sure the signature of the first partition is the same as the signature of the original exported program.

Issue #14443, opened 2025-09-19
Pybindings can't load on Windows when installing wheel


GregoryComer:
When installing ET 1.0 from the wheel, pybindings fail with a DLL load error. This is likely the same underlying issue that we solved for Mac + Linux in https://github.com/pytorch/executorch/pull/14422.

```
>>> from executorch.extension.pybindings._portable_lib import _load_for_executorch
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: DLL load failed while importing _portable_lib: The specified module could not be found.
```

cc @larryliu0820

GregoryComer:
Looks like the recommended way to handle this on Windows is to call os.add_dll_directory in __init__.py before attempting to load the dll.

GregoryComer:
PR and pick are merged.

PR #14446, opened 2025-09-19
Add pybind extension to the DLL search path on Windows

GregoryComer:
### Summary
When installing ExecuTorch from a wheel on Windows, the native pybinding extension DLL fails to load (https://github.com/pytorch/executorch/issues/14443). This is because it's not on the search path. From a quick Google search, it seems like the recommended way to handle this is to call `os.add_dll_directory` prior to loading the extension.

I tested this locally by installing from wheel in a clean env and then patching the installed copy in site-packages. With this change, I was able to import executorch.extension.pybindings.portable_lib. I also ran the XNNPACK add op tests as a sanity check, which use pybindings.

### DIFF
--- a/extension/pybindings/portable_lib.py
+++ b/extension/pybindings/portable_lib.py
@@ -13,6 +13,9 @@
     This API is experimental and subject to change without notice.
 """
 
+import logging
+import os
+import sys
 import warnings as _warnings
 
 import executorch.exir._warnings as _exir_warnings
@@ -27,6 +30,21 @@ _warnings.warn(
 # related libs, ensuring that the pybindings lib can resolve those runtime
 # dependencies.
 import torch as _torch
+
+logger = logging.getLogger(__name__)
+
+# Update the DLL search path on Windows. This is the recommended way to handle native
+# extensions.
+if sys.platform == "win32":
+    try:
+        # The extension DLL should be in the same directory as this file.
+        pybindings_dir = os.path.dirname(os.path.abspath(__file__))
+        os.add_dll_directory(pybindings_dir)
+    except Exception as e:
+        logger.error(
+            "Failed to add the pybinding extension DLL to the search path. The extension may not work.",
+            e,
+        )
 
 # Let users import everything from the C++ _portable_lib extension as if this
 # python file defined them. Although we could import these dynamically, it


extension/pybindings/portable_lib.py
```diff
@@ -28,6 +30,13 @@
 # dependencies.
 import torch as _torch
 
+# Update the DLL search path on Windows. This is the recommended way to handle native
+# extensions.
+if sys.platform == "win32":
+    # The extension DLL should be in the same directory as this file.
+    pybindings_dir = os.path.dirname(__file__)
```
mergennachin: Need absolute path

```
pybindings_dir = os.path.dirname(os.path.abspath(__file__))
```

extension/pybindings/portable_lib.py
```diff
@@ -28,6 +30,13 @@
 # dependencies.
 import torch as _torch
 
+# Update the DLL search path on Windows. This is the recommended way to handle native
+# extensions.
+if sys.platform == "win32":
+    # The extension DLL should be in the same directory as this file.
```
mergennachin: try:
              pybindings_dir = os.path.dirname(os.path.abspath(__file__))
              if pybindings_dir and os.path.isdir(pybindings_dir):
                  os.add_dll_directory(pybindings_dir)
          except Exception:
              pass

extension/pybindings/portable_lib.py
```diff
@@ -28,6 +30,13 @@
 # dependencies.
 import torch as _torch
 
+# Update the DLL search path on Windows. This is the recommended way to handle native
+# extensions.
+if sys.platform == "win32":
```
mergennachin, line 38:
```
if sys.platform.startswith("win"):
```

so that it can catch "win64" and "windows"

GregoryComer: Win32 is actually the stable value for sys.platform for both 32-bit and 64-bit Windows builds. I'm happy to change it, though it seems like direct comparison to "win32" is the idiomatic appoach.

extension/pybindings/portable_lib.py
GregoryComer: I'm a little hesitant to swallow the exception here, since it would be indicative of something being pretty wrong. I'll at least add a warning in this case.

extension/pybindings/portable_lib.py
GregoryComer: Updated.

extension/pybindings/portable_lib.py
GregoryComer: Update to log the error message and attempt to continue.

mergennachin (approved):

PR #14447, opened 2025-09-19
equip the first partitioned graph with same example inputs as original graph

Gasoonjia:


JacobSzwejbka:
Can you explain why we need to preserve the example inputs after partitioning @Gasoonjia

PR #14448, opened 2025-09-19
Use weight cache for quantized tensor scale data

GregoryComer:
Summary:
When enabling the XNNPACK weight cache and running a model with qb4 or qc8-quantized linear weights, it triggers an assertion that is intended to make sure all data is in the weight cache. This can be reproduced by running the XNNPACK backend linear op tests with weight cache enabled.

The root cause appears to be that tensor scale data is bypassing the weight cache - likely an oversight. This isn't a correctness issue, but does cause the aforementioned assert to fail and uses marginally more memory than it otherwise needs to.

This PR updates the XNNPACK compileModel call to use the weight cache for scale data (instead of putting it in the unpacked_buffers list). With this change, the linear op tests pass with weight cache enabled.

Differential Revision: D82862629

### DIFF
--- a/backends/xnnpack/runtime/XNNCompiler.cpp
+++ b/backends/xnnpack/runtime/XNNCompiler.cpp
@@ -174,13 +174,12 @@ payload (deprecated) or via offsets to the constant_data_ptr. If no constant
 data associated with the tensor value, then returns nullptr.
 */
 const uint8_t* getConstantDataPtr(
-    const fb_xnnpack::XNNTensorValue* tensor_value,
+    uint32_t buffer_idx,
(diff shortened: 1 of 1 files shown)


GregoryComer:
Adding to 1.0 release blockers, as weight cache is enabled by default in several presets (including Android). QC8 and QB4 linear is broken when it's enabled, so we'll want to cherry pick this fix.

digantdesai (approved):

digantdesai (approved):
can we also run test with a draft PR to see if we enable this by default what fails if anything..

lucylq (approved):

PR #14449, opened 2025-09-19
Fix outdated lintrunner directions

swolchok:
They had the wrong version, but also should just point people toward the script that will install the right version.


swolchok:
* __->__ #14449

mergennachin (approved):

PR #14456, opened 2025-09-22
[Backend Tester] Migrate to pytest

GregoryComer:
Refactor the backend test suites to use pytest. This includes the following changes:
* Define pytest markers for each backend and test flow (recipe). This allows for easy filter, such as by running `pytest some/path/... -m backend_xnnpack`.
* Use a parameterized pytest fixture to handle test generation / expansion for each test flow.
* Switch to using the pytest-json-report plugin for reporting. Update the markdown generation script to take json.
* Shim the existing unittest-based logic for op tests.
  * I've updated add.py to show what they should look like long-term. I've also just updated the model tests, since there aren't as many. I'll update the remaining op tests later in this stack, though this is purely to clean up the code. The shimming logic makes them work properly with pytest in this PR.
 * Update the backend test CI to use pytest.

This also has the benefit of making the jobs much faster by leveraging parallel execution. I've also added a repro command to the markdown summary.


GregoryComer:
* __->__ #14456

backends/test/suite/__init__.py
```diff
@@ -55,6 +56,11 @@ def get_test_flows() -> dict[str, TestFlow]:
     return _ALL_TEST_FLOWS
 
 
+def dtype_to_str(dtype: torch.dtype) -> str:
```
GregoryComer, line 59: This utility function is used for generating the display name for parameterized tests.

backends/test/suite/operators/__init__.py
```diff
@@ -66,112 +61,46 @@ def dtype_test(func):
...
-    def wrapped_test(self):
-        with TestContext(test_name, test_base_name, flow.name, params):
-            if flow.should_skip_test(test_name):
-                raise unittest.SkipTest(
-                    f"Skipping test due to matching flow {flow.name} skip patterns"
-                )
 
-            test_kwargs = copy.copy(params) or {}
-            test_kwargs["flow"] = flow
+def wrap_test(original_func, test_type):
```
GregoryComer, line 76: This is shim code to make the existing op tests work with the new pytest setup. It will be removed once I update all of the op tests. Note that this is not a functional change, just code cleanup. It's using the proper pytest fixtures and parameterization with this PR, just through this shim instead of directly.

.ci/scripts/test_backend_linux.sh
```diff
@@ -54,7 +54,7 @@ fi
 PYTHON_EXECUTABLE=python CMAKE_ARGS="$EXTRA_BUILD_ARGS" .ci/scripts/setup-linux.sh --build-tool cmake --build-mode Release --editable true
 
 EXIT_CODE=0
-python -m executorch.backends.test.suite.runner $SUITE --flow $FLOW --report "$REPORT_FILE" || EXIT_CODE=$?
+pytest -c /dev/nul -n auto backends/test/suite/$SUITE/ -m flow_$FLOW --json-report --json-report-file="$REPORT_FILE" || EXIT_CODE=$?
```
digantdesai: FYI we don't exactly love pytest when not working in OSS.

.ci/scripts/test_backend_macos.sh
```diff
@@ -24,7 +24,7 @@ PYTHON_EXECUTABLE=python
 ${CONDA_RUN} --no-capture-output .ci/scripts/setup-macos.sh --build-tool cmake --build-mode Release
 
 EXIT_CODE=0
-${CONDA_RUN} --no-capture-output python -m executorch.backends.test.suite.runner $SUITE --flow $FLOW --report "$REPORT_FILE" || EXIT_CODE=$?
+${CONDA_RUN} python -m pytest -c /dev/nul -n auto backends/test/suite/$SUITE/ -m flow_$FLOW --json-report --json-report-file="$REPORT_FILE" || EXIT_CODE=$?
 
 # Generate markdown summary.
-${CONDA_RUN} --no-capture-output python -m executorch.backends.test.suite.generate_markdown_summary "$REPORT_FILE" > ${GITHUB_STEP_SUMMARY:-"step_summary.md"} --exit-code $EXIT_CODE
+${CONDA_RUN} python -m executorch.backends.test.suite.generate_markdown_summary_json "$REPORT_FILE" > ${GITHUB_STEP_SUMMARY:-"step_summary.md"} --exit-code $EXIT_CODE
```
digantdesai: for uniform experience and output structure, you can use bash functions and share them across these scripts.

GregoryComer: Updated to unify the scripts.

.ci/scripts/test_backend_linux.sh
GregoryComer: Do you have concerns with using pytest? Or suggested changes?

backends/test/suite/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+        for f in all_flows().values()
+    ],
+    ids=str,
+)
+def test_runner(request):
+    return TestRunner(request.param, request.node.name, request.node.originalname)
+
+
+@pytest.hookimpl(optionalhook=True)
+def pytest_json_runtest_metadata(item, call):
```
digantdesai, line 86: For `pytest-json-report` I assume?

.github/workflows/_test_backend.yml
```diff
@@ -81,4 +81,4 @@ jobs:
         # This is needed to get the prebuilt PyTorch wheel from S3
         ${CONDA_RUN} --no-capture-output pip install awscli==1.37.21
 
-        source .ci/scripts/test_backend_macos.sh "${{ matrix.suite }}" "${{ matrix.flow }}" "${RUNNER_ARTIFACT_DIR}"
+        source .ci/scripts/test_backend.sh "${{ matrix.suite }}" "${{ matrix.flow }}" "${RUNNER_ARTIFACT_DIR}"
```
digantdesai, line 84: +1

digantdesai (approved):
Thanks!

PR #14462, opened 2025-09-22
Arm backend: Enable run on vulkan runtime by default and add xfails

mansnils:
Enables runing on vulkan runtime by default and add xfails for those tests that can not yet run.

cc @SS-JIA @manuelcandales @cbilgin @digantdesai @freddan80 @per @zingo @oscarandersson8218

### DIFF
--- a/backends/arm/test/tester/test_pipeline.py
+++ b/backends/arm/test/tester/test_pipeline.py
@@ -922,7 +922,7 @@ class VgfPipeline(BasePipelineMaker, Generic[T]):
         test_data: T,
         aten_op: str | List[str],
         exir_op: Optional[str | List[str]] = None,
-        run_on_vulkan_runtime: bool = False,
+        run_on_vulkan_runtime: bool = True,
         vgf_compiler_flags: Optional[str] = "",
         tosa_version: str = "TOSA-1.0+FP",
         symmetric_io_quantization: bool = False,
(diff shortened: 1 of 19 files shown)


robell (approved):

backends/arm/test/tester/test_pipeline.py
```diff
@@ -922,7 +922,7 @@ def __init__(
         test_data: T,
         aten_op: str | List[str],
         exir_op: Optional[str | List[str]] = None,
-        run_on_vulkan_runtime: bool = False,
+        run_on_vulkan_runtime: bool = True,
```
digantdesai, line 925: I wonder if we should spin vk tests in a different CI job? @zingo refactored Ethos ones perhaps he has an opinion ..

zingo: Yes Its a plan do it soon,  @Michiel-Olieslagers is testing this first in our Arm internal CI.

mansnils: So essentially this is not even running in CI yet. Only in our internal CI. And yeah the plan is to do it soon for sure..

mansnils: Anyway regardless of this split we would still have run on vulkan runtime eq true here. The same way as run_on_fvp is true by default now. Essentially it just means before most VGF tests failed and now most are passing.

digantdesai (approved):

zingo:
Arm oobt fail is fixed in another PR

PR #14464, opened 2025-09-22
Enable Exynos backend Quatization

Jiseong-oh:
### Summary
- Implemented quantized strategies for enn-backend.
- Added support for ENN's quantization strategies.
- Successfully verified multiple quantized models.

### Test plan
python -m executorch.examples.samsung.scripts.${MODEL_NAME} -c e9955 -p A8W8


backends/samsung/utils/export_utils.py
```diff
@@ -29,24 +45,57 @@ def get_edge_compile_config():
             exir_ops.edge.aten._safe_softmax.default,
             exir_ops.edge.aten.layer_norm.default,
             exir_ops.edge.aten.matmul.default,
+            exir_ops.edge.aten.hardsigmoid.default,
         ],
     )
 
 
+def get_enn_pass_list(edge_program: ExportedProgram) -> List[PassType]:
```
SS-JIA: Why not apply these passes in `enn_preprocess.py` instead? With the current pattern, users will have to remember to call `get_enn_pass_list()` and pass the result to `to_edge_transform_and_lower()` which creates a point of failure if they forget to call it.

It seems to me that at the very least `AnnotateQparamsPass` should be moved to `enn_preprocess.py`, since it is required to preserve the quantization parameters. If it is not applied, then the `FoldQDQPass()` applied at the start of `enn_preprocess.py` will erase all quantization parameter information.

backends/samsung/_passes/annotate_qparams.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def get_quant_attrs(
+        self, quant_node: torch.fx.Node, key_map: Optional[Dict] = None
+    ) -> Dict[str, Any]:
+        quant_attr_keys = [arg.name for arg in quant_node.target._schema.arguments]
+        quant_attrs = dict.fromkeys(quant_attr_keys)
+        for key, attr in zip(quant_attr_keys[1:], quant_node.args[1:]):
+            # For channel-wise quantization, params are stored by buffer nodes.
+            if isinstance(attr, torch.fx.Node):
+                assert isinstance(attr.target, str), "Not supported now. "
```
SS-JIA: What specifically is the case that is not supported here? I see that we attempt to get the underlying tensor data via `get_buffer`, perhaps it would be better to check that `attr` is associated with a buffer?

My suggestions for this check is:
1. Make the assert message more descriptive; it's hard to tell right now what exactly is not supported
2. Instead of checking `isinstance(attr.target, str)`, there may be a better condition to check for that more precisely identifies the unsupported conditions.

backends/samsung/_passes/annotate_qparams.py
```diff
@@ -0,0 +1,243 @@
...
+        for node in graph_module.graph.nodes:
+            if key_map := QuantConstants.DEQUANT_OPS_KEY_MAP.get(node.target, None):
+                # We will fold node with constant output in the future pass as a constant node
+                # example: Constant->Q->DQ->nodeN->Q->DQ, this seq will be folded to one
+                # We need to store the q-params from last DQ params for quantizing constant value
+                quant_attrs = self.get_quant_attrs(node, key_map)
+                node.meta["quantize_attrs"] = quant_attrs
+                continue
+            else:
+                key_map = QuantConstants.QUANT_OPS_KEY_MAP.get(node.target, None)
```
SS-JIA: Why is `node.meta["quantize_attrs"]` not set for quant ops?

backends/samsung/_passes/annotate_qparams.py
```diff
@@ -0,0 +1,243 @@
...
+                continue
+            source_node = node.args[0]
+            if source_node.target in (
+                *QuantConstants.QUANT_OPS_KEY_MAP,
+                *QuantConstants.DEQUANT_OPS_KEY_MAP,
+            ):
+                # Currently, don't add quant info for d_qd node here.
+                continue
+            quant_attrs = self.get_quant_attrs(node, key_map)
+            assert node.args[0].target != operator.getitem, "Not supported now."
```
SS-JIA: This is ok for now, but it may easy to support. For `operator.getitem` nodes, you just have to get `node.args[0]` of the `getitem` node and use that as the `source_node`.

backends/samsung/_passes/annotate_qparams.py
```diff
@@ -0,0 +1,243 @@
...
+                    res_list.append(node)
+                else:
+                    _impl(user, res_list)
+
+        res_list: List[Node] = []
+        for user in node.users:
+            _impl(user, res_list)
+        return res_list
+
+    def _deliver_quant_params(self, node: Node):
```
SS-JIA: What does it mean to "deliver" the quant params? My current understanding is that since some operators (i.e. `linear`) are decomposed into a sequence of operators (i.e. `permute` + `mm`), this function attempts to propagate the quant params registered to the first node in the sequence to the entire sequence. Just want to check my understanding here.

backends/samsung/_passes/annotate_qparams.py
```diff
@@ -0,0 +1,243 @@
...
+                if node.target in QuantConstants.QUANT_OPS_KEY_MAP:
+                    node.args[0].meta["real_out"] = True
+                else:
+                    node.meta["real_out"] = True
+
+    def _annotate_real_in(self, graph_module: GraphModule):
+        for in_node in filter(
+            lambda x: is_graph_input(self.edge_program, x), graph_module.graph.nodes
+        ):
+            in_node.meta["real_in"] = True
```
SS-JIA: What's the purpose of this annotation? I don't see it being used anywhere. Although I do see the `"real_out"` annotation being used.

backends/samsung/_passes/fold_qdq.py
```diff
@@ -0,0 +1,35 @@
...
+            if node.target not in (
+                *QuantConstants.QUANT_OPS_KEY_MAP.keys(),
+                *QuantConstants.DEQUANT_OPS_KEY_MAP.keys(),
+            ):
+                continue
+            for user in [user for user in node.users.keys()]:  # noqa: C416
+                user.replace_input_with(node, node.args[0])
+            graph_module.graph.erase_node(node)
+
+    def call(self, graph_module: GraphModule):
```
SS-JIA, line 31: Would it be possible to use the `"real_in"` and `"real_out"` annotations to skip folding Q/DQ chains for graph inputs and outputs? If I understand correctly, this would remove the need to have a separate `insert_qdq.py` pass.

backends/samsung/_passes/fold_qdq.py
```diff
@@ -0,0 +1,35 @@
...
+        self._fold(graph_module)
+        graph_module.recompile()
+        dead_code_elimination_pass(graph_module)
```
SS-JIA, line 34: Typically, for passes that insert or remove nodes from the graph I recommend re-tracing the graph to validate that the graph structure is valid after executing the pass.

For example:

```python
graph_module.recompile()
dead_code_elimination_pass(graph_module)
# Re-trace to validate everything is ok
graph_module = super().call(graph_module).graph_module

```

SS-JIA: There are also some other passes added in this PR for which this comment also applies

backends/samsung/_passes/remove_useless_ops.py
```diff
@@ -0,0 +1,50 @@
...
+                node.target == exir_ops.edge.aten._to_copy.default
+                or node.target == exir_ops.edge.dim_order_ops._to_dim_order_copy.default
+            ) and "memory_format" not in node.kwargs:
+                continue
+
+            for user in [user for user in node.users.keys()]:  # noqa: C416
+                user.replace_input_with(node, node.all_input_nodes[0])
+            graph_module.graph.erase_node(node)
+
+    def call(self, graph_module: GraphModule):
```
SS-JIA, line 82: Could the functionality of `fold_redundant_as_strided.py` be merged into this pass? This can help consolidate the graph transform codebase and help make it more maintainable.

backends/samsung/serialization/enn_graph_schema.py
```diff
@@ -83,3 +126,22 @@ def finish(self):
 
     def serialize(self):
         return self.graph.Serialize()
+
+    @staticmethod
+    def _affine_meta_param(param: Any) -> str:
+        type_str_affine_table = {
+            torch.int8: "AINT8",
+            torch.int32: "FLOAT32",  # INT32 just used for HW quant.
```
SS-JIA: what is HW quant?

backends/samsung/enn_preprocess.py
```diff
@@ -53,12 +58,15 @@ def preprocess(
 
         enn_preprocess_passes = PassManager(
             passes=[
+                FoldQDQPass(),
```
SS-JIA, line 63: see comments in `export_utils.py`; IMO the passes returned by `get_enn_pass_list` should be applied here instead.

backends/samsung/serialization/enn_graph_schema.py
```diff
@@ -46,22 +54,57 @@ def define_op(
...
+            )
+            q_dtype = self._affine_meta_param(
+                quant_param[QuantConstants.QUANT_KEY.quant_dtype]
+            )
+            tensor.AddQuantizeParam(q_dtype, scales, zero_points)
+
+            if need_quantize and data is not None:
+                if isinstance(data, np.ndarray):
+                    data = torch.tensor(data)
+                data = quantize_tensor(
```
SS-JIA, line 96: I'm curious in what situations this is needed. Typically, after the pt2e quantization process, weight tensors should already be quantized and have int8 data type.

backends/samsung/_passes/fold_qdq.py
```diff
@@ -0,0 +1,35 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from executorch.backends.samsung.utils.constants import QuantConstants
+from executorch.exir.pass_base import ExportPass, PassResult
+from executorch.exir.passes import dead_code_elimination_pass
+from torch.fx import GraphModule
+
+
+class FoldQDQPass(ExportPass):
```
SS-JIA, line 13: The ENN delegate only supports static quantization (i.e. not dynamic quantization) at the moment, is that right?

SS-JIA:
Also, please run `lintrunner` to fix the lint errors.

```bash
# install lintrunner dependencies
pip install lintrunner
pip install lintrunner_adapters

cd executorch
lintrunner -a --verbose
```

backends/samsung/_passes/fold_qdq.py
Jiseong-oh: yes, right. we only support static quantization.

backends/samsung/utils/export_utils.py
Chen03ZhaoSamsung: We want some passes to be done before ```to_backend```, and we define ```to_edge_transform_and_lower_enn``` to make it easy to use. ```AnnotateQparamsPass``` could be moved to ```enn_preprocess.py```, so we have moved it now. Thanks for your recommendation.

backends/samsung/_passes/annotate_qparams.py
Chen03ZhaoSamsung: We confirmed this condition will not be triggered. We removed the assert expression. Thanks.

backends/samsung/_passes/annotate_qparams.py
Chen03ZhaoSamsung: Both quant ops and dequant ops don't need the ```node.meta["quantize_attrs"]``` now. Now it's removed, thanks for your comment.

backends/samsung/_passes/annotate_qparams.py
Chen03ZhaoSamsung: Thanks! Done.

backends/samsung/_passes/annotate_qparams.py
Chen03ZhaoSamsung: Yes you are right. The 'deliver' here is to propagate the quant params to successor nodes. We renamed this function to ```_propagate_quant_params```to avoid misunderstanding.

backends/samsung/_passes/annotate_qparams.py
Chen03ZhaoSamsung: Features need ```real_in``` and ```real_out``` are not contained in this version of code. So now we removed them. Thanks for your comment.

backends/samsung/_passes/fold_qdq.py
Chen03ZhaoSamsung: Thanks. Now we use ```super().call(graph_module).graph_module``` to validate the graph, except for ```insert_qdq```(because this pass will give a customized graph for our backend). Otherwise, we need to keep the graph same has object with ```exported_program```, so we apply it like ```_ = super().call(graph_module).graph_module```

backends/samsung/_passes/remove_useless_ops.py
Chen03ZhaoSamsung: Very good suggestion! Thank you!

backends/samsung/serialization/enn_graph_schema.py
Chen03ZhaoSamsung: torch.int32 comes from some experimental quantization strategy for our hardware, but it's not contained in this version. Now it is removed. Thanks for your comment.

backends/samsung/enn_preprocess.py
Chen03ZhaoSamsung: We moved ```annotate_qparams``` here, as the previous comment said. Thanks.

backends/samsung/serialization/enn_graph_schema.py
Chen03ZhaoSamsung: It seems constant tensors are not quantized after  ```convert_pt2e```, but just has quant/dequant node after the constant tensor nodes. So we quantize all the constant tensors here.

SS-JIA: Would you happen to have an example of this behaviour? My understanding is that the activation tensor will have quant/dequant, but the weight tensor will have only a quant node. If you inspect the tensor data backing the weight tensor, it should have int8 type.

For instance, here are some logs I collected while debugging MobileNet V2 quantized with XNNPACK Quantizer that I got a while back:

```python
# dq input node:
torch.float32: torch.Size([1, 384, 14, 14]) = quantized_decomposed::dequantize_per_tensor(torch.int8: torch.Size([1, 384, 14, 14]), 0.06534181535243988, 16, -128, 127, torch.int8,  ...)
# dq weight node:
torch.float32: torch.Size([96, 384, 1, 1]) = quantized_decomposed::dequantize_per_channel(torch.int8: torch.Size([96, 384, 1, 1]), torch.float32: torch.Size([96]), torch.int64: torch.Size([96]), 0, -127, 127, torch.int8,  ...)
# weight tensor:
tensor: torch.Size([96, 384, 1, 1]), torch.int8, -127, 127
# scales tensor:
tensor: torch.Size([96]), torch.float32, 0.004114292096346617, 0.009169017896056175
# zeros tensor:
tensor: torch.Size([96]), torch.int64, 0, 0
# conv node:
torch.float32: torch.Size([1, 96, 14, 14]) = aten::conv2d(torch.float32: torch.Size([1, 384, 14, 14]), torch.float32: torch.Size([96, 384, 1, 1]), torch.float32: torch.Size([96]),  ...)

```

SS-JIA:
Overall LGTM, but before I stamp:

1. Could you guys rebase to latest main and re-submit the PR? I want to see if the failures get resolved.
2. There's a remaining comment regarding manually quantizing constant tensors when serializing the model. To me, I don't think this should be required since the quantization workflow *should* be quantizing the constant tensor for you already.

backends/samsung/utils/export_utils.py
Sangsooko: Thanks for pointing this out. I'll address this item in the next update.

backends/samsung/serialization/enn_graph_schema.py
Sangsooko: As you can see in the following code which is a part of mobilebert, the div has quantized input and a constant input.
The above code changes the constant number of 5.656854249492381 into a quantized one.
Because this changing is only for the internal operation with real HWs, the code can be moved to our backend codes.
```
    %div : [num_users=1] = call_function[target=torch.ops.aten.div.Tensor](args = (%dequantize_per_tensor_default_424, 5.656854249492381), kwargs = {})
    %quantize_per_tensor_default_425 :   [num_users=1] = call_function[target=torch.ops.quantized_decomposed.quantize_per_tensor.default](args = (%div, 0.15715381503105164, -78, -128, 127, torch.int8), kwargs = {})
    %dequantize_per_tensor_default_425 : [num_users=1] = call_function[target=torch.ops.quantized_decomposed.dequantize_per_tensor.default](args = (%quantize_per_tensor_default_425, 0.15715381503105164, -78, -128, 127, torch.int8), kwargs = {out_dtype: torch.float16})
```

If you think the additional quantization process after convert_pt2e is  not appropriate according to the ExecuTorch code policy, we will move this code to our backend in the next update.

SS-JIA (approved):
LGTM! Thanks for addressing the comments.

Would you mind rebasing one more time before merging? Just to ensure that the merge base is fairly up to date. Thanks!

SS-JIA (changes requested):
Sorry, one more thing. In addition to rebasing, can you include the fix for getting the AI_LITECORE_API_KEY that's implemented in https://github.com/pytorch/executorch/pull/14866?

The fix was to move up when `secrets-env` is declared.

backends/samsung/serialization/enn_graph_schema.py
SS-JIA: Understood. I figured it might have been for binary ops that operate with constant args. In that case LGTM!

SS-JIA:
I decided to just merge https://github.com/pytorch/executorch/pull/14866, so please just rebase past it and this PR LGTM!

SS-JIA (approved):

SS-JIA:
Validated that samsung test is passing via https://github.com/pytorch/executorch/pull/14977. I believe that this PR cannot access the repository API secret because the merge branch is from a fork.

SS-JIA:
cherry-pick --onto release/1.0 -c fixnewfeature

PR #14465, opened 2025-09-22
Make extension_runner_util install relative

AdrianLundell:
This removes a reference to the binary tree in the install tree.


AdrianLundell:
Fails does not look related

extension/runner_util/CMakeLists.txt
```diff
@@ -28,7 +29,7 @@ target_compile_options(extension_runner_util PUBLIC ${_common_compile_options})
 install(
   TARGETS extension_runner_util
   EXPORT ExecuTorchTargets
-  DESTINATION ${CMAKE_BINARY_DIR}/lib
+  DESTINATION lib
```
digantdesai, line 32: Assuming CMAKE_BINARY_DIR == CMAKE_INSTALL_PREFIX?

digantdesai (approved):

extension/runner_util/CMakeLists.txt
AdrianLundell: My intention is to install the lib in CMAKE_INSTALL_PREFIX instead of the binary dir, if someone has a reason for placing it there they can make sure to set CMAKE_BINARY_DIR == CMAKE_INSTALL_PREFIX but I don't see why it would be a requirement.

digantdesai: No idea, but if the CI is green then I guess this is OK. Let me add trunk just to be sure.

digantdesai: Like this?
https://github.com/pytorch/executorch/blob/b3f3111bd619a41621d4713d6a9a9ae03d52ebb2/examples/cadence/CMakeLists.txt#L94

AdrianLundell: This change should only affect the install tree so if you want to use the build tree library instead that should still be in the same place, in the Arm example we use FindPackage to avoid direct paths and that works well. 

The reason why we want to make this change is to make it possible to move around the install tree without braking anything, for example to help parallelizing testing.

digantdesai: Good point. Thanks.

PR #14466, opened 2025-09-22
Arm backend: Replace module() calls with graph_module

### LINKED ISSUE #14417: test-arm-ootb-linux is failing in release branch
See our dashboard: 
https://hud.pytorch.org/hud/pytorch/executorch/release%2F1.0/1?per_page=50&mergeEphemeralLF=true

See the error message:
https://github.com/pytorch/executorch/actions/runs/17839511969/job/50725076407

Here's the PR: 
https://github.com/pytorch/executorch/pull/14393

Looks like we might need to update the Arm notebook?

tom-arm:
Using ExportedProgram.module() in later versions of torch introduce guards on the input. These guards cause an error in the ExportInterpreter which forbid the use of call_module.

Fixes: https://github.com/pytorch/executorch/issues/14417
Change-Id: I0cb68af9671dd1e3b543660c1cc49ff28cbffddc

### DIFF
--- a/backends/arm/scripts/TOSA_minimal_example.ipynb
+++ b/backends/arm/scripts/TOSA_minimal_example.ipynb
@@ -62,7 +62,7 @@
     "model = Add()\n",
     "model = model.eval()\n",
     "exported_program = torch.export.export(model, example_inputs)\n",
-    "graph_module = exported_program.module()\n",
+    "graph_module = exported_program.graph_module\n",
     "\n",
     "_ = graph_module.print_readable()"
    ]
@@ -201,7 +201,7 @@
     "            config=ExecutorchBackendConfig(extract_delegate_segments=False)\n",
     "        )\n",
     "\n",
-    "executorch_program_manager.exported_program().module().print_readable()\n",
+    "executorch_program_manager.exported_program().graph_module.print_readable()\n",
     "\n",
     "# Save pte file\n",
     "pte_name = base_name + \".pte\"\n",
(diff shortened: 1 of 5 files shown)


mergennachin (approved):
Thanks for fixing.

Please cherry-pick as well

https://github.com/pytorch/executorch/issues/14288

tom-arm:
> Thanks for fixing.
> 
> Please cherry-pick as well
> 
> #14288

Yes will do. And thank you for the prompt review

JacobSzwejbka:
> Using ExportedProgram.module() in later versions of torch introduce guards on the input.  

Ah you probably already know this now but .module() transforms the EP back into an executable object and tries to restore the original semantics. .graph_module is not intended to be run, but instead mostly is just a reference to the lifted graph. .graph_module does not give you the full semantics of the model. For that you need .graph_module + .graph_signature.

tom-arm:
> > Using ExportedProgram.module() in later versions of torch introduce guards on the input.
> 
> Ah you probably already know this now but .module() transforms the EP back into an executable object and tries to restore the original semantics. .graph_module is not intended to be run, but instead mostly is just a reference to the lifted graph. .graph_module does not give you the full semantics of the model. For that you need .graph_module + .graph_signature.

Given it doesn't give the full semantics, is this not the recommended approach for getting a GraphModule to pass to `prepare_pt2e`?

digantdesai (approved):

PR #14475, opened 2025-09-22
Revert "NXP backend: Resolve limitations of uncertain tensor formats"

kimishpatel:
Reverts pytorch/executorch#13942

Broke a buck target test_neutoron_backend because of "ModuleNotFoundError: No module named 'executorch.backends.nxp._passes". _passes needs a buck target specified to be included as dep


SS-JIA (approved):

MartinPavella:
Hi @kimishpatel ,
how did you discover the issue? I couldn't find any relevant failing builds.
Could you please instruct me on how I can reproduce the failure?

kimishpatel:
@MartinPavella it was one of the internal builds that was failing. Just resubmit it and i will import to internal and make sure the failure is not there.

In this specific case it is actually whats in the summary "ModuleNotFoundError: No module named 'executorch.backends.nxp._passes". You need to create a buck target and add that as dep here https://github.com/pytorch/executorch/blob/b3f3111bd619a41621d4713d6a9a9ae03d52ebb2/backends/nxp/tests/TARGETS#L45, or include transitively in one of the other deps

MartinPavella:
> @MartinPavella it was one of the internal builds that was failing. Just resubmit it and i will import to internal and make sure the failure is not there.
> 
> In this specific case it is actually whats in the summary "ModuleNotFoundError: No module named 'executorch.backends.nxp._passes". You need to create a buck target and add that as dep here
> 
> https://github.com/pytorch/executorch/blob/b3f3111bd619a41621d4713d6a9a9ae03d52ebb2/backends/nxp/tests/TARGETS#L45
> , or include transitively in one of the other deps

@kimishpatel Thank you for the reply.
I have implemented the changes in a new PR: https://github.com/pytorch/executorch/pull/14576
Could you please check if your internal build works now?

PR #14478, opened 2025-09-22
cuda export supported

Gasoonjia:
Summary: this diff introuce the cuda backend that compiles the partitioned model graph to run on CUDA devices. It uses the AOTInductor compiler to generate optimized CUDA kernels for the model's operators with libtorch-free. The compiled model can be executed on CUDA devices using the Executorch runtime.

Differential Revision: D82987410

### DIFF
--- a/backends/cuda/cuda_backend.py
+++ b/backends/cuda/cuda_backend.py
@@ -1,0 +1,171 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import contextlib
+import os
+import typing
+
+from typing import Any, Dict, final, List, Optional, Set
+
+import torch
+from executorch.exir._serialize._named_data_store import NamedDataStore
+from executorch.exir._warnings import experimental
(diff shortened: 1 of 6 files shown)


larryliu0820 (approved):

backends/cuda/cuda_backend.py
```diff
@@ -0,0 +1,167 @@
...
+        named_data_store.add_named_data("so_blob", so_data, 1, "aoti_cuda_blob")
+
+        # Clean up the generated so file; it has been packaged into the NamdeDataStore
+        # pyre-ignorep[6]: Incompatible parameter type
+        os.remove(so_path)
+
+        return PreprocessResult(
+            processed_bytes=b"",
+            debug_handle_map={},
+            data_store_output=named_data_store.get_named_data_store_output(),
```
JacobSzwejbka, line 170: Why are you putting this in the named_data_store since the .so is not actually shareable? Just legacy from when we were going to share with nativeRT?

Gasoonjia: we just want to make sure in et we are using the correct pipeline. 
in the future we need to find the way to load .so directly from ptd file which benefits both et loading efficiency and other partners like nativeRT.

PR #14487, opened 2025-09-22
Revert "Make _get_program_from_buffer work for bundled programs"

kimishpatel:
Reverts pytorch/executorch#14435

`buf_len` and some other variables unused related stuff is breaking a bunch of internal builds

PR #14488, opened 2025-09-22
Shift Left: Run selective build CI test with more restrictions

mergennachin:
Revert reason for https://github.com/pytorch/executorch/pull/14487 was that internal tests were running with "-Werror,-Wunused-variable" (https://www.internalfb.com/diff/D82981124) option but not OSS tests

### DIFF
--- a/codegen/tools/CMakeLists.txt
+++ b/codegen/tools/CMakeLists.txt
@@ -24,8 +24,17 @@ target_include_directories(
 
 # Compile options
 target_compile_options(
-  selective_build PUBLIC -Wno-deprecated-declarations -fPIC -frtti -fexceptions
+  selective_build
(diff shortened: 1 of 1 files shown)


codegen/tools/selective_build.cpp
```diff
@@ -186,6 +186,11 @@ get_kernel_tensor_metadatas_from_execution_plan(
 
 const executorch_flatbuffer::Program* _get_program_from_buffer(
     const py::bytes& buffer) {
+  
+  // TEST CODE to trigger CI failure
+  const std::string_view sv = buffer.cast<std::string_view>();
+  void* buf_ptr = const_cast<void*>(static_cast<const void*>(sv.data()));
```
mergennachin: will remove this before merging.

lucylq (approved):

PR #14499, opened 2025-09-23
Shift Left: Try building and testing selective_build in OSS

mergennachin:
- Enable buck targets in OSS under codegen/tools/targets.bzl
- Build all buck targets in OSS under codegen/tools/... (Can't test yet due to torch not available in OSS buck)
- Run test_tools_selective_build.py in pytest
- Building on top of https://github.com/pytorch/executorch/pull/14488

PR #14503, opened 2025-09-23
Make _get_program_from_buffer work for bundled programs take 2

Erik-Lundell:
This is a re-upload of #14435, with variables moved inside #ifdef to not get unused variables when building without devtools. 
---------------------------------------
Add some cmake to only do this if executorch is built with bundleio.

codegen/tools subdirectory include needs to be moved in top-level CmakeLists.txt
to have access to the bundled_program target.

Follow-up patch to enable the fix in arm backend.


Erik-Lundell:
Merge should wait until #14488 to make sure Meta's internal build is not broken

codegen/tools/CMakeLists.txt
```diff
@@ -28,6 +29,10 @@ target_compile_options(
 )
 
 # Link against required libraries
+if(TARGET bundled_program)
```
mergennachin, line 41: Should you check EXECUTORCH_BUILD_DEVTOOLS?

Something like 

```
if (EXECUTORCH_BUILD_DEVTOOLS and TARGET bundled_program)
   ... 
```

Erik-Lundell: I went with this since it is done that way on L754 in top-level CMakeLists.txt. Is there a benefit in additionally checking the flag or just to be on the safe side?

codegen/tools/CMakeLists.txt
```diff
@@ -28,6 +29,10 @@ target_compile_options(
...
+  target_compile_definitions(selective_build PRIVATE -DET_BUNDLE_IO)
+  target_link_libraries(selective_build PRIVATE bundled_program)
```
mergennachin, line 43: You also wanna change the targets.bzl file too 

I added a CI to build buck target for selective_build

https://github.com/pytorch/executorch/pull/14499

For future reference, here's how you can download buck: 

```
python tools/cmake/resolve_buck.py --cache_dir==/tmp/
# it will download and output a buck executable file within /tmp/ file
```

Once you have the buck file, you can do something like this: 

```
/tmp/buck2 build //codegen/tools/... 
# build all targets within //codegen/tools/...
```


```
/tmp/buck2 build codegen/tools:selective_build
# build all targets
```

Erik-Lundell: I had a look and the bundled_program seems to have multiple targets `runtime` and `runtime_aten`, depending on some get_aten_mode_options(). I need some assistance to figure this out. From what I can tell, if I add the different targets to selective_build that target is also split which means that targets depending on it will also split and so on... That doesn't sound right.

mergennachin: @Erik-Lundell -- no need to bifurcate on aten vs not. You can add dependency on `"//executorch/devtools/bundled_program:runtime` 

Historically, `runtime_aten` is mainly for internal customers when portable ops coverage was low. That's not the case, so you should be able to just depend on `"//executorch/devtools/bundled_program:runtime`  directly

digantdesai (approved):
Looks good with internal CI as well. Stamping..

PR #14504, opened 2025-09-23
Arm backend: Use default run_on_fvp=True unless explicitly set to False

SaoirseARM:
### Summary
Updates the unit test to use the default run_on_fvp unless explicitly set otherwise.


backends/arm/test/ops/test_sin.py
```diff
@@ -67,7 +67,6 @@ def test_sin_u55_INT(test_data: Tuple):
         (test_data,),
         aten_op,
         exir_ops=[],
-        run_on_fvp=False,
```
digantdesai, line 70: wouldn't these fail for no-fvp jobs without the xfails?

digantdesai (approved):
Thanks!

digantdesai:
Let me make sure we are not running into some random issue with internal FVP setup with this..

zingo:
@digantdesai Thanks, did the internal test work OK ?

digantdesai:
Slice tests like these `test_slice_tensor_u55_INT` are failing rest are OK. Let me take a look.

digantdesai:
We use bundled_io for some internal testing, and seems like `None` in the test_suite for Slice is causing a problem when creating bundled_io. Still digging..

digantdesai:
I have a fwd fix. I can land this.. thanks for the patience.

PR #14506, opened 2025-09-23
Arm backend: Add dtype check for aten.where in operator support check

YufengShi-dudu:
- Previously, aten.where with non-quantized FP inputs can pass the the operator support check in INT profile, get partitioned, and then be removed from the partition due to FP inputs.

- This will introduce dependency cycles, an invalid re-entry pattern `partition -> outside -> partition`, to the graph.

- Workaround: when aten.where(cond, x, y) has FP x and y in INT profile, only partition if both come from dequantize ops (DQ_OPS). Note this may over-reject cases like `dq -> op1 -> aten.where -> q` that could be partitioned.

- Don't partition aten.where with unsupported input dtype in FP profle.

Change-Id: I0cef30ffe36eeb6d9c8fae7618649292df242887


digantdesai (approved):

zingo:
Arm oobt fail is unrelated and will be fixed by another PR

PR #14508, opened 2025-09-23
[test_llama] CI: enable logging

digantdesai:
The rationale is to see tps and other common stats not just on CI but also when using this script for local developments.


mergennachin:
Don't you wanna just add to the llm present too?

mergennachin (approved):

digantdesai:
I am not sure about preset, since this isn't strictly necessary.

PR #14510, opened 2025-09-23
Cortex_m backend: Add cortex_m tester + test_add

AdrianLundell:
Note that tests are currently failing, comparing for example the call to arm_elementwise_add_s8 in op_quantized_add.cpp https://github.com/pytorch/executorch/blob/ab3100715afd21e5f0cee48675d9187152775d86/backends/cortex_m/ops/op_quantized_add.cpp#L88 with the definition in CMSIS-NN
https://github.com/ARM-software/CMSIS-NN/blob/88f1982a69c00ed13dd633a63da1009c48abbb4d/Include/arm_nnfunctions.h#L1923 it seems that the args are listed in the wrong order. This will be fixed in a future patch.

Minor fixes to get this to work:
- Add init file to make test names unique
- Update conftest to not crash is_option_enabled for tests running from external folder

### DIFF
--- a/backends/cortex_m/test/cortex_m_tester.py
+++ b/backends/cortex_m/test/cortex_m_tester.py
@@ -1,0 +1,102 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+
+from executorch.backends.arm.quantizer.arm_quantizer import (
+    get_symmetric_quantization_config,
+    TOSAQuantizer,
+)
+from executorch.backends.arm.test.common import get_u55_compile_spec
+from executorch.backends.arm.test.tester.arm_tester import Serialize
+from executorch.backends.cortex_m.passes.quantized_op_fusion_pass import (
+    QuantizedOpFusionPass,
+)
+
+from executorch.backends.cortex_m.passes.replace_quant_nodes_pass import (
+    ReplaceQuantNodesPass,
+)
+from executorch.backends.test.harness import Tester as TesterBase
+from executorch.backends.test.harness.stages import (
+    Export,
+    Quantize,
+    RunPasses,
+    StageType,
+    ToEdgeTransformAndLower,
+    ToExecutorch,
+)
+from executorch.backends.xnnpack._passes import XNNPACKPassManager
+
+
+class CortexMQuantize(Quantize):
+    def __init__(self):
+        compile_spec = get_u55_compile_spec()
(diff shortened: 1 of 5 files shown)


backends/cortex_m/test/cortex_m_tester.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+
+from executorch.backends.arm.quantizer.arm_quantizer import (
+    get_symmetric_quantization_config,
+    TOSAQuantizer,
```
digantdesai: Let's use XNNPACK quantizer for time being?

backends/cortex_m/test/tester.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
```
digantdesai, line 1: s/cortex_m_tester.py/tester.py

backends/cortex_m/test/ops/test_add.py
```diff
@@ -0,0 +1,175 @@
...
+# LICENSE file in the root directory of this source tree.
+
+
+import torch
+from executorch.backends.arm.test.common import parametrize
+from executorch.backends.cortex_m.test.cortex_m_tester import CortexMTester, McuTestCase
+from executorch.backends.test.suite.operators.test_add import Model, ModelAlpha
+
+
+class SelfAdd(torch.nn.Module):
```
digantdesai:
```suggestion
class SelfCortexMAdd(torch.nn.Module):
```

digantdesai (approved):
LGTM, let's get the ball rolling.. thanks.

backends/cortex_m/test/cortex_m_tester.py
AdrianLundell: Done

backends/cortex_m/test/tester.py
AdrianLundell: Done

backends/cortex_m/test/ops/test_add.py
AdrianLundell: I added CortexM as a prefix to all models here instead, I prefer to keep to one same naming scheme and not put the target description part of the name (CortexM) in the middle of the model description part (SelfAdd).

AdrianLundell:
I have not added buck targets in this PR since it is a new structure which makes it difficult to guess how it should look, but if you add it to this one I can follow that pattern for upcoming PR:s.

digantdesai:
@psiddh  - do you want to introduce buck targets here?

psiddh:
> @psiddh - do you want to introduce buck targets here?

@AdrianLundell  Once you land this PR , I can add the buck targets to it as a follow up

PR #14511, opened 2025-09-23
Arm Backend: Expose Vela's Debug Database

jmahbs:
Follow on from https://github.com/pytorch/executorch/pull/14401
Enables dumping of Vela's debug database to a specified directory . This gives us generic information on operators in  our model, and can be combined with the trace output to provide more detailed profiling analysis.


examples/arm/aot_arm_compiler.py
```diff
@@ -420,11 +420,14 @@ def get_compile_spec(
             tosa_spec = TosaSpecification.create_from_string("TOSA-1.0+INT")
         compile_spec = TosaCompileSpec(tosa_spec)
     elif "ethos-u" in target:
+        extra_flags = ["--verbose-operators", "--verbose-cycle-estimate"]
```
digantdesai, line 326: This is ok for side debugging, I wonder what are your thoughts on `ETRecord` and leveraging that infra?

jmahbs: I think this is a good idea. Re-using `ETRecord` for vela debugging simplifies the debugging workflow. This could be a good follow up PR, as it's out of scope with what this current change is enabling.

digantdesai (approved):

PR #14513, opened 2025-09-23
Arm backend: Add TOSA dialect op for RESIZE

oscarandersson8218:
Add TOSA backend dialect op for TOSA RESIZE. The dialect op replaces upsample_nearest2d and upsample_bilinear_2d in RewriteUpsamplePass. Also the Nodevisitors of upsample_nearest2d and upsample_bilinear2d are replaced by one NodeVisitor for the resize backend dialect op.


digantdesai (approved):
Nice!

PR #14514, opened 2025-09-23
NXP backend: added aten.sub operator support

novak-vaclav:
### Summary

adds support for aten.sub operator

### Test plan


backends/nxp/backend/edge_program_converter.py
```diff
@@ -30,6 +30,7 @@
     exir_ops.edge.aten._adaptive_avg_pool2d.default: AdaptiveAvgPool2dConverter,  # noqa F405
     exir_ops.edge.aten.addmm.default: AddMMConverter,  # noqa F405
     exir_ops.edge.aten.add.Tensor: AddTensorConverter,  # noqa F405
+    exir_ops.edge.aten.sub.Tensor: SubTensorConverter,  # noqa F405
```
MartinPavella: Minor:
Please maintain alphabetical order.

backends/nxp/backend/ir/converter/node_converters/ops_converters/sub_tensor_converter.py
```diff
@@ -0,0 +1,62 @@
...
+            return False
+
+        if hasattr(node.kwargs, "alpha"):
+            return False
+
+        return True
+
+    # sub.Tensor Node format: (Tensor self, Tensor other, *, Scalar alpha=1)
+    def convert(self, node: Node):
+        """Convert 'sub_tensor' operator to TFLite 'sub'."""
```
MartinPavella: Minor:
We prefer to use NeutronIR instead of TFLite, because the intermediate format is tailored for Neutron to the point where it is no longer pure TFLite.
And the `Sub` should be capitalized.

backends/nxp/backend/ir/converter/node_converters/ops_converters/sub_tensor_converter.py
```diff
@@ -0,0 +1,62 @@
...
+    def _is_supported_in_IR(
+        node: Node,
+        parameters_mapping: dict[str, Parameter],
+        custom_delegation_options: CustomDelegationOptions,
+    ) -> bool:
+        if len(node.args) != 2:
+            return False
+
+        if hasattr(node.kwargs, "alpha"):
+            return False
```
MartinPavella, line 47: I suppose this is here to simplify the implementation, but it is not quite correct.
The `sub` with an `alpha` attribute **can** be represented in the IR by using 2 operators (`Mul` and `Sub`).
Therefore, I would add a comment here stating something like:

```python
# The `alpha` attribute can be represented by adding an extra `Mul` operator.
#  However, this is not implemented as `alpha` is rarely used.
```

backends/nxp/neutron_partitioner.py
```diff
@@ -194,6 +194,7 @@ def tag_qdq_clusters(self, nodes: List[torch.fx.Node]):
     exir_ops.edge.aten._adaptive_avg_pool2d.default: AdaptiveAvgPool2dConverter,  # noqa F405
     exir_ops.edge.aten.addmm.default: AddMMConverter,  # noqa F405
     exir_ops.edge.aten.add.Tensor: AddTensorConverter,  # noqa F405
+    exir_ops.edge.aten.sub.Tensor: SubTensorConverter,  # noqa F405
```
MartinPavella: Minor:
Alphabetical order.

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -16,6 +16,7 @@
     AdaptiveAvgPoolPattern,
     AddmmPattern,
     AddTensorPattern,
+    SubTensorPattern,
```
MartinPavella: Minor:
Alphabetical order. (Same below.)

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,145 @@
+import numpy as np
```
MartinPavella, line 5: The header is missing (I know it was missing in the `test_add_tensor_converter.py` too, but let's add it to this new file.)
```python
# Copyright 2025 NXP
#

```

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,145 @@
...
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+
+    # Run conversion
+    _ = to_quantized_edge_program(model, [input_shape, input_shape])
+
+    # Capture generated model
+    tflite_flatbuffers_model, io_formats = converter_spy.spy_return
+
+    # Capture converted program
+    exported_program: ExportedProgram = converter_spy.call_args.args[1]
```
MartinPavella, line 54: Please add an assertion to make sure the `exported_program` contains the `aten.sub.tensor`.

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,145 @@
...
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+
+    # Run conversion
+    _ = to_quantized_edge_program(model, input_shape)
+
+    # Capture generated model
+    tflite_flatbuffers_model, io_formats = converter_spy.spy_return
+
+    # Capture converted program
+    exported_program: ExportedProgram = converter_spy.call_args.args[1]
```
MartinPavella, line 93: Please add an assertion to make sure the `exported_program` contains the `aten.sub.tensor`.
It is especially useful for this case (single input) as some optimization may be added in the future which replaces the node with `aten.zeros` for example.

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,145 @@
...
```
MartinPavella, line 127: Please add an assertion to make sure the `exported_program` contains the `aten.sub.tensor`.
It is useful here as if the `aten.sub` was for some reason **not** delegated and the `aten.convolution` was, the current test would pass.

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,145 @@
...
+    model = SubTensorModule()
+
+    # Run conversion
+    edge_program = to_quantized_edge_program(
+        model, [x_input_shape, y_input_shape]
+    ).exported_program()
+    nodes = list(edge_program.graph.nodes)
+
+    # Broadcast is not supported, node is not converted
+    assert nodes[6].target.__name__ == "aten.sub.Tensor"  # Sub Tensor is not delegated.
```
MartinPavella: Minor:
It is better to compare the target directly (not through its name).
```python
from executorch.exir.dialects._ops import ops as exir_ops

...

assert nodes[6].target == exir_ops.edge.aten.sub.Tensor
```

backends/nxp/tests/models.py
```diff
@@ -424,6 +424,34 @@ def forward(x):
...
+
+
+class SubTensorConvModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv = Conv2dModule(padding=1, stride=1)
+
+    def forward(self, x):
+        x = self.conv(x)
+        return x - x
```
MartinPavella: Minor:
It is not important for this test to use the single input `sub`, and its use actually reduces the quality of the test (since it always returns `0`, some incorrect behavior may go unnoticed).
I would definitely use 2 inputs:
```python
def forward(self, x, y):
    x = self.conv(x)
    return x - y
```

MartinPavella (changes requested):
I mainly added small suggestions and tips (which will be applicable to any of your future ExecuTorch PRs).
Once the changes are implemented, LGTM.

backends/nxp/tests/models.py
novak-vaclav: There is a code that tests two inputs `test_sub_tensor_quant_conversion` which I believe does what you describe after I added a second random input.
Should I keep this `test_sub_tensor_one_input_quant_conversion` or delete it?

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
robert-kalmar: The comment is resolved, but I don't see the file header.

backends/nxp/tests/models.py
robert-kalmar: The goal of test with Convolution is to force the channel-first to channel last conversion, and verify it. 

The ExecuTorch/PyTorch default memory layout is channel first (N, C, H, W) /Batch, channels, height, width), whereas neutron strictly requires channel last (N, H, W, C). Not channel last/first memory layout relates to only a snippet of operations, which computes on the level of 'channels', like convolutions (convolution, depthwise, transpose). 
The test using the `SubTensorModule` does not trigger the channel-first to channel-last conversion as sub does not have meaning for channels. By adding the Convolution, we set the model to have channel-last memory layout, and test if sub works correctly - the second operand not coming from Conv must be also transposed accordingly. This is what we test here. 
Now if you use `x - x`, the result is always 0, irrespecivelly if the channel-first to last is correctly handled.

robert-kalmar (changes requested):

backends/nxp/tests/models.py
novak-vaclav: oh now I get it, thank you! :)
is the following fix of the test okay?

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -103,34 +103,42 @@ def test_sub_tensor_one_input_quant_conversion(mocker, input_shape):
...
-        pytest.param((1, 4, 8, 8), id="4D."),
-        pytest.param((1, 4, 5, 5), id="4D, product of dims is not a multiple of 8."),
+        pytest.param((1, 4, 8, 8), (1, 8, 8, 8), id="4D."),
+        pytest.param(
+            (1, 4, 5, 5), (1, 8, 5, 5), id="4D, product of dims is not a multiple of 8."
+        ),
     ],
 )
-def test_sub_tensor_w_conv_quant_conversion(mocker, input_shape):
+def test_sub_tensor_w_conv_quant_conversion(mocker, x_input_shape, y_input_shape):
```
MartinPavella: The `y_input_shape` parameter doesn't seem necessary here. Since the convolution output channels are fixed at 8 and not configurable, the shape of second input is entirely determined by `x_input_shape`. This makes `y_input_shape` redundant and it adds unnecessary complexity to the test parameters.

Please remove the parameter and compute the `y_input_shape` inside the test function.

backends/nxp/quantizer/neutron_quantizer.py
```diff
@@ -207,6 +208,7 @@ def __init__(self):
                 NeutronAtenQuantizer(AbsPattern(), static_qconfig),
                 NeutronAtenQuantizer(AdaptiveAvgPoolPattern(), static_qconfig),
                 NeutronAtenQuantizer(AddTensorPattern(), static_qconfig),
+                NeutronAtenQuantizer(SubTensorPattern(), static_qconfig),
```
MartinPavella: Please maintain alphabetical order.

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -0,0 +1,174 @@
...
+    # Capture generated model
+    tflite_flatbuffers_model, io_formats = converter_spy.spy_return
+
+    # Capture converted program
+    exported_program: ExportedProgram = converter_spy.call_args.args[1]
+
+    input_data = (np.random.random(input_shape).astype(np.float32) * 50).astype(np.int8)
+
+    nodes = list(exported_program.graph.nodes)
+    assert nodes[2].name == "aten_sub_tensor"
```
MartinPavella: For these assertions, it is much safer co compare the target directly, not through its name. For example:
```python
assert nodes[2].target == exir_ops.edge.aten.sub.Tensor
```

Please change this for all applicable instances in this module.

MartinPavella (changes requested):

backends/nxp/tests/ir/converter/node_converter/test_sub_tensor_converter.py
```diff
@@ -95,27 +95,28 @@ def test_sub_tensor_one_input_quant_conversion(mocker, input_shape):
...
     ],
 )
-def test_sub_tensor_w_conv_quant_conversion(mocker, x_input_shape, y_input_shape):
+def test_sub_tensor_w_conv_quant_conversion(mocker, x_input_shape):
     model = SubTensorConvModule()
 
     converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
 
+    n, c, w, h = x_input_shape
+    y_input_shape = (n, 8, w, h)
```
MartinPavella: The shapes are NCHW (not NCWH).

MartinPavella (changes requested):

MartinPavella (approved):
LGTM. Once the tests pass, we can merge,

robert-kalmar (approved):

PR #14517, opened 2025-09-23
Move BUCK targets

JakeStevens:
Summary: As title

Differential Revision: D83053817

### DIFF
--- a/backends/nxp/runtime/targets.bzl
+++ b/backends/nxp/runtime/targets.bzl
@@ -1,20 +1,24 @@
 load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime")
+load("@fbsource//tools/target_determinator/macros:ci.bzl", "ci")
 
 def define_common_targets():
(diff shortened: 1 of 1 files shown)


JacobSzwejbka (approved):

kimishpatel (approved):

backends/nxp/runtime/targets.bzl
```diff
@@ -1,20 +1,24 @@
 load("@fbsource//xplat/executorch/build:runtime_wrapper.bzl", "runtime")
+load("@fbsource//tools/target_determinator/macros:ci.bzl", "ci")
 
 def define_common_targets():
     runtime.cxx_library(
-        name = "nxp_backend",
+        name = "nxp_backend_base",
```
mergennachin, line 6: Do you wanna protect this in OSS CI? 

See an example PR to build: https://github.com/pytorch/executorch/pull/14542

That way, we can know if a buck target has issue in OSS.

PR #14524, opened 2025-09-23
solve arm call_module issue

Gasoonjia:
disable check_guards in all exported_program.module() calls in ARM examples before convert_qt2e to prevent generating call_module nodes, which are not supported by et ARM quantization passes.

Differential Revision: D83073648


kimishpatel:
@digantdesai  can you tag someone from arm?

mergennachin:
Is this necessary after this PR? 

https://github.com/pytorch/executorch/pull/14466

mergennachin:
cc @tom-arm

PR #14526, opened 2025-09-23
Re-introduce some logging after backout

metascroy:
Summary:
Since D80715432 landed, we also landed D81256207 to add more logging.  To prevent conflicts, we backout in 3 steps:

* In diff 1/3, we backout D81256207, which added more logging
* In diff 2/3, we backout D80715432, which is the problematic diff we're trying to revert
* In diff 3/3, we introduce some of the logging we backed out in diff 1/3

Differential Revision: D82581441


mergennachin (approved):

mergennachin:
@metascroy did you figure out the root cause of the original revert?

metascroy:
> @metascroy did you figure out the root cause of the original revert?

Still investigating with internal teams.  I do not think there was an issue with the OSS builds, though.

PR #14528, opened 2025-09-23
add qnn eval CI

cccclai:
As title, add the script to prevent eval script regression. The test will calibrate qwen model and run eval, and test against threshold. If it's above the threshold, the test fail

image

### DIFF
--- a/.ci/scripts/test_qnn_static_llama_eval.sh
+++ b/.ci/scripts/test_qnn_static_llama_eval.sh
@@ -7,6 +7,26 @@
 
 set -euo pipefail
 
+source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
+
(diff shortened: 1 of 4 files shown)


kimishpatel:
what is the motivation to add this PR? In general I agree with protecting code, but it is not clear what kind of breakage we expect to prevent with this and enabling this on every pull request seems expensive.

cccclai:
> pull

It is super easy to use to evaluate accuracy with different modes and changes from qcom breaks it easily without tests

kimishpatel (approved):
discussed moving to trunk.yml if we cannot figure out how to make this specific job run only when changes are related to qualcomm

cccclai:
Will follow up on narrowing down the path for the job

PR #14529, opened 2025-09-23
Documentation on reproducing BUCK CI failures

mergennachin:
I found having resolve_buck.py script helpful to reproduce buck tests locally.

I know @swolchok wanted to delete, but how about having it in our codebase, but actually use it in our CI?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -329,6 +329,24 @@ the CI (continuous integration) jobs. If `main` is broken, consider rebasing
 your PR onto the `viable/strict` branch, which points to the most recent
 all-green commit.
 
+### BUCK tests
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -329,6 +329,24 @@ the CI (continuous integration) jobs. If `main` is broken, consider rebasing
 your PR onto the `viable/strict` branch, which points to the most recent
 all-green commit.
 
+### BUCK tests
+In addition to CMake, we run `buck` tests in our CI pipeline to ensure continued integration with certain customers. We recognize that `buck` is not a widely used build system and may not be installed on your local machine. If you encounter a failure in the `buck` CI and wish to reproduce it locally, follow these steps:
```
swolchok, line 333: > `buck` is not a widely used build system and may not be installed on your local machine

even if you do have buck2 installed, you need the exact version that goes with the version of the buck2 prelude we're using

PR #14540, opened 2025-09-24
Arm backend: Fix annotation of inplace ReLU

gggekov:
The ResNet18 model uses a lot of ReLUs with inplace=True As a result of the correct annotation, we can pass the numerical accuracy check on resnet with lower atol.


backends/arm/test/models/test_resnet18.py
```diff
@@ -54,7 +54,7 @@ def test_resnet_tosa_INT(per_channel_quantization):
         exir_op=[],
         use_to_edge_transform_and_lower=True,
         per_channel_quantization=per_channel_quantization,
-        atol=0.5,
+        atol=0.25,
```
digantdesai, line 57: 👍

digantdesai (approved):

PR #14542, opened 2025-09-24
Shift Left: Build and test extension/llm/... buck packages packages whenever possible

mergennachin:
We had a few reverts last week's oncall. 

Having these in place would have prevented them.

Test Plan: 

unittest-buck CI job


kimishpatel (approved):

kirklandsign (approved):

PR #14559, opened 2025-09-24
Remove reduce_range as it is not relevant for HTP

navsud:
Summary:
`reduce_range=True` reduces the available bit width by 1, in cases where quant_min, quant_max are not provided. It was originally intended for intel `fbgemm` kernels but I don't think this quantization setting is relevant for HTP. 
Also, PTQ quantization config doesn't use it, so removing it in all the QAT configs. This helped improve the QAT model quality.

Differential Revision: D82867843


billmguo:
@haowhsu-quic  can u help review this PR?

haowhsu-quic (approved):
Thank you.

cccclai (approved):
stamp on behalf of qcom's team review

PR #14560, opened 2025-09-24
use the same pytorch/torchao version same as ET repro in the wheel package test

cccclai:
As title, it was hardcoded to use nightly before, causing gaps as the main executorch flow

### DIFF
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -139,8 +139,13 @@ run_core_tests () {
   echo "=== [$LABEL] Installing wheel & deps ==="
   "$PIPBIN" install --upgrade pip
   "$PIPBIN" install "$WHEEL_FILE"
-  "$PIPBIN" install torch=="2.9.0.dev20250906" --index-url "https://download.pytorch.org/whl/nightly/cpu"
-  "$PIPBIN" install --pre torchao --index-url "https://download.pytorch.org/whl/nightly/cpu"
+
+  NIGHTLY_VERSION=$(python -c 'import re; txt=open("install_requirements.py").read(); print(re.search(r"NIGHTLY_VERSION\s*=\s*[\"\']([^\"\']+)[\"\']", txt).group(1))')
+
+  "$PIPBIN" install torch=="2.10.0.${NIGHTLY_VERSION}" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+  pushd "$REPO_ROOT" / third-party/ao > /dev/null
+  USE_CPP=0 python setup.py develop
+  popd > /dev/null
 
   echo "=== [$LABEL] Import smoke tests ==="
   "$PYBIN" -c "import executorch; print('executorch imported successfully')"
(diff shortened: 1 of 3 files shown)


install_requirements.py
```diff
@@ -36,6 +36,7 @@
 #
 # NOTE: If you're changing, make the corresponding supported CUDA versions in
 # SUPPORTED_CUDA_VERSIONS above if needed.
+TORCH_VERSION = "2.10.0"
```
kimishpatel: Can you put this in a separate PR?

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -139,8 +139,30 @@ run_core_tests () {
...
+)
+  echo "=== [$LABEL] Install torch==${TORCH_VERSION}.${NIGHTLY_VERSION} ==="
+
+  # Install torchao based on the pinned PyTorch version
+  "$PIPBIN" install torch=="${TORCH_VERSION}.${NIGHTLY_VERSION}" --index-url "https://download.pytorch.org/whl/nightly/cpu"
+
+  # Install torchao based on the pinned commit from third-party/ao submodule
+  pushd "$REPO_ROOT/third-party/ao" > /dev/null
+  USE_CPP=0 "$PYBIN" setup.py develop
+  popd > /dev/null
```
kimishpatel, line 165: Given how these are stored in install_requirements.py, it feels a bit flaky. I think you can just refactor install_requirements.py so that you have torch_pin.py that defines TORCH_VERSION and NIGHTLY_VERSION. That would be cleaner.

kimishpatel: That comment was for torch

kimishpatel, line 165: I think this dependency is for cpp kernels. Is that what we use for pip wheel as well? Can you confirm?

kimishpatel:
cc @GregoryComer. lets make sure there is one way to get the torchpin

.ci/scripts/test_wheel_package_qnn.sh
cccclai: @metascroy is the submodule we use for the torchao?

cccclai: Sure, add it in https://github.com/pytorch/executorch/pull/14631

kimishpatel: ok rebase on top of it and fix this?

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -139,8 +139,30 @@ run_core_tests () {
...
+txt = pathlib.Path("torch_pin.py").read_text()
+print(re.search(r'TORCH_VERSION\s*=\s*["\']([^"\']+)["\']', txt).group(1))
+PY
+)
+
+  NIGHTLY_VERSION=$(
+  "$PYBIN" - <<'PY'
+import re, pathlib
+txt = pathlib.Path("torch_pin.py").read_text()
+print(re.search(r'NIGHTLY_VERSION\s*=\s*["\']([^"\']+)["\']', txt).group(1))
```
kimishpatel: can you not just import torch_pin.py and get the TORCH_VERSION and NIGHTLY_VERSION from that instead of parsing the file?

kimishpatel (approved):
if you can import torch_pin.py that is better than parsing. approving but see if you can do that

.ci/scripts/test_wheel_package_qnn.sh
cccclai: updated

PR #14564, opened 2025-09-24
Dedup delegate blobs in emitter

lucylq:
Summary:
Previously we deduplicated entire 'BackendDelegate' blobs using the preprocessed blob. If two BackendDelegate fields have different id or compile specs, it would be disregarded.

This diff:
1. Only deduplicates the preprocessed blob. BackendDelegate retains its own compile specs, etc.
2. Removes the per-method 'delegate_cache', as we have a program-wide delegate cache.
3. Adds a test to confirm we have one delegate segment but two BackendDelegate references pointing at it.

Differential Revision: D83162107

### DIFF
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -147,8 +147,6 @@ class _EmitterState:
     operators: List[Operator]
     delegates: List[BackendDelegate]
     operator_cache: Dict[Tuple[str, str], int]
-    # delegate_cache: the key is hash(delegated_payload) and the value is the index in delegates
-    delegate_cache: Dict[str, int]
     emit_stacktrace: bool
     emit_mutable_buffer_names: bool
 
@@ -1092,7 +1090,7 @@ class _Emitter(torch.fx.Interpreter):
         delegate's blob."""
         processed_bytes = lowered_module.processed_bytes
         hashed = hashlib.sha256(processed_bytes).hexdigest()
-        delegate_index = self.emitter_state.delegate_cache.get(hashed)
+        delegate_index = self.program_state.backend_delegate_data_cache.get(hashed)
         delegate_ret = None
 
         if isinstance(self.node.meta["spec"], list):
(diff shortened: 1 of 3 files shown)


JacobSzwejbka (approved):

exir/emit/_emitter.py
```diff
@@ -1092,7 +1090,7 @@ def _emit_delegate(
         delegate's blob."""
         processed_bytes = lowered_module.processed_bytes
         hashed = hashlib.sha256(processed_bytes).hexdigest()
-        delegate_index = self.emitter_state.delegate_cache.get(hashed)
+        delegate_index = self.program_state.backend_delegate_data_cache.get(hashed)
```
cccclai, line 1093: what's the difference between emitter state and program state? How is `backend_delegate_data_cache` different than `delegate_cache`

exir/emit/_emitter.py
```diff
@@ -1130,28 +1128,20 @@ def _emit_delegate(
...
+                BackendDelegateInlineData(data=processed_bytes)
             )
-            if data_index is None:
-                data_index = len(self.program_state.backend_delegate_data)
-                self.program_state.backend_delegate_data_cache[hashed] = data_index
-                self.program_state.backend_delegate_data.append(
-                    BackendDelegateInlineData(data=processed_bytes)
-                )
-
-            backend_delegate = BackendDelegate(
```
cccclai, line 1144: I'm not quite sure why the previous logic didn't work but the new one does. It seems the logic is the same here

exir/emit/_emitter.py
```diff
@@ -1164,7 +1154,12 @@ def _emit_delegate(
             delegate_args.append(elem.id)
 
         self.chain.instructions.append(
-            Instruction(DelegateCall(delegate_index=delegate_index, args=delegate_args))
+            Instruction(
+                DelegateCall(
+                    delegate_index=len(self.emitter_state.delegates) - 1,
```
cccclai, line 1159: hmm why is the delegate index defined as `len(self.emitter_state.delegates) - 1`?

exir/emit/test/test_emit.py
```diff
@@ -1770,6 +1770,63 @@ def forward(self, x):
...
+        model_inputs = (torch.ones(1),)
+        edge_prog = to_edge(export(composite_model, model_inputs)).to_executorch()
+
+        exported_program = edge_prog.exported_program()
+        program = emit_program({"method1": exported_program}, False).program
+        self.assertEqual(len(program.execution_plan), 1)
+
+        plan = program.execution_plan[0]
+        # Two delegates that point to the same blob.
+        self.assertEqual(len(plan.delegates), 2)
```
cccclai, line 1818: Is it for checking the number of call_delegate instruction?

exir/emit/_emitter.py
lucylq: emitter state is per-method, program state covers all the methods in the program

exir/emit/test/test_emit.py
lucylq: Yes

exir/emit/_emitter.py
lucylq: The previous logic refers to the BackendDelegate created for the deduplicated processed blob. We may have a different compile specs, but the same processed blob, in which case the compile specs are lost. You can take a look at the test case as well, and try it on the old code.

exir/emit/_emitter.py
lucylq: the new logic creates a separate BackendDelegate for each call_delegate, so it corresponds to the length self.emitter_state.delegates.

PR #14570, opened 2025-09-25
Back FreeableBuffer with int64_t

lucylq:
Summary:
This diff backs FreeableBuffer with an int64_t, instead of a void*

## Usecase
PTE file lives on a 32-bit system, where void* is 4 bytes.
PTD file lives on a 36-bit system, which requires int64_t to address it.

We want to fetch addresses from the ptd file and pass them to accelerator. Executorch APIs return a FreeableBuffer when fetching data (data_loader.h, named_data_map.h). FreeableBuffer is currently backed by a void*, which could truncate a 36-bit address, when on a 32-bit system.

Note that we still want existing void* behavior to load segments etc., and only want int64_t behavior when fetching weights from the named_data_map.

## Potential concerns
* Increased memory usage; additional 4 bytes for each FreeableBuffer + some extra for the std::variant template. For the PTE file, this is order of the number of segments, which is usually small.
* Increased runtime latency; calls to the existing void* API perform truncation checks now.

## Alternatives
Why we choose this solution? Seems to be the least intrusive out of a number of solutions.
1. Compiler macro to switch the backing value of FreeableBuffer to int64_t for specific builds. However  void* and int64_ are both required - void* for the existing use cases (fetching delegate blobs). Both APIs must exist.
2. Template FreeableBuffer on void* and int64_t. Messy, as the template is contagious; it will need to be applied to data_loader.h, named_data_map.h, and potentially program/method. Imagine this will bloat the core runtime with templating.
3. Store int64_t addresses in FreeableBuffer, and ask user to parse the address to load the data. This avoids changing FreeableBuffer, but is not semantically correct as the API is not returning a buffer of data, but an address that the user must parse and then load data from.
4. Add a specific API to named_data_map.h that returns an int64_t buffer. Not a good use of API surface.

https://docs.google.com/document/d/11dMXh1N66rfY-8aO3N-ra2dDPx0RGCoc1rlCSN80Zf8/edit?tab=t.0

Differential Revision: D83007972

### DIFF
--- a/runtime/core/freeable_buffer.h
+++ b/runtime/core/freeable_buffer.h
@@ -9,6 +9,12 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdint>
+#include <variant>
+
+#include <executorch/runtime/core/error.h>
+#include <executorch/runtime/core/result.h>
+#include <executorch/runtime/platform/assert.h>
 
 namespace executorch {
 namespace runtime {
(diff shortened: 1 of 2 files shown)


runtime/core/freeable_buffer.h
```diff
@@ -105,8 +175,16 @@ class FreeableBuffer final {
   FreeableBuffer& operator=(const FreeableBuffer& rhs) = delete;
 
   FreeFn free_fn_;
+  FreeUInt64Fn free_uint64_fn_;
```
swolchok: it looks like at most one free function can be set, and it needs to correspond to the contents of data_, right? if so, I would reorganize things like so:

```
struct PointerData {
  FreeFn free_fn_;
  const void* data_;
};

struct Int64Data {
  FreeUInt64Fn free_fn_;
  uint64_t data_;
};

std::variant<PointerData, Int64Data> data_;
```

this way we only increase the size of FreeableBuffer by 8 bytes instead of 16.

lucylq: Thank you! This is really helpful I'll try it.

swolchok (approved):
feel free to ping me if I don't re-review within a day next time

zingo:
This seem to have broken the coretex-m size test :(

PR #14574, opened 2025-09-25
cuda export supported

Gasoonjia:
Summary: this diff introuce the cuda backend that compiles the partitioned model graph to run on CUDA devices. It uses the AOTInductor compiler to generate optimized CUDA kernels for the model's operators with libtorch-free. The compiled model can be executed on CUDA devices using the Executorch runtime.

Reviewed By: angelayi, larryliu0820

Differential Revision: D82987410


larryliu0820 (approved):

PR #14577, opened 2025-09-25
Arm backend: Add U55 operator check for cast

oscarandersson8218:
U55 does not support casting from int32 or any cast involving booleans. This patch introduces a new U55 operator check that will reject any such casts. Note that bool->bool and int32->int32 are considered ok as these will not result in any cast operation.


zingo:
@digantdesai would this be OK to merge?

digantdesai (approved):

digantdesai:
Seems low risk for internal 🤞

PR #14578, opened 2025-09-25
Arm backend: Add chronological dependencies for passes

martinlsm:
The passes in the Arm backend have an attribute called `_passes_required_after` which is a set specifying which passes must run after the pass itself. This patch sets these dependencies for all the passes.


martinlsm:
Failures look unrelated

digantdesai (approved):

PR #14579, opened 2025-09-25
Arm backend: add DeiTTiny evaluator and deterministic shuffled calibration subsets

tirwu01:
Change-Id: I7f61120772906ae0fec5d1f2b9cfcc0aa2c2c7af

### Summary
Add DeiTTiny evaluator for model accuracy evaluation. 

### DIFF
--- a/backends/arm/util/arm_model_evaluator.py
+++ b/backends/arm/util/arm_model_evaluator.py
@@ -1,5 +1,4 @@
 # Copyright 2024-2025 Arm Limited and/or its affiliates.
-# All rights reserved.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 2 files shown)


backends/arm/util/arm_model_evaluator.py
```diff
@@ -1,5 +1,4 @@
 # Copyright 2024-2025 Arm Limited and/or its affiliates.
-# All rights reserved.
```
digantdesai, line 2: unintentional delete?

backends/arm/util/arm_model_evaluator.py
```diff
@@ -197,3 +224,69 @@ def evaluate(self) -> dict[str, Any]:
...
+                "Top 1: {}".format(top1_correct / ((i + 1) * self.__batch_size))
+            )
+            logger.info(
+                "Top 5: {}".format(top5_correct / ((i + 1) * self.__batch_size))
+            )
+        top1_accuracy = top1_correct / len(dataset)
+        top5_accuracy = top5_correct / len(dataset)
+        return top1_accuracy, top5_accuracy
+
+    def evaluate(self) -> dict[str, Any]:
```
digantdesai, line 367: would it better to refactor this and MobileNetV2Evaluator to share much of the code?

tirwu01: good suggestion, will do.

backends/arm/util/arm_model_evaluator.py
mansnils: It is intentional since file is anyway modified and there is only an Arm copyright.

mansnils (approved):

PR #14583, opened 2025-09-25
Arm backend: Small refactoring of aot_arm_compiler

Erik-Lundell:
* Move evaluation logic out of aot_arm_compiler
 * Remove example models
    
    It doesn't make sense to have example models in
    the aot_arm_compiler script anymore, they are too
    simple, and there are other ways to do examples.
    
    Add and softmax can be replaced with example/models tests,
    add2 and add3 are deprecated.
    q-models can't be removed yet since the new testing is not
    in place but they should be as soon as it is.


Erik-Lundell:
From what I can tell, failing checks are unrelated.

digantdesai:
> q-models can't be removed yet since the new testing is not in place but they should be as soon as it is.

cc @AdrianLundell

digantdesai (approved):
Thanks!

PR #14584, opened 2025-09-25
Arm backend: Move rescale ops out of node visitors

martinlsm:
Some TOSA ops do not support INT8 as inputs and outputs. Instead, only INT32 is supported as a whole number type. Prior to this patch, affected node visitors inserted rescale ops between the data types INT8 and INT32 before and after the operator such that it will accept its input and output.

Change this by moving the insertion of the rescale ops to a new pass called `InsertRescaleInt32Pass`. This will further enable optimizations to the graph by fusing the rescale nodes.

Only comparison, ABS, MAXIMUM and MINIMUM operators are handled in this patch; the remaining ones are left out to be done in another patch.

### Test plan
This is refactoring which means that external behavior is not altered. A new pass `InsertRescaleInt32Pass` has been added and it comes with a new unit test in backends/arm/test/passes/test_insert_rescale_i32_pass.py.


martinlsm:
failures look unrelated

backends/arm/_passes/insert_rescales_pass.py
```diff
@@ -65,3 +70,185 @@ def call(self, graph_module: GraphModule) -> PassResult:
...
+        def int32_qargs(s):
+            return QuantArgs(
+                scale=s,
+                zp=0,
+                qmin=torch.iinfo(torch.int32).min,
+                qmax=torch.iinfo(torch.int32).max,
+                dtype=torch.int32,
+            )
+
+        if target in [
```
digantdesai, line 119: may be this boat has sailed already but for logical grouping perspective, I would rather keep the util call in respective ops instead of a pass.. but that's just me. Not a strong objection.

martinlsm: @digantdesai Not sure what you mean fully. Can you elaborate a bit?

digantdesai: Instead of inserting rescale ops in the tosa graph through this pass, I would prefer (mild preference) of keeping the way things were, i.e. an op lowering (op_<kind>.py) would inject an op when needed. 

This is mainly to keep things together logically and keep the tosa graph 'sound' after each op lowering.

That said I understand the code duplication aspect, but that could be tackled with keeping these `insert rescale` as util functions.

Again, not a strong objection. Stamping to unblock you, and leaving it up to you to decide.

Thanks!

digantdesai (approved):

PR #14585, opened 2025-09-25
Arm backend: Update full quantization annotation

oscarandersson8218:
full, full.default and fill_.Scalar were previously part of _one_to_one_shared_input_or_input_act_qspec without having any input nodes. This meant that these nodes were never annotated and solely relied on the next node to annotate its input. This patch changes so that full, full.default and fill_.Scalar are annotated in the same way as scalar_tensor.default.

Also adds these targets to _is_large_scalar().


digantdesai:
Arm no-fvp CI is failing, else lgtm.

digantdesai (approved):

zingo:
re-run seem to help, it a unrelated flakey test I have flagged it internaly at Arm.

PR #14588, opened 2025-09-25
Add strict-flag to ExportSession

Erik-Lundell:
**Add strict export option to ExportRecipe**
    
    Default is True, mirroring earlier behavior.
    Also update ExportSession to handle this.

### DIFF
--- a/export/recipe.py
+++ b/export/recipe.py
@@ -1,5 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -151,6 +152,7 @@ class ExportRecipe:
         executorch_backend_config: Optional backend configuration for ExecuTorch
         pipeline_stages: Optional list of stages to execute, defaults to a standard pipeline.
         mode: Export mode (debug or release)
+        strict: Set the strict flag in the torch export call.
     """
 
     name: Optional[str] = None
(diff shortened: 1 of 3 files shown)


export/recipe.py
```diff
@@ -163,6 +165,7 @@ class ExportRecipe:
     executorch_backend_config: Optional[ExecutorchBackendConfig] = None
     pipeline_stages: Optional[List[StageType]] = None
     mode: Mode = Mode.RELEASE
+    strict: bool = True
```
digantdesai, line 168: Nit update docstrings..

export/stages.py
```diff
@@ -174,10 +177,12 @@ def __init__(
         transform_passes: (
             None | List[Callable[[str, ExportedProgram], List[PassType]]]
         ) = None,
+        post_edge_passes: list[PassType] | None = None,
```
digantdesai: I wonder overloading this and tobackend stage what about adding another stage for "post_edge_transform"

digantdesai (approved):
I think this is going to be a general problem for all "non-partitioner" op-lib backends. My thinking is we should either hide them behind a pseudo-partitioner wrapper to make it more composible. For now this might be OK. See some comments I added here. Stamping to unblock. Let's try to resolve the comments before merging.

export/stages.py
```diff
@@ -499,10 +509,12 @@ def __init__(
...
```
abhinaykukkadapu: @Erik-Lundell Thanks for making the changes, to me `post_edge_passes` seems ambiguous as the current transform passes run after `to_edge()` in both `to_edge_transform_lower()` and `to_edge() -> transform() to_backend()` paths. If these passes are run after `to_backend()` we can name something that reflects this, may be "post_to_backend_passes"?

export/stages.py
abhinaykukkadapu: @Erik-Lundell +1 to this, we can make it into its own stage, and have the arm backend recipe provider to publish the stages it want, something like this: https://github.com/pytorch/executorch/blob/main/backends/xnnpack/recipes/xnnpack_recipe_provider.py#L40 and provider can also publish the recipe stage sequence at: https://github.com/pytorch/executorch/blob/main/export/recipe.py#L164

Erik-Lundell:
Ack, I removed the post_edge commit from this PR and will create a new with the post-to-backend stage.

PR #14592, opened 2025-09-25
Fix pyproject.toml license classifier deprecation

tmi:
Gets rid of the 'deprecated' warnings that pop up multiple times during build/install

Bumps setuptools requirement to accept the new license declaration format

### Summary
Just a tiny PR, no change to API or code or anything. The license itself is as before, it just changes the manner in which it is declared -- as recommended by the PyPA guidelines https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

I tried to find any issue related to this but found none. And I guess not worth it creating one

### Test plan
Since this does not change any of the code, I just tested that the package can be installed/built as before via `./install_executorch.sh`, and that the deprecation warnings vanish

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,10 @@
 [build-system]
 requires = [
   "cmake>=3.29,<4.0.0",  # For building binary targets in the wheel. 4.0.0 breaks third-party CMake build so temporarily pin the version.
+  "packaging>=24.2", # Lower bound required by setuptools
   "pip>=23",  # For building the pip package.
   "pyyaml",  # Imported by the kernel codegen tools.
-  "setuptools>=63",  # For building the pip package contents.
+  "setuptools>=77.0.3",  # For building the pip package contents.
   "wheel",  # For building the pip package archive.
   "zstd",  # Imported by resolve_buck.py.
   "certifi",  # Imported by resolve_buck.py.
@@ -21,7 +22,8 @@ readme = "README-wheel.md"
 authors = [
   {name="PyTorch Team", email="packages@pytorch.org"},
 ]
-license = {file = "LICENSE"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
 keywords = ["pytorch", "machine learning"]
 # PyPI package information.
 classifiers = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -21,7 +21,8 @@ readme = "README-wheel.md"
 authors = [
   {name="PyTorch Team", email="packages@pytorch.org"},
 ]
-license = {file = "LICENSE"}
+license = "BSD-2-Clause"
```
mergennachin: We're actually BSD-3-Clause 

https://github.com/pytorch/executorch/blob/main/LICENSE

The last clause there makes it 3-clause BSD

tmi: thank you for the correction, updated

mergennachin:
Why did you upgrade setuptools?

tmi:
@mergennachin without the lower bound, the `license` fields fail to be parsed -- the first CI run failed because of that (as the runner defaulted to a cached version, 60-smth). I just grabbed the bound from a similar issue https://github.com/vllm-project/vllm/pull/17389. I can leave a comment in the pyproject.toml linking the PEP that requries this bound. Thinking of it, we may also want to add an upper bound

The failures of the recent runs, like `cannot import name 'PerBlock' from 'torchao.quantization.pt2e'` or `ValueError: QNN backend is not available.` are not related, right? Or is it possible that the setuptools bump caused some silent failure of a prerequisite build?

mergennachin:
Okay got it

That seems to be the right version. 

The other failures seem unrelated.

Please also update https://github.com/pytorch/executorch/blob/main/requirements-dev.txt

mergennachin:
Test failures for "Test CUDA Builds" look legit 

Can you also update packaging>=24.2 in here? https://github.com/pytorch/executorch/blob/7ed926693fbaf471ec8072ff8896090f9fe5fd44/pyproject.toml#L61

tmi:
> Test failures for "Test CUDA Builds" look legit. Can you also update packaging>=24.2 in here? 

right, thx. I updated it, but in the `build-system`'s requirements, not in the `depedendencies` -- I think thats more suitable place, given the test failure. Not sure if the runtime `packaging` dependency needs a lower bound too since there is no runtime `setuptools` dependency -- I'd guess not

mergennachin (approved):

mergennachin:
Thank you @tmi

PR #14593, opened 2025-09-25
[ET-VK][ez] Do not allow using glslc from Android NDK

SS-JIA:
Summary:
Title says it all. The `glslc` binary distributed with the Android NDK can no longer be used to build the Vulkan delegate since it doesn't support the GL_EXT_integer_dot_product extension.

This PR updates the CMakeLists.txt to look for the `glslc` installed on the system instead by default.


cccclai (approved):

SS-JIA:
Merged https://github.com/pytorch/executorch/pull/14596 instead

PR #14599, opened 2025-09-25
[Llava] Add max_context_len CLI arg

### LINKED ISSUE #14474: LLaVA allocates too much memory on iOS
### 🐛 Describe the bug

When running LLaVA model on iOS it requires 6,137 Mb of RAM which is too much for a high-end iPhone to handle.

Image

The above image shows the memory usage on Simulator, where there are virtually no memory restrictions.
LLaVa used to work in the past when the memory requirement was lower, well under 4Gb.

### Versions

Git clone and run https://github.com/meta-pytorch/executorch-examples/tree/main/llm/apple
It'll use the latest 1.0.0 runtime build. Switch to a nightly if needed as explained in https://docs.pytorch.org/executorch/main/using-executorch-ios.html#xcode
Copy llava.pte and tokenizer.bin to Simulator or device, pick the model and tokenizer files in UI and start a conversation

GregoryComer:
### Summary
Add a required max_context_len argument to the Llava example model export. When set to 768, this reduces the memory consumption (~6GiB -> ~4.8GiB RSS) at the cost of a smaller context length and thus fixes https://github.com/pytorch/executorch/issues/14474.

### Test plan
Ran ./test_llava.sh and validated the reported memory consumption on an x86 Linux machine.

```
I 00:00:18.433471 executorch:main.cpp:172] Starting generation...
I 00:00:18.433500 executorch:multimodal_runner.cpp:95] RSS after loading model: 4746.726562 MiB (0 if unsupported)
I 00:00:18.433554 executorch:multimodal_runner.cpp:119] Prefilling input 0/3, type: text
I 00:00:19.484581 executorch:multimodal_runner.cpp:119] Prefilling input 1/3, type: image
I 00:00:19.484710 executorch:multimodal_prefiller.cpp:83] Image tensor dim: 3, dtype: Byte
I 00:00:30.442685 executorch:multimodal_runner.cpp:119] Prefilling input 2/3, type: text
I 00:00:30.951938 executorch:multimodal_runner.cpp:138] RSS after multimodal input processing: 4847.933594 MiB (0 if unsupported)
I 00:00:30.952000 executorch:multimodal_runner.cpp:148] Max new tokens resolved: 153, pos_ 615, max_context_len 768
```

### DIFF
--- a/examples/models/llava/export_llava.py
+++ b/examples/models/llava/export_llava.py
@@ -281,6 +281,7 @@ def create_llava_config_from_args(args):
     llm_config = LlmConfig()
 
     llm_config.model.use_sdpa_with_kv_cache = args.use_sdpa_with_kv_cache
+    llm_config.export.max_context_length = args.max_context_len
     llm_config.export.max_seq_length = args.max_seq_len
     llm_config.export.output_name = args.pte_name
     llm_config.debug.profile_memory = args.profile_memory
@@ -295,6 +296,12 @@ def main():
         default=True,
         action=BooleanOptionalAction,
         help="Use sdpa_with_kv_cache custom op in LLava text model.",
+    )
+    parser.add_argument(
+        "--max-context-len",
+        required=True,
+        type=int,
+        help="Maximum context length for the text model.",
     )
     parser.add_argument(
         "--max-seq-len",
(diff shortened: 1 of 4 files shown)


examples/models/llava/export_llava.py
```diff
@@ -296,6 +297,12 @@ def main():
         action=BooleanOptionalAction,
         help="Use sdpa_with_kv_cache custom op in LLava text model.",
     )
+    parser.add_argument(
+        "--max-context-len",
+        default=768,
```
kimishpatel: i think we should make this a required arg. this default could be a footgun as well

kimishpatel (changes requested):
would like to make max_context_len required arg and if it is not in export_llm I think it should be. Or at least improve documentation to include this arg in the export CLI example

examples/models/llava/export_llava.py
GregoryComer: I've updated it to be required.

GregoryComer:
> would like to make max_context_len required arg and if it is not in export_llm I think it should be. Or at least improve documentation to include this arg in the export CLI example

I've updated the arg to be required and made corresponding changes to the example README and test_llava.sh script. I'd recommend better documenting the different between the user facing max_context_len and max_seq_len args in the export_llava.py script, though I'm likely not the right owner for this.

kimishpatel:
> > would like to make max_context_len required arg and if it is not in export_llm I think it should be. Or at least improve documentation to include this arg in the export CLI example
> 
> I've updated the arg to be required and made corresponding changes to the example README and test_llava.sh script. I'd recommend better documenting the different between the user facing max_context_len and max_seq_len args in the export_llava.py script, though I'm likely not the right owner for this.

It is also a bit hard to explain the difference between the two unless user understands how to use it for better memory footprint. I would just opt for better default for max_seq_len

kimishpatel (approved):

larryliu0820 (approved):
Thank you for the fix!

GregoryComer:
Noting that CI failures appear to be pre-existing or flaky. Merging.

GregoryComer:
I'll submit a pick request once the trunk jobs complete.

PR #14603, opened 2025-09-25
set a dependency job to limit qnn jobs to specific paths

cccclai:
without dependency: https://github.com/pytorch/executorch/actions/runs/18017807043/job/51267174825?pr=14603

with the dependency: eems like the job is skipped https://github.com/pytorch/executorch/actions/runs/18017569385/job/51266354486?pr=14603

with the dependency and trigger with qcom changes: https://github.com/pytorch/executorch/actions/runs/18017856742/job/51267377173?pr=14603


cccclai:
follow up on https://github.com/pytorch/executorch/pull/14528

kimishpatel (approved):

PR #14618, opened 2025-09-26
Qualcomm AI Engine Direct - Suite Operator Test Support (Part1)

winskuo-quic:
### Summary

- Support Add/Sub with alpha values
- Support Conv3d
- Support TransposeConv3d

### Test plan
UT added


winskuo-quic:
Hi @cccclai,
This PR should resolve some suite operator's failure.
It would be appreciated if you could guide me through the following:
1. How could I see the most up-to-date suite operator results in mainline, like the pdf file you shared with us through email.
2. Could you share the command to run the suite operator tests? The command I use is:
`python -m executorch.backends.test.suite.runner operators --flow qnn --filter "test_conv3d_basic_qnn"`.
I can see op validation and preprocess is passing, however, I keep getting an unknown error at the end, causing the test to skip, as shown in the image below:
image

cccclai:
> Hi @cccclai, This PR should resolve some suite operator's failure. It would be appreciated if you could guide me through the following:
> 
> 1. How could I see the most up-to-date suite operator results in mainline, like the pdf file you shared with us through email.
> 2. Could you share the command to run the suite operator tests? The command I use is:
>    `python -m executorch.backends.test.suite.runner operators --flow qnn --filter "test_conv3d_basic_qnn"`.
>    I can see op validation and preprocess is passing, however, I keep getting an unknown error at the end, causing the test to skip, as shown in the image below:
> 
> image

@GregoryComer can you share the steps?

cccclai (approved):

cccclai:
@winskuo-quic from the latest run, it seems the number improves https://github.com/pytorch/executorch/actions/runs/18053810243

cccclai:
@winskuo-quic from the latest run, it seems the number improves https://github.com/pytorch/executorch/actions/runs/18053810243

GregoryComer:
> https://github.com/pytorch/executorch/actions/runs/18053810243

Thanks for the fixes! I updated the README here with some more repro examples: https://github.com/pytorch/executorch/blob/main/backends/test/suite/README.md. I've also updated to use pytest, which should make it easier to run the tests.

Regarding skips on conv3d, that's effectively a pass. What's happening is that the operator or model has an undelegated conv3d, which we don't have a portable implementation for. If nothing is delegated, it gets considered a pass, as the backend worked as intended. If it's partially delegated and has missing portable ops, it's considered skipped as it can't test the model. But, from the backend perspective, it's passing. I should maybe just update the test logic to mark as pass instead of skip for this case.

cccclai:
> I should maybe just update the test logic to mark as pass instead of skip for this case.

@GregoryComer yes let's do that

PR #14619, opened 2025-09-26
Remove explicit device arguments

navsud:
Summary:
As part of enabling QAT for HTP model, we need to run QAT on the model that we use during export. Currently Rope is explicitly hardcoded to "cpu". This change enables us to create rope params on "cuda" if it is run on GPU machine.

Differential Revision: D82239525

### DIFF
--- a/examples/models/llama/model_args.py
+++ b/examples/models/llama/model_args.py
@@ -63,6 +63,9 @@ class ModelArgs:
     use_sdpa_with_kv_cache_op: bool = (
         False  # Use custom sdpa op that updates kv cache in-place
     )
+    # Device to use for the model: "cpu" or "cuda" (needed for QAT)
+    # Only used for creating Rope parameters
+    device: str = "cpu"
     # Generate logits for all inputs. When it's True, it would take big memory usage
     # at runtime. Enable it only necessary (e.g., use perplexity tools that requires
     # logits for all input tokens.)
--- a/examples/models/llama/rope.py
+++ b/examples/models/llama/rope.py
@@ -138,7 +138,11 @@ class RotaryEmbedding(torch.nn.Module):
 # and https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_rope_utils.py#L242.
 # Current only support non-long rope.
 def hf_precompute_freqs_cis(
-    dim: int, end: int, theta: float, partial_rotary_factor: float = 1.0
+    dim: int,
+    end: int,
+    theta: float,
+    partial_rotary_factor: float = 1.0,
+    device: Union[str, torch.device] = "cpu",
 ):
     # Partial rotary embeddings.
     dim = int(dim * partial_rotary_factor)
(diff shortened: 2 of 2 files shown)


examples/models/llama/model.py
```diff
@@ -195,7 +195,7 @@ def __init__(self, llm_config: Optional[LlmConfig] = None):
 
         # Within the device="meta" context, tensors that are created do not carry data.
         # They possess all other metadata a tensor carries such as size, stride, requires_grad.
-        with torch.device("meta"):
+        with torch.device("cpu"):
```
lucylq: Hi @navsud, could you provide some more context on this?

> Update: This was failing multiple export tests, as the Llama2Model (at llama/model.py) was instantiating the transformer with "meta" device, which needed the rope params to be explicitly instantiated on "cpu" device. Changed "meta" to "cpu" to fix this issue.

Using torch.device("meta") provides some memory benefits during export because we do not load all the tensors into memory until necessary.

navsud: Full context is: We use the same rope implementation for QAT of HTP model using pt2e flow. But since "cpu" was hardcoded for Rope, it was a bit messy to move to GPU (especially when we are doing multi-GPU training). 
So, I removed "device" argument in rope implementation. 
However, it was causing fails in export, because - during export models are built with "meta" device, and then during checkpoint load the weights are "assigned". However "rope" freq params are not present in the checkpoint, hence not loaded, causing the fail in the export. 
So, I'm creating the model directly on "cpu" thus rope freq params are also created on the cpu, thus avoiding the need for explicit hardcoding of "cpu" only for rope.

navsud: @lucylq 
> Using torch.device("meta") provides some memory benefits during export because we do not load all the tensors into memory until necessary.

Well, it is really a very short-term benefit in the export flow, because right after the model is built with "meta" [here](https://github.com/pytorch/executorch/blob/main/examples/models/llama/model.py#L200), we are either (a) loading weights from checkpoint - at which point the model takes memory [here](https://github.com/pytorch/executorch/blob/main/examples/models/llama/model.py#L279-L287) or (b) initializing the model weights with zeros - which also takes memory [here](https://github.com/pytorch/executorch/blob/main/examples/models/llama/model.py#L289-L296).

lucylq: Thanks for the context @navsud. 

Noob q: if the device is set to cpu when creating the model, does this cause issues when moving to GPU later?

I'm concerned we're regressing export performance; do you see memory/latency regressions at export with this change? FWIW, we were able to see significant savings when using device=meta: [D54871495](https://www.internalfb.com/diff/D54871495)

navsud: @lucylq 
Thanks for the context around memory/latency savings with export[D54871495](https://www.internalfb.com/diff/D54871495).
So I'm left with creating with device set to cpu and then move to GPU later. I will try out and revise this PR accordingly.

navsud: @lucylq 
Updated to an implementation that doesn't disturb the way export works (so all of the memory/latency benefits with device=meta is preserved), but still works for my usecase.

billmguo (approved):

examples/models/llama/model.py
lucylq: Thanks for the changes @navsud, appreciate it.

navsud: @lucylq 
Had to make another revision as some other tests were failing. Please do another quick round of review.  Thanks.

lucylq (approved):

PR #14621, opened 2025-09-26
use ET_CHECK macro for sanity checks in memory shim layer

Gasoonjia:
Summary: this is a comprehensive update to use ET_CHECK macro to replace original if..else check for better follow et's law

Reviewed By: larryliu0820

Differential Revision: D83315545


larryliu0820 (approved):

PR #14623, opened 2025-09-26
Arm Backend: Add tests for stack.default

agrima1304:
Stack is not in the list of core ATen ops and is decomposed automatically when lowering the graph (https://docs.pytorch.org/docs/main/export.html#export-ir-decompositions), so only the tests need to be added.

stack is in this decomp table: https://github.com/pytorch/pytorch/blob/5d749ceb92c2c28bcfbdf918b4ab99b1a91fcb50/torch/_decomp/__init__.py#L466 


digantdesai (approved):
Thanks. Welcome to ET.

zingo:
@agrima1304 there is a fail in an Arm test, it might be a flaky test and a re-run might help, if so and it turns green we can merge and ask in about it in our internal slack nd maybe create a ticket. Let me know if you dont know how to re-run the test and Ill show you.

PR #14624, opened 2025-09-26
Arm backend: Fix torch.matmul() failures for 2D tensor inputs

YufengShi-dudu:
- ConvertMmToBmmPass converts an MM node to BMM nodes, turns input and output tensors from rank-2 to rank-3 via unsqueeze/squeeze, and inserts q-dq before and after BMM node when necessary.

- After ConvertMmToBmmPass:
```
  x -> q   -> dq   -> unsqueeze -> q_2 -> dq_2 ->
                                                 \
                                                bmm -> q_4 -> dq_4
                                                 /
  y -> q_1 -> dq_1 -> unsqueeze -> q_3 -> dq_3 ->
```

- Therefore, if the original matmul was 2D, the bmm already has DQ nodes on its inputs and Q node on its output. If AnnotateDecomposedMatmulPass (#10654) is still applied in this case, it produces illegal sequences such as: x -> q -> unsqueeze -> q_2 (invalid)

- Fix by checking whether the BMM is already surrounded by DQ nodes on its inputs and Q nodes on its output.

Change-Id: I9949d59b0b4a96fa34a88b0734014567ea6f24cc


YufengShi-dudu:
If possible, I suggest we get this fix into the 1.0 release branch

digantdesai (approved):
Thanks @YufengShi-dudu  for the bug fix. Please mark the PR to be cherry-picked in 1.0. Thanks again.

PR #14625, opened 2025-09-26
Arm backend: Fix Arm tester issue for inplace ops

mansnils:
Deep-copying the input avoids it getting mutated by the first reference run.


backends/arm/test/tester/arm_tester.py
```diff
@@ -463,6 +463,10 @@ def run_method_and_compare_outputs(
         for run_iteration in range(num_runs):
             reference_input = inputs if inputs else next(self.generate_random_inputs())
 
+            # Avoid issues with inplace operators
```
digantdesai, line 466: Interesting. I am sure we will run this in other places, but I see just Silu, which is good. :)

digantdesai (approved):
Thanks!

PR #14626, opened 2025-09-26
Arm backend: Add 6D tensor and pixel shuffle/unshuffle support

mansnils:
Adds 6D tensor support required by pixel_shuffle/pixel_unshuffle when given 4D inputs, which means for now we only support 4D inputs. Adds TOSA, VGF and xfailing Ethos-U85 unit tests.


digantdesai (approved):
Nice!

PR #14633, opened 2025-09-26
Try to get nightly wheel build work with qnn

cccclai:
Our current nightly/release wheel package is done following https://github.com/pytorch/test-infra/wiki/Using-Nova-Reusable-Build-Workflows

As described by https://github.com/pytorch/test-infra/blob/5398e1a00c39939f43251f29031c37e6d0c84647/.github/workflows/build_wheels_linux.yml#L4, The docker image infra team used to release nightly/release package is from https://github.com/pypa/manylinux, and it's currently using https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux_2_28-almalinux-8-based. It means the glibc version is 2.28 and GCC is 14. 

The issue is that, QNN .so files are not compatible with 2.28. The minimum version is 2.34 (I tried 2.29 the first time when it failed and asked for 2.29, but it still fails). 

In this PR, instead of checking glibc and failed directly when minimum version isn't matched, we will download the glibc 2.34 to /tmp. A different strategy compared with glibc++ is that, we don't load them, because the python process itself start with the system glibc 2.28. We need to re-execute the process with the new glibc

### DIFF
--- a/backends/qualcomm/scripts/download_qnn_sdk.py
+++ b/backends/qualcomm/scripts/download_qnn_sdk.py
@@ -437,8 +437,7 @@ def install_qnn_sdk() -> bool:
     Returns:
         True if both steps succeeded (or were already satisfied), else False.
     """
-    if check_glibc_exist_and_validate():
-        if _ensure_libcxx_stack():
-            if _ensure_qnn_sdk_lib():
-                return True
+    if _ensure_libcxx_stack():
+        if _ensure_qnn_sdk_lib():
+            return True
     return False
--- a/backends/qualcomm/scripts/download_qnn_sdk.py
+++ b/backends/qualcomm/scripts/download_qnn_sdk.py
@@ -300,14 +300,17 @@ def _install_glibc_234():
 
         logger.info("[glibc] Configuring build...")
         env = os.environ.copy()
+
         # Remove LD_LIBRARY_PATH to satisfy configure checks
         env.pop("LD_LIBRARY_PATH", None)
-        # allow warnings
-        # env["CFLAGS"] = "-O2 -Wno-error"
-        # env["CFLAGS"] = "-O2 -Wno-error"  # prevent warnings-as-errors
-        # env["CC"] = "gcc"  # explicit, avoids surprises
+
+        # Allow warnings (disable -Werror promoted by newer GCC)
+        # Add -fcommon to avoid "multiple definition" issues on some glibc versions
         env["CFLAGS"] = (
-            "-O2 -Wno-error=array-parameter -Wno-error=stringop-overflow -Wno-error"
+            "-O2 -fPIC -fcommon "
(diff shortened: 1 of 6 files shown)


backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -241,6 +182,128 @@ def _extract_tar(archive_path: pathlib.Path, prefix: str, target_dir: pathlib.Pa
...
+GLIBC_LIBDIR = GLIBC_ROOT / "lib"
+GLIBC_REEXEC_GUARD = "QNN_GLIBC_REEXEC"
+MINIMUM_LIBC_VERSION = GLIBC_VERSION
+
+RPM_URL = (
+    "https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/35/"
+    "Everything/x86_64/os/Packages/g/glibc-2.34-7.fc35.x86_64.rpm"
+)
+RPM_PATH = pathlib.Path("/tmp/glibc.rpm")
+WORKDIR = pathlib.Path("/tmp/glibc-extracted")
```
kimishpatel: Can you use tmpdir pacakge?

cccclai: I think this code currently applies to both when building the wheel package and user install the wheel package. If we use tmpdir package, it will affect the user's side. The path will be different every time users run this code and they need to do the process again.

kimishpatel: Wait what do you mean? Even when users do pip install setup.py, storing in this temp artifact is only for the during of `pip insta...`. Is that right?

cccclai: > users do pip install setup.py

Do you mean user install or use wheel package? Like do you mean 1) `pip install executorch_xx.whl`  or  2) building the wheel packages (`python setup bdist`)?

backends/qualcomm/scripts/download_qnn_sdk.py
```diff
@@ -34,68 +37,59 @@ def is_linux_x86() -> bool:
...
+        # On Windows, prefer %LOCALAPPDATA%, fallback to ~/AppData/Local
+        base = pathlib.Path(
+            os.environ.get("LOCALAPPDATA", pathlib.Path.home() / "AppData" / "Local")
         )
-    return exists
+    elif is_linux_x86():
+        # On Linux/Unix, prefer $XDG_CACHE_HOME, fallback to ~/.cache
+        base = pathlib.Path(
+            os.environ.get("XDG_CACHE_HOME", pathlib.Path.home() / ".cache")
+        )
```
kimishpatel, line 83: I liked this

kimishpatel (approved):
Thanks Chen for the changes. Looks good

PR #14636, opened 2025-09-26
Pin MacOS runners to MacOS 14

GregoryComer:
### Summary
A recent change seems to have updated macos-m1-stable to OS 15 and broken many ET jobs due to an issue building torchaudio with Clang 16 (see https://github.com/pytorch/executorch/actions/runs/18023896317/job/51287146415#step:9:46287). This appears to be unique to the release branch. Perhaps the issue was fixed on the development branch of torchaudio or somewhere upstream.

To unbreak the release branch, I'm trying pinning the runner version to "macos-m1-14" instead of "macos-m1-stable".

### Test plan
I'm running trunk CI on this job. If all the MacOS jobs pass, then we're good.

### DIFF
--- a/.ci/scripts/gather_test_models.py
+++ b/.ci/scripts/gather_test_models.py
@@ -18,7 +18,7 @@ from examples.xnnpack import MODEL_NAME_TO_OPTIONS, QuantType
 
 DEFAULT_RUNNERS = {
     "linux": "linux.2xlarge",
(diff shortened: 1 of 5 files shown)


GregoryComer:
OS 14 runners are still failing (they are still using Clang 16). Abandoning this PR in favor of https://github.com/pytorch/executorch/pull/14639.

PR #14639, opened 2025-09-26
[RELEASE ONLY CHANGES] Use prebuilt torch + domain libraries for MacOS jobs

GregoryComer:
### Summary
Skip building torch, torchvision, and torchaudio from source on the release branch. We install the pinned versions in install_requirements.py. This is an attempt to fix the broken CI on the release/1.0 branch, which is failing to install torchaudio from source. There seems to be an upstream incompatibility with torchaudio's current release/2.9 branch head and AppleClang 16+. Since we don't actually need to build from source on the release branch (we're pinning the versions), we can just skip the build from source.

### Test plan
CI, including trunk

### DIFF
--- a/.ci/scripts/utils.sh
+++ b/.ci/scripts/utils.sh
@@ -107,7 +107,7 @@ install_pytorch_and_domains() {
   TORCHVISION_VERSION=$(cat .github/ci_commit_pins/vision.txt)
   export TORCHVISION_VERSION
 
-  install_domains
+  #install_domains
 
   popd || return
   # Print sccache stats for debugging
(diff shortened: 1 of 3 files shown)


GregoryComer:
First attempt seems to break the build with errors in the torch headers. To try to unblock quickly, I'm going to try install torch from source (as we have been) but skip the domain libs.

GregoryComer:
Now failing in the ET pybind build with the same error:
```
[100%] Linking CXX shared library _portable_lib.cpython-311-darwin.so
  ld: duplicate LC_RPATH '@loader_path' in '/Users/ec2-user/runner/_work/_temp/conda_environment_18051662538/lib/python3.11/site-packages/torch/lib/libtorch_cpu.dylib'
  clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```
https://github.com/pytorch/executorch/actions/runs/18051662538/job/51374563715?pr=14639

Going to try linking with `-Wl,-ld_classic`. Seems like this is a known issue on PyTorch: https://github.com/pytorch/pytorch/issues/151592.

PR #14642, opened 2025-09-27
[Release only] Install torch, vision, and audio RC in the release branch

huydhn:
It's wrong to build torch, vision, and audio during release, we should just use their release candidates instead.  Not only we could avoid the need to rebuild these packages, it's also the correct setup to test against RC during release


mergennachin:
cc @GregoryComer , @psiddh, @abhinaykukkadapu please update the release manual and checklist with this step

GregoryComer:
Thanks! I see some AO failures on a few jobs, but it isn't obviously related. Maybe we were implicitly relying on some installation of AO. Given that the release branch is broken, I'm going to merge as is and take the few remaining failures as a fast follow-up.

GregoryComer (approved):

PR #14643, opened 2025-09-27
[multimodal] Allow generate and prefill to take move sematics

larryliu0820:
As titled

### DIFF
--- a/extension/llm/runner/multimodal_runner.cpp
+++ b/extension/llm/runner/multimodal_runner.cpp
@@ -62,6 +62,11 @@ Error MultimodalRunner::load() {
     ET_LOG(Info, format, __VA_ARGS__);     \
   }
 
+Error MultimodalRunner::prefill(std::vector<MultimodalInput>&& inputs) {
+  // Forward to the const reference version
+  return prefill(inputs);
+}
+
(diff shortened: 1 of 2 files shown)


shoumikhin (approved):

kirklandsign (commented):
Sorry I am still not sure about the use case. `const vector&` is a superset of `vector&&` ?

kirklandsign:
Or lost prefill cache>

larryliu0820:
> Sorry I am still not sure about the use case. `const vector&` is a superset of `vector&&` ?

If only `const vector&` then it doesn't take `std::move(inputs)`, basically rvalues. With this new API you can do stuff like:

```cpp

runner.generate({make_text_input("my prompt"), make_image_input(image_data)});
```

kirklandsign:
We can't do this with `const vector&`?
```
runner.generate({make_text_input("my prompt"), make_image_input(image_data)});
```

larryliu0820:
> We can't do this with `const vector&`?
> 
> ```
> runner.generate({make_text_input("my prompt"), make_image_input(image_data)});
> ```

Can't do this without this PR, we are creating an rvalue and generate doesn't take it.

PR #14645, opened 2025-09-28
Rescale add int16 correctly

3l1:
Differential Revision: D82906134


digantdesai (approved):

per (approved):

PR #14646, opened 2025-09-28
Update mul int16 test

3l1:
Differential Revision: D83437473


digantdesai (approved):

PR #14653, opened 2025-09-29
Arm backend: Backend test TOSA FP, INT and Ethos-U55/U85

zingo:
### Summary
Create arm_ethos_u55 and arm_ethos_u85 test flows and add them to CI
Build a semihosted runner for testing on the Corstone3x0 FVP
And split the arm_tosa test job that tested TOSA-1.0+FP into arm_tosa_fp and arm_tosa_int to also test TOSA-1.0+INT

### Test plan
This will add new tests for arm_tosa_int arm_ethos_u55 and arm_ethos_u85


backends/test/suite/flows/arm.py
```diff
@@ -1,24 +1,72 @@
...
+        return Quantize(quantizer, quantization_config)
 
-def _create_tosa_flow() -> TestFlow:
     return TestFlow(
-        "arm_tosa",
+        name,
         backend="arm",
-        tester_factory=_create_arm_tester_tosa_fp,
+        tester_factory=_create_arm_tester,
         supports_serialize=False,
```
digantdesai, line 47: Why not enable this? This should work if we can provide a runner somehow to the FVP? This might be an alternative to pybinding assuming we want to do it this way.

zingo: We hope to enable more and more here over time. Plan is to do more in later PR but would be nice to get som INT stuff into the flow already now 🙂

zingo: As for supports_serialize=False, specificalli, we have not looked deep enough yet to check it, its basicaly just left as is from the first PR by @GregoryComer

GregoryComer (approved):
Thanks. Looks good to me. FYI I did migrate the test suite to use pytest (https://github.com/pytorch/executorch/pull/14456), though this shouldn't affect your PR much, if at all.

This is an evolving effort, so feel free to make suggestions if the test harness or logic isn't working for your use case or if you have suggestions for improvements.

zingo:
Spotted the pytest that worked out great.
I have a quant change for EthosU on testing. Thanks for starting this framework. It makes sense to sycronize backend test.

PR #14655, opened 2025-09-29
Arm backend: Add VGF tests to StableDiffusion module tests

YufengShi-dudu:
Also refactor the StableDiffusion module tests to use test_pipeline instead of ArmTester directly.

Change-Id: Iac07da7edc5a638d95c9bb6ad4b134c44aedac05


backends/arm/test/models/stable_diffusion/test_CLIPTextModelWithProjection.py
```diff
@@ -69,47 +75,93 @@ def prepare_model_and_inputs(self):
...
+    )
+    with torch.no_grad():
+        pipeline = VgfPipeline[input_t](
+            text_encoder_model,
+            text_encoder_model_inputs,
+            aten_op=[],
+            exir_op=[],
+            tosa_version="TOSA-1.0+FP",
+            use_to_edge_transform_and_lower=True,
+            atol=4,  # TODO: Investiage numerical issue: MAX Diff ~50%
```
digantdesai, line 135: Let's make sure this test isn't flaky.

digantdesai (approved):

backends/arm/test/models/stable_diffusion/test_CLIPTextModelWithProjection.py
YufengShi-dudu: Yes, that's a good point. I tested it locally and 4 should be enough. But we can increase it later if we see flaky issue upstream.

PR #14657, opened 2025-09-29
Arm backend: Add test for monitoring memory allocation

perheld:
Simple test to monitor memory allocations when running the "add" model in fvp.

Change-Id: Ib1a47fc8236dab4ae70f49fda2294f6e1503f62b

### DIFF
--- a/backends/arm/test/test_arm_baremetal.sh
+++ b/backends/arm/test/test_arm_baremetal.sh
@@ -366,5 +366,20 @@ test_smaller_stories_llama() {
     echo "${TEST_SUITE_NAME}: PASS"
 }
 
+test_memory_allocation() {
+    echo "${TEST_SUITE_NAME}: Test ethos-u memory allocation with run.sh"
+
+    mkdir -p arm_test/test_run
+    # Ethos-U85
+    echo "${TEST_SUITE_NAME}: Test target Ethos-U85"
+    examples/arm/run.sh --et_build_root=arm_test/test_run --target=ethos-u85-128 --model_name=examples/arm/example_modules/add.py &> arm_test/test_run/full.log
+    python3 backends/arm/test/test_memory_allocator_log.py --log arm_test/test_run/full.log \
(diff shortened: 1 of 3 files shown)


perheld:
Ouch, readonly fs. Didnt think about that. Need to change the pythonscript so it can parse through stdin and just have the results piped into it. Tomorrows work.

perheld:
> Ouch, readonly fs. Didnt think about that. Need to change the pythonscript so it can parse through stdin and just have the results piped into it. Tomorrows work.

Nope, turns out I was trying to write to the dir before it exists. Added a mkdir for it.

perheld:
@digantdesai I need some input what you think about this since I'm touching trunk.yml

backends/arm/test/test_arm_baremetal.sh
```diff
@@ -366,5 +366,20 @@ test_smaller_stories_llama() {
     echo "${TEST_SUITE_NAME}: PASS"
 }
 
+test_memory_allocation() {
+    echo "${TEST_SUITE_NAME}: Test ethos-u memory allocation with run.sh"
+
+    mkdir -p arm_test/test_run
+    # Ethos-U85
+    echo "${TEST_SUITE_NAME}: Test target Ethos-U85"
+    examples/arm/run.sh --et_build_root=arm_test/test_run --target=ethos-u85-128 --model_name=examples/arm/example_modules/add.py &> arm_test/test_run/full.log
```
digantdesai, line 375: are you planning to extend this to more "interesting" cases, like MV2 (with weights and intermediate tensors) or something which doesn't fully delegate etc.

digantdesai (approved):
❤️

backends/arm/test/test_arm_baremetal.sh
zingo: Yes  at least some bigger model is planned maybe mv2, we didn't think of not fully delegated, mainly as we intend to delegate it all :) But could be interesting, maybe we add one of we find a good candidate.

PR #14663, opened 2025-09-29
delete qnnpack backend

JacobSzwejbka:
Summary: Afaik this is unused and not maintained

Reviewed By: larryliu0820

Differential Revision: D83448512


larryliu0820 (approved):

cccclai:
why are we deleting this? also can we at least move the test internal? This is the only test covered two backends

PR #14674, opened 2025-09-29
Android set different maven package names of flavors

kirklandsign:
Different flavor name generates different maven packages


shoumikhin (approved):

mergennachin:
@kirklandsign cherry-pick?

PR #14675, opened 2025-09-29
support qnn mean (dim=None)

cccclai:
Summary: Address mean op lower failure. When dim is not specified, it will take mean across all axes. For QNN, we need to get axes based on input shape

Differential Revision: D83520776

### DIFF
--- a/backends/qualcomm/builders/op_mean_dim.py
+++ b/backends/qualcomm/builders/op_mean_dim.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import cast, Dict, List
+from typing import cast, Dict
 
 import executorch.backends.qualcomm.python.PyQnnWrapperAdaptor as PyQnnWrapper
 
@@ -40,7 +40,22 @@ class MeanDim(NodeVisitor):
         )
 
         # mean dims and keep dims
-        mean_dims = cast(List[int], node.args[1])
+        rank = len(input_node.meta["val"].shape)
+
+        if rank == 0:
+            raise RuntimeError(
+                "Mean doesn't support 0d input, please report a bug in https://github.com/pytorch/executorch/issues"
+            )
+
+        dim_arg = node.args[1]
+
+        if dim_arg is None or len(dim_arg) == 0:
+            mean_dims = list(range(rank))  # reduce over all dims
+        elif isinstance(dim_arg, int):
+            mean_dims = [dim_arg]
+        else:
+            mean_dims = list(dim_arg)
+        print("mean_dims: ", mean_dims, "rank: ", rank)
         mean_dims = [
             mean_dim % len(input_node.meta["val"].shape) for mean_dim in mean_dims
         ]
--- a/backends/qualcomm/tests/models.py
+++ b/backends/qualcomm/tests/models.py
@@ -4,8 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import torch
+from typing import List, Optional, Tuple, Union
 
+import torch
 
 # module with related operator only
 
@@ -1332,20 +1333,20 @@ class MaxPool2d(torch.nn.Module):
         return self.max_pool2d(x)
 
 
-class MeanWKeppDim(torch.nn.Module):
-    def __init__(self):
+class Mean(torch.nn.Module):
+    def __init__(
+        self,
+        dim: Optional[Union[int, Tuple[int, ...], List[int]]] = None,
+        keepdim: bool = False,
+        dtype: Optional[torch.dtype] = None,
+    ):
         super().__init__()
+        self.dim = dim
+        self.keepdim = keepdim
+        self.dtype = dtype
 
     def forward(self, x):
-        return torch.mean(x, (-1, -2), keepdim=True)
-
-
-class MeanWOKeppDim(torch.nn.Module):
-    def __init__(self):
-        super().__init__()
-
-    def forward(self, x):
-        return torch.mean(x, (-1, -2))
+        return torch.mean(x, dim=self.dim, keepdim=self.keepdim, dtype=self.dtype)
 
 
 class MaskedFill(torch.nn.Module):
--- a/backends/qualcomm/tests/test_qnn_delegate.py
+++ b/backends/qualcomm/tests/test_qnn_delegate.py
@@ -1018,12 +1018,61 @@ class TestQNNFloatingPointOperator(TestQNN):
         sample_input = (torch.randn(4, 3, 24, 24),)
         self.lower_module_and_test_output(module, sample_input)
 
-    def test_qnn_backend_mean_dim(self):
-        modules = [MeanWKeppDim(), MeanWOKeppDim()]  # noqa: F405
-        sample_input = (torch.randn([2, 5, 1, 3]),)
-        for i, module in enumerate(modules):
+    def test_qnn_backend_mean(self):
+        test_comb = [
+            # Reduce over last two dims, keepdim=True
+            {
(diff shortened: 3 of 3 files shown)


backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -1025,6 +1025,13 @@ def test_qnn_backend_mean_dim(self):
             with self.subTest(i=i):
                 self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_mean(self):
+        modules = [Mean(), Mean()]  # noqa: F405
```
haowhsu-quic: Do you need more configurations here? like `Mean(dim=0)`, `Mean(dim=0, keepdim=True)`.

backends/qualcomm/tests/models.py
```diff
@@ -1348,6 +1348,22 @@ def forward(self, x):
         return torch.mean(x, (-1, -2))
 
 
+class Mean(torch.nn.Module):
```
haowhsu-quic, line 1336: Could you help remove the legacy models / tests for `MeanWOKeppDim`, `MeanWKeppDim`? thanks.

cccclai: Yes updated

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: updated

cccclai:
Does it look good? I think it can fix 5 failing op tests

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -2666,13 +2710,59 @@ def test_qnn_backend_max_pool2d(self):
...
+
+        for i, test in enumerate(test_comb):
             with self.subTest(i=i):
-                module = self.get_qdq_module(module, sample_input)
-                self.lower_module_and_test_output(module, sample_input)
+                print("test i", i)
+                module = self.get_qdq_module(
+                    test[QCOM_MODULE], test[QCOM_SAMPLE_INPUTS]
+                )
+                module = self.get_qdq_module(module, test[QCOM_SAMPLE_INPUTS])
```
haowhsu-quic: Might need to remove L2759~2762, thanks.

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -1018,12 +1018,56 @@ def test_qnn_backend_max_pool2d(self):
...
+            {
+                QCOM_MODULE: Mean(dim=(1, 3), keepdim=True),
+                QCOM_SAMPLE_INPUTS: (torch.randn(2, 3, 4, 5, 6),),
+            },
+        ]
+
+        for i, test in enumerate(test_comb):
             with self.subTest(i=i):
-                self.lower_module_and_test_output(module, sample_input)
+                module = self.get_qdq_module(
```
haowhsu-quic: I think we don't need `get_qdq_module` for floating point test.

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: Good catch, thank you for the review

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -1018,12 +1018,53 @@ def test_qnn_backend_max_pool2d(self):
...
+            },
+            # Default: reduce all dims
+            {
+                QCOM_MODULE: Mean(),
+                QCOM_SAMPLE_INPUTS: (torch.randn(10, 10),),
+            },
+            # Scalar case
+            {
+                QCOM_MODULE: Mean(),
+                QCOM_SAMPLE_INPUTS: (torch.tensor([5.0]),),
```
chenweng-quic: Should it be torch.tensor(5.0) if you want to test scalar?

backends/qualcomm/builders/op_mean_dim.py
```diff
@@ -40,7 +40,16 @@ def define_node(
         )
 
         # mean dims and keep dims
-        mean_dims = cast(List[int], node.args[1])
+        rank = len(input_node.meta["val"].shape)
+        dim_arg = node.args[1]
+
+        if dim_arg is None:
```
chenweng-quic: Maybe edge case, could dim_arg be empty list?

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -2666,13 +2707,55 @@ def test_qnn_backend_max_pool2d(self):
...
+                QCOM_MODULE: Mean(dim=(1, 3), keepdim=True),
+                QCOM_SAMPLE_INPUTS: (torch.randn(2, 3, 4, 5, 6),),
+            },
+        ]
+
+        for i, test in enumerate(test_comb):
             with self.subTest(i=i):
-                module = self.get_qdq_module(module, sample_input)
-                self.lower_module_and_test_output(module, sample_input)
+                module = self.get_qdq_module(module, test[QCOM_SAMPLE_INPUTS])
```
haowhsu-quic: I guess this should be `module = self.get_qdq_module(test[QCOM_MODULE], test[QCOM_SAMPLE_INPUTS])`

backends/qualcomm/builders/op_mean_dim.py
cccclai: Yes I verify the eager mode and it does support empty list. Updated

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: Thanks, good catch

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: Sorry I should be more careful..

backends/qualcomm/tests/test_qnn_delegate.py
cccclai: Actually this is indeed a good test. ReduceMean doesn't support 0d tensor, I think we need to have a pass to convert 0d to 1d tensor if the input of mean is 0d. I have some sketch code here. What do you think? For now, I comment out the test case and think we can follow up on this
```
import torch
from executorch.exir.pass_base import ExportPass, PassResult

class Rank0ToRank1(ExportPass):
    """
    For selected ops and selected input positions, if the input is rank-0 (scalar),
    insert a reshape to [1] before the op.
    """

    def __init__(self, op_input_map=None) -> None:
        super().__init__()
        # key is the op, value is the input indices to be reshaped
        self.op_input_map = {
          torch.ops.aten.mean.dim: [0],
        }

    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
        graph = graph_module.graph
        changed = False

        for node in list(graph.nodes):
            if node.op == "call_function" and node.target in self.op_input_map:
                input_indices = self.op_input_map[node.target]

                new_args = list(node.args)
                for idx in input_indices:
                    if idx < len(new_args):
                        inp = new_args[idx]
                        if hasattr(inp, "meta"):
                            val = inp.meta.get("val", None)
                            if val is not None and hasattr(val, "shape") and val.shape == ():
                                # Insert reshape right before the op
                                with graph.inserting_before(node):
                                    reshape_node = graph.call_function(
                                        torch.ops.aten.reshape.default,
                                        args=(inp, (1,))
                                    )
                                    reshape_node.meta["val"] = val.reshape(1,)

                                # Replace arg idx with reshape_node
                                new_args[idx] = reshape_node
                                changed = True

                # update node args if modified
                node.args = tuple(new_args)

        if changed:
            graph_module.recompile()
        return PassResult(graph_module, changed)

```

chenweng-quic (approved):

kirklandsign (approved):

PR #14654, opened 2025-09-29
refactor:(samsung backend): replace pkg_resources with importlib.resources for schema loading

onuralpszr:
This PR refactors the Samsung backend schema loading logic in compile_options.py by replacing pkg_resources with importlib.resources. This modernizes resource access, improves compatibility with Python packaging standards, and removes the dependency on setuptools. No functional changes to the compile options logic; only the resource loading mechanism is updated.

PR #14691, opened 2025-09-30
Fix deprecation of pkg_resources

tmi:
### Summary

Replaces deprecated `pkg_resources` (which is scheduled for removal in 2025-11-30 !) with `importlib.resources` -- part of python stdlib since like python 3.9. No observable change expected (beyond disappearance of a bunch of warnings)

Note: there's other pkg_resources usage coming from https://github.com/facebook/buck2-prelude, I'll make a PR there too

### Test plan
Multiple existing tests rely on this function, so I assume the test coverage is sufficient

### DIFF
--- a/backends/samsung/serialization/compile_options.py
+++ b/backends/samsung/serialization/compile_options.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import importlib.resources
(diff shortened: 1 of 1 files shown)


mergennachin:
Looks like a duplicate PR: https://github.com/pytorch/executorch/pull/14654

tmi:
ah narrowly missed that one, apologies & thx for the notice

PR #14694, opened 2025-09-30
Arm backend: Add TOSA dialect op for MATMUL

oscarandersson8218:
Adds TOSA backend dialect op for MATMUL and associating pass to rewrite edge.aten.bmm to tosa.MATMUL.


backends/arm/_passes/rewrite_matmul.py
```diff
@@ -0,0 +1,87 @@
...
+            tosa_matmul_node.replace_all_uses_with(rescale_node)
+            rescale_node.args = (
+                tosa_matmul_node,
+                torch.int8,
+                scale,
+                0,
+                output_qparams.get_zp_per_tensor(),
+            )
+
+    def call(self, graph_module):
```
digantdesai, line 52: curious why not do this in the aten.bmm node visitor?

oscarandersson8218: aten.bmm is not a one-to-one mapping to TOSA.MATMUL due to TOSA outputting an INT32 for INT8 inputs while aten.bmm would output an INT8. So a rescale needs to be inserted and we're trying to avoid inserting rescales in multiple node visitors.

digantdesai: > we're trying to avoid inserting rescales in multiple node visitor

Yeah I saw #14584 

I called this out is because this pass feels a lot like bmm lowering which we have been doing through the node visitors.

backends/arm/operators/op_tosa_matmul.py
```diff
@@ -26,20 +25,13 @@
...
+class MatmulVisitor(NodeVisitor):
+    """Provide a visitor that serializes TOSA ``MATMUL``."""
 
-    INT8 accumulates into INT32; add a rescale to INT8 using SINGLE_ROUND
-    rounding and output zero-point.
-
-    """
-
-    target = "aten.bmm.default"
+    target = "tosa.MATMUL.default"
```
digantdesai, line 34: nit: perhaps we should rename this file a bit further to `op_tosa_matmul.py` and others, if any, just to distinguish from aten op target lowering?

digantdesai (approved):

backends/arm/operators/op_tosa_matmul.py
oscarandersson8218: Sounds good to me!

PR #14710, opened 2025-09-30
support argmax/argmin without dim kwargs and fix adaptive_max_pool3d

cccclai:
Summary: As title, in PyTorch, when dim is not set, it will flatten the input and get argmax as dim=0. Add a pass to reshape the input when dim is not set and consolidate test case

edit:
1. Apply to argmin too
2. Add `exir_ops.edge.aten.adaptive_max_pool3d.default` to the to be implemented op list to pass the error

Differential Revision: D83606497

### DIFF
--- a/backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
+++ b/backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
@@ -1,0 +1,59 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from executorch.exir.pass_base import ExportPass, PassResult
+from executorch.exir.passes import dead_code_elimination_pass
+
+
+class InsertReshapeForReduceOps(ExportPass):
+    """
+    Rewrite `aten.argmax.default` with `dim=None` into
+    a reshape-to-1D followed by argmax(dim=0).
+
+    PyTorch semantics:
+      torch.argmax(x, dim=None) -> flatten(x) then argmax along axis=0
+
+    QNN requires an explicit axis, so we insert the reshape.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.op_map = {torch.ops.aten.argmax.default, torch.ops.aten.argmin.default}
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        graph = graph_module.graph
+        modified = False
+
+        for n in graph.nodes:
+            if n.target in self.op_map:
+                dim_arg = None if len(n.args) == 1 else n.args[1]
+
+                if dim_arg is None:
+                    inp = n.args[0]
+
+                    # Insert reshape before argmax
+                    with graph.inserting_before(n):
+                        reshape_node = graph.create_node(
+                            "call_function",
+                            torch.ops.aten.reshape.default,
+                            (inp, [-1]),
+                            {},
+                        )
+                        reshape_node.meta = dict(inp.meta)
+                        if "val" in inp.meta:
+                            reshape_node.meta["val"] = inp.meta["val"].reshape(-1)
+
+                    # Rewrite argmax: take reshape_node as input, set dim=0
+                    n.args = (reshape_node, 0, *n.args[2:])
+
+                modified = True
+
+        if modified:
+            graph_module.recompile()
+            dead_code_elimination_pass(graph_module)
+
+        return PassResult(graph_module, modified)
--- a/backends/qualcomm/partition/common_defs.py
+++ b/backends/qualcomm/partition/common_defs.py
@@ -17,6 +17,7 @@ not_supported_operator = [
 to_be_implemented_operator = [
     exir_ops.edge.aten._adaptive_avg_pool3d.default,
     exir_ops.edge.aten.adaptive_max_pool2d.default,
+    exir_ops.edge.aten.adaptive_max_pool3d.default,
     exir_ops.edge.aten.avg_pool3d.default,
     exir_ops.edge.aten.div.Tensor_mode,
     exir_ops.edge.aten.log10.default,
--- a/backends/qualcomm/tests/test_passes.py
+++ b/backends/qualcomm/tests/test_passes.py
@@ -1,0 +1,54 @@
+import unittest
+
+import torch
+from executorch.backends.qualcomm._passes import InsertReshapeForReduceOps
+
+
+class TestPasses(unittest.TestCase):
+    def test_insert_reshape_for_argmax(self):
+        class ArgmaxModule(torch.nn.Module):
+            def forward(self, x):
+                return torch.argmax(x, dim=None)
+
+        mod = ArgmaxModule()
+
+        x = torch.tensor([[1.0, 5.0], [3.0, 2.0]])
+        ep = torch.export.export(mod, (x,))
+        # Run original module for reference
+        ref = mod(x)
+
+        reshape_nodes = [
+            n for n in ep.graph.nodes if n.target == torch.ops.aten.reshape.default
+        ]
+        argmax_nodes = [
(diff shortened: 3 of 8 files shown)


backends/qualcomm/_passes/insert_reshape_for_argmax.py
```diff
@@ -0,0 +1,59 @@
...
+# All rights reserved
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from executorch.exir.pass_base import ExportPass, PassResult
+from executorch.exir.passes import dead_code_elimination_pass
+
+class InsertReshapeForArgmax(ExportPass):
```
haowhsu-quic: I think we might have other ops like `argmin` which can leverage this pass as well. Do you mind rewording the class name into more generic one? Something like `InsertReshapeForReduceOp`.

backends/qualcomm/_passes/insert_reshape_for_argmax.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.op_map = {torch.ops.aten.argmax.default}
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        graph = graph_module.graph
+        modified = False
+
+        for n in list(graph.nodes):
```
haowhsu-quic: I think we can use the iterator directly: `for n in graph.nodes:`.

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
```diff
@@ -0,0 +1,59 @@
...
+                dim_arg = None if len(n.args) == 1 else n.args[1]
+
+                if dim_arg is None:
+                    inp = n.args[0]
+
+                    # Insert reshape before argmax
+                    with graph.inserting_before(n):
+                        reshape_node = graph.create_node(
+                            "call_function",
+                            torch.ops.aten.reshape.default,
```
haowhsu-quic, line 42: I think we use `torch` namespace for graph which needs decomposition before `prepare_pt2e` and `exir_ops` namespace for floating point graph which goes directly to `to_edge_transform_and_lower` API.

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
```diff
@@ -0,0 +1,59 @@
...
+                            {},
+                        )
+                        reshape_node.meta = dict(inp.meta)
+                        if "val" in inp.meta:
+                            reshape_node.meta["val"] = inp.meta["val"].reshape(-1)
+
+                    # Rewrite argmax: take reshape_node as input, set dim=0
+                    n.args = (reshape_node, 0, *n.args[2:])
+                
+                modified = True
```
haowhsu-quic: I think you need one more indention to have this statement fall into branch.

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
cccclai: I see, do you mean I need to use the op with different namespace here?

backends/qualcomm/tests/test_passes.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+import torch
+from executorch.backends.qualcomm._passes import InsertReshapeForArgmax
+
+class TestPasses(unittest.TestCase):
```
haowhsu-quic, line 7: Thank you for adding this file, it's helpful for us to test all the passes thoroughly.

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
haowhsu-quic: It's a convention we're using but I guess the op builder could still work since it only recognizes part of op's target string.

cccclai: I was actually using it for both `transform_for_annotation_pipeline` and `transform_for_export_pipeline`. I was thinking it could be a good idea to have it in preprocess, but preprocess happens after partition, and op_argmax.py will run during partition as well. I think in this case, we need to have it similar to those decompose passes, which means we use `torch` namespace

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
cccclai: updated

backends/qualcomm/_passes/insert_reshape_for_argmax.py
cccclai: updated

backends/qualcomm/_passes/insert_reshape_for_argmax.py
cccclai: updated

winskuo-quic: I think the new commit wasn't pushed successfully. Could you try again? Thanks

cccclai: oh sorry, just push

backends/qualcomm/builders/op_argmax.py
```diff
@@ -46,7 +46,7 @@ def define_node(
             nodes_to_wrappers,
         )
         argmax_output_tensors = [argmax_out_tensor_wrapper]
-
+        print("argmax node.args: ", node.args)
```
mergennachin: remove

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) Qualcomm Innovation Center, Inc.
```
mergennachin: You, representing Meta, contributed this file, right? If so, add Meta copyright

backends/qualcomm/builders/op_argmax.py
cccclai: updated

backends/qualcomm/_passes/insert_reshape_for_reduce_ops.py
cccclai: updated

cccclai:
Can I get another round of review on this? I fix a few more failing tests including `argmin`, and `adaptive_max_pool3d`

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -173,14 +173,60 @@ def test_qnn_backend_arange(self):
                 self.lower_module_and_test_output(module, sample_input)
 
     def test_qnn_backend_argmax(self):
-        module = Argmax()  # noqa: F405
-        sample_input = (torch.randn(16, 3, 4, 4),)
-        self.lower_module_and_test_output(module, sample_input)
+        test_cases = [
+            {
+                "module": Argmax(),  # noqa: F405
```
haowhsu-quic: I think we have keywords such as "QCOM_MODULE", "QCOM_SAMPLE_INPUTS". It would be great if you could use those.

cccclai: Yeah let me update

haowhsu-quic (approved):
Thank you!

backends/qualcomm/partition/common_defs.py
```diff
@@ -17,6 +17,7 @@
 to_be_implemented_operator = [
     exir_ops.edge.aten._adaptive_avg_pool3d.default,
     exir_ops.edge.aten.adaptive_max_pool2d.default,
+    exir_ops.edge.aten.adaptive_max_pool3d.default,
```
kimishpatel, line 20: is this pr just about argmax? i see maxpool added here. if it is about reduce ops then please do update the title

kimishpatel (approved):

backends/qualcomm/partition/common_defs.py
cccclai: Yeah updated

PR #14711, opened 2025-10-01
Support sine operator on XNNPACK

GregoryComer:
Summary: Wire up the unary sine operator in xnnpack for fp32 and fp16.

Differential Revision: D83623086


GregoryComer:
Cosine is coming in D83623619, will export as a PR once this is merged because I didn't use ghexport.

digantdesai (approved):

digantdesai:
cherry pick on to 1.0?

GregoryComer:
> cherry pick on to 1.0?

I wasn't originally planning to, given that it's late, but it's reasonably low risk so I'm okay to pick it if you'd like.

backends/xnnpack/test/ops/test_sin.py
```diff
@@ -0,0 +1,87 @@
...
+            ),
+        )
+        self._test_sin(inputs, legacy_mode=False)
+
+    def test_fp32_sin_legacy_mode(self):
+        inputs = (
+            torch.Tensor(
+                [
+                    [0.0, 0.1, 0.5, 0.785398],
+                    [-0.5, -0.785398, 1.5708, -1.5708],
```
digantdesai, line 83: XNNPACK does use a faster approximation algorithm, and does have [tests](https://github.com/google/XNNPACK/blob/da9a34d9bb68f339c35d2da480ab0734b0a26429/test/unary-ops.h#L420-L443) against std::sin(), with atol/rotl = `3*std::numeric_limits<T>::epsilon()` and `5*std::numeric_limits<T>::epsilon()`, which is also used by ET::Portable::sin(), so some tolerance will be required. I guess we can increase the tensor sizes here and perhaps find the tolerance required, and live with that for now?

PR #14715, opened 2025-10-01
Aoti support multi method

larryliu0820:
This pull request introduces several improvements to the CUDA backend. The main changes include adding a new graph pass to replace unnecessary `slice_copy` operations, improving how method names are tracked in compilation artifacts, and making the preprocessing pipeline more robust and accurate.

**Key changes:**

### Graph optimization and preprocessing

* Introduced `ReplaceSliceCopyWithSlicePass`, a new export pass that replaces non-mutated `slice_copy` operations with more efficient `slice` view operations in the computational graph (`replace_slice_copy_with_slice.py`, used in `cuda_backend.py`). [[1]](diffhunk://#diff-c4a228b182f50f778545991d472609ad705d2325994342174093ff374738851dR1-R113) [[2]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aR115-R117)
* Added context management for attention kernel selection and no-grad mode during AOT compilation to ensure correct backend selection for decomposition. This is needed in the short term until we have a flash attention cuda kernel.

### Method name and compile specification handling

* Added a `COMPILE_SPEC_KEYS` enum and utility methods (`generate_method_name_compile_spec`, `method_name_from_compile_specs`) to consistently embed and retrieve the method name in compile specs and as a key in the data store, improving traceability of compiled artifacts. [[1]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aL24-R35) [[2]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aL161-R158) [[3]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aR169-R195)

### Code cleanup and maintainability

* Minor refactor in `cuda_partitioner.py` to clarify delegation tag assignment.
* Improved imports and code organization for clarity in `cuda_backend.py`.

These changes collectively improve the reliability, performance, and maintainability of the CUDA backend pipeline.

### DIFF
--- a/backends/cuda/cuda_backend.py
+++ b/backends/cuda/cuda_backend.py
@@ -12,7 +12,9 @@ from enum import Enum
 from typing import Any, Dict, final, List, Optional, Set
 
 import torch
-from executorch.backends.cuda.replace_slice_copy_with_slice import ReplaceSliceCopyWithSlicePass
+from executorch.backends.cuda.replace_slice_copy_with_slice import (
+    ReplaceSliceCopyWithSlicePass,
+)
 from executorch.exir._serialize._named_data_store import NamedDataStore
 from executorch.exir._warnings import experimental
 from executorch.exir.backend.backend_details import (
@@ -31,8 +33,10 @@ supported_fallback_kernels: Dict[str, Any] = {}
 # required fallback kernels but not supported
 missing_fallback_kernels: Set[str] = set()
 
+
 class COMPILE_SPEC_KEYS(Enum):
     METHOD_NAME = "method_name"
+
 
 # context manager for non-fallback guarantee
 # it will raise exception when generating fallback kernels during aoti compile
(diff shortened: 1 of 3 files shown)


backends/cuda/cuda_backend.py
```diff
@@ -108,6 +112,9 @@ def preprocess(
         # Move the edge_program from CPU to CUDA for aoti compile
         cuda_edge_program = move_to_device_pass(edge_program, "cuda")
 
+        # replace slice_copy with slice
+        ReplaceSliceCopyWithSlicePass()(cuda_edge_program.graph_module)
```
JacobSzwejbka, line 120: Seems pretty hacky to run the force functionalization pass and then come through and undo it (but only for slice). Wont you in practice have to do this for all view ops?

Does AOTI lowering typically happen on functionalized IR?

backends/cuda/cuda_backend.py
```diff
@@ -132,7 +139,8 @@ def preprocess(
             "max_autotune_conv_backends": "TRITON",
         }
 
-        with collect_unsupported_fallback_kernels():
+        with collect_unsupported_fallback_kernels(), torch.nn.attention.sdpa_kernel([SDPBackend.MATH]), torch.no_grad():
+            torch._logging.set_logs(post_grad_graphs=True)
```
JacobSzwejbka: Should this be landed?

JacobSzwejbka (approved):

backends/cuda/cuda_backend.py
larryliu0820: Inductor's reinplace pass reverts most of the functionalization. I don't think it handles `slice_copy` though, since it comes from this pass: https://github.com/pytorch/executorch/blob/main/exir/passes/replace_broken_ops_with_function_ops_pass.py#L13

The other option we can do is to optionally run this pass in `to_edge()`.

backends/cuda/cuda_backend.py
larryliu0820: I'll revert

cccclai:
I think method name should be in the preprocess_blob instead of the compile spec....if you use compile spec, users need to pass in the graph name as part of the compile spec every time, while in preprocess blob, users don't need to.

PR #14716, opened 2025-10-01
Add Phi4 test and fix regex parsing.

shoumikhin:
Summary:
- Add Phi4 test
- Handle the invert flag properly
- Skip regex lookahead since it's not supported by RE2
- Catch std::exception instead of narrow json parser errors
- Make the special tokens arg optional in ObjC/Swift API

Differential Revision: D83641294


mergennachin (approved):

larryliu0820 (approved):

PR #14721, opened 2025-10-01
Arm backend: Convert ExportPasses to ArmPasses

martinlsm:
All passes in backends/arm/_passes have been set to inherit from `ArmPass` instead of `ExportPass`.

Furthermore, the attribute `exported_program` in `ArmPass` has been removed to instead let each subclasses set it. The reason is that it was type annotated as an `Optional[ExportedProgram]]` in `ArmPass` and Mypy complained about that when a subclass tried to access it without checking for `None`. By moving the attribute down to each subclass that needs it, the confusion around whether the value is `None` or not is elimininated.

### Test plan
The persistent tests in the Arm backend should cover this change.

### DIFF
--- a/backends/arm/_passes/arm_pass.py
+++ b/backends/arm/_passes/arm_pass.py
@@ -9,16 +9,11 @@ import traceback
 from abc import abstractmethod
 from typing import List, Optional, Set, Type
 
-import torch
 from executorch.exir.pass_base import ExportPass, NodeMetadata
 
 
 class ArmPass(ExportPass):
     """Base class for Arm passes"""
-
-    def __init__(self, exported_program: Optional[torch.export.ExportedProgram] = None):
-        super(ArmPass, self).__init__()
-        self.exported_program = exported_program
 
     @property
     @abstractmethod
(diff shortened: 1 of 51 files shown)


backends/arm/_passes/arm_pass.py
```diff
@@ -9,17 +9,12 @@
...
 from typing import List, Optional, Set, Type
 
-import torch
 from executorch.exir.pass_base import ExportPass, NodeMetadata
 
 
 class ArmPass(ExportPass):
     """Base class for Arm passes"""
 
-    def __init__(self, exported_program: Optional[torch.export.ExportedProgram] = None):
```
digantdesai, line 19: I think this is a step in the right direction.

Just a word of caution for the passes which do use exported_program mainly with an intention to add modes which ends up updating the graph signature i.e. using utils like, `create_constant_placeholder` is that if the graph inside the exported program and the graph the pass got are out of sync then all sorts of weird errors can happen. So it might be wise to assert something like `id(exported_progra.graph_module) == if(pass_arg_graph_module)`

digantdesai (approved):

Sebastian-Larsson (commented):
Unrelated CI failures

PR #14725, opened 2025-10-01
Fix const prop pass when a const prop tensor has zero stride, make it contiguous

abhinaykukkadapu:
Summary:
Found out that xnnpack lowering fails due to rejecting a zero stride tensor by executorch during to_backend phase. The fix is to identify such tensors and make them contiguous.

For context: https://fb.workplace.com/groups/764610762549390/permalink/1868907967337731/

Differential Revision: D83631522


larryliu0820 (approved):

PR #14740, opened 2025-10-02
Use FusedMovingAvgObsFakeQuantize instead of FakeQuantize for faster QAT

navsud:
Summary:
FusedMovingAvgObsFakeQuantize speeds up by fusing FakeQuantize and MovingAverageMinMaxObserver into one CUDA op. Using it should give good speedups. This change updates the QAT qconfigs to accordingly. 

Tested on llama model on HTP and got ~4x QAT speedup.

Differential Revision: D83583655

### DIFF
--- a/backends/qualcomm/quantizer/qconfig.py
+++ b/backends/qualcomm/quantizer/qconfig.py
@@ -200,7 +200,7 @@ def get_16a8w_qnn_qat_config(
     act_observer=MovingAverageMinMaxObserver,
 ) -> QuantizationConfig:
     extra_args: Dict[str, Any] = {"eps": 2**-20}
-    act_fake_quant_ctr = FakeQuantize.with_args(
+    act_fake_quant_ctr = FusedMovingAvgObsFakeQuantize.with_args(
         dtype=torch.int32,
         quant_min=torch.iinfo(torch.uint16).min,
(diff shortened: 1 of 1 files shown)


billmguo (approved):

backends/qualcomm/quantizer/qconfig.py
```diff
@@ -200,7 +200,7 @@ def get_16a8w_qnn_qat_config(
     act_observer=MovingAverageMinMaxObserver,
 ) -> QuantizationConfig:
     extra_args: Dict[str, Any] = {"eps": 2**-20}
-    act_fake_quant_ctr = FakeQuantize.with_args(
+    act_fake_quant_ctr = FusedMovingAvgObsFakeQuantize.with_args(
```
cccclai, line 203: What's the difference between `FakeQuantize` and `FusedMovingAvgObsFakeQuantize`

navsud: `FusedMovingAvgObsFakeQuantize` - as the name suggests, has a combined op for FakeQuantize and MovingAvgObserver which makes it faster than two separate ops: FakeQuantize and MovingAvgObserver.

billmguo (approved):

PR #14744, opened 2025-10-02
Summary: Use javaClassStatic() for class references stored in static …

psiddh:
…variables - creates global references safe for persistence

findClassLocal() returns a local reference. Storing it in static auto exceptionClass = ... could result in potential 'invalid local reference:' as local references become invalid when the JNI frame ends

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/extension/android/jni/jni_helper.cpp
+++ b/extension/android/jni/jni_helper.cpp
@@ -13,10 +13,13 @@ namespace executorch::jni_helper {
 void throwExecutorchException(uint32_t errorCode, const std::string& details) {
   // Get the current JNI environment
   auto env = facebook::jni::Environment::current();
+  if (!env) {
+    return;
+  }
 
-  // Find the Java ExecutorchRuntimeException class
-  static auto exceptionClass = facebook::jni::findClassLocal(
-      "org/pytorch/executorch/ExecutorchRuntimeException");
+  // stable/global class ref — safe to cache
+  static const auto exceptionClass =
+      JExecutorchRuntimeException::javaClassStatic();
 
   // Find the static factory method: makeExecutorchException(int, String)
   static auto makeExceptionMethod =
--- a/extension/android/jni/jni_helper.h
+++ b/extension/android/jni/jni_helper.h
@@ -23,4 +23,11 @@ namespace executorch::jni_helper {
  */
 void throwExecutorchException(uint32_t errorCode, const std::string& details);
 
+// Define the JavaClass wrapper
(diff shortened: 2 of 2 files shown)


extension/android/jni/jni_helper.cpp
```diff
@@ -10,13 +10,17 @@
 
 namespace executorch::jni_helper {
 
+// Define the JavaClass wrapper
+struct JExecutorchRuntimeException : public facebook::jni::JavaClass<JExecutorchRuntimeException> {
+  static constexpr auto kJavaDescriptor = "Lorg/pytorch/executorch/ExecutorchRuntimeException;";
+};
+
```
mergennachin: What are your thoughts on moving this definition to the jni_helper.h file

extension/android/jni/jni_helper.cpp
```diff
@@ -10,13 +10,17 @@
...
 void throwExecutorchException(uint32_t errorCode, const std::string& details) {
   // Get the current JNI environment
   auto env = facebook::jni::Environment::current();
 
   // Find the Java ExecutorchRuntimeException class
-  static auto exceptionClass = facebook::jni::findClassLocal(
-      "org/pytorch/executorch/ExecutorchRuntimeException");
+  static const auto exceptionClass = JExecutorchRuntimeException::javaClassStatic();
```
mergennachin:
```
  auto env = facebook::jni::Environment::current();
  if (!env) {
    // Handle case where JNI environment is not available
    return;
  }
```

mergennachin: Also, any thoughts on catching C++ exceptions? Something like this:


```
  try {
    static const auto exceptionClass = JExecutorchRuntimeException::javaClassStatic();
    // ... rest of the code
  } catch (const std::exception& e) {
    // Log the error or handle gracefully
    // Avoid throwing C++ exceptions across JNI boundary
  }
```

psiddh: I think catching C++ exceptions inside a helper like this can mask real initialization bugs and make debugging harder.

psiddh: Doesn't hurt to add defensive check (although checking for null is not very typical  I think), but handling cases where  it is null can be tricky, We can't talk to java layer as jni env is not valid, and it is not ideal to throw c++ exception from here.

mergennachin (approved):

PR #14745, opened 2025-10-02
Extend FuseViewCopyTransform to fuse more views

AdrianLundell:
Extends the pass to find chains of unary elementwise ops and fuse all views in each chain. This gives the same result since the shape does not matter for elementwise ops.

This change allows to fuse patterns like view -> clone -> view.

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai @psiddh


AdrianLundell:
Fails not related.

AdrianLundell:
@digantdesai Do you think this is OK to add to the general pass or should I move it to the arm backend?

digantdesai:
Let me make sure internal CI is happy since it has a wider surface

digantdesai (approved):

digantdesai:
Internal tests are OK

zingo:
Thansk for testing, mergin!

PR #14746, opened 2025-10-02
Cortex-M backend: Add mul and linear tests

AdrianLundell:
Minor included fixes:
- Make quantized_linear_fusion_pass an XNNPACK pass to initialize it with an exported program
- Add TO_EXECUTORCH as a valid stage after RUN_PASSES
- Add ramp_tensor function to simplify creating dummy data

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai @psiddh


AdrianLundell:
@digantdesai @psiddh What do you think?

psiddh (approved):
lgtm, the PR as it covers all basic scenarios given the current state of Passes
xfails properly document known pass limitations

psiddh:
Looks like "int32 NameError" specifically happens in the dialect transformation stage for these two test cases: linear_rank2_pos and linear_bias. This appears more fundamental issue , isn't ?

AdrianLundell:
I think it has to do with how meta-data is propagated in the pass if I remember correctly, the error is fixed in an upcoming patch at least!

backends/cortex_m/test/tester.py
```diff
@@ -44,7 +47,12 @@ def __init__(self):
 class CortexMRunPasses(RunPasses):
     def __init__(self):
         super().__init__(
-            XNNPACKPassManager, pass_list=[QuantizedOpFusionPass, ReplaceQuantNodesPass]
+            XNNPACKPassManager,
```
digantdesai, line 50: Let's fork it and put it under `backends/cortex_m`

I was thinking `s/XNNPACK/MCU` but we can do the rebranding later and if you want to stick to cortex_m then `s/XNNPACK/CortexM` ?

backends/cortex_m/test/ops/test_linear.py
```diff
@@ -0,0 +1,211 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
```
digantdesai, line 1: ❤️

backends/cortex_m/test/tester.py
digantdesai, line 90: should we merge `test_dialect` and `test_impl`?

AdrianLundell: I kind of like the split since the python implementation is so separate from the c++ implementation and it's nice to get feedback on only the thing you are working on. On the other hand you can simply modify the stages of the tester during development to get this so it should be fine!

backends/cortex_m/test/tester.py
AdrianLundell: I went with CortexM in here https://github.com/pytorch/executorch/commit/57a79037b1973836705527a59c789de7a0348152 so it will have to be rebranded later

backends/cortex_m/test/tester.py
digantdesai: Given, (1) python eager dq->op->q, (2) python impl of the cortex-m op, and (3) C++ impl of the cortex-m op, I guess it makes sense to do A: (1) vs. (2) and B: (1) vs. (3) but when I suggested to merge I assumed we don't really care about (2) besides unit-tests, unless the second comparison (i.e. B) is (2) vs. (3).

PR #14747, opened 2025-10-02
Arm Backend: Add tests for rsub.Scalar, transpose_copy.int and t_copy.default

agrima1304:
These ops are not in the list of core ATen ops and are decomposed automatically when lowering the graph (https://docs.pytorch.org/docs/main/export.html#export-ir-decompositions), so only the tests need to be added.

All three of these ops are in this decomp table: https://github.com/pytorch/pytorch/blob/5d749ceb92c2c28bcfbdf918b4ab99b1a91fcb50/torch/_decomp/__init__.py#L295

Also adds a small comment in the sub op tests.


digantdesai (approved):

Sebastian-Larsson:
Unrelated CI errors.

PR #14748, opened 2025-10-02
Arm backend: Mark test in test_bmm.py as flaky

martinlsm:
Add a new argument to `common.parametrize`, `flakies`, which selects which parametrized test cases to mark as flaky. With this new argument, mark the test test_bmm.py::test_bmm_vgf_FP_single_input[rand_big_1] as flaky.


digantdesai (approved):
Is this the recommended pytest way to selectively mark flaky from "parameterized" inputs? Just curious.

martinlsm:
> Is this the recommended pytest way to selectively mark flaky from "parameterized" inputs? Just curious.

@digantdesai Good question. Since we are wrapping the parametrization decorator anyways, I thought it was easiest to just extend it like this.

PR #14749, opened 2025-10-02
Add null pointer check for evalues

JakeStevens:
Summary: The to* functions could de-reference a NULL pointer. Adding a check.

Differential Revision: D83742246

### DIFF
--- a/runtime/core/evalue.h
+++ b/runtime/core/evalue.h
@@ -63,7 +63,11 @@ class BoxedEvalueList {
    * unwrapped vals.
    */
   BoxedEvalueList(EValue** wrapped_vals, T* unwrapped_vals, int size)
-      : wrapped_vals_(wrapped_vals, size), unwrapped_vals_(unwrapped_vals) {}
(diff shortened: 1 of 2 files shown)


runtime/core/evalue.h
```diff
@@ -335,6 +344,9 @@ struct EValue {
 
   executorch::aten::ArrayRef<double> toDoubleList() const {
     ET_CHECK_MSG(isDoubleList(), "EValue is not a Double List.");
+    ET_CHECK_MSG(
+        payload.copyable_union.as_double_list_ptr != nullptr,
```
JacobSzwejbka, line 357: We should check this in the ctors no?

JacobSzwejbka (changes requested):

runtime/core/evalue.h
JakeStevens: Kept the original checks and added in boxed ctor and each evalue ctor

mergennachin (approved):

JacobSzwejbka (approved):

PR #14750, opened 2025-10-02
Cortex_m backend: Add script for building test runner

AdrianLundell:


AdrianLundell:
Fails unrelated

AdrianLundell:
@digantdesai @psiddh Even though this calls scripts in the arm directory I think it makes sense to add now, partly since it is more user friendly to use scripts in the same folder as you are working in and partly since it gives allows for more separation from the arm backend. For example we can add selected aten ops to the executor runner only when building it for the cortex_m backend easily.

psiddh (approved):

digantdesai:
100%. From user point of view, this is an independent backend and different from `backends/arm`. However, from implementation point of view, we do what makes most sense which is do rely on infrastructure from backends/{arm,xnnpack,transforms} etc. when it makes sense.

PR #14756, opened 2025-10-02
Export lora weights to sep file

lucylq:
* __->__ #14756
* #14767

Differential Revision: [D83777195](https://our.internmc.facebook.com/intern/diff/D83777195/)


examples/models/llama/main.cpp
```diff
@@ -8,6 +8,8 @@
  */
 
 #include <gflags/gflags.h>
+#include <sstream>
+#include <vector>
```
shoumikhin, line 12: nit: maybe some of these are already included by runer.h, etc

shoumikhin (approved):

PR #14779, opened 2025-10-03
Restore backend tests

JakeStevens:
Summary: Now that the SDK update has been ingested, CI tests pass again.

Differential Revision: D83849826


abhinaykukkadapu (approved):

backends/nxp/tests/TARGETS
robert-kalmar, line 47: Irrelevant to this change, but what only `test_quantizer.py` and `test_neutron_backend.py` are defined here? You want to limit the tests only to cover high level tests, what can be considered as integration tests?

robert-kalmar (approved):

PR #14816, opened 2025-10-06
NXP backend: Replace pass to fuse activations functions with joint quantization with activation

roman-janik-nxp:
### Summary
This PR replaces optimizations 'fuse_activation_functions.py' by quantization of Conv 2D and Linear ops together with fusable activations - selected activations supported by Neutron (Relu, Relu6, Sigmoid, Tanh). Logic is determined by target specs, now supporting Neutron-C. Tests updated. Relu has now non-shared, standalone quantization.

### Test plan
Unit tests provided (test_edge_passes.py, test_quantizer.py).


backends/nxp/quantizer/patterns.py
```diff
@@ -385,23 +403,12 @@ class HardTanhPattern(QuantizationPattern):
...
-            biases=[],
-            output=[(node,)],
-        )
 
     def replacement_op(self):
         raise AssertionError()
 
 
-class HardTanhInPlacePattern(QuantizationPattern):
+class HardTanhInPlacePattern(SingleInputBasicPattern):
```
MartinPavella, line 489: Didn't you forget to remove the `get_anchors()` method?

backends/nxp/backend/edge_helper.py
```diff
@@ -4,10 +4,29 @@
...
+    ]
+
+
+def _is_quantize(node_: Node) -> bool:
+    return node_.op == "call_function" and node_.target.__name__ in [
+        "quantize_per_tensor.default",
+        "quantized_decomposed.quantize_per_tensor.default",
+        "quantize_per_channel.default",
+        "quantized_decomposed.quantize_per_channel.default",
+    ]
```
MartinPavella: Why not compare the targets directly? It would be more reliable.

backends/nxp/backend/neutron_target_spec.py
```diff
@@ -7,16 +7,92 @@
...
 
 
 class NeutronHWVersion(Enum):
     N1 = 1
     N3 = 2
 
 
+class NeutronTargetNeutronC:
+    @staticmethod
+    def is_supported_fused_activation(node_: Node) -> bool:
```
MartinPavella: Perhaps the suffix `_aten` should be added to the name. Same for `is_fusable_conv_or_linear()`.
It is not clear that the function is meant for the aten dialect.

backends/nxp/neutron_partitioner.py
```diff
@@ -80,6 +80,10 @@ class QDQCluster:
         operator.getitem,
         exir_ops.edge.aten.view_copy.default,
         exir_ops.edge.aten.permute_copy.default,
+        exir_ops.edge.aten.hardtanh.default,
+        exir_ops.edge.aten.relu.default,
+        exir_ops.edge.aten.sigmoid.default,
+        exir_ops.edge.aten.tanh.default,
```
MartinPavella, line 80: Is this necessary?
I would expect that since the `Move*AuxiliaryOperatorIntoSeparateQDQClusterPass` now supports these operators, they should be the main nodes within their own QDQ clusters.

backends/nxp/quantizer/patterns.py
```diff
@@ -208,11 +214,25 @@ def get_anchors(
...
+        output = [(addmm_node,)]
+        if len(
+            addmm_node.users
+        ) == 1 and self.neutron_target_info.is_supported_fused_activation(
+            activation := next(iter(addmm_node.users))
+        ):
+            activation_quantizer = self.neutron_quantizer.op_to_quantizer[
+                activation.target
+            ]
+            activation_quantizer.annotate(gm)
```
MartinPavella, line 227: Minor:
The code only makes sense if the `activation_quantizer` only quantizes the output, and not the input. Perhaps a comment explaining that would be useful here. Same for the 3 patterns below.

backends/nxp/backend/edge_helper.py
roman-janik-nxp: I created it out of existing code that used name attribute. you are right, I will change it.

backends/nxp/neutron_partitioner.py
roman-janik-nxp: I've looked into into it. The reason is it's in the `QDQClusterRecognizer` `AUXILIARY_OPS` list not only for the use in partitioner, but because the `QDQClusterRecognizer` is also used in the `Move*AuxiliaryOperatorIntoSeparateQDQClusterPass`: 

https://github.com/pytorch/executorch/blob/33f153a0d0ee61c30f6229dee6e190f51c664eff/backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py#L233

backends/nxp/quantizer/patterns.py
```diff
@@ -376,7 +474,7 @@ def partition_types(self):
         return [torch.ops.aten.flatten.using_ints]
 
 
-class HardTanhPattern(QuantizationPattern):
+class HardTanhPattern(SingleInputBasicPattern):
     """
     Quantizer for HardTanh operator. Shared quantization spec is selected, as activation functions usually follows
```
StrycekSimon: There is no longer a shared quantization in use, please update the comments.

backends/nxp/quantizer/patterns.py
```diff
@@ -385,23 +483,11 @@ class HardTanhPattern(QuantizationPattern):
...
-        )
-
     def replacement_op(self):
         raise AssertionError()
 
 
-class HardTanhInPlacePattern(QuantizationPattern):
+class HardTanhInPlacePattern(SingleInputBasicPattern):
     """
     Quantizer for HardTanh operator with param inplace=True. Shared quantization spec is selected, as activation
```
StrycekSimon: As I mentioned before, please, update the comments.

backends/nxp/tests/executorch_pipeline.py
```diff
@@ -85,13 +97,17 @@ def to_quantized_edge_program(
...
     target="imxrt700",
-    neutron_converter_flavor="SDK_25_09",
+    neutron_converter_flavor=default_neutron_converter_flavor,
     remove_quant_io_ops=False,
     custom_delegation_options=CustomDelegationOptions(),  # noqa B008
-    get_quantizer_fn=lambda: NeutronQuantizer(),
+    get_quantizer_fn=None,
 ) -> EdgeProgramManager:
-    calibration_inputs = get_calibration_inputs_fn(to_model_input_spec(input_spec))
+    _neutron_target_spec = NeutronTargetSpec(target, neutron_converter_flavor)
```
StrycekSimon, line 109: It is probably out of scope of this PR but I don't really like that `NeutronTargetSpec` is initialized multiple times during conversion now. (When passin to `NeutronQuantizer` and in `NeutronPartitioner`.) Can we do something about it?

StrycekSimon (commented):
Please remove the comments about shared quantization if it is not relevant.

examples/nxp/aot_neutron_compile.py
```diff
@@ -114,7 +115,10 @@ def post_training_quantize(
     # Based on executorch.examples.arm.aot_amr_compiler.quantize
     logging.info("Quantizing model")
     logging.debug(f"---> Original model: {model}")
-    quantizer = NeutronQuantizer()
+    neutron_target_spec = NeutronTargetSpec(
+        target="imxrt700", neutron_converter_flavor="SDK_25_09"
```
MartinPavella: The `target` and the `neutron_converter_flavor` should really be the function's parameters, and the CLI options from the user should be passed in as arguments.

MartinPavella (approved):

PR #14817, opened 2025-10-06
Arm backend: Backend test serializes and uses EthosUQuant on Ethos-U flows

zingo:
### Summary

Serialize and quantize automaticaly when possible. This make Ethos-U flows work.

### Test plan
This is runned by the backed suit testing for Ethos-U

### DIFF
--- a/backends/test/suite/flows/arm.py
+++ b/backends/test/suite/flows/arm.py
@@ -3,70 +3,66 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# Create flows for Arm Backends used to test operator and model suits
 
-from executorch.backends.arm.quantizer import (
-    get_symmetric_quantization_config,
-    TOSAQuantizer,
-)
+from executorch.backends.arm.common.arm_compile_spec import ArmCompileSpec
+from executorch.backends.arm.quantizer import get_symmetric_quantization_config
 from executorch.backends.arm.test import common
 from executorch.backends.arm.test.tester.arm_tester import ArmTester
+from executorch.backends.arm.tosa.compile_spec import TosaCompileSpec
+from executorch.backends.arm.util._factory import create_quantizer
 from executorch.backends.test.suite.flow import TestFlow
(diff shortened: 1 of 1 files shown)


backends/test/suite/flows/arm.py
```diff
@@ -3,70 +3,66 @@
...
+    compile_spec: ArmCompileSpec,
     symmetric_io_quantization: bool = False,
     per_channel_quantization: bool = True,
 ) -> TestFlow:
 
     def _create_arm_tester(*args, **kwargs) -> ArmTester:
         kwargs["compile_spec"] = compile_spec
+        return ArmTester(*args, **kwargs)
+
+    support_serialize = not isinstance(compile_spec, TosaCompileSpec)
```
digantdesai, line 29: we should be able to run on tosa reference model, no?

backends/test/suite/flows/arm.py
```diff
@@ -3,70 +3,66 @@
...
 )
-ARM_TOSA_INT_FLOW = _create_tosa_flow(
+ARM_TOSA_INT_FLOW = _create_arm_flow(
     "arm_tosa_int",
     common.get_tosa_compile_spec(tosa_spec="TOSA-1.0+INT"),
-    quantize=True,
 )
-
-ARM_ETHOS_U55_FLOW = _create_tosa_flow(
+ARM_ETHOS_U55_FLOW = _create_arm_flow(
```
digantdesai, line 61: When running the model won't this flow run into long runtime issues?

digantdesai (approved):
Stamping to unblock, left some comments. Thanks @zingo !

backends/test/suite/flows/arm.py
zingo: We do not have a "runner" for it. We only have a special testflow.
Enabling serilize right now causes more fails. Need to investigate more so I left it as a future exercise

backends/test/suite/flows/arm.py
zingo: runs kind of OK on my laptop, modeltest took about 30-40min I think. Lets see :)

PR #14818, opened 2025-10-06
Arm backend: Bump tosa version to remove mlplatform dependencies

ArmRyan:
The only change with this version is that submodules will
 no longer point to mlplatform and will instead point to gitlab.

Change-Id: I99e78b9401eaffa2b3e4ae8e840c84bd69ac5ab2


robell (approved):

digantdesai:
> The only change with this version is that submodules will
no longer point to mlplatform and will instead point to gitlab.

Thanks for this.

digantdesai (approved):

PR #14838, opened 2025-10-07
Build pthreadpool with hidden visibility on Apple

GregoryComer:
### Summary
We are seeing pthreadpool-related crashes on Mac when running with pybindings. This appears to be due to XNNPACK using the Google fork of pthreadpool and extension/threadpool using the pthreadpool in libtorch_cpu. See https://github.com/pytorch/executorch/issues/14321 for more details.

Beyond the obvious one definition rule issues, the specific failure happens because the pthreadpool functions in the copy of pthreadpool built with ET are marked as weak on Apple platforms. The functions are not marked as weak in source code or in the build, and the behavior appears to be specific to Apple's toolchain.

Weak symbols are compiled as indirect calls and can be overridden at runtime by strong symbols in another dylib. For reasons that I don't fully understand, the pthreadpool symbols in libtorch_cpu are strong. Also, the calls in XNNPACK prefer the symbols from the local pthreadpool

This PR works around the issue by building pthreadpool with -fvisibility=hidden, which causes the symbols to not be exposed in the final dylib, and thus not end up in the symbol table as an indirect symbol. Instead, the call to pthreadpool_create in extension_threadpool is compiled as a direct call to the pthreadpool_create in the pthreadpool built by executorch.

This isn't a proper fix for the issue, as there are still two pthreadpool implementations in the process whenever we link libtorch_cpu. However, it does appear to mitigate the symptoms and thus prevent crashes. Long-term, we'll need to find a proper solution, such as namespacing the pthreadpool fork.

### Test plan
In addition to validating this change on CI (including trunk CI), I manually verified the fix by testing the repro in https://github.com/pytorch/executorch/issues/14321 before and after the change. I verified that ASan does not trip upon resetting the threadpool. I also verified with `nm` and `otool` that `pthreadpool_create` does not show up in the indirect symbol table, and thus cannot (to my knowledge) be overridden at runtime by the implementation in libtorch_cpu.


GregoryComer:
Samsung model failure is due to running on a fork. Demo app build is broken on trunk, so CI is otherwise green on this PR.

mergennachin (approved):
In 

https://github.com/pytorch/executorch/blob/1b8d380bf1db79ce22fba5096aefb80c2224e5a3/extension/threadpool/threadpool.cpp#L54

What are your thoughts on adding this logging to see if the old and new threadpool address space is drastically different.

```
+  void* old_pool = threadpool_.get();
  threadpool_.reset(pthreadpool_create(new_thread_count));
+  ET_LOG(Info, "Reset threadpool from %p to %p (size %u)", old_pool, threadpool_.get(), new_thread_count);
```

kimishpatel:
> In
> 
> https://github.com/pytorch/executorch/blob/1b8d380bf1db79ce22fba5096aefb80c2224e5a3/extension/threadpool/threadpool.cpp#L54
> 
> What are your thoughts on adding this logging to see if the old and new threadpool address space is drastically different.
> 
> ```
> +  void* old_pool = threadpool_.get();
>   threadpool_.reset(pthreadpool_create(new_thread_count));
> +  ET_LOG(Info, "Reset threadpool from %p to %p (size %u)", old_pool, threadpool_.get(), new_thread_count);
> ```

what does that help with?

GregoryComer:
> In
> 
> https://github.com/pytorch/executorch/blob/1b8d380bf1db79ce22fba5096aefb80c2224e5a3/extension/threadpool/threadpool.cpp#L54
> 
> What are your thoughts on adding this logging to see if the old and new threadpool address space is drastically different.
> 
> ```
> +  void* old_pool = threadpool_.get();
>   threadpool_.reset(pthreadpool_create(new_thread_count));
> +  ET_LOG(Info, "Reset threadpool from %p to %p (size %u)", old_pool, threadpool_.get(), new_thread_count);
> ```

Since the google and upstream pthreadpool use the same allocator for the threadpool state, I wouldn't expect to be able to extract much from the addresses. We should probably have some logging though, so I'll add some. Thanks.

digantdesai:
> by building pthreadpool with -fvisibility=hidden

Clever!

PR #14842, opened 2025-10-07
Reapply "Add EXECUTORCH_THREADPOOL_SIZE options, default to u… (#14307)

GregoryComer:
This reverts commit 750cba7321e6f21bb85819d27241e40aff4bf461.

Re-applying the better threadpool size defaults from https://github.com/pytorch/executorch/pull/14090 with the fix from https://github.com/pytorch/executorch/pull/14838. This gives a 2-4x speedup for many models and platforms (I measured 4x speedup on M1 with MobileNet V3 + XNNPACK). On high core count server platforms (doing evals, for example), this can give a 100x speedup out of box.

### DIFF
--- a/extension/threadpool/CMakeLists.txt
+++ b/extension/threadpool/CMakeLists.txt
@@ -20,6 +20,16 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
+# Threadpool size specifiers. Mutual exclusion is checking in default.cmake.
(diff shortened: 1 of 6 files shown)


GregoryComer:
Voxtral CI failure is due to running on a fork. Samsung failure is also pre-existing.

kimishpatel (approved):
assuming this is reapply. stamping

GregoryComer:
> assuming this is reapply. stamping

Yeah, this is a direct re-apply of the previously landed PR. Thanks.

PR #14843, opened 2025-10-07
Arm backend: Add TOSA (DW) Conv2d dialect operator

oscarandersson8218:
Adds backend dialect operators for:
- Conv2D
- Depthwise Conv2D

Also removes AddBiasPass and moves that logic to new RewriteConv2dPass.


backends/arm/operators/op_tosa_conv2d.py
```diff
@@ -31,13 +31,9 @@
```
digantdesai, line 31: Thanks for the rename :)

digantdesai (approved):
Thanks! I like this tosa op abstraction layer direction :)

Sebastian-Larsson (approved):
Trivial changes since approval that should not have an impact on buck2.

CI failures unrelated.

PR #14848, opened 2025-10-07
Qualcomm AI Engine Direct - Suite Operator Test Support Part 2

winskuo-quic:
### Summary

Support following OPs

- Threshold OP
- negative dims permute
- sqrt unit test modified to use desired input rather than random values
- rsqrt unit test modified to use desired input rather than random values
- per channel conv3d support

For the sqrt/rsqrt, I believe the sample input for each UT is using `rand` instead of `randn` on purpose to prevent negative numbers input, however, if we don't set `generate_random_test_inputs=False`, then later on it will be using random values consisting of negative numbers, causing `nan` showing up on output.

If everything works as expected, we should pass 6 more tests, bringing pass rate from **90.7% -> 91.5%**

### Test plan
UT added

cc @cccclai @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

### DIFF
--- a/backends/test/suite/operators/test_rsqrt.py
+++ b/backends/test/suite/operators/test_rsqrt.py
@@ -37,15 +37,28 @@ class TestRsqrt(OperatorTest):
 
     def test_rsqrt_shapes(self, flow: TestFlow) -> None:
         # Test with different tensor shapes
-
         # 1D tensor
-        self._test_op(RsqrtModel(), (torch.rand(20) + 0.01,), flow)
-
+        self._test_op(
+            RsqrtModel(),
+            (torch.rand(20) + 0.01,),
+            flow,
+            generate_random_test_inputs=False,
+        )
         # 2D tensor
-        self._test_op(RsqrtModel(), (torch.rand(5, 10) + 0.01,), flow)
+        self._test_op(
+            RsqrtModel(),
+            (torch.rand(5, 10) + 0.01,),
+            flow,
+            generate_random_test_inputs=False,
+        )
 
         # 3D tensor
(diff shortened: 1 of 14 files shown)


winskuo-quic:
Hi @cccclai,
This PR should support about 5 to 6 more QNN Suite tests. Please have a look. Thanks,

Hi @GregoryComer,
I am working on supporting more QNN operations to pass the suite operator tests. I have made some changes to the suite operator test files. Please let me know if the changes look fine to you. Thanks!

backends/test/suite/operators/test_rsqrt.py
```diff
@@ -37,15 +37,28 @@ def test_rsqrt_dtype(self, flow: TestFlow, dtype) -> None:
 
     def test_rsqrt_shapes(self, flow: TestFlow) -> None:
         # Test with different tensor shapes
-
         # 1D tensor
-        self._test_op(RsqrtModel(), (torch.rand(20) + 0.01,), flow)
-
+        self._test_op(
```
cccclai, line 41: @GregoryComer does the changes in `backends/test/suite/operators/*` look good to you?

GregoryComer:
> Hi @cccclai, This PR should support about 5 to 6 more QNN Suite tests. Please have a look. Thanks,
> 
> Hi @GregoryComer, I am working on supporting more QNN operations to pass the suite operator tests. I have made some changes to the suite operator test files. Please let me know if the changes look fine to you. Thanks!

Looks good to me. Thanks for catching this and making these updates.

GregoryComer (approved):

cccclai (approved):
The changes under qualcomm/* looks good. @GregoryComer please review the changes in the `test/suite`

PR #14860, opened 2025-10-07
Fix circulate/reflection/replication padding and conv2d with different padding

cccclai:
Differential Revision: D84071939

This PR make changes for the Pad and Conv2d with pad option
1. For Pad op, first of all, QNN can support it so we don't decompose it. `circular` mode is not supported by QNN so it has to be converted anyway. However, for `replicate`, ideally it should work with `OpPad.Scheme.EDGE` but I have been getting wrong result. In this PR, I convert pad with `circular` mode and `replicate` mode for the pass. We can follow up on why replicate didn't work with `OpPad.Scheme.EDGE` and remove the condition

2. Slice copy op needs to be fixed with conv2d node, because we convert pad to slice copy in step 1)

3. For `Conv2d` with pad option, it just work out of box once we fix the pad op

Add unit test for pad and conv2d with pad option, also add test for the new pass

### DIFF
--- a/backends/qualcomm/_passes/convert_pad_to_slice_concat.py
+++ b/backends/qualcomm/_passes/convert_pad_to_slice_concat.py
@@ -1,0 +1,236 @@
+import operator
+import torch
+from torch.fx import GraphModule
+from executorch.exir.pass_base import ExportPass, PassResult
+
+import operator
+import torch
+from torch.fx import GraphModule
+from executorch.exir.pass_base import ExportPass, PassResult
+
+class ConvertPadToSliceConcat(ExportPass):
+    """
+    Replace aten.pad(..., mode in {'circular','replicate'}) with slice+cat (+expand for replicate).
+    Supports 1D/2D (NCL / NCHW-like). SymInt-safe for torch.export graphs.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    # ---------- small helpers ----------
+
+    def _copy_meta(self, src, dst, val_transform=None):
+        dst.meta = dict(getattr(src, "meta", {}))
+        if "val" in getattr(src, "meta", {}) and isinstance(src.meta["val"], torch.Tensor):
+            v = src.meta["val"]
+            if val_transform is not None:
+                try:
+                    v = val_transform(v)
+                except Exception:
+                    pass
+            dst.meta["val"] = v
+
+    def _set_scalar_meta(self, node, dtype=torch.int64):
+        node.meta = getattr(node, "meta", {})
+        node.meta["val"] = torch.tensor(0, dtype=dtype)
+
+    def _sym_size(self, graph, x, dim):
+        if hasattr(torch.ops.aten, "sym_size"):
+            n = graph.create_node("call_function", torch.ops.aten.sym_size.int, (x, dim))
+        else:
+            n = graph.create_node("call_function", torch.ops.aten.size.int, (x, dim))
+        self._set_scalar_meta(n)
+        return n
+
+    def _sym_sub(self, graph, a, b):
+        n = graph.create_node("call_function", operator.sub, (a, b))
+        self._set_scalar_meta(n)
+        return n
+
+    def _rank_from_meta(self, t):
+        r = None
+        if hasattr(t, "meta") and isinstance(t.meta.get("val", None), torch.Tensor):
+            r = t.meta["val"].dim()
+        return r
+
+    def _expand_along_dim(self, graph, t, dim, new_len, before):
+        """
+        Build aten.expand(t, new_sizes) where only 'dim' changes to new_len.
+        Works with SymInt sizes. new_len is a python int.
+        """
+        with graph.inserting_before(before):
+            rank = self._rank_from_meta(t)
+            if rank is None:
+                # Fallback: grab sizes with sym_size one-by-one assuming up to 8 dims
+                # (most models are 4D here; if meta is missing, 4 is reasonable)
+                rank = 4
+            sizes = []
+            # convert negative dim to pos
+            pdim = dim % rank
+            for d in range(rank):
+                if d == pdim:
+                    sizes.append(int(new_len))
+                else:
+                    sizes.append(self._sym_size(graph, t, d))
+            n = graph.create_node("call_function", torch.ops.aten.expand.default, (t, sizes))
+            # meta: broadcast view to the new shape if we have it
+            def _vt(v):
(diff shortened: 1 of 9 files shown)


cccclai:
@winskuo-quic it seems like you're fixing failing. I have this pr fixing some of the conv2d with different mode but still missing circular mode.

backends/qualcomm/tests/test_qnn_delegate.py
```diff
@@ -2011,6 +2083,22 @@ def test_qnn_backend_conv1d(self):
...
                 self.lower_module_and_test_output(module, sample_input)
 
+    def test_qnn_backend_conv1d_mode(self):
+        sample_input = (torch.arange(4 * 3 * 16 * 16).reshape(4, 3, 16, 16).float() / 1000,)
+        for mode in ["zeros", "reflect", "replicate", "circular"]:
+            module = Conv1d(padding=1, padding_mode=mode)  # noqa: F405
+            module = self.get_qdq_module(module, sample_input)
+            self.lower_module_and_test_output(module, sample_input)
+
+    def test_qnn_backend_conv1d_mode(self):
```
winskuo-quic, line 2093: I think the function name is duplicated.

backends/qualcomm/_passes/convert_pad_to_slice_concat.py
```diff
@@ -0,0 +1,236 @@
...
+    # ---------- rewrites ----------
+    def _insert_circular(self, graph, x, pad, ndim, before):
+        with graph.inserting_before(before):
+            if ndim == 1:
+                left, right = pad
+                w = self._sym_size(graph, x, -1)
+                start = self._sym_sub(graph, w, left)
+                left_slice = graph.create_node("call_function", torch.ops.aten.slice.Tensor, (x, -1, start, w))
+                right_slice = graph.create_node("call_function", torch.ops.aten.slice.Tensor, (x, -1, 0, right))
+                self._copy_meta(x, left_slice)
```
chenweng-quic, line 127: Is meta['val'].shape correct for slice and cat?
I feel like their shapes cannot be copied directly and need some transform function.

backends/qualcomm/_passes/convert_pad_to_slice_concat.py
```diff
@@ -0,0 +1,236 @@
...
+    Replace aten.pad(..., mode in {'circular','replicate'}) with slice+cat (+expand for replicate).
+    Supports 1D/2D (NCL / NCHW-like). SymInt-safe for torch.export graphs.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    # ---------- small helpers ----------
+
+    def _copy_meta(self, src, dst, val_transform=None):
```
winskuo-quic, line 22: Could we possibly reuse the `copy_meta` under 
`from executorch.backends.qualcomm._passes.utils import copy_meta`

backends/qualcomm/builders/op_pad.py
```diff
@@ -49,48 +55,72 @@ def define_node(
...
+        if len(node.args) >= 3 and isinstance(node.args[2], str):
+            mode = node.args[2]
+        else:
+            # default to constant
+            mode = "constant"
 
+        scheme_map = {
+            "constant": OpPad.Scheme.CONSTANT,
+            "reflect":  OpPad.Scheme.MIRROR_REFLECT,
+            "replicate": OpPad.Scheme.EDGE, # I think this is supposed to be correct, but the result is wrong
```
winskuo-quic, line 87: I'm actually getting the UT passed even by commenting the `ConvertPadToSliceConcat` pass out when running quantized flow.

winskuo-quic (commented):
Thanks for supporting the conv2d padding mode!
 
I noticed `replicate` mode generates correct output when running quantized flow, while FP has wrong output, which I am guessing is caused by QNN SDK.
 
If this is the case, maybe we don't need to convert `replicate` mode when running quant flow. We can use a flag like following:
https://github.com/pytorch/executorch/blob/8fbc42c0c3a42a40293b4f76fddf9ecd12e712b2/backends/qualcomm/_passes/decompose_linalg_vector_norm.py#L37
 to differentiate whether the pass is called during `transform_for_annotation_pipeline` or `transform_for_export_pipeline`.
 
To reproduce FP error, I have written a simple `pad` test:
```
def test_qnn_backend_pad_mode(self):
        sample_input = (torch.tensor([[[[1, 2, 3],
                               [4, 5, 6],
                               [7, 8, 9]]]], dtype=torch.float),)
        module = PadMode()  # noqa: F405
```
 
```
class PadMode(torch.nn.Module):
    def __init__(self):
        super().__init__()
 
    def forward(self, x):
        return torch.nn.functional.pad(x, pad=(1, 1, 1, 1), mode='replicate')
```

backends/qualcomm/builders/op_pad.py
```diff
@@ -49,48 +55,72 @@ def define_node(
...
 
-        pad_op.AddScalarParam(
-            OpPad.param_pad_constant_value,
-            QNN_TENSOR_TYPE_MAP[type(pad_amount_val)],
-            {QCOM_DATA: pad_amount_val},
-        )
+        # pad_constant_value only for constant mode
+        if mode == "constant":
+            pad_value = None
+            if len(node.args) > 2 and not isinstance(node.args[2], str):
```
winskuo-quic, line 107: Actually, since we are now supporting 2 targets, 
`aten.constant_pad_nd.default` and `aten.pad.default`,
the schema can be different.
I believe there's a change that the constant value is now stored in `node.args[3]`.
Maybe we can add another condition like:
```
if len(node.args) >=3 and not isinstance(node.args[2], str):
                pad_value = node.args[2]
elif len(node.args) >= 4 and not isinstance(node.args[3], str):
                pad_value = node.args[3]
```

backends/qualcomm/builders/op_pad.py
```diff
@@ -18,7 +18,13 @@
 
 @register_node_visitor
 class Pad(NodeVisitor):
-    target = ["aten.constant_pad_nd.default"]
+    target = [
+        "aten.constant_pad_nd.default",
+        "aten.pad.default",
```
winskuo-quic, line 23: Could we also add `aten.pad.default` to `LayoutTransform` pass?
https://github.com/pytorch/executorch/blob/8fbc42c0c3a42a40293b4f76fddf9ecd12e712b2/backends/qualcomm/_passes/layout_transform.py#L76

PR #14861, opened 2025-10-07
Introduce public MergedDataMap

lucylq:
Summary:
Add public merged data map. Module can use this to resolve multiple named data maps.
Creating as a sep dependency rather than inside module/ so it can be used independently of module. (think there may be some other internal usages soon)

Add support for BUCK and CMake. On CMake side, have to add the dep for everyone who uses module ..

Reviewed By: swolchok

Differential Revision: D83527299


swolchok (approved):

GregoryComer (approved):

PR #14862, opened 2025-10-07
Make determinism of channels_last more conservative

kimishpatel:
Summary:
When in doubt because both is_contiguous and is_contiguous(channels_last)
return true, assume it is channels first

Differential Revision: D83998877

### DIFF
--- a/backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
+++ b/backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
@@ -110,7 +110,9 @@ class ChannelsLastTaggedReshapePass(XNNPACKPass):
             if len(quantize_node.all_input_nodes) > 0:
                 actual_node = quantize_node.args[0]
                 if actual_node.op == "placeholder":
-                    return not actual_node.meta["val"][0].is_contiguous()
+                    return ChannelsLastTaggedReshapePass._is_nhwc_tensor(
+                        actual_node.meta["val"][0]
+                    )
                 else:
                     return actual_node.meta.get(
                         ChannelsLastTaggedReshapePass.XNN_NHWC_NODE, False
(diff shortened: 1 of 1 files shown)


GregoryComer (approved):

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
```diff
@@ -125,14 +127,36 @@ def is_nchw_node(node: torch.fx.Node) -> bool:
...
                         ChannelsLastTaggedReshapePass.XNN_NHWC_NODE, False
                     )
 
         return not ChannelsLastTaggedReshapePass.is_nhwc_node(node)
 
+    @staticmethod
+    def _is_nhwc_tensor(tensor: torch.Tensor) -> bool:
+        nhwc = tensor.is_contiguous(memory_format=torch.channels_last)
+        nchw = tensor.is_contiguous()
+        # if both are true false
```
digantdesai, line 144: Should we just use `Tensor.dim_order(ambiguity_check=True)`?

digantdesai (commented):
IIUC this is about ambiguous cases right? If yes, when we have ambiguous shape [1,1,1,1] this may cause issues when the shape changes at runtime and our assumptions about the memory format (one way or the other) breaks down and results in silently calculating incorrect results.

The right solution might be to get rid of `torch.memory_format` from XNNPACK switch to unambiguous (at least with the new API with check) and deal with it properly.

backends/xnnpack/_passes/channels_last_tagged_reshape_pass.py
kimishpatel: it raises runtime error when ambiguity_check=True. Thats why i had to do this

kimishpatel:
> IIUC this is about ambiguous cases right? If yes, when we have ambiguous shape [1,1,1,1] this may cause issues when the shape changes at runtime and our assumptions about the memory format (one way or the other) breaks down and results in silently calculating incorrect results.
> 

How would you determine the unambiguous cases? What I found for one of the models, which was not using channels in the conventional sense, it was just a 4D tensor, it resulted in failures to lower.

> The right solution might be to get rid of `torch.memory_format` from XNNPACK switch to unambiguous (at least with the new API with check) and deal with it properly.

Yes, but that challenge is what is unambiguous. Problem is we dont have a language in pytorch to unambiguously say what is the dim order. I think the right solution would be that. Although when I think about it, it feels fairly non-trivial. User would have to be explicit in their intent within the model to tell you what is the dim order.

I think safe thing to do would be start out wit making everything contiguous/channels first and then introduce any dim order changes explicitly and this would have to happen within XNNPACK backend passes.

GregoryComer:
> IIUC this is about ambiguous cases right? If yes, when we have ambiguous shape [1,1,1,1] this may cause issues when the shape changes at runtime and our assumptions about the memory format (one way or the other) breaks down and results in silently calculating incorrect results.
> 
> The right solution might be to get rid of `torch.memory_format` from XNNPACK switch to unambiguous (at least with the new API with check) and deal with it properly.

Are there any specific examples of this that you're aware of that yield silent correctness issues?

digantdesai:
> How would you determine the unambiguous cases?

if a tensor of a given shape can be mapped to multiple dim_order ==> it is ambiguous. 

https://github.com/pytorch/pytorch/blob/85801126821d4f509f3cf5aafa24dbcd3cd11183/torch/_tensor.py#L1532

PR #14864, opened 2025-10-07
Add xnnpack pass to propagate custom meta field to q/dq nodes

lucylq:
### Summary
Enable quantization with program-data separation.

To select weights for separation, we tag nodes on the eager model. After quantization, qdq nodes are generated. These do not contain the external tags that their inputs have. This PR propagates the tags to the qdq nodes, so that quant weights are moved to external file and can be shared.

### Test plan
```
python -m unittest executorch.backends.xnnpack.test.passes.test_propagate_custom_meta_pass
```

### DIFF
--- a/backends/test/harness/stages/quantize.py
+++ b/backends/test/harness/stages/quantize.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, Sequence, Tuple
+from typing import Any, Callable, Optional, Sequence, Tuple
 
 import torch
 
@@ -15,6 +15,8 @@ from torchao.quantization.pt2e.quantize_pt2e import (
     prepare_qat_pt2e,
 )
 from torchao.quantization.pt2e.quantizer import Quantizer
+from torchao.quantization.quant_api import quantize_
+from torchao.utils import unwrap_tensor_subclass
 
 
 class Quantize(Stage):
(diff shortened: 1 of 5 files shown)


backends/test/harness/stages/quantize.py
```diff
@@ -79,3 +88,47 @@ def graph_module(self) -> str:
 
     def run_artifact(self, inputs):
         return self.converted_graph.forward(*inputs)
+
+class Quantize_(Stage):
+    """
```
lucylq, line 87: @GregoryComer let me know if I should move this into the test class, instead of the test harness, or rename.

GregoryComer (approved):
Thanks for adding the tester stage! I don't really have a better idea on the name, and it's not part of the public API surface, so I'm good with these changes.

backends/xnnpack/test/passes/test_propagate_custom_meta_pass.py
```diff
@@ -0,0 +1,175 @@
...
+        tester.export(self.Export())
+        tester.to_edge_transform_and_lower(
+            ToEdgeTransformAndLower([partitioner])
+        ).to_executorch()
+        tester.run_method_and_compare_outputs()
+
+        exec = tester.get_artifact()
+        program_buffer = exec.buffer
+        self.assertEqual(len(exec._tensor_data), 1)
+        data_buffer = bytes(exec._tensor_data.pop("model"))
```
digantdesai, line 109: Do we want to assert the size of it? Just to make sure it is indeed the quantized weight tensor. Also I would like (somehow) to validate that we also didn't put this in the blob, perhaps by asserting that the blob size is < weight_size (if we have large-ish weights).

backends/test/harness/stages/quantize.py
digantdesai: FWIW, I like it here.

digantdesai (approved):
Thanks @lucylq. I hope we have quantized LoRA working OK with XNNPACK now?

lucylq:
@digantdesai Yes! Thanks a lot for the help. There were a few issues with export_llama and emitter, but I was able to generate quantized lora files. They work well on desktop, need to debug load issues on android.

backends/xnnpack/test/passes/test_propagate_custom_meta_pass.py
lucylq: Thanks - I'll add a check on size.

Re validating that it's not in the blob, we can check that forward fails when we do not pass in the data buffer.

lucylq: @digantdesai added a check on size, and check on accuracy.

Verified locally that we're missing the weight if we do not pass in the data buffer. The test segfaults after ~4 runs though, maybe something isn't cleaned up properly in pybindings. Left that as a todo for now.

PR #14885, opened 2025-10-08
Arm backend: build with NAMED_DATA_MAP=ON for vgf

Erik-Lundell:
PR #14861 introduced the NAMED_DATA_MAP extension which is required to be on when BUILD_EXTENSION_MODULE is on. Therefore, turn it on where needed.


zingo:
Fails unrelated

zingo (approved):

PR #14891, opened 2025-10-08
Arm backend: Make NSS unit tests public

Michiel-Olieslagers:
NSS has now been made public so unit tests
have been moved to public executorch repository.

Change-Id: I604acd50d8461e809cd6e85e2a7498ccdf56e8d7


backends/arm/scripts/install_models_for_test.sh
```diff
@@ -6,3 +6,16 @@
 
 set -e
 pip install -r backends/arm/requirements-arm-models-test.txt
+
+# Install model gym repository
+git clone https://github.com/arm/neural-graphics-model-gym.git
```
digantdesai, line 11: do you want to pin it to a tag or sha?

digantdesai (approved):

backends/arm/scripts/install_models_for_test.sh
Michiel-Olieslagers: Great suggestion, we can probably do this as part of another patch later on!

PR #14893, opened 2025-10-08
Use quantizable LSTM in test when flow has quantize=True

Erik-Lundell:
It makes more sense to use the quantizable version of the LSTM. For example, right now  the xnnpack int8 tests pass, even though all tensors are float, since the quantizer is not triggered.


GregoryComer (approved):
Good catch. Thanks.

PR #14898, opened 2025-10-08
Enable named data map extension in CUDA build

larryliu0820:
Since #14861 we need to specify `EXECUTORCH_BUILD_EXTENSION_NAMED_DATA_MAP=ON` along with `EXECUTORCH_BUILD_EXTENSION_MODULE=ON`. Adding that to fix CUDA CI.


lucylq (approved):

zingo:
#14861 just got put into the 1.0 release "train track" as #14899 you might need/want to do the same with this PR

lucylq:
If the other cuda work isn't in 1.0, might not need to cherrypick this one either?

zingo:
> If the other cuda work isn't in 1.0, might not need to cherrypick this one either?

Ah sorry yes that is correct, I don't know whats on 1.0 or not regarding this code, I just spoted you needed to have the same fix in this PR as our backend and it is on 1.0 also.

PR #14903, opened 2025-10-08
introduce cuda stream into runtime backend

Gasoonjia:
* #14905
* #14904
* __->__ #14903
* #14902

This diff introduces CUDA streams into the Executorch runtime backend. The changes include:

* Adding CUDA stream support to the `cuda_backend.cpp` file
* Including the `cuda_runtime.h` header file in `cuda_backend.cpp`
* Adding a `void* cuda_stream` field to the `AOTInductorModelContainer` struct in `aoti_model_container.h` to store the CUDA stream
* Defining a new macro `ET_CHECK_OR_LOG` in `log.h` to check a condition and log an error message if the condition is false.

Differential Revision: [D84128173](https://our.internmc.facebook.com/intern/diff/D84128173/)


runtime/platform/log.h
```diff
@@ -181,6 +181,20 @@ using ::executorch::runtime::LogLevel;
...
+ * @param[in] _condition The condition to check.
+ * @param[in] _format Log message format string.
+ */
+#define ET_CHECK_OR_LOG(_condition, _format, ...) \
+  do {                                            \
+    if (!(_condition)) {                          \
+      ET_LOG(Error, _format, ##__VA_ARGS__);      \
+    }                                             \
+  } while (0)
+
```
larryliu0820, line 197: Make it explicit that we are logging at `Error` level. Maybe `ET_CHECK_OR_LOG_ERROR`? The other thing you can do is let this macro take a logging level.

larryliu0820 (approved):

PR #14904, opened 2025-10-08
refactor cuda_backend.cpp

Gasoonjia:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #14920
* #14919
* #14905
* __->__ #14904
* #14902

This diff does a comprehensive refactor on cuda_backend.cpp. Two main points:
1.  Reuse ExecuTorch standard macros (ET_CHECK_OR_RETURN_ERROR and others) to replaces exiting if..else + ET_LOG branches
2. Introduced LOAD_SYMBOL macro to concentrate the symbol loading pipeline.

Differential Revision: [D84135844](https://our.internmc.facebook.com/intern/diff/D84135844/)

PR #14905, opened 2025-10-08
update cuda delegate resource free pipeline for safety and segfault-free

Gasoonjia:
* #14920
* #14919
* __->__ #14905
* #14904
* #14902

This diff survives `clear_all_tensors()` function and enable it during backend destroy stage. Furthermore, we defer the container handle deletion to OS to avoid potential segfault if there's more than one .so files.

Differential Revision: [D84135792](https://our.internmc.facebook.com/intern/diff/D84135792/)


backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -318,16 +306,14 @@ class ET_EXPERIMENTAL CudaBackend final
...
-      ET_CHECK_OR_LOG(
-          delete_result == Error::Ok,
-          "Failed to delete AOTInductorModelContainer with error code %d",
-          delete_result);
-      handle->container_handle = nullptr;
-    }
+    // We noticed that AOTInductorModelContainerDelete doesn't work well with
+    // mutitple .so files when we tried to use it to delete container handle,
+    // since freeing one of them will free some sharing resources, leading to
+    // segfault when trying to free the other .so files. Now we do not explicted
```
larryliu0820: You can let AI proofread these.
```suggestion
    // segfault when trying to free the other .so files. Now we do not explicitly
```

larryliu0820 (approved):

PR #14932, opened 2025-10-09
Arm backend: Decompose sub/add with alpha!=1

Erik-Lundell:
This was previously not supported, causing crashes in quantization, and incorrect output in floating point.


zingo:
trunk / test-arm-cortex-m-size-test (zephyr-preset) / linux-job (push) was broken already by 
https://github.com/pytorch/executorch/pull/14570

zingo (approved):

PR #14955, opened 2025-10-09
Tag constants after quantization

lucylq:
### Summary
Tag constants after quantization.

If we tag before quantization, the graph is mutated and tags are lost.

### Test plan
CI (lora test)

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -874,6 +874,7 @@ def _to_edge_and_lower_llama_xnnpack(
     xnnpack_extended_ops: bool = False,
     generate_etrecord: bool = False,
     verbose: bool = False,
+    gen_tag_fn: Optional[Callable[[torch.fx.Node], Optional[str]]] = None,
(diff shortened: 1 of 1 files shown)


examples/models/llama/export_llama_lib.py
```diff
@@ -876,9 +877,27 @@ def _to_edge_and_lower_llama_xnnpack(
     if generate_etrecord:
         builder_exported.generate_etrecord = True
 
-    builder = builder_exported.pt2e_quantize(quantizers).to_edge_transform_and_lower(
-        partitioners
-    )
+    builder = builder_exported.pt2e_quantize(quantizers)
```
digantdesai, line 900: Just moving here after `pt2e_quantize` or something else has also changed?

digantdesai (approved):

examples/models/llama/export_llama_lib.py
lucylq: Yep, splitting it so we can tag after`pt2e_quantize` and before `to_edge_transform_and_lower`

PR #14956, opened 2025-10-09
Support for batched matmul

DrJessop:
Summary: Matmul was relying on linear infra which didn't support batched second argument. This adds support.

Differential Revision: D84279595


hsharma35 (approved):

davorchap:
batched matmul :)

PR #14957, opened 2025-10-09
pin bump with better architecture

Gasoonjia:
Summary:
This diff bump torch pin to 1008.

also expose cuda version into torch_pin.py for better update if needed;

Differential Revision: D84280496


JacobSzwejbka (approved):

JacobSzwejbka (approved):

larryliu0820 (approved):

PR #14962, opened 2025-10-09
Add -fexceptions to quantized kernel generated libs

Daksh-Shami:
Summary:
Added compiling with `-fexceptions` since quantized kernels have exception based code in them. Example error:
```
buck-out/ABC/gen/fbsource/6e53edb0a9a0d828/xplat/executorch/kernels/quantized/__generated_lib_combined__/out/RegisterCodegenUnboxedKernelsEverything.cpp:77:44: error: exception handling disabled, use '-fexceptions' to enable
   77 |             } catch (const std::exception& ex) {
      |           
```
This means the generated code uses exceptions in its code, so we should explicitly flag it to the compiler via the relevant BUCK targets.

Differential Revision: D84284541

### DIFF
--- a/kernels/quantized/targets.bzl
+++ b/kernels/quantized/targets.bzl
@@ -96,41 +96,46 @@ def define_common_targets():
             ],
         )
 
-        executorch_generated_lib(
-            name = "generated_lib" + aten_suffix,
-            deps = [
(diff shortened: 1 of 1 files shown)


swolchok (changes requested):

swolchok (approved):

PR #14976, opened 2025-10-10
backends/cuda: use async malloc/free

swolchok:
Found device synchronize in aoti_torch_delete_tensor_object via Linux perf. This change appears to significantly improve self-reported latency from voxtral_runner as found in https://github.com/pytorch/executorch/blob/main/.github/workflows/cuda.yml#L111-L172:

Baseline:
Run latency (ms):
audio_encoder: 575.797
token_embedding: 14.571
text_decoder: 3095.356

With this PR:
Run latency (ms):
audio_encoder: 175.807
token_embedding: 8.799
text_decoder: 344.367

### DIFF
--- a/backends/cuda/runtime/shims/memory.cpp
+++ b/backends/cuda/runtime/shims/memory.cpp
@@ -225,7 +225,7 @@ AOTITorchError aoti_torch_empty_strided(
 
   if (device_type == static_cast<int32_t>(SupportedDevices::CUDA)) {
     ET_CUDA_CHECK_OR_RETURN_ERROR(
-        cudaMallocManaged(&ptr, static_cast<size_t>(nbytes)));
+        cudaMallocAsync(&ptr, static_cast<size_t>(nbytes), cudaStreamDefault));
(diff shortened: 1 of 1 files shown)


swolchok:
* __->__ #14976

Gasoonjia (approved):

larryliu0820 (approved):

swolchok:
test-qnn-wheel-packages-linux is broken on main. test-multimodal-linux (gemma3-4b) looks like it's a bit flaky on main and segfaulting, so the segfault here is not blocking considering that we have more specific tests for this PR that are passing and this PR should be unrelated. merging.

larryliu0820:
> test-qnn-wheel-packages-linux is broken on main. test-multimodal-linux (gemma3-4b) looks like it's a bit flaky on main and segfaulting, so the segfault here is not blocking considering that we have more specific tests for this PR that are passing and this PR should be unrelated. merging.

Let me see if I can fix the gemma3 issue

PR #14980, opened 2025-10-10
[aoti-et] Enable multimodal runner for Voxtral on CUDA

larryliu0820:
This pull request introduces changes to the CUDA workflow, model artifact handling, and multimodal runner logic. The main changes include restructuring the GitHub Actions workflow to separate model export, benchmarking, and end-to-end testing for the Voxtral CUDA pipeline, improving artifact management and reproducibility. Additionally, the multimodal runner now supports automatic conversion of audio tensors to bfloat16, ensuring compatibility with expected input types. There are also enhancements to caching and symbol registration in the CUDA backend, and build system updates to support linking the CUDA backend.

**Workflow and Artifact Management Improvements:**

* Refactored `.github/workflows/cuda.yml` to split the Voxtral CUDA pipeline into three jobs: `export-voxtral-cuda-artifact` (exports and stores model artifacts), `benchmark-voxtral-cuda` (benchmarks using exported artifacts), and `test-voxtral-cuda-e2e` (runs full end-to-end tests with artifact download and audio input). Improved artifact handling, reproducibility, and added explicit checks for required files. [[1]](diffhunk://#diff-29abea04e0613c2569973e5c8e3c89e04846d408c855eeb1f3efcfae7cfa6f89L90-R91) [[2]](diffhunk://#diff-29abea04e0613c2569973e5c8e3c89e04846d408c855eeb1f3efcfae7cfa6f89R107) [[3]](diffhunk://#diff-29abea04e0613c2569973e5c8e3c89e04846d408c855eeb1f3efcfae7cfa6f89R134-R185) [[4]](diffhunk://#diff-29abea04e0613c2569973e5c8e3c89e04846d408c855eeb1f3efcfae7cfa6f89R196-R267) [[5]](diffhunk://#diff-29abea04e0613c2569973e5c8e3c89e04846d408c855eeb1f3efcfae7cfa6f89R122)

**Multimodal Runner Logic:**

* Added automatic conversion of audio tensors to bfloat16 in `MultimodalPrefiller::prefill` and implemented a helper function `convert_to_bfloat16` in `util.h` to support this. This ensures that audio inputs match the expected dtype for the encoder, improving robustness for multimodal inference. [[1]](diffhunk://#diff-ad4fcb32ffc5f1f7b4f87b5ee58927cb948a8c0976295befd10e3de445913ae4L96-R136) [[2]](diffhunk://#diff-db4801445eaa3bb4f1370fe41d3a00ae2e3ef354a23ad4d5ace141ecc3c6f413R144-R180)

**CUDA Backend and Caching Enhancements:**

* Improved caching logic in `common_shims.cpp` for tensor strides and sizes by validating cached values and updating them when necessary. This prevents stale cache issues and ensures correct tensor metadata. [[1]](diffhunk://#diff-1e7c9d572d434c9a85c9d466e7f406877bc974a373c370fe7ddb3fe32852c1f2R54-R81) [[2]](diffhunk://#diff-1e7c9d572d434c9a85c9d466e7f406877bc974a373c370fe7ddb3fe32852c1f2R104-R130)
* Added dynamic symbol re-registration in `CudaBackend` to handle multiple shared objects in the same process, ensuring correct execution when switching between models.
* Removed redundant logging statements in CUDA backend for cleaner output. [[1]](diffhunk://#diff-a4b17eccf1aa933837671c5184e02bc815d934a362344bb2b17b789cdfaa5375L226) [[2]](diffhunk://#diff-a4b17eccf1aa933837671c5184e02bc815d934a362344bb2b17b789cdfaa5375L256)

**Build System Updates:**

* Updated `CMakeLists.txt` and `executorch-config.cmake` to include and link the CUDA backend (`aoti_cuda`) when building Voxtral and other components, improving build flexibility and CUDA support. [[1]](diffhunk://#diff-606feb24310595f592d98d021a2c90618346977d94decb80b35b7e26ed8ccc1eR89-R95) [[2]](diffhunk://#diff-6a78a155992483ff6f35d595ff6cef63b477d1c853f6482e77acae6ef443f0e4R56)

**Debugging and Tuning Options:**

* Added support for enabling debug compilation in `cuda_backend.py` via the `DEBUG` environment variable, allowing easier troubleshooting and development.

### DIFF
--- a/backends/aoti/common_shims.cpp
+++ b/backends/aoti/common_shims.cpp
@@ -56,18 +56,16 @@ AOTITorchError aoti_torch_get_strides(Tensor* tensor, int64_t** ret_strides) {
   if (it == internal::tensor_to_strides.end()) {
     needs_update = true;
   } else {
-    // Check if cached values are still valid
+    // CRITICAL: Multimodal models reuse tensors with different shapes across
+    // executions (e.g., variable-length audio). We MUST validate cached
+    // metadata matches current tensor state, or CUDA kernels will receive
+    // incorrect shapes leading to memory corruption and segfaults.
     auto tensor_strides = tensor->strides();
-    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
-      needs_update = true;
-    } else {
-      for (int i = 0; i < tensor->dim(); i++) {
-        if (it->second[i] != tensor_strides[i]) {
-          needs_update = true;
-          break;
-        }
-      }
-    }
+    needs_update = !std::equal(
+        it->second.begin(),
+        it->second.end(),
+        tensor_strides.begin(),
+        tensor_strides.end());
   }
 
   if (needs_update) {
@@ -106,18 +104,16 @@ AOTITorchError aoti_torch_get_sizes(Tensor* tensor, int64_t** ret_sizes) {
   if (it == internal::tensor_to_sizes.end()) {
     needs_update = true;
   } else {
-    // Check if cached values are still valid
+    // CRITICAL: Multimodal models reuse tensors with different shapes across
+    // executions (e.g., variable-length audio). We MUST validate cached
+    // metadata matches current tensor state, or CUDA kernels will receive
+    // incorrect shapes leading to memory corruption and segfaults.
     auto tensor_sizes = tensor->sizes();
-    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
-      needs_update = true;
-    } else {
-      for (int i = 0; i < tensor->dim(); i++) {
-        if (it->second[i] != tensor_sizes[i]) {
-          needs_update = true;
-          break;
-        }
-      }
-    }
+    needs_update = !std::equal(
+        it->second.begin(),
+        it->second.end(),
+        tensor_sizes.begin(),
+        tensor_sizes.end());
   }
 
   if (needs_update) {
--- a/backends/aoti/common_shims.cpp
+++ b/backends/aoti/common_shims.cpp
@@ -50,14 +50,36 @@ AOTITorchError aoti_torch_get_storage_offset(
 }
 
 AOTITorchError aoti_torch_get_strides(Tensor* tensor, int64_t** ret_strides) {
-  std::vector<int64_t> strides(tensor->dim());
-  auto tensor_strides = tensor->strides();
-  for (ssize_t i = 0; i < tensor->dim(); i++) {
-    strides[i] = static_cast<int64_t>(tensor_strides[i]);
+  auto it = internal::tensor_to_strides.find(tensor);
+  bool needs_update = false;
+
+  if (it == internal::tensor_to_strides.end()) {
+    needs_update = true;
+  } else {
+    // Check if cached values are still valid
+    auto tensor_strides = tensor->strides();
+    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
+      needs_update = true;
+    } else {
+      for (int i = 0; i < tensor->dim(); i++) {
+        if (it->second[i] != tensor_strides[i]) {
+          needs_update = true;
+          break;
+        }
+      }
+    }
   }
-  auto it =
-      internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides))
-          .first;
+
+  if (needs_update) {
+    std::vector<int64_t> strides(tensor->dim());
+    auto tensor_strides = tensor->strides();
+    for (int i = 0; i < tensor->dim(); i++) {
+      strides[i] = tensor_strides[i];
+    }
+    it =
+        internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides))
+            .first;
+  }
 
   // For 0D tensors, data() returns nullptr on empty vectors, but we need to
   // return a valid pointer
@@ -78,13 +100,35 @@ AOTITorchError aoti_torch_get_dtype(Tensor* tensor, int32_t* ret_dtype) {
 }
 
 AOTITorchError aoti_torch_get_sizes(Tensor* tensor, int64_t** ret_sizes) {
-  std::vector<int64_t> sizes(tensor->dim());
-  auto tensor_sizes = tensor->sizes();
-  for (ssize_t i = 0; i < tensor->dim(); i++) {
-    sizes[i] = static_cast<int64_t>(tensor_sizes[i]);
+  auto it = internal::tensor_to_sizes.find(tensor);
+  bool needs_update = false;
+
+  if (it == internal::tensor_to_sizes.end()) {
+    needs_update = true;
+  } else {
+    // Check if cached values are still valid
+    auto tensor_sizes = tensor->sizes();
+    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
+      needs_update = true;
+    } else {
+      for (int i = 0; i < tensor->dim(); i++) {
+        if (it->second[i] != tensor_sizes[i]) {
+          needs_update = true;
+          break;
+        }
+      }
+    }
   }
-  auto it = internal::tensor_to_sizes.insert_or_assign(tensor, std::move(sizes))
-                .first;
+
+  if (needs_update) {
+    std::vector<int64_t> sizes(tensor->dim());
+    auto tensor_sizes = tensor->sizes();
+    for (int i = 0; i < tensor->dim(); i++) {
+      sizes[i] = tensor_sizes[i];
+    }
+    it = internal::tensor_to_sizes.insert_or_assign(tensor, std::move(sizes))
+             .first;
+  }
 
   // For 0D tensors, data() returns nullptr on empty vectors, but we need to
   // return a valid pointer
--- a/backends/aoti/common_shims.cpp
+++ b/backends/aoti/common_shims.cpp
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
 
 AOTITorchError aoti_torch_get_strides(Tensor* tensor, int64_t** ret_strides) {
   auto it = internal::tensor_to_strides.find(tensor);
+  bool needs_update = false;
+
   if (it == internal::tensor_to_strides.end()) {
+    needs_update = true;
+  } else {
+    // Check if cached values are still valid
+    auto tensor_strides = tensor->strides();
+    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
+      needs_update = true;
+    } else {
+      for (int i = 0; i < tensor->dim(); i++) {
+        if (it->second[i] != tensor_strides[i]) {
+          needs_update = true;
+          break;
+        }
+      }
+    }
+  }
+
+  if (needs_update) {
     std::vector<int64_t> strides(tensor->dim());
     auto tensor_strides = tensor->strides();
     for (int i = 0; i < tensor->dim(); i++) {
       strides[i] = tensor_strides[i];
     }
-    it = internal::tensor_to_strides.emplace(tensor, std::move(strides)).first;
+    it =
+        internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides))
+            .first;
   }
 
   // For 0D tensors, data() returns nullptr on empty vectors, but we need to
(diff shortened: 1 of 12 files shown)


backends/cuda/cuda_backend.py
```diff
@@ -135,6 +135,8 @@ def preprocess(
             "aot_inductor.link_libtorch": False,
             # Package model constants and other generated files directly in the shared object (.so) file
             "aot_inductor.package_constants_in_so": True,
+            # Enable debug mode if the DEBUG environment variable is set
```
Gasoonjia: please get rid of the debug info

extension/llm/runner/multimodal_prefiller.cpp
```diff
@@ -93,14 +93,47 @@ Result<uint64_t> MultimodalPrefiller::prefill(
...
+
+    // Convert to expected dtype if needed
+    if (audio_tensor->scalar_type() != expected_dtype) {
+      if (expected_dtype == ::executorch::aten::ScalarType::BFloat16) {
+        // Convert to bfloat16
+        audio_tensor = ET_UNWRAP(
+            convert_to_bfloat16(audio_tensor),
+            "Failed to convert audio tensor to bfloat16");
+      } else {
+        ET_LOG(
```
Gasoonjia: please use `ET_CHECK_OR_RETURN_ERROR` instead

backends/aoti/common_shims.cpp
```diff
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
...
+
   if (it == internal::tensor_to_strides.end()) {
+    needs_update = true;
+  } else {
+    // Check if cached values are still valid
+    auto tensor_strides = tensor->strides();
+    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
+      needs_update = true;
+    } else {
+      for (int i = 0; i < tensor->dim(); i++) {
```
Gasoonjia: we should use `ssize_t` instead of `int` for i's type.

backends/aoti/common_shims.cpp
```diff
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
...
+  if (needs_update) {
     std::vector<int64_t> strides(tensor->dim());
     auto tensor_strides = tensor->strides();
     for (int i = 0; i < tensor->dim(); i++) {
       strides[i] = tensor_strides[i];
     }
-    it = internal::tensor_to_strides.emplace(tensor, std::move(strides)).first;
+    it =
+        internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides))
+            .first;
```
Gasoonjia, line 79: The current caching logic doesn't provide pretty much pref update due to the O(n_dim) check.

I think we can simplify the logic by always update. Whenever we get a new tensor, we calculate its stride and add it to tensor_to_strides. This way we make the logic simple while maintain same complexity. 
```suggestion
  std::vector<int64_t> strides(tensor->dim());
  auto tensor_strides = tensor->strides();
  for (ssize_t i = 0; i < tensor->dim(); i++) {
    strides[i] = static_cast<int64_t>(tensor_strides[i]);
  }
  it = internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides)).first;
```

Same as aoti_torch_get_sizes function

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -165,6 +165,13 @@ class ET_EXPERIMENTAL CudaBackend final
       Span<EValue*> args) const override {
     AOTIDelegateHandle* handle = (AOTIDelegateHandle*)handle_;
 
+    // Need to re-register all the symbols from the so_handle hosted by this
+    // CudaBackend instance. The reason is that these symbols are
+    // static/singleton across the whole process. When we share multiple methods
+    // (meaning multiple so_handle) in the same process, we need to re-register
+    // the symbols from the so_handle that is being used in this execution.
+    register_shared_library_functions(handle->so_handle);
```
Gasoonjia: please  add ET_CHECK_OK_OR_RETURN_ERROR  for check potential error msg.
Also remove the registration during the init stage.

examples/models/voxtral/CMakeLists.txt
```diff
@@ -86,6 +86,13 @@ list(
   extension_flat_tensor
 )
 
+# Link CUDA backend
+if(EXECUTORCH_BUILD_CUDA)
+  find_package(CUDAToolkit REQUIRED)
+  list(APPEND link_libraries aoti_cuda)
```
Gasoonjia, line 92: is there anyway we can guarantee cudatoolkit is presented when aoti_cuda is available in the cmake system? maybe a preset?

Gasoonjia (commented):
Thansk for your great work!
The size/stride change for me is pretty strange: i con't image a case that the tensor ptr keeps the same while its size/stride got changed

backends/cuda/cuda_backend.py
larryliu0820: I think this is helpful. This introduces a config that will be turned on if you run with `DEBUG=1` environment variable

examples/models/voxtral/CMakeLists.txt
larryliu0820: I think user needs to install by themselves

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: We still need this in init() since we want to call `AOTInductorModelContainerCreateWithDevice`

backends/cuda/cuda_backend.py
Gasoonjia: you can also set `AOT_INDUCTOR_DEBUG_COMPILE` env var to control it. We don't need to have something on our own,

Gasoonjia (approved):

backends/aoti/common_shims.cpp
```diff
@@ -50,15 +50,14 @@ AOTITorchError aoti_torch_get_storage_offset(
...
-    }
-    it = internal::tensor_to_strides.emplace(tensor, std::move(strides)).first;
+  std::vector<int64_t> strides(tensor->dim());
+  auto tensor_strides = tensor->strides();
+  for (ssize_t i = 0; i < tensor->dim(); i++) {
+    strides[i] = static_cast<int64_t>(tensor_strides[i]);
   }
+  auto it =
+      internal::tensor_to_strides.insert_or_assign(tensor, std::move(strides))
+          .first;
```
swolchok: why are we now allocating a vector unconditionally? this seems less efficient than the old code.

backends/aoti/common_shims.cpp
```diff
@@ -79,15 +78,13 @@ AOTITorchError aoti_torch_get_dtype(Tensor* tensor, int32_t* ret_dtype) {
...
 AOTITorchError aoti_torch_get_sizes(Tensor* tensor, int64_t** ret_sizes) {
-  auto it = internal::tensor_to_sizes.find(tensor);
-  if (it == internal::tensor_to_sizes.end()) {
-    std::vector<int64_t> sizes(tensor->dim());
-    auto tensor_sizes = tensor->sizes();
-    for (int i = 0; i < tensor->dim(); i++) {
-      sizes[i] = tensor_sizes[i];
-    }
-    it = internal::tensor_to_sizes.emplace(tensor, std::move(sizes)).first;
+  std::vector<int64_t> sizes(tensor->dim());
```
swolchok: ditto

extension/llm/runner/util.h
```diff
@@ -141,6 +141,43 @@ inline runtime::Result<TensorPtr> populate_start_pos_or_cache_position(
...
+
+  // Convert float to bfloat16
+  for (size_t i = 0; i < num_elements; ++i) {
+    // bfloat16 is the upper 16 bits of float32
+    uint32_t float_bits;
+    std::memcpy(&float_bits, &float_data[i], sizeof(float));
+
+    // Rounding: add 0x7FFF to round to nearest even
+    uint32_t rounding_bias = 0x7FFF + ((float_bits >> 16) & 1);
+    bf16_data[i] = static_cast<uint16_t>((float_bits + rounding_bias) >> 16);
```
swolchok: why can't we use ExecuTorch's BFloat16 class (which is c10::BFloat16 underneath) for this?

swolchok (changes requested):

backends/aoti/common_shims.cpp
Gasoonjia: i just believe the original logic is too complex.
We can also do inplace update:  if there's tensor in the map, we can reuse the vector instead of creating a new one.
this will have same order of memory consumption, while make the code logic cleaner

swolchok: the branch on whether the tensor is already present in the map before creating and filling out a new vector is very important; it's the difference between doing a heap allocation once and doing it every time.

larryliu0820:
@swolchok take another look?

backends/aoti/common_shims.cpp
```diff
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
 
 AOTITorchError aoti_torch_get_strides(Tensor* tensor, int64_t** ret_strides) {
   auto it = internal::tensor_to_strides.find(tensor);
+  bool needs_update = false;
+
   if (it == internal::tensor_to_strides.end()) {
+    needs_update = true;
+  } else {
+    // Check if cached values are still valid
```
swolchok: oh, I didn't see this problem!

backends/aoti/common_shims.cpp
```diff
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
...
+    if (it->second.size() != static_cast<size_t>(tensor->dim())) {
+      needs_update = true;
+    } else {
+      for (int i = 0; i < tensor->dim(); i++) {
+        if (it->second[i] != tensor_strides[i]) {
+          needs_update = true;
+          break;
+        }
+      }
+    }
```
swolchok: `needs_update = !std::equal(it->second.begin(), it->second.end(), tensor_strides.begin(), tensor_strides.end());`?

extension/llm/runner/util.h
```diff
@@ -141,6 +141,30 @@ inline runtime::Result<TensorPtr> populate_start_pos_or_cache_position(
...
+      "BFloat16 conversion only supported from Float source data");
+
+  const auto num_elements = static_cast<size_t>(src_tensor->numel());
+  const float* float_data = src_tensor->const_data_ptr<float>();
+
+  auto bf16_tensor = ::executorch::extension::empty_like(
+      src_tensor, ::executorch::aten::ScalarType::BFloat16);
+  auto* bf16_data = bf16_tensor->mutable_data_ptr<::c10::BFloat16>();
+  for (size_t i = 0; i < num_elements; ++i) {
+    bf16_data[i] = ::c10::BFloat16(float_data[i]);
```
swolchok: we should probably use the executorch name for this type in case we ever need to bifurcate again, but I guess it doesn't matter much

swolchok (approved):
objections withdrawn. I think with some work you can further simplify the sizes()/strides() update stuff, up to you how much of it you want to do right now

backends/aoti/common_shims.cpp
```diff
@@ -51,13 +51,34 @@ AOTITorchError aoti_torch_get_storage_offset(
 
 AOTITorchError aoti_torch_get_strides(Tensor* tensor, int64_t** ret_strides) {
   auto it = internal::tensor_to_strides.find(tensor);
+  bool needs_update = false;
+
```
mergennachin, line 55: Can you make docblock something like this? 

  // CRITICAL: Multimodal models reuse tensors with different shapes across
  // executions (e.g., variable-length audio). We MUST validate cached metadata
  // matches current tensor state, or CUDA kernels will receive incorrect shapes
  // leading to memory corruption and segfaults.

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -165,6 +165,14 @@ class ET_EXPERIMENTAL CudaBackend final
...
     AOTIDelegateHandle* handle = (AOTIDelegateHandle*)handle_;
 
+    // Need to re-register all the symbols from the so_handle hosted by this
+    // CudaBackend instance. The reason is that these symbols are
+    // static/singleton across the whole process. When we share multiple methods
+    // (meaning multiple so_handle) in the same process, we need to re-register
+    // the symbols from the so_handle that is being used in this execution.
+    ET_CHECK_OK_OR_RETURN_ERROR(
+        register_shared_library_functions(handle->so_handle));
+
```
mergennachin, line 175: If we're loading the model once and doing execute/inference multiple times, it will register multiple times, no? 

Can you do something like this? 

```
  void* last_registered_handle = nullptr;

  if (handle->so_handle != last_registered_handle) {
      ET_CHECK_OK_OR_RETURN_ERROR(
          register_shared_library_functions(handle->so_handle));
      last_registered_handle = handle->so_handle;
  }
```

mergennachin (approved):
See inline

mergennachin:
Also can you update the https://github.com/pytorch/executorch/blob/main/examples/models/voxtral/README.md to include additional CUDA instructions too?

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: So the `so_handle` won't change. It's just we are mapping the symbols differently, especially `AOTInductorModelContainerRun`. Let's say we do the following:

1. load(token_embedding)
2. load(audio_encoder)
3. load(text_decoder)
4. run(audio_encoder) <-- here `AOTInductorModelContainerRun` maps to the symbol in text_decoder.so, so we need to remap the symbol to audio_encoder.so

mergennachin: @larryliu0820 

Can you store the AOTInductorModelContainerRunFunc inside AOTIDelegateHandle? 

```
  struct AOTIDelegateHandle {
      void* so_handle;
      std::string so_path;
      AOTInductorModelContainerHandle container_handle;
      void* cuda_stream;

      AOTInductorModelContainerRunFunc run_func;
      // ... etc for all symbols
  };
  ```


```
 Result<DelegateHandle*> init(...) const override {
      AOTIDelegateHandle* handle = new AOTIDelegateHandle();
      handle->so_handle = so_handle;

      // Load symbols into THIS handle's struct (not global)
      handle->run_func = reinterpret_cast<AOTInductorModelContainerRunFunc>(
          dlsym(so_handle, "AOTInductorModelContainerRun"));
      // ... etc

      ET_CHECK_OR_RETURN_ERROR(
          handle->run_func != nullptr,
          AccessFailed,
          "Failed to load AOTInductorModelContainerRun");

      return (DelegateHandle*)handle;
  }
  ```


```
  Error execute(..., DelegateHandle* handle_, ...) const override {
      AOTIDelegateHandle* handle = (AOTIDelegateHandle*)handle_;

      // NO re-registration, use the handle's local symbols
      AOTIRuntimeError error = handle->run_func(
          ...)

      // ... rest of execution ...
  }
  ```

PR #14990, opened 2025-10-10
Add Post to backend stage to recipes

Erik-Lundell:
Add PostToBackend stage to Recipes    
    After a discussion in #14588, it was decided to
    create an additional recipe stage to run passes after partitioning.
    This is meant for backends that convert ops directly instead of
    partitioning.
    

Add RequireExportProgram mixin for passes    
    It is a common pattern that passes require an exported program,
    which means that the pass needs to be constructed right before
    it is ran, with the correct exported program. Today there is no
    unified way this is done, which makes it tricky for general pass handling
    to figure out whether to pass an exported program to the constructor or not.
    This commit solves this by introducing a RequireExportProgram mixin,
    and makes a best effort to migirate passes to this system.
    
    Note that there are passes which require other arguments that
    still will break general pass handling.


zingo:
Hi @abhinaykukkadapu and @digantdesai I think you commented on the original issue leading up to this change, is it OK to merge this?

exir/pass_base.py
```diff
@@ -734,6 +736,24 @@ def migrate_meta_val(
         return res
 
 
+class RequireExportedProgram:
+    """Mixin to require a pass to take an exported program, which is accessed by the exported_program property.
```
abhinaykukkadapu, line 740: @Erik-Lundell @zingo Sorry for the late response and thanks for following up on this change, i've done some _"standardization"_ when passes needs an `ExportedProgram`. QNN backend has exact issue where some passes wants ExportedProgram, i think we can follow similar pattern here too, here is the code ptr: https://github.com/pytorch/executorch/blob/main/backends/qualcomm/recipes/qnn_recipe_provider.py#L142

Basic idea is that the recipe will get a callable similar to `aten_transform_passes` and the callable can have a sequence of passes instantiated with or without EP. Let me know if you have any questions. 

After you are done, i will probably take a stab at creating a single stage for all the pass related steps and then that stage shall be plugged in at various stages pre/post et stages.

PR #15018, opened 2025-10-10
Use TensorPtr in Image and Audio structs to avoid data duplication.

shoumikhin:
Summary: .

Differential Revision: D84381425

### DIFF
--- a/extension/llm/runner/pybindings.cpp
+++ b/extension/llm/runner/pybindings.cpp
@@ -277,11 +277,11 @@ PYBIND11_MODULE(_llm_runner, m) {
       .def_property_readonly(
           "uint8_data",
           static_cast<const std::vector<uint8_t>& (Image::*)() const&>(
-              &Image::get_uint8_data))
+              &Image::copy_uint8_data))
       .def_property_readonly(
           "float_data",
           static_cast<const std::vector<float>& (Image::*)() const&>(
-              &Image::get_float_data))
+              &Image::copy_float_data))
       .def("__repr__", [](const Image& img) {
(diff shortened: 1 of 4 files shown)


extension/llm/runner/pybindings.cpp
```diff
@@ -317,15 +317,14 @@ PYBIND11_MODULE(_llm_runner, m) {
...
+              &Audio::copy_uint8_data))
       .def_property_readonly(
           "float_data",
           static_cast<const std::vector<float>& (Audio::*)() const&>(
-              &Audio::get_float_data))
+              &Audio::copy_float_data))
       .def_property_readonly("batch_size", &Audio::get_batch_size)
       .def_property_readonly("n_bins", &Audio::get_n_bins)
       .def_property_readonly("n_frames", &Audio::get_n_frames)
-      .def("toTensor", &Audio::toTensor)
```
larryliu0820, line 328: Instead of removing it, I actually think this is very useful. Can you convert ETensor to torch.tensor here?

extension/llm/runner/pybindings.cpp
```diff
@@ -317,15 +317,14 @@ PYBIND11_MODULE(_llm_runner, m) {
       .def_property_readonly(
           "uint8_data",
           static_cast<const std::vector<uint8_t>& (Audio::*)() const&>(
-              &Audio::get_uint8_data))
+              &Audio::copy_uint8_data))
```
larryliu0820, line 320: Since it's a readonly property, I'm curious if we can get away of copying, by providing a view of the underlying data.

PR #15020, opened 2025-10-10
Add Metal backend core ETMetal runtime.

manuelcandales:
This commit introduces the foundational Metal backend runtime.

Key features:
 - ETMetalStream for managing Metal devices, command queues, buffers, and synchronization.
 - ETMetalShaderLibrary for compiling Metal shader source and caching pipeline states.
 - ETMetalKernelFunction for kernel argument binding, dispatching, and synchronization with stream-managed encoders.
 - Added global buffer management and pointer tracking between host and Metal buffers.
 - Added global stream management utilities and synchronization helpers

This provides the necessary runtime primitives for executing compute shaders and MPSGraph workloads.


manuelcandales:
* #15024
* #15023
* #15022
* #15021
* __->__ #15020

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,872 @@
...
+void ETMetalStream::commitAndContinue() {
+    if (!commandBuffer_) {
+        ET_LOG(Error, "ETMetalStream::commitAndContinue: No command buffer to commit");
+        return;
+    }
+
+    // Commit buffer and allow immediate reuse for better performance
+    [commandBuffer_ commit];
+    ET_LOG(Debug, "ETMetalStream::commitAndContinue: Committed buffer %p with continue", commandBuffer_);
+
```
mergennachin, line 704: Don't you need to release the buffer after commit?

```
      [commandBuffer_ release];
      commandBuffer_ = nil; 
```

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,872 @@
...
+}
+
+// =======================
+// ETMetalKernelFunction Implementation
+// =======================
+
+ETMetalKernelFunction::ETMetalKernelFunction(id<MTLComputePipelineState> cps, id<MTLFunction> func)
+    : cps_(cps), func_(func), encoder_(nil) {
+    if (cps_) [cps_ retain];
+    if (func_) [func_ retain];
```
mergennachin, line 275: Why do you need these `retain` lines? Aren't these already owned by the class?

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,872 @@
...
+        if (cps_) {
+            [cps_ release];
+            cps_ = nil;
+        }
+        if (func_) {
+            [func_ release];
+            func_ = nil;
+        }
+
+        encoder_ = nil; // Clear reference without releasing
```
mergennachin, line 291: Just this?

```
cps_ = nil;
func_ = nil;
encoder_ =nil;
```

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,872 @@
...
+        @autoreleasepool {
+            endKernelCoalescing();
+
+            [mpsGraph encodeToCommandBuffer:commandBuffer()
+                                      feeds:feeds
+                           targetOperations:nil
+                          resultsDictionary:results
+                        executionDescriptor:nil];
+
+            //synchronize(syncType);
```
mergennachin: Why commented out?

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,271 @@
...
+  NONE, // no commit to command buffer
+  COMMIT, // commit and flush the command buffer
+  COMMIT_AND_WAIT, // flush and wait for command buffer execution to finish
+  COMMIT_AND_CONTINUE, // commit and continue with a new underlying command
+                       // buffer
+  COMMIT_ADAPTIVE, // commit adaptively based on available memory
+};
+
+// =======================
+// ETMetalShaderLibrary - ExecuTorch Metal shader library management
```
mergennachin, line 78: Add a bit more comments. ETMetalShaderLibrary

- Compiles Metal Shading Language (MSL) source to MTLLibrary
- Caches compiled pipeline states for reuse
- Creates ETMetalKernelFunction lazily
- Lifetime: Persistent (application lifetime)

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,271 @@
...
+  std::string shaderSource_;
+  MTLLibrary_t library_;
+  std::unordered_map<
+      std::string,
+      std::pair<MTLComputePipelineState_t, MTLFunction_t>>
+      pipelineStates_;
+};
+
+// =======================
+// ETMetalKernelFunction - ExecuTorch Metal kernel function execution
```
mergennachin, line 135:
```
- Encodes kernel dispatches into command buffers
- Lifetime: Per-kernel-invocation
```

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,271 @@
...
+  void runCommandBlock(std::function<void(void)> f);
+
+ private:
+  MTLComputePipelineState_t cps_;
+  MTLFunction_t func_;
+  MTLComputeCommandEncoder_t encoder_;
+};
+
+// =======================
+// ETMetalStream - Metal command buffer and synchronization management
```
mergennachin, line 203:
```
 ETMetalStream                                                  
   - Manages Metal device, command queue, and command buffers
   - Provides synchronization primitives (5 modes)               
   - Implements kernel coalescing for batch optimization      
   - Executes MPSGraph operations for AOTI-compiled models  
```

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,271 @@
...
+void synchronize_metal_stream_with_type(int sync_type);
+
+// =======================
+// Metal helper functions (C interface)
+// =======================
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Memory management functions for Metal
```
mergennachin, line 355: Docblock about the memory management aspect of this design, such as buffer lifecycle, thread safety etc.

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,271 @@
...
+// Helper functions to access Metal objects
+MTLDevice_t get_metal_device();
+MTLCommandQueue_t get_metal_command_queue();
+
+#ifdef __cplusplus
+}
+
+// C++ only - expose the Metal buffer mapping
+#ifdef __OBJC__
+extern std::unordered_map<void*, MTLBuffer_t> ptr_to_mtl_buffer;
```
mergennachin, line 375: do you need a lock and thread safety to access ptr_to_mtl_buffer?

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,872 @@
...
+            [blitEncoder endEncoding];
+            synchronize(syncType);
+
+            ET_LOG(Debug, "ETMetalStream::fill: Filled buffer with value %u, length %zu, offset %zu", value, length, offset);
+        }
+    });
+}
+
+void ETMetalStream::copy(id<MTLBuffer> srcBuffer, id<MTLBuffer> dstBuffer, size_t length,
+                        size_t srcOffset, size_t dstOffset, SyncType syncType) {
```
mergennachin, line 745: before copying check that offsets are within the buffer bounds

mergennachin (commented):
See inline

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: I think retaining ensures that the objects remain valid for the lifetime of the ETMetalKernelFunction instance, preventing them from being deallocated elsewhere. Without the retains cps_ and func_ could become dangling pointers.
Similar patter followed in PyTorch [here](https://github.com/pytorch/pytorch/blob/2395d7d7dad80bb887872e85b3ae8cbd38c70f1c/aten/src/ATen/native/mps/OperationUtils.mm#L1158-L1159)

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: No, we need to release them, because we own them now.
Same pattern followed in PyTorch [here](https://github.com/pytorch/pytorch/blob/2395d7d7dad80bb887872e85b3ae8cbd38c70f1c/aten/src/ATen/native/mps/OperationUtils.mm#L1161-L1164)

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: With commitAndContinue we commit and continue reusing the buffer, until we flush. So, we don't release it after commit.

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: left over of debugging, deleted.

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: done!

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: done!

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: done!

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: It is currently working without one, because the operations are being executed in sequence. I don't see any CPU threading in the AOTI MPS side. But again, I want to replace this design with something more robust.

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: Buffer management is something I want to change. I think I want to have some kind of RAII instead of a global map. But I was hoping to do that after this first landing. Is it ok if I add documentation here later?

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,891 @@
...
+
+int metal_copy_memory(void* dst, const void* src, size_t nbytes, bool src_is_device, bool dst_is_device) {
+    if (!src || !dst || nbytes == 0) {
+        ET_LOG(Error, "Metal copy: Invalid parameters");
+        return -1;
+    }
+
+    @autoreleasepool {
+        // Case 1: Device-to-device copy - use GPU blit encoder (most efficient)
+        if (src_is_device && dst_is_device) {
```
larryliu0820, line 112: Do we need this? Shouldn't we already have unified memory for all the M series?

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -0,0 +1,891 @@
...
+// =======================
+// Global Variables and Storage
+// ================
+
+
+// Global Metal buffer mapping - accessible for MPS shim
+std::unordered_map<void*, id<MTLBuffer>> ptr_to_mtl_buffer;
+
+// Global storage to keep shared_ptr alive while raw pointers are used
+static std::unordered_map<ETMetalKernelFunction*, std::shared_ptr<ETMetalKernelFunction>> function_storage;
```
larryliu0820, line 51: Huh. Can you explain why we need this? Is the raw pointer mapping to its own shared_ptr version?

backends/apple/metal/runtime/shims/et_metal.h
```diff
@@ -0,0 +1,382 @@
...
+  void runCommandBlock(std::function<void(void)> f);
+
+ private:
+  MTLComputePipelineState_t cps_;
+  MTLFunction_t func_;
+  MTLComputeCommandEncoder_t encoder_;
+};
+
+// =======================
+// ETMetalStream - Metal command buffer and synchronization management
```
larryliu0820, line 203: noob question, how much of https://github.com/pytorch/executorch/blob/main/backends/apple/mps/runtime/MPSStream.h can be reused here?

larryliu0820 (approved):

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: We need this to keep the shared_ptr alive, while the raw pointer is still in use

backends/apple/metal/runtime/shims/et_metal.h
manuelcandales: They are all adaptations of the PyTorch's MPSStream.h.
I think with enough time (not right now) we should refactor PyTorch's MPS classes in order to be ATen agnostic, and  be able to reuse them.

manuelcandales: So, basically what I am saying is, I don't think we should reuse code from the ET MPS backend, instead, we should refactor PyTorch's code and reuse that.

PR #15021, opened 2025-10-10
Metal backend: Add AOTI shims for memory management

manuelcandales:
Implement the following shims:
- aoti_torch_create_tensor_from_blob_v2
- aoti_torch_empty_strided
- aoti_torch_delete_tensor_object
- aoti_torch_copy_
- aoti_torch__reinterpret_tensor

PR #15022, opened 2025-10-10
Metal backend: Implement the AOTI MPS shim

manuelcandales:
Includes:
- Shader library management
- Kernel function handling
- Command buffer execution
- Metal memory operations


manuelcandales:
* #15024
* #15023
* __->__ #15022
* #15021
* #15020

backends/apple/metal/runtime/shims/shim_mps.mm
```diff
@@ -0,0 +1,540 @@
...
+
+            uint8_t* src_contents = static_cast<uint8_t*>([src_mtl_buffer contents]);
+            uint8_t* dst_contents = static_cast<uint8_t*>([dst_mtl_buffer contents]);
+
+            if (!src_contents || !dst_contents) {
+                ET_LOG(Error, "aoti_torch_mps_copy_buffer: Failed to get buffer contents");
+                return Error::Internal;
+            }
+
+            memcpy(dst_contents + dst_offset, src_contents + src_offset, data_size);
```
mergennachin, line 478: aoti_torch_mps_free and aoti_torch_mps_memcpy expect contents ptr, but aoti_torch_mps_copy_buffer expects MTLBuffer objects. Is this intentional?  

Isn't it something like this? 

```
              auto src_it = ptr_to_mtl_buffer.find(src_buffer);
              auto dst_it = ptr_to_mtl_buffer.find(dst_buffer);

              if (src_it == ptr_to_mtl_buffer.end()) {
                  ET_LOG(Error, "aoti_torch_mps_copy_buffer: src_buffer %p not found", src_buffer);
                  return Error::InvalidArgument;
              }

              if (dst_it == ptr_to_mtl_buffer.end()) {
                  ET_LOG(Error, "aoti_torch_mps_copy_buffer: dst_buffer %p not found", dst_buffer);
                  return Error::InvalidArgument;
              }

              id<MTLBuffer> src_mtl_buffer = src_it->second;
              id<MTLBuffer> dst_mtl_buffer = dst_it->second;

              ETMetalStream* stream = getCurrentMetalStream();
              stream->copy(src_mtl_buffer, dst_mtl_buffer, data_size, src_offset, dst_offset, SyncType::NONE);
```

backends/apple/metal/runtime/shims/shim_mps.mm
```diff
@@ -0,0 +1,540 @@
...
+                return Error::Internal;
+            }
+            id<MTLBuffer> subBuffer = [device newBufferWithBytesNoCopy:buffer_pointer + constant_offset
+                                                                length:data_size
+                                                               options:MTLResourceCPUCacheModeWriteCombined | MTLResourceStorageModeShared
+                                                           deallocator:nil];
+
+            if (constant_offset != 0) {
+                ptr_to_mtl_buffer[buffer_pointer + constant_offset] = subBuffer;  // Map contents to buffer
+            }
```
mergennachin, line 437: Why do you need this at all?  subBuffer doesn't seem to be used anywhere

backends/apple/metal/runtime/shims/shim_mps.mm
```diff
@@ -0,0 +1,540 @@
...
+    } catch (const std::exception& e) {
+        ET_LOG(Error, "aoti_torch_mps_delete_shader_library exception: %s", e.what());
+        return Error::Internal;
+    } catch (...) {
+        ET_LOG(Error, "aoti_torch_mps_delete_shader_library: unknown exception");
+        return Error::Internal;
+    }
+}
+
+AOTITorchError aoti_torch_mps_get_kernel_function(
```
mergennachin, line 90: Call @autoreleasepool?

backends/apple/metal/runtime/shims/shim_mps.mm
```diff
@@ -0,0 +1,540 @@
...
+    } catch (const std::exception& e) {
+        ET_LOG(Error, "aoti_torch_mps_get_kernel_function exception: %s", e.what());
+        return Error::Internal;
+    } catch (...) {
+        ET_LOG(Error, "aoti_torch_mps_get_kernel_function: unknown exception");
+        return Error::Internal;
+    }
+}
+
+AOTITorchError aoti_torch_mps_start_encoding(
```
mergennachin, line 130: call @autoreleasepool

backends/apple/metal/runtime/shims/shim_mps.mm
```diff
@@ -0,0 +1,540 @@
...
+    } catch (...) {
+        ET_LOG(Error, "aoti_torch_mps_dispatch_single_with_group_size: unknown exception");
+        return Error::Internal;
+    }
+}
+
+// Pure C dispatch functions - array versions
+AOTITorchError aoti_torch_mps_dispatch_array(
+    AOTIMetalKernelFunctionHandle func,
+    const uint64_t* length,
```
mergennachin, line 267: Validate length != nullptr

backends/apple/metal/runtime/shims/shim_mps.mm
manuelcandales: It is intentional to make things work in this first landing. It is a workaround, because right now I can't create an ET tensor with a Metal buffer, I need the contents pointer. I think to avoid this workaround I need to make changes in executorch::extension::from_blob. I want to look into this later.

backends/apple/metal/runtime/shims/shim_mps.mm
manuelcandales: This is another workaround, because right now AOTI MPS creates a single buffer with all the constants (data tensors). This is one key difference between AOTI MPS vs AOTI CUDA.
However, when I am calling ops implemented with MPSGraph, which take in data tensors (right now convolution and sdpa), I need to call initWithMTLBuffer to create an MPSGraphTensor from the buffer. But initWithMTLBuffer doesn't let me pass an offset. So, I need to have the data tensor in its own buffer.

larryliu0820 (approved):

PR #15023, opened 2025-10-10
Metal backend: Add operator implementations

manuelcandales:
Adds bfloat16/float32 working implementations of the following AOTI shim ops:
 - aoti_torch_mps_mm_out
 - aoti_torch_mps_convolution
 - aoti_torch_mps__scaled_dot_product_attention_math_for_mps

 Adds a stub implementation of aoti_torch_mps_addmm_out


manuelcandales:
* #15024
* __->__ #15023

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+      AOTITensorHandle output_tensor_handle = nullptr;
+
+      AOTITorchError create_result = aoti_torch_create_tensor_from_blob_v2(
+          tensor_data,
+          static_cast<int64_t>(output_sizes_int64.size()),  // ndim
+          output_sizes_int64.data(),
+          output_strides.data(),
+          0,  // storage_offset
+          dtype,  // dtype
+          2,  // device_type (MPS)
```
mergennachin: If we change to 11 or 13, make sure to change this consistently

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+        AOTITensorHandle attn_tensor_handle = nullptr;
+
+        AOTITorchError create_out_result = aoti_torch_create_tensor_from_blob_v2(
+            out_contents_ptr,
+            4,  // ndim
+            output_sizes.data(),
+            out_strides.data(),
+            0,  // storage_offset
+            dtype,
+            2,  // device_type (MPS)
```
mergennachin: .

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+        );
+
+        AOTITorchError create_attn_result = aoti_torch_create_tensor_from_blob_v2(
+            attn_contents_ptr,
+            4,  // ndim
+            attn_sizes.data(),
+            attn_strides.data(),
+            0,  // storage_offset
+            dtype,
+            2,  // device_type (MPS)
```
mergennachin: .

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+      // Convert AOTITensorHandle to ExecutorTorch tensors
+      auto out_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(out);
+      auto self_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(self);
+      auto mat1_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(mat1);
+      auto mat2_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(mat2);
+
+      ET_LOG(Debug, "aoti_torch_mps_addmm_out: Converted tensor handles to ET tensors");
+
+      // For now, just zero out the output tensor to get the right shape
+      // TODO: Implement actual matrix multiplication: out = beta * self + alpha * (mat1 @ mat2)
```
mergennachin: Probably need addmm for functionality for most models

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+        if (value_it != ptr_to_mtl_buffer.end()) {
+          value_buffer = value_it->second;
+        }
+
+        // Create temporary Metal buffers if not found in mapping
+        if (!query_buffer) {
+          size_t query_size = query_tensor->numel() * element_size;
+          query_buffer = [device newBufferWithBytes:query_data_ptr
+                                             length:query_size
+                                            options:MTLResourceStorageModeShared];
```
mergennachin: how and where do you clean this up?

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+
+      ET_LOG(Debug, "aoti_torch_mps_addmm_out: Converted tensor handles to ET tensors");
+
+      // For now, just zero out the output tensor to get the right shape
+      // TODO: Implement actual matrix multiplication: out = beta * self + alpha * (mat1 @ mat2)
+
+      // Get output data pointer and size
+      float* out_data = static_cast<float*>(out_tensor->mutable_data_ptr());
+      size_t out_numel = out_tensor->numel();
+
```
mergennachin: Why hardcode to float? What if we want to support bfloat for example

```
  void* out_data = out_tensor->mutable_data_ptr();
  size_t out_numel = out_tensor->numel();
  size_t element_size = out_tensor->element_size();
  std::memset(out_data, 0, out_numel * element_size);
```

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+          element_size = sizeof(uint16_t);  // bfloat16 is 16 bits
+        } else {
+          ET_LOG(Error, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: Unsupported data type: %d", dtype);
+          throw std::runtime_error("Unsupported data type for scaled dot product attention");
+        }
+
+        ET_LOG(Debug, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: mps_dtype=%d, element_size=%zu", mps_dtype, element_size);
+
+        // Calculate scale factor
+        double scale_factor = scale ? *scale : (1.0 / sqrt(static_cast<double>(headSize)));
```
mergennachin, line 941: check headsize is not equal to 0

backends/apple/metal/runtime/shims/et_metal_ops.h
```diff
@@ -0,0 +1,86 @@
...
+    AOTITensorHandle mat1,
+    AOTITensorHandle mat2,
+    double beta,
+    double alpha);
+
+/**
+ * ExecutorTorch implementation of aoti_torch_mps_mm_out.
+ * Performs simple matrix multiplication: out = self @ mat2
+ */
+AOTITorchError aoti_torch_mps_mm_out(
```
mergennachin, line 25: Does custom ops use caching mechanism like the ETMetalShaderLibrary?

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+          ET_LOG(Debug, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: MPSGraph execution completed successfully");
+
+        } @catch (NSException *exception) {
+          ET_LOG(Error, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: NSException caught: %s - %s",
+                 [[exception name] UTF8String], [[exception reason] UTF8String]);
+          throw std::runtime_error("MPSGraph operation failed with NSException");
+        }
+
+        // For attention weights, zero-fill the GPU buffer (shared memory allows CPU memset)
+        std::memset(attn_contents_ptr, 0, attn_size_bytes);
```
mergennachin, line 1285: do you need zero filling here

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -0,0 +1,1255 @@
...
+
+        // Mark that we own the memory for these tensors
+        auto* out_et_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(out_tensor_handle);
+        auto* attn_et_tensor = reinterpret_cast<executorch::runtime::etensor::Tensor*>(attn_tensor_handle);
+        is_tensor_own_memory[out_et_tensor] = true;
+        is_tensor_own_memory[attn_et_tensor] = true;
+
+        // Set output tensor handles
+        *ret0 = out_tensor_handle;
+        *ret1 = attn_tensor_handle;
```
mergennachin, line 1337: Is ret1 actually populated or just zerod

backends/apple/metal/runtime/shims/et_metal_ops.h
manuelcandales: No, not yet. These fallback ops are implemented using MPSGraph, so, here we would be caching the graph. This is something I want to look into later when optimizing performance. But this deserves time. In particular, since I never understood why MPSGraph operations have a non-trivial CPU overhead in PyTorch, in spite of PyTorch having a caching mechanism for MPSGraphs.

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: Yeah, but I will delete this for this initial landing.

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: done!

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: deleted addmm

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: Well, I though it was nicer to return 0, rather than some random stuff.

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: This is just zeroed. 
We are using MPSGraph's [scaledDotProductAttention](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/scaleddotproductattention(query:key:value:mask:scale:name:)) which only returns the output tensor.
We need to return an attention tensor because we need to match _scaled_dot_product_attention_math_for_mps signature. But we don't really need it, it gets thrown away [here](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp#L806)

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: I removed this. The op should never directly allocate buffers. It is all done through the malloc shim, which handles tracking.

larryliu0820 (approved):

mergennachin (approved):

PR #15026, opened 2025-10-10
Bump torch nightly pin to 20251012

manuelcandales:


torch_pin.py
```diff
@@ -1,2 +1,2 @@
 TORCH_VERSION = "2.10.0"
-NIGHTLY_VERSION = "dev20251003"
+NIGHTLY_VERSION = "dev20251010"
```
mergennachin: You also need to update this file 

https://github.com/pytorch/executorch/blob/d00279d4ab19bf38f7b130f66be11352802ab7cc/.ci/docker/ci_commit_pins/pytorch.txt

You can find the hash from https://hud.pytorch.org/hud/pytorch/pytorch/nightly/1?per_page=50&mergeEphemeralLF=true

Looks like a57a14868dcfd9dabf9bd19b6b11f31967c80c87

mergennachin (approved):

PR #15039, opened 2025-10-11
Fix alignment calculation in XNNWeightsCache

GregoryComer:
### Summary
We're seeing crashes on Android when running XNNPACK-delegated models. I tracked it down to a bug in the alignment calculation for weight cache memory. To make the calculation, it casts the void* to a (signed) intptr_t. When the address is in the upper half of the address space, it becomes negative. This causes the modulo to return a negative value and increment the address too much - leading to out of bounds access.

https://github.com/pytorch/executorch/blob/cc6cb837d6ac92f52a2d30a405900caf115f0556/backends/xnnpack/runtime/XNNWeightsCache.cpp#L166-L168

Walking through the numbers I captured in https://github.com/pytorch/executorch/issues/14831:
* The raw (unaligned) address of the data buffer is 0xb40000763d4bfa90.
* The target alignment is 64 bytes.
* Casting the address to intptr_t gives -5476376639047992688.
  * Mod 64 is -48.
  * The total offset applied  is 64 - (-48) = 112.
* Since the allocation size is N + 64, increasing the start by 112 means the new region extends 48 bytes past the end of the allocation.

To resolve this, I replaced the alignment code with a call to std::align. Casing to uintptr_t also resolves it, but using the standard implementation seems less error prone.

### Test plan
I've validated that the repro in https://github.com/pytorch/executorch/issues/14831 does not crash with this change.

### DIFF
--- a/backends/xnnpack/runtime/XNNWeightsCache.cpp
+++ b/backends/xnnpack/runtime/XNNWeightsCache.cpp
@@ -11,6 +11,9 @@
 #include <executorch/runtime/core/memory_allocator.h>
 #include <sys/stat.h>
 #include <xnnpack.h>
+#include <exception>
+#include <memory>
+#include <new>
 #include <string>
 #include <vector>
 
@@ -155,21 +158,45 @@ size_t XNNWeightsCache::look_up(
   return packed_weight_entry->second.offset;
 }
 
+/**
+ * Reserve space in the weight cache for n bytes of weight data, aligned to
+ * context->kPackedAllocationAlignment. This function will return nullptr if
+ * the allocation fails.
+ */
 void* XNNWeightsCache::reserve_space(XNNWeightsCache* context, size_t n) {
   // MemoryAllocator* allocator = context->runtime_allocator_;
   // void* reserved_pointer = allocator->allocate(n,
   // context->kPackedAllocationAlignment);
 
   // return reserved_pointer;
-  std::string data_container;
-  data_container.resize(n + context->kPackedAllocationAlignment);
-  void* maybe_aligned_space = data_container.data();
-  void* aligned_space = (void*)((intptr_t)maybe_aligned_space + 64 -
-                                (intptr_t)maybe_aligned_space % 64);
+  try {
+    std::string data_container;
+    size_t raw_allocation_size = n + context->kPackedAllocationAlignment - 1;
+    data_container.resize(raw_allocation_size);
 
-  context->packed_pointer_to_container_[aligned_space] =
-      std::move(data_container);
-  return aligned_space;
+    void* maybe_aligned_space = data_container.data();
+    void* aligned_space = std::align(
+        context->kPackedAllocationAlignment,
+        n,
+        maybe_aligned_space,
+        raw_allocation_size // Note that std::align mutates this value.
+    );
+    ET_CHECK_MSG(aligned_space != nullptr, "Memory alignment failed.");
+
+    context->packed_pointer_to_container_[aligned_space] =
+        std::move(data_container);
+    return aligned_space;
+  } catch (std::bad_alloc& e) {
+    // XNNPACK can gracefully handle allocation failures, so return nullptr.
+    // We want to be able to recover from a failed attempt to load a large
+    // model without a crash.
+    ET_LOG(
+        Error,
+        "XNN weight cache failed to allocate %zu bytes: %s.",
+        n,
+        e.what());
+    return nullptr;
+  }
 }
 
 size_t XNNWeightsCache::look_up_or_insert(


GregoryComer:
CI failures are due to running on a fork or broken trunk.

mergennachin:
This is great, thank you @GregoryComer 

Is it possible to write a regression test?

backends/xnnpack/runtime/XNNWeightsCache.cpp
```diff
@@ -162,10 +163,14 @@ void* XNNWeightsCache::reserve_space(XNNWeightsCache* context, size_t n) {
 
   // return reserved_pointer;
   std::string data_container;
-  data_container.resize(n + context->kPackedAllocationAlignment);
+  auto raw_allocation_size = n + context->kPackedAllocationAlignment;
+  data_container.resize(raw_allocation_size);
   void* maybe_aligned_space = data_container.data();
-  void* aligned_space = (void*)((intptr_t)maybe_aligned_space + 64 -
-                                (intptr_t)maybe_aligned_space % 64);
+  void* aligned_space = std::align(
```
mergennachin: assert the aligned_space  != nullptr

mergennachin: Also add a docblock to reserve_space method

backends/xnnpack/runtime/XNNWeightsCache.cpp
```diff
@@ -155,21 +157,45 @@ size_t XNNWeightsCache::look_up(
...
+        n,
+        maybe_aligned_space,
+        raw_allocation_size // Note that std::align mutates this value.
+    );
+    ET_CHECK_MSG(aligned_space != nullptr, "Memory alignment failed.");
+
+    context->packed_pointer_to_container_[aligned_space] =
+        std::move(data_container);
+    return aligned_space;
+  } catch (std::exception& e) {
```
mergennachin: is there any particular exception type you're expecting? usually not a good practice to catch base exception, recommend specifying the exception type

digantdesai:
>  it casts the void* to a (signed) intptr_t. 

sigh

backends/xnnpack/runtime/XNNWeightsCache.cpp
digantdesai: Why do we need to use `try..catch` in the first place?

digantdesai (approved):
stamping to unblock, let's cp in 1.0

backends/xnnpack/runtime/XNNWeightsCache.cpp
```diff
@@ -155,21 +157,45 @@ size_t XNNWeightsCache::look_up(
...
-  void* maybe_aligned_space = data_container.data();
-  void* aligned_space = (void*)((intptr_t)maybe_aligned_space + 64 -
-                                (intptr_t)maybe_aligned_space % 64);
-
-  context->packed_pointer_to_container_[aligned_space] =
-      std::move(data_container);
-  return aligned_space;
+  try {
+    std::string data_container;
+    auto raw_allocation_size = n + context->kPackedAllocationAlignment;
```
psiddh: super nits:
- better to use size_t for simple types than 'auto', makes it better in terms of readability
- Also 
size_t raw_allocation_size = n + context->kPackedAllocationAlignment - 1;  ( 1 less extra byte )

backends/xnnpack/runtime/XNNWeightsCache.cpp
GregoryComer: This is here because Mergen requested asserting non-null output from std::align (it returns null on error if it can't align to the requested params). We also want non-fatal OOM handling. XNNPACK can gracefully handle null weight pointers (at least from a quick glance).

Since asserts are fatal, I needed to distinguish between null in the context of memory allocation failed vs null due to std::align failing. String wont return null data, but since I'm adding the assert, I want to make it robust for if we swap to using the ET allocator later. String can throw both bad_alloc and length_error, which both should be treated as an allocation failure. I'm not entirely sure why the code is using a string over a unique_ptr<uint8_t[]> or similar, but I wanted to keep the changes minimal. I'm open to suggestions.

backends/xnnpack/runtime/XNNWeightsCache.cpp
GregoryComer: Updated.

GregoryComer:
> Is it possible to write a regression test?

This is a bit tricky as reproing the issue as is requires the allocation to be in the upper half of the address space. We do cover this code in CI in a number of places, but not on the right platforms. We could refactor it a bunch to allow a mock allocator in, but this is largely just testing the alignment logic. Since it's using std::align now, I don't necessarily see a ton of value in that.

I think the best takeaway would be to run more end to end model tests on Android (maybe via emulator?) in CI. That likely would have caught the bug, given that we see it pretty reliably on the Android demo app.

extension/android/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ if(TARGET vulkan_backend)
 endif()
 
 if(EXECUTORCH_BUILD_KERNELS_LLM)
-  list(APPEND link_libraries $<LINK_LIBRARY:WHOLE_ARCHIVE,custom_ops>)
```
mergennachin, line 140: intentional?

mergennachin (approved):

backends/xnnpack/runtime/XNNWeightsCache.cpp
GregoryComer: Updated.

extension/android/CMakeLists.txt
GregoryComer: Oops, fixed.

GregoryComer:
I'm re-testing locally and in CI after rebasing and minor cleanup. Hitting some seemingly unrelated errors with class loading of the Android LLM extension. Will land once I figure this out if everything looks good. Hopefully within the next hour or two.

kirklandsign:
Hi @GregoryComer what is the android error? Is it consistent?

GregoryComer:
> Hi @GregoryComer what is the android error? Is it consistent?

Here's what I'm seeing. Any ideas?
```
Abort message: 'JNI DETECTED ERROR IN APPLICATION: JNI NewGlobalRef called with pending exception java.lang.NoSuchMethodError: no static or non-static method "Lorg/pytorch/executorch/extension/llm/LlmModule;.initHybrid(ILjava/lang/String;Ljava/lang/String;FLjava/lang/Object;)Lcom/facebook/jni/HybridData;"
```

This started after I rebased and switched branches, as well as closed and re-setup my terminal env, so I'm guessing it's on my end. CI seems to pass. I'm testing on the base commit now.

kirklandsign:
> > Hi @GregoryComer what is the android error? Is it consistent?
> 
> Here's what I'm seeing. Any ideas?
> 
> ```
> Abort message: 'JNI DETECTED ERROR IN APPLICATION: JNI NewGlobalRef called with pending exception java.lang.NoSuchMethodError: no static or non-static method "Lorg/pytorch/executorch/extension/llm/LlmModule;.initHybrid(ILjava/lang/String;Ljava/lang/String;FLjava/lang/Object;)Lcom/facebook/jni/HybridData;"
> ```
> 
> This started after I rebased and switched branches, as well as closed and re-setup my terminal env, so I'm guessing it's on my end. CI seems to pass. I'm testing on the base commit now.

Does https://github.com/pytorch/executorch/pull/15067 fix the issue?

GregoryComer:
> > > Hi @GregoryComer what is the android error? Is it consistent?
> > 
> > 
> > Here's what I'm seeing. Any ideas?
> > ```
> > Abort message: 'JNI DETECTED ERROR IN APPLICATION: JNI NewGlobalRef called with pending exception java.lang.NoSuchMethodError: no static or non-static method "Lorg/pytorch/executorch/extension/llm/LlmModule;.initHybrid(ILjava/lang/String;Ljava/lang/String;FLjava/lang/Object;)Lcom/facebook/jni/HybridData;"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This started after I rebased and switched branches, as well as closed and re-setup my terminal env, so I'm guessing it's on my end. CI seems to pass. I'm testing on the base commit now.
> 
> Does #15067 fix the issue?

I'm still seeing it after rebasing onto the latest main. Since CI is clean, I'm hopeful that it's a local issue. I'm going to merge, pick, and build RC3 and validate there.

kirklandsign:
Hi @GregoryComer what device did you validate?

GregoryComer:
> Hi @GregoryComer what device did you validate?

I validated on an emulated Pixel 9 from an M1 host.

**Edit: I tried the RC3 AAR on the same emulator instance and it works - no segfault.**

PR #15043, opened 2025-10-13
Qualcomm backend documentation update

cccclai:
Added detailed sections on QnnPartitioner, quantization, operator support, and custom ops support.


digantdesai (commented):
Split Quantizer, Partitioner, Op_support into different pages?

cccclai:
> Split Quantizer, Partitioner, Op_support into different pages?

We discussed that they can populated in the page but doesn't necessarily separate. Otherwise, it's just a really short page with a few sentences...

digantdesai (approved):

PR #15049, opened 2025-10-13
Update top-level README.md file

mergennachin:


psiddh (approved):
Few links use /main/ branch, Looks like we need to explicitly update them to point to 1.0 branch when this gets cherry-picked

Issue #14419, opened 2025-09-18
Pin torchao 0.14


GregoryComer:
We will need to pin torchao 0.14 for the 1.0 release when available. We're targeting 0.13 currently, but it is not fully compatible with pytorch 2.9. The release for ao 0.14 is not yet published, but it is planned and we will need to update the pin to 0.14 before releasing ExecuTorch 1.0.

GregoryComer:
AO 0.14.0 has been fully released. I've put up https://github.com/pytorch/executorch/pull/15071 to test the pin update. Hopefully everything is smooth. If so, I'll merge that and close this issue.

GregoryComer:
Done with #15071.

PR #15071, opened 2025-10-13
[RELEASE ONLY] Use torchao 0.14.0

GregoryComer:
### Summary
AO has officially released 0.14.0 (see https://pypi.org/project/torchao/#history). Update the pyproject.toml and pins to use 0.14. I also bumped the submodule to the latest release/0.14.0 head, though this only picks up a minor GH actions workflow change.

Fixes https://github.com/pytorch/executorch/issues/14419.

### Test plan
CI

### DIFF
--- a/install_requirements.py
+++ b/install_requirements.py
@@ -111,6 +111,7 @@ def install_requirements(use_pytorch_nightly):
     )
 
     LOCAL_REQUIREMENTS = [
+        "third-party/ao",  # We need the latest kernels for fast iteration, so not relying on pypi.
     ] + (
         [
             "extension/llm/tokenizers",  # TODO(larryliu0820): Setup a pypi package for this.
--- a/install_requirements.py
+++ b/install_requirements.py
@@ -111,7 +111,6 @@ def install_requirements(use_pytorch_nightly):
     )
 
     LOCAL_REQUIREMENTS = [
-        "third-party/ao",  # We need the latest kernels for fast iteration, so not relying on pypi.
     ] + (
         [
             "extension/llm/tokenizers",  # TODO(larryliu0820): Setup a pypi package for this.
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -78,6 +78,7 @@ dependencies=[
   "scikit-learn==1.7.1",
   "hydra-core>=1.3.0",
   "omegaconf>=2.3.0",
+  "torchao=0.14.0",
```
mergennachin: Double equals

install_requirements.py
```diff
@@ -111,7 +111,6 @@ def install_requirements(use_pytorch_nightly):
     )
 
     LOCAL_REQUIREMENTS = [
-        "third-party/ao",  # We need the latest kernels for fast iteration, so not relying on pypi.
```
mergennachin, line 114: we actually keep it since we need low-bit kernels

GregoryComer: My intent was that it would be installed with pyproject.toml (instead of through this script). Is there a case I'm missing where that isn't sufficient?

pyproject.toml
GregoryComer: Fixed.

install_requirements.py
mergennachin: cc @metascroy

mergennachin: Iirc correctly, in the default case, pyproject.toml is sufficient. But there are cases where we need to build LLM runner from source especially if we need low-bit kernels like 2bit kernels.

GregoryComer: Ah, okay. Thanks. I'll remove this change.

GregoryComer: Updated.

GregoryComer:
CI looks clean. Marking as ready for review.

lucylq (approved):

PR #15073, opened 2025-10-13
Core ML doc updates

metascroy:
CoreML doc updates


docs/source/backends/coreml/coreml-op-support.md
```diff
@@ -0,0 +1,5 @@
+# Op support
+
+The Core ML backend supports almost all PyTorch operators.
```
digantdesai, line 3: Is there a reference to the coreml repo page or something we want to link here?

docs/source/backends/coreml/coreml-overview.md
```diff
@@ -0,0 +1,112 @@
...
+To target the Core ML backend during the export and lowering process, pass an instance of the `CoreMLPartitioner` to `to_edge_transform_and_lower`. The example below demonstrates this process using the MobileNet V2 model from torchvision.
+
+```python
+import torch
+import torchvision.models as models
+from torchvision.models.mobilenetv2 import MobileNet_V2_Weights
+from executorch.backends.apple.coreml.partition import CoreMLPartitioner
+from executorch.exir import to_edge_transform_and_lower
+
+mobilenet_v2 = models.mobilenetv2.mobilenet_v2(weights=MobileNet_V2_Weights.DEFAULT).eval()
```
digantdesai, line 44: curious, not v3?

digantdesai (approved):
Thanks @metascroy for a quick turn around on the new template.

docs/source/backends/coreml/coreml-op-support.md
metascroy: Not a doc page.  There is a code page that people can search, but we'd also need to keep it synced with the commit that ET uses.

They also have a python utility that you can use to query whether an op is supported and this is what the partitioner uses.

docs/source/backends/coreml/coreml-overview.md
metascroy: I didn't change the content, just rearranged into different places.

When I first put this page together for 0.6 I matched the example that XNNPACK used (which was mv2 I guess)

PR #15075, opened 2025-10-13
Added support for Stable Diffusion LCM model

suryasidd:
### Summary
Added initial support for running stable diffusion lcm models. 

### Test plan
To test this PR, after installing the requirements. 
Use export_lcm.py to export  
`python export_lcm.py --output_dir` ./lcm_models --device CPU
Use openvino_lcm.py to infer
`python openvino_lcm.py --models_dir ./lcm_models --prompt "Sunset over mountains" --step 4`


examples/openvino/stable_diffusion/README.md
```diff
@@ -0,0 +1,48 @@
+# Stable Diffusion LCM with OpenVINO Backend
+
+This example demonstrates how to run Latent Consistency Models (LCM) for fast text-to-image generation on Intel hardware using ExecuTorch with the OpenVINO backend.
+
+## Overview
+
+Latent Consistency Models (LCMs) are optimized diffusion models that generate high-quality images in just 4-8 steps, compared to 25-50 steps required by traditional Stable Diffusion models. This implementation:
+
```
cavusmustafa, line 8: Maybe some content is missing here after "This implementation:" ?

examples/openvino/stable_diffusion/openvino_lcm.py
```diff
@@ -0,0 +1,311 @@
...
+logger = logging.getLogger(__name__)
+
+
+class OpenVINOLCMPipeline:
+    """OpenVINO optimized Latent Consistency Model pipeline for Intel hardware"""
+
+    def __init__(self, device: str = "CPU", dtype: torch.dtype = torch.float16):
+        self.device = device
+        self.dtype = dtype
+        self.models = {}
```
cavusmustafa: lintrunner MYPY fails on my side for multiple parts of the code including this:

Need type annotation for "models" (hint: "models: dict[<type>, <type>]
    = ...")
    
 They might need to be fixed manually

examples/openvino/stable_diffusion/openvino_lcm.py
```diff
@@ -0,0 +1,311 @@
...
+from typing import Optional
+
+import torch
+from PIL import Image
+
+try:
+    from diffusers import LCMScheduler
+    from transformers import CLIPTokenizer
+except ImportError:
+    raise ImportError("Please install diffusers and transformers: pip install diffusers transformers")
```
cavusmustafa: lintrunner UFMT fails on my side in multiple parts including this line. These can be resolved by running below:

lintrunner init
lintrunner -a

cavusmustafa (changes requested):

examples/openvino/stable_diffusion/requirements.txt
```diff
@@ -0,0 +1 @@
+diffusers==0.29.0
```
ynimmaga: This seems to be an older diffusers version. Is this a hard requirement or minimum requirement?

examples/openvino/stable_diffusion/README.md
suryasidd: Fixed this

examples/openvino/stable_diffusion/openvino_lcm.py
suryasidd: Fixed the mypy issues

examples/openvino/stable_diffusion/openvino_lcm.py
suryasidd: Verified that lintrunner passes for all the files

examples/openvino/stable_diffusion/requirements.txt
suryasidd: Made this the minimum requirement.

examples/openvino/stable_diffusion/export_lcm.py
```diff
@@ -0,0 +1,375 @@
...
+                def forward(self, input_ids):
+                    # Call text encoder and extract last_hidden_state
+                    output = self.text_encoder(input_ids, return_dict=True)
+                    return output.last_hidden_state
+
+            text_encoder_wrapper = TextEncoderWrapper(self.text_encoder)
+            text_encoder_wrapper.eval()
+
+            # Create dummy input for text encoder
+            dummy_input_ids = torch.ones(1, 77, dtype=torch.long)
```
kimishpatel: is it with static input size?

kimishpatel (commented):
looks largely good. Can you add some perf numbers on different backends

examples/openvino/stable_diffusion/export_lcm.py
```diff
@@ -0,0 +1,375 @@
+# Copyright (c) Intel Corporation
```
mergennachin, line 1: Can you decouple the model definition and OpenVINO specific lowering? 

For the model definition, you can add directly in examples/models/... that is backend agnostic components (but we should still keep the examples/openvino/stable_diffusion/export_lcm.py script that contains OpenVINO specific lowering) 

That way, we can try on different backends too, such as on CPU via XNNPACK

mergennachin (changes requested):
Putting back in the queue. See my minor previous comment

examples/openvino/stable_diffusion/export_lcm.py
suryasidd: Hey mergen, makes sense. We were also thinking about making more improvements in subsequent PRs like also having a common runner for stable diffusion that can be used with cpp as well, quantization etc. Currently there is a separate one only for Qualcomm i think. I will decouple the model definition and put in examples/models

examples/openvino/stable_diffusion/export_lcm.py
suryasidd: I added it with static shapes as there was an issue with dynamic shapes in OpenVINO backend. I will try to resolve those issues and create another PR as an improvement.

suryasidd:
> Putting back in the queue. See my minor previous comment

Separated the model definition and lowering.

mergennachin (approved):

cavusmustafa (approved):

PR #15100, opened 2025-10-14
Cortex_m backend: Fix add implementation

AdrianLundell:
- Call CMSIS-NN kernel with correct argument order and signs
- Change python implementation to reflect CMSIS-NN behaviour
- Fix scale calculations
- Remove broken broadcasting support
- Add pass to lower scalar version ops
- Remove unused definitions/ implementations in operators.py, operators.yaml and op_quantized_add.cpp

Note: arm_elementwise_add_s8 does not natively support broadcasting, so simply resizing the output tensor will not work. Enabling this in an efficient way is not straight forward, so avoid fusing these ops for now to avoid break graphs.


backends/cortex_m/ops/op_quantized_add.cpp
```diff
@@ -47,13 +48,6 @@ Tensor& quantized_add_out(
       output_shift,
       out);
```
psiddh, line 50: Should we the add an explicit shape check before running the add operation ?
If shapes do not match, return an error and log a clear message (e.g., "Input tensors must have the same shape for CMSIS-NN add operation. Broadcasting is not supported.").  (or) something like that..

backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -49,6 +50,12 @@ inline void validate_cmsis_nn_tensor_requirements(
       "Output dtype must be %hhd, got %hhd",
       expected_dtype,
       output.scalar_type());
+  ET_CHECK_MSG(
+      input1.sizes() == input2.sizes(),
+      "Input1 and Input2 must have the same sizes");
+  ET_CHECK_MSG(
+      output.sizes() == input1.sizes(),
+      "Output must have the same sizes as inputs");
```
psiddh, line 58: Adding these two checks in this helper makes it more stricter , making it harder for this hlper function to be reused for other kernel implementations that could support 'broadcast'.

One way could be to move these functions to a separate helper function : validate_cmsis_nn_no_broadcast_requirements(...)

// Add a specific one for non-broadcast operations
validate_cmsis_nn_no_broadcast_requirements(input1, input2, out);

backends/cortex_m/passes/quantized_op_fusion_pass.py
```diff
@@ -58,7 +60,16 @@ def _get_quant_targets(self) -> Set:
...
+        is_supported = (
+            node.op == "call_function" and node.target in self.SUPPORTED_OPS_MAPPING
+        )
+        if not is_supported:
+            return False
+
+        shape1 = node.args[0].meta["val"].shape
+        shape2 = node.args[1].meta["val"].shape
+        is_broadcast = shape1 != shape2
+        return not is_broadcast
```
psiddh, line 72: Does this mean all CMSIS-NN binary ops don't support broadcasting ?

psiddh: lgtm, Pls look at my comments and if applicable , address them before merging

psiddh (approved):

backends/cortex_m/ops/cortex_m_ops_common.h
AdrianLundell: We can always change this later if needed since this is WIP, I will merge it for now.

backends/cortex_m/passes/quantized_op_fusion_pass.py
AdrianLundell: Min/max does I think, anyways this is temporary since we are planning to do something about broadcasting pretty soon.

GregoryComer:
It looks like the test-mcu-models (cmake) / linux-job CI job has been failing since this PR merged. Example failure: https://github.com/pytorch/executorch/actions/runs/18869591982/job/53844591126. Can you take a look? Thanks.
CC @AdrianLundell @psiddh

AdrianLundell:
We should remove that test IMO, it was only testing the lowering and not that we were getting reasonable results, and we have pytests which covers that now. I can create a patch for that today.

PR #15101, opened 2025-10-14
Arm backend: Add support for sigmoid and tanh int16x8

SaoirseARM:
### Summary
Adds support for sigmoid and tanh to arm backend for int16x8 support 
Removes unnecessary test_sigmoid_16bit as these are now covered by test_sigmoid


backends/arm/test/ops/test_sigmoid.py
```diff
@@ -269,22 +270,23 @@ def get_symmetric_a16w8_sigmoid_quantizer(per_channel_quantization=False):
     }
 
     quantizer = TOSAQuantizer(tosa_profiles[tosa_version])
+
+    # Use a smaller episilon value to not greatly inflate [qmin, qmax]
     quantizer.set_global(
-        get_symmetric_a16w8_quantization_config(is_per_channel=per_channel_quantization)
+        get_symmetric_a16w8_quantization_config(
+            is_per_channel=per_channel_quantization, epsilon=2**-16
```
Ninja91, line 277: How does this change impact other ops?

backends/arm/test/ops/test_sigmoid.py
```diff
@@ -311,7 +313,7 @@ def test_sigmoid_16a8w_tosa_INT(test_data: torch.Tensor):
 @common.parametrize("test_data", test_data_suite)
 @common.XfailIfNoCorstone300
 @pytest.mark.xfail(
-    reason="Vela compilation fails with 'Invalid arguments' for int16 sigmoid operations"
+    reason="MLETORCH-707: AssertionError: Output 0 does not match reference output."
```
Ninja91, line 316: This indicates we don't yet have full support for U55. Can you comment on what's remaining? cc: @3l1

Ninja91 (commented):
Thanks for adding this support! Hopefully we will have full support in next few days.

backends/arm/test/ops/test_sigmoid.py
SaoirseARM: Thanks for review! 
There could be other Table operators that may need this change I will update those as I go through all operators. But, this change will not impact others as its stands because its only set in the unit tests.

backends/arm/test/ops/test_sigmoid.py
SaoirseARM: This is more so a limitation in terms of U55 support and not a int16x8 specific issue. 
Thanks for review!

backends/arm/test/ops/test_sigmoid.py
Ninja91: It's set in unit tests so that the sig/tanh can be partitioned to U55/U85, right?

backends/arm/quantizer/arm_quantizer.py
```diff
@@ -161,6 +161,7 @@ def get_symmetric_a16w8_quantization_config(
     is_dynamic: bool = False,
     weight_qmin: int = -127,
     weight_qmax: int = 127,
+    epsilon: float = 2**-12,
```
Ninja91, line 164: Ok so in order for sig/tanh to be able to partition, we need to set this to 2**-16 right?

SaoirseARM: It can be partitioned without it, this is more so for numerical behavior. If the set epsilon is too high, the quantization arguments can be inflated resulting in an incorrect output. 
Thank you for review!

backends/arm/test/ops/test_sigmoid.py
SaoirseARM: As mentioned below, this is just to ensure we get the correct values on the output. 
Thanks for review!

Ninja91 (approved):

PR #15103, opened 2025-10-14
Arm backend: Move rescales from MUL visitor to pass

martinlsm:
Some TOSA ops do not support INT8 as inputs and outputs. Instead, only
INT32 is supported as a whole number type. Prior to this patch, the MUL
node visitor inserted rescale ops between the data types INT8 and INT32
before and after the operator such that it will accept its input and
output.

Change this by moving the insertion of the rescale ops to the pass
`InsertRescaleInt32Pass`. This will further enable optimizations
to the graph by fusing the rescale nodes.

### Test plan
Test coverage is found the modified test_insert_rescale_i32_pass.py.


Sebastian-Larsson (approved):
Unrelated CI failures

PR #15119, opened 2025-10-14
Switch ruamel-yaml default from 0.17.x to 0.18.15

mkillianey:
Summary: Change the default version of ruamel-yaml from 0.17.10 to 0.18.15 everywhere.

Reviewed By: parulgupta1004

Differential Revision: D83863515

### DIFF
--- a/exir/dialects/edge/_ops.py
+++ b/exir/dialects/edge/_ops.py
@@ -16,7 +16,6 @@ from executorch.exir.dialects.edge.dtype.supported import regular_tensor_str_to_
 from executorch.exir.dialects.edge.op.api import to_variant
 from executorch.exir.dialects.edge.spec.utils import get_tensor_variable_names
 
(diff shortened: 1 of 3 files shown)


mergennachin (approved):

exir/dialects/edge/_ops.py
```diff
@@ -16,7 +16,6 @@
 from executorch.exir.dialects.edge.op.api import to_variant
 from executorch.exir.dialects.edge.spec.utils import get_tensor_variable_names
 
-# pyre-ignore
```
manuelcandales, line 19: why are you removing these pyre-ignores?

mkillianey: Type information was not available in ruamel 0.17.x, but is in 0.18.15.  (I.e., the linter complained that this directive is no longer necessary after upgrading the package)

PR #15131, opened 2025-10-14
Fix precommit hook to extract commit hash from nightly commit title

larryliu0820:
Instead of using the nightly commit hash

As titled.

### DIFF
--- a/.github/scripts/update_pytorch_pin.py
+++ b/.github/scripts/update_pytorch_pin.py
@@ -4,7 +4,6 @@ import json
 import re
 import sys
 import urllib.request
-from datetime import datetime
 
 
 def parse_nightly_version(nightly_version):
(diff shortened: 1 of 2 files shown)


.github/scripts/update_pytorch_pin.py
```diff
@@ -75,13 +74,20 @@ def get_commit_hash_for_nightly(date_str):
         # Check if the first line of commit message matches
         first_line = commit_msg.split("\n")[0].strip()
         if first_line == target_title or first_line.startswith(f"{date_str} nightly"):
```
Gasoonjia: looks like the second condition cover the first one?

Gasoonjia (approved):
lgtm and i believe you've test locally right?

PR #15139, opened 2025-10-15
Dedup constants in emitter

lucylq:
### Summary
Fix bug where if we have two constants with different names but the same data, the second one is not saved and results in missing tensor. (This happened in quant flows).

### Test plan
```
python -m unittest executorch.exir.emit.test.test_emit.TestEmit.test_constant_tagged_tensor_dedup 
```

### DIFF
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -372,6 +372,21 @@ class _Emitter(torch.fx.Interpreter):
             )
         return allocation_info
 
+    def _save_to_external_constant_map(
+        self,
+        fqn: str,
+        buffer_idx: int,
+        constant_tag: str,
+    ) -> None:
+        """
+        Saves external constant to the map.
+        """
+        # buffer data should be in the external_constant_buffer already.
+        assert buffer_idx < len(self.program_state.external_constant_buffer)
+        if constant_tag not in self.program_state.external_constant_map:
+            self.program_state.external_constant_map[constant_tag] = {}
+        self.program_state.external_constant_map[constant_tag][fqn] = buffer_idx
+
     def _save_new_const_tensor(
         self,
         spec: TensorSpec,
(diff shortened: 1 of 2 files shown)


exir/emit/_emitter.py
```diff
@@ -466,6 +466,15 @@ def _tensor_spec_to_evalue(
...
                 buffer_idx = self.program_state.external_constant_hash.get(hashed, -1)
+                if buffer_idx != -1:
+                    # Save the constant tag for the external tensor
+                    if constant_tag not in self.program_state.external_constant_map:
+                        # pyre-ignore Undefined attribute [16]: `Optional` has no attribute `fully_qualified_name`.
+                        self.program_state.external_constant_map[constant_tag] = {}
+                    # pyre-ignore Undefined attribute [16]: `Optional` has no attribute `fully_qualified_name`.
+                    self.program_state.external_constant_map[constant_tag][
+                        spec.extra_tensor_info.fully_qualified_name
+                    ] = buffer_idx
```
mergennachin: also please assert that constant_tag is not None
can you deduplicate this code. i think there's another place in this file doing the same logic

exir/emit/_emitter.py
```diff
@@ -466,6 +466,15 @@ def _tensor_spec_to_evalue(
                 and spec.extra_tensor_info.location == TensorDataLocation.EXTERNAL
             ):
                 buffer_idx = self.program_state.external_constant_hash.get(hashed, -1)
+                if buffer_idx != -1:
+                    # Save the constant tag for the external tensor
```
mergennachin: Write a big inline comment on why this is necessary. This is quite subtle.

exir/emit/test/test_emit.py
```diff
@@ -1719,6 +1719,45 @@ def forward(self, x):
         self.assertEqual(external_map["linear.weight"], 0)
         self.assertEqual(external_map["linear.bias"], 1)
 
+    def test_constant_tagged_tensor_dedup(self) -> None:
```
mergennachin, line 1722: also add another test with 4 identical constants

mergennachin (approved):

lucylq:
thanks for the comments @mergennachin! Updated.

PR #15148, opened 2025-10-15
NXP Backend: Add pass to remove unnecessary Quantize/Dequantize nodes.

MartinPavella:
### Summary
This PR adds an edge dialect pre-processing pass to remove some Q/DQ nodes. This enables some non-delegated nodes (which run on the CPU) to run in directly in int8 and avoid the QDQ compute overhead. This improves the inference speed (by eliminating the need to artificially quantize and de-quantize input and output values.

### Test plan
Unit tests provided.

cc @robert-kalmar

### DIFF
--- a/backends/nxp/backend/edge_helper.py
+++ b/backends/nxp/backend/edge_helper.py
@@ -125,3 +125,14 @@ def previous_non_qdq_node(node: Node, input_index: int = 0) -> Node | None:
             current_node = current_node.args[0]
         else:
             return current_node
+
+
+Scale = list[float] | float
+ZeroPoint = list[int] | int
+
+
+def get_quantization_parameters_for(node: Node) -> tuple[Scale, ZeroPoint] | None:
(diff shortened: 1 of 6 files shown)


backends/nxp/tests/test_turning_batch_first_gru_to_time_major.py
roman-janik-nxp, line 1: This shouldn't be there.

roman-janik-nxp (approved):

robert-kalmar:
Update the Summary, the pass has different intention: 

This PR adds an edge dialect pre-processing pass to remove some Q/DQ nodes. This enables some non-delegated nodes (which run on the CPU) to run in directly in int8 and avoid the QDQ compute overhead. This improves the inference speed (by eliminating the need to artificially quantize and de-quantize input and output values.

robert-kalmar (approved):

PR #15167, opened 2025-10-15
Add Llama xnnpack recipe

jackzhxng:


examples/models/llama/config/llama_xnnpack.yaml
```diff
@@ -0,0 +1,17 @@
...
+  metadata: '{"get_bos_id":128000, "get_eos_ids":[128009, 128001]}'
+
+model:
+  use_sdpa_with_kv_cache: True
+  use_kv_cache: True
+  dtype_override: fp32
+
+quantization:
+  qmode: 8da4w
+  group_size: 128
```
digantdesai, line 11: does this work OK, i.e. produce correct output?

jackzhxng: I haven't tried recently but I've definitely used this combination of parameters a few months ago when I was working on export_llm and it worked

jackzhxng: (aka these are some of the most common options)

digantdesai: Let me run it..

digantdesai: Seems to produce sane output. Let's land. CI?

digantdesai (approved):

examples/models/llama/config/llama_xnnpack.yaml
jackzhxng: Should already be covered in one of our many Llama CI tests! This combination params is pretty standard

PR #15168, opened 2025-10-15
Use torch 2.9 release packages

GregoryComer:
### Summary
Use release 2.9 packages for pytorch and domain libraries in preparation for the final 1.0 release. I'm following the pattern from https://github.com/pytorch/executorch/pull/13248.

### DIFF
--- a/.ci/docker/common/install_pytorch.sh
+++ b/.ci/docker/common/install_pytorch.sh
@@ -12,11 +12,11 @@ source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
 
 install_domains() {
   echo "Install torchvision and torchaudio"
-  pip install torchvision==0.24.0 torchaudio==2.9.0
+  pip install torchvision==0.24.0 torchaudio==2.9.0 --index-url https://download.pytorch.org/whl/cpu
 }
 
 install_pytorch_and_domains() {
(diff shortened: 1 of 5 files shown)


mergennachin:
Do you still need these lines? 

https://github.com/pytorch/executorch/blob/release/1.0/install_requirements.py#L150-L167

This will be invoked for people install from source in release branch. We should just rely on requirements-examples.txt

GregoryComer:
CI errors are apparently due to installing the CUDA version of torch. Looking into this now.
```
CMakeLists.txt:794 (find_package_torch)
  CMake Error at /opt/conda/envs/py_3.10/lib/python3.10/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:90 (message):
    Your installed Caffe2 version uses CUDA but I cannot find the CUDA
    libraries.  Please set the proper CUDA prefixes and / or install CUDA.
```

Edit: Added index-url back into the install_pytorch.sh script to use cpu variants. Re-testing.

GregoryComer:
> Do you still need these lines?
> 
> https://github.com/pytorch/executorch/blob/release/1.0/install_requirements.py#L150-L167
> 
> This will be invoked for people install from source in release branch. We should just rely on requirements-examples.txt

Done. Good catch.

kirklandsign (approved):

larryliu0820 (approved):

PR #15170, opened 2025-10-15
[CoreML] Add retry logic to database/key-value store

metascroy:
This PR introduces changes to the database and key-value store to retry if there are certain kinds of failures.

I've confirmed it can recover and re-create the DB if the DB is deleted during predictions.

The "std::cout" statements will be removed before landing.

### DIFF
--- a/backends/apple/coreml/runtime/kvstore/database.cpp
+++ b/backends/apple/coreml/runtime/kvstore/database.cpp
@@ -8,6 +8,7 @@
 
 #include <database.hpp>
 
+#include <iostream> // @nocommit
 #include <sqlite_error.hpp>
 
 namespace {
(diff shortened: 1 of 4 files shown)


metascroy:
@cymbalrush can you have a look?

cymbalrush:
@metascroy do we know for sure if the database file is being deleted?
All reads and writes to the database happen on the sync queue ([ETCoreMLAssetManager.mm](https://github.com/pytorch/executorch/blob/main/backends/apple/coreml/runtime/delegate/ETCoreMLAssetManager.mm)), so access should be fully serialized. An SQLite transaction shouldn’t fail under serialized access.

If the database file is being deleted, that’s concerning — the OS will not automatically remove files from the Application Support directory. If the application itself is deleting it, can we exclude this file from that process?

Also, could we log the SQLite error code to confirm what’s actually failing? is it failing when the app is backgrounded?

PR #15180, opened 2025-10-16
[aoti-et] Store weights outside of .so

larryliu0820:
This pull request introduces significant changes to how model constants (weights) are packaged and loaded in the CUDA backend, improving modularity and flexibility. It also updates the runner APIs to support more flexible loading modes and bumps PyTorch commit pins and nightly version references.

**Packaging and Loading Model Constants:**

* Model constants are now separated from the `.so` file and stored as a binary blob on disk, rather than being packaged directly into the shared object. The preprocessing logic in `cuda_backend.py` is updated to handle the new file outputs and manage cleanup. [[1]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aL149-R153) [[2]](diffhunk://#diff-5b5ea2257772b3aba04b2534f5ea1429a0c631bfd25a7ef531f526e76c471d7aL165-R211)
* The C++ CUDA backend (`cuda_backend.cpp`) now loads the new weights blob from the named data map and feeds it into the model container using a newly added API function. The buffer is freed immediately after use for better resource management. [[1]](diffhunk://#diff-a4b17eccf1aa933837671c5184e02bc815d934a362344bb2b17b789cdfaa5375R153-R154) [[2]](diffhunk://#diff-a4b17eccf1aa933837671c5184e02bc815d934a362344bb2b17b789cdfaa5375R183-R195)

**API and Infrastructure Updates:**

* A new function pointer type, `AOTInductorModelUpdateConstantsFromBlobFunc`, is added to the delegate handle structure in `aoti_delegate_handle.h` to support updating model constants from a binary blob. [[1]](diffhunk://#diff-0598c198d53bf756f6013186ea3155f15ddef247aa863e83ef30f27991b3a0a7R74-R78) [[2]](diffhunk://#diff-0598c198d53bf756f6013186ea3155f15ddef247aa863e83ef30f27991b3a0a7R95)
* The CUDA backend now loads this new symbol from the shared object at runtime.

**Runner API Improvements:**

* The multimodal runner API is updated to accept a `Module::LoadMode` parameter, allowing for more flexible loading options such as memory mapping. This change is propagated through helper functions and their headers. [[1]](diffhunk://#diff-0ac16dbe4eaefa08e21fbda582fe2cd2b482f43aaedfc1bf2f31becf5e7bb843L322-R322) [[2]](diffhunk://#diff-005ac94c6b217e02d652aafc206d36b2ec1190af36aa0a632fd406975dfc2600L271-R272) [[3]](diffhunk://#diff-005ac94c6b217e02d652aafc206d36b2ec1190af36aa0a632fd406975dfc2600L281-R284) [[4]](diffhunk://#diff-ac7a381a7828a6f1a543d2beab4cf503c2d3547ab86821c8e1777df9305108aaL143-R144)

**Dependency Updates:**

* The PyTorch commit pin is updated in `.ci/docker/ci_commit_pins/pytorch.txt` and the nightly version is bumped in `torch_pin.py` for compatibility with the new packaging logic. [[1]](diffhunk://#diff-e873e85ae7aa52ebeadb13a27cf83eff1891b1011e27f94ec040eb8407893c5eL1-R1) [[2]](diffhunk://#diff-9665391232bd21d4ee0a293cbc7f76d99db902ab1e6e045a59f9a132325babc9L2-R2)


backends/cuda/cuda_backend.py
```diff
@@ -162,25 +165,50 @@ def preprocess(
...
+            if path.endswith(".wrapper.so"):
+                so_path = path
+            elif path.endswith(".wrapper_weights.blob"):
+                blob_path = path
+
+        if so_path is None:
+            raise RuntimeError(
+                f"Could not find .wrapper.so file in compiled paths, got {paths}"
+            )
+
```
mergennachin, line 190: check if existence/non-existence of blob_path matches the options

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -83,6 +83,12 @@ class ET_EXPERIMENTAL CudaBackend final
 
     LOAD_SYMBOL(handle, run, AOTInductorModelContainerRun, so_handle);
 
+    LOAD_SYMBOL(
+        handle,
+        update_constants_from_blob,
+        AOTInductorModelUpdateConstantsFromBlob,
+        so_handle);
```
mergennachin: This is not backwards compatible

LOAD_SYMBOL fails if the symbol doesn't exist. 

We sometimes need to load the old style .so files -- without the weight embedded into so

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -172,6 +180,19 @@ class ET_EXPERIMENTAL CudaBackend final
...
+    auto buffer_res = named_data_map->get_data(weights_blob_key.c_str());
+    if (buffer_res.ok()) {
+      ET_LOG(Info, "Found %s in named data map", weights_blob_key.c_str());
+      const void* weights_blob = buffer_res->data();
+      // Feed the weights blob into the container. Under the hood it's copying
+      // weights, so we should free the buffer immediately.
+      ET_CHECK_OK_OR_RETURN_ERROR(handle->update_constants_from_blob(
+          handle->container_handle, static_cast<const uint8_t*>(weights_blob)));
+      buffer_res->Free();
+    }
```
mergennachin, line 188: if !buffer_res.ok() then why does it keep continue? in the else case, assert that handle->update_constants_from_blob == nullptr so that we can handle the old-style .so file without weights

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -172,6 +180,19 @@ class ET_EXPERIMENTAL CudaBackend final
...
     handle->container_handle = container_handle;
 
+    // Look into named data map for constant data
+    std::string weights_blob_key =
+        method_name.empty() ? "weights_blob" : method_name + "_weights_blob";
+    auto buffer_res = named_data_map->get_data(weights_blob_key.c_str());
+    if (buffer_res.ok()) {
+      ET_LOG(Info, "Found %s in named data map", weights_blob_key.c_str());
+      const void* weights_blob = buffer_res->data();
+      // Feed the weights blob into the container. Under the hood it's copying
```
mergennachin, line 183: assert that update_constants_from_blob!=nullptr

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: My understanding is this will always be there? @yushangdi can you confirm?

larryliu0820: I guess we won't have it before this API is added to pytorch/pytorch that's true.

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -60,29 +53,31 @@ class ET_EXPERIMENTAL CudaBackend final
   Error load_function_pointers_into_handle(
       void* so_handle,
       AOTIDelegateHandle* handle) const {
-    LOAD_SYMBOL(
-        handle,
-        create_with_device,
-        AOTInductorModelContainerCreateWithDevice,
-        so_handle);
+#define LOAD_SYMBOL(member, name)                                           \
```
JacobSzwejbka: can you do this in a follow up so there arent merge conflicts with https://github.com/pytorch/executorch/pull/15183

examples/models/voxtral/multimodal.cpp
```diff
@@ -319,7 +319,7 @@ int32_t main(int32_t argc, char** argv) {
   // Create multimodal runner
   std::unique_ptr<::executorch::extension::llm::MultimodalRunner> runner =
       ::executorch::extension::llm::create_multimodal_runner(
-          model_path, std::move(tokenizer), data_path);
+          model_path, std::move(tokenizer), data_path, Module::LoadMode::Mmap);
```
JacobSzwejbka, line 322: lol so we were loading from disk -> writing to heap -> writing to disk -> loading temp from disk -> writing to cuda.

extension/llm/runner/llm_runner_helper.h
```diff
@@ -140,6 +140,7 @@ ET_EXPERIMENTAL std::unique_ptr<TextLLMRunner> create_text_llm_runner(
 ET_EXPERIMENTAL std::unique_ptr<MultimodalRunner> create_multimodal_runner(
     const std::string& model_path,
     std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
-    std::optional<const std::string> data_path = std::nullopt);
+    std::optional<const std::string> data_path = std::nullopt,
+    Module::LoadMode load_mode = Module::LoadMode::File);
```
JacobSzwejbka, line 144: I think mmap is the better default. Its experimental so changing the behavior is fine

backends/cuda/runtime/cuda_backend.cpp
mergennachin: @larryliu0820 - i guess my question is whether there will be .so file only without the weight_blob

larryliu0820: Yep updated

examples/models/voxtral/multimodal.cpp
larryliu0820: That's for the .so yeah. For the weights, we mmap it and then copy to cuda memory.

backends/cuda/runtime/cuda_backend.cpp
swolchok: would `aoti_dso_buffer` or `aoti_so_buffer` be a better name for this variable? I thought it was either a GPU buffer or a buffer containing cubin, but it's neither.

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -172,6 +169,19 @@ class ET_EXPERIMENTAL CudaBackend final
...
 
+    // Look into named data map for constant data
+    std::string weights_blob_key =
+        method_name.empty() ? "weights_blob" : method_name + "_weights_blob";
+    auto buffer_res = named_data_map->get_data(weights_blob_key.c_str());
+    if (buffer_res.ok() && handle->update_constants_from_blob != nullptr) {
+      ET_LOG(Info, "Found %s in named data map", weights_blob_key.c_str());
+      const void* weights_blob = buffer_res->data();
+      // Feed the weights blob into the container. Under the hood it's copying
+      // weights, so we should free the buffer immediately.
```
swolchok, line 184: the weights are mmapped, so this isn't halving the maximum amount of weights we can handle, right? even so, seems unfortunate that we have to copy and therefore can't keep them simply mmapped though; peak CPU memory now needs to hold them, right?

backends/cuda/cuda_backend.py
```diff
@@ -162,25 +165,49 @@ def preprocess(
...
 
         named_data_store = NamedDataStore()
         method_name = CudaBackend.method_name_from_compile_specs(compile_specs)
+
+        # Keep the so file in the NamedDataStore, so that it can be packaged into the .pte file.
+        named_data_store.add_named_data(method_name + "_so_blob", so_data, 1, None)
+
+        # Add weights blob to named data store
+        with open(blob_path, "rb") as f:
+            blob_data = f.read()
```
swolchok, line 203: I imagine we may want to mmap this in the future so that we can work with weights that don't necessarily all fit into CPU memory at once, but I guess that's not a goal for our first cut at the problem?

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: Feel free to merge #15183, I can rebase this one.

backends/cuda/runtime/cuda_backend.cpp
yushangdi: > My understanding is this will always be there? @yushangdi can you confirm?

It will always exist for anything compiled using the pytorch commit after the PR landed. https://github.com/pytorch/pytorch/commit/77bf23d85c47d6b89a6bd7839b41fa97f6ede7c4

> my question is whether there will be .so file only without the weight_blob

It's possible depending on which flag you use. If you use `package_constant_on_disk_format = "binary_blob"`, then the weight_blob file will always exist. Otherwise, there will not be the weight blob file.

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: > this isn't halving the maximum amount of weights we can handle, right?

Good point, let me test this on my RTX 5080.

> seems unfortunate that we have to copy and therefore can't keep them simply mmapped though

Yeah would be good if aoti can just take it without copying.

backends/cuda/cuda_backend.py
larryliu0820: the .so file should never contain weights, since it will have to go through a read-write roundtrip to be able to dlopen().

extension/llm/runner/llm_runner_helper.h
larryliu0820: I'll leave it to another PR. Here I'm not changing the default.

backends/cuda/runtime/cuda_backend.cpp
yushangdi: > Yeah would be good if aoti can just take it without copying.

I might be missing something here, but I assume you're mmaping into the CPU memory right? AOTI copies it into the CUDA memory, and since we're running this on CUDA, we have to copy it to CUDA some time.

larryliu0820: Is there a mmap equivalent on CUDA? If so, on ET side we can create a dataloader to directly load into CUDA memory.

yushangdi: I guess you can use GPUDirect Storage.

JacobSzwejbka (approved):

backends/cuda/cuda_backend.py
mergennachin: Only check for blob_path is None if package_constant_on_disk_format option is set to "binary_blob"

GregoryComer:
FYI Unit test jobs on trunk seem to be failing consistently since this PR due to OOM. I'm looking at bumping the runner size in #15400, but figured I'd mention it since there might a peak memory regression. I'm not sure why it passed on the PR, though. CC @larryliu0820

larryliu0820:
> FYI Unit test jobs on trunk seem to be failing consistently since this PR due to OOM. I'm looking at bumping the runner size in #15400, but figured I'd mention it since there might a peak memory regression. I'm not sure why it passed on the PR, though. CC @larryliu0820

Hmm majority of the changes in this PR are specific to CUDA. I would assume the OOM is caused by pytorch pin bump which happen to be in this PR as well.

PR #15182, opened 2025-10-16
Qualcomm AI Engine Direct - Suite operator fix part 3

winskuo-quic:
### Summary

- Supports transpose_conv with per_channel_quant. Previously, transpose_conv per_channel_quant is not done correctly, as the channel should be ch_axis=1 instead of ch_axis=0.
- Some test suite operators are using random values instead of desired input. These operations require specific range to prevent `NaN` as output.

### Test plan
UT added

cc @cccclai @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

### DIFF
--- a/backends/qualcomm/builders/node_visitor.py
+++ b/backends/qualcomm/builders/node_visitor.py
@@ -153,6 +153,13 @@ class NodeVisitor:
         scales, scale_offset, quantized_scales = quant_attrs[QCOM_SCALE], [], []
         # channel in observers defaults to zero
         num_channels = node.meta["val"].shape[0]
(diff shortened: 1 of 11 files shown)


winskuo-quic:
Hi @cccclai,
This PR works on fixing  conv_transpose per_channel quant. We should be able to get 5 or more conv_transpose tests fixed for each quant config.

Hi @GregoryComer 
I have updated some suite_operator test that is using **random input** instead of **desired random input**. Please let me know if it makes sense to you.

Thanks

cccclai (approved):
Thank you!

PR #15183, opened 2025-10-16
platform layer for windows and linux compatibility in cuda_backend

JacobSzwejbka:
### Summary
Abstract away posix apis which are not on windows.

Future PRs are coming quick. Should be a no-op for linux builds though and this file is under a lot of active refactoring and development so I want to quit having to rebase changes.


larryliu0820 (approved):

larryliu0820:
How do you test this?

JacobSzwejbka:
> How do you test this?

I have not tested the windows impl really (half of ET/extensions doesnt build with msvc yet). Linux is protected by CI though and the bigger thing to get through here is the refactor so that we dont keep merge conflicting each other.

PR #15197, opened 2025-10-16
Msvc inline macros

JacobSzwejbka:
### Summary
Add MSVC support to inline macro

### Test plan
Manual testing for now. Will add CI support when I get a full msvc build working.


Gasoonjia (approved):
trust you

runtime/platform/compiler.h
```diff
@@ -55,9 +55,17 @@
  */
 
 #define ET_NORETURN [[noreturn]]
+
+// Inline/NoInline
+#if defined(_MSC_VER)
+#define ET_NOINLINE __declspec(noinline)
+#define ET_INLINE __forceinline
+#define ET_INLINE_ATTRIBUTE
```
swolchok: looks like ET_INLINE_ATTRIBUTE is unused and should simply be deleted?

swolchok (approved):
matches c10, seems legit

PR #15200, opened 2025-10-16
Remove ET_UNWRAP usage

JacobSzwejbka:
Summary: Has issues with MSVC, so reintroduce the minor boilerplate. 

Differential Revision: D84844855


shoumikhin:
What was the issue?
It'd be nice to keep using the macro for brevity.

JacobSzwejbka:
> What was the issue? It'd be nice to keep using the macro for brevity.

Statement expressions are not supported in msvc so in general its not very portable. I have run into issues with it on windows MSVC trying to set up the cuda voxtral demo.

larryliu0820 (approved):

PR #15202, opened 2025-10-16
Add fpu embedded target to EthosUBackend

Daksh-Shami:
Summary: We need to add embedded fpu target as well for EthosUBackend as Arm FVP platform can use that for tests for some models.

Differential Revision: D84284543


swolchok (approved):

PR #15214, opened 2025-10-17
audio float API

kirklandsign:
### Summary
Some audio inputs are float[]
### Test plan
CI

cc @cbilgin

### DIFF
--- a/extension/android/jni/jni_layer_llama.cpp
+++ b/extension/android/jni/jni_layer_llama.cpp
@@ -340,6 +340,29 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
   }
 
   // Returns status_code
+  jint append_audio_input_float(
+      facebook::jni::alias_ref<jfloatArray> data,
+      jint batch_size,
+      jint n_bins,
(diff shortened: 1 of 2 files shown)


larryliu0820 (approved):

extension/android/jni/jni_layer_llama.cpp
```diff
@@ -339,6 +339,29 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
...
+  jint append_audio_input_float(
+      facebook::jni::alias_ref<jfloatArray> data,
+      jint batch_size,
+      jint n_bins,
+      jint n_frames) {
+    if (data == nullptr) {
+      return static_cast<jint>(Error::EndOfMethod);
+    }
+    auto data_size = data->size();
+    if (data_size != 0) {
```
psiddh, line 352: So is this condition '    if (data_size == 0) {'   also considered ok / success in this flow ? We seem to return ok at Ln 362

kirklandsign: Good point. We can add a check in java. I just don't want it crash here.

PR #15215, opened 2025-10-17
Use tokenizer eos

lucylq:
Summary:
Currently we check the tokenizer eos, and then override it with method eos. Sounds like this was for legacy cases where tokenizer did not store eos.

Use tokenizer eos as source of truth.

This is causing issues when we have special_tokens_map.json (from huggingface), which contains eos as a special id. It is overridden by the method, and llm generates excess output because we do not terminate correctly.

see https://github.com/meta-pytorch/tokenizers/pull/139 where support for `special_tokens_map.json` is added.

Differential Revision: D84865804

### DIFF
--- a/extension/llm/runner/llm_runner_helper.cpp
+++ b/extension/llm/runner/llm_runner_helper.cpp
@@ -154,27 +154,6 @@ std::unordered_set<uint64_t> get_eos_ids(
     tokenizers::Tokenizer* tokenizer,
     Module* module) {
   std::unordered_set<uint64_t> eos_ids = {tokenizer->eos_tok()};
-  // Get EOS IDs if available
-  auto method_names_result = module->method_names();
-  if (method_names_result.error() != Error::Ok) {
-    ET_LOG(Error, "Failed reading method names");
-    return eos_ids;
-  }
-  const auto& method_names = method_names_result.get();
-
-  if (method_names.count(llm::kEosIds)) {
-    eos_ids.clear();
-    auto execute_result = module->execute(llm::kEosIds);
-    if (execute_result.error() != Error::Ok) {
-      ET_LOG(Error, "Failed to execute %s", llm::kEosIds);
-      return eos_ids;
-    }
-    for (const auto& eos_id : execute_result.get()) {
-      auto value = eos_id.toScalar().to<int64_t>();
-      eos_ids.emplace(value);
-      ET_LOG(Info, "eos_id = %" PRId64, value);
-    }
-  }
   return eos_ids;
 }
 


jackzhxng (commented):
Which llm are you seeing this issue for? And do you know which models store eos and which don't? What about models that don't have a `special_tokens_map.json`?

lucylq:
> Which llm are you seeing this issue for? 

@jackzhxng it's overriding the correct eos token for a fine-tuned llama3_2 1B. See here: https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/blob/main/special_tokens_map.json
I added `special_tokens_map.json` parsing to tokenizers here: https://github.com/meta-pytorch/tokenizers/pull/139

> And do you know which models store eos and which don't? 

No, from discussion with @larryliu0820 it was legacy/older tokenizers/internal case but I'm not sure.

> What about models that don't have a `special_tokens_map.json`?

The eos should be in the main tokenizer.json. If there are multiple eos tokens, I think `special_tokens_map.json` tells you which one to use. e.g. I think <|eot_id|> is used for chat conversations (end of turn), where as <|end_of_text|> for non-chat.

jackzhxng (commented):
Hmm okay seems reasonable but where does the method eos come from originally? And we should be sure the tokenizers will always contain an eos

lucylq:
> Hmm okay seems reasonable but where does the method eos come from originally?

I think it comes from export time, if you export with 'get_eos_ids'.
It's still retrievable from the module, this removes it from the runner api.

>  And we should be sure the tokenizers will always contain an eos

Agree, though I think this requires us to expose this in the tokenizer API... currently tokenizer->eos_tok is an int, so hard to tell if it's valid or not...

extension/llm/runner/llm_runner_helper.cpp
```diff
@@ -154,27 +154,6 @@ std::unordered_set<uint64_t> get_eos_ids(
     tokenizers::Tokenizer* tokenizer,
     Module* module) {
   std::unordered_set<uint64_t> eos_ids = {tokenizer->eos_tok()};
-  // Get EOS IDs if available
```
larryliu0820, line 157: Yeah previously we rely on metadata inside of .pte to determine what eos to use. Partially because some models don't respect the tokenizer's eos. I think we should get rid of this logic to avoid confusion.

lucylq:
Update: I exported with `base.metadata='"{\"get_bos_id\":128000, \"get_eos_ids\":[128009, 128001]}"'` and that works well.

Not good UX for the user to manually specify eos ids because the tokenizer one is overridden, though.

PR #15218, opened 2025-10-17
Arm backend: Fuse duplicate user ops

AdrianLundell:
Adds a pass which checks if a node has multiple users performing equivalent operations its output. If that is the case, it fuses these ops into one.


zingo:
I'll rebase to get the test-mcu-model error out of the way

zingo:
Failed with 
FAILED backends/arm/test/ops/test_tanh.py::test_tanh_16a8w_tosa_INT[rand] - AssertionError: Output 0 does not match reference output.
	Given atol: 0.0010152587890625, rtol: 0.001.
	Output tensor shape: torch.Size([10, 10]), dtype: torch.float32
	Difference: max: 1.52587890625e-05, abs: 0.00152587890625, mean abs error: 2.6092529296875e-05.
	-- Model vs. Reference --
	 Numel: 100, 100
	Median: -0.0678253173828125, -0.0678253173828125
	  Mean: -0.0142254638671875, -0.014200897216796874
	   Max: 0.4605865478515625, 0.4621124267578125
	   Min: -0.4584808349609375, -0.4584808349609375
= 1 failed, 3050 passed, 2534 skipped, 2239 xfailed, 115 xpassed, 982 warnings in 2753.46s (0:45:53) =

This is probably just a flakey test Retrigger

zingo (approved):

zingo:
Added files in _passes is part of a buck2 *.py rule so this should work in buck2
Fails are unrelated

PR #15220, opened 2025-10-17
Vulkan backend: Add missing include of cstdint

martinlsm:
backends/vulkan/runtime/utils/StorageUtils.h was referring to definitions like `uint8_t` without including the header cstdint. This resulted in a risk of facing compilation errors on some systems. Fix this problem by including the header.


martinlsm:
This problem is causing issues in our CI-tests so it would be great if we could merge this patch ASAP

SS-JIA (approved):

digantdesai:
Thanks @SS-JIA

PR #15223, opened 2025-10-17
Remove internal executorch dependency on torchao.quantization.subclass

andrewor14:
Summary:
**Summary:** This is a really old quantization API that we
recently removed in torchao (D84842047). No one should be
calling it anymore. For BC, let's just copy the base class
into executorch for now. We should delete this in the future.

**Test Plan:** CI

Differential Revision: D84921134

### DIFF
--- a/examples/models/llama/experimental/subclass.py
+++ b/examples/models/llama/experimental/subclass.py
@@ -20,7 +20,11 @@
 #
 # This layout is handled internally in the tensor subclass.
 import torch
-from torchao.quantization.subclass import QuantizedLinearWeightBase
+from torch.utils._python_dispatch import return_and_correct_aliasing
+from typing_extensions import deprecated
+
+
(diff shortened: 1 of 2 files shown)


andrewor14:
@larryliu0820

examples/models/llama/experimental/subclass.py
```diff
@@ -129,6 +132,172 @@ def to_float(
     return a * scale.unsqueeze(1)
 
 
+class QuantizedLinearWeightBase(torch.Tensor):
```
lucylq, line 137: Can we tag this as deprecated? 

```
from executorch.exir._warnings import deprecated
.. 
@deprecated("QuantizedLinearWeightBase is deleted from torchao ... ")
class QuantizedLinearWeightBase(torch.Tensor):
```

lucylq (approved):

PR #15226, opened 2025-10-17
Msvc ops changes

JacobSzwejbka:
### Summary
MSVC really doesn't like the constexpr variables that I've removed in this PR. It was complaining about them not actually being static at compile time. Searching around suggests that its the act of putting them in a variable that breaks it and if I just place the value directly in all the used locations it should be fine. I tried that and it works.

The other changes are fixing up some compiler flags, adding a missing include, fixing an interaction between lambdas the ternary operator and  auto, and removing unneeded statement expressions

### Test plan
current ci to verify no regression on linux/mac. Manual testing of msvc will add ci soon


kernels/portable/cpu/op_view_as_real_copy.cpp
```diff
@@ -64,13 +64,13 @@ Tensor& view_as_real_copy_out(
...
       ctx, tensors_have_same_dim_order(self, out), InvalidArgument, out);
 
-  constexpr auto op_name = "view_as_real_copy.out";
-
-  ET_SWITCH_COMPLEXH_TYPES(self.scalar_type(), ctx, op_name, CTYPE_IN, [&] {
-    ET_SWITCH_FLOATH_TYPES(out.scalar_type(), ctx, op_name, CTYPE_OUT, [&] {
-      _to_impl<CTYPE_IN, CTYPE_OUT>(self, out);
-    });
-  });
+  ET_SWITCH_COMPLEX_TYPES(
```
JacobSzwejbka: oh lost the H on complexh

kernels/optimized/cpu/op_bmm.cpp
```diff
@@ -150,15 +150,14 @@ Tensor& opt_bmm_out(
   ET_KERNEL_CHECK(
       ctx, check_bmm_out_args(self, mat2, out), InvalidArgument, out);
 
-  constexpr auto name = "bmm.out";
```
swolchok, line 153: I don't suppose we are lucky enough that `static constexpr auto name = "bmm.out";` would fix it?

kernels/portable/cpu/util/elementwise_util.h
```diff
@@ -85,7 +85,6 @@ inline void dtype_specialized_elementwise_fn_impl(
   static_assert(
       (std::is_same_v<Args, std::pair<const Tensor*, SupportedTensorDtypes>> &&
        ...));
-  constexpr auto kNumInputs = sizeof...(inputs);
```
JacobSzwejbka, line 88: I tried a couple things to make msvc recognize kNumInputs here as being constant at compile time but it refused. Weirdly it does still recognize sizeof...(inputs) so I can just update the usage

kernels/optimized/CMakeLists.txt
```diff
@@ -21,7 +21,10 @@ if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
-set(_common_compile_options -Wno-deprecated-declarations)
+set(_common_compile_options
+    $<$<CXX_COMPILER_ID:MSVC>:/wd4996>
+    $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wno-deprecated-declarations>
+)
```
swolchok, line 27: I don't have a strong preference here, just noting that the following is also valid:

```
if (MSVC)
  set(_common_compile_options /wd4996)
else()
  set(_common_compile_options -Wno-deprecated-declarations)
endif()
```

I like that it avoids the ugly `$<$<NOT:$<X>>>`, but I don't like that it duplicates the variable name. up to your sense of taste.

kernels/portable/cpu/util/elementwise_util.h
swolchok, line 88: the strings are understandable (I had to be pretty careful just to get clang to accept them IIRC), but this one is pretty disappointing and also seems to work on godbolt. have you tried generating a minimal example?

kernels/portable/cpu/op_topk.cpp
```diff
@@ -118,19 +118,30 @@ void perform_topk(
...
-      const auto elem_greater = [](const elem_t& x, const elem_t& y) -> bool {
-        return float_less_than(y.first, x.first);
-      };
-      const auto elem_less = [](const elem_t& x, const elem_t& y) -> bool {
-        return float_less_than(x.first, y.first);
-      };
-      const auto cmp = largest ? elem_greater : elem_less;
-      if (use_partial_sort) {
-        std::partial_sort(queue, queue + k, queue + dim_size, cmp);
+      if (largest) {
```
swolchok: what's the problem with this one? these don't seem to be constexpr variables at all. error message might be helpful

swolchok (commented):
if this requires more than 1-2 more review rounds, you might be able to move faster by splitting out the uncontroversial strings changes; I could approve those right away.

kernels/portable/cpu/op_topk.cpp
JacobSzwejbka: Piping lambdas into the ternary was having issues. It couldnt resolve that they had the same return type

swolchok: Oh, you should be able to fix that by explicitly using the appropriate function pointer type instead of auto

kernels/optimized/cpu/op_bmm.cpp
JacobSzwejbka: It does work :). Ok Ill do that then...

JacobSzwejbka: Ill revist the size one and the lambdas too

kernels/optimized/cpu/op_bmm.cpp
```diff
@@ -150,9 +150,10 @@ Tensor& opt_bmm_out(
...
   auto self_type = self.scalar_type();
 
+  static constexpr auto name = "bmm.out";
+
   if (executorch::runtime::isComplexType(self_type)) {
```
swolchok, line 157: nit: probably want to minimize the change here

kernels/portable/cpu/op_view_as_real_copy.cpp
```diff
@@ -64,13 +64,13 @@ Tensor& view_as_real_copy_out(
   ET_KERNEL_CHECK(
       ctx, tensors_have_same_dim_order(self, out), InvalidArgument, out);
 
-  constexpr auto op_name = "view_as_real_copy.out";
```
JacobSzwejbka, line 67: oh missed one.

kernels/portable/cpu/op_topk.cpp
```diff
@@ -118,13 +118,16 @@ void perform_topk(
...
         return float_less_than(y.first, x.first);
       };
-      const auto elem_less = [](const elem_t& x, const elem_t& y) -> bool {
+      bool (*elem_less)(const elem_t&, const elem_t&) =
+          [](const elem_t& x, const elem_t& y) -> bool {
         return float_less_than(x.first, y.first);
       };
-      const auto cmp = largest ? elem_greater : elem_less;
+      bool (*cmp)(const elem_t&, const elem_t&) =
+          largest ? elem_greater : elem_less;
```
swolchok, line 131: FWIW, you can probably continue to use `auto` on the declarations of elem_greater and elem_less so long as you cast them when defining `cmp`. either way I would recommend a quick note that this is meant to pacify MSVC to save future readers from wasting time trying to clean it up

kernels/portable/cpu/op_view_as_real_copy.cpp
swolchok, line 67: why can't this one use `static`?

swolchok (approved):
looks good, just a couple minor comments

kernels/portable/cpu/op_view_as_real_copy.cpp
JacobSzwejbka: it can version history was just messed up so I missed it.

JacobSzwejbka:
Test failures do not look related

PR #15228, opened 2025-10-17
make aoti_torch_empty_strided support creating incontiguous tensor

Gasoonjia:
* #15282
* #15241
* __->__ #15228

This diff modifies the `aoti_torch_empty_strided` function to support the creation of incontiguous tensors.

To achieve it, this diff:
1. update the way to calculate the memory size by using both tensor size and the stride
2. skip stride check in ETensor by adding and checking cmake macro `USE_CUDA_BACKEND` when building with CUDA backend support.

we will soon bring the ETensor check back for every backend after migrating to use slimtensor.

Differential Revision: [D84938258](https://our.internmc.facebook.com/intern/diff/D84938258/)


backends/cuda/runtime/shims/memory.cpp
```diff
@@ -223,7 +219,29 @@ AOTITorchError aoti_torch_empty_strided(
...
+    if (sizes_ptr[i] == 0) {
+      storage_size = 0;
+      break;
+    }
+    // For each dimension, add stride[i] * (size[i] - 1)
+    // This gives us the maximum offset in that dimension
+    int64_t stride_i = (strides_ptr != nullptr) ? strides_ptr[i] : 0;
+    if (strides_ptr == nullptr) {
+      // Calculate contiguous stride if not provided
+      stride_i = 1;
```
msluszniak:
```suggestion
    int64_t stride_i = (strides_ptr != nullptr) ? strides_ptr[i] : 1;
    if (strides_ptr == nullptr) {
      // Calculate contiguous stride if not provided
```

extension/tensor/tensor_ptr.cpp
```diff
@@ -79,6 +79,11 @@ TensorPtr make_tensor_ptr(
       });
     }
   }
+
+// Skip stride calculation and incontiguous tensor check for CUDA backend since
+// AOTI-CUDA handles both contiguous and incontiguous tensors. This will be
+// removed after SlimTensor migration.
+#ifndef USE_CUDA_BACKEND
```
JacobSzwejbka: 😭

larryliu0820: This is not needed anymore?

Gasoonjia: you are absolutely right. Removed.

larryliu0820 (approved):

PR #15231, opened 2025-10-17
Remove default bos/eos from metadata

lucylq:
Summary:
Also see: https://github.com/pytorch/executorch/pull/15215

Currently:
- default eos/bos tokens are embedded into the pte
- llama3 instruct has a different set of eos/bos tokens
- users must manually specify at export time the llama3 instruct eos/bos tokens, because the runner overrides tokenizer eos/bos with the values in the PTE

This diff:
- removes the defaults
- rely on tokenizer for eos/bos UNLESS the user explicitly specifies in the metadata, in which case use the eos/bos saved in PTE.

Differential Revision: D84942718

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -15,7 +15,6 @@ import json
 import logging
 import re
 import shlex
-from enum import Enum
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


jackzhxng (approved):
Glad to get rid of fairseq

PR #15256, opened 2025-10-19
Enable int16 for op permute

3l1:
Summary: Enable int16 for op permute

Reviewed By: Ninja91

Differential Revision: D84948536

### DIFF
--- a/backends/arm/test/ops/test_permute.py
+++ b/backends/arm/test/ops/test_permute.py
@@ -9,8 +9,11 @@
 from typing import Tuple
 
 import torch
-
-from executorch.backends.arm.test import common
+from executorch.backends.arm.quantizer.arm_quantizer import (
+    get_symmetric_a16w8_quantization_config,
+    TOSAQuantizer,
+)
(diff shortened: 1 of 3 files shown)


backends/arm/test/ops/test_permute.py
```diff
@@ -42,10 +41,10 @@ class SimplePermute(torch.nn.Module):
     def __init__(self, dims: list[int]):
         super().__init__()
 
-        self.permute = Permute(dims=dims)
+        self.dims = dims
 
     def forward(self, x):
-        return self.permute(x)
+        return torch.permute(x, self.dims)
```
digantdesai, line 52: Is this just a cleanup?

digantdesai, line 52: Add int16 test?

digantdesai (approved):

backends/arm/test/ops/test_permute.py
3l1: it is to avoid importing torchvision

PR #15257, opened 2025-10-19
Fix conv2d bias INT48 type hint removed by fusing placeholder nods

3l1:
Summary: Fix conv2d bias INT48 type hint removed by fusing placeholder nods

Reviewed By: Ninja91

Differential Revision: D84948170

### DIFF
--- a/backends/arm/_passes/fuse_equal_placeholders_pass.py
+++ b/backends/arm/_passes/fuse_equal_placeholders_pass.py
@@ -84,6 +84,13 @@ class FuseEqualPlaceholdersPass(ArmPass):
                     common_persistent,
                 )
 
+                # TBD: Find a principled way to merge node.meta across all fused node
+                # For now, i specifically transfer over the TosaSpecialDtype.meta_key() of the rep_node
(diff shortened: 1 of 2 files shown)


backends/arm/_passes/fuse_equal_placeholders_pass.py
```diff
@@ -83,6 +83,13 @@ def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
                     rep_tensor,
                     common_persistent,
                 )
+                
+                # TBD: Find a principled way to merge node.meta across all fused node
```
digantdesai, line 87: Yeah this is a problem for 'non-standard' meta keys.

digantdesai (approved):
Good debugging @3l1.

CC: @per as FYI

backends/arm/_passes/fuse_equal_placeholders_pass.py
digantdesai: Is there a way to write a conv2d test which fails without this change, just to safeguard.

PR #15264, opened 2025-10-20
Use NoThreadPoolGuard in executor_runner if -cpu_threads is set to 1

swolchok:
Doing so removes some overhead and is what I think we should recommend to customers.

### DIFF
--- a/examples/portable/executor_runner/executor_runner.cpp
+++ b/examples/portable/executor_runner/executor_runner.cpp
@@ -41,6 +41,9 @@
 #if defined(ET_USE_THREADPOOL)
 #include <executorch/extension/threadpool/cpuinfo_utils.h>
 #include <executorch/extension/threadpool/threadpool.h>
+#include <executorch/extension/threadpool/threadpool_guard.h>
+
+#include <optional>
 #endif
 
 static uint8_t method_allocator_pool[4 * 1024U * 1024U]; // 4 MB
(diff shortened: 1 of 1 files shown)


swolchok:
* __->__ #15264

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -163,6 +166,11 @@ int main(int argc, char** argv) {
     ::executorch::extension::threadpool::get_threadpool()
         ->_unsafe_reset_threadpool(num_performant_cores);
   }
+  std::optional<::executorch::extension::threadpool::NoThreadPoolGuard>
+      opt_guard;
+  if (cpu_threads == 1) {
+    opt_guard.emplace();
```
digantdesai, line 172: After this, does the perf look same as `ET_USE_THREADPOOL` not defined? Curious if there is anything else...

digantdesai (approved):

examples/portable/executor_runner/executor_runner.cpp
swolchok: It won't look the same, because we still wrap the function in a runtime::FunctionRef so there's an extra layer of function call.

swolchok:
unittest-arm-backend-with-no-fvp and unittest-nxp-neutron both flaky/broken since at least Friday, merging

PR #15265, opened 2025-10-20
Use runtime::FunctionRef in threadpool.h itself

swolchok:
I previously added runtime::FunctionRef and used it in parallel_for to save code size, but didn't carry that through to the Threadpool wrapper. Profiling showed that we were doing heap allocations on each parallel_for invocation, so let's fix that.


swolchok:
* __->__ #15265

digantdesai (approved):

swolchok:
unittest-nxp-neutron has been broken since Friday and unittest-arm-backend-with-no-fvp has been flaky, merging

PR #15267, opened 2025-10-20
Arm backend: Support per-channel in TOSA.RESCALE

oscarandersson8218:
Adds support for per-channel rescaling in TOSA dialect RESCALE op.


Sebastian-Larsson (approved):
Unrelated CI failures

PR #15268, opened 2025-10-20
Arm backend: remove some xpassing xfails

Erik-Lundell:
That are not 16bit-related, vgf related or due to
numerical instability.


Sebastian-Larsson (approved):
Unrelated test fails

PR #15269, opened 2025-10-20
Arm backend: Use reshape instead of view before edge

Erik-Lundell:
The view operator can't handle non-contigious strides, such as the result of an expand. These are normalized after to_edge, but in the transform_for_annotation_pipeline we shouldn't use views for that reason.
Reshape is the equivalent operator that can handle such strides.


Sebastian-Larsson (approved):
Unrelated CI failures

PR #15270, opened 2025-10-20
NXP backend: Improve `view_copy` delegation

MartinPavella:
### Summary
The conversion of `aten.view_copy` to NeutronIR may require the insertion of extra `Transpose` operations, which may not be supported. This PR makes sure the `view_copy` is only delegated if the extra operations are supported too.

### Test plan
Unit tests provided.

cc @robert-kalmar

### DIFF
--- a/backends/nxp/backend/edge_helper.py
+++ b/backends/nxp/backend/edge_helper.py
@@ -136,3 +136,51 @@ def get_quantization_parameters_for(node: Node) -> tuple[Scale, ZeroPoint] | Non
         return None
 
     return node.args[1], node.args[2]  # Scale and zero_point
+
+
+def get_non_qdq_users(node: Node) -> list[Node]:
+    """Return a list of nodes which consume the output of `node`, but Quantize/Dequantize nodes from QDQ clusters are
+     ignored. Meaning, the list of nodes [<user_1>, ..., <user_N>] from the illustration below is returned.
+
+    If the graph does not follow the QDQ pattern, an empty list is returned.
+
+                │
+            ┌───▼────┐
+            │ `node` │
+            └───┬────┘
+           ┌────▼─────┐
+           │ Quantize │
+           └────┬─────┘
+                ├─────── ... ──────┐
+          ┌─────▼──────┐     ┌─────▼──────┐
+          │ Dequantize │ ... │ Dequantize │
+          └─────┬──────┘     └─────┬──────┘
+           ┌────▼─────┐       ┌────▼─────┐
+           │ <user_1> │  ...  │ <user_N> │
+           └────┬─────┘       └────┬─────┘
+
+    """
+
+    quant_nodes = list(node.users)
+    if len(quant_nodes) != 1 or quant_nodes[0].target not in [
+        exir_ops.edge.quantized_decomposed.quantize_per_tensor.default,
+        exir_ops.edge.quantized_decomposed.quantize_per_channel.default,
+    ]:
+        return []
+
+    dequant_nodes = list(quant_nodes[0].users)
+    if any(
+        dequant_node.target
+        not in [
+            exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
+            exir_ops.edge.quantized_decomposed.dequantize_per_channel.default,
+        ]
+        for dequant_node in dequant_nodes
+    ):
+        return []
+
+    res = []
+    for dequant_node in dequant_nodes:
+        res.extend(list(dequant_node.users))
+
+    return res
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
@@ -6,12 +6,18 @@
 import numpy as np
 
 from executorch.backends.nxp.backend.edge_helper import (
+    get_non_qdq_users,
     input_tensor,
     output_tensor,
     tensor_rank,
 )
 from executorch.backends.nxp.backend.ir.converter import quantization_utils
 from executorch.backends.nxp.backend.ir.converter.conversion.common import OpsList
+from executorch.backends.nxp.backend.ir.converter.conversion.translator import (
+    apply_permutation_to,
+    create_channels_first_to_channels_last_permutation,
+    create_channels_last_to_channels_first_permutation,
+)
 from executorch.backends.nxp.backend.ir.converter.node_converter import (
     CustomDelegationOptions,
     is_not_qdq_node,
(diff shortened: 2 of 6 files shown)


backends/nxp/backend/edge_helper.py
```diff
@@ -125,3 +125,21 @@ def previous_non_qdq_node(node: Node, input_index: int = 0) -> Node | None:
             current_node = current_node.args[0]
         else:
             return current_node
+
+
+def get_non_qdq_users(node: Node) -> list[Node]:
```
robert-kalmar, line 141: I don't understand the logic here. You are looking the list of node uses of users (i.e. not direct users but list of users of the first users node)?
Why?

MartinPavella: The function returns a list of nodes which use `node` as an input, while ignoring the Q/Dq nodes. I.e. the function returns [<user_1>, ..., <user_N>] from the graphic:
```
                   │
               ┌───▼────┐
               │ `node` │
               └───┬────┘
              ┌────▼─────┐
              │ Quantize │
              └────┬─────┘
                   ├─────── ... ──────┐
             ┌─────▼──────┐     ┌─────▼──────┐
             │ Dequantize │ ... │ Dequantize │
             └─────┬──────┘     └─────┬──────┘
              ┌────▼─────┐       ┌────▼─────┐
              │ <user_1> │  ...  │ <user_N> │
              └────┬─────┘       └────┬─────┘
```

Thanks to your comment I found a mistake in the implementation. I have fixed the mistake as well as improved the documentation for clarity.

backends/nxp/backend/edge_helper.py
```diff
@@ -125,3 +125,51 @@ def previous_non_qdq_node(node: Node, input_index: int = 0) -> Node | None:
...
+        dequant_node.target
+        not in [
+            exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
+            exir_ops.edge.quantized_decomposed.dequantize_per_channel.default,
+        ]
+        for dequant_node in dequant_nodes
+    ):
+        return []
+
+    res = []
```
roman-janik-nxp, line 182: Why not simply use list comprehension?
```suggestion
    res = [user for dequant_node in dequant_nodes for user in dequant_node.users]
```

backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
```diff
@@ -66,6 +79,72 @@ def supports_partitioning_result(
...
+                # The `Transpose` would have to be removed by the `PermuteFullyConnectedWeightsAfterReshape` pass.
+                # Make sure it will be applied.
+                users = get_non_qdq_users(node)
+                if len(users) != 1 or (linear_node := users[0]).target not in [
+                    exir_ops.edge.aten.addmm.default,
+                    exir_ops.edge.aten.mm.default,
+                ]:
+                    return False
+
+                if linear_node not in view_copy_partitions[0].nodes:
```
roman-janik-nxp, line 114: Could you please explain me more this condition? Transposition to channel first is not supported on Neutron and the user of `view_copy` is `addmm` or `mm` linear node. The linear node is however outside of current partition and it has to be together to be delegatable. Is that right?

backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
```diff
@@ -66,6 +79,72 @@ def supports_partitioning_result(
...
+            # Make sure these transpositions are supported.
+            if not (
+                transposition_is_supported_on_neutron(
+                    input_shape, to_nchw_perm, neutron_target_spec
+                )
+                and transposition_is_supported_on_neutron(
+                    output_shape, to_nhwc_perm, neutron_target_spec
+                )
+            ):
+                return False
```
roman-janik-nxp, line 151: Why is this needed? Both input an output formats are channel first, so why check for transpositions? There shouldn't be any, or?

backends/nxp/tests/ir/converter/node_converter/test_view_copy_converter.py
```diff
@@ -32,9 +31,11 @@
 )
 from executorch.backends.nxp.tests.executors import (
     convert_run_compare,
+    graph_contains_any_of_ops,
     ToNCHWPreprocess,
```
roman-janik-nxp: Please refactor it to `ToChannelFirstPreprocess` and `ToChannelLastPreprocess` while you are at changing the tests.

backends/nxp/tests/ir/converter/node_converter/test_view_copy_converter.py
```diff
@@ -266,3 +295,207 @@ def test_view_w_conv_linear_quant_conversion(mocker, input_shape, channels_view_
         tfl_model=tflite_flatbuffers_model,
         atol=1.0,
     )
+
+
+@pytest.mark.parametrize(
+    "bias",
+    [True, False],
+)
+def test__context_dependent__channels_first_to_formatless__transpose_fused(
```
roman-janik-nxp: I would add view_copy to the test names as it is common for all other node converter tests.

roman-janik-nxp (commented):
Please check that our internal integration test plan is passing too.

robert-kalmar (approved):

backends/nxp/backend/edge_helper.py
MartinPavella: Because it doesn't really matter, and I'd argue that the list comprehension is a bit less clear.

backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
MartinPavella: That's right.
As mentioned by the comment on line `100`, the `view_copy` can be delegated only if the pass `PermuteFullyConnectedWeightsAfterReshape` will remove the inserted `Transpose` operator. And this pass will only be applied if the `Transpose`, `Reshape` and `FullyConnected` (`mm`/`addmm`) are in the same Neutron IR model (same delegated partition).

backends/nxp/backend/ir/converter/node_converters/ops_converters/view_copy_converter.py
MartinPavella: If the input and output are channels first, they will both be channels last during conversion. In this case, a generic solution is to insert a `Transpose` before and after, to make the input and output of the `Reshape` channels first.

An illustrative example:

ExecuTorch:
[1,2,3,4] -> view_copy -> [2,3,1,4]

Neutron IR:
Naive conversion:
[1,3,4,2] -> Reshape -> [2,1,4,3]    (the data of the output will be shuffled, and not equal to an NHWC version of the ExecuTorch output)
Correct conversion:
[1,3,4,2] -> Transpose -> [1,2,3,4] -> Reshape -> [2,3,1,4] -> Transpose -> [2,1,4,3]    (the output is an NHWC equivalent of the ExecuTorch output)

MartinPavella:
> Please check that our internal integration test plan is passing too.

It is passing: https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH23-9

PR #15276, opened 2025-10-20
Arm backend: Improve error logging when running Corstone FVP

zingo:
Instead of putting the whole FVP runtime command and log in the runtime error text only a condensed error is used, the command and the full log is outputed in the log instead.

As we now always log the command and log, you can now look and compare FVP output logs also on passed test by running run pytests with "-rP".


Sebastian-Larsson (approved):
Unrelated CI failures.

PR #15277, opened 2025-10-20
update backend cadence md for branch cut

zonglinpeng:
Summary: ~

Differential Revision: D85064213

### DIFF
--- a/docs/source/archive/backends-cadence-legacy.md
+++ b/docs/source/archive/backends-cadence-legacy.md
@@ -1,12 +1,19 @@
+# Cadence Xtensa Backend (Legacy / Outdated)
+
+```{warning}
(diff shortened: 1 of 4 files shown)


docs/source/archive/backends-cadence-legacy.md
```diff
@@ -0,0 +1,238 @@
+# Cadence Xtensa Backend (Legacy / Outdated)
```
mergennachin, line 1: Can just delete this page since we can trust git version control?

mcremon-meta: We can, I think keeping it around is still useful in case people rely on it or we want to make it a first class citizen again

mcremon-meta (approved):

PR #15278, opened 2025-10-20
Arm backend: ArmTester support testing with portable ops

zingo:
### Summary
This will enable model testing where some ops are not fully delegated,
like done in the backend flow test suite.

This also remove the unused fast_fvp pytest flag and always run the
Corstone-300/320 in fast mode as we are never interesting in PMU cycle
numbers in the pytesting anyway, this gives more then a 2x speedboost
on some models.

This will also extend the FVP timeout to 20 min for backend flows and
increase allocation pool to 80MB to make it possible to tests larger
models.

For normal unit tests the timeout is set to 10 min

### Test plan
Arm backend suite test will tests this.


oscarandersson8218 (approved):

zingo:
Failed tests are unrelated

PR #15279, opened 2025-10-20
Runner logs

lucylq:
### Summary
Move some runner logs around.
1. Move the stats so it doesn't interfere with generation
2. Add a different message to end of generation to indicate we cut off because tokens, not because of eos

### Test plan
lora demo

### DIFF
--- a/extension/llm/runner/text_token_generator.h
+++ b/extension/llm/runner/text_token_generator.h
@@ -128,9 +128,13 @@ class ET_EXPERIMENTAL TextTokenGenerator {
       if (eos_ids_->find(cur_token) != eos_ids_->end()) {
         printf("\n");
         ET_LOG(Info, "\nReached to the end of generation");
-        break;
+        return pos - start_pos;
       }
     }
+    ET_LOG(
+        Info,
+        "\nFinished generation. Generated %" PRIi32 " tokens.",
+        start_pos + max_new_tokens);
     return pos - start_pos;
   }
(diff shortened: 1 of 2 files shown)


extension/llm/runner/text_token_generator.h
```diff
@@ -128,9 +128,13 @@ class ET_EXPERIMENTAL TextTokenGenerator {
       if (eos_ids_->find(cur_token) != eos_ids_->end()) {
         printf("\n");
```
JacobSzwejbka, line 129: not your change but weird we have a printf

extension/llm/runner/text_token_generator.h
```diff
@@ -128,9 +128,13 @@ class ET_EXPERIMENTAL TextTokenGenerator {
...
         ET_LOG(Info, "\nReached to the end of generation");
-        break;
+        return pos - start_pos;
```
JacobSzwejbka, line 131: can you put this in a variable and share it

JacobSzwejbka: oh its in the loop. Nvm I guess

JacobSzwejbka (approved):

extension/llm/runner/text_token_generator.h
lucylq: Leaving as-is, because one is inside a while loop and another is not.

PR #15286, opened 2025-10-20
Fix text_llm_runner kv cache pos count and use it for generate()

kirklandsign:
### Summary
pos_ should advance by prefill and generated prompt size.
### Test plan
CI

### DIFF
--- a/extension/llm/runner/text_llm_runner.cpp
+++ b/extension/llm/runner/text_llm_runner.cpp
@@ -199,6 +199,8 @@ Error TextLLMRunner::generate(
   }
   int64_t num_generated_tokens = generate_result.get();
 
+  pos_ += num_generated_tokens;
+
   stats_->inference_end_ms = time_in_ms();
   if (!config.warming) {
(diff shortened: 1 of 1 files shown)


larryliu0820 (approved):

extension/llm/runner/text_llm_runner.cpp
```diff
@@ -199,6 +199,8 @@ Error TextLLMRunner::generate(
   }
   int64_t num_generated_tokens = generate_result.get();
 
+  pos_ += num_generated_tokens;
```
larryliu0820, line 202: you don't need this anymore right? `pos_` is being updated.

kirklandsign: No :(

https://github.com/pytorch/executorch/blob/main/extension/llm/runner/text_token_generator.h#L53

https://github.com/pytorch/executorch/blob/main/extension/llm/runner/multimodal_runner.cpp#L180-L190

PR #15289, opened 2025-10-21
fix: alert users that darwin is not supported

Mistobaan:
### Summary

Add a note for the users that try to follow the tutorials on Apple/Darwin systems


cccclai:
Thanks for contributing! We can merge after CI passes

jackzhxng (approved):

PR #15298, opened 2025-10-21
Arm backend: Merge passes that replace scalars

martinlsm:
Besides having obsolete names, ReplaceScalarWithTensorPassTOSAMI and ReplaceScalarWithTensorPassTOSABI was causing difficulties for defining chronological pass dependencies with the `_passes_required_after` attribute. This is because, with the current design, there is no way to distinguish which profile is referred to when defining `_passes_required_after`; a pass simply declares its chronological dependencies globally.

Solve this by merging the two pass classes together into one called `ReplaceScalarWithTensorByProfilePass`. This means that a pass should include this new pass in `_passes_required_after` no matter which TOSA profile its working towards.

### Test plan
This is a refactoring patch. No behavior is modified. Thus the tests already in place are enough to test this change.


oscarandersson8218 (approved):
Unrelated failures

PR #15299, opened 2025-10-21
Arm backend: Move rescales from SUM visitor to pass

martinlsm:
In the SUM node visitor, an INT8->INT32 RESCALE node is inserted prior to the SUM node; similarly, an INT32->INT8 RESCALE node is inserted after.

This patch moves the insertion to `InsertRescaleInt32Pass`. Since SUM is decomposed, insertion of RESCALE nodes should be carried out before `DecomposeSumPass` (which decomposes SUM into a chain of single dim SUMs). The ordering is important to avoid redundant INT8/INT32 RESCALE nodes being inserted between each SUM node in the chain after decomposition. Only one INT8->INT32 RESCALE is needed before the chain, and an INT32->INT8 after it; between the SUM nodes in the chain, the edges are already in the correct INT32 data type.

### Test plan
Tests exercising the modified pass of this patch have been added to backends/arm/test/passes/test_insert_rescale_i32_pass.py.

### DIFF
--- a/backends/arm/_passes/arm_pass_manager.py
+++ b/backends/arm/_passes/arm_pass_manager.py
@@ -194,7 +194,6 @@ class ArmPassManager(PassManager):
         self.add_pass(ConvertExpandCopyToRepeatPass())
         self.add_pass(UnsqueezeBeforeRepeatPass())
         self.add_pass(CastInt64BuffersToInt32Pass(exported_program))
-        self.add_pass(DecomposeSumPass())
         self.add_pass(DecomposeCumsumPass(exported_program))
         self.add_pass(Conv1dUnsqueezePass())
         self.add_pass(DecomposeMaxPool2DPass())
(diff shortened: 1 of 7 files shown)


Sebastian-Larsson (approved):
Unrelated CI failures

Sebastian-Larsson (changes requested):
This is failing 
[trunk / test-arm-backend (test_pytest_ops_ethosu_fvp) / linux-job](https://hud.pytorch.org/pr/pytorch/executorch/15299#53477949080) ([gh](https://github.com/pytorch/executorch/actions/runs/18747420773/job/53477949080))

martinlsm:
> This is failing [trunk / test-arm-backend (test_pytest_ops_ethosu_fvp) / linux-job](https://hud.pytorch.org/pr/pytorch/executorch/15299#53477949080) ([gh](https://github.com/pytorch/executorch/actions/runs/18747420773/job/53477949080))

This must be a flaky test. It has not been seen in our internal CI nor have I managed to reproduce it locally by looping this test.

zingo:
It will probably work in that case if you retrigget the job?

Sebastian-Larsson (approved):

PR #15301, opened 2025-10-21
Arm backend: Remove `# type: ignore` from tosa_serializer import

Sebastian-Larsson:
A leftover from TOSA 0.80 was the need for `# type: ignore` after the tosa serializer import. This ignore is now removed, along with moving the imports to the top of the file instead of having it as a dynamic import.


oscarandersson8218 (approved):
CI failures seem unrelated

PR #15305, opened 2025-10-21
[ET-VK] Add redirect for backends-vulkan

SS-JIA:
Summary:
Title says it all!

cc @manuelcandales @digantdesai @cbilgin


docs/source/backends/vulkan/vulkan-op-support-table.csv
```diff
@@ -0,0 +1,113 @@
+Namespace,Operator,Notes
```
SS-JIA, line 1: due to .gitignore settings this file was accidentally left out of the original commit updating the documentation. Decided to just bundle it with this fix.

mergennachin (approved):

Issue #14791, opened 2025-10-03
Fix Quick Start


psiddh:
### 📚 The doc issue

- Flow chart for the User 
  - Add pip installation steps,
  - Export flow
- etc
https://docs-preview.pytorch.org/pytorch/executorch/14720/index.html

Issue #14759, opened 2025-10-02
Update Building from Source Docs


GregoryComer:
### 📚 The doc issue

Our current building from source docs can use some improvements to be clearer and flow better. Derrick Johnson made the following suggestions, which I would like to make for 1.0.

* Clarify which systems provide prebuilt pip packages and which require building from source.
* Move Windows-specific env setup to near the rest of the env setup info.
* Call out git submodule sync and update/init.
* Move ccache info to the bottom as to not disrupt the main flow.
* Use -S and -B args to cmake invocations instead of changing directory
* Move CMake option descriptions up near where we tell people to do cmake configuration.

GregoryComer:
This is partially complete, `Use -S and -B args to cmake invocations instead of changing directory` still needs some work.

XAheli:
Hi @GregoryComer would like to work on this :)

GregoryComer:
Sure, thanks! I've assigned the task to you. Feel free to reach out if you have any questions.

XAheli:
Hi @GregoryComer ! Does the pr #19392 closes this issue? If not please lmk any other points I should address! Thanks :)

GregoryComer:
Closing as done. Thanks for the contribution!

PR #15311, opened 2025-10-21
Update build from source and getting started docs

GregoryComer:
### Summary

Update Getting Started and Build from Source Docs:
 * Integrate Windows steps into the main flow with minor Windows-specific callouts.
 * Clarify top-level flow for building from source - add a table by use case.
 * Clarify building ET as a submodule vs standalone build.
 * Re-order, re-word, and clean up the content related to building from source.
 * Add info on NDK build for Android.

Tracked in https://github.com/pytorch/executorch/issues/14791 and https://github.com/pytorch/executorch/issues/14759.

cc @mergennachin @AlannaBurke @byjlw

### DIFF
--- a/docs/source/using-executorch-building-from-source.md
+++ b/docs/source/using-executorch-building-from-source.md
@@ -17,8 +17,8 @@ ExecuTorch is tested on the following systems, although it should also work in s
  * macOS (x86_64/ARM64)
     * Big Sur (11.0)+
  * Windows (x86_64)
+    * Windows 10+ with Visual Studio 2022+ and [Clang-CL](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170)
     * Windows Subsystem for Linux (WSL) with any of the Linux options
-    * Windows 10+ with Visual Studio 2022+ (experimental)
 
 ### Software Requirements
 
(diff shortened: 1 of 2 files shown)


docs/source/using-executorch-building-from-source.md
```diff
@@ -17,8 +17,8 @@ ExecuTorch is tested on the following systems, although it should also work in s
  * macOS (x86_64/ARM64)
     * Big Sur (11.0)+
  * Windows (x86_64)
-    * Windows Subsystem for Linux (WSL) with any of the Linux options
     * Windows 10+ with Visual Studio 2022+ (experimental)
+    * Windows Subsystem for Linux (WSL) with any of the Linux options
```
mergennachin, line 21: Is "Windows 10+ with Visual Studio 2022+" still considered experimental?

GregoryComer: I'm okay to drop the experimental wording. My main concern is that support is still very new and they might hit some surprises. However, we do have CI coverage for the core runtime, XNNPACK, and key extensions (data loader, module, etc.). Would you recommend that I drop the experimental wording?

mergennachin (approved):
I see that you removed but I recommend adding back a brief "Verify Your Build" section after the Building section that contains 

  - One-line command to export a simple model
  - One-line command to run it with the executor_runner

docs/source/using-executorch-building-from-source.md
mergennachin: Yeah, let's drop experimental and explicitly say "ClangCL"? Otherwise, people may use with MSVC and expect it work

GregoryComer:
> I see that you removed but I recommend adding back a brief "Verify Your Build" section after the Building section that contains
> 
> * One-line command to export a simple model
> * One-line command to run it with the executor_runner

Sounds good. I'll re-add. The reason I removed it was partly because we don't actually build the executor_runner for most presets and it requires multiple CLI options to include. I'm a little conflicted on this, but maybe we should build by default?

I think it does make sense to have a quick verification step. I'll re-add.

docs/source/using-executorch-building-from-source.md
GregoryComer: Updated.

GregoryComer:
> > I see that you removed but I recommend adding back a brief "Verify Your Build" section after the Building section that contains
> > 
> > * One-line command to export a simple model
> > * One-line command to run it with the executor_runner
> 
> Sounds good. I'll re-add. The reason I removed it was partly because we don't actually build the executor_runner for most presets and it requires multiple CLI options to include. I'm a little conflicted on this, but maybe we should build by default?
> 
> I think it does make sense to have a quick verification step. I'll re-add.

Added.

PR #15327, opened 2025-10-21
Properly record load time

larryliu0820:
In multimodal_runner we should record load time whenever we actually run load on modules.


JacobSzwejbka (approved):

larryliu0820:
The change is merged.

PR #15336, opened 2025-10-21
Minor doc fixes

GregoryComer:
### Summary
Fix heading level for build verification, cleanup wording on Windows reqs, update android presets since they were split by arch.


docs/source/getting-started.md
```diff
@@ -12,7 +12,7 @@ The following are required to install the ExecuTorch host libraries, needed to e
 - g++ version 7 or higher, clang++ version 5 or higher, or another C++17-compatible toolchain.
 - Linux (x86_64 or ARM64), macOS (ARM64), or Windows (x86_64).
     - Intel-based macOS systems require building PyTorch from source (see [Building From Source](using-executorch-building-from-source.md) for instructions).
-- On Windows, Visual Studio 2022 or later. Clang build tools are needed to build from source.
+- On Windows, Visual Studio 2022 or later.
```
GregoryComer, line 15: Removing the build from source callout as it's covered in the building from source section. We do still need VS 2022 because torch.export does C++ compilation (via MSVC) in strict mode.

kirklandsign (approved):

psiddh (approved):

PR #15338, opened 2025-10-22
Bulk cherry-pick doc updates

GregoryComer:
### Summary
Bulk pick doc updates using https://github.com/pytorch/executorch/blob/main/scripts/pick_doc_commits.py.

I also manually included https://github.com/pytorch/executorch/commit/eb05056623d944f0351c7c339b96261e78472e2d, which has some example changes, since it's a doc update PR.

I sanity checked this by running `git diff HEAD upstream/main -- docs/source` and the diff didn't show any surprises. There were some changes on main related to larger PRs that haven't been picked (NXP visualizer, for example), but everything else is in sync.


psiddh (approved):
Lets Go!

PR #15340, opened 2025-10-22
add CI configs for msvc build

JacobSzwejbka:


setup.py
```diff
@@ -751,9 +751,14 @@ def run(self):  # noqa C901
             f"-DCMAKE_BUILD_TYPE={cmake_build_type}",
         ]
 
-        # Use ClangCL on Windows.
+        # Use ClangCL on Windows by default, unless EXECUTORCH_USE_MSVC is set.
```
JacobSzwejbka: Oh these changes arent needed anymore. They are also the wrong idea

PR #15341, opened 2025-10-22
Metal backend: store weights outside .so

manuelcandales:
This pull request refactors how model weight constants are packaged and loaded in the Metal backend. The main change is separating weight constants from the shared object (`.so`) file, storing them as a binary blob on disk, and updating the runtime to load these weights from the blob. This speeds up loading time.

### DIFF
--- a/backends/apple/metal/metal_backend.py
+++ b/backends/apple/metal/metal_backend.py
@@ -108,8 +108,11 @@ class MetalBackend(BackendDetails):
         options: dict[str, typing.Any] = {
             # Do not link against the full PyTorch/libtorch library
             "aot_inductor.link_libtorch": False,
-            # Package model constants and other generated files directly in the shared object (.so) file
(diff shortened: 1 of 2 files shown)


manuelcandales:
* #15346
* #15343
* #15342
* __->__ #15341

backends/apple/metal/metal_backend.py
```diff
@@ -108,34 +108,62 @@ def preprocess(
         options: dict[str, typing.Any] = {
```
larryliu0820, line 108: Is there a reason why we need metal_backend.py? It feels very much like cuda_backend.py

manuelcandales: Yes, I agree. Next week we should begin refactoring, and moving most of this code to a new aoti_backend.py. Same with most of the code in metal_backend.cpp/cuda_backend.cpp

larryliu0820 (approved):

PR #15350, opened 2025-10-22
clean unnecessary CMAKE_ARGS="-DEXECUTORCH_BUILD_CUDA=ON"

Gasoonjia:
Summary: we no longer need this arg for et installation. clean it to avoid misleading.

Differential Revision: D85223111


larryliu0820 (approved):

PR #15356, opened 2025-10-22
Arm Backend: Enable Pybindings for tosa_serialization lib

ArmRyan:
These bindings should drastically increase serialization
 performance, both speed wise and memory usage wise.

 This patch also enables serialization of tosa flatbuffers
 that are over 2GB in size.

 - Ensure each tosa operator has an attribute.
 - Replace TosaOp.Op() calls with Op enum
 - Convert tosa clamp data to numpy int8 to allow c++ byte serialization
 - Replace "import tosa_serializer.serializer" with "import tosa_serializer" for pybind
 - Remove serialize to json.
 - Remove Darwin support workaround in mlsdk dependencies
 - Update mlsdk to latest tag, but overwrite the tosa_mlir_translator to a version that supports offset buffers
 - Update tosa-tools to the monorepo
 - Update slice to clamp end and start dim. This avoids the object not having a number field from being too large or small.

Change-Id: Id03582b6b7b9eef296480f9ea3a4744a08b22603

### DIFF
--- a/examples/arm/setup.sh
+++ b/examples/arm/setup.sh
@@ -35,6 +35,9 @@ toolchain_url=""
 toolchain_dir=""
 toolchain_md5_checksum=""
 
+# Load logging helpers early so option parsing can emit status messages.
+source "$et_dir/backends/arm/scripts/utils.sh"
+
 
 # List of supported options and their descriptions
 OPTION_LIST=(
@@ -145,7 +148,7 @@ function check_options() {
                 shift
                 ;;
             --setup-test-dependency)
-                echo "Installing test dependency..."
+                log_step "deps" "Installing test dependency..."
                 source $et_dir/backends/arm/scripts/install_models_for_test.sh
                 exit 0
                 ;;
@@ -218,7 +221,6 @@ if [[ $is_script_sourced -eq 0 ]]; then
     check_options "$@"
 
     # Import utils
-    source $et_dir/backends/arm/scripts/utils.sh
     source $et_dir/backends/arm/scripts/fvp_utils.sh
     source $et_dir/backends/arm/scripts/toolchain_utils.sh
     source $et_dir/backends/arm/scripts/vulkan_utils.sh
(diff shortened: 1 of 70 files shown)


mansnils (approved):

ArmRyan:
failures unrelated

robell (approved):
Internal review completed so we're happy with this.

digantdesai:
This was a huge change (esp in terms of tosa serializer changes), it is breaking internal CI. Let me try to fwd fix. In the future if we are touching to many files, we should also pull internally and test (at least until we improve OSS testing for buck :().

examples/arm/setup.sh
```diff
@@ -277,11 +279,39 @@ if [[ $is_script_sourced -eq 0 ]]; then
...
+        pip install --no-dependencies -r "$et_dir/backends/arm/requirements-arm-tosa.txt"
+
+    pushd "$root_dir"
+    if [[ ! -d "tosa-tools" ]]; then
+        git clone https://git.gitlab.arm.com/tosa/tosa-tools.git
+    fi
+
+    pushd tosa-tools
+    git checkout 8468d041c50c6d806f3c1c18c66d7ef641e46580 # serialization lib pybindings
+    if [[ ! -d "reference_model" ]]; then
```
digantdesai, line 294: I guess `tosa/tosa-reference-model` repo on gitlab.arm.com is dead (but its there) and long live `tosa/tosa-tools`?

examples/arm/setup.sh
```diff
@@ -277,11 +279,39 @@ if [[ $is_script_sourced -eq 0 ]]; then
...
+    CMAKE_POLICY_VERSION_MINIMUM=3.5 \
+        pip install --no-dependencies -r "$et_dir/backends/arm/requirements-arm-tosa.txt"
+
+    pushd "$root_dir"
+    if [[ ! -d "tosa-tools" ]]; then
+        git clone https://git.gitlab.arm.com/tosa/tosa-tools.git
+    fi
+
+    pushd tosa-tools
+    git checkout 8468d041c50c6d806f3c1c18c66d7ef641e46580 # serialization lib pybindings
```
digantdesai, line 293: Any impact on BC esp on Vela 4.4.[12]? I see the CI is OK so should be low risk.

ArmRyan: Should be fine, tosa is supposed to be backwards compatible so unless we use new tosa features ( we dont atm ) then nothing should change for vela. The main changes are:
Pybind11 instead of pure python backend, so better perf when lowering. eg. Llama 3.2 1b in int8 goes from ~1h lowering on my laptop to like 3 minutes, and ram usage goes from over 30gb to ~9.
Support for >2GB flatbuffers, these modify the flatbuffer layout if it is >2GB, but there is a guard around ethos-u backend to try and block those from entering vela until its supported there

examples/arm/setup.sh
ArmRyan: Yeah individual repos probably wont change anymore, but should stay online. tosa-tools is a monorepo and is the future of the tooling as far as I'm aware

ArmRyan:
> This was a huge change (esp in terms of tosa serializer changes), it is breaking internal CI. Let me try to fwd fix. In the future if we are touching to many files, we should also pull internally and test (at least until we improve OSS testing for buck :().

It *should* just be as simple as a setup.sh run, front-end usage should be identical, its purely the internals of the backend that are api broken, but they shouldnt really be exposed so as long as the library exists there should be no change

examples/arm/setup.sh
digantdesai: re. the tosa-serializer, IIUC we have both c++ (through pybind) and direct python APIs, i.e. `tosa_serializer` vs. `_tosa_serializer` why do we have two? Any differences? Are you going to deprecate the direct python ones over pybind ones for perf?

ArmRyan: I think its mostly just a quick of the reference model atm, when you pip install reference model it installs old python api ( import serializer.tosa_serializer ), but if you pip install serializer you get c++ ( import tosa_serializer ). Im not sure what the tosa teams plans are, but i think the pure python api is likely going to be frozen where it is, it already lags behind the c++ anyway. For us in ET we have no plans to use the old python version anymore. 

Theres not really any difference that I can see, the graphs should serialize to the same flatbuffer and have the same graph, I dont think either one has more or less features right now, but that could change

digantdesai: Hmm I am having difficulty in buckify the C++ variant, and since this is breaking things internally, let me revert this for now, we can try to land this again later this week.

robell: it will likely impact quite a few follow on fixes which depend on this change if we have to revert this (and so a revert chain) - is there anything we can do to help get this through?

digantdesai: The problem is, because of the travel etc. this is blocking a bunch of changes internally since we can't land this due to failures. Do you know which PRs are depending on this? Besides the import name change, this PR (on its own) seems quite independent.

ArmRyan: I think its mostly the large revert chain upstream and internally that we were hoping to avoid

PR #15363, opened 2025-10-23
Arm backend: support mean.default

Erik-Lundell:


Sebastian-Larsson (approved):
Unrelated CI failures.

PR #15364, opened 2025-10-23
Arm backend: Tag control flow submodules in partitioner

Erik-Lundell:
Control flow operators are implemented with get_attr/ placeholders nodes pointing to submodules in the graph_module. To know if we can handle such an operators, we need to figure out whether we can partition all of the submodule the operator points to. We can do this simply by applying the partition logic we already have to submodules. This has the added benefit that we will partition parts of submodules into normal partitions, even if we can't partition the full submodule.


Sebastian-Larsson (approved):
Unrelated CI failures

PR #15366, opened 2025-10-23
Add check for negative index in cpu kernel op_index

Sebastian-Larsson:
Fast path cannot be taken when the index is negative in op_index. It results in a failure in `check_fast_path_args`. Add a check to `check_fast_path_conditions` for negative index and return `false` if one is encountered.

### DIFF
--- a/kernels/portable/cpu/op_index.cpp
+++ b/kernels/portable/cpu/op_index.cpp
@@ -1,11 +1,13 @@
 /*
  * Copyright (c) Meta Platforms, Inc. and affiliates.
  * All rights reserved.
+ * Copyright 2025 Arm Limited and/or its affiliates.
  *
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
 
+#include <algorithm>
 #include <cinttypes>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


kernels/portable/cpu/op_index.cpp
```diff
@@ -47,6 +48,23 @@ bool check_fast_path_conditions(
...
       }
+
+      // Fast path only supports non-negative indices.
+      if (ix_type == ScalarType::Int) {
+        const int32_t* const data = index.const_data_ptr<int32_t>();
+        for (const auto j : c10::irange(index.numel())) {
+          if (data[j] < 0) {
+            return false;
+          }
+        }
```
swolchok: you may wish to consider writing this as

```
if (std::any_of(data, data + index.numel(), [](const auto x) { return x < 0; })) {
  return false;
}
```

up to your personal judgement and taste though.

kernels/portable/cpu/op_index.cpp
```diff
@@ -47,6 +48,23 @@ bool check_fast_path_conditions(
...
+      } else { // ScalarType::Long
+        const int64_t* const data = index.const_data_ptr<int64_t>();
```
swolchok, line 62: ditto

swolchok (changes requested):
looks fine; inline comments are for your consideration and not blocking. However, please also add a test for this behavior (and feel free to @ me for re-review to avoid extra delay).

kernels/portable/cpu/op_index.cpp
Sebastian-Larsson: Sounds good :)

kernels/portable/cpu/op_index.cpp
Sebastian-Larsson: Done!

Sebastian-Larsson:
Changed to your suggestions and added tests @swolchok

kernels/test/op_index_test.cpp
```diff
@@ -480,6 +481,36 @@ TEST_F(OpIndexTensorOutTest, AllDtypesSupportedForIndex) {
...
+  Tensor expected = tf.make({1}, {3.});
+
+  std::array<optional<Tensor>, 1> indices = {
+      optional<Tensor>(tfl.make({1}, {-1}))};
+
+  Tensor ret = op_index_tensor_out(x, indices, out);
+  EXPECT_TENSOR_EQ(ret, expected);
+}
+
+TEST_F(OpIndexTensorOutTest, NegativeIndexSupportedForInt) {
```
swolchok, line 499: I would prefer to see a templated test, but it's probably not worth blocking on

swolchok (approved):

swolchok:
looks like we have a bunch of infra failures and known flakes in CI. merging.

PR #15368, opened 2025-10-23
Add bundled IO support to portable executor runner

mansnils:
Executor runner supports both models with/wo bundled io in same path.
To enable bundled IO EXECUTORCH_BUILD_DEVTOOLS and
EXECUTORCH_ENABLE_BUNDLE_IO are required.

Adds tests in Arm backend for testing this/depending on this.
Except for enabling bundle-io for VGF backend where applicable,
some additional resnets model tests are enabled as well.

Avoids narrowing conversion errors in pte_to_header script by switching
char to unsigned char.


mansnils:
Need to either add DEVTOOLS=ON to all cmake recipes that builds the executor runner or use an optional build flag.

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -460,5 +560,52 @@ int main(int argc, char** argv) {
     ET_CHECK_MSG(status == Error::Ok, "Failed to save ETDump file.");
   }
 
+  if (bundle_io) {
+    // With bundled io we can check the result.
```
JacobSzwejbka, line 574: This file gets copy pasted a ton as a basic example of a runner. Im pretty resistant to something like this which will tempt people to use it it as a tester. Im already not a huge fan of having the benchmarking stuff in here.

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -92,6 +105,17 @@ using executorch::runtime::Span;
 using executorch::runtime::Tag;
 using executorch::runtime::TensorInfo;
 
+#if defined(ET_BUNDLE_IO_ATOL)
```
JacobSzwejbka: just use gflags to pass these as an arg

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -172,89 +213,138 @@ int main(int argc, char** argv) {
...
+  // Inputs can come from bundleio, from ptd file, as optional input file(s), or
+  // everything hardcoded to ones.
   std::vector<std::string> inputs_storage;
   std::vector<std::pair<char*, size_t>> input_buffers;
+  std::unique_ptr<FileDataLoader> ptd_loader;
+  std::unique_ptr<FlatTensorDataMap> ptd_data_map;
 
-  std::stringstream list_of_input_files(FLAGS_inputs);
-  std::string path;
+  if (!bundle_io) {
```
JacobSzwejbka, line 250: are .ptd and bundledProgram mutually exclusive? @lucylq  @Gasoonjia.

ptd is only used to pass weights along its not another way to pass inputs to this file afaik

JacobSzwejbka (changes requested):

mansnils:
> Need to either add DEVTOOLS=ON to all cmake recipes that builds the executor runner or use an optional build flag.

This comment should be resolved now. 
Apologies for force push after your comments @JacobSzwejbka

examples/portable/executor_runner/executor_runner.cpp
mansnils: This is now an optional build flag.

examples/portable/executor_runner/executor_runner.cpp
mansnils: Will do

examples/portable/executor_runner/executor_runner.cpp
```diff
@@ -347,7 +452,6 @@ int main(int argc, char** argv) {
   et_timestamp_t time_spent_executing = 0;
   // Run the model.
   for (uint32_t i = 0; i < FLAGS_num_executions; i++) {
-    ET_LOG(Debug, "Preparing inputs.");
     // Allocate input tensors and set all of their elements to 1 or to the
     // contents of input_buffers if available. The `inputs`
     // variable owns the allocated memory and must live past the last call to
```
mansnils, line 445: TODO: problem here that inputs loose lifetime which is likely why some test fails

mansnils: Done

examples/portable/executor_runner/executor_runner.cpp
mansnils: Done

examples/portable/executor_runner/executor_runner.cpp
mansnils: Done

JacobSzwejbka:
@mansnils Do you have strong preference to not fork the runner? 

We currently have a slight preference in this case to have you fork it rather then add the ifdef stuff. But if you feel really strongly not too we would hear you out.

mansnils:
> @mansnils Do you have strong preference to not fork the runner?
> 
> We currently have a slight preference in this case to have you fork it rather then add the ifdef stuff. But if you feel really strongly not too we would hear you out.

Yes we really prefer not to fork the runner.

robell:
> @mansnils Do you have strong preference to not fork the runner?
> 
> We currently have a slight preference in this case to have you fork it rather then add the ifdef stuff. But if you feel really strongly not too we would hear you out.

we're integrated with the Vulkan flows too, so running vulkan or VGF (vulkan extension features) delegates in the same runner and same system configs, sometimes just with a vulkan layer driver, is much preferred.

JacobSzwejbka (approved):
Ok that works for me

PR #15375, opened 2025-10-24
llm/custom_ops: follow test code conventions

gpires-meta:
Summary:
Rename Test Suites in `test_sdpa_with_kv_cache` to follow internal tooling conventions and make test target static listable.

Previously, some of the Test Suites in the modified test file did not satisfy the condition:
* Test Suite's name must contain `Test`

This diff makes it so that the Test Suites follow the expected conventions.

### Benefits
Following the conventions grants the test targets:
* Faster test listing
* Better dev experience in VSCode

Differential Revision: D85428922


digantdesai (approved):
Make sure the Github CI is green before merging. Stamping to unblock. Thanks.

PR #15380, opened 2025-10-24
Arm backend: support torch.sum.default

Erik-Lundell:
When sum is called without arguments,
it is traced as sum.default rather than sum.dim_IntList, which 1) is lowered to edge with dims=[] instead of None, and 2) is not annotated in the quantizer.

This commit fixes those issues and adds tests.


zingo:
@Erik-Lundell this has a tests fail in test-arm-backed (test_model_ethos-u55) something to fix or retrigger?

zingo:
This will probably go awaiw with a rebase trunk / test-mcu-models (cmake)
So Ill rebase and we see if the other error is still there.

zingo (approved):

PR #15394, opened 2025-10-27
NXP backend: Replace move relu before concat optimization

roman-janik-nxp:
### Summary
This PR replaces optimization in `move_relu_before_concat.py` by `MoveActivationBeforeConcat` aten pass. The pass moves selected activations that are supported for fusion on Neutron (Relu, Relu6, Sigmoid, Tanh) before the `concat` node if concat input nodes are either Conv 2D or Linear 2D. The whole node Logic is determined by target specs, now supporting Neutron-C. Tests updated. 

### Test plan
Unit tests provided (test_move_activation_before_concatenation.py).

cc @robert-kalmar


backends/nxp/aten_passes/move_activation_before_concat.py
```diff
@@ -0,0 +1,102 @@
...
+                )
+            ):
+                continue
+
+            # Loop all Cat input nodes and insert new activation after node.
+            for input_node in cat_node.all_input_nodes:
+                with module.graph.inserting_after(input_node):
+                    new_activation = module.graph.call_function(
+                        activation.target,
+                        args=(*activation.args[1:],),
```
MartinPavella: Minor:
Skipping the input argument here and then overwriting the `args` with the input node + these `args` (on line 93) seems confusing to me. Perhaps assigning an empty tuple here, and then using `new_activation.args = (input_node, *activation.args[1:])` on line 93 would be clearer.

backends/nxp/quantizer/patterns.py
```diff
@@ -745,3 +746,147 @@ def get_anchors(
...
+        def _activations_are_equal(activation1: Node, activation2: Node) -> bool:
+            if (  # Targets are equal also with their inplace variants
+                activation1.target in hardtanh_t
+                and activation2.target in hardtanh_t
+                or activation1.target in relu_t
+                and activation2.target in relu_t
+                or activation1.target in tanh_t
+                and activation2.target in tanh_t
+                or activation1.target == torch.ops.aten.sigmoid.default
+                and activation2.target == torch.ops.aten.sigmoid.default
```
MartinPavella: Minor:
Brackets would be useful.

MartinPavella (approved):

roman-janik-nxp:
The failing tests are not caused by this PR.

PR #15398, opened 2025-10-27
Pin huggingface_hub<1.0

GregoryComer:
### Summary
Trying to fix failing HF CI jobs. Looks like our transformers version requires huggingface_hub<1.0, but we're installing the latest 1.0 version which just released today. Until we bump transformers, I'll just pin our install of huggingface_hub to below 1.0.

Here's an example job failure:
```
transformers 4.56.1 requires huggingface-hub<1.0,>=0.34.0, but you have huggingface-hub 1.0.0 which is incompatible.
...
/exec: line 13: huggingface-cli: command not found
```
(from https://github.com/pytorch/executorch/actions/runs/18851188516/job/53787937311)

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -128,7 +128,7 @@ jobs:
         echo "::endgroup::"
 
         echo "::group::Setup Huggingface"
-        pip install -U "huggingface_hub[cli]" accelerate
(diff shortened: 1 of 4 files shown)


GregoryComer:
Abandoning to re-create not on a fork, since HF jobs require secrets.

PR #15397, opened 2025-10-27
Revert ARM PRs fdfeaa4..4675292

GregoryComer:
### Summary

Temporarily reverting https://github.com/pytorch/executorch/pull/15356 and stacked Arm backend PRs due to a buck build breakage blocking code sync. Will re-land when resolved. CC @digantdesai

PR #15400, opened 2025-10-27
Bump linux unittest runner to linux.2xlarge.memory

GregoryComer:
### Summary
Unittest jobs are dying since https://github.com/pytorch/executorch/pull/15180. I'm not entirely sure why the jobs passed on the PR, but they are failing consistently since on trunk. Testing a runner size bump (2xlarge -> 2xlarge.memory) to see if that resolves it.

Example logs (from another [PR](https://github.com/pytorch/executorch/pull/15397)). Note that the exit code (137 / SIGKILL) likely means OOM kill.
```
The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
...
Process completed with exit code 137.
```

### DIFF
--- a/.github/workflows/_unittest.yml
+++ b/.github/workflows/_unittest.yml
@@ -32,7 +32,7 @@ jobs:
       id-token: write
       contents: read
     with:
-      runner: linux.2xlarge
+      runner: linux.2xlarge.memory
       docker-image: ${{ inputs.docker-image }}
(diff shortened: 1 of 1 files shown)


GregoryComer:
Voxtral and samsung failures are due to running on a fork (I put up #15401 and #15402 to fix this).

SS-JIA (approved):

PR #15406, opened 2025-10-28
Arm backend: Avoid not decomposing linears we reject

Erik-Lundell:
If a linear is not quantized properly, we will reject it when partitioning. However, if we tell Executorch to _not_ not decompose an op, we are required to partition it. We thus need to figure out if we will partition the linear or not in the ops_not_to_decompose filter function.

Also turn off grad in the arm tester to solve an error that popped up in the GRU model. Since we only do inference, grad is never relevant.

### DIFF
--- a/backends/arm/test/models/test_nn_modules.py
+++ b/backends/arm/test/models/test_nn_modules.py
@@ -137,8 +137,6 @@ def test_nn_Modules_FP(test_data):
     "test_data",
     test_parameters,
     xfails={
-        "GRUModule": "RuntimeError: Node aten_linear_default with op <EdgeOpOverload: aten.linear[...]> was not decomposed or delegated.",
-        "PReLUModule": "RuntimeError: mul(): functions with out=... arguments don't support automatic differentiation, but one of the arguments requires grad.",
         "TransformerModule": "AssertionError: Output 0 does not match reference output.",
     },
 )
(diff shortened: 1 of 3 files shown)


backends/arm/tosa/partitioner.py
```diff
@@ -355,31 +356,45 @@ def filter_fn(node: torch.fx.Node) -> bool:
                 bool: True to keep the op intact; otherwise, False.
 
             """
-            dq = torch.ops.quantized_decomposed.dequantize_per_tensor.default
-            q = torch.ops.quantized_decomposed.quantize_per_tensor.default
+            dq = (
+                torch.ops.quantized_decomposed.dequantize_per_tensor.default,
+                torch.ops.quantized_decomposed.dequantize_per_channel.default,
```
zingo, line 361: Nice

backends/arm/test/models/test_nn_modules.py
```diff
@@ -137,8 +137,6 @@ def test_nn_Modules_FP(test_data):
     "test_data",
     test_parameters,
     xfails={
-        "GRUModule": "RuntimeError: Node aten_linear_default with op <EdgeOpOverload: aten.linear[...]> was not decomposed or delegated.",
```
zingo, line 140: Splendid

zingo (approved):

PR #15431, opened 2025-10-28
Support cosine operator on XNNPACK

GregoryComer:
Summary: Wire up the unary cosine operator in xnnpack for fp32 and fp16.

Differential Revision: D83623619


backends/xnnpack/operators/op_cos.py
```diff
@@ -0,0 +1,52 @@
...
+        vals_to_ids: Dict[torch.fx.Node, int],
+        debug_handle: int,
+    ) -> None:
+        self.define_nodes_tensor_inputs_outputs(node, xnn_graph, vals_to_ids)
+
+        # input
+        input_id = vals_to_ids[get_input_node(node, 0)]
+
+        # output
+        output_id = vals_to_ids[node]
```
digantdesai, line 42: Nit: assert dtype?

backends/xnnpack/partition/config/generic_node_configs.py
```diff
@@ -702,3 +702,10 @@ def check_constraints(self, node: torch.fx.Node, ep: ExportedProgram) -> bool:
             return False
 
         return True
+
+
+class CosConfig(GenericNodePartitionerConfig):
+    target_name = "cos.default"
+
+    def supported_precision_types(self) -> List[ConfigPrecisionType]:
+        return [ConfigPrecisionType.FP32]
```
digantdesai, line 711: we need to add FP16 and later BF16 here.

digantdesai (approved):

backends/xnnpack/partition/config/generic_node_configs.py
GregoryComer: Yeah. The way the partitioner is currently written, FP32 implies FP16. I'll likely refactor this a little bit when we add BF16 support.

backends/xnnpack/operators/op_cos.py
GregoryComer: Will take as a follow-up, if that's okay. This pattern is shared between many ops, so I might do a larger refactor.

PR #15433, opened 2025-10-28
Remove AWS Device Farm benchmark jobs

GregoryComer:
### Summary
Take down the AWS device farm benchmarking jobs. We are dropping them due to performance data being unreliable on non-rooted devices.


GregoryComer:
Note: I'm dropping the build-benchmark-app jobs, since they're built per benchmark run. If we want to keep those, I can move them to pull/trunk jobs to make sure they build. Let me know and I'll do that.

GregoryComer:
@mergennachin are we ready to take down the benchmark jobs?

Gasoonjia (approved):

PR #15434, opened 2025-10-28
[aoti-et] Add dtype bool support

larryliu0820:
As titled.

### DIFF
--- a/backends/cuda/runtime/utils.h
+++ b/backends/cuda/runtime/utils.h
@@ -62,6 +62,7 @@ enum class SupportedDTypes : int32_t {
   INT32 = 3, // PyTorch's int32 dtype code
   INT64 = 4, // PyTorch's int64 dtype code
   FLOAT32 = 6, // PyTorch's float32 dtype code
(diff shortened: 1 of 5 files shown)


backends/cuda/runtime/utils.h
```diff
@@ -96,13 +98,14 @@ inline AOTITorchError validate_dtype(int32_t dtype) {
   ET_CHECK_OR_RETURN_ERROR(
       is_dtype_supported_in_et_cuda(dtype),
       InvalidArgument,
-      "Unsupported dtype: %d. Supported dtypes: %d (int8), %d (int16), %d (int32), %d (int64), %d (float32), %d (bfloat16)",
+      "Unsupported dtype: %d. Supported dtypes: %d (int8), %d (int16), %d (int32), %d (int64), %d (float32), %d (bool), %d (bfloat16)",
```
Gasoonjia, line 101: any macro we can use here? seems it is getting longer and longer loll

Gasoonjia (approved):

PR #15446, opened 2025-10-29
Arm backend: Fix index bug in InsertRescaleInt32Pass

Erik-Lundell:
qparams = node.meta["input_qparams"] maps arg index to qparam, which is not guaranteed to go 0..N. In the pass,
range(len(qparams)) was used, which will always create indices 0..N. This caused a KeyError in several torch audio models.


zingo (approved):
Nice! Lets merge of all Arm tests passes

PR #15452, opened 2025-10-29
[llm] Beef up wav loader to read audio format 3 (float format)

larryliu0820:
This PR enables audio format 3 (IEEE float format) on `wav_loader.h`, which allows direct reading of float values without normalization. Also adds a unit test.

### DIFF
--- a/extension/llm/runner/test/test_wav_loader.cpp
+++ b/extension/llm/runner/test/test_wav_loader.cpp
@@ -133,6 +133,40 @@ std::vector<uint8_t> make_float_wav_bytes(
   return bytes;
 }
 
+std::vector<uint8_t> make_wav_bytes_with_format(
+    uint16_t audio_format,
+    int bits_per_sample,
+    const std::vector<uint8_t>& sample_data,
+    uint16_t num_channels = 1,
+    uint32_t sample_rate = 16000) {
+  const size_t bytes_per_sample = static_cast<size_t>(bits_per_sample / 8);
+  const uint32_t subchunk2_size = static_cast<uint32_t>(sample_data.size());
+  const uint32_t byte_rate = sample_rate * num_channels * bytes_per_sample;
+  const uint16_t block_align = num_channels * bytes_per_sample;
+  const uint32_t chunk_size = 36 + subchunk2_size;
+
+  std::vector<uint8_t> bytes;
+  bytes.reserve(44 + subchunk2_size);
+
+  append_bytes(bytes, "RIFF");
+  append_le32(bytes, chunk_size);
+  append_bytes(bytes, "WAVE");
+  append_bytes(bytes, "fmt ");
+  append_le32(bytes, 16);
+  append_le16(bytes, audio_format);
+  append_le16(bytes, num_channels);
+  append_le32(bytes, sample_rate);
+  append_le32(bytes, byte_rate);
+  append_le16(bytes, block_align);
+  append_le16(bytes, static_cast<uint16_t>(bits_per_sample));
+  append_bytes(bytes, "data");
+  append_le32(bytes, subchunk2_size);
+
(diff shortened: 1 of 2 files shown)


Gasoonjia (approved):

extension/llm/runner/wav_loader.h
```diff
@@ -168,18 +168,29 @@ inline std::vector<float> load_wav_audio_data(const std::string& fp) {
...
   if (bits_per_sample == 32) {
     size_t num_samples = data_size / 4;
     audio_data.resize(num_samples);
-    const int32_t* input_buffer =
-        reinterpret_cast<const int32_t*>(data + data_offset);
 
-    for (size_t i = 0; i < num_samples; ++i) {
-      audio_data[i] = static_cast<float>(
-          static_cast<double>(input_buffer[i]) * kOneOverIntMax);
+    if (audio_format == 3) {
```
Gasoonjia: can we use some constant variable to replace 3 here to make to more readbale?

extension/llm/runner/test/test_wav_loader.cpp
```diff
@@ -51,6 +51,13 @@ void append_le32(std::vector<uint8_t>& out, uint32_t value) {
   out.push_back(static_cast<uint8_t>((value >> 24) & 0xFF));
 }
 
+void append_float(std::vector<uint8_t>& out, float value) {
+  const uint8_t* bytes = reinterpret_cast<const uint8_t*>(&value);
```
msluszniak: Guess `const auto* bytes = value.data()` will work as well

extension/llm/runner/test/test_wav_loader.cpp
```diff
@@ -91,6 +98,75 @@ std::vector<uint8_t> make_pcm_wav_bytes(
...
+  return bytes;
+}
+
+std::vector<uint8_t> make_wav_bytes_with_format(
+    uint16_t audio_format,
+    int bits_per_sample,
+    const std::vector<uint8_t>& sample_data,
+    uint16_t num_channels = 1,
+    uint32_t sample_rate = 16000) {
+  const size_t bytes_per_sample = static_cast<size_t>(bits_per_sample / 8);
```
msluszniak: This is preference but in case of static casts the type is obvious so you can use shorter auto (also less bug prone), see: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-auto.html#cast-expressions. Same for other files

extension/llm/runner/test/test_wav_loader.cpp
```diff
@@ -91,6 +98,75 @@ std::vector<uint8_t> make_pcm_wav_bytes(
...
+  const uint32_t subchunk2_size = static_cast<uint32_t>(sample_data.size());
+  const uint32_t byte_rate = sample_rate * num_channels * bytes_per_sample;
+  const uint16_t block_align = num_channels * bytes_per_sample;
+  const uint32_t chunk_size = 36 + subchunk2_size;
```
msluszniak: 36 and 44 in this file are kinda magic numbers

extension/llm/runner/wav_loader.h
Gasoonjia: please use kWavFormatIeeeFloat to replace 3 here

extension/llm/runner/wav_loader.h
```diff
@@ -168,18 +171,38 @@ inline std::vector<float> load_wav_audio_data(const std::string& fp) {
...
+      // IEEE float format - read directly as floats
+      const float* input_buffer =
+          reinterpret_cast<const float*>(data + data_offset);
+      for (size_t i = 0; i < num_samples; ++i) {
```
msluszniak: probably can be replaced with `std::vector::assign`

PR #15459, opened 2025-10-30
Cortex_m backend: Add quantizer + avoid linear decomp

AdrianLundell:
Changes to_edge_and_transform to to_edge which supports the preserver_ops arg of the EdgeCompileConfig to avoid decomposing of the linear op. This significantly simplifies lowering the linear operator as it does not have to be re-fused.

Adds a cortex_m quantizer, with the intention to be general enough to be used for a general MCU. It is implemented as a ComposableQuantizer using multiple instances of a new OperatorConfigQuantizer class. This gives a number of abstraction levels for configuration

- McuQuantizer
- ComposableQuantizer
- OperatorConfig
- QuantizerConfig
- QuantizationSpec

The new quantizer also adds a transform_for_annotation pass pipeline which allows to fix scalar + tensor operations.


backends/cortex_m/passes/cortex_m_pass_manager.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+from executorch.backends.arm._passes import ScalarsToAttributePass
```
psiddh, line 7: Do you plan to add "EdgeCompileConfig(preserve_ops...." step in CortexMPassManager to avoid decomposition ?

psiddh: Or "Do you plan to implement a CortexMPartitioner with `ops_to_not_decompose()` method to work with `to_edge_transform_and_lower()`, similar to how the Cadence backend handles operation preservation?"
https://github.com/pytorch/executorch/blob/007ccc6f65cb098d794b5627da48683c06f68972/backends/cadence/aot/compiler.py#L275-L277

AdrianLundell: I am rather neutral towards how the list of ops to not preserve should be implemented, do you have a preference? From what I noticed preserve_ops in EdgeCompileConfig does not do anything in to_edge_transfrom_and_lower however, only in to_edge.

psiddh (approved):

PR #15471, opened 2025-10-30
Properly set modified for variety of cadence passes

DrJessop:
Summary: Update a large batch of passes to use the new remove or replace interface

Differential Revision: D85817305


abeakkas (approved):
With pending Phabricator comment.

PR #15478, opened 2025-10-30
Fix for missing ' executorch.backends.arm.common.type' (See #15367)

psiddh:
Summary: Same as title ^^^

Differential Revision: D85886023


GregoryComer (approved):

AdrianLundell:
Thanks for the fix!

PR #15488, opened 2025-10-31
Arm backend: Fix mypy warnings in test/passes

perheld:
Change-Id: I7915f4560fe0b2b1c949e21139612dae8acb8ee9

### DIFF
--- a/backends/arm/test/passes/test_broadcast_args_pass.py
+++ b/backends/arm/test/passes/test_broadcast_args_pass.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import operator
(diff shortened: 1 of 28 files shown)


zingo:
@digantdesai this touch a lot of files but only fixing mypy problem, do you see a need to check it internally?

zingo:
Hi @SS-JIA Im not sure if you want to test this internally or not if so let us know, it just changes small stuff in many files but does not move code around or add/remove files so we consider it safe but try to alert when when do "larger" changes.

SS-JIA (approved):
Should be ok I think. Just adds typing annotations, etc. so shouldn't break anything. LGTM!

PR #15492, opened 2025-10-31
Add pytorch-tokenizers as a dependency of ExecuTorch

larryliu0820:
So that we can `pip install executorch` and run python code like `examples/models/llama/runner/generation.py` without installing `pytorch-tokenizers` separately.


GregoryComer (approved):

lucylq (approved):
Nice!

PR #15504, opened 2025-10-31
Add torch.tensor to named_data_store

lucylq:
Summary: Allow adding torch.Tensor to named_data_store, and infer the TensorLayout.

Differential Revision: D85992938


larryliu0820 (approved):

PR #15507, opened 2025-11-01
FastGelu custom htp kernel

cccclai:
```
python3 examples/qualcomm/custom_op/custom_ops_fast_gelu.py --build_folder build-android  -m SM8750 --op_package_dir examples/qualcomm/custom_op/fastgelu_op_package_htp/FastGeluOpPackage --build_op_package 
```


cccclai:
Hi team, I add a FastGelu example, but I didn't use HTP intrinsics so the perf is still not optimized. Would like to know where to put these examples

cccclai:
Can I get a review on this PR?

shewu-quic:
> Hi team, I add a FastGelu example, but I didn't use HTP intrinsics so the perf is still not optimized. Would like to know where to put these examples

Thanks for your contribution. I think it is ok to put this custom op in examples/qualcomm/custom_op.

PR #15510, opened 2025-11-03
Allow data casting along with cloning.

shoumikhin:
Summary: .

Differential Revision: D86070966

### DIFF
--- a/extension/tensor/tensor_ptr.cpp
+++ b/extension/tensor/tensor_ptr.cpp
@@ -164,7 +164,9 @@ TensorPtr make_tensor_ptr(
       [data = std::move(data)](void*) {});
 }
 
-TensorPtr clone_tensor_ptr(const executorch::aten::Tensor& tensor) {
+TensorPtr clone_tensor_ptr(
+    const executorch::aten::Tensor& tensor,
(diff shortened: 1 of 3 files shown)


extension/tensor/tensor_ptr.cpp
```diff
@@ -178,23 +180,63 @@ TensorPtr clone_tensor_ptr(const executorch::aten::Tensor& tensor) {
...
+              CTYPE_TO* data_ptr = reinterpret_cast<CTYPE_TO*>(data.data());
+              std::transform(
+                  tensor_data_ptr,
+                  tensor_data_ptr + tensor_numel,
+                  data_ptr,
+                  [](const CTYPE_FROM& val) {
+                    return static_cast<CTYPE_TO>(val);
+                  });
+            });
+      });
```
larryliu0820, line 232: Can this code be moved to kernels/portable? maybe clone_ops_util or something like that

larryliu0820 (approved):

extension/tensor/tensor_ptr.cpp
shoumikhin: Maybe, but we probably don't want the tensor stuff depending on some kernels utils

PR #15514, opened 2025-11-03
Summary: ARM backend: Add skip markers for 16A8W quantization known t…

psiddh:
…est failures

Added skip functionality (because buck based targets do not seem to comply with xfail marker) to the test parametrize framework to handle known failing tests in ARM backend 16A8W quantization. The model_linear_rank4_zeros, model_linear_rank4_negative_ones, and model_linear_rank4_negative_large_rand test cases are now skipped due to bias quantization accuracy issues tracked in MLETORCH-1452. This prevents buck based CI from blocking while maintaining visibility of the known issues through proper test annotations.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/backends/arm/test/common.py
+++ b/backends/arm/test/common.py
@@ -237,6 +237,7 @@ def parametrize(
     arg_name: str,
     test_data: dict[str, Any],
     xfails: dict[str, xfail_type] | None = None,
+    skips: dict[str, str] | None = None,
     strict: bool = True,
     flakies: dict[str, int] | None = None,
 ) -> Decorator:
@@ -249,6 +250,8 @@ def parametrize(
     """
     if xfails is None:
         xfails = {}
+    if skips is None:
+        skips = {}
     if flakies is None:
         flakies = {}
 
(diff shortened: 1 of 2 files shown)


AdrianLundell:
Seems a bit strange to me that xfails would not work for a particular test case, or am I misunderstanding something? Also, if this is a temporary fix I would prefer if you could break out the odd test cases into a new test and skip that rather than modifying the testing infrastructure.

metascroy (approved):

psiddh:
> Seems a bit strange to me that xfails would not work for a particular test case, or am I misunderstanding something? Also, if this is a temporary fix I would prefer if you could break out the odd test cases into a new test and skip that rather than modifying the testing infrastructure.

Not sure if this is modifying any testing infra. It is only adding 'skip' marker to parameter impl as fwd fix. Only for the tests that are failing on internal CI (buck based), added the skip marker for now so that we don't have to backout a PR , satisfying both OSS CI & internal buck based CI.  We can discuss a cleaner fix if you have better ideas / suggestions but we need to keep the CI train for now moving as it has been stalled since Friday due to these issues

PR #15518, opened 2025-11-03
Arm backend: Merge FP and INT pass pipelines

oscarandersson8218:
Merges FP and INT pass pipelines into one pipeline.

### DIFF
--- a/backends/arm/_passes/annotate_decomposed_matmul.py
+++ b/backends/arm/_passes/annotate_decomposed_matmul.py
@@ -64,6 +64,7 @@ class AnnotateDecomposedMatmulPass(ArmPass):
         )
         matmul_targets = {
             exir_ops.edge.aten.bmm.default,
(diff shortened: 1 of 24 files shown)


zingo:
@digantdesai @mergennachin This is touching many files but will not add file och move API around so I think it should be fine form buck2 perspective to merge. Ill wait a few day before merging to let any of you have time to stop up if you think it needs internal testing anyway.

zingo:
Hi @SS-JIA Im not sure if you want to test this internally or not if so let us know, it just changes small stuff in many files but does not move code around or add/remove files so we consider it safe but just try to alert you when when do "larger" changes.

SS-JIA:
Should be ok but just to err on the side of caution I will trigger internal test.

SS-JIA (approved):
Having some trouble importing the diff at the moment; but the changes look fairly inocuous so LGTM to me.

PR #15526, opened 2025-11-03
Cortex_m backend: Simplify add + linear fusion passes

AdrianLundell:
Reuses the FoldAndAnnotateQParamsPass from the Arm backend to greatly simplify the logic for fusing the ops.

Additionally updates the linear kernel to be numerically correct and computes the kernel_sum aot in the quantized_linear_fusion pass. Note that since this replaces the bias node it typically causes no extra memory usage.

Updates the Linear tests to mirror this, including removing the various matmul tests. Since the linear is handled as a separate op rather than a particular type of matmul these tests are not related anymore.

Removes unnecessary stub definitions in operators.py, operators.yaml and op_quantized_linear.cpp

Leaving a few TODO:s since the patch is large already.


psiddh:
@AdrianLundell  Just want to better understand the rationale behind removing per channel support / scratch buffer code and in general stateful ops related code ?  (I do understand the FoldDQQ related changes to the pass though),  (edited)

AdrianLundell:
The implementation didn't produce numerically results originally so I wanted to fix that before adding per-channel support. Not doing that in this patch was mostly to keep the patch size down  and prioritizing the most important functionality before going into details.

Regarding the stateful scratch buffer, since we compute the kernel sum which the buffer was used for AOT and replace the bias, there is no reason to do that in the runtime as I see it. The only use-case I see where that would make sense (using the MVE implementation) is if you are very tight on memory and prefer to spend some extra time during each interference to compute the kernel sum in a scratch buffer rather than keeping it in memory. But then again there is no reason to do everything in one single patch, better to get a good minimal working example running IMO.

psiddh:
> The implementation didn't produce numerically results originally so I wanted to fix that before adding per-channel support. Not doing that in this patch was mostly to keep the patch size down and prioritizing the most important functionality before going into details.
> 
> Regarding the stateful scratch buffer, since we compute the kernel sum which the buffer was used for AOT and replace the bias, there is no reason to do that in the runtime as I see it. The only use-case I see where that would make sense (using the MVE implementation) is if you are very tight on memory and prefer to spend some extra time during each interference to compute the kernel sum in a scratch buffer rather than keeping it in memory. But then again there is no reason to do everything in one single patch, better to get a good minimal working example running IMO.

- Yes, we need scratch buffer for DSP & MVE (Helium) implementations, I think
- Given that, I would suggest, to keep CMSISScratchBufferContext file as-is , for now. The idea was to evolve this into a generic ScratchBuffer impl that could be reused for couple of diff optimization scenarios
- How are you testing these changes ? E2E on FVP (or) some real device for numerical results.  Can you share the steps please ?
- Can we add regression tests specifically for the numerical correctness fixes mentioned ?

backends/cortex_m/ops/op_quantized_linear.cpp
```diff
@@ -20,151 +20,90 @@ using KernelRuntimeContext = torch::executor::KernelRuntimeContext;
...
-  }
-
-  // Initialize scratch buffer context (validates early)
-  CMSISScratchBufferContext scratch_ctx(
-      const_cast<Tensor&>(scratch_buffer), weights, weight_zero_point, bias);
 
-  scratch_ctx.compute_kernel_sums_if_needed();
-  cmsis_nn_context ctx = scratch_ctx.get_cmsis_ctx();
+  cmsis_nn_context ctx;
+  ctx.size = 2;
```
psiddh: Can you add comments why this is '2' ?

backends/cortex_m/test/ops/test_linear.py
```diff
@@ -4,21 +4,18 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import pytest
 import torch
+from executorch.backends.arm.test.common import parametrize
```
psiddh, line 8: Are these tests running in CI ?

psiddh:
Also can you add comments in the code explaining why scratch buffer context is simplified and when it might still be needed ? Overall I think this is a solid refactor in the right direction

backends/cortex_m/test/ops/test_linear.py
AdrianLundell: It is running in internal CI for us but I have not added it to upstream CI yet, that is on m TODO list.

AdrianLundell:
> 
>     * Yes, we need scratch buffer for DSP & MVE (Helium) implementations, I think
> 
We will need scratch buffer for some ops yes, but I am not sure if the header solution for keeping track of the state will ever be needed. As I see it, we can either do the computation AOT as in this case, or it will be reset with every inference as for CONV. Maybe I am missing some use-case here?

>     * Given that, I would suggest, to keep CMSISScratchBufferContext file as-is , for now. The idea was to evolve this into a generic ScratchBuffer impl that could be reused for couple of diff optimization scenarios
> 
Agree we can keep it for potential future use.

>     * How are you testing these changes ? E2E on FVP (or) some real device for numerical results.  Can you share the steps please ?
>
The python op implementation is tested in the test_dialect_linear in backends/cortex_m/test/ops/test_linear.py and the cpp implementation is tested on FVP in the test_implementation_linear tests. The tests for running this after having downloaded executorch are:
```
python3 -m venv env
source env/bin/activate
pip install -U pip setuptools wheel cmake pytest-cov zstd ninja
./install_executorch.sh --editable
examples/arm/setup.sh --i-agree-to-the-contained-eula
source examples/arm/ethos-u-scratch/setup_path.sh

#To test installation
examples/arm/run.sh

# To build cortex-m test runner
backends/cortex_m/test/build_test_runner.sh

# To run linear tests:
pytest --config-file=backends/arm/test/pytest.ini backends/cortex_m/test/ops/test_linear.py
```
>     * Can we add regression tests specifically for the numerical correctness fixes mentioned ?
These are available in test_linear.py. These tests were not passing before the patch.

backends/cortex_m/ops/op_quantized_linear.cpp
AdrianLundell: Good catch, this is not used so I will add a comment explaining that instead.

AdrianLundell:
> Also can you add comments in the code explaining why scratch buffer context is simplified and when it might still be needed ? Overall I think this is a solid refactor in the right direction

I think the logic is well explained in the commit message + the quantized_linear_fusion_pass.py

psiddh:
Thanks,  looks good over all, approving as we also need to move faster.
Also can you check FVP CI failures (not related ?)

[EDIT:] Looks like there are compilation issues / CI failures  (not sure if it is related to https://github.com/pytorch/executorch/pull/15612)

psiddh (approved):

AdrianLundell:
Fails in XNNPACK/ios-arm64-coreml-fp16 not related

PR #15528, opened 2025-11-03
[aoti-backend-consolidation 2/3] backend.py 

Gasoonjia:
Summary:

# Summary

This diff consolidates the backend functionality into a single target `//executorch/backends/aoti:aoti_backend` and simplifies the cuda backend target by making it dependent on the consolidated backend target. 

The following changes are made in this diff:

*   Creation of a new target `//executorch/backends/aoti:aoti_backend` in `fbcode/executorch/backends/aoti/targets.bzl` which includes the necessary dependencies for the AOTI backend.
*   Update of the `//executorch/backends/cuda:cuda_backend` target in `fbcode/executorch/backends/cuda/TARGETS` to depend on the new `//executorch/backends/aoti:aoti_backend` target instead of individual AOTI backend dependencies.
*   Creation of a new file `fbcode/executorch/backends/aoti/aoti_backend.py` which imports the necessary dependencies and passes for the AOTI backend.
*   Simplification of the `xplat/executorch/backends/cuda/cuda_backend.py` file by removing unnecessary imports and using the new `AotiBackend` class from the `aoti_backend.py` file.
ghstack-source-id: 319556735

Reviewed By: larryliu0820

Differential Revision: D85704977

### DIFF
--- a/backends/aoti/aoti_backend.py
+++ b/backends/aoti/aoti_backend.py
@@ -17,11 +17,7 @@ from executorch.backends.aoti.passes.replace_view_copy_with_view import (
 )
 from executorch.exir._serialize._named_data_store import NamedDataStore
 from executorch.exir._warnings import experimental
-from executorch.exir.backend.backend_details import (
-    BackendDetails,
-    ExportedProgram,
-    PreprocessResult,
-)
+from executorch.exir.backend.backend_details import ExportedProgram, PreprocessResult
 from executorch.exir.backend.compile_spec_schema import CompileSpec
 from torch._inductor.codegen.cpp_wrapper_cpu import CppWrapperCpu
 from torch.export.passes import move_to_device_pass
@@ -34,12 +30,16 @@ class COMPILE_SPEC_KEYS(Enum):
 @experimental(
     "This API and all of aoti-driven backend related functionality are experimental."
 )
-class AotiBackend(BackendDetails, ABC):
+class AotiBackend(ABC):
     """
-    Base backend class for AOTInductor-based backends.
+    Base mixin class for AOTInductor-based backends.
 
     This class provides common functionality for compiling models using AOTInductor
     with different device targets (CUDA, Metal, etc.).
+
+    This is a mixin class, not an actual backend object, for aoti-driven backens.
+    Concrete backends (e.g., CudaBackend, MetalBackend) should inherit from both
+    BackendDetails and AotiBackend to get the full functionality.
     """
 
     @staticmethod
--- a/backends/aoti/aoti_backend.py
+++ b/backends/aoti/aoti_backend.py
@@ -39,13 +39,13 @@ class AotiBackend(BackendDetails, ABC):
     Base backend class for AOTInductor-based backends.
 
     This class provides common functionality for compiling models using AOTInductor
-    with different device targets (CUDA, Metal/MPS, etc.).
+    with different device targets (CUDA, Metal, etc.).
     """
 
     @staticmethod
     @abstractmethod
     def get_device_name() -> str:
-        """Return the device name for this backend (e.g., 'cuda', 'mps')."""
+        """Return the device name for this backend (e.g., 'cuda', 'metal')."""
         pass
 
     @staticmethod
--- a/backends/aoti/aoti_backend.py
+++ b/backends/aoti/aoti_backend.py
@@ -148,7 +148,9 @@ class AotiBackend(ABC):
         options = cls.get_aoti_compile_options()
 
         # Move the edge_program to the target device
-        device_edge_program = move_to_device_pass(edge_program, device_name)
+        device_edge_program = move_to_device_pass(
+            edge_program, device_name if device_name != "metal" else "mps"
+        )
 
         # Replace view_copy with view
         ReplaceViewCopyWithViewPass()(device_edge_program.graph_module)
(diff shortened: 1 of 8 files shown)


larryliu0820 (approved):

backends/apple/metal/metal_backend.py
```diff
@@ -117,83 +54,3 @@ def preprocess(
...
-
-        # Keep the so file in the NamedDataStore, so that it can be packaged into the .pte file.
-        named_data_store.add_named_data(method_name + "_so_blob", so_data, 1, None)
-
-        # Add weights blob to named data store
-        with open(blob_path, "rb") as f:
-            blob_data = f.read()
-
-        named_data_store.add_named_data(
-            method_name + "_weights_blob", blob_data, 1, "aoti_metal_blob"
```
manuelcandales, line 159: what's the new name?

backends/aoti/aoti_backend.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Create named data store
+        named_data_store = NamedDataStore()
+        method_name = cls.method_name_from_compile_specs(compile_specs)
+
+        # Add SO and weights blob separately
+        named_data_store.add_named_data(method_name + "_so_blob", so_data, 1, None)
+        weights_blob_data_type = f"aoti_{device_name}_blob"
+        named_data_store.add_named_data(
+            method_name + "_weights_blob", blob_data, 1, weights_blob_data_type
```
Gasoonjia, line 239: here's the blob name in the new setting @manuelcandales

backends/apple/metal/metal_backend.py
Gasoonjia: here's the new name: https://github.com/pytorch/executorch/pull/15528/files#r2496118034

Gasoonjia: ahh i called it mps now instead of metal; thanks let me solve it

backends/aoti/aoti_backend.py
```diff
@@ -0,0 +1,261 @@
...
+        with open(blob_path, "rb") as f:
+            blob_data = f.read()
+
+        # Create named data store
+        named_data_store = NamedDataStore()
+        method_name = cls.method_name_from_compile_specs(compile_specs)
+
+        # Add SO and weights blob separately
+        named_data_store.add_named_data(method_name + "_so_blob", so_data, 1, None)
+        weights_blob_data_type = f"aoti_{device_name}_blob"
```
manuelcandales, line 237: yeah, this will now be `aoti_mps_blob`, so, need to update this in metal.yml (uses aoti_metal_blob)

exir/backend/backend_api.py
```diff
@@ -581,9 +581,21 @@ def lower_all_submodules_to_backend(
...
+        subclasses = set()
+        if len(cls.__subclasses__()) == 0:
+            return {cls}
+        else:
+            for subclass in cls.__subclasses__():
+                # Recursively check subclasses
+                subclasses.update(_get_all_final_backend_details_subclasses(subclass))
+        return subclasses
+
     backend_name_to_subclass = {
```
cccclai, line 584: I remember it was discussed at the beginning of the delegate meeting - the backend implementation should be final and we didn't want to get involve in nested backends https://github.com/pytorch/executorch/blob/3e9629a4591ff038b380e2f59827f92d466d49a1/exir/backend/backend_api.py#L111 it's hard to guard in python though

cccclai: In the example we put, we did try to put a final https://github.com/pytorch/executorch/blob/3e9629a4591ff038b380e2f59827f92d466d49a1/exir/backend/test/backend_with_compiler_demo.py#L28

cccclai: If there are shared logic, maybe put up a shared folder?

Gasoonjia: Thanks for comments. I make the AOTIBackend as a collector of sharing logics but not a actual backend, and make CUDABackend and MetalBackend inherit from both AOTIBackend and BackendDetails to avoid making any update on backend_api.py.

manuelcandales (approved):

backends/aoti/aoti_backend.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class COMPILE_SPEC_KEYS(Enum):
+    METHOD_NAME = "method_name"
+
+
+@experimental(
+    "This API and all of aoti-driven backend related functionality are experimental."
+)
+class AotiBackend(ABC):
```
cccclai, line 33: Is AotiBackend an actual class or an abstract class?

setup.py
```diff
@@ -465,13 +483,9 @@ def run(self):
...
         try:
             # Following code is for building the Qualcomm backend.
-            from backends.qualcomm.scripts.download_qnn_sdk import (
-                _download_qnn_sdk,
-                is_linux_x86,
-            )
-
             if is_linux_x86():
                 os.environ["EXECUTORCH_BUILDING_WHEEL"] = "1"
+                from backends.qualcomm.scripts.download_qnn_sdk import _download_qnn_sdk
```
cccclai: Mengwei has this PR https://github.com/pytorch/executorch/pull/15546 maybe can coordinate how to land together or maybe after he merged his change

backends/aoti/aoti_backend.py
cccclai: After reading, it seems like we will just have metal backend and cuda backend as the actual backend?

Gasoonjia: yes it is just a abstract class not a real backend for meet backend_details' requirement

setup.py
Gasoonjia: have reverted my changes

PR #15538, opened 2025-11-03
Mmap windows

JacobSzwejbka:
Current windows implementation had issues with 64 bit files. This expands the coverage to work properly and add some tests

### DIFF
--- a/extension/data_loader/mman.h
+++ b/extension/data_loader/mman.h
@@ -13,6 +13,7 @@
 
 #include <executorch/runtime/platform/compiler.h>
 #include <sys/stat.h>
+#include <cstdint>
 
 #ifndef _WIN32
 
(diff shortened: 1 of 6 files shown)


extension/data_loader/mman.h
```diff
@@ -40,4 +61,23 @@ ET_INLINE long get_os_page_size() {
   return pagesize;
 }
 
+/**
+ * Platform-specific file stat function.
+ */
+ET_INLINE int get_file_stat(int fd, size_t* out_size) {
```
lucylq, line 67: So, having `out_size` as a `size_t*` should not pose an issue as we wouldn't be able to map >4gb on 32bit system anyways, right?

lucylq (approved):
LGTM, thanks for adding the test!

cc @SS-JIA for windows changes as well.

extension/data_loader/mman.h
JacobSzwejbka: Yeah

Issue #12804, opened 2025-07-24
Support for Shared Weights between multiple entry points in PTE export


pkmandke:
Is there (ideally, backend agnostic) support for sharing weights in a multiple entry point PTE exported file? [#11738](https://github.com/pytorch/executorch/issues/11738) seems to track this for XNNPack. However, it doesn't seem to be supported based on the last update on the thread (?)

----

On similar lines, alternatively, is it possible to serialize and export just the (FX) graph/s and load the weights separately (say from a [flat_tensor](https://github.com/pytorch/executorch/tree/main/extension/flat_tensor))?

A wishful API would be something like this:
```python
ep = export(model, inputs)
ep = _export_forward_backward(ep)

ep = to_edge(ep)
# Optionally ep.to_backend()
ep = ep.to_executorch()

# Now a wishful workflow would look something like:
ep.save_graph_only(path)
ep.save_weights_only(path)
```

cc @JacobSzwejbka @angelayi

lucylq:
Hi @pkmandke, 

> Is there (ideally, backend agnostic) support for sharing weights in a multiple entry point PTE exported file

Yes, weight-sharing across multiple entry points in a PTE is supported for XNNPACK, and I believe is also the default.

The issue you're referencing also tracks sharing memory/state between entrypoints, which doesn't have good support at the moment cc @JacobSzwejbka 


> is it possible to serialize and export just the (FX) graph/s and load the weights separately (say from a [flat_tensor](https://github.com/pytorch/executorch/tree/main/extension/flat_tensor))?

Yes, there is support for saving the program and weights separately. You can see a demo here: https://github.com/pytorch-labs/executorch-examples/tree/main/program-data-separation

This will save your file to flat_tensor (PTD) format. Again this supported for XNNPACK but not other delegates at the moment.

pkmandke:
>Yes, weight-sharing across multiple entry points in a PTE is supported for XNNPACK, and I believe is also the default.

It doesn't seem to work for me. Here is the example I am using along with the latest executorch off `main`:

```py
class Net(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(2048, 128256)
        self.linear2 = nn.Linear(128256, 2048)

    def forward(self, x):
        return self.linear2(F.sigmoid(self.linear(x)))

    def forward_2(self, x):
        return self.linear2(F.relu(self.linear(x)))

net = Net()
x = torch.randn(1, 2048)
[...]

PR #15541, opened 2025-11-04
Don't cache source partitions

lucylq:
### Summary
This caching breaks multi-method; for the second method, we are using the graph of the first method, and do not delegate properly. I think this is because the source partitions are based on the first graph, so the addmm nodes do not match up/do not give a partition, so the graph doesn't get tagged by the partitioner.

See issue: https://github.com/pytorch/executorch/issues/12804

In the example from the issue, the first method `fwd` is delegated well, and the second method `fwd_2` does not have the linear functions delegated (it is using portable kernels, and the weights are duplicated in the constant segment). 

Also add test_partitioner to buck targets. 

### Test plan
python -m unittest backends.xnnpack.test.test_xnnpack_partitioner

### DIFF
--- a/backends/xnnpack/partition/config/gemm_configs.py
+++ b/backends/xnnpack/partition/config/gemm_configs.py
@@ -458,9 +458,7 @@ class AddmmConfig(GEMMConfig):
         a bool indicating if the deps are valid and a list of all the
         dep nodes. This handles the src partition for
         """
-        if self.src_partitions is None:
-            # Cache src partitions so we don't have to recompute them every time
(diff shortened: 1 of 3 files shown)


GregoryComer (approved):
Thanks for adding this. FYI We've deprecated the to_edge + to_backend flow for XNNPACK and will throw up a big warning when used. Fixing the caching issue is probably good regardless, even if it mostly only affects to_edge + to_backend (we preserve linear w/ to_edge_transform_and_lower).

PR #15544, opened 2025-11-04
Add CUDA backend to pybind

larryliu0820:
This PR integrates CUDA and AOTI support into the pybind build system. The implementation starts by updating setup.py to automatically detect CUDA availability using install_utils.py functions, replacing the problematic sys.path hack with a clean importlib-based approach. This enables automatic building of CUDA and AOTI targets when CUDA is detected on the system.

The changes then extend to CUDA runtime shims for SlimTensor support, update AOTI build targets and common shims, and enhance CI workflows for CUDA testing. The implementation ensures proper symbol resolution when loading AOTI-produced shared libraries.

## Test Plan

- [x] Verified setup.py imports work correctly without sys.path modifications
- [x] Confirmed CUDA detection functionality works as expected
- [x] Validated that install_utils functions are accessible via importlib approach
- [x] Tested setup.py basic functionality (--help, --version) still works
- [x] Confirmed Python syntax validation passes for all modified files

Co-Authored-By: Claude Sonnet 4 <noreply@anthropic.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -668,6 +668,16 @@ if(EXECUTORCH_BUILD_CORTEX_M)
   list(APPEND _executorch_backends coretex_m_backend)
 endif()
 
+# If CUDA, Metal, or pybind will need Torch, find it at root scope first so that
+# imported targets (e.g. MKL::MKL) aren't created in child directory scopes and
+# then duplicated when found again at root scope.
+if(EXECUTORCH_BUILD_CUDA
+   OR EXECUTORCH_BUILD_METAL
+   OR EXECUTORCH_BUILD_PYBIND
+)
+  find_package_torch()
+endif()
+
 # Build common AOTI functionality if needed by CUDA or Metal backends
 if(EXECUTORCH_BUILD_CUDA OR EXECUTORCH_BUILD_METAL)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/backends/aoti)
@@ -885,6 +895,15 @@ if(EXECUTORCH_BUILD_PYBIND)
       util
       torch
   )
+
+  # Build common AOTI functionality if needed by CUDA or Metal backends
+  if(EXECUTORCH_BUILD_CUDA)
+    # CUDA uses SlimTensor-based shims
+    list(APPEND _dep_libs aoti_cuda_backend)
+  elseif(EXECUTORCH_BUILD_METAL)
+    # Metal still uses ETensor-based shims (for now)
+    list(APPEND _dep_libs aoti_common)
+  endif()
 
   # RPATH for _portable_lib.so
   set(_portable_lib_rpath "$ORIGIN/../../../torch/lib")
--- a/backends/cuda/runtime/platform/platform.cpp
+++ b/backends/cuda/runtime/platform/platform.cpp
@@ -11,6 +11,7 @@
 #include <executorch/runtime/core/error.h>
 #include <executorch/runtime/core/result.h>
 #include <filesystem>
+#include <mutex>
 #include <string>
 
 #ifdef _WIN32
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -769,6 +783,12 @@ def run(self):  # noqa C901
             item for item in re.split(r"\s+", os.environ.get("CMAKE_ARGS", "")) if item
         ]
 
+        # Check if CUDA is available, and if so, enable building the CUDA
+        # backend by default.
+        if install_utils.is_cuda_available() and install_utils.is_cmake_option_on(
```
JacobSzwejbka, line 704: So the torch dep is determined automatically but if you want backend support in the pybind you have to set the CMAKE flag to ON?

larryliu0820: the cmake option default is True. this gives us a way to disable it.

backends/cuda/runtime/platform/platform.cpp
```diff
@@ -41,6 +41,22 @@ executorch::runtime::Result<void*> load_library(
   }
 
 #else
+  // Before loading the delegate .so, we need to ensure symbols from the current
+  // process (e.g., _portable_lib.so) are globally visible. Python loads modules
+  // with RTLD_LOCAL by default, so we re-open the current module with
+  // RTLD_GLOBAL | RTLD_NOLOAD to promote its symbols to global visibility.
+  // This allows the delegate .so to resolve symbols like aoti_torch_dtype_*.
+  static bool symbols_promoted = false;
+  if (!symbols_promoted) {
```
lucylq: noob q: is this flag necessary?

CMakeLists.txt
```diff
@@ -886,6 +815,19 @@ if(EXECUTORCH_BUILD_PYBIND)
       torch
   )
 
+  # Build common AOTI functionality if needed by CUDA or Metal backends
+  if(EXECUTORCH_BUILD_CUDA)
+    # CUDA uses SlimTensor-based shims
+    list(APPEND _dep_libs aoti_common_shims_slim)
```
larryliu0820: @Gasoonjia I am using it here

setup.py
```diff
@@ -62,6 +63,16 @@
 from pathlib import Path
 from typing import List, Optional
 
+# Clean dynamic import using importlib
```
Gasoonjia, line 66: what's this for?

setup.py
```diff
@@ -743,6 +761,10 @@ def run(self):  # noqa C901
         if cmake_cache.is_enabled("EXECUTORCH_BUILD_EXTENSION_LLM_RUNNER"):
             cmake_build_args += ["--target", "_llm_runner"]
 
+        if cmake_cache.is_enabled("EXECUTORCH_BUILD_CUDA"):
+            cmake_build_args += ["--target", "aoti_cuda_backend"]
```
Gasoonjia, line 765: aoti_cuda_backend covers aoti_common_shims_slim content, no matter https://github.com/pytorch/executorch/pull/17217 land or not. I think we can get rid of `aoti_common_shims_slim`.

CMakeLists.txt
```diff
@@ -886,6 +815,19 @@ if(EXECUTORCH_BUILD_PYBIND)
...
+  elseif(EXECUTORCH_BUILD_METAL)
+    # Metal still uses ETensor-based shims (for now)
+    list(APPEND _dep_libs aoti_common)
+  endif()
+
+  if(EXECUTORCH_BUILD_CUDA)
+    list(APPEND _dep_libs aoti_cuda_backend)
```
Gasoonjia: combine the two if(EXECUTORCH_BUILD_CUDA)?

backends/cuda/runtime/platform/platform.cpp
larryliu0820: Yeah? We don't want to do this twice. This is doing exposing symbols in the pybind shared library _portable_lib.so

backends/cuda/runtime/platform/platform.cpp
```diff
@@ -41,6 +42,28 @@ executorch::runtime::Result<void*> load_library(
   }
 
 #else
+  // Before loading the delegate .so, we need to ensure symbols from the current
+  // process (e.g., _portable_lib.so) are globally visible. Python loads modules
+  // with RTLD_LOCAL by default, so we re-open the current module with
+  // RTLD_GLOBAL | RTLD_NOLOAD to promote its symbols to global visibility.
```
Gasoonjia, line 48: Is this the way we solve the duplicate-declaration issue for shim layers?

backends/aoti/CMakeLists.txt
```diff
@@ -92,8 +92,9 @@ if(EXECUTORCH_BUILD_CUDA)
   target_link_libraries(aoti_common_shims_slim PUBLIC CUDA::cudart)
 endif()
 
+# SlimTensor is header-only, no library to link against
 target_link_libraries(
-  aoti_common_shims_slim PUBLIC slimtensor extension_tensor ${CMAKE_DL_LIBS}
+  aoti_common_shims_slim PUBLIC extension_tensor ${CMAKE_DL_LIBS}
```
Gasoonjia: i would like to land https://github.com/pytorch/executorch/pull/17217 first then we no longer need this update.

Gasoonjia (approved):
LGTM, thansk for adding pybinding support!

backends/cuda/runtime/platform/platform.cpp
larryliu0820: actually we don't have duplicate symbols. It's likely that libtorch symbols are not exported, but I didn't dive deep. This is to resolve the issue that _portable_lib.so symbols are not visible to aoti backend lib.

setup.py
larryliu0820: so that I can use install_utils.py in setup.py. We can't directly import it before installation is completed.

backends/cuda/runtime/platform/platform.cpp
Gasoonjia: interesting cuz that's the issue me and some other guy encountered previously. But it works it works loll

PR #15545, opened 2025-11-04
Arm backend: Use correct log-level in tests

oscarandersson8218:
Arm tests logged too much as comparisons with logger.level were used instead of logger.getEffectiveLevel(). logger.level will always be logging.NOTSET unless explicitly set with logger.setLevel() which we want to avoid. Instead, we should use logger.getEffectiveLevel() which will inherit the level from its parent.

### DIFF
--- a/backends/arm/test/conftest.py
+++ b/backends/arm/test/conftest.py
@@ -6,10 +6,11 @@
 import logging
 import os
 import random
-import sys
 from typing import Any
 
 import pytest
+
+logger = logging.getLogger(__name__)
 
 """
 This file contains the pytest hooks, fixtures etc. for the Arm test suite.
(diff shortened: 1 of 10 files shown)


backends/arm/test/conftest.py
```diff
@@ -98,7 +97,7 @@ def set_random_seed():
                 "ARM_TEST_SEED env variable must be integers or the string RANDOM"
             )
 
-    print(f" ARM_TEST_SEED={seed} ", end=" ")
+    logger.info(f"ARM_TEST_SEED={seed}")
```
zingo: I think this was on purpose to ensure ARM_TEST_SEED=xxx is always printed so it can be reproduced with the same seed if needed later.

oscarandersson8218: Good catch! I'll add it back.

zingo (approved):

zingo:
Fails unrelated, buck2 files should not be needed to be adopted

PR #4998, opened 2024-08-30
Qualcomm AI Engine Direct - Check the version QNN API and backend API

shewu-quic:
## Summary:

QNN API version format is major.minor.patch.
- If major given by the user does not match the built major, it will return and show error message.
- If minor does not match, it will show warning message.
## Scenarios
| User QNN API Version | Build QNN API Version | Log Output |
|--------------|---------------|------------|
| 2.17.0      | 2.15.0       | [WARNING] [Qnn ExecuTorch]: Qnn backend library version 2.17.0 is used. The version is tested against 2.15.0.|
| 2.14.0      | 2.15.0      | [WARNING] [Qnn ExecuTorch]: Qnn backend library version 2.14.0 is mismatched. The minimum supported version is 2.15.0. Please make sure you have the correct backend library version.|

| User QNN HTP Version | Build QNN HTP Version | Log Output |
|--------------|---------------|------------|
| 5.24.0       | 5.22.0       | [WARNING] [Qnn ExecuTorch]: Backend kHtpBackend  library version 5.24.0 is used. The version is tested against 5.22.0.|
| 5.20.0       | 5.22.0      | [WARNING] [Qnn ExecuTorch]: Backend kHtpBackend library version 5.20.0 is mismatched. The minimum supported version is 5.22.0. Please make sure you have the correct backend library version.|

Author: yuyazhua

PR #15546, opened 2025-11-04
Move QNN SDK into backends/qualcomm/CMakeLists.txt

larryliu0820:
### Summary

### Test plan

### DIFF
--- a/backends/qualcomm/CMakeLists.txt
+++ b/backends/qualcomm/CMakeLists.txt
@@ -64,6 +64,13 @@ if(EXECUTORCH_BUILD_PYBIND)
     )
   endif()
   set(ENV{QNN_SDK_ROOT} ${QNN_SDK_ROOT})
+elseif(NOT DEFINED QNN_SDK_ROOT)
+  message(
+    FATAL_ERROR
+      "Please define QNN_SDK_ROOT, e.g. cmake <..> -DQNN_SDK_ROOT=<...>"
+  )
+elseif(CMAKE_TOOLCHAIN_FILE MATCHES ".*(iOS|ios\.toolchain)\.cmake$")
+  message(FATAL_ERROR "ios is not supported by Qualcomm AI Engine Direct")
 endif()
 
 message(STATUS "Using qnn sdk root ${QNN_SDK_ROOT}")
--- a/backends/qualcomm/CMakeLists.txt
+++ b/backends/qualcomm/CMakeLists.txt
@@ -23,6 +23,7 @@ get_filename_component(
   _common_include_directories "${EXECUTORCH_SOURCE_DIR}/.." ABSOLUTE
 )
 
+# We only download QNN SDK when we build pip wheel for ExecuTorch.
 if(EXECUTORCH_BUILD_PYBIND)
   set(_qnn_default_sdk_dir "${CMAKE_CURRENT_BINARY_DIR}/sdk/qnn")
 
--- a/backends/qualcomm/CMakeLists.txt
+++ b/backends/qualcomm/CMakeLists.txt
@@ -24,36 +24,44 @@ get_filename_component(
 )
 
 if(EXECUTORCH_BUILD_PYBIND)
-  find_package(
-    Python3
-    COMPONENTS Interpreter
-    REQUIRED
+  set(_qnn_default_sdk_dir "${CMAKE_CURRENT_BINARY_DIR}/sdk/qnn")
+
+  if(EXISTS "${_qnn_default_sdk_dir}" AND EXISTS
+                                          "${_qnn_default_sdk_dir}/lib"
   )
-  message(STATUS "QNN_SDK_ROOT not provided, downloading Qualcomm SDK")
-  execute_process(
-    COMMAND
-      ${Python3_EXECUTABLE}
-      ${EXECUTORCH_SOURCE_DIR}/backends/qualcomm/scripts/download_qnn_sdk.py
-      --print-sdk-path
-    WORKING_DIRECTORY ${EXECUTORCH_SOURCE_DIR}
-    RESULT_VARIABLE _qnn_sdk_download_result
-    OUTPUT_VARIABLE _qnn_sdk_download_output
-    ERROR_VARIABLE _qnn_sdk_download_error
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-  )
-  if(NOT _qnn_sdk_download_result EQUAL 0 OR _qnn_sdk_download_output STREQUAL
-                                             ""
-  )
-    message(
-      FATAL_ERROR
-        "Failed to download Qualcomm SDK. stdout: ${_qnn_sdk_download_output}\n"
-        "stderr: ${_qnn_sdk_download_error}"
+    message(STATUS "Found cached Qualcomm SDK at ${_qnn_default_sdk_dir}")
+    set(QNN_SDK_ROOT
+        ${_qnn_default_sdk_dir}
+        CACHE PATH "Qualcomm SDK root directory" FORCE
+    )
+  else()
+    message(STATUS "Downloading Qualcomm SDK")
+    execute_process(
+      COMMAND
+        ${Python3_EXECUTABLE}
+        ${EXECUTORCH_SOURCE_DIR}/backends/qualcomm/scripts/download_qnn_sdk.py
+        --dst-folder ${_qnn_default_sdk_dir}
+        --print-sdk-path
+      WORKING_DIRECTORY ${EXECUTORCH_SOURCE_DIR}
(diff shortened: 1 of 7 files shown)


cccclai:
Thanks! I added the binary ci job to make sure the wheel package publishing path is still working

cccclai:
Hi @haowhsu-quic, can you review the changes inside `backends/qualcomm`?

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,15 +23,66 @@ get_filename_component(
...
-  message(
-    FATAL_ERROR
-      "Please define QNN_SDK_ROOT, e.g. cmake <..> -DQNN_SDK_ROOT=<...>"
-  )
-elseif(CMAKE_TOOLCHAIN_FILE MATCHES ".*(iOS|ios\.toolchain)\.cmake$")
+if(CMAKE_TOOLCHAIN_FILE MATCHES ".*(iOS|ios\.toolchain)\.cmake$")
   message(FATAL_ERROR "ios is not supported by Qualcomm AI Engine Direct")
 endif()
 
+set(_qnn_host_is_linux_x86 FALSE)
```
shewu-quic: I think #L274 can be replaced with this flag.

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,15 +23,66 @@ get_filename_component(
...
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR MATCHES
+                                          "^(x86_64|amd64|i.86)$"
+)
+  set(_qnn_host_is_linux_x86 TRUE)
+endif()
+
+if(NOT DEFINED QNN_SDK_ROOT)
```
cccclai: I think the logic below is for wheel only, so probably better to have a separate env variable to control

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,15 +23,66 @@ get_filename_component(
...
+  if(_qnn_host_is_linux_x86)
+    set(_qnn_default_sdk_dir "${CMAKE_CURRENT_BINARY_DIR}/sdk/qnn")
```
cccclai: I feel like it will cause weird logic like pip install will have the sdk in pip site packages but then when we clone the executorch repro for runtime, we are going to download a seperate version inside executorch github. I'm not sure what the behavior will be

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,15 +23,66 @@ get_filename_component(
...
 
+set(_qnn_host_is_linux_x86 FALSE)
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR MATCHES
+                                          "^(x86_64|amd64|i.86)$"
+)
+  set(_qnn_host_is_linux_x86 TRUE)
+endif()
+
+if(NOT DEFINED QNN_SDK_ROOT)
+  if(_qnn_host_is_linux_x86)
```
shewu-quic: I’m not sure I understand the reason for this condition, as both Android and x86 builds require the QNN SDK.

larryliu0820: This is for pip wheels only. Like Chen said, maybe I should condition on that

backends/qualcomm/CMakeLists.txt
larryliu0820: > pip install will have the sdk in pip site packages 

No pip install won't package the sdk because the sdk is downloaded to build directory.

larryliu0820: > when we clone the executorch repro for runtime, we are going to download a seperate version inside executorch github.

How is the behavior different than before this PR?

backends/qualcomm/CMakeLists.txt
larryliu0820: @shewu-quic can you give some advice on how to move forward? The purpose of this PR is quite simple: to move the qnn build logic from `setup.py` to `backends/qualcomm/CMakeLists.txt`. Notice that the logic in `setup.py` is unique to wheel building and I'm trying to wrap my head on why building the runtime doesn't require downloading the sdk.

shewu-quic: Based on my understanding, you want to include libqnn_executorch_backend.so and the pybind files in the wheel build so users can install via pip without needing to set QNN_SDK_ROOT or download qnn sdk.

We could add a step to download the QNN SDK if QNN_SDK_ROOT is not set, similar to your approach, and use "if is_linux_x86():" in setup.py to avoid installation on unsupported systems for now. Does this approach work for you?

backends/qualcomm/CMakeLists.txt
shewu-quic: I believe, as Chen mentioned, we have a runtime check to verify if the qnn version of libqnn_executorch_backend.so is different from the qnn version used by the user during runtime.
https://github.com/pytorch/executorch/pull/4998

backends/qualcomm/CMakeLists.txt
larryliu0820: Thank you for the reply, I think that should work. Let me sync with @cccclai to have more understanding.

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,13 +23,46 @@ get_filename_component(
   _common_include_directories "${EXECUTORCH_SOURCE_DIR}/.." ABSOLUTE
 )
 
-if(NOT DEFINED QNN_SDK_ROOT)
-  message(
-    FATAL_ERROR
-      "Please define QNN_SDK_ROOT, e.g. cmake <..> -DQNN_SDK_ROOT=<...>"
+if(EXECUTORCH_BUILD_PYBIND)
```
larryliu0820: @cccclai what do you think about this

backends/qualcomm/CMakeLists.txt
larryliu0820: I think it makes sense to gate this change by `EXECUTORCH_BUILD_PYBIND`. After talking to @cccclai I think we agree that this code path should not be used by cases other than building ET wheel.

backends/qualcomm/CMakeLists.txt
cccclai: I think it should be wheel package only, not just pybind. I think `EXECUTORCH_BUILD_WHEEL_DEV_USE_ONLY` will be better or maybe @shewu-quic has a better idea

backends/qualcomm/CMakeLists.txt
```diff
@@ -23,7 +23,48 @@ get_filename_component(
...
+          "stderr: ${_qnn_sdk_download_error}"
+      )
+    endif()
+    set(QNN_SDK_ROOT
+        ${_qnn_sdk_download_output}
+        CACHE PATH "Qualcomm SDK root directory" FORCE
+    )
+  endif()
+  set(ENV{QNN_SDK_ROOT} ${QNN_SDK_ROOT})
+elseif(NOT DEFINED QNN_SDK_ROOT)
```
cccclai: Maybe just end if instead of else, because they are not in the same flow

backends/qualcomm/CMakeLists.txt
shewu-quic: I think EXECUTORCH_BUILD_WHEEL_DEV_USE_ONLY would be a better name as well.

larryliu0820: They are essentially the same thing. But if you insist sure

backends/qualcomm/CMakeLists.txt
cccclai: It's true that we have a version check. But tracking down where the malformed sdk path is quite annoying especially we're using the generic `LD_LIBRARY_PATH`. If possible, I'd prefer to have one sdk path in the system

backends/qualcomm/CMakeLists.txt
cccclai: They're not quite the same. I think my original proposal in the chat thread was to use the env variable `DO_NOT_USE_BUILD_FOR_WHEEL_ONLY` to ensure users won't touch this code at all. There are two purposes of the env variable: 1) guard the flow 2) users won't touch it. 

`EXECUTORCH_BUILD_WHEEL_DEV_USE_ONLY` is better than the original `EXECUTORCH_BUILD_PYBIND`, but I still want to have `DO_NOT_USE` prefix to ensure users even want to try it. Can you add this change as well? Afterwards, @shewu-quic can you review the changes?

shewu-quic (approved):
LGTM. Thanks!

cccclai (approved):
Thank you for making the changes

PR #15549, opened 2025-11-04
Arm backend: Initial support of conditional operator

Erik-Lundell:
- Add partition check to make sure that the submodules with
    the if/else codepaths are fully delegated.
    - Fix some partitioning issues with submodule nodes, since they
    point to a submodule rather than a tensor they dont have a fake tensor.
    - Add node visitor.
    - Add tests.
    
Arm backend: Use output.name in node visitors
    
    As mentioned in #15381, TOSA tensors need unique naming,
    which gets tricky with submodules. It is handled in
    the TosaArg object, and therefore node visitors need
    to use output.name rather than node.name when creating
    new tensors.


zingo:
Hi again @SS-JIA  here is a lager patch touching many file. Sorry we got so many in a row right now, it's usually calmer :)

zingo:
@SS-JIA we think this should be buck2 build safe as the new cond op should be handled by the 
srcs = glob(["op_*.py", "ops_*.py"]), rule.

SS-JIA (approved):
Should be okay!

zingo:
Failed tests seems unrelated

PR #15557, opened 2025-11-04
Arm backend: Fix test_lstm_arm.py after mypy changes

perheld:
Use keywords when calling pipeline.change_args()

Change-Id: Iba70b4c26eb1b96c16ee47382002cd8f7c36852b


zingo (approved):

zingo:
Fails seems unrelated

PR #15581, opened 2025-11-04
More overflow checks in memory allocator.

shoumikhin:
Summary:
- Safer overflow math for high alignments
- Only over-allocate up to alignment - 1 bytes when a higher-than-malloc alignment is requested
- Move EXECUTORCH_TRACK_ALLOCATION to after a successful allocation and use the final size

Differential Revision: D86228757

### DIFF
--- a/extension/memory_allocator/malloc_memory_allocator.h
+++ b/extension/memory_allocator/malloc_memory_allocator.h
@@ -10,6 +10,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <vector>
 
 #include <executorch/runtime/core/memory_allocator.h>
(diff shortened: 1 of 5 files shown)


runtime/core/memory_allocator.h
```diff
@@ -184,21 +196,18 @@ class MemoryAllocator {
   /**
    * Returns true if the value is an integer power of 2.
    */
-  static bool isPowerOf2(size_t value) {
-    return value > 0 && (value & ~(value - 1)) == value;
+  static constexpr bool isPowerOf2(size_t value) {
```
JacobSzwejbka, line 202: constexpr?

shoumikhin: Yeah, thought on using it in some compile time expressions like static assert, not strictly needed rn though, just nice to have just in case

extension/memory_allocator/malloc_memory_allocator.h
```diff
@@ -56,30 +55,29 @@ class MallocMemoryAllocator : public executorch::runtime::MemoryAllocator {
...
       // To get higher alignments, allocate extra and then align the returned
-      // pointer. This will waste an extra `alignment` bytes every time, but
+      // pointer. This will waste an extra `alignment - 1` bytes every time, but
       // this is the only portable way to get aligned memory from the heap.
-
-      // Check for overflow before adding alignment to size
-      if (size > SIZE_MAX - alignment) {
-        ET_LOG(
-            Error, "Size %zu + alignment %zu would overflow", size, alignment);
+      const size_t extra = alignment - 1;
```
JacobSzwejbka, line 60: oh nice

JacobSzwejbka (approved):

PR #15586, opened 2025-11-05
[BC Breaking] Remove runtime/core/portable_type/optional.h

cyyever:
### Summary
The usage of customised optional type has been replaced by `std::optional`.

cc @larryliu0820 @JacobSzwejbka @lucylq


JacobSzwejbka:
Hmm ET exposed optional as a public type, so to avoid being bc breaking you probably need to leave the header and alias the std::optional to the et namespaces

edit: Ah the header already was doing that. Ok then probably just leave it. Its probably safer

cyyever:
This is BC breaking, so it is up to the community to decide when to apply.

lucylq:
Importing to internal as well to make sure we're not breaking any internal tests

PR #15589, opened 2025-11-05
Arm backend: Fix bug in ConvertExpandToRepeat pass

Erik-Lundell:
The pass assumed that if all repeat multiples are one, the op is a no-op. However, it can still change the rank.


zingo (approved):

zingo:
Fails unrelated, buck2 build should not need to be changed

PR #15590, opened 2025-11-05
Cortex_m backend: Add IO quantizers + tests of non rescaling ops

AdrianLundell:
A number of ops only handles shape/meta-data without changing the dynamic range. In these cases, no rescaling needs to be performed and the int8 portable_ops kernel can be used directly.

A new test is added to ensure this behaviour, as well as a test showing how operators which does change the dynamic range (SUB) are not supported.

To support quantization of graphs with no-rescale ops in the beginning/ end of the graph, two new quantizers InputQuantizer and OutputQuantizer are introduced. By explicitly stating the dtpye of the input/output, no-rescale ops inherit dtypes from them as with any other op.

### DIFF
--- a/backends/cortex_m/quantizer/quantizer.py
+++ b/backends/cortex_m/quantizer/quantizer.py
@@ -16,6 +16,9 @@ from executorch.backends.cortex_m.quantizer.operator_configs import (
     INT8_BINARY_OPS_OPERATOR_CONFIG,
     INT8_LINEAR_OPERATOR_CONFIG,
 )
+from executorch.backends.cortex_m.quantizer.quantization_configs import (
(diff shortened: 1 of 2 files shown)


psiddh:
Is this the idea in general ?

// Current state:
input(float32) → ops → output(float32)

// Annotate **(This PR):**
input(float32) 📋"Annotate Q!" → ops → 📋"Annotate DQ!" output(float32)

// What later passes will do:
input(float32) → [Q] → ops(int8) → [DQ] → output(float32)
                 ↑---------------------------------↑
            "OK, inserting  ------------- ---------"OK, inserting
             quantize!" -------------------------- "dequantize!"

psiddh (approved):

PR #15591, opened 2025-11-05
Cortex_m backend: Add mul operator

AdrianLundell:


backends/cortex_m/ops/op_quantized_mul.cpp
```diff
@@ -0,0 +1,87 @@
...
+      output_multiplier,
+      output_shift,
+      out);
+
+  // Extract quantization parameters
+  const int32_t zp1 = extractScalarToInt32(input1_zero_point);
+  const int32_t zp2 = extractScalarToInt32(input2_zero_point);
+  const int32_t out_zp = extractScalarToInt32(output_zero_point);
+  const int32_t output_mult = extractScalarToInt32(output_multiplier);
+  const int32_t output_shift_val = extractScalarToInt(output_shift);
```
psiddh: Should this also be extractScalarToInt32 ?

backends/cortex_m/ops/op_quantized_mul.cpp
```diff
@@ -0,0 +1,87 @@
...
+
```
psiddh, line 74: No Broadcasting support is needed ?

backends/cortex_m/ops/op_quantized_mul.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // Call CMSIS-NN elementwise multiply kernel
+  arm_cmsis_nn_status status = arm_elementwise_mul_s8(
+      input1_int8.const_data_ptr<int8_t>(),
+      input2_int8.const_data_ptr<int8_t>(),
+      -static_cast<int32_t>(zp1),
```
psiddh, line 79: the negative cast seems little surprising,  If it is indeed needed, can you add comments why CMSIS api expect negative offsets

backends/cortex_m/ops/op_quantized_mul.cpp
```diff
@@ -0,0 +1,87 @@
...
+      -static_cast<int32_t>(zp2),
+      out.mutable_data_ptr<int8_t>(),
```
psiddh, line 81: General question, I have, does it have to be contiguous memory ? If so, do we have to validate it ?

backends/cortex_m/ops/operators.py
```diff
@@ -168,6 +174,66 @@ def quantized_add_impl(
...
+    self: torch.Tensor,
+    self_zero_point: int,
+    other: torch.Tensor,
+    other_zero_point: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    # Broadcast to output shape
+    assert (
```
psiddh: nits: assert self.shape == other.shape, (
    "Cortex-M quantized_mul: broadcasting is not yet supported — "
    f"got self.shape={self.shape}, other.shape={other.shape}"
)   makes debugging better

backends/cortex_m/ops/operators.py
```diff
@@ -168,6 +174,66 @@ def quantized_add_impl(
...
+    other: torch.Tensor,
+    other_zero_point: int,
+    output_zero_point: int,
+    output_multiplier: int,
+    output_shift: int,
+) -> torch.Tensor:
+    # CMSIS-NN kernel multiplies raw int8 tensors (after zero-point offset) and
+    # only uses the output multiplier/shift for rescaling. Mirror that here to
+    # keep the composite implementation numerically aligned with the backend.
+    assert (
```
psiddh: same nits as above

backends/cortex_m/ops/operators.py
```diff
@@ -168,6 +174,66 @@ def quantized_add_impl(
...
+        self.shape == other.shape
+    ), "Broadcasting not yet supported in Cortex-M backend."
+    self_int = self.to(torch.int32) - self_zero_point
+    other_int = other.to(torch.int32) - other_zero_point
+    result_fp = self_int * other_int
```
psiddh, line 235: Just curious ,
If add_impl uses SHIFT_INT8, why add requires it and not mul , Maybe we should add that difference explicitly in comments.

backends/cortex_m/ops/op_quantized_mul.cpp
AdrianLundell: Broadcasting is not supported natively in CMSIS-NN similar as with the add operator, if it is required for an important model we might have to implement it then but I'm leaving it for now.

backends/cortex_m/ops/op_quantized_mul.cpp
AdrianLundell: It has to be in contiguous memory but we have the same req in the arm backend and we have never had a requirement for it so I do not believe we have to validate it. If it comes up we can look into it later.

backends/cortex_m/test/ops/test_mul.py
```diff
@@ -91,11 +90,11 @@ class CortexMTensorMul(Model):
     ),
```
psiddh, line 90: Suggest adding boundary-value tests for quantized_mul in a follow-up PR to act as guardrails against off-by-one, clamping, sign, and requantization errors on extreme int8 inputs.

psiddh (approved):

backends/cortex_m/test/ops/test_mul.py
AdrianLundell: Good idea!

AdrianLundell:
Fails not related

PR #15592, opened 2025-11-05
Arm backend: Cherry-pick tosa patch to speed up large models

ArmRyan:
* Only affects models >2gb

Change-Id: Id5ec79eecc4f1ae8d864e35ed13b34bbf761d30b

### DIFF
--- a/backends/arm/tosa/backend.py
+++ b/backends/arm/tosa/backend.py
@@ -133,7 +133,14 @@ class TOSABackend(BackendDetails):
         if not artifact_path:
             artifact_path = ""
 
-        tosa_graph = ts.TosaSerializer(artifact_path)
+        version = tosa_spec.version
+        tosa_graph = ts.TosaSerializer(
+            artifact_path,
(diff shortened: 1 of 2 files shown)


zingo:
@SS-JIA here is a change bumping TOSA slightly that you might also want to test, we think it will work if you keep the old version internally if you prefer, this change should only be needed on larger models that are not yet in the tests flow.

ArmRyan:
> @SS-JIA here is a change bumping TOSA slighly that you might also wen to test, we also think to will work if you keep the old version internally, this change should only be needed on larger models that are not yet in the tests flow.

This is just a 2line change in an existing function it doesnt add anything new or change any behaviour, it just resereves memory correctly so nothing changes functionally and no new files or structure changes - wont affect anything

YufengShi-dudu (approved):
It has been tested internally and shows a significant speedup on large models.

SS-JIA (approved):

PR #15594, opened 2025-11-05
Arm backend: Fix mypy warnings in test/models

perheld:
Add ignores for third-party modules and change import common into using the correct path.

Change-Id: I0b1cfa4c5b7a1625b72c6cfa59da8f427a25a204


zingo (approved):

zingo:
fails unrelated, only changes within files so buck2 build files should not need any update.

PR #15595, opened 2025-11-05
Arm backend: Upcast index argument to int64 for aten.index_copy ops

YufengShi-dudu:
Change-Id: I01dbb2b69c5689dbc8b5b56534996eb7d74828ff


ArmRyan (approved):
Approved and reviewed internally

zingo (approved):

zingo:
Fails are unrelated, only editing existing files so buck2 builds files should not need to be changed.

PR #12707, opened 2025-07-22
Use std::optional

cyyever:
### Summary
Move remaining usage of optional to `std::optional`.

PR #15612, opened 2025-11-05
Revert "Use std::optional"

metascroy:
Reverts pytorch/executorch#12707

This causes 1000s of internal CI failures, e.g.,

error: no template named 'optional' in namespace 'torch::executor'; did you mean simply 'optional'?
using optional = torch::executor::optional<T>;
                 ^~~~~~~~~~~~~~~~~~~~~~~~~
                 optional
note: 'optional' declared here
using ::executorch::aten::optional;


JacobSzwejbka (approved):

JacobSzwejbka:
cc @cyyever I think the root cause of the failures is what I mentioned on your follow up diff of we cant just remove the name "torch::executor::optional" you will need to leave that exposed as an alias for std::optional

larryliu0820:
> cc @cyyever I think the root cause of the failures is what I mentioned on your follow up diff of we cant just remove the name "torch::executor::optional" you will need to leave that exposed as an alias for std::optional

I think you just need to add this line:

```
#include <executorch/runtime/core/portable_type/optional.h> // @manual
```
To runtime/core/exec_aten/exec_aten.h

cyyever:
@JacobSzwejbka Is it better to migrate internal tests to std::optional before changing OSS?

PR #15629, opened 2025-11-06
Add src to mypy_path in .mypy.ini

perheld:
Add src to mypy_path to get rid of lots of missing import warnings.

Change-Id: I4ed1a08cfe185aa956c6408fefa5aadeb567d4c1

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -24,7 +24,7 @@ files =
     test,
     util
 
-mypy_path = executorch
+mypy_path = executorch,src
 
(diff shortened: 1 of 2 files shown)


zingo:
@mergennachin this helps a a lot when checking with mypy is it OK to do?
The src folders seem to be a bit of a work in progress.

perheld:
lintrunner-mypy is not running for this PR because no python file is changed. Is there some clever way I can force run it?

PR #14770, opened 2025-10-03
Arm backend: Enable int16 rsqrt on Ethos-U55/U85

3l1:
Summary:
Fix Rsqrt op for int16
Add unit tests

Differential Revision: D83802158

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai

PR #15631, opened 2025-11-06
Arm backend: Add tests for int16 rsqrt on Ethos-U55/U85

Ninja91:
Summary:
Fix Rsqrt op for int16
Relands D83802158

Differential Revision: D86402524

### DIFF
--- a/backends/arm/test/ops/test_rsqrt.py
+++ b/backends/arm/test/ops/test_rsqrt.py
@@ -8,9 +8,14 @@
 
 from typing import Tuple
 
+import pytest
 import torch
+from executorch.backends.arm.quantizer.arm_quantizer import (
+    get_symmetric_a16w8_quantization_config,
+    TOSAQuantizer,
+)
+from executorch.backends.arm.test import common, conftest
 
-from executorch.backends.arm.test import common
 from executorch.backends.arm.test.tester.test_pipeline import (
     EthosU55PipelineINT,
     EthosU85PipelineINT,
@@ -18,7 +23,8 @@ from executorch.backends.arm.test.tester.test_pipeline import (
     TosaPipelineINT,
     VgfPipeline,
 )
-
+from executorch.backends.arm.tosa import TosaSpecification
+from executorch.backends.xnnpack.test.tester import Quantize
 
 aten_op = "torch.ops.aten.rsqrt.default"
 input_t1 = Tuple[torch.Tensor]  # Input x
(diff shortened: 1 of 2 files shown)


Ninja91:
test-lama-runner-mac test seemes unrrelated.
cc: @zingo  @digantdesai  @SaoirseARM

backends/arm/test/ops/test_rsqrt.py
```diff
@@ -104,3 +110,96 @@ def test_rsqrt_vgf_INT(test_tensor: torch.Tensor):
...
+    return Quantize(
+        quantizer,
+        get_symmetric_a16w8_quantization_config(
+            is_per_channel=per_channel_quantization
+        ),
+    )
+
+
+@common.parametrize("test_tensor", Rsqrt.test_parameters)
+def test_rsqrt_int16_tosa_INT(test_tensor: torch.Tensor):
```
SaoirseARM: It looks like there are some failures on this particular testcase. For rand_4d_1 and rand _4d_2. It might be best to see these as expected failures like below. Thanks for adding this!

zingo:
Hi @Ninja91 this is the fails we see,

[pull / unittest-arm-backend-with-no-fvp (test_pytest_ops) / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/19161380882/job/54772221446?pr=15631)
FAILED backends/arm/test/ops/test_rsqrt.py::test_rsqrt_int16_tosa_INT[rand_4d_1]
FAILED backends/arm/test/ops/test_rsqrt.py::test_rsqrt_int16_tosa_INT[rand_4d_2]
FAILED backends/arm/test/ops/test_rsqrt.py::test_rsqrt_int16_tosa_INT[rand_3d]

 if you can break out the fails in it's own tests you can add a tag 
@pytest.mark.xfail( reason="text" )
for now, or just commen them out.

Also we see https://github.com/pytorch/executorch/pull/14770  should that be closed?

backends/arm/test/ops/test_rsqrt.py
zingo: Also test name should follow this naming convention   `test_<op>_16a8w_<target>`

backends/arm/test/ops/test_rsqrt.py
```diff
@@ -104,3 +110,96 @@ def test_rsqrt_vgf_INT(test_tensor: torch.Tensor):
...
+    # Run the pipeline
+    pipeline.run()
+
+
+@common.parametrize("test_tensor", Rsqrt.test_parameters)
+@common.XfailIfNoCorstone300
+@pytest.mark.xfail(
+    reason="MLETORCH-707: AssertionError: Output 0 does not match reference output."
+)
+def test_rsqrt_int16_u55_INT16(test_tensor: torch.Tensor):
```
zingo: Test name should follow this naming convention   `test_<op>_16a8w_<target>`

backends/arm/test/ops/test_rsqrt.py
```diff
@@ -104,3 +110,96 @@ def test_rsqrt_vgf_INT(test_tensor: torch.Tensor):
...
+    )
+    pipeline.run()
+
+
+@common.parametrize("test_tensor", Rsqrt.test_parameters)
+@common.XfailIfNoCorstone320
+@pytest.mark.xfail(
+    reason="MLETORCH-707: AssertionError: Output 0 does not match reference output."
+)
+def test_rsqrt_int16_u85_INT16(test_tensor: torch.Tensor):
```
zingo: Test name should follow this naming convention   `test_<op>_16a8w_<target>`

zingo (changes requested):

backends/arm/test/ops/test_rsqrt.py
Ninja91: Updated. Thanks for clarifying.

backends/arm/test/ops/test_rsqrt.py
gggekov: Regarding the numerical difference between PyTorch eager mode & TOSA reference model, one thing I would add is that we use a LUT to perform rsqrt in 8x8 and 16x8. 

For the 8x8 case, the HW has a LUT that has 256 entries so we have a 1:1 mapping between quantized value and what gets written to the LUT of the HW. And for 8x8, we can pass the unit tests between TOSA ref model & PyTorch eager mode even with atol=qtol=0.

However, for the 16x8 case, the HW doesn't have a LUT with 65536 entries as that would be too impractical/too costly from HW perspective. Instead, we do an interpolation algorithm as explained in https://www.mlplatform.org/tosa/tosa_spec_1_0_0.html#_table where you still get high precision from the 16-bit operation, but we can't be bit-accurate. Agree with @SaoirseARM we need to mark the tests as expected failures and probably relax the atol a bit to pass. 

I agree with @zingo 's comments about using the same naming convention as the rest of the code.

zingo:
Thanks for the update, lets see if CI passes now. 

Maybe we want/need to carve out just the failed tests and only mark those but lets see, we should probably add atol/rtol instead and remove xfails long term anyway so if CI is happy I'm happy :)

zingo:
@Ninja91 I see 
[Meta Internal-Only Changes Check](https://github.com/pytorch/executorch/pull/15631/checks?check_run_id=54813544132) — The Diff and Pull Request are not in sync!

It stops me from merging, I don't know if it's bacause I edited the commit heade adding "Arm backend: "  Maybe you need to reimport it or something else? Sorry about that.

backends/arm/test/ops/test_rsqrt.py
zingo: Seem to work now, lets just tests run to end. marking resloved.

zingo (approved):

limintang (approved):

zingo:
Arm Backend Fail 
FAILED backends/arm/test/ops/test_sub.py::test_sub_tensor_tosa_INT_alpha[rand_4D_pi]
is unrelated and has a error rate on about 2-3% normaly we should retrigger to prove it but due to this is needed to be merged Ill just state it here this time.
Rest of the fails are also unrelated

backends/arm/test/ops/test_rsqrt.py
Ninja91: @gggekov  Didn't follow:
> we can pass the unit tests between TOSA ref model & PyTorch eager mode even with atol=qtol=0.

Do we mean we don't expect 8x8 useful from numerics perspective?

PR #15632, opened 2025-11-06
Arm backend: Backend flow should not share compile spec object

zingo:
### Summary
Delay compile-spec creation in the backend test flow to prevent sharing the temp directory between tests.

Previously, using a shared compile spec implied a shared temp directory. After we began cleaning the temp directory after each test, this sharing caused conflicts.

### Test plan
This is tested by the Backend test flow

### DIFF
--- a/backends/test/suite/flows/arm.py
+++ b/backends/test/suite/flows/arm.py
@@ -5,19 +5,22 @@
 
 # Create flows for Arm Backends used to test operator and model suits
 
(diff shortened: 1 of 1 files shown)


zingo:
Test fails unrelated

Sebastian-Larsson (approved):

zingo:
Fails unrelated, no buck2 changes should be needed for this.

PR #15633, opened 2025-11-06
Dev tools: Fix filename in model explorer install suggestion

zingo:
### Summary
Fix the filename in the log to match the file.

### Test plan
Tested by hand


cccclai (approved):
Thanks for the fix!

PR #15635, opened 2025-11-06
Arm backend: Lower more int8/int16 permutations for Ethos-U55

AdrianLundell:
- Add (0,3,1,2) and (0,2,3,1) as permutations supported for large shapes.
- Lower permutations expressable as views ('singleton permutations') to views to allow them to run on the Ethos-U55.

All unittests added were previosuly not lowered which leads for example to 19 permutes delegated on the
convnext_tiny model from torchvision.


zingo (approved):

zingo:
fails are unrelated, added file in _passes should be covered by the *.py glob for buck2 so no need to change any buck2 files.

PR #15611, opened 2025-11-05
[Executorch] Introduce caching cpu memory allocator

kimishpatel:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #15730
* #15729
* #15728
* __->__ #15611
* #15610

Meant to use this for temp allocator for kernels. Specifically for sdpa, it seems that on iOS there is a significant overhead coming from allocations

Differential Revision: [D85532079](https://our.internmc.facebook.com/intern/diff/D85532079/)

PR #15653, opened 2025-11-06
[Executorch][llm] Fix flakyness of quantized sdpa test

kimishpatel:
* #15730
* #15729
* #15728
* #15611
* #15610
* #15609
* #15608
* #15607
* __->__ #15653

Without this fix when running diabled tets it fails with
OMP: Error #131: Thread identifier invalid

Differential Revision: [D86447540](https://our.internmc.facebook.com/intern/diff/D86447540/)


extension/llm/custom_ops/test_quantized_sdpa.py
```diff
@@ -40,6 +41,11 @@ def setUp(self):
         self.q_shape = None
         self.kv_shape = None
         self.is_seq_at_dim_2 = True
+        # For some reason 4 threads doesnt work
```
metascroy, line 44: Is there a concern with users choose a thread count that doesn't work?

kimishpatel: I think this is specifically related to OMP which I am definitely a bit surprised why it occurs here. This hasnt been observed on-device

metascroy (approved):

PR #15667, opened 2025-11-07
Manually cherry-pick "Pin huggingface_hub<1.0"

GregoryComer:
### Summary
This is a manual cherry-pick of https://github.com/pytorch/executorch/pull/15399 and https://github.com/pytorch/executorch/pull/15403 to resolve CI failures caused by an incompatibility between the newly released huggingface_hub package and our pinned version of transformers. Due to differences between main and release/1.0, the automated cherry-pick failed, so I've reverted to a manual pick and some small manual edits to examples/models/llama/install_requirements.sh.

### Test Plan
This PR is running trunk CI.

### DIFF
--- a/.ci/scripts/test_wheel_package_qnn.sh
+++ b/.ci/scripts/test_wheel_package_qnn.sh
@@ -136,7 +136,7 @@ run_core_tests () {
   local PIPBIN="$2"     # path to pip
   local LABEL="$3"      # label to print (conda/venv)
 
- 
+  echo "=== [$LABEL] Installing wheel & deps ==="
   "$PIPBIN" install --upgrade pip
   "$PIPBIN" install "$WHEEL_FILE"
   "$PIPBIN" install torch=="2.9.1"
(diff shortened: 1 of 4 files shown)


GregoryComer:
Note to reviewers: there are more failing/flaky jobs. These are pre-existing and I will fix as a follow-up, as the cause is unrelated to this PR.

.ci/scripts/test_wheel_package_qnn.sh
```diff
@@ -139,7 +139,7 @@ run_core_tests () {
   echo "=== [$LABEL] Installing wheel & deps ==="
   "$PIPBIN" install --upgrade pip
   "$PIPBIN" install "$WHEEL_FILE"
-  "$PIPBIN" install torch=="2.9.0"
+  "$PIPBIN" install torch=="2.9.1"
```
lucylq, line 142: is this intended?

lucylq (approved):

.ci/scripts/test_wheel_package_qnn.sh
GregoryComer: Yeah - good question. It's not strictly part of the original pick but we want torch==2.9.1 for 1.0.1 and I was seeing a related issue on CI, so I stuck that fix in.

PR #15672, opened 2025-11-07
Arm backend: Enable mypy for tests

perheld:
Add test directory in arm backend to lintrunner. Except the ops directory which has to be fixed later.

About "disallow_untyped_decorators = False":
The only way to make mypy happy seems to be a wrapper around the decorator for each file it is used. I have read numerous of threads about it and there seems to be no good solution. Hence the added ignore for backends/arm.

Change-Id: I0430041cb3308b51798d74da61cc4d310966772a

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -367,7 +367,7 @@ exclude_patterns = [
     '**/third-party/**',
     'scripts/check_binary_dependencies.py',
     'profiler/test/test_profiler_e2e.py',
-    'backends/arm/test/**',
+    'backends/arm/test/ops/*.py',
 ]
 command = [
     'python',
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -24,7 +24,7 @@ files =
     test,
     util
 
-mypy_path = executorch
+mypy_path = executorch,src
 
 [mypy-executorch.backends.*]
 follow_untyped_imports = True
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -29,6 +29,9 @@ mypy_path = executorch,src
 [mypy-executorch.backends.*]
 follow_untyped_imports = True
 
+[mypy-backends.arm.*]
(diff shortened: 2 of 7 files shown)


perheld:
This has  to be the last one to go in after:
https://github.com/pytorch/executorch/pull/15629
https://github.com/pytorch/executorch/pull/15636
https://github.com/pytorch/executorch/pull/15671

And possibly others if there are more commits going in that adds warnings. :)

perheld:
I added all needes commits to this PR to make life easier.

zingo:
@mergennachin and/or @SS-JIA there is one line in this PR that is not 100% Arm specific, it adds the "new" src folder to .mypy.ini  this clean up some lintrunner problem as project starting to use it also.

 e.g.
mypy_path = executorch,src

I'm not sure if this is something you would like to test internally or not, we let this PR be up for some days so you get a chance to weight in. As there is a inflow of PR that might breaks this as it not enabled until this lands, we will need to rebase and maybe add more lintrunner fixes in this PR before merging and we hope/plan to do this "Europe" time so can test it and retrigger some random PR to make sure there is no problem and do a fast fix or revert while less people is active in the project.

perheld:
Fix incoming for the actual lint error. :)

perheld:
> Fix incoming for the actual lint error. :)

And the lintrunner-mypy got green with that commit. So Im confident it is running the right thing.

SS-JIA (approved):
Should be ok, MYPY/lintrunner is not run internally

zingo:
Thanks

zingo:
Ill rebase and check that lintrunner is still happy :) 
... I manage to merge a conflict just before :) Well we check it next week and fix it more calmly.

perheld:
Failed are unrelated. Ive rebased locally and it still passes lintrunner.

PR #15684, opened 2025-11-07
Error out if git symlink not enabled on Windows

larryliu0820:
When I install executorch on windows I run into this error:   error:
[Errno 2] No such file or directory:
'pip-out\\lib.win-amd64-cpython-312\\executorch\\version.py'

This is because we are not enabling symlink in git.

Need to run:

```
 git config --global core.symlinks true
```

and reclone the repo.

This PR fails the installation when that happens.

### DIFF
--- a/install_executorch.bat
+++ b/install_executorch.bat
@@ -1,4 +1,5 @@
 @ECHO OFF
+setlocal EnableDelayedExpansion
 
 rem Copyright (c) Meta Platforms, Inc. and affiliates.
 rem All rights reserved.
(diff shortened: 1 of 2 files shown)


GregoryComer (approved):

mergennachin:
@GregoryComer @larryliu0820 cherry-pickable to 1.0.1?

larryliu0820:
Oops this patch doesn't fix the underlying problem. If we don't turn on symlink on windows: `git config --global core.symlinks true`, then the wheel doesn't have all the python files such as exir/backend/:

```
PS C:\executorch> tar -tf ".\dist\executorch-1.1.0a0+a6b20e6-cp312-cp312-win_amd64.whl" | Select-String exir

executorch/exir/_serialize/program.fbs
executorch/exir/_serialize/scalar_type.fbs
```

Probably will abandon this.

GregoryComer:
> @GregoryComer @larryliu0820 cherry-pickable to 1.0.1?

Following up on Mengwei's comment, I had a task for 1.0 to at least catch this any maybe try to automatically fix the symlinks. I dropped it from release blockers due to priorities, but I'll pull it into 1.0.1.

larryliu0820:
I updated the PR to error out in `install_executorch.bat` when symlink is not enabled on the environment.

GregoryComer (approved):
Thanks for adding this. Looks good to me.

PR #15685, opened 2025-11-07
Metal backend: Add Whisper to CI workflow

manuelcandales:
This pull request refactors the model export and e2e scripts to support both CUDA and Metal backends, and updates the Metal CI workflow to generalize model export and testing for multiple models and quantization options. It expands CI model coverage to also include Whisper


Gasoonjia (approved):
LGTM!

larryliu0820 (approved):

mergennachin:
This is great, @manuelcandales 

We should also update https://github.com/pytorch/executorch/blob/main/examples/models/whisper/README.md

PR #15689, opened 2025-11-08
Qualcomm AI Engine Direct - supprot SA8255

TadayukiOkada:
### Summary

- add soc SA8255

### Test plan
`python -m examples.qualcomm.scripts.deeplab_v3 -b build-android -m SA8255 --compile_only --download`

push necessary files on the target
```
export BASEDIR=/data/local/tmp/et/
adb shell "cd ${BASEDIR} \
    && export ADSP_LIBRARY_PATH=\"${BASEDIR};/vendor/lib/rfsa/adsp;/vendor/lib64;\" \
    && export LD_LIBRARY_PATH=${BASEDIR}:/vendor/lib64/ \
    && ./qnn_executor_runner --model_path ./dl3_qnn_q8.pte"
```


backends/qualcomm/serialization/qc_compiler_spec.fbs
```diff
@@ -37,6 +37,7 @@ enum QcomChipset: int {
   SM8450 = 36,
   SM8475 = 42,
   SM8550 = 43,
+  SA8255 = 52,
```
cccclai: I feel like we need to append at the end to avoid BC breacking. Thoughts from QCOM team?

cccclai (commented):
Thank you for contributing!

backends/qualcomm/serialization/qc_compiler_spec.fbs
shewu-quic: Yes, I think that makes sense.

backends/qualcomm/serialization/qc_schema.py
```diff
@@ -43,6 +43,7 @@ class QcomChipset(IntEnum):
     SM8450 = 36  # v69
     SM8475 = 42  # v69
     SM8550 = 43  # v73
+    SA8255 = 52  # v73
```
shewu-quic: Could you please append at the end? Thanks

TadayukiOkada: I'm fine with changing the order, but I believe my change is still backward-compatible since I'm only adding a new enum name with a previously unused value. Anyway, if you prefer, I can append it at the end instead.

TadayukiOkada: @shewu-quic I moved the SA8255 entry at the end.

shewu-quic (approved):
LGTM. Thanks for your contribution.

cccclai (approved):

PR #15691, opened 2025-11-10
Qualcomm AI Engine Direct - GLM1.5B

winskuo-quic:
### Summary

GLM Enablement
`python examples/qualcomm/oss_scripts/llama/llama.py -b build-android -s $DEVICE -m SM8750 --temperature 0 --model_mode kv --max_seq_len 128  --decoder_model glm-1_5b --prompt "Could you tell me about Facebook?"`

### Test plan
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleLLMScript.test_static_glm1_5b --model SM8750 --build_folder build-android/ --executorch_root . -s $DEVICE --artifact ./glm1_5b`

### DIFF
--- a/examples/models/glm/config/1_5b_config.json
+++ b/examples/models/glm/config/1_5b_config.json
@@ -1,0 +1,17 @@
+{
+  "dim": 2048,
+  "ffn_dim_multiplier": 1,
+  "hidden_dim": 6144,
+  "n_heads": 16,
+  "head_dim": 128,
+  "n_kv_heads": 4,
+  "n_layers": 28,
+  "norm_eps": 1e-05,
+  "rope_theta": 10000.0,
+  "use_scaled_rope": false,
+  "vocab_size": 59264,
+  "use_hf_rope": true,
+  "attention_qkv_bias": false,
+  "use_qk_norm": false,
+  "model_architecture" : "GlmForCausalLM"
+}
--- a/examples/models/llama/model_args.py
+++ b/examples/models/llama/model_args.py
@@ -131,7 +131,9 @@ class ModelArgs:
     attention_kwargs: Dict[str, Any] = dataclasses.field(default_factory=dict)
     # Hybrid models can have layer types different from attention
     layer_types: Optional[list] = None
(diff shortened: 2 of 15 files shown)


cccclai (approved):
Looks good, can you fix the lint?

examples/models/llama/model_args.py
```diff
@@ -122,6 +122,7 @@ class ModelArgs:
     attention_kwargs: Dict[str, Any] = dataclasses.field(default_factory=dict)
     # Hybrid models can have layer types different from attention
     layer_types: Optional[list] = None
+    model_architecture: Optional[str] = None
```
cccclai: Can you add comments to explain this variable?

examples/models/glm/config/1_5b_config.json
```diff
@@ -0,0 +1,17 @@
...
+  "n_kv_heads": 4,
+  "n_layers": 28,
+  "norm_eps": 1e-05,
+  "rope_theta": 10000.0,
+  "use_scaled_rope": false,
+  "vocab_size": 59264,
+  "use_hf_rope": true,
+  "attention_qkv_bias": false,
+  "use_qk_norm": false,
+  "model_architecture" : "GlmForCausalLM"
```
cccclai, line 16: Do we have any existing variable that can be used for this?

examples/models/llama/model_args.py
winskuo-quic: Added. Thanks

examples/models/glm/config/1_5b_config.json
winskuo-quic: Thanks for the suggestion.
I was actually thinking of reusing `base_model_name_or_path`. However, it seems like this variable is used in optimum for some other purpose, like referring to actual model path, so I created a new variable to prevent any conflict in future.
Another reason of creating this config is that as we are enabling more models, we noticed minor differences among models. For example, GLM FeedForward is different from other model's FeedForward. We need some variables to differentiate GLM and other LLM models.

PR #15694, opened 2025-11-10
Arm backend: Fix decomposition logic for FP

Erik-Lundell:
If the tosa_spec supports floating point, incorrect quantization is not a problem, and we can always
not decompose the ops_to_not_decompose.


Erik-Lundell:
Unrelated failure, should be buck safe.

oscarandersson8218 (approved):

PR #15695, opened 2025-11-10
Arm backend: Improve support for multiple output ops

Erik-Lundell:
The pattern of using .name from the TosaArg output is used in essentially all node visitors. Refactoring output to possibly be a list of TosaArgs would be very cumbersome. Instead, use a new multiple_output_names field to provide the names of all getitem tensors. This also means that we don't have to insert identity ops for getitem operators, they just need to insert a tensor.


Erik-Lundell:
Unrelated failure, should be buck safe

zingo (approved):

zingo:
fails are unrelated

PR #15696, opened 2025-11-10
Qualcomm AI Engine Direct - VIT Optimization

winskuo-quic:
### Summary
QNN doesn't have much 5D permute optimization, which causes Vit running slower than CPU.
image
Switched pattern from unsqueeze->permute 5d->squeeze to permute 4d.

Improvements: 150ms/inference -> 4.2ms/inference.

### Test plan
Pass Vit UT

### DIFF
--- a/examples/qualcomm/scripts/torchvision_vit.py
+++ b/examples/qualcomm/scripts/torchvision_vit.py
@@ -7,12 +7,14 @@
 import json
 import logging
 import os
+from contextlib import contextmanager
 
 from multiprocessing.connection import Client
 
 import numpy as np
 
 import torch
+import torch.nn.functional as F
 from executorch.backends.qualcomm.quantizer.quantizer import QuantDtype
 from executorch.examples.models.torchvision_vit.model import TorchVisionViTModel
 from executorch.examples.qualcomm.utils import (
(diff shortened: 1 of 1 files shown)


winskuo-quic:
Hi @cccclai,
This PR is to address the issue you have reported about QNN running slower than CPU for Vit.
I have tested SM8750 with this PR, and inference speed improves from **140ms/inference -> 4.1ms/inference**, which should be around **35 times** faster.
I have made a source level transformation on the Vit model.
The bottleneck is caused by the 5D permute pattern here: https://github.com/pytorch/pytorch/blob/3cfbf98ea9d937d23f3700168b22706c957308ce/torch/nn/functional.py#L5825.
I did source level transformation because this pattern is a bit specific and it is hard to generalize.
Please have a look and let me know if you have any questions.
Thanks

examples/qualcomm/scripts/torchvision_vit.py
```diff
@@ -25,6 +27,56 @@
 )
 
 
+# Copied from torch/nn/functional.py
+# QNN does not have 5D permute optimization. Fuse to a single 4D optimization
+# Changed unsqueeze(0).transpose(0, -2).squeeze(-2) to permute(2, 0, 1, 3)
```
cccclai, line 32: It can be a pass ideally, is it correct?

cccclai (approved):
oh just saw that you mentioned the pattern is a bit hard to generalize hence we're doing source transform here

examples/qualcomm/scripts/torchvision_vit.py
winskuo-quic: Yes. This can be a pass. If this pattern occurs frequently in the future, we will see if there's a better way to do this.

PR #15697, opened 2025-11-10
Arm backend: Return false on non-VK_SUCCESS in process_vgf

YufengShi-dudu:
Replace incorrect `return result;` (VkResult) in a bool path to avoid false success.

Change-Id: I842ffbed90f3f511695fc06820c167246f92e13a


ArmRyan (approved):
Approved in internal review

PR #15698, opened 2025-11-10
Arm backend: Fix incorrect qparams propagation

oscarandersson8218:
For some passes after q-dq-folding, metadata was naively copied to all new nodes in decomposition. This can lead to problems if these nodes were to be computed ahead of time. This is solved by introducing a new optional argument to create_node, called inherit_qparams that defaults to false. This means that you explicitly need to set it to True in order to inherit qparams from the original node. For call_operator passes, the metadata dict is copied and modified in each pass. The patch also removes a few duplicate passes in ArmPassManager.


zingo:
Testfail with rsub :( 
[pull / unittest-arm-backend-with-no-fvp (test_pytest_ops) / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/19235065690/job/54982687494?pr=15698)

zingo:
rebased to retrigger all test

zingo:
rebase worked, other fails are unrelated

zingo (approved):

PR #15699, opened 2025-11-10
Add get_cond_while_submodules utility

Erik-Lundell:
Adding while to get_control_flow_submodules seems risky, could have unintended side effects.

The new function is similar to get_control_flow_submodules though, so refactor that to share logic.

cc @JacobSzwejbka @angelayi @freddan80 @per @zingo @oscarandersson8218 @digantdesai


JacobSzwejbka:
What is the context for the change?

Erik-Lundell:
> What is the context for the change?

Our backend is planning to support cond and while, which means that there will be many places where we want to access all submodules coming from such ops. See for example #15381 and #15364 that prepare for cond support, but I want to change get_control_flow_submodules -> get_cond_while_submodules when adding while support.

JacobSzwejbka (approved):

PR #15300, opened 2025-10-21
Arm backend: Propagate node info from quantizer to backend

oscarandersson8218:
Use the Node meta 'custom' field to propagate information from quantizer to partitioner using a new ArmAnnotationInfo data class. This  allows us to track quantized node reliably which is useful in order to track which nodes should 'fold' it's quantization parameter and which should be kept in fp when mixing integer and float in a sub-graph.

cc @freddan80 @per @zingo @digantdesai

PR #15701, opened 2025-11-10
Arm backend: Simplify testing of int inputs

oscarandersson8218:
INT-test pipelines always checked that quantized models produced quantize and dequantize nodes, and that these are removed in to_edge. This meant that when you add a test with inputs that do not need to be quantized, e.g. integer and boolean inputs, you would have to pop those stages from the test pipeline.

This patch removes the need for popping those stages by making sure that they are only added if at least one input is in floating point.


zingo:
Some arm tests fails :(

zingo:
@SS-JIA Hi here is a many files PR again :)

oscarandersson8218:
I believe there's a merge conflict with the revert of https://github.com/pytorch/executorch/pull/15300, so I'll park this PR until the new [PR](https://github.com/pytorch/executorch/pull/15778) is merged.

zingo:
As it's merged now, Ill rebase this

zingo (approved):

PR #15709, opened 2025-11-10
[Executorch] Make module constructors uniform across

kimishpatel:
* #15710
* __->__ #15709
* #15708
* #15611
* #15610
* #15609
* #15608
* #15607
* #15653

Existing constructors dont compose well such that if you want data loader or data files constructor then you cannot get to override memory allocator.
Fix that.

Differential Revision: [D86120037](https://our.internmc.facebook.com/intern/diff/D86120037/)

### DIFF
--- a/extension/module/module.h
+++ b/extension/module/module.h
@@ -63,6 +63,8 @@ class Module {
   explicit Module(
       const std::string& file_path,
       const LoadMode load_mode = LoadMode::File,
+      std::unique_ptr<runtime::MemoryAllocator> memory_allocator = nullptr,
+      std::unique_ptr<runtime::MemoryAllocator> temp_allocator = nullptr,
       std::unique_ptr<runtime::EventTracer> event_tracer = nullptr);
 
   /**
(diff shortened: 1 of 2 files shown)


extension/module/module.h
```diff
@@ -63,6 +63,8 @@ class Module {
   explicit Module(
       const std::string& file_path,
       const LoadMode load_mode = LoadMode::File,
+      std::unique_ptr<runtime::MemoryAllocator> memory_allocator = nullptr,
```
JacobSzwejbka, line 66: nit: method_allocator

extension/module/module.h
```diff
@@ -63,6 +63,8 @@ class Module {
...
+      std::unique_ptr<runtime::MemoryAllocator> temp_allocator = nullptr,
       std::unique_ptr<runtime::EventTracer> event_tracer = nullptr);
```
JacobSzwejbka, line 68: This is technically a bc breaking change

kimishpatel: That is true. Is this part of core API btw given that it is in extension?

I do think we should unify the interface though.

extension/module/module.h
kimishpatel: oh I see. This is supposed to be called method_allocator?

PR #15713, opened 2025-11-10
check in lib file for now and throw it in the pip install

JacobSzwejbka:
the .lib file comes from compiling aoti_cuda_shims on windows. Ideally we will not check this in but rather cross compile it on the fly from linux. Thats coming soonish. 

We need the .lib file in order to lower for cuda windows backend. Cuda windows uses AOTI as the acceleration lib, AOTI generates executable code that calls into the shim. On windows you must provide a reference to all symbols you will link against (unlike linux) so we need the .lib artifact.


setup.py
```diff
@@ -918,5 +918,11 @@ def run(self):  # noqa C901
             is_dynamic_lib=True,
             dependent_cmake_flags=["EXECUTORCH_BUILD_KERNELS_LLM_AOT"],
         ),
+        BuiltFile(
+            src_dir="backends/cuda/runtime/",
+            src_name="aoti_cuda_shims.lib",
+            dst="executorch/data/bin/",
```
larryliu0820: should it match the src_dir?

larryliu0820 (approved):

setup.py
```diff
@@ -918,5 +918,11 @@ def run(self):  # noqa C901
...
+            dependent_cmake_flags=[],
```
larryliu0820, line 925: I think you need to dependent on EXECUTORCH_BUILD_CUDA

setup.py
JacobSzwejbka: I think this is probably clearer, though if theres an executorch/data/lib thatd probably be even better let me go check

setup.py
JacobSzwejbka: I dont want to be dependent on that build flag. I want to always put it in the pip package because this is for the AoT flow which is no longer gated by the cmake flag

JacobSzwejbka: I could technically try to gate this behind if we are building cuda, but its just a temporary hack anyway till we have cross compiling so I think its fine to just land.

PR #15716, opened 2025-11-10
Improve warnings for qnn runner

abhinaykukkadapu:
Summary:
Changes:
1. add `--seq_len` param to llama script to distinguish max_seq_len which is compile time param
2. Add warnings in the runner when `seq_len` is clamped to `max_seq_len` to avoid silently clamping it.
3. Add warnings in the token generator when EOS is not reached due to insufficient seq_len or max_seq_len.

Differential Revision: D86696759

## Tests

Use --seq_len=600, prompt_len=512

```
I 00:00:02.883890 executorch:token_generator.cpp:333] Warning: Generation stopped at seq_len limit (600) without reaching EOS token. Response may be incomplete.
I 00:00:02.884094 executorch:token_generator.cpp:346] - seq_len (600) is less than compiled max_seq_len (1024). Consider increasing --seq_len (up to 1024).
```

Use --seq_len=2048, prefill_ar_len=1024

```
I 00:00:00.546967 executorch:runner.cpp:385] Warning: Requested seq_len (2048) exceeds compiled max_seq_len (1024). Clamping to 1024.
```

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
+++ b/examples/qualcomm/oss_scripts/llama/runner/runner.cpp
@@ -376,7 +376,22 @@ Error Runner<T>::generate_from_prompt_or_file(
   stats_.inference_start_ms = time_in_ms();
 
   int32_t seq_len = config.seq_len;
-  seq_len = (seq_len > 0 && seq_len <= context_len_) ? seq_len : context_len_;
+  if (seq_len > context_len_) {
+    ET_LOG(
+        Info,
+        "Warning: Requested seq_len (%d) exceeds compiled max_seq_len (%d). Clamping to %d.",
+        seq_len,
+        context_len_,
+        context_len_);
+    seq_len = context_len_;
+  } else if (seq_len <= 0) {
+    ET_LOG(
+        Info,
+        "Warning: Invalid seq_len (%d). Using compiled max_seq_len (%d).",
+        seq_len,
+        context_len_);
+    seq_len = context_len_;
+  }
   int32_t n_bos = (cur_pos_ == 0) ? 1 : 0;
 
   // encode the (string) prompt into tokens sequence
--- a/examples/qualcomm/oss_scripts/llama/runner/token_generator.cpp
+++ b/examples/qualcomm/oss_scripts/llama/runner/token_generator.cpp
@@ -323,6 +323,30 @@ Result<int64_t> TokenGenerator<T>::generate(
       break;
     }
   }
+
+  // Check if generation was truncated due to seq_len limit (no EOS token)
+  if (eos_ids_->count(cur_token) == 0 && pos >= seq_len - 1) {
+    printf("\n");
+    ET_LOG(
+        Info,
+        "Warning: Generation stopped at seq_len limit (%d) without reaching EOS token. Response may be incomplete.",
+        seq_len);
+    if (seq_len >= metadata_.context_len) {
+      ET_LOG(
+          Info,
+          "- seq_len (%d) already equals compiled max_seq_len (%d). Consider recompiling with larger --max_seq_len.",
+          seq_len,
+          metadata_.context_len);
+    } else {
+      ET_LOG(
+          Info,
+          "- seq_len (%d) is less than compiled max_seq_len (%d). Consider increasing --seq_len (up to %d).",
+          seq_len,
+          metadata_.context_len,
+          metadata_.context_len);
+    }
+  }
+
   return pos - start_pos;
 }
 // Explicit instantiations


winskuo-quic (commented):
Hi @abhinaykukkadapu,
Thanks for the PR.
I would like to know if we could achieve the same thing with the combination of 
`--pre_gen_pte & --max_seq_len`.

For example:
During compile time, you can provide:
`--max_seq_len 1024 --compile_only`

During inference, you can provide:
`--max_seq_len 512 --pre_gen_pte ./path_to_pregen_pte`

abhinaykukkadapu:
> Hi @abhinaykukkadapu, Thanks for the PR. I would like to know if we could achieve the same thing with the combination of `--pre_gen_pte & --max_seq_len`.
> 
> For example: During compile time, you can provide: `--max_seq_len 1024 --compile_only`
> 
> During inference, you can provide: `--max_seq_len 512 --pre_gen_pte ./path_to_pregen_pte`

@winskuo-quic Thanks for the quick review, the goal of this additional param is to avoid confusing the users of the script thinking that --max_seq_len can be dynamic but it is a static param and is fixed during compilation.

Currently, one can pass `--max_seq_len` for inference which actually made me think that total context length can be changed dynamically and i only found out after digging through the code that is is piped as `seq_len` to the runner. With this new param, we have a clear distinction, that `--max_seq_len` is used for compile time and `--seq_len` is during runtime/ inference. Functionally this change is a no-op and it only improves user experience and making it visible to the user when we clamp it down.

examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -1170,11 +1171,18 @@ def _build_parser():
...
         "--max_seq_len",
-        help="This refers to maximum number of tokens that the model can process & consider at once to generate predictions/responses.",
+        help="[Compile-time] Maximum sequence length compiled into the model (sets buffer sizes and context_len). This is the hard limit for the model's context window.",
         default=512,
         type=int,
     )
 
+    parser.add_argument(
+        "--seq_len",
+        help="[Runtime-time] Maximum number of tokens to generate (prompt + output). If not specified, uses --max_seq_len. Will be clamped to compiled max_seq_len if exceeded.",
```
winskuo-quic: Maybe `[Runtime-time]` -> `[Runtime]`

winskuo-quic:
> @winskuo-quic Thanks for the quick review, the goal of this additional param is to avoid confusing the users of the script thinking that --max_seq_len can be dynamic but it is a static param and is fixed during compilation.
> 
> Currently, one can pass `--max_seq_len` for inference which actually made me think that total context length can be changed dynamically and i only found out after digging through the code that is is piped as `seq_len` to the runner. With this new param, we have a clear distinction, that `--max_seq_len` is used for compile time and `--seq_len` is during runtime/ inference. Functionally this change is a no-op and it only improves user experience and making it visible to the user when we clamp it down.

I see.
I think it makes sense to add some warning messages in `.cpp` files to guide users, which is helpful.
However, for `llama.py`, I would like to know if you think flag `--max_seq_len` is misleading? The `--max_seq_len` flag can actually be set to a different number every time during execution, as long as it is shorter than the `--max_seq_len` used during compilation.

abhinaykukkadapu:
> I think it makes sense to add some warning messages in `.cpp` files to guide users, which is helpful.

@winskuo-quic Right, i think we should be transparent in these, i've already added the messages that i think would be helpful, but suggest if you have any more in mind.

> However, for llama.py, I would like to know if you think flag --max_seq_len is misleading? The --max_seq_len flag can actually be set to a different number every time during execution, as long as it is shorter than the --max_seq_len used during compilation.

Yeah i think this param is misleading as it clearly represents max context a model can have, so using it to during inference is a bit misleading for someone new to Qcom delegate, who might not know we use static llama and might think this can change total context length of the model dynamically. Also, all i did is use the same param of qnn runner (--seq_len) and bubbled up to llama.py script.

If you think this adds to the confusion, I'm also open to remove it and only keep warning messages in this PR.

examples/qualcomm/oss_scripts/llama/llama.py
```diff
@@ -941,7 +941,11 @@ def post_process():
         with open(f"{args.artifact}/outputs/outputs.txt", "r") as f:
             outputs.append(f.read())
 
-    seq_len = args.max_seq_len
+    # Use --seq_len if provided (inference-only), otherwise fall back to --max_seq_len
```
cccclai: I don't quite follow why we need seq_len, can you share more? I feel like it might further causing the confusion...

abhinaykukkadapu: The runner itself uses a param named `--seq_len` it is the script llama.py repurposing `--max_seq_len` and using it as `seq_len` for the runner. For context, if you've followed the internal discussion on benchmark numbers, we thought we swept the benchmarks for max_seq_len and prompt_length but the sweeping was not valid for max_seq_len because it is a compile time param and is ignored if it is more than what the model is compiled with. 

Looking at CoreML and they use separate params as well: https://github.com/pytorch/executorch/blob/main/examples/apple/coreml/llama/run.py#L97-L103

Open to suggestions, if there are better ways to distinguish this param during compile time vs runtime from ux perspective.

cccclai: I think we need to re-visit the seq_len in qnn llama runner because it seems it's only for debugging/profiling purpose. Users wouldn't need to use it and I want to make sure we don't make it more confusing.

winskuo-quic (approved):
Thanks for the explanation. 
I am slightly leaning toward leaving warnings messages in runtime and reuse the `--max_seq_len` flag in `llama.py`, which aligns with your new commit.
Thanks for the help on improving user's experience!

cccclai (approved):

cccclai:
Let's rename the PR title because it's not introducing seq_len anymore

PR #15719, opened 2025-11-10
Only block when building with MSVC

JacobSzwejbka:


tools/cmake/preset/llm.cmake
```diff
@@ -16,13 +16,21 @@ set_overridable_option(EXECUTORCH_BUILD_EXTENSION_TENSOR ON)
...
   set_overridable_option(EXECUTORCH_BUILD_KERNELS_QUANTIZED ON)
   set_overridable_option(EXECUTORCH_BUILD_KERNELS_LLM ON)
+else()
+  if(NOT EXECUTORCH_BUILD_CUDA)
+    message(
+      WARNING
+        "The llm custom kernels and the quantized kernels will not be built when using MSVC on Windows. "
+        "If you need them (since you appear to be building for CPU) try building with -T ClangCL"
+    )
+  endif()
```
larryliu0820, line 33: I don't think you need the else() here

larryliu0820: well I guess it's fine.

larryliu0820 (approved):

PR #15731, opened 2025-11-11
Enable Exynosbackend e2e test

Jiseong-oh:
### Summary
Enable E2E test for the Exynos backend.
Modify it to run the testing using a device from devicefarm that can be used directly in CI.

### Test plan
Each test case was verified on a real device.


SS-JIA (approved):
Very cool - LGTM!

PR #15755, opened 2025-11-11
Update ReplaceLogicalNotBooleanWhereWithWherePass to use new pass interface

DrJessop:
Summary: As titled, more efficient now and properly updates the modified bit. Also, this pass was missing tests, so added some for a variety of different cases.

Differential Revision: D86782910


abeakkas (approved):
With phabricator comments.

PR #15764, opened 2025-11-11
Arm backend: Make remaining passes inherit from ArmPass

martinlsm:


zingo (approved):

zingo:
Fails unrelated 
Passes are handled with a glob for buck2 so no TARGET files should need and update

PR #15770, opened 2025-11-12
CortexM Tests - Add Boundary tests to test_mul

psiddh:
### Summary

### Test plan

### DIFF
--- a/backends/cortex_m/test/ops/test_mul.py
+++ b/backends/cortex_m/test/ops/test_mul.py
@@ -111,6 +111,14 @@ test_cases = {
             ramp_tensor(-5, 5, (1, 2, 1, 2)),
         ),
     ),
+    "boundary_extreme_positive": McuTestCase(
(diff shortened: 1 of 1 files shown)


backends/cortex_m/test/ops/test_mul.py
```diff
@@ -111,6 +111,14 @@ class CortexMTensorMul(Model):
             ramp_tensor(-5, 5, (1, 2, 1, 2)),
         ),
     ),
+    "boundary_extreme_positive": McuTestCase(
```
AdrianLundell, line 114: I don't think this is testing exactly what is intended since the input values are quantized so 127, -128 won't be at the boundary any more than any other value (depending on how the model is calibrated). Have you compared the values in the INT domain with the other tests and made sure that this is actually testing something new?

psiddh: Yeah you are right , I will look into this later

PR #15776, opened 2025-11-12
Revert "Arm backend: Don't run pass call_operator in submodules"

per:
Reverts pytorch/executorch#15736


zingo:
Forward fix was landed instead

PR #15759, opened 2025-11-11
Arm backend: forward fix internal failures

SS-JIA:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #15759

Title says it all! forward fix of D86741047, which caused build failures from introducing a new file.

Differential Revision: [D86786510](https://our.internmc.facebook.com/intern/diff/D86786510/)

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai

PR #15778, opened 2025-11-12
Arm backend: Propagate node info from quantizer to backend (re-uploaded) 

oscarandersson8218:
Use the Node meta 'custom' field to propagate information from quantizer to partitioner using a new ArmAnnotationInfo data class. This  allows us to track quantized node reliably which is useful in order to track which nodes should 'fold' it's quantization parameter and which should be kept in fp when mixing integer and float in a sub-graph.

### DIFF
--- a/backends/arm/TARGETS
+++ b/backends/arm/TARGETS
@@ -17,11 +17,7 @@ runtime.python_library(
 )
 runtime.python_library(
     name = "common",
-    srcs = [
-        "common/__init__.py",
-        "common/debug.py",
-        "common/type.py",
-    ],
+    srcs = glob(["common/*.py"]),
     deps = [
         "fbsource//third-party/tosa_tools:serializer",
         "//caffe2:torch",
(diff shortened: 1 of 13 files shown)


SS-JIA:
@oscarandersson8218 still failing due to 

```
E   ModuleNotFoundError: No module named 'executorch.backends.arm.common.annotation_meta'
```

Mind adding the buck file change from https://github.com/pytorch/executorch/pull/15759?

zingo:
Hi @SS-JIA as this and the internal Meta PR  is different now this needs a re-import to get OK to merge e.g. we now get:

[Meta Internal-Only Changes Check](https://github.com/pytorch/executorch/pull/15778/checks?check_run_id=55382145040) — The Diff and Pull Request are not in sync!

Thanks for your help with this PR (and other)

zingo (approved):

SS-JIA:
Tests look good now - feel free to merge!

PR #15780, opened 2025-11-12
Add a CUDA memory tracker and use it in voxtral runner

larryliu0820:
This PR adds a CUDA memory tracker and integrates it into `Stats` of LLM Runner. We added `gpu_total_bytes`, `gpu_free_before_load_bytes`, `gpu_free_after_load_bytes`, `gpu_free_after_generate_bytes` and  `gpu_peak_usage_mb` information.

See logging:

```
PyTorchObserver {"prompt_tokens":387,"generated_tokens":68,"model_load_start_ms":1762976881583,"model_load_end_ms":1762976883487,"inference_start_ms":1762976887396,"inference_end_ms":1762976888589,"prompt_eval_end_ms":1762976887815,"first_token_ms":1762976887815,"aggregate_sampling_time_ms":17,"gpu_total_bytes":17094475776,"gpu_free_before_load_bytes":15589179392,"gpu_free_after_load_bytes":11455692800,"gpu_free_after_generate_bytes":10530848768,"gpu_peak_usage_mb":4824,"SCALING_FACTOR_UNITS_PER_SECOND":1000}
```

### DIFF
--- a/extension/llm/runner/multimodal_runner.cpp
+++ b/extension/llm/runner/multimodal_runner.cpp
@@ -62,14 +62,16 @@ Error MultimodalRunner::load() {
   if (is_loaded()) {
     return Error::Ok;
   }
+  stats_->model_load_start_ms = time_in_ms();
   ET_CHECK_OK_OR_RETURN_ERROR(multimodal_prefiller_->load());
   ET_CHECK_OK_OR_RETURN_ERROR(text_token_generator_->load());
+  stats_->model_load_end_ms = time_in_ms();
 
 #ifdef CUDA_AVAILABLE
-    cuda_memory_tracker_->log_sample("after_load");
-    stats_->gpu_total_bytes = cuda_memory_tracker_->total_bytes();
-    stats_->gpu_free_after_load_bytes = cuda_memory_tracker_->last_free_bytes();
-    stats_->gpu_peak_usage_mb = cuda_memory_tracker_->peak_usage_mb();
+  cuda_memory_tracker_->log_sample("after_load");
+  stats_->gpu_total_bytes = cuda_memory_tracker_->total_bytes();
+  stats_->gpu_free_after_load_bytes = cuda_memory_tracker_->last_free_bytes();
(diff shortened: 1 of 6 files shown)


extension/llm/runner/multimodal_runner.cpp
```diff
@@ -86,9 +109,7 @@ Error MultimodalRunner::generate(
   }
 
   if (!is_loaded()) {
-    stats_->model_load_start_ms = time_in_ms();
```
Gasoonjia, line 89: why do we want to delete the loading time recording?

extension/llm/runner/stats.h
```diff
@@ -75,6 +83,11 @@ struct ET_EXPERIMENTAL Stats {
     aggregate_sampling_time_ms = 0;
     num_prompt_tokens = 0;
     num_generated_tokens = 0;
+    gpu_total_bytes = static_cast<uint64_t>(-1);
+    gpu_free_before_load_bytes = static_cast<uint64_t>(-1);
+    gpu_free_after_load_bytes = static_cast<uint64_t>(-1);
+    gpu_free_after_generate_bytes = static_cast<uint64_t>(-1);
+    gpu_peak_usage_mb = -1.0;
```
Gasoonjia, line 90: I think we can rename it for better description. I don't think it is actually the peak gpu usage, it is the memory usage gap between start and finish running. It could be happen if the peak moment happen in the middle of execution. 

Same as other places for "peak gpu memory"

Gasoonjia (approved):
Thanks for adding that! The logic LGTM, i just have some concern if we should use the word `peak gpu memory`

extension/llm/runner/multimodal_runner.cpp
larryliu0820: Moved it inside `load()`

extension/llm/runner/stats.h
larryliu0820: The meaning of this field is "among all the samples that I collected using `log_sample()` what is the peak gpu memory". I think we can add a few more sample points, but from what I observed, the memory usage has been quite stable after `load()` finishes.

PR #15790, opened 2025-11-12
Arm backend: Sort passes in transform_for_annotation_pipeline

martinlsm:
The passes listed in ArmPassManager.transform_for_annotation_pipeline can feel a bit arbitrary because there is no clearly intended structure or pattern being applied there. Restructure the list into clearly labelled blocks to make the code easier to read and maintain.

### DIFF
--- a/backends/arm/_passes/decompose_remainder_pass.py
+++ b/backends/arm/_passes/decompose_remainder_pass.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Set, Type
+from typing import Dict, Set, Type
 
 import torch
 from executorch.backends.arm._passes import ArmPass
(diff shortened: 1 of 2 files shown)


backends/arm/_passes/decompose_remainder_pass.py
```diff
@@ -17,46 +17,50 @@
 
 Op = OpOverload | EdgeOpOverload
 
-
-def _get_remainder_decomposition_ops(op: Op) -> tuple[Op, Op, Op]:
-    """
```
zingo, line 22: Is removing this help text on purpose or a missed rebase thing?

 Returns the (div_mode_op, mul_op, sub_op) needed to lower the provided
    remainder operator. The concrete ops depend on whether the remainder op is
    the aten or edge variant.

martinlsm: It is intended actually. I thought it was a bit redundant info so I removed it (and got a +2). Good that you're asking

zingo (approved):

PR #15792, opened 2025-11-12
Support eq.Scalar

cccclai:
It's reported from Discord and this pattern fails to lower to QNN
```
class EqualFromInplaceCopyDecomp(torch.nn.Module):
    def __init__(self, hidden_size=4):
        super().__init__()
        # a small state tensor
        self.register_buffer("h", torch.zeros((1, hidden_size)))

    def forward(self, x):
        self.h[0] = x
        return self.h[0]

```

Differential Revision: D86891707

### DIFF
--- a/backends/qualcomm/builders/node_visitor.py
+++ b/backends/qualcomm/builders/node_visitor.py
@@ -466,6 +466,7 @@ class NodeVisitor:
             tensor_source_node, target_build_node
         )
         dtype = self.get_data_type(tensor, quant_configs)
+        print(f"tensor_name: {tensor_name}, tensor_type: {tensor_type}, dtype: {dtype}")
         if isinstance(tensor, torch._subclasses.fake_tensor.FakeTensor):
             tensor_wrapper = PyQnnWrapper.TensorWrapper(
                 tensor_name,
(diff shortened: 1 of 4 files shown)


cccclai:
@haowhsu-quic @winskuo-quic @shewu-quic @DannyYuyang-quic  Hi team, I might need some help for this pattern. I managed to lower the pattern to fp, but it's failing with the quantized version. I think a copy_ node shows up in the graph indicating it's a mutable buffer. Do you know the best way moving forward with this pattern?

winskuo-quic:
> @haowhsu-quic @winskuo-quic @shewu-quic @DannyYuyang-quic Hi team, I might need some help for this pattern. I managed to lower the pattern to fp, but it's failing with the quantized version. I think a copy_ node shows up in the graph indicating it's a mutable buffer. Do you know the best way moving forward with this pattern?

Hi @cccclai,

I am able to reproduce the issue where I get the following graph finalize error message:
`No graph inputs present for graph [0]`

I believe this has to do with the behavior of run_decomposition here: https://github.com/pytorch/executorch/blob/b433278ec47a2ff9274fc456b62a0511a941dee7/exir/program/_program.py#L1176

The input node is actually not connected to the graph after `run_decomposition` for quantization flow, as shown in the image below. The copy_ node also disappeared.
   
image


On the other hand, floating point flow is working properly because the input node is still connected to the graph after `run_decomposition`.
image

PR #15798, opened 2025-11-13
Remove model_name param from Whisper-Metal

seyeong-han:
# Issue

I found that the tokenizer was wrong when I run `open-ai/whisper-tiny` model using `whisper_runner` by looking at the transcription result.

- Expected 
```
<\|en\|><\|transcribe\|><\|notimestamps\|> This week, I traveled to Chicago to deliver my final farewell address to the nation, following in the tradition of presidents before me. It was not opportunity to say thank you. Whether we've seen IDI or rarely agreed at all, my conversations with you, the American people, in living rooms and schools,<\|endoftext\|>
```
- Result
```
<|startoftranscript|><|translate|><|10.00|> So.<|24.00|><|endoftext|>
```

Since HuggingFace has updated all Whisper model tokenizers to the v3 format, we don't need to care about the `decoder_start_token_id` manually.

## Solution

- **Removed `model_name` argument** from `run.sh` and `main.cpp`
- Hardcoded `decoder_start_token_id=50258` for all models
- Fixes tokenizer compatibility issue where all Whisper models from HuggingFace now use the v3 tokenizer format
- Eliminates confusion about which model name to pass at runtime

@manuelcandales


Gasoonjia (approved):

cccclai:
This PR is breaking tests listed in  https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=whisper and I'm reverting in https://github.com/pytorch/executorch/pull/15856

PR #15800, opened 2025-11-13
Introduce PTEFile class

lucylq:
* __->__ #15800

PTEFile class holds the components of a PTE file: the program, mutable constants and named data.

Currently, the `program` definition does not contain mutable constants and named data; they are always stored in segments and not inline. This means when we deserialize, they are lost, because we only deserialize into the `program` concept.

Now, segment data is included in the PTEFile class.

Differential Revision: [D86814175](https://our.internmc.facebook.com/intern/diff/D86814175/)


backends/cadence/runtime/runtime.py
```diff
@@ -45,7 +45,7 @@ def get_op_names(program: et_schema.Program, execution_plan_id: int = 0) -> set[
             op_names |= get_op_names(
                 deserialize_pte_binary(
                     program.backend_delegate_data[delegate.processed.index].data
-                )
+                ).program
```
larryliu0820, line 48: Is this a bug?

larryliu0820: oh nvm you changed the return type of `deserialize_pte_binary`

larryliu0820 (approved):

zingo (commented):
It might be nice it bundleIO BPTE files could be mixed into this in some way and handled similarly.

lucylq:
> It might be nice it bundleIO BPTE files could be mixed into this in some way and handled similarly.

@zingo thanks for the suggestion - could you comment a bit more on what you're looking for?

zingo:
> > It might be nice it bundleIO BPTE files could be mixed into this in some way and handled similarly.
> 
> @zingo thanks for the suggestion - could you comment a bit more on what you're looking for?

Im thinking the it might be god to have if you made tools parsing the different files. As Im not writing those kind of tools right now Im not sure its fully applicable but I feel the bpte are sometimes forgotten or need special handling.

PR #15801, opened 2025-11-13
Use PTEFile class in serialize_pte_binary

lucylq:
* __->__ #15801
* #15800

Take in a PTE file for serialization instead of program, mutable_segments, named_data segments separately.

Differential Revision: [D86908241](https://our.internmc.facebook.com/intern/diff/D86908241/)


devtools/bundled_program/test/test_bundle_data.py
```diff
@@ -18,7 +18,7 @@
 from executorch.devtools.bundled_program.util.test_util import (
     get_common_executorch_program,
 )
-from executorch.exir._serialize import _serialize_pte_binary
+from executorch.exir._serialize import _PTEFile, _serialize_pte_binary
```
larryliu0820, line 21: why `_PTEFile` but not `PTEFile`

larryliu0820: Is `PTEFile` supposed to be private only?

larryliu0820 (approved):

devtools/bundled_program/test/test_bundle_data.py
lucylq: Mostly following _serialize_pte_binary and _deserialize_pte_binary ...

I think those are fairly public though haha

PR #15813, opened 2025-11-13
CUDA delegate: fuse int4 kernel for better performance

mergennachin:


desertfire:
Is it possible to make this a part of Inductor freezing optimization? I think it will benefit all models beyond ET use cases.

PR #15815, opened 2025-11-13
Add a fast path for _clone_dim_order

GregoryComer:
### Summary
Add a direct memcpy fast path for the portable _clone_dim_order op, as it can be a performance bottleneck. I'd like to more aggressively optimize these out of the graph, but this fast path should reduce the perf impact significantly.

### Test plan
Existing correctness tests for the _clone_dim_order implementation should cover it.

For performance, I did a quick test with a default dim order (1, 128, 256, 256) element tensor on an x86 server. This is mainly intended as a quick smoke test and not a proper benchmark. I included numbers for both optimized and debug builds. Optimized matters more, but super long debug runs can be painful for development.

[Optimized Build]
Before: 27.9 ms
After: 6.4 ms

[Debug Build]
Before: 5947.01 ms
After: 7.2 ms

### DIFF
--- a/kernels/portable/cpu/op__clone_dim_order.cpp
+++ b/kernels/portable/cpu/op__clone_dim_order.cpp
@@ -10,6 +10,9 @@
 #include <executorch/kernels/portable/cpu/util/copy_ops_util.h>
 #include <executorch/runtime/kernel/kernel_includes.h>
 
+#include <algorithm>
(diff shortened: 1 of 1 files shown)


GregoryComer:
Note that the moshi failure is pre-existing.

Gasoonjia (approved):
LGTM!

PR #15822, opened 2025-11-14
Introduce Makefile and some runner targets

larryliu0820:
This way we can do

```
make llama-cpu
```
to build llama runner with CPU backend, instead of the original:

```
cmake --preset llm -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=cmake-out

cmake --build cmake-out -j16 --target install --config Release

cmake -DCMAKE_INSTALL_PREFIX=cmake-out \
	-DCMAKE_BUILD_TYPE=Release \
	-Bcmake-out/examples/models/llama \
	examples/models/llama

cmake --build cmake-out/examples/models/llama -j16 --config Release
```

I added a bunch of runners such as whisper, gemma3, voxtral and llava.

Another good thing about `Makefile` is that it supports tab completion.

### Test plan
Manually tested a few of them. Changed CI job to use these commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,96 @@
+# ==============================================================================
+# ExecuTorch Targets Makefile
+# ==============================================================================
+#
+# This Makefile provides convenient targets for building ExecuTorch model runners
+# with different backend configurations (CPU, CUDA, Metal), as well as other
+# binary targets.
+#
+# WHAT THIS BUILDS:
+# -----------------
+# Each target builds:
+# 1. ExecuTorch core libraries with the specified backend (CPU, CUDA, or Metal)
+# 2. The model-specific runner executable in cmake-out/examples/models/<model>/
+#
+# SUPPORTED MODELS:
+# -----------------
+# - voxtral:  Multimodal voice + text model (CPU, CUDA, Metal)
+# - whisper:  Speech recognition model (CPU, CUDA, Metal)
+# - llama:    Text generation model (CPU)
+# - llava:    Vision + language model (CPU)
+# - gemma3:   Text generation model (CPU, CUDA)
+#
+# USAGE:
+# ------
+# make <model>-<backend>    # Build a specific model with a backend
+# make help                  # Show all available targets
+# make clean                 # Remove all build artifacts
+#
+# Examples:
+#   make voxtral-cuda        # Build Voxtral with CUDA backend
+#   make llama-cpu           # Build Llama with CPU backend
+#   make whisper-metal       # Build Whisper with Metal backend (macOS)
+#
+# HOW TO ADD A NEW MODEL:
+# -----------------------
+# To add a new model (e.g., "mymodel"), follow these steps:
+#
+# 1. Create a CMakePresets.json in examples/models/mymodel/:
+#    - Define configurePresets for each backend (base, cpu, cuda, metal)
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -0,0 +1,108 @@
...
+	@echo "  llava-cpu      - Build Llava runner with CPU backend"
+	@echo "  gemma3-cuda    - Build Gemma3 runner with CUDA backend"
+	@echo "  gemma3-cpu     - Build Gemma3 runner with CPU backend"
+	@echo "  clean          - Clean build artifacts"
+
+voxtral-cuda:
+	@echo "==> Building and installing ExecuTorch with CUDA..."
+	cmake --workflow --preset llm-release-cuda
+	@echo "==> Building Voxtral runner with CUDA..."
+	cd examples/models/voxtral && cmake --workflow --preset voxtral-cuda
```
lucylq, line 110: So the difference here is:

1. different preset for each backend
2. different source code dir for each model

right?

Makefile
```diff
@@ -0,0 +1,108 @@
+.PHONY: voxtral-cuda voxtral-cpu voxtral-metal whisper-cuda whisper-cpu whisper-metal llama-cpu llava-cpu gemma3-cuda gemma3-cpu clean help
```
lucylq, line 90: Does this need a license header?

examples/models/gemma3/CMakePresets.json
```diff
@@ -0,0 +1,76 @@
+{
+    "version": 6,
```
lucylq, line 2: what is this version?

Makefile
```diff
@@ -0,0 +1,108 @@
...
+
+help:
+	@echo "Available targets:"
```
lucylq: Short description as well?

```
This Makefile adds targets to build runners for various models on various backends. Run using make <target>.
```
Might be worth adding a README on how to add a new target, or point to an example.

examples/models/gemma3/CMakePresets.json
```diff
@@ -0,0 +1,76 @@
...
+    "configurePresets": [
+        {
+            "name": "gemma3-base",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/../../../cmake-out/examples/models/gemma3",
```
lucylq, line 7: noob q, how is `sourceDir` set?

lucylq (approved):
This is very clean! Thank you

Makefile
larryliu0820: Yeah, first workflow builds ET C++ libraries, with different flavors (CUDA enabled, Metal enabled etc). Second workflow builds the actual runner that depends on the libraries.

Makefile
larryliu0820: I don't see it in pytorch/pytorch so maybe not? https://github.com/pytorch/pytorch/blob/main/Makefile

examples/models/gemma3/CMakePresets.json
larryliu0820: Screenshot 2025-11-13 at 5 46 31 PM

examples/models/gemma3/CMakePresets.json
larryliu0820: The directory of the current file

PR #14109, opened 2025-09-09
Arm backend: Enable pytest VKML op unit testing in ciflow/trunk

mansnils:
Enable VKML op unit test for Arm backend.

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218

PR #15826, opened 2025-11-14
Arm backend: test without output re-order workaround

mansnils:
By default outputs are re-ordered to correct order during TOSA lowering. However this is seen as a workaround as it should not be needed. Furthermore the output issue is not easily reproduced, rather it seems to happen randomly. Therefore we add a test case without the workaround, which is currently passing.
In case it won't pass without the workaround at some point, the new changes might give some hints on why the workaround is needed and how to fix it.
In case it continues to pass, we may switch the default and potentially even remove the workaround.


zingo:
Timeout problem [trunk / test-arm-backend (test_pytest_ops_ethosu_fvp) / linux-job (push)](https://github.com/pytorch/executorch/actions/runs/19368534024/job/55418045900?pr=15826)Cancelled after 121m

zingo:
Seem to be happening after https://github.com/pytorch/executorch/pull/14109 was merged, trying to revert that  PR

mansnils:
See comment for more context of workaround: https://github.com/pytorch/executorch/pull/13454#discussion_r2288743796

zingo (approved):

PR #15838, opened 2025-11-15
Remove no-op clones

GregoryComer:
Remove clones that don't change memory layout / dim order. As the graph is assumed to be functional, these are no-ops. We've seen some non-zero performance impact. Some backends optimize them out, but it makes sense to do it at the to_edge level, as the optimization is global.

As this impacts several consumers, I've synced with NXP (see PR comments) and with RL / Jarvis to verify that the change is acceptable.

Differential Revision: D86588171

### DIFF
--- a/backends/nxp/tests/ir/converter/node_converter/test_clone_converter.py
+++ b/backends/nxp/tests/ir/converter/node_converter/test_clone_converter.py
@@ -76,6 +76,7 @@ class KWSFinalBlock(torch.nn.Module):
         return self.block(x)
 
 
+@unittest.skip("Clones are optimized out of the graph.")
 class TestCloneConverter(unittest.TestCase):
     __test__ = False  # Prevent interfering with PyTest tests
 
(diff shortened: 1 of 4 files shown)


GregoryComer:
@MartinPavella @StrycekSimon I'm looking at optimizing out clones more aggressively from the graph during to_edge. I see that the NXP backend has some test coverage looking at clones in the context of the dropout op. Do you have any concerns with removal of clones from the graph? If not, I'm happy to take a pass at updating the tests accordingly.

MartinPavella:
> @MartinPavella @StrycekSimon I'm looking at optimizing out clones more aggressively from the graph during to_edge. I see that the NXP backend has some test coverage looking at clones in the context of the dropout op. Do you have any concerns with removal of clones from the graph? If not, I'm happy to take a pass at updating the tests accordingly.

Hi @GregoryComer ,
we have no issues with removing clones which leave the dim order unchanged. Feel free to skip our tests for now. We will update them accordingly later if need be.

MartinPavella (approved):

robert-kalmar (approved):

PR #15846, opened 2025-11-17
Arm backend: Fix ubsan warning in example runner

perheld:
Avoid dangerous pointer arithmetic by making sure EValue lands at its natural alignment (8 bytes on this target).

Change-Id: I3efee39f5c36ea76e08373fccc5ff10c01cbc588


zingo (approved):

perheld:
Memory allocator test fails so I need to update that. Hold for a second commit in this PR.

perheld:
> Memory allocator test fails so I need to update that. Hold for a second commit in this PR.

Fixed

PR #15856, opened 2025-11-17
Revert "Remove model_name param from Whisper-Metal"

cccclai:
Reverts pytorch/executorch#15798


manuelcandales:
Let's wait on https://github.com/pytorch/executorch/pull/15858. The fix should be as simple as that.

cccclai:
forward pass seems good

PR #15858, opened 2025-11-17
Fix Metal/CUDA ci: remove model_name argument

manuelcandales:


cccclai (approved):
Thanks! Let's merge if the test passes

PR #15862, opened 2025-11-18
Save external constant tensors to custom filename

lucylq:
Summary: Adding a Callable option in the config, so we can customize the nodes to save, and the filenames to save to.
The pass is applied in _program.py, it's hard to do this on the eager model (like we do with delegates) as we have to run the pass after SpecPropPass.

Differential Revision: D87280747


ethansfng (approved):

JacobSzwejbka (approved):

PR #15872, opened 2025-11-18
Cortex_m backend: Add SharedQspecQuantizer + maximum/minimum

AdrianLundell:
This patch reworks how ops w/o rescaling are handled. Instead of not annotating them and counting on the retracing to make them get correct dtype, they must now be annotated by the SharedQspecQuantizer.

The reason for this is that it simplifies ensuring that non-rescaling with multiple inputs has the same scale on both inputs. It also becomes easier to tell which dtype each op will receive before folding.


zingo (approved):
Nice

PR #15879, opened 2025-11-19
Manually pick moshi requirements update from #15180

GregoryComer:
### Summary
The torchcodec nightly version we are pinning is out of retention. It was updated on main as part of #15180. This PR picks just that change to the requirements file to avoid picking up all of the other large changes in the PR. The follow-up change to the ffmpeg install in #15821 is also picked here, since it's needed for the previous change to work.

### DIFF
--- a/examples/models/moshi/mimi/install_requirements.sh
+++ b/examples/models/moshi/mimi/install_requirements.sh
@@ -7,10 +7,10 @@
 
 set -x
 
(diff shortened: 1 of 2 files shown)


GregoryComer:
Note that the OpenVINO failure is pre-existing (fixed in https://github.com/pytorch/executorch/pull/15880).

mergennachin (approved):

lucylq (approved):

PR #15884, opened 2025-11-19
Remove no-op clones in xnnpack

GregoryComer:
ATen clone ops can end up in the graph from a few sources. Since the graph is functional, we don't actually need these and they are slow. This PR runs the no-op clone removal pass for XNNPACK.

In addition to this, I ran into an issue where XNNPACK delegate doesn't currently handle inputs being directly forwarded to partition outputs. There has to be at least one operator. To solve this, I updated the removal pass to leave these clone ops in and added copy support in the XNN delegate to direct copy to the output.

In the long-run, I want to remove these no-ops higher up as part of to_edge, but this requires alignment and changes with a few more backends. See https://github.com/pytorch/executorch/pull/15838. But resolving for XNNPACK will mitigate the issue for CPU models, at least.

Differential Revision: D87405074

### DIFF
--- a/backends/transforms/remove_clone_ops.py
+++ b/backends/transforms/remove_clone_ops.py
@@ -25,8 +25,9 @@ class RemoveCloneOpsTransform(ExportPass):
         exir_ops.edge.dim_order_ops._clone_dim_order.default,
     }
 
-    def __init__(self) -> None:
+    def __init__(self, preserve_input_output_copies: bool = False) -> None:
         super().__init__()
+        self._preserve_input_output_copies = preserve_input_output_copies
 
     def _remove(self, graph_module: torch.fx.GraphModule) -> None:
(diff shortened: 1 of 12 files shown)


exir/tests/test_memory_format_ops_pass_utils.py
```diff
@@ -79,15 +79,15 @@ def __init__(self):
         super().__init__()
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        return x.clone(memory_format=torch.contiguous_format)
+        return x.clone(memory_format=torch.contiguous_format) * 2
```
GregoryComer: This change is intended to create a test graph that includes at least one non-clone op.

shoumikhin (approved):

Gasoonjia (approved):

PR #15885, opened 2025-11-19
Fix Neutron backend API compatibility and multiprocessing fallback

apullin:
Summary:
This fixes two issues preventing the Neutron backend from building in sandcastle/build environments:

1. **NeutronAtenPassManager API compatibility**: The `NeutronAtenPassManager` class was updated to require a `neutron_target_spec` parameter, but the call site  wasn't updated. This caused a `TypeError` at runtime. Fixed by passing the already-available `neutron_target_spec` to the pass manager constructor.

2. **Multiprocessing fallback**: The Neutron converter uses multiprocessing for isolation when converting models, but this fails in restricted build environments (sandcastle) with an `EOFError` when trying to create a multiprocessing Manager. Added a try/except block to catch `EOFError` and `OSError` and fall back to direct execution of the converter when multiprocessing is unavailable. This maintains isolation in normal environments while allowing builds to succeed in restricted ones.

Both issues were pre-existing on trunk and not caused by recent changes.

Differential Revision: D87252895

### DIFF
--- a/backends/nxp/backend/neutron_converter_manager.py
+++ b/backends/nxp/backend/neutron_converter_manager.py
@@ -78,23 +78,35 @@ class NeutronConverterManager:
         cctx.compilationOpts.minNumOpsPerGraph = 1
         cctx.compilationOpts.excludeGraphPasses = "MergeTranspose"
 
-        logger = multiprocessing.log_to_stderr()
-        logger.setLevel(logging.WARNING)
-        queue = multiprocessing.Manager().Queue()
+        # Try to use multiprocessing for isolation, but fall back to direct execution
+        # if the environment doesn't support it (e.g., in sandcastle/build environments)
+        try:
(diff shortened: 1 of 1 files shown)


robert-kalmar:
@apullin , LGTM, although, don't see the change related to 

> NeutronAtenPassManager API compatibility: The NeutronAtenPassManager class was updated to require a neutron_target_spec parameter, but the call site wasn't updated. This caused a TypeError at runtime. Fixed by passing the already-available neutron_target_spec to the pass manager constructor.

I checked the executorch OSS code base and did not find the above mentioned API compatibility break. Is is something in your internal code base?

robert-kalmar (approved):

apullin:
> I checked the executorch OSS code base and did not find the above mentioned API compatibility break. Is is something in your internal code base?

This is on an internal SDK, yes. My first time syncing between, so I should have labeled that.

PR #15887, opened 2025-11-19
Arm backend: Add docstrings to vgf/backend.py and ethosu/backend.py

Sebastian-Larsson:


zingo (approved):

zingo:
To many fails even if most/all probably are unrelated, Rebased to see if we can get a "OK" amount :)

PR #15889, opened 2025-11-19
NXP backend: added aten.slice support

novak-vaclav:
### Summary

adds support for aten.slice operator

### Test plan


backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,128 @@
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import numpy as np
+from backends.nxp.backend.edge_helper import input_tensor
+from backends.nxp.backend.ir.converter.conversion.common import OpsList
```
MartinPavella: Please don't use relative imports.

The tests pass because Python finds the `backends` directory in our development environment, but it won't work in deployment as there is no `backends` top-level module in the installed Python package.

So, replace all `from backends...` with `from executorch.backends...`.

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,128 @@
...
+
+class SliceTensorConverter(NodeConverter):
+    @staticmethod
+    def _is_supported_on_target(
+        node: Node,
+        neutron_target_spec: NeutronTargetSpec,
+        parameters_mapping: dict[str, Parameter],
+        custom_delegation_options: CustomDelegationOptions,
+    ) -> bool:
+        if node_uses_shape_broadcasting(node):
```
MartinPavella: Shape broadcasting is something that allows us to perform element-wise operations on 2 tensors, that don't necessarily have the same shape. For example if you have 2 tensors with shapes (2, 3) and (2, 3), you can trivially add (or multiply, subtract, divide, ...) their elements together. E.g.:
[[1, 2, 3],
 [4, 5, 6]]
+
[[0, -1, 1],
 [2, -2, 3]]
\=
[[1, 1, 4],
 [6, 3, 9]]

If your tensors have different shapes, for example (2, 3) and (5, 4), the element-wise operations are not defined (how would you add them together?). But for some combinations of shapes, they **are** defined. Going from the right, the dimensions must either match, or one of them must be `1`, and the ranks don't even need to be the same. For example with shapes (2, 3) and (1):
[[1, 2, 3],
 [4, 5, 6]]
+
[-1]
\=
[[0, 1, 2],
 [3, 4, 5]]

or for shapes (2, 3) and (3):
[[1, 2, 3],
 [4, 5, 6]]
+
[-1, 0, 1]
\=
[[0, 2, 4],
 [3, 5, 7]]

Other combinations that work are for example: 
(2, 4, 6, 8) and (4, 1, 8)
(42, 1) and (2, 4, 1, 8)
...

The fact that the tensors can have a different rank (number of dimensions) causes some issues when we are handling the conversion from ExecuTorch's `channels_first` format to NeutronIR's `channels_last` format. These issues may require the insertion of `Transpose` operators to solve, hence the check on line 33.

But this is only the case for operators that support shape broadcasting (Add, Sub, Mul, Div, Pow, ...). Slice is not one of these element-wise operations, therefore this check does not make sense here, and it should be removed.

Sorry for the long comment, but I believe it is important to understand these issues going forward.

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Provisional solution - slice conversion works for neutron software 2.2.1+
+        neutron_flavor = neutron_target_spec.neutron_target.__module__.split(".")[0]
+        if neutron_flavor != "neutron_converter_SDK_25_12":
+            return False
+
+        input_shape = input_tensor(node, 0).shape
+        dim = node.args[1]
+
+        # The rank of the dimension that we want to slice must be divisible by num_macs
```
MartinPavella: What does `The rank of the dimension` mean?
Generally, by `rank` we mean the number of dimensions of a tensor.

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,128 @@
...
+        t_op.builtin_options = slice_options.Slice()
+
+        ops = OpsList(middle_op=t_op)
+        # Insert forward and backward transpose
+        ops.add_pre(self.builder.create_transpose_operator_before(t_op, 0, perm))
+        ops.add_post(self.builder.create_transpose_operator_after(t_op, 0, perm))
+
+        self.builder.append_operators(ops.flatten())
+
+    Dim = Start = End = int
```
MartinPavella, line 131: Nice 👍🏻

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @staticmethod
+    def _get_clipped_slice_args(node: Node) -> (Dim, Start, End):
```
MartinPavella: This type hint is not valid in Python. Use the following instead:
```python
def _get_clipped_slice_args(node: Node) -> tuple[Dim, Start, End]:
```

backends/nxp/backend/edge_program_converter.py
```diff
@@ -41,6 +41,7 @@
     exir_ops.edge.aten.permute_copy.default: PermuteCopyConverter,  # noqa F405
     exir_ops.edge.aten.relu.default: ReLUConverter,  # noqa F405
     exir_ops.edge.aten._softmax.default: SoftmaxConverter,  # noqa F405
+    exir_ops.edge.aten.slice_copy.Tensor: SliceTensorConverter,  # noqa F405
```
MartinPavella, line 44: Please maintain alphabetical order.

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -0,0 +1,229 @@
...
+            (0, 1),
+            (0, 0),
+            (24, 31),
+            id="2D, end arg not divisible by num_macs",
+        ),
+        pytest.param((24, 32), (1, 0), (16, 0), (32, 24), id="2D, mixed dim args"),
+        pytest.param((24, 32), (0, -1), (0, 16), (24, 32), id="2D, negative dim arg"),
+    ],
+)
+def test_slice_tensor_quant_conversion(mocker, x_input_shape, dims, starts, ends):
```
MartinPavella, line 84: I would replace
```python
if neutron_converter_flavor != "SDK_25_12":
    pytest.skip("Neutron Software must be version 2.2.1 or higher.")
```
with 
```python
if neutron_converter_flavor == "SDK_25_09":
    pytest.skip("Neutron Software must be version 2.2.1 or higher.")
```
because there is a risk that we won't update the test and the condition will remain there for `SDK_26_03` and the test will become useless (always skipped).

---


Secondly, the whole test needs to be much more strict to ensure we either detect a failure after the Nutron update, or that the passing test correctly indicates expected (and correct behavior.

Right now, you are not checking the whole exported program at all:
```python
_ = to_quantized_edge_program(model, x_input_shape).exported_program()
```
replace it with something like
```python
full_exported_program = to_quantized_edge_program(model, x_input_shape).exported_program()
```
and make sure the slice operators are indeed delegated. Because right now, you are literally only checking that the last delegated partition produces the same output in ExecuTorch and in NeutronIR. So it is possible that the slice(s) will not be delegated, but some other operator will. `converter_spy.call_args.args[1]` will then return only this unrelated operator, and you will not notice that you are not testing the slice at all.

I realize that the `SliceTensorModule` only contains `slice` operators, but you never know what auxiliary ops are added during the export to aten dialect. This applies to all tests in general, and it's something we should always do.

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -0,0 +1,229 @@
...
+        pytest.param(
+            (1, 4, 32, 48),
+            (0, 1, 2, 3),
+            (0, 0, 8, 0),
+            (1, 8, 32, 48),
+            id="4D, handle channel order swap",
+        )
+    ],
+)
+def test_slice_tensor_w_conv_quant_conversion(
```
MartinPavella, line 133: Same as for the test above.
Since this one uses `convolution` it is even more necessary to properly check that the slice was delegated. As if it wasn't, your test would only capture the Conv, successfully compare its ExecuTorch and NeutronIR output, and from the successful tests we would happily assume the `slice` delegation works.


Edit:
I ran the test and this is indeed what's already happening. Only the `convolution` is being delegated. Please have a look at the test.

Image

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2025 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import numpy as np
+import pytest
+import torch
+
+from backends.nxp.tests.models import SliceTensorModule
```
MartinPavella: Please fix the import (see other comment).

backends/nxp/tests/models.py
```diff
@@ -172,6 +172,46 @@ def forward(self, x):
...
+
+class SliceTensorConvModule(torch.nn.Module):
+    def __init__(self, dims, starts, ends):
+        super().__init__()
+        self.conv = Conv2dModule(in_channels=4, out_channels=8, kernel_size=3, stride=1)
+        self.dims = dims
+        self.starts = starts
+        self.ends = ends
+
+    def do_slice(self, x):
```
MartinPavella: Minor:
Why define the `do_slice` method again instead of using the whole `SliceTensorModule`?

backends/nxp/neutron_partitioner.py
```diff
@@ -210,6 +210,7 @@ def tag_qdq_clusters(self, nodes: list[torch.fx.Node]):
     exir_ops.edge.aten.mm.default: MMConverter,  # noqa F405
     exir_ops.edge.aten.permute_copy.default: PermuteCopyConverter,  # noqa F405
     exir_ops.edge.aten.relu.default: ReLUConverter,  # noqa F405
+    exir_ops.edge.aten.slice_copy.Tensor: SliceTensorConverter,  # noqa F405
```
MartinPavella, line 215: Minor:
Pleas keep alphabetical order. (I know it was already broken by `Sigmoid`, but you can fix both ;) )

MartinPavella (changes requested):

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
novak-vaclav: Ohhh now I understand, thank you for the explanation! 😊
fixed

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
novak-vaclav: Yeah, it doesn't have anything to do with ranks, sorry.
fixed

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
novak-vaclav: thanks 😊

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
novak-vaclav: makes sense, fixed

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
novak-vaclav: Fixed parameters of the test and added a check of delegation.

novak-vaclav:
Thank you for the review, fixed found issues. 
Please check the changes @MartinPavella

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -90,18 +80,25 @@ def _convert_to_slice(self, t_op, main_input, input_rank, dim, start, end) -> No
...
         t_op.builtin_options = slice_options.Slice()
-
         ops = OpsList(middle_op=t_op)
-        # Insert forward and backward transpose
-        ops.add_pre(self.builder.create_transpose_operator_before(t_op, 0, perm))
-        ops.add_post(self.builder.create_transpose_operator_after(t_op, 0, perm))
+
+        # If slicing along non-channels dimension, we need to swap it with channels dimension.
+        # Otherwise Neutron will not convert it.
+        if dim != -1:
```
MartinPavella: The `dim` could also be positive (for example `3` for a 4D input tensor) which would still refer to the last dimension. Actually, as this method is called from the `convert()` method of this class where you have:
```python
if t_op.tmp_inputs[0].tensor_format.is_channels_last():
    dim = translator.create_channels_last_to_channels_first_permutation(
        t_op.tmp_inputs[0].rank
    )[dim]
```
the `dim` will always be positive in case of a channels last input.

Make sure to compare against the positive variant as well.

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -86,13 +87,18 @@ def test_slice_tensor_quant_conversion(mocker, x_input_shape, dims, starts, ends
...
     converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
 
     # Run conversion
-    _ = to_quantized_edge_program(model, x_input_shape).exported_program()
+    edge_program = to_quantized_edge_program(model, x_input_shape).exported_program()
+    edge_nodes = list(edge_program.graph.nodes)
+
+    # Check if slices and potential transposes were delegated
+    node_delegated = edge_nodes[3]
+    assert node_delegated.op == "call_function" and node_delegated.target.__name__ == "executorch_call_delegate"
```
MartinPavella: This is not very strict. You should make sure that there are no `slice` nodes in model. For example:
```python
assert not any("slice" in n.name for n in edge_nodes)
```

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -132,7 +138,12 @@ def test_slice_tensor_w_conv_quant_conversion(
     converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
 
     # Run conversion
-    _ = to_quantized_edge_program(model, x_input_shape)
+    edge_program = to_quantized_edge_program(model, x_input_shape, use_neutron_for_format_conversion=False).exported_program()
+    edge_nodes = list(edge_program.graph.nodes)
+
+    # Check if slices and transposes were delegated
+    node_delegated = edge_nodes[3]
+    assert node_delegated.op == "call_function" and node_delegated.target.__name__ == "executorch_call_delegate"
```
MartinPavella: Not strict enough, just like above. The fact that there is a delegate call doesn't mean that all the nodes in the model are delegated.

Edit:
I ran the test and indeed, the `slice` nodes are not being delegated, 
When developing tests, it is good practice to always manually inspect the graph first to make sure the result is as expected. Please double check al the tests you are adding.

image

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -145,12 +156,15 @@ def test_slice_tensor_w_conv_quant_conversion(
...
 
+    conversion_config = ConversionConfig()
+    conversion_config.use_neutron_for_format_conversion = False
     convert_run_compare(
         exported_program,
         input_data=input_data,
         tflite_input_preprocess=ToChannelLastPreprocess(),
         tfl_model=tflite_flatbuffers_model,
         tflite_output_preprocess=ToChannelFirstPreprocess(),
+        conversion_config=conversion_config
```
MartinPavella: Minor:
This is not necessary as you are already passing the `tfl_model`.

MartinPavella (changes requested):

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
novak-vaclav: The issue with these slices not being delegated is bad Neutron Converter Flavor. Just like in `test_slice_tensor_quant_conversion`, the conversion works for SDK_25_12+. Sorry, I missed this.
Added a conditional skip of the test.

novak-vaclav:
Fixed issues from second round of code review. Thank you for the feedback.

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -0,0 +1,124 @@
...
+        # If slicing along non-channels dimension, we need to swap it with channels dimension.
+        # Otherwise Neutron will not convert it.
+        if dim != -1 and dim != input_rank - 1:
+            # Create permutation for swapping
+            perm = list(range(0, input_rank))
+            perm[dim], perm[-1] = perm[-1], perm[dim]
+
+            # Insert forward and backward transpose
+            ops.add_pre(self.builder.create_transpose_operator_before(t_op, 0, perm))
+            ops.add_post(self.builder.create_transpose_operator_after(t_op, 0, perm))
```
MartinPavella, line 127: The `_is_supported_on_target()` method should verify that these inserted `Transpose` operators are supported. You have access to the node format in `node.meta[NXP_NODE_FORMAT]`, which you can use to determine if the `dim` corresponds to the channels or not, and therefore whether the `Transpose` ops will be necessary.

However, as you are prohibiting `slice` delegation altogether until `SDK_25_12`, this is not necessary. Currently, only some transpositions are supported on Neutron, but in the next release (hopefully) all these "simple" transpositions should be working.

So you should either correctly call `transposition_is_supported_on_neutron()` in `_is_supported_on_target()` for both the transpositions if they will be inserted, or you should somehow make sure that when we switch to `SDK_25_12`, we verify if this is necessary. Perhaps a comment in the `_is_supported_on_target()` near the SDK version restriction will be enough. I'll leave the choice to you.

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
```diff
@@ -0,0 +1,244 @@
...
+            (24, 32, 64, 48),
+            (0, 1, 2, 3),
+            (0, 0, 8, 0),
+            (24, 32, 64, 48),
+            id="4D, one transpose",
+        ),
+        pytest.param(
+            (24, 32, 64), (0, 1, 2), (8, 8, 0), (24, 32, 64), id="3D, two transposes"
+        ),
+        # bug in neutron-converter will not properly convert models in these test cases
```
MartinPavella, line 63: What bug is this? Does the Neutron team know about it? Can we determine when this bug will happen and prohibit delegation of the cases?

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
novak-vaclav: Added the check using `transposition_is_supported_on_neutron()` to `_is_supported_on_target()`.

backends/nxp/tests/ir/converter/node_converter/test_slice_tensor_converter.py
novak-vaclav: The bug is reported [here](https://jira.sw.nxp.com/browse/AIR-13665).

The bug occurs when a model contains a slice operator that slices multiple non-channel dimensions simultaneously. During conversion to the `aten` dialect, such slices are broken into several simpler ones—for example, a two-dimensional slice becomes two separate one-dimensional slices. When transposes are inserted around these slices, `neutron_converter` sometimes fails to merge them into a single accelerated `Neutron` node, even though documentation suggests it should. A concrete example is in the ticket.
Since this issue depends on operator sequences, it dont think it can be detected during individual node conversion.

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
```diff
@@ -33,8 +34,19 @@ def _is_supported_on_target(
             return False
 
         input_shape = input_tensor(node, 0).shape
+        input_rank = len(input_shape)
         dim = node.args[1]
 
+        # Slicing is only allowed along the channel dimension. 
+        # Therefore, we must verify that Neutron supports swapping the channel dimension 
+        # with the dimension intended for slicing.
+        if dim != -1 and dim != input_rank - 1:
```
MartinPavella, line 56: During conversion, the `dim` can be modified depending on the format of the input tensor. You need to consider the format here too. And the input shape will also be permuted during conversion in this case. So something like:
```python
if node.args[0].meta[NXP_NODE_FORMAT].is_channels_first():
    dim = translator.create_channels_last_to_channels_first_permutation(len(input_shape))[dim]
    channels_last_input_shape = translator.apply_permutation_to(
        input_shape, 
        translator.create_channels_first_to_channels_last_permutation(len(input_shape))
    )
```
should be inserted before you check the `Transpose` support.

MartinPavella (changes requested):

backends/nxp/backend/ir/converter/node_converters/ops_converters/slice_tensor_converter.py
novak-vaclav: fixed, thanks

MartinPavella (commented):
LGTM.
Once the tests start working (and pass), we can merge.

novak-vaclav:
Latest tests passed 🥳

MartinPavella:
> Latest tests passed 🥳

Only 3 checks were run. Meta has currently disabled their CI due to some firewall changes. We will have to wait for them to fix it.

In the meantime, feel free to keep this branch up-to-date with `main`.

MartinPavella:
Please update the branch to enable the tests.

MartinPavella (approved):

MartinPavella:
We will need to get back to this once Neutron 25_12 is released. But for now, we can merge.

PR #15903, opened 2025-11-20
Qualcomm AI Engine Direct - Refactor QNN Manager

shewu-quic:
## Summary:
- Include the contextmanager QnnManagerContext and QnnManagerRegistry to handle the life cycle management of QnnManager in AOT.
- Introduce a singleton QnnBackendUnifiedRegistry to manage shared QNN components.
- Separate the QnnManager Init function into InitBackend and InitContext to prevent creating the same backend and device for each context in AOT and runtime.
  - InitBackend should initialize shared QNN components such as QnnBackend, QnnDevice, QnnLogger, and the QNN library.
  - InitContext will be responsible for setting up QnnContext and QnnGraph.
- Resolved the outdated link in the documentation

## Background
- It takes some time to re-initialize the QNN components for both qnn_partitioner and qnn_preprocess in AOT.
  - We currently create and initialize a QnnManager in the qnn_partitioner, and then repeat this process for each partition in the qnn_preprocess. This could be improved by using a single QnnManager for both qnn_partitioner and qnn_preprocess to enhance lowering time.
- It takes some fixed time to do QNN component initialization during QNN ExecuTorch runtime, especially  to QNN device API.
  - We can use static initialization of the QNN device to improve model loading times when the model is initialized frequently.

## Goals and Objectives
- In AOT, make sure to load the QNN library and configure the QNN Backend and Device, ensuring these are shared between both qnn_partitioner and qnn_preprocess.
- During runtime, we can create one backend and device once in QnnExecutorchBackend and sharing for all partitions and multi-methods

cc: @haowhsu-quic


cccclai:
It seems like a big change, and I assume it's BC breaking?

shewu-quic:
> It seems like a big change, and I assume it's BC breaking?

Yes, it will result in BC breaking since I remove graph_name in compile_spec. I replace graph_name with method name.

cccclai:
Sorry for being late on reviewing this PR, can you rebase?

backends/qualcomm/runtime/QnnManager.h
```diff
@@ -30,7 +31,9 @@ class QnnManager {
       const QnnExecuTorchContextBinary& qnn_executorch_context_binary);
 
   ~QnnManager();
-  executorch::runtime::Error Init();
+  executorch::runtime::Error InitBackend();
```
cccclai, line 35: Can you add documentation for initbackend and init context?

backends/qualcomm/runtime/QnnManager.h
```diff
@@ -30,7 +31,9 @@ class QnnManager {
...
+  executorch::runtime::Error InitContext(
+      std::optional<std::vector<std::string>> graph_names = std::nullopt);
```
cccclai, line 40: Can I do something like
```
initContext(methods_a)
destroyContext(method_a)
initContext(method_b)
```

backends/qualcomm/runtime/QnnManager.h
```diff
@@ -47,7 +50,11 @@ class QnnManager {
       const std::string& graph_name,
       executorch::runtime::EventTracer* event_tracer);
 
+  // Destroy all QNN components and decrease reference count of shared QNN
+  // resource
   void Destroy();
+  // Only destroy non-shared QNN components
+  void DestroyContext();
```
cccclai, line 61: Does it destroy all context? It seems like the argument is differant than initContext

backends/qualcomm/runtime/backends/QnnBackendUnifiedRegistry.h
```diff
@@ -0,0 +1,106 @@
...
+namespace executorch {
+namespace backends {
+namespace qnn {
+
+// A bundle struct to hold all shared QNN backend resources
+struct QnnBackendBundle {
+  std::unique_ptr<QnnImplementation> implementation;
+  std::unique_ptr<QnnLogger> qnn_logger_ptr;
+  std::unique_ptr<QnnBackend> qnn_backend_ptr;
+  std::unique_ptr<QnnDevice> qnn_device_ptr;
```
cccclai, line 30: What's difference between Backend and Device?

backends/qualcomm/serialization/qc_compiler_spec.fbs
```diff
@@ -217,10 +213,6 @@ table QnnExecuTorchOptions {
   /// Optional backend specific options for the HTP backend.
   backend_options:QnnExecuTorchBackendOptions;
 
-  /// Optional parameter to create qnn graph if QNN context blob is not given
-  /// It could be a list of names only when doing weight-sharing lowering
-  graph_name:[string];
```
cccclai, line 268: Are we enforcing graph_name now so we are removing this?

backends/qualcomm/utils/utils.py
```diff
@@ -442,15 +436,15 @@ def ensure_graph_specific_dict(value, graph_names):
...
-
-    return to_edge_transform_and_lower(
-        aten_programs,
-        transform_passes=transform_passes,
-        partitioner=qnn_partitioners,
-        constant_methods=constant_methods,
-        compile_config=qnn_edge_config(),
-        generate_etrecord=generate_etrecord,
-    )
+    with QnnManagerContext(compiler_specs):
```
cccclai, line 441: Does it mean the lifecycle for the qnn manager will be alive with this context?

backends/qualcomm/runtime/QnnManager.h
shewu-quic: Yes, that's what I do in qnn_preprocess.py.

backends/qualcomm/runtime/QnnManager.h
shewu-quic: Sure!

backends/qualcomm/runtime/QnnManager.h
shewu-quic: I think the current naming could be improved. 
In QNN, several components need to be created, including QnnBackend, QnnDevice, QnnContext, and QnnGraph:
* QnnBackend handles op validation and custom op registration.
* QnnDevice is a software abstraction of a hardware accelerator platform, providing constructs needed to associate preferred hardware resources like platform info and power configuration.
* QnnContext allows interoperability between multiple graphs by offering a shared memory space for exchanging tensors, such as weight sharing and spill-fill buffers. It can also be cached as a serialized object for future use, like a context binary. Multiple contexts are created when there are multiple partitions.
* QnnGraph represents a loadable network model. When using the multi-method, multiple graphs are created, and each graph’s name corresponds to the method name.

QnnBackend and QnnDevice are expected to be shared across all contexts and graphs when using the same backend.

image

backends/qualcomm/serialization/qc_compiler_spec.fbs
shewu-quic: Because we use method name as graph_name.

backends/qualcomm/utils/utils.py
shewu-quic: Yes, that’s exactly what I expected.

backends/qualcomm/runtime/QnnManager.h
shewu-quic: For multi-method and multi-partition, it should be 
```
for each partition:
    # create multi graphs in one context
    initContext({methods_a, methods_b...}) 
    # compile graphs and serialized context binary
    compile({methods_a, methods_b...})
    # destroy context and graphs
    destroyContext()
```

cccclai (approved):

PR #15909, opened 2025-11-20
Arm backend: Move support_extension to base class

oscarandersson8218:
Move support_extension to TosaSpecification base class to avoid having to check whether the TosaSpecification is an instance of TosaSpecification_1_00.


zingo (approved):
Nice cleanup :)

PR #15912, opened 2025-11-20
pin bump optimum

Gasoonjia:
Summary: bump up optimum to keep sdpa kernel not being decomposed during lowering to cuda delegate

Differential Revision: D87519599


larryliu0820:
why jobs are canceled

larryliu0820 (approved):

larryliu0820 (approved):

PR #15917, opened 2025-11-20
Cortex-M backend: Fuse Relu, Hardtanh and Hardsigmoid

AdrianLundell:
Implements a new pass which fuses activation passes with preceeding cortex-m ops if possible.

Removed quantization of conv1d, conv3d as they are not tested
+ moves Conv+relu test to test_activations.

Propagate qmin, qmax to conv kernel.

### DIFF
--- a/backends/cortex_m/passes/activation_fusion_pass.py
+++ b/backends/cortex_m/passes/activation_fusion_pass.py
@@ -1,0 +1,170 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import logging
+
+import executorch.backends.cortex_m.ops.operators  # noqa: F401
+from executorch.backends.arm._passes.quant_args import QuantArgs
+
+from executorch.exir.dialects._ops import ops as exir_ops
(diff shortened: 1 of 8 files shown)


backends/cortex_m/passes/activation_fusion_pass.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2025 Arm Limited and/or its affiliates.
```
zingo, line 1: Add file to backends/cortex_m/passes/TARGETS ???
Or maybe use glob in that file?
    srcs = glob(["*.py"]),

zingo (changes requested):
Nice!

backends/cortex_m/passes/activation_fusion_pass.py
AdrianLundell: @metascroy I'm a bit uncertain if this is needed or not here, could you confirm/ pull it internally and see if anything breaks?

metascroy: Imported as D87791350.  Will let you know outcome of CI

metascroy:
Internal testing is good apart from OSS failures

zingo (approved):

PR #15918, opened 2025-11-20
Arm backend: Update bias quantization range

oscarandersson8218:
Bias range was [-2147483648, 2147483646] which isn't really symmetric. This patch changes the range to [-2147483647, 2147483647].


zingo (approved):

zingo:
Macos fails unrelated

PR #15919, opened 2025-11-20
[Executorch][llama] Split main in lib + binary

kimishpatel:
* __->__ #15919

Main reason to do so is to allow internal benchmarking binary to link against main_lib

Differential Revision: [D87547090](https://our.internmc.facebook.com/intern/diff/D87547090/)


examples/models/llama/targets.bzl
```diff
@@ -5,8 +5,8 @@ def define_common_targets():
         if not runtime.is_oss or not aten:
             aten_suffix = "_aten" if aten else ""
 
-            runtime.cxx_binary(
-                name = "main" + aten_suffix,
+            runtime.cxx_library(
+                name = "main" + aten_suffix + "_lib",
```
lucylq: nit: "main_lib" + aten_suffix

examples/models/llama/targets.bzl
```diff
@@ -25,3 +25,11 @@ def define_common_targets():
                 ],
                 **get_oss_build_kwargs()
             )
+
+            runtime.cxx_binary(
+                name = "main" + aten_suffix,
+                srcs = [],
+                deps = [
+                    ":main" + aten_suffix + "_lib",
```
lucylq: same here, `":main_lib" + aten_suffix,`

lucylq (approved):

examples/models/llama/targets.bzl
kimishpatel: makes sense

PR #15925, opened 2025-11-20
Tag scales for external data

lucylq:
### Summary
Currently scales are only saved to PTE segments (not PTD file). Allow them to be added to PTD as well.

### Test plan
Test with linear module quantized with torchao quantize_

### DIFF
--- a/backends/xnnpack/operators/node_visitor.py
+++ b/backends/xnnpack/operators/node_visitor.py
@@ -275,7 +275,7 @@ class NodeVisitor:
         return dtype
 
     def get_quant_params(
-        self, quant_params: QuantParams, xnn_graph: XNNGraph
+        self, quant_params: QuantParams, xnn_graph: XNNGraph, external_tag: str = None
     ) -> XNNQuantParams:
         if quant_params.per_channel:
             scale = cast(torch.Tensor, quant_params.scale)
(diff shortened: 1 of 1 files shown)


backends/xnnpack/operators/node_visitor.py
```diff
@@ -291,13 +291,18 @@ def get_quant_params(
                 ctypes.POINTER(ctypes.c_char * num_bytes),
             ).contents
             scale_name = hashlib.sha256(bytes(scale_array)).hexdigest()
+            scale_name = "scale_" + scale_name
```
lucylq, line 294: add this for debugging purposes

backends/xnnpack/operators/node_visitor.py
```diff
@@ -614,7 +623,7 @@ def get_serialized_buffer_index(
             f"Serializing constant data node {tensor} but tensor value has no bytes",
         )
         sha256_hash = hashlib.sha256(bytes(array))
-        named_key = sha256_hash.hexdigest()
+        named_key = tensor.name + "_" + sha256_hash.hexdigest()
```
lucylq, line 628: also debugging here

metascroy: Do you plan on landing this PR?

lucylq: @metascroy yeah - adding the tensor name to the hash for general debuggability until we can use names instead of hash

metascroy (approved):

PR #15807, opened 2025-11-13
Qualcomm AI Engine Direct - Quantization Recipe for LLM

DannyYuyang-quic:
### Summary

Qualcomm AI Engine Direct - Quantization Recipe for LLM 
 - add a fine-grained quantization annotation mechanism – quantization recipe
 - applied to LLM models with fine-grained quantization configs

### Test plan
All LLM CI under TestExampleLLMScript:
``` bash
python -m backends.qualcomm.tests.test_qnn_delegate.TestExampleLLMScript -s ${device_id} -H ${host_id} -m ${soc} -b build-android
```

PR #15926, opened 2025-11-20
forward fix unit test

cccclai:
Summary:
Forward fix test failure in https://github.com/pytorch/executorch/pull/15807

Main reason is that this API is called internally. In this PR, I recovered some of the deleted functions in the previous PRs

Differential Revision: D87566729


cccclai:
not needed

PR #15931, opened 2025-11-20
Fix fake mode detection

angelayi:
Differential Revision: D87577206


cccclai (approved):

zingo:
Hi this might have broken something in the Cortex-M backend. 
https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=cortex-m
@oscarandersson8218 has a fix ongoing that will remove the Cortex-M scalar tests from add and mul ops as they don't make sense.

PR #15932, opened 2025-11-20
Move user-defined passes after SpecPropPass

lucylq:
Summary:
1. Use `ExecutorchBackendConfig` instead of inserting the pass.
2. Move SpecPropPass after user-defined passes. It was originally swapped here: D80840528. cc ethansfng, who has a pass that should run before SpecPropPass

Differential Revision: D87576772


larryliu0820 (approved):

JacobSzwejbka (approved):

PR #15934, opened 2025-11-20
Add ETDump event tracer support to LLaMa runner

navsud:
Summary:
for op-level profiling of od-llms

**Addresses reviewer feedback:**
- Added `ET_EVENT_TRACER_ENABLED` ifdef guards around ETDump code to ensure normal builds compile without event tracer support
- Replaced hardcoded `/data/local/tmp/etdump.bin` path with configurable `--etdump_path` flag (default: `"etdump.in"`)
- ETDumpGen is created only when compiled with event tracer support enabled

Differential Revision: D87122487

### DIFF
--- a/examples/models/llama/main.cpp
+++ b/examples/models/llama/main.cpp
@@ -7,11 +7,14 @@
  * @lint-ignore-every CLANGTIDY facebook-hte-Deprecated
  */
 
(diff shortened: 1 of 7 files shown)


larryliu0820 (approved):

kimishpatel (approved):

PR #15936, opened 2025-11-21
[Do not land] Static numeric experiment

metascroy:
Here we do experiments to see why iOS26 produces bad numerics (see https://github.com/pytorch/executorch/issues/15833).

Export with:

```
 python export.py -n model2.pte -p $HOME/models/llama1b/params.json -c $HOME/models/llama1b/llama1b.pth --seq_length 16 --max_seq_length 512 --dtype fp16 --coreml-quantize custom &> log.txt
```

Run with:
```
python run.py -m model2.pte -t $HOME/models/llama1b/tokenizer.model --prompt "Once upon a time,"
```

What we've found:
* Quantizing embeddings seems to lead to bad results on iOS26
* Using CoreMLRMSNorm instead of RMSNorm produces better numerics in FP16 on iOS26
* Quantizing q, k, v, o, and lm_head on iOS26 seems OK
* Quantizing w1, w1, or w3 on iOS26 leads to bad numerics.  It seems to be related to the 8096 matrix size.  If we use in/out feature splitting on those linear layers, we can recover good numerics.

Even after all of the above changes, on my mac CoreML is scheduling this model to run on CPU.  Have yet to try on iPhone.

**Update:**

We can quantize everything with somewhat OK text generation if we decompose SDPA into matmuls + softmax.  So the key changes are the following settings in llama_transformer.py:
```
_RMS_NORM = CoreMLRMSNorm
_DECOMPOSE_SDPA = True
_USE_SOFTMAX = True
_USE_SILU_APPROX = False
```

Repro commands:

Export as one model:

```
python export.py -n model2.pte -p $HOME/models/llama1b/params.json -c $HOME/models/llama1b/llama1b.pth --seq_length 16 --max_seq_length 512 --dtype fp16 --coreml-quantize c4w -E "8,0" --target_split_size 1024 --max_splits 8 --export_in_pieces_mode 0 &> log.txt
python run.py -m model2.pte -t $HOME/models/llama1b/tokenizer.model --prompt "Once upon a time,"
```

Export as collection of models:
```
python export.py -n model2.pte -p $HOME/models/llama1b/params.json -c $HOME/models/llama1b/llama1b.pth --seq_length 16 --max_seq_length 512 --dtype fp16 --coreml-quantize c4w -E "8,0" --export_in_pieces_mode 1 &> log.txt
python run.py -m model2 -t $HOME/models/llama1b/tokenizer.model --prompt "Once upon a time,"
```

### DIFF
--- a/examples/apple/coreml/llama/export.py
+++ b/examples/apple/coreml/llama/export.py
@@ -28,6 +28,7 @@ from executorch.extension.export_util.utils import save_pte_program
 
 from torchao.quantization.granularity import PerAxis, PerGroup
 from torchao.quantization.quant_api import IntxWeightOnlyConfig, quantize_
+from torchao.prototype.quantization.codebook_coreml import CodebookWeightOnlyConfig
 from torchao.utils import unwrap_tensor_subclass
 
 
@@ -77,7 +78,7 @@ def main() -> None:
     parser.add_argument(
         "--coreml-quantize",
         default=None,
-        choices=["b4w", "c4w"],
+        choices=["b4w", "c4w", "custom",],
         help="This option is only for coreml: Use coreml quantization, e.g. b4w (for blockwise 4 bit weight), c4w (for channelwise 4 bit weight)",
     )
     parser.add_argument(
@@ -117,6 +118,8 @@ def main() -> None:
 
     model.eval()
     model.to(float_dtype)
+
+    print("MODEL", model)
 
     if export_args.target_split_size is not None:
         replace_linear_with_split_linear(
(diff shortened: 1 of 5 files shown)


examples/apple/coreml/llama/export.py
```diff
@@ -163,6 +166,34 @@ def main() -> None:
...
+        )
+        quantize_(
+            model,
+            IntxWeightOnlyConfig(
+                weight_dtype=torch.int4,
+                granularity=PerAxis(0),
+            ),
+            lambda m, fqn: (
+                isinstance(m, torch.nn.Linear) 
+                and any(fqn.endswith(suffix) for suffix in ["wq", "wk", "wv", "wo", "output", "w1", "w3", "w2"])
```
metascroy: This needs to be suffix in fqn instead of fqn.endswith(suffix) in order to capture the split linear modules.  This is because once split linear is applied, the fqn no longer ends with w1, w2, and w3.

PR #15944, opened 2025-11-21
Expose get_num_threads via pybind

kimishpatel:
Summary:
Main objective is to allow querying number of threads being used from python.

Later diff uses this mainly for testing.

Reviewed By: larryliu0820

Differential Revision: D87510750


larryliu0820 (approved):

PR #15951, opened 2025-11-22
Decompose after export in export_llama

lucylq:
### Summary
`unwrap_tensor_subclass` was not unwrapping nested lora linears. This meant qdata/scale/zero were bundled together in the subclass, and separated at run decompositions inside to_edge_transform_and_lower. This is after nodes are tagged, meaning that the scales were not tagged, and remained in the PTE file after the rest of the weights were moved to a PTD file.

It's recommended to move away from `unwrap_tensor_subclass` and rely on export + decomps. This PR adds a decomp after exporting in export_llama, and removes cases of `unwrap_tensor_subclass`. 

TODO: remove all cases of `unwrap_tensor_subclass` in ET.

### Test plan
Add check that quantized weights are in PTD file (not PTE file) after quantization. This is a simple check, nested linears seem to be the real issue that decomposing resolves. TODO to add a test for that (probably e2e test with stories in subsequent PR)
```
 python -m unittest executorch.backends.xnnpack.test.passes.test_propagate_custom_meta_pass
```


metascroy:
LGTM!  Address any failing CI tests before landing.

metascroy (approved):

PR #15959, opened 2025-11-24
Adding a payload version to become compatible with new Neutron Converter

jirioc:
### Summary
The new neutron converter introduces extra output tensors. This requires changes in the backend. These changes are backward compatible, meaning that the PTE converted by the older versions can be run by the new NeutronBackend.cpp, but not vice versa.

### Test plan
The unit tests have been adjusted to accommodate the new payload format.

### DIFF
--- a/backends/nxp/neutron_node_extraction.py
+++ b/backends/nxp/neutron_node_extraction.py
@@ -21,6 +21,7 @@ class NeutronNodeArtifacts:
     microcode: np.ndarray
     weights: np.ndarray
     kernels: np.ndarray
+    payload_version: int
 
 
 def extract_artifacts_from_neutron_node(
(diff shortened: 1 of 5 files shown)


backends/nxp/neutron_node_extraction.py
```diff
@@ -123,7 +124,11 @@ def extract_artifacts_from_neutron_node(
     output_names = []
     output_indices = []
     graph_outputs = sub_graph.OutputsAsNumpy()
+    payload_version = 0
     node_outputs = neutron_node.OutputsAsNumpy()[:-1]
+    if len(graph_outputs) == len(node_outputs) - 2:
```
MartinPavella, line 130: Could you please add a comment explaining the `- 2`?

backends/nxp/tests/test_neutron_backend.py
```diff
@@ -36,7 +36,10 @@ def test_neutron_backend__single_conv_model__payload_header_channels_last():
     assert payload[4] == 0x1  # Channels last 0-th Neutron output
     assert payload[5] == 0x0  # Map 0-th Neutron input to 0-th model input
     assert payload[6] == 0x0  # Map 0-th Neutron output to 0-th model output
-    assert all(byte == 0x0 for byte in payload[7:16])  # Aligned to 16 bytes
+    assert (
+        payload[7] == 0x0 or payload[7] == 0x1
```
MartinPavella, line 40: Do you think it would be useful to use `NeutronTargetSpec` to determine the expected payload version for a stricter check?

MartinPavella (approved):

backends/nxp/tests/test_neutron_backend.py
jirioc: The payload version is not dependent on Target, it depends on neutron converter version.

backends/nxp/neutron_node_extraction.py
jirioc: I added a comment explaining all 3 extra tensors.

roman-janik-nxp (approved):

PR #15967, opened 2025-11-24
Portable aoti

JacobSzwejbka:
Dynamically link against cuda runtime at the lib level since the aoti blob dynamically links and its probably bad if they end up being different versions. Also add the setting so that the aoti blob generates a fatbin. 

Not totally sure how to add CI for this one since we really only expect portability of aoti blobs from a weaker gpu to a stronger one. Not super sure what our access to machines looks like in CI maybe can figure it out as I add windows CI tomorrow.


backends/cuda/CMakeLists.txt
```diff
@@ -28,6 +28,9 @@ if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
+# Use dynamic linking for CUDA runtime
+set(CUDA_USE_STATIC_CUDA_RUNTIME OFF)
```
larryliu0820, line 32: Is this for CUDA to consume?

larryliu0820 (approved):

backends/cuda/CMakeLists.txt
JacobSzwejbka: like their cmake? Yes.

examples/models/whisper/CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(EXECUTORCH_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../..")
```
JacobSzwejbka, line 13: Just made this file look like voxtrals

PR #15969, opened 2025-11-25
custom comparator support in calculate_numeric_gap

Gasoonjia:
Summary: Enable users to define their own metrics for numeric gap.

Differential Revision: D87840446


kimishpatel (approved):

devtools/inspector/_inspector.py
```diff
@@ -1416,7 +1419,10 @@ def calculate_numeric_gap(
         compare the intermediate outputs from the AOT and the runtime.
 
         Args:
-            distance: The metrics the inspector will use for gap calculation. Should be one of "MSE", "L1" and "SNR".
+            distance: The metrics the inspector will use for gap calculation. Can be either:
+                - A string: one of "MSE", "L1", or "SNR" for built-in comparators.
```
winskuo-quic, line 1423: Should this be `- A string: one of "MSE", "L1", "SNR", or user-defined comparators`?

winskuo-quic (commented):
Thanks for supporting this feature and providing users more flexibility!

devtools/inspector/_inspector.py
Gasoonjia: i separated the bulit-in comparators and custom comparators into two lines

PR #15971, opened 2025-11-25
NXP backend: added aten.mul support

novak-vaclav:
### Summary

adds support for aten.mul operator

### Test plan


backends/nxp/quantizer/patterns.py
```diff
@@ -673,6 +673,47 @@ def get_anchors(
...
+        )
+
+        # appending custom qspec before the "Mul" operator caused two unnecessary dequantize-quantize sequences,
+        # which prevented the partitioner from functioning properly. This way I set only one dequantize-quantize sequence
+        # with correct quantization parameters
+        for input_node in input_nodes:
+            input_node.meta["quantization_annotation"].output_qspec = qspec
+
+        return PartitionAnchors(
+            inputs=[(node, NodeArgsIdx(0), qspec), (node, NodeArgsIdx(1), qspec)],
```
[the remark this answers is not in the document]
novak-vaclav: does not apply here

backends/nxp/quantizer/patterns.py
```diff
@@ -673,6 +673,47 @@ def get_anchors(
...
+            scale=1.0 / 256.0,
+            zero_point=0,
+            quant_min=-128,
+            quant_max=127,
+            qscheme=torch.per_tensor_affine,
+        )
+
+        # Directly setting output_qspec on input nodes avoids inserting redundant dequantize-quantize sequences
+        # before the "Mul" operator. This ensures only one dequantize-quantize sequence is present,
+        # with the correct quantization parameters, allowing the partitioner to function properly.
```
MartinPavella: I think this comment should be re-phrased. Consider something like:

The `Mul` operator in Neutron IR requires a specific scale and zero_point (defined above) for its inputs. Since these input nodes have already been annotated by their own patterns which didn't take the requirements of `Mul` into account, we need to overwrite the existing `quantization_annotation`.

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
```diff
@@ -0,0 +1,230 @@
...
+        model, [x_input_shape, x_input_shape]
+    ).exported_program()
+    edge_nodes = list(edge_program.graph.nodes)
+
+    # Check "Mul" was delegated
+    mul_delegated = edge_nodes[5]
+    assert (
+        mul_delegated.op == "call_function"
+        and mul_delegated.target.__name__ == "executorch_call_delegate"
+    )
```
MartinPavella: A better way to check a node was delegated is to make sure that none of the nodes in the model are `mul`. For example:
```python
assert not any("mul" in n.name for n in edge_nodes)
```

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
```diff
@@ -0,0 +1,230 @@
...
+    [
+        pytest.param((1, 4, 16, 16), id="4D."),
+        pytest.param((1, 4, 5, 5), id="4D, product of dims is not a multiple of 8."),
+    ],
+)
+def test_mul_tensor_w_conv_quant_conversion(mocker, x_input_shape):
+    model = MulTensorConvModule()
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+
+    c, _, w, h = x_input_shape
```
MartinPavella: Is this correct? What sort of format are you using?
I would expect:
```python
n, c, h, w = x_input_shape
```

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
```diff
@@ -0,0 +1,230 @@
...
+        model, [x_input_shape, y_input_shape], use_neutron_for_format_conversion=False
+    ).exported_program()
+    edge_nodes = list(edge_program.graph.nodes)
+
+    # Check "Mul" was delegated
+    mul_delegated = edge_nodes[5]
+    assert (
+        mul_delegated.op == "call_function"
+        and mul_delegated.target.__name__ == "executorch_call_delegate"
+    )
```
MartinPavella: Minor:
You should check that both `mul` and `convolution` are delegated within the same partition. I know the code on lines `189-190` does that, but at least the comment here should reflect it too. Or perhaps you could also add an assertion for the length of the `edge_nodes`.

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
```diff
@@ -0,0 +1,230 @@
...
+    assert exported_nodes[12].target == exir_ops.edge.aten.convolution.default
+    assert exported_nodes[15].target == exir_ops.edge.aten.mul.Tensor
+
+    conversion_config = ConversionConfig()
+    conversion_config.use_neutron_for_format_conversion = False
+    convert_run_compare(
+        exported_program,
+        input_data=input_data,
+        tfl_model=tflite_flatbuffers_model,
+        conversion_config=conversion_config,
```
MartinPavella: Minor:
The use of the `conversion_config` is not necessary, as you are providing an argument for the `tfl_model` parameter.

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
```diff
@@ -0,0 +1,230 @@
...
+@pytest.mark.parametrize(
+    "x_input_shape, y_input_shape",
+    [
+        pytest.param((1, 4, 7), (4, 7), id="3D -> 2D."),
+        pytest.param((4, 4, 8), (1, 4, 4, 8), id="3D -> 4D."),
+        pytest.param((1, 1, 4, 4, 8), (1, 4, 4, 8), id="5D -> 4D."),
+        pytest.param((4,), (1, 4), id="1D -> 2D."),
+        pytest.param((4,), (1, 1, 4), id="1D -> 3D."),
+        pytest.param((6, 6), (1, 1, 6, 6), id="2D -> 4D."),
+        pytest.param((1, 6), (6,), id="2D -> 1D."),
```
MartinPavella, line 195: Minor:
Using so many tests for the broadcasting is not necessary. One test case is enough to test that you are calling the 
```python
node_uses_shape_broadcasting(node)
```
in your converter. Using multiple test cases only verifies the correct behavior of the function `node_uses_shape_broadcasting()`, which has been tested thoroughly on many occasions before.

My suggestion is to keep only 1 test case (for example `pytest.param((8, 8, 8), (1, 8, 8, 8), id="3D -> 4D.")`), to minimize the run-time of our tests.

MartinPavella (approved):
I only have small suggestions for improvements. Please have a look at them. 
Otherwise, well done 👍🏻

MartinPavella:
A general tip: 
We prefer to use `rebase` instead of `merge` to keep up with the base branch (`main`). It keeps the history nicer, and the `lintrunner` runs significantly faster.

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
novak-vaclav: Yeah thats a typo, thanks

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
novak-vaclav: I think it is a good idea to keep at least two, one going from higher dim to lower dim and the second going from lower dim to higher dim. Hope that is fine by you.
I removed all other test cases.

backends/nxp/tests/ir/converter/node_converter/test_mul_tensor_converter.py
novak-vaclav: Added a test to check that `convolution` is not present. Also modified the test to be more robust.

novak-vaclav:
> A general tip: We prefer to use `rebase` instead of `merge` to keep up with the base branch (`main`). It keeps the history nicer, and the `lintrunner` runs significantly faster.

I had a discussion about it with @robert-kalmar and in the next PRs, I will try to keep a more structured commit history.

novak-vaclav:
All issues found by @MartinPavella are fixed.

MartinPavella:
Please update the branch to enable the tests.

PR #15977, opened 2025-11-25
Arm backend: Add support for int32 clamp

ArmRyan:
Tosa does not support int32 dtype for clamp operator
 so instead we need to use Min/Max to implement clamp.

Change-Id: Iea442901b2227610ebb5e5a0f1bca8d236e70d9d


oscarandersson8218 (changes requested):
I think this should be done in a pass instead.

ArmRyan:
> I think this should be done in a pass instead.

Done

oscarandersson8218 (approved):
Failures seem unrelated.

PR #15979, opened 2025-11-25
docs: Add logging guidance for debug partition

### LINKED ISSUE #15914: Matrix multiplication not delegated to backend during export (Whisper)
### 🐛 Describe the bug

During the export of **whisper-tiny.en** model to ExecuTorch format, I encountered the following problem: one of the matrix multiplication operations is not being delegated to the XNNPACK backend, causing a significant inference speed slowdown (up to 50% of the whole inference time).

The operation is located within the forward method of Decoder module from *openai-whisper* package:
```
def forward(self, x: Tensor, xa: Tensor, kv_cache: Optional[dict] = None):
        """
        x : torch.LongTensor, shape = (batch_size, <= n_ctx)
            the text tokens
        xa : torch.Tensor, shape = (batch_size, n_audio_ctx, n_audio_state)
            the encoded audio features to be attended on
        """
        offset = next(iter(kv_cache.values())).shape[1] if kv_cache else 0
        x = (
            self.token_embedding(x)
            + self.positional_embedding[offset : offset + x.shape[-1]]
        )
        x = x.to(xa.dtype)

        for block in self.blocks:
            x = block(x, xa, mask=self.mask, kv_cache=kv_cache)

        x = self.ln(x)
        logits = (
            # ISSUE: the below matrix multiplication is not being delegated
            x @ torch.transpose(self.token_embedding.weight.to(x.dtype), 0, 1)
        ).float()

        return logits
```

### Steps to reproduce
I use the following code to export the module with static shapes:
```
import torch
import whisper
from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
from executorch.exir import to_edge_transform_and_lower

model = whisper.load_model("tiny.en").decoder
model.eval()

inputs = (
  torch.randint(1, 100, (1, 128), dtype=torch.int),
  torch.randn(1, 1500, 384, dtype=torch.float32)
)

# Export
exported_program = torch.export.export(model, inputs)

executorch_program = to_edge_transform_and_lower(
    exported_program,
    partitioner = [XnnpackPartitioner()]
).to_executorch()

with open("whisper/exported/decoder.pte", "wb") as file:
    executorch_program.write_to_file(file)
```
To profile the exported model, I follow the instructions from [ExecuTorch docs](https://docs.pytorch.org/executorch/stable/runtime-profiling.html).

### Actual behavior
A single native_call_mm.out operation not delegated to the XNNPACK backend is responsible for approximately 50% of the inference time. 
The profiling results are available [here](https://gist.github.com/IgorSwat/e10b501b42e43fc00f0326d3e0b2bbe0).

### What I tried
- Using other matrix multiplication function (`matmul()` and `mm()`) - no effect
- Replacing matrix multiplication with an equivalent nn.Linear call with static shapes - resolves the delegation issue, and produces a speedup from approximately 145 ms to 80 ms inference time (~45%).

### Versions

PyTorch version: 2.10.0.dev20250916
Is debug build: False
CUDA used to build PyTorch: None

OS: macOS 15.7.1 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.3.19.1)
CMake version: version 4.1.2
Libc version: N/A

Python version: 3.12.11 (main, Jun  3 2025, 15:41:47) [Clang 17.0.0 (clang-1700.0.13.3)] (64-bit runtime)
Python platform: macOS-15.7.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M1 Pro

[pip3] executorch==1.0.0.dev20250916
[pip3] mypy_extensions==1.1.0
[pip3] numpy==2.3.4
[pip3] optimum-executorch==0.2.0.dev0
[pip3] pytorch_tokenizers==0.1.0
[pip3] torch==2.10.0.dev20250916
[pip3] torchao==0.14.0.dev20250916+cpu
[pip3] torchaudio==2.8.0.dev20250916
[pip3] torchvision==0.25.0.dev20250916
[pip3] openai-whisper==20250625

yujiaoliang:
### Summary
This PR adds instructions for enabling internal debug logging in the  partitioner.
The logging can help users quickly identify why certain nodes are not delegated.

Fixes #15914 

cc @mergennachin @byjlw @GregoryComer @digantdesai @cbilgin

### DIFF
--- a/docs/source/backends/xnnpack/xnnpack-troubleshooting.md
+++ b/docs/source/backends/xnnpack/xnnpack-troubleshooting.md
@@ -23,3 +23,27 @@ The XNNPACK backend is built by default for Python, Android, iOS, and in most CM
  * Use [op-level profiling](/tutorials/devtools-integration-tutorial) to understand which operators are taking the most time. <!-- @lint-ignore linter doesn't like this link for some reason -->
    * The XNNPACK backend provides operator-level timing for delegated operators.
  * See general performance troubleshooting tips in [Performance Troubleshooting](/using-executorch-faqs.md#inference-is-slow-performance-troubleshooting).
+
+
+## Debugging Why Nodes Are Not Partitioned
+
+* To debug cases where operators are not delegated to XNNPACK,
+you can enable internal debug logs before **to_edge_transform_and_lower** from the partitioner. This will print diagnostic messages explaining why specific nodes fail
+to partition.
+
+``` python
+
+# caption: Enable internal partition debug logging
+
+import logging
+
+logger = logging.getLogger("executorch.backends.xnnpack.partition")
+logger.setLevel(logging.DEBUG)
+
+if not logger.handlers:
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.DEBUG)
+    formatter = logging.Formatter("[%(levelname)s] %(name)s: %(message)s")
+    ch.setFormatter(formatter)
+    logger.addHandler(ch)
+```


yujiaoliang:
Hi @cccclai , just a gentle ping on this PR.
Please take a look when convenient. Thanks!

README.md
```diff
@@ -73,6 +73,18 @@ import torch
 from executorch.exir import to_edge_transform_and_lower
 from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
 
+# 0、enable log for partition debug
```
cccclai:
```suggestion
# 0、[Optional] enable log for partition debug
```

cccclai:
I'm not sure if we want to have it in the top readme. Thoughts? @GregoryComer @mergennachin

GregoryComer:
@yujiaoliang Thanks for contributing. What do you think about adding to the troubleshooting page here: https://docs.pytorch.org/executorch/stable/backends/xnnpack/xnnpack-troubleshooting.html?

Here's the corresponding .md file: https://github.com/pytorch/executorch/blob/1864fb09c6412473d9712ba3d81e320674daad57/docs/source/backends/xnnpack/xnnpack-troubleshooting.md.

yujiaoliang:
> @yujiaoliang Thanks for contributing. What do you think about adding to the troubleshooting page here: https://docs.pytorch.org/executorch/stable/backends/xnnpack/xnnpack-troubleshooting.html?
> 
> Here's the corresponding .md file: https://github.com/pytorch/executorch/blob/1864fb09c6412473d9712ba3d81e320674daad57/docs/source/backends/xnnpack/xnnpack-troubleshooting.md.

Thanks for the suggestion. I agree that adding this information to the troubleshooting page would keep the docs structure cleaner. I'll update the PR later today and move the content into `xnnpack-troubleshooting.md`. Happy to further refine it if needed.

cccclai (approved):
Thank you!

PR #15980, opened 2025-11-25
Arm backend: Add support for aten.clone.default

ArmRyan:
Add clone to tosa identity op factory. Only present with using
 _skip_dim_order flag.

Change-Id: Ieda1990fa2d1a9627d821db122d201c032dc517b


oscarandersson8218:
@AdrianLundell can you have a look at this and make sure that this is aligned with our dim-order/memory-format handling?

AdrianLundell:
I have a few questions:
1. How is the memory_format arg of clone handled here?
2. Why are we supporting skip_dim_order=True? I was under the impression that it isn't supposed to be used.
3. Why do we have two ways of handling noops? (remove_noop_pass vs ops_identity, I prefer the pass here).

PR #15982, opened 2025-11-25
Fix -Wimplicit-const-int-float-conversion error.

hsharma35:
Summary: as titled

Reviewed By: skrtskrtfb

Differential Revision: D87882617

### DIFF
--- a/backends/cadence/vision/kernels/kernels.cpp
+++ b/backends/cadence/vision/kernels/kernels.cpp
@@ -11,7 +11,6 @@
 #include <algorithm>
 #include <cstring>
 #include <limits>
-#include <numeric>
 
 namespace impl {
 namespace vision {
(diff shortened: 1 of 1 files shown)


zonglinpeng (approved):

backends/cadence/vision/kernels/kernels.cpp
```diff
@@ -25,8 +25,8 @@ void* allocate_temp_memory(KernelRuntimeContext& ctx, size_t size) {
 // Quantize a fp32 value to an int8_t/uint8_t value
 template <typename T>
 T quantize(const float x, float scale, int32_t zero_point) {
-  constexpr float min_val = std::numeric_limits<T>::min();
-  constexpr float max_val = std::numeric_limits<T>::max();
+  constexpr float min_val = static_cast<float>(std::numeric_limits<T>::min());
```
msluszniak: Same where applicable
```suggestion
  constexpr auto min_val = static_cast<float>(std::numeric_limits<T>::min());
```

hsharma35: This seems to go against executorch style guide
https://google.github.io/styleguide/cppguide.html#Type_deduction

msluszniak: I will interpret this as a second point from this screen (goes from the google style guide you've sent):

Screenshot_20251128_094820_Chrome

hsharma35: I see. I'm sticking to the decision section below the text you shared about the pros/cons.

> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type.

Saving one character (float -> auto) does not make the code more readable, IMO.

PR #15984, opened 2025-11-25
Make ExecuTorch CoreML asset/trash/databse folders configurable.

metascroy:
Summary:
The ExecuTorch CoreML backend uses various folder locations to store compiled assets, to trash old assets, and to store a databse.

With this diff, all of these are configurable with the following compile-time flags:

* EXECUTORCH_COREML_ASSETS_DIRECTORY_PATH (default value is [cache]/executorchcoreml)
* EXECUTORCH_COREML_TRASH_DIRECTORY_PATH (default value is [tmp]/executorchcoreml)
* EXECUTORCH_COREML_DATABASE_DIRECTORY_PATH (default value is [app support]/executorchcoreml)

Differential Revision: D85064722

### DIFF
--- a/backends/apple/coreml/runtime/delegate/ETCoreMLStrings.mm
+++ b/backends/apple/coreml/runtime/delegate/ETCoreMLStrings.mm
@@ -101,39 +101,50 @@
 }
 
 + (nullable NSString *)assetsDirectoryPath {
-    static dispatch_once_t onceToken;
-    static NSString *result = nil;
-    dispatch_once(&onceToken, ^{
-        NSArray<NSString *> *paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
-        if (paths.count > 0) {
-            result = [paths.lastObject stringByAppendingPathComponent:self.productName];
-        }
-    });
-    
-    return result;
+    #if defined(EXECUTORCH_COREML_ASSETS_DIRECTORY_PATH)
+        return @(EXECUTORCH_COREML_ASSETS_DIRECTORY_PATH);
+    #else
+        static dispatch_once_t onceToken;
+        static NSString *result = nil;
+        dispatch_once(&onceToken, ^{
+            NSArray<NSString *> *paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+            if (paths.count > 0) {
+                result = [paths.lastObject stringByAppendingPathComponent:self.productName];
+            }
+        });
+        
+        return result;
+    #endif
 }
 
 + (nullable NSString *)trashDirectoryPath {
-    static dispatch_once_t onceToken;
-    static NSString *result = nil;
-    dispatch_once(&onceToken, ^{
-        result = [NSTemporaryDirectory() stringByAppendingPathComponent:self.productName];
-    });
-    
-    return result;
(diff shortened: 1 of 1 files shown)


backends/apple/coreml/runtime/delegate/ETCoreMLStrings.mm
```diff
@@ -101,39 +101,50 @@ + (NSString *)debugSymbolToHandlesKeyName {
...
-    static NSString *result = nil;
-    dispatch_once(&onceToken, ^{
-        NSArray<NSString *> *paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
-        if (paths.count > 0) {
-            result = [paths.lastObject stringByAppendingPathComponent:self.productName];
-        }
-    });
-    
-    return result;
+    #if defined(EXECUTORCH_COREML_ASSETS_DIRECTORY_PATH)
```
cccclai, line 104: Is it like runtime feature? What will happen if I use one path for one run and use a different path for another run?

backends/apple/coreml/runtime/delegate/ETCoreMLStrings.mm
```diff
@@ -101,39 +101,50 @@ + (NSString *)debugSymbolToHandlesKeyName {
...
+        return @(EXECUTORCH_COREML_ASSETS_DIRECTORY_PATH);
+    #else
+        static dispatch_once_t onceToken;
+        static NSString *result = nil;
+        dispatch_once(&onceToken, ^{
+            NSArray<NSString *> *paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+            if (paths.count > 0) {
```
shoumikhin, line 111: nit: technically, neither of `paths.count > 0` checks are needed

backends/apple/coreml/runtime/delegate/ETCoreMLStrings.mm
metascroy: This is a compile time flag for the CoreML backend, not a feature that can be changed between predictions.   If the flag is defined at compile time, users can control what directories are used for model caches.  If not defined, a default folder is used.

cccclai: If I do path = A and model caches will be there. Then compile again with path = B, what will happen for path A? Will the cache stay as is or clear out?

metascroy: The cache will stay there.  On next model run, we will get a cache miss and it will write model to updated path B.

cccclai: I see, will they last there forever or will OS try to clean it up? I'm just worried that there might be caches in the system and no one can find them...

metascroy: Whether the OS cleans them up depends on what folder they live in.  The default folder if these compile flags are NOT set (cache folder) will be cleaned up by the OS under disk pressure, but if a app developer sets these paths to something specific in the app container, they will have to be careful about cleaning up old model paths.  But I don't know what you're proposing we do about that.  Setting these is an advanced feature that some internal teams have requested.

Note that on iOS everything is isolated in an app container, so if the user uninstalls the app, the model caches are cleared no matter where they are stored.

cccclai (approved):
Seems fine to me as it's a requested feature

PR #15992, opened 2025-11-26
Arm Backend: improve non-persistent placeholder and bool handling

robell:
Rework bool handling pass to make it more generic.
* Improving the pass to handle type promotion in cases where one operand is not bool.
* Rename bool promotion pass
* Add tests for promote bool operands pass

Handle non-persistent buffer placeholders.


robell:
@zingo - do we want release notes: arm? nothing notable for release notes in here i don't think so i left it off.

If you don't mind can you also approve it as reviewed? - already gone via internal code review as usual. should then just be merge-able if tests pass. Though i moved a file so i don't know if buck related updates are needed?

zingo:
> @zingo - do we want release notes: arm? nothing notable for release notes in here i don't think so i left it off.
> 
> If you don't mind can you also approve it as reviewed? - already gone via internal code review as usual. should then just be merge-able if tests pass. Though i moved a file so i don't know if buck related updates are needed?

You/we can set Release note: none in that case

zingo:
I usually review and merge at the same time when all arm test and CI looks OK, so I have not looked at this one yet. I assume non of the other reviewers have either.

oscarandersson8218 (approved):
Looks good to me.

PR #15993, opened 2025-11-26
Arm backend: Add support for quant. decomposition

oscarandersson8218:
For mixed type models we need to be able switch between FP and INT,
meaning quantize and dequantize at runtime. As there are no quantize or
dequantize operators in TOSA, we need to decompose these operators to
TOSA operators. This commit introduces a pass that decomposes q-dq nodes
into more primitive nodes. 

This also affects Cortex-M backend as it uses the qdq-folding pass in arm/_passes.


metascroy:
There are no internal failures.  Please feel free to review/merge

zingo (approved):

Issue #15732, opened 2025-11-11
QNN backend fails on second model load (DMA-BUF preregistration issue)


nambn007:
### 🐛 Describe the bug

Hi, I am currently facing an issue when using ExecuTorch with the QNN backend on Android.
The model loads and runs correctly the first time, but after I unload (reset) the model and try to load it again within the same app session, the QNN backend fails with the error PreRegisterMem failed to get file descriptor / Fail to initialize Qnn Manager.

```
PreRegisterMem failed to get file descriptor.
Fail to pre register custom memory handle
Fail to initialize Qnn Manager

```

```
#if defined(__aarch64__)
  ET_CHECK_OR_RETURN_ERROR(
      PreRegisterMem() == Error::Ok,
      Internal,
      "Fail to pre register custom memory handle");
#endif
```

This happens specifically after QNN tries to preregister external memory (DMA-BUF / AHardwareBuffer).

### Versions

SM8550

ExecuTorch version: 1.0 / Nightly (please update if different)

Backend: QNN (HTP / DSP)

Device: Qualcomm Snapdragon (CDSP)

OS / Platform: Android (via JNI)

Model: LLaMA .pte (Hybrid KV cache mode enabled)

cc @cccclai @winskuo-quic @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

cccclai:
@haowhsu-quic @shewu-quic @winskuo-quic @DannyYuyang-quic can we have someone to take a look at this?

allerou4:
No more logs or QNN error codes?

shewu-quic:
Thank you for your effort. This error is actually due to some legacy code, which we no longer use. I believe we can remove it.
https://github.com/pytorch/executorch/blob/b1e3e28bb611e06d484138be27221faffd89f565/backends/qualcomm/runtime/QnnManager.cpp#L359

nambn007:
Thanks for your reply. Do you have any plans to update or improve this part in the future?

> Thank you for your effort. This error is actually due to some legacy code, which we no longer use. I believe we can remove it.
> 
> [executorch/backends/qualcomm/runtime/QnnManager.cpp](https://github.com/pytorch/executorch/blob/b1e3e28bb611e06d484138be27221faffd89f565/backends/qualcomm/runtime/QnnManager.cpp#L359)
> 
> Line 359 in [b1e3e28](/pytorch/executorch/commit/b1e3e28bb611e06d484138be27221faffd89f565)
[...]

PR #16000, opened 2025-11-27
Qualcomm AI Engine Direct - Remove legacy code related to the shared buffer

shewu-quic:
Summary:
- Remove PreRegisterMem, as it is legacy code used for registering custom memory before execution.

cc: @haowhsu-quic

### DIFF
--- a/backends/qualcomm/runtime/SharedBuffer.h
+++ b/backends/qualcomm/runtime/SharedBuffer.h
@@ -59,18 +59,9 @@ class SharedBuffer final {
   // memory handle is registered during execution
   void AddCusomMemTensorAddr(void* tensor_addr, void* custom_mem);
 
-  // memory handle can be registered before execution
-  void AddCusomMemTensorInfo(const CustomMemTensorInfo& info);
-
   size_t GetAllocatedSize(void* buf);
 
   void* GetCustomMemBase(void* buf);
-
-  void* GetUnAlignedAddr(void* buf);
-
-  const std::unordered_set<CustomMemTensorInfo>& GetCustomMemTensorInfoSet() {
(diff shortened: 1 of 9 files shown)


shewu-quic:
Hi @cccclai,

This PR addresses [the error that occurs when running llama_runner multiple times within a single process.](https://github.com/pytorch/executorch/issues/15732)
Could you please take a look?

Thanks

backends/qualcomm/runtime/SharedBuffer.h
```diff
@@ -93,7 +84,10 @@ class SharedBuffer final {
   std::unordered_map<void*, size_t> allocated_size_map_;
   // Maps for the custom memory
   std::unordered_map<void*, void*> tensor_addr_to_custom_mem_;
-  std::unordered_set<CustomMemTensorInfo> custom_mem_tensor_info_set_;
+  // After the custom memory is freed, we will ensure that no tensor addresses
+  // remain linked to this custom memory.
+  std::unordered_map<void*, std::unordered_set<void*>>
```
cccclai, line 89: Can you add a comment to describe how custom memory work (like the life cycle and ownership etc) in this stack? I received the question multiple times.

cccclai:
Also our CI isn't working last week. Can you rebase to see if it's recovered?

backends/qualcomm/runtime/SharedBuffer.h
shewu-quic: Thanks for your advice. I have added a section that explains how to use the shared buffer in the QNN backend.

cccclai (approved):
Thank you!

PR #16002, opened 2025-11-27
[OpenVINO] NNCF Data-Aware Compression Algorithms Support for OVQuantizer

anzr299:
### Summary
This PR introduces the use of `nncf.compress_pt2e()` API which allows users to pass a quantizer object which is compatible with the Torch AO Quantizers API and the Torch FX model. It then returns the model with weights only compression applied to it along with additional algorithms that can be applied from NNCF like AWQ, Scale Estimation, etc.


backends/openvino/requirements.txt
```diff
@@ -1,2 +1,2 @@
 transformers
-git+https://github.com/openvinotoolkit/nncf@3d753ac#egg=nncf
+git+https://github.com/openvinotoolkit/nncf@88445b32d6a622e177945c331beef95e222dbe00
```
mergennachin: Ideally, I'd like this to point to stable pypi version. I understand this is temporary

Can you create an issue to track this?

mergennachin: Also add a newline here

backends/openvino/quantizer/llm_compression.py
```diff
@@ -0,0 +1,112 @@
...
+    the given quantizer and optional AWQ / scale estimation enabled.
+
+    :param builder_exported: LLMEdgeManager containing the FX graph, tokenizer path,
+        calibration prompt, and max sequence length.
+    :param quantizer: TorchAO quantizer to use for compression.
+    :param awq: If True, enables Activation-aware Weights Quantization (AWQ).
+    :param scale_estimation: If True, enables NNCF's scale estimation algorithm.
+    :return: The updated LLMEdgeManager with compressed torch FX model
+    """
+    tokenizer = get_tokenizer(builder_exported.tokenizer_path)
```
mergennachin: check for the existence of tokenizer path

backends/openvino/quantizer/llm_compression.py
```diff
@@ -0,0 +1,112 @@
...
+
```
mergennachin: double check calibration data is in fact present

backends/openvino/quantizer/llm_compression.py
```diff
@@ -0,0 +1,112 @@
+# Copyright (c) Intel Corporation
```
mergennachin, line 1: Can you add a basic unit test that calls this flow?

mergennachin (changes requested):

backends/openvino/quantizer/llm_compression.py
anzr299: Done

backends/openvino/quantizer/llm_compression.py
anzr299: Done

backends/openvino/quantizer/llm_compression.py
anzr299: Done

backends/openvino/requirements.txt
anzr299: > Ideally, I'd like this to point to stable pypi version. I understand this is temporary
> 
> Can you create an issue to track this?

Done:
https://github.com/pytorch/executorch/issues/17148.
We are close to releasing the new version. Perhaps we can wait on this PR until then to merge.

mergennachin (approved):

anzr299:
@mergennachin the new version of nncf is released and I have updated the [requirements.txt](https://github.com/anzr299/executorch/blob/0093592df4a3f7cedf803207c656e74b6e8ee7d6/backends/openvino/requirements.txt) accordingly

PR #16004, opened 2025-11-27
Add Python 3.13 support

### LINKED ISSUE #8292: Support Python 3.13

PyTorch 2.5 and above support Python 3.13 as as experimental. It would be desirable for us to support 3.13, as well. It would be nice to have, from a developer experience perspective, as it makes it more likely that ET "just works" for users on a wider variety of systems.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

PetarTerziev-ROLI:
### Summary

coremltools versions prior to 9.0 did not support Python 3.13. Now executorch uses 9.0, we can finally enable building wheels with Python 3.13.

Fixes #8292 

### Test plan
Python test suite ran and compared against a dev install of the target branchh in a 3.12 virtual environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,11 +47,10 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
 ]
 
-# Python dependencies required for use.
-# coremltools has issue with python 3.13, see https://github.com/apple/coremltools/issues/2487
-requires-python = ">=3.10,<3.13"
+requires-python = ">=3.10,<3.14"
 dependencies=[
   "expecttest",
   "flatbuffers",
@@ -132,4 +131,4 @@ first_party_detection = false
 # Emit syntax compatible with older versions of python instead of only the range
 # specified by `requires-python`. TODO: Remove this once we support these older
 # versions of python and can expand the `requires-python` range.
-target-version = ["py38", "py39", "py310", "py311", "py312"]
+target-version = ["py38", "py39", "py310", "py311", "py312", "py313"]
--- a/.github/workflows/build-wheels-aarch64-linux.yml
+++ b/.github/workflows/build-wheels-aarch64-linux.yml
@@ -32,7 +32,7 @@ jobs:
       test-infra-ref: main
       with-cuda: disabled
       with-rocm: disabled
-      python-versions: '["3.10", "3.11", "3.12"]'
+      python-versions: '["3.10", "3.11", "3.12", "3.13"]'
 
   build:
     needs: generate-matrix
--- a/.github/workflows/build-wheels-linux.yml
+++ b/.github/workflows/build-wheels-linux.yml
@@ -32,7 +32,7 @@ jobs:
       test-infra-ref: main
       with-cuda: disabled
       with-rocm: disabled
-      python-versions: '["3.10", "3.11", "3.12"]'
+      python-versions: '["3.10", "3.11", "3.12", "3.13"]'
 
   build:
     needs: generate-matrix
--- a/.github/workflows/build-wheels-macos.yml
+++ b/.github/workflows/build-wheels-macos.yml
@@ -32,7 +32,7 @@ jobs:
       test-infra-ref: main
       with-cuda: disabled
       with-rocm: disabled
-      python-versions: '["3.10", "3.11", "3.12"]'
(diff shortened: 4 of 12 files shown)


PetarTerziev-ROLI:
I need to add a "release notes: misc" label, but I don't seem to have the ability to.

PetarTerziev-ROLI:
the tests under `examples/models/llama3_2_vision/preprocess` hang on a call into AOTInductor when torch nightly is installed.

Need to look into this further.

PetarTerziev-ROLI:
> the tests under `examples/models/llama3_2_vision/preprocess` hang on a call into AOTInductor when torch nightly is installed.
> 
> Need to look into this further.

Things are better in with a newer nightly build of torch, so I've bumped the pinned version as part of this PR too.

pyproject.toml
```diff
@@ -47,11 +47,12 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
 ]
 
 # Python dependencies required for use.
 # coremltools has issue with python 3.13, see https://github.com/apple/coremltools/issues/2487
```
larryliu0820: Delete this comment?

larryliu0820:
@PetarTerziev-UL can you please bump the nightly version of pytorch to 1122 instead? I think the nightly job failed to build 1124 nightly.

PetarTerziev-ROLI:
> @PetarTerziev-UL can you please bump the nightly version of pytorch to 1122 instead? I think the nightly job failed to build 1124 nightly.

Done! Is there a way to check nightly job statuses? I was looking at wheels available at download.pytorch.org and drawing assumptions.

PetarTerziev-ROLI:
@larryliu0820 I've rolled back to an earlier nightly version still.

Looks like the failing CI job ends up with a versions of torch and torchaudio which are incompatible with each other down to an API change in C10 Logging.

However, after looking at the test script itself, it seems like moshi requires torch <2.8,>=2.2.0 which is in conflict with building executorch agaist a nightly build of torch. I'm looking further into this...

EDIT:

Actually, I just found out that the Linux docker container also has a git hash that pins a torch version. Will can/will update that too to match the `torch_pin.py` version.

It'd also be good to try dropping the nightly bump commit and running the CI, to see whether the hang that prompted me to update that was unique to my local system. Then the nightly bump can go into a separate PR as it is turning out to be a bit of a can of worms 😁 

EDIT 2: I've changed the version of torch the Docker container to match the nightly I bumped to originally.

PetarTerziev-ROLI:
@larryliu0820 (and anyone else interested)

I've dropped the nightly update commit. It triggers a new docker container build, which then installs the latest pip (25.3) which removes the `--no-use-pep517` option for `install`. Please see [relevant changelogs](https://pip.pypa.io/en/stable/news/#v25-3). I fear this means that next time the container needs updating, it will require a full migration to a pyproject.toml based build. Though I guess everywhere a new conda env is created will suffer the same problem.

I'll raise an issue in the repo and update this comment with it.

In the mean time, I hope that the problem which prompted me to bump the nightly version was unique to me, and this can get merged.

EDIT: I see that torch, audio, and vision (probably others too) use `setup-py` instead of `python-build-pacakge` based builds. And as the projects are quite intertwined, migrating is quite an undertaking. I guess as a workaround we could force conda to install pip 25.2.

mergennachin:
@PetarTerziev-UL 

Do you need specifically torch 11-22 nightly? 

I am trying to upgrade to 11-20 in this PR

https://github.com/pytorch/executorch/pull/16198

PetarTerziev-ROLI:
> @PetarTerziev-UL
> 
> Do you need specifically torch 11-22 nightly?
> 
> I am trying to upgrade to 11-20 in this PR
> 
> #16198

@mergennachin Nope, that's also fine. Happy to wait for that to go in first?

mergennachin (approved):

kirklandsign:
Seems to cause windows issue https://github.com/pytorch/executorch/actions/runs/20177557359/job/57929443045

PetarTerziev-ROLI:
> Seems to cause windows issue https://github.com/pytorch/executorch/actions/runs/20177557359/job/57929443045

@kirklandsign Taking a look at this.

EDIT: ah it's because there's no pytorch_tokenizers wheel available from pip so it tries to build it and it blows up the Windows max character path limit :(

It's not obvious how the pytorch/tokenizers project builds and uploads its wheels, I'll try find out to see the best way to sort it.

PR #16005, opened 2025-11-28
Arm backend: Do not partition view_copy

Christoffer-JL:
aten::view_copy is now considered no-compute by
the partitioner. This prevents the case
where a partition is left empty if two
view_copy ops cancels each other out,
leaving the partition empty.

This was the case when running
torch_audio_hdemucs_high_musdb(_plus),
which made Vela output an error.

Additionaly, some incorrectly named tests in test_sum have been renamed from view to sum


zingo (approved):

zingo:
Fails are unrelated

PR #16009, opened 2025-11-28
Arm backend: Add support for asan in executor_runner

perheld:
A lightweight AddressSanitizer runtime tailored for the ExecuTorch bare metal examples. The goal is to provide basic memory safety diagnostics while keeping the runtime self-contained.

This uses the kernel asan so it is not full asan.

Change-Id: I7ca5a6d39450b13ed461f0d1b91e1f17ab69a42e


zingo (approved):

zingo:
Fails are unrelated

PR #16016, opened 2025-12-01
Cortex_M backend: Fuse clamp + hardswish decompostion

AdrianLundell:
Adds quantization and fusion of clamp.

This is in turn used to decompose the hardswish operator in two passes, one clamping the dynamic range before quantization, and one decomposing the reminder of the operation into a maximum and mul op.

The tests in this patch exposes an issue in the runtime dim_order check as it cannot differ between channels_last/channels_first for tensors with C=1 or H=W=1. Therefore the check is removed and added as a TBD instead.

Additionally fixes per_tensor quantization for conv2d.

### DIFF
--- a/backends/cortex_m/ops/cortex_m_ops_common.h
+++ b/backends/cortex_m/ops/cortex_m_ops_common.h
@@ -69,11 +69,7 @@ inline void validate_cmsis_nn_tensor_requirements(
         "Output must have the same sizes as inputs");
   }
 
-  // Dim order consistency
-  ET_CHECK_MSG(
-      executorch::runtime::tensors_have_same_dim_order(input1, input2, output),
-      "Tensors must have same dimension order");
-
+  // TBD: Validate dim_order
   // TBD: Validate memory alignment (CMSIS-NN requirement)
 }
 
--- a/backends/cortex_m/ops/cortex_m_ops_common.h
+++ b/backends/cortex_m/ops/cortex_m_ops_common.h
@@ -69,7 +69,7 @@ inline void validate_cmsis_nn_tensor_requirements(
         "Output must have the same sizes as inputs");
   }
 
-  // TBD: Validate dim_order
+  // TBD (#16032): Validate dim_order
   // TBD: Validate memory alignment (CMSIS-NN requirement)
 }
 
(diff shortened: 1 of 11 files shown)


backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -69,11 +69,7 @@ inline void validate_cmsis_nn_tensor_requirements(
         "Output must have the same sizes as inputs");
   }
 
-  // Dim order consistency
```
psiddh, line 72: Removing this is bit risky.  I have debugged this issue further, and created a follow up task:
https://github.com/pytorch/executorch/issues/16032   (Pls update the comment with this issue no. just so that it doesn't fall through the cracks)

I will take the AI to fix the util function  + add test cases to test these edge cases and then add this check back

backends/cortex_m/passes/clamp_hardswish_pass.py
```diff
@@ -0,0 +1,37 @@
...
+
+import torch
+
+from executorch.exir.dialects.edge._ops import EdgeOpOverload
+from executorch.exir.pass_base import ExportPass, NodeMetadata, ProxyValue
+from torch.fx.node import Argument
+
+
+class ClampHardswishPass(ExportPass):
+    """
```
psiddh, line 16: +1 to the comment, this is cool

Are there test cases demonstrating the improved accuracy?  Also wondering if this paradigm could be applicable to other activation functions as well.

if so (not part of this PR, but in some future PR) consider extending this to a generic ClampActivationPass that works with configurable activation types and ranges.

backends/cortex_m/passes/decompose_hardswish_pass.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Decomposes hardswish like
+
+        hardswish(x) = x * (clamp(x, -3, 3) + 3)/6
+
+    where the add and division is implemented by modifying the quantization parameters similar
+    to hardsigmoid in the activation_fusion_pass. Note that this pass assumes
+    that the output range of the preceding op is already clamped to [-3, inf] during
+    quantization by the clamp_hardswish_pass, removing the need for the negative clamp.
+    """
```
psiddh, line 34: +1 to the comment, helps the reviewer :D

psiddh:
lgtm overall, although there is 1 major concern regd. dim order removal. But we have a clear action plan to address it properly and add it back (see https://github.com/pytorch/executorch/issues/16032 for more details)

psiddh (approved):

backends/cortex_m/ops/cortex_m_ops_common.h
AdrianLundell: Thanks!

PR #16019, opened 2025-12-01
Switch to pnpm for dependency management

mergennachin:
This PR hardens the executorch WASM test CI pipeline against the [Shai Hulud 2.0 supply chain attack](https://www.wiz.io/blog/shai-hulud-2-0-ongoing-supply-chain-attack) (and similar NPM-based threats) by migrating the test runner environment from npm to pnpm and enforcing strict script blocking.

Implements --ignore-scripts during dependency installation. This ensures that even if a compromised package is downloaded, its malicious payload (usually a script that installs Bun/malware) cannot execute.

Uses curl to install pnpm, bypassing the NPM registry entirely for the tool installation.

Explicitly pins pnpm (v10.24.0) and jest (v30.2.0) versions to prevent "drift" into potentially compromised future releases.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -862,15 +862,17 @@ jobs:
         # Install Node.js and Emscripten
         source .ci/scripts/setup-emscripten.sh
 
+        npm install -g pnpm
+
         # Test selective build
         bash scripts/build_wasm_tests.sh ${{ matrix.enable-etdump }}
 
         # Install Jest
         cd cmake-out-wasm/extension/wasm/test
-        npm install --save-dev jest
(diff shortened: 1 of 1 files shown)


.github/workflows/pull.yml
```diff
@@ -861,16 +861,18 @@ jobs:
 
         # Install Node.js and Emscripten
         source .ci/scripts/setup-emscripten.sh
+        
+        npm install -g pnpm@10.24.0 --ignore-scripts
```
larryliu0820: How about `curl -fsSL https://get.pnpm.io/install.sh | sh -`

larryliu0820 (approved):

PR #16028, opened 2025-12-01
Scan support

JacobSzwejbka:
Add support for higher order ops scan. Its inefficient today because we are manually deep copying from output to input for every carry. We could do better by shallow swapping the pointers but Ill do that in a follow up if needed. 

Test plan: Unit tests and internal verification against harder patterns


exir/emit/test/test_emit.py
```diff
@@ -812,6 +812,323 @@ def map_fn(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        module = to_edge(
+            export(f, inputs, strict=True),
+            compile_config=exir.EdgeCompileConfig(_check_ir_validity=False),
+        )
+        program = module.to_executorch().executorch_program
+
+        op_table = program.execution_plan[0].operators
+        instructions = program.execution_plan[0].chains[0].instructions
+
+        # Collect all operator names in the program
```
larryliu0820, line 841: honestly all the ops seem like implementation details and should not be tested

JacobSzwejbka: I was using it as a sort of a proxy that the general pattern was emitted. If you want we can just test the end 2 end behavior though.

larryliu0820: Don't have a strong opinion, but you might have to maintain this test if there's a change to the exported graph in the future

exir/emit/_emitter.py
```diff
@@ -944,22 +944,258 @@ def forward(self, x,y):
...
+              Working carry buffers, initialized from init, updated each iteration
+            - y_outputs (subemitter_binding_output_values[num_carry:]):
+              Pre-allocated stacked output buffers, filled via et_copy_index
+
+        The combine_fn writes to its own temporary output buffers (concrete_output_ids).
+        After each iteration:
+            1. Copy combine_fn's carry output -> carry_outputs (for next iteration)
+            2. et_copy_index(y_outputs, combine_fn's y output, iter_idx)
+
+        This explicit copy approach is used because in-place op.out(x, out=x) is unsafe.
```
larryliu0820, line 980: I was under the impression that this might be fine. We basically emit scan at the very end of the lowering process and I'm not convinced we still require the graph to be functional.

exir/emit/_emitter.py
```diff
@@ -944,22 +944,258 @@ def forward(self, x,y):
...
+        """
+        combine_fn, init, xs, additional_inputs = args
+
+        assert isinstance(
+            subemitter_binding_output_values, (list, tuple)
+        ), f"Expected list for subemitter_binding_output_values. Got {subemitter_binding_output_values}."
```
larryliu0820: This print can be overwhelming. Maybe print the type first and then the actual value?

exir/emit/_emitter.py
```diff
@@ -944,22 +944,258 @@ def forward(self, x,y):
...
+        assert isinstance(additional_inputs, (list, tuple))
+
+        num_carry = len(init)
+        num_xs = len(xs)
+
+        carry_outputs = list(subemitter_binding_output_values[:num_carry])
+        y_outputs = list(subemitter_binding_output_values[num_carry:])
+
+        if num_xs < 1:
+            raise RuntimeError("Scan requires at least one xs tensor to scan over.")
```
larryliu0820:
```suggestion
            raise RuntimeError(f"Scan requires at least one xs tensor to scan over but got {num_xs}")
```

larryliu0820 (approved):

exir/emit/_emitter.py
JacobSzwejbka: No the problem isnt being functional its that aten (and ET ops) are not guaranteed to work when in and out alias the same memory. 

You could very easily write before read over sections of the tensor.

exir/emit/test/test_emit.py
JacobSzwejbka: The ops we are querying over are the ones /not/ in the original model definition but instead created by the emitter to maintain the semantics of scan

exir/pass_base.py
```diff
@@ -545,6 +550,40 @@ def call_map(
             meta,
         )
 
+    def call_scan(
```
JacobSzwejbka, line 553: @angelayi can you check that Im not doing anything stupid here

exir/pass_base.py
```diff
@@ -545,6 +550,40 @@ def call_map(
...
+        ]
+        num_init = len(init)
+        # The x_element placeholders are at indices [num_init : num_init + num_xs]
+        xs_element_data = []
+        for i, x_proxy in enumerate(xs):
+            ph = combine_fn_placeholders[num_init + i]
+            # Use the placeholder's val which has the correct shape
+            xs_element_data.append(ph.meta["val"])
+
+        combine_fn_result = self.call_submodule(
```
JacobSzwejbka, line 574: I mostly copied torch.cond here with running call_submodul. Is this just so the subgraph also gets a chance to be run over by spec prop before callign the original? It just seems weird Im calling scan on this subgraph instead of the original one passed in as an arg

exir/pass_base.py
```diff
@@ -545,6 +550,40 @@ def call_map(
...
+        combine_fn_placeholders = [
+            n for n in combine_fn.graph.nodes if n.op == "placeholder"
+        ]
+        num_init = len(init)
+        # The x_element placeholders are at indices [num_init : num_init + num_xs]
+        xs_element_data = []
+        for i in range(0, len(xs)):
+            ph = combine_fn_placeholders[num_init + i]
+            # Use the placeholder's val which has the correct shape
+            xs_element_data.append(ph.meta["val"])
```
angelayi, line 572: i think this part is a little sus where you look at the subgraph's placeholder nodes. I think the xs_element_data should just be something like, xs[0]?

PR #16030, opened 2025-12-01
make triton kernel usage user controlable

Gasoonjia:
Summary: We observed perf regression on some models after using triton sdpa kernel, compare with decomposed operator. This diff makes triton kernel controalable by user to prevent such perf regression.

Reviewed By: larryliu0820

Differential Revision: D88096054


larryliu0820 (approved):

PR #16052, opened 2025-12-02
Remove fairseq from export_llama

lucylq:
Summary: This is not being used anymore; remove it and simplify export_llama

Differential Revision: D87831086


larryliu0820 (approved):
Thank you for cleaning this up!

PR #16054, opened 2025-12-02
Modify depthwise int8 conv2d to reduce register/memory pressure

GregoryComer:
## Summary

Modify the Vulkan int8 depthwise convolution shader to reduce register pressure by not loading the full input window upfront. On Mali, it seems to spill to main memory with significant performance impact. This is a relatively naive implementation that largely just loads as needed. It shows significant speedup on Mali-G720. Adreno performance regresses slightly (~10%) on some benchmarks, but actually shows a consistent end-to-end speedup of ~9% on a skin segmentation model on S25 (Adreno 830).

There is likely a lot of room for additional optimization here. In particular, optimizing cycle count by reducing bounds checks, looking at improvements to read coalescing, and using spec constants for the major params.

Differential Revision: D88183224

### DIFF
--- a/backends/vulkan/runtime/graph/ops/glsl/conv2d_dw_q8_utils.glslh
+++ b/backends/vulkan/runtime/graph/ops/glsl/conv2d_dw_q8_utils.glslh
@@ -11,20 +11,6 @@
 
 #extension GL_EXT_control_flow_attributes : require
 
-struct InputWindow1D {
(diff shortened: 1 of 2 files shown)


SS-JIA (changes requested):

SS-JIA (approved):
Thanks for the great work!

PR #16056, opened 2025-12-02
[CUDA]: GPU Device Caching for Encoder Output in CUDA Backend

mergennachin:
Summary:

In encoder-decoder models like Whisper, the encoder output tensor is
used as input to every decoder iteration, and doing unnecessary
CPU->GPU->CPU->GPU cpies.

Implemented a "keep on device" caching mechanism in the CUDA backend
that:

-  Caches encoder output in persistent GPU memory after the encoder runs
-  Uses fast GPU-to-GPU copies decoder iterations instead of slow CPU-to-GPU copies

Test Plan:

make whisper-cuda

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/backends/cuda/runtime/cuda_backend.cpp
+++ b/backends/cuda/runtime/cuda_backend.cpp
@@ -8,6 +8,7 @@
 
 #include <cuda_runtime.h>
 #include <executorch/runtime/backend/interface.h>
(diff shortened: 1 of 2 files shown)


mergennachin:
Newer version here: https://github.com/pytorch/executorch/pull/16060/

PR #16057, opened 2025-12-03
ANE llama runner fixes for iOS26

metascroy:
This PR:

* Resolves https://github.com/pytorch/executorch/issues/15833.  The main issue is that iOS26 has a bug in its ANE implementation of SDPA, so we break SDPA into matmuls and softmax to avoid this.
* iOS26 also has large performance regressions over iOS18, so we tweak the model slightly (by exporting in parts) to achieve iOS18-level performance.
* We also add a C++ runner for the model split in parts.

With these changes, we estimate performance of Llama1B on iPhone 15 Pro / iOS 26 at:
* 30 tok/sec decode
* 1900 tok/sec prefill

Differential Revision: D88083155

### DIFF
--- a/examples/apple/coreml/llama/export.py
+++ b/examples/apple/coreml/llama/export.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import argparse
+import os
 
 import coremltools as ct
 import torch
@@ -14,6 +15,7 @@ from executorch.backends.apple.coreml.partition import CoreMLPartitioner  # pyre
 from executorch.examples.apple.coreml.llama.llama_transformer import (
     InputManager,
     load_model,
+    load_model_in_pieces_ITO,
 )
 from executorch.examples.apple.coreml.llama.utils import (
     replace_linear_with_split_linear,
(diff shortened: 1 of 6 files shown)


examples/apple/coreml/llama/export.py
```diff
@@ -102,67 +106,69 @@ def main() -> None:
...
-        replace_linear_with_split_linear(
-            model,
-            out_target_split_size=export_args.target_split_size,
-            out_max_splits=export_args.max_splits,
-            # I have not found splitting on in_features to be beneficial,
-            # and it often leads to OOM so I set in_max_splits to 1
-            in_target_split_size=1,
-            in_max_splits=1,
-        )
+    def maybe_split_model(model):
```
cccclai, line 121: Is maybe_split_model mostly just split linear or something else

metascroy: It splits the linear only.  But I have changes planned here.

metascroy:
We are unifying iOS26 fixes and future development around the static model https://github.com/pytorch/executorch/pull/16184

PR #16060, opened 2025-12-03
[CUDA]: GPU Device Caching for Encoder Output in CUDA Backend

mergennachin:
**Summary**

This PR implements a "keep on device" optimization for the CUDA backend that stores encoder output tensors on the GPU and reuses them via fast device-to-device (D2D) copies
during decoder iterations. This avoids redundant CPU→GPU transfers in encoder-decoder models like Whisper.

**Motivation**

In encoder-decoder architectures like Whisper, the encoder runs once and produces an output tensor that the decoder consumes on every iteration. Without this optimization,
the flow is:

Encoder: CPU input → [H2D copy] → GPU compute → [D2H copy] → CPU output
Decoder (×N tokens): CPU inputs (including encoder output) → [H2D copy] → GPU compute → [D2H copy] → CPU output

The encoder output (~2.3 MB for Whisper) is copied from CPU→GPU on every decoder iteration, even though it never changes. With N=109 tokens, that's 109 redundant H2D copies.

**Design**

The optimization introduces a simple GPU tensor storage mechanism:

Encoder: CPU input → [H2D copy] → GPU compute → [store on GPU] → [D2H copy] → CPU output
Decoder (×N tokens): CPU inputs → [D2D copy for encoder output, H2D for others] → GPU compute → [D2H copy] → CPU output

**Key Design Decisions**

1. Name-based storage: Tensors are stored by name (e.g., "encoder_output") rather than by slot index, making the API more intuitive and less fragile.
2. Size-based matching: When looking for a stored tensor to use as input, the backend matches by tensor size rather than requiring explicit slot mapping.
3. Explicit opt-in: The optimization is controlled via set_option() calls, making it backwards compatible and non-intrusive to existing code paths.
4. RAII cleanup: A TensorCleanup guard ensures GPU tensors are freed on all exit paths, preventing memory leaks when errors occur during execution.

**API**

Three backend options control the behavior:

| Option             | Type   | Description                                        |
|--------------------|--------|----------------------------------------------------|
| store_output       | string | Store first output tensor under this name          |
| use_stored_input   | string | Use stored tensor for inputs matching by size      |
| reset_stored_input | bool   | Clear the input setting (tensor remains in memory) |

**Lifecycle**

1. Before encoder: set_option("store_output", "encoder_output")
2. Run encoder: output tensor stored on GPU
3. Before decoder loop: set_option("use_stored_input", "encoder_output")
4. Run decoder ×N: each iteration uses D2D copy for encoder output
5. After decoder loop: set_option("reset_stored_input", true)
6. On destroy(): all stored tensors freed

**Changes**

backends/cuda/runtime/cuda_backend.cpp (+212 lines)
- Added GpuTensorRef struct to hold GPU tensor references with ownership
- Added gpu_tensors_ map for named tensor storage with documented lifetime contract
- Implemented set_option() with validation for option types
- Added RAII TensorCleanup guard to prevent memory leaks on error paths
- Added validation for single-output constraint
- Cleanup of stored tensors in destroy()

extension/asr/runner/runner.cpp (+42 lines)
- Set store_output before encoder execution
- Set use_stored_input before decoder loop
- Reset after decoder loop completes
- Consistent error logging at Warning level

**Performance**

Profiling with nsys confirms the optimization works:

| Operation  | Count | Total Size | Avg Time |
|------------|-------|------------|----------|
| H2D copies | 722   | 521.7 MB   | 125 µs   |
| D2D copies | 109   | 251.1 MB   | 13.6 µs  |

The 109 D2D copies (one per decoder token) each transfer 2.304 MB (encoder output) ~9x faster than equivalent H2D copies would.

**Test plan**

- Build succeeds
- Whisper transcription produces correct output
- nsys profile confirms D2D copies are occurring
- No memory leaks (RAII cleanup on all paths)

### DIFF
--- a/backends/cuda/runtime/cuda_backend.cpp
+++ b/backends/cuda/runtime/cuda_backend.cpp
@@ -8,6 +8,7 @@
 
 #include <cuda_runtime.h>
 #include <executorch/runtime/backend/interface.h>
+#include <executorch/runtime/backend/options.h>
 #include <executorch/runtime/core/error.h>
 #include <executorch/runtime/core/evalue.h>
 #include <executorch/runtime/core/exec_aten/util/tensor_util.h>
@@ -16,6 +17,7 @@
 #include <filesystem>
 #include <fstream>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 // Include our shim layer headers
@@ -46,9 +48,27 @@ using executorch::runtime::Result;
 using executorch::runtime::Span;
 using executorch::runtime::etensor::Tensor;
 
+// Structure to hold cached GPU tensor data for "keep on device" optimization
+struct CachedGpuData {
+  void* data_ptr;           // GPU memory pointer
+  size_t size_bytes;        // Total size in bytes
+  int32_t scalar_type;      // Data type
+  std::vector<int64_t> sizes;  // Original shape
+};
+
+// Global device cache - maps name to cached GPU data
+// Using raw GPU pointers instead of tensor handles for format independence
+static std::unordered_map<std::string, CachedGpuData> g_device_cache;
+
 class ET_EXPERIMENTAL CudaBackend final
     : public ::executorch::runtime::BackendInterface {
  private:
+  // Cache control options (set via set_option before execute)
+  mutable int cache_output_slot_ = -1;        // Which output slot to cache (-1 = none)
+  mutable std::string cache_output_name_;     // Name to cache output under
+  mutable int use_cache_input_slot_ = -1;     // Which input slot to use cache for (-1 = none)
+  mutable std::string use_cache_input_name_;  // Name of cached tensor to use
+
   Error load_function_pointers_into_handle(
       void* so_handle,
       AOTIDelegateHandle* handle) const {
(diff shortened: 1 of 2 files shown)


backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -46,9 +48,88 @@ using executorch::runtime::Result;
...
+
 class ET_EXPERIMENTAL CudaBackend final
     : public ::executorch::runtime::BackendInterface {
  private:
+  // ============================================================================
+  // GPU Tensor Storage for D2D Copy Optimization
+  // ============================================================================
+  //
+  // This backend supports storing GPU tensors between execute() calls to enable
+  // device-to-device (D2D) copies instead of slower host-to-device (H2D)
```
larryliu0820, line 67: I'm curious why do we still need to copy? Can you just make_tensor using the GPU data pointer

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -46,9 +48,88 @@ using executorch::runtime::Result;
...
+  //
+  //   1. Before encoder: set_option("store_output", "encoder_output")
+  //   2. Execute encoder (output is stored on GPU)
+  //   3. Before decoder loop: set_option("use_stored_input", "encoder_output")
+  //   4. Execute decoder N times (D2D copies for encoder output input)
+  //   5. After decoder loop:
+  //        set_option("reset_stored_input", true)
+  //        set_option("clear_stored_tensor", "encoder_output")
+  //
+  // ============================================================================
```
cccclai, line 102: Trying to understand the intention, is it trying to use backend option to have let the method encode/decode share the output memory? In an ideal word, if encode/decode methods can share memory planning, does it mean we don't have to use this?

backends/cuda/runtime/cuda_backend.cpp
mergennachin: Yeah, I tried not copying initially and it was segfaulting. Because they're completely two different graphs, the output from the first graph and input from the second graph had different underlying layout assumptions, so had to explicitly copy.

backends/cuda/runtime/cuda_backend.cpp
JacobSzwejbka: >is it trying to use backend option to have let the method encode/decode share the output memory?

Its trying to avoid cpu->gpu copies. If we had device tensor we wouldnt need this, but its wip and perf here is time sensitive so Mergen is hacking around it until its properly fixed upstream

cccclai: Ah I see, that seems fine to me. Maybe worth adding this as part of the comment because I can't tell from the PR

mergennachin:
Closing in lieu of https://github.com/pytorch/executorch/pull/16235

PR #16061, opened 2025-12-03
Cortex_M backend: Remove int8 requirement for zero_points

AdrianLundell:
The zero point would sometimes get out of range in the hardswish decomposition (depending on the range of the quantized output) as it adjusts the zero point. Since the zero_point is applied in int32 accumulation there is no actual reason for the limitation, so we can just remove it.

Additionally change the input to linear_hardswish to reliably trigger the behaviour causing the crash before the fix was introduced to catch future regressions.


oscarandersson8218 (approved):
LGTM!

AdrianLundell:
Test failures unrelated

PR #16071, opened 2025-12-04
Arm backend: Make sure DW-conv weights are reshaped once

oscarandersson8218:
Make sure RewriteConv2DPass only reshapes shared weights once.

### Test plan
Functionality is tested in backends/arm/test/misc/test_dw_convs_with_shared_weights.py.


zingo:
Fails are unrelated

zingo (approved):

PR #16072, opened 2025-12-04
Arm backend: Make SupportedTOSAOperatorChecks work for INT+FP

martinlsm:


oscarandersson8218 (approved):
cortex-m failures are unrelated.

PR #16073, opened 2025-12-04
Arm backend: Improve quantizer configuration in tests

Erik-Lundell:
Updating the quantizer in test pipelines has been rather cumbersome. As we anticipate more tests with different quantization, we want to make it easy.

The quantizer can now be accessed, and modified using pipeline.quantizer.set_[...]

See the patch for examples.


zingo (approved):

zingo:
fails unrelated

PR #16074, opened 2025-12-04
Append to cflag and honor CMAKE_POSITION_INDEPENDENT_CODE OFF

zingo:
### Summary

With this change we append the _common_compile_options instead of replacing it and only set -fPIC if CMAKE_POSITION_INDEPENDENT_CODE is not set to OFF

This fixes a problem when building from other projects like ZephyrOS as it already replaced setup _common_compile_options instead of just appending the flags. It also and always set -fPIC independent of how CMAKE_POSITION_INDEPENDENT_CODE was set. This was causing incompatible build binaries when building the objects files.

### Test plan

Tested by invoking the project from ZephyrOS an look at generated build instructions after cmake has been executed.


zingo:
Hi @metascroy this is outside of Arm and more of a global change and might need your check/verdict or just need a review from that point of view :)

zingo:
Rebased to keep in synch with latest code.

zingo:
Hi @digantdesai do you think this would be problematic or OK, as it blocks ZephyrOS PR it would be good to know if this is not along the lines of something that can be merged and we need to add some cmake remove PIC logic in various plaices  in the ZephyrOS PR instead.

digantdesai (approved):
Should be OK to update this default globally. Assuming no size regression. Perf we won't know. Thanks.

PR #16076, opened 2025-12-04
Arm backend: Move FuseConstantArgsPass

oscarandersson8218:
Move FuseConstantArgsPass to run after ComputeConstantOpsAOTPass to make sure that ops without input nodes are pre-computed first.


zingo (approved):
Nice find!

PR #16085, opened 2025-12-04
disable sdpa in static attention mha forward

billmguo:
Differential Revision: D87944616


metascroy:
@billmguo are the changes in llama_transformer.py identical to the ones here https://github.com/pytorch/executorch/pull/16057 (except this PR doesn't implement the 3-part split)?

PR #16086, opened 2025-12-04
Metal backend: SDPA metal implementation

manuelcandales:
Replaces SDPA MPSGraph's implementation with Metal implementation (adapted from MLX implementation, with several modifications, to support transposed middle dimensions, and floating point attention masks).

Speeds up voxtral/whisper by 2-3x

Fixes BFloat16 issue on macOS 26.1

### DIFF
--- a/backends/apple/metal/runtime/shims/et_metal_ops.mm
+++ b/backends/apple/metal/runtime/shims/et_metal_ops.mm
@@ -19,6 +19,7 @@
 #include <executorch/backends/apple/metal/runtime/shims/memory.h>
 #include <functional>
 #include <unordered_map>
+#include <memory>
 
 namespace executorch {
 namespace backends {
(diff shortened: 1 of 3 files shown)


manuelcandales:
* __->__ #16086

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -953,560 +1181,347 @@ AOTITorchError aoti_torch_mps__scaled_dot_product_attention_math_for_mps(
...
+        ET_LOG(Error, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: Unsupported dtype for Metal kernel");
+        throw std::runtime_error("Unsupported dtype for Metal SDPA kernel");
+      }
+
+      // Select head_dim - must match exactly one of the supported sizes (64, 96, 128)
+      int64_t head_dim = headSize;
+      if (head_dim != 64 && head_dim != 96 && head_dim != 128) {
+        ET_LOG(Error, "aoti_torch_mps__scaled_dot_product_attention_math_for_mps: Unsupported head_dim %lld (must be 64, 96, or 128)", head_dim);
+        throw std::runtime_error("Unsupported head_dim for Metal SDPA kernel - must be exactly 64, 96, or 128");
+      }
```
mergennachin, line 1336: What's the main reason for only limiting to these head sizes?

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -122,6 +123,233 @@ void logStats() {
...
+
+      // Add mask value to score if mask is present
+      if (has_mask) {
+        score += static_cast<U>(mask[0]);
+      }
+
+      // Update the accumulators
+      U new_max = max(max_score, score);
+      U factor = metal::fast::exp(max_score - new_max);
+      U exp_score = metal::fast::exp(score - new_max);
```
mergennachin, line 256: Nitpick/micro optimizatoin: Can you save on exponentation if max_score == score?

mergennachin (commented):
- Are you ignoring "is_causal" altogether? https://github.com/pytorch/executorch/blob/a09a4b74feb4680c840f37c458fc291e36ce78bd/backends/apple/metal/runtime/shims/et_metal_ops.h#L61

- Can we compile the metal shader at build time? Isn't it jit compiling?

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: The head size is a compile-time constant. See the template instantiation lines:
```
#define INSTANTIATE_SDPA_VECTOR_HEADS(DTYPE)        \
  INSTANTIATE_SDPA_VECTOR(DTYPE, 64, 64);           \
  INSTANTIATE_SDPA_VECTOR(DTYPE, 96, 96);           \
  INSTANTIATE_SDPA_VECTOR(DTYPE, 128, 128);
```
This is also how MLX does it, and I am pretty sure it is done for performance reasons (loop unrolling, register optimization, etc).

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: We only do one actual exponentiation. One of those two exponentiations is trivial (zero exponent). Here is the full snippet:
```
      U new_max = max(max_score, score);
      U factor = metal::fast::exp(max_score - new_max);
      U exp_score = metal::fast::exp(score - new_max);
```
`new_max` is either the existing `max_score`, or the new `score`.

manuelcandales:
> * Are you ignoring "is_causal" altogether? 

Yes, for now. I added a check, returning `Error:NotImplemented` when `is_causal` is `True`

manuelcandales:
> * Can we compile the metal shader at build time? Isn't it jit compiling?

We don't have yet that infrastructure in the Metal backend. I don't think that's a priority though, because AOTI generated kernels are also jit compiled, and the amount of code we are jit compiling there is much more than what we are compiling in the single SDPA kernel. So, I don't think it would make any difference right now.

mergennachin (approved):

PR #16131, opened 2025-12-08
Cortex_m backend: Support channels-broadcasting for ADD/MUL

AdrianLundell:
Adds support for broadcasting in the special case where one input contains only channels which are broadcasted onto every spatial element of the other tensor, e.g. [1,C,1,1] + [N, C, H, W] for channel-last tensors.

This is a needed for mobilenet_v3.

### DIFF
--- a/backends/cortex_m/ops/cortex_m_ops_common.h
+++ b/backends/cortex_m/ops/cortex_m_ops_common.h
@@ -129,6 +129,43 @@ inline void validate_quantization_params(
       "Single quant Output");
 }
 
+inline bool is_channels_last_tensor(const Tensor& tensor) {
+  if (tensor.dim() != 4) {
+    return false;
+  }
+
+  // When channels or spatial dims are 1 the layout information is ambiguous.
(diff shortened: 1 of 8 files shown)


psiddh:
This PR does not implement general broadcasting semantics, but rather adds support for a very specific broadcasting case required for MobileNet v3 (MV3).
In general, should we aim to support full broadcasting semantics?

AdrianLundell:
My two cents is that full broadcasting support takes engineering effort, code size, and likely performance tradeoffs, and on the other hand the selection of networks which makes sense to run on MCUs is fairly limited and might not use a lot of broadcasting.

So in my view only implementing what is used makes sense here, until we see a large need for full broadcasting support (if ever).

psiddh:
> My two cents is that full broadcasting support takes engineering effort, code size, and likely performance tradeoffs, and on the other hand the selection of networks which makes sense to run on MCUs is fairly limited and might not use a lot of broadcasting.
> 
> So in my view only implementing what is used makes sense here, until we see a large need for full broadcasting support (if ever).

Seems reasonable, if more usecases / models start requiring general broadcasting, we could look at supporting it more broadly

psiddh (approved):

PR #16132, opened 2025-12-08
Implement guardrail to check for existing branches

shoumikhin:
Added a guardrail step to prevent overwriting existing branches in the Apple workflow to avoid accidental binaries overwrite on S3.


GregoryComer (approved):
Looks good to me. Copilot nit about the name seems reasonable but it's your call.

PR #16138, opened 2025-12-09
Prevent signed integer overflow in pixel_shuffle size calculation

GregoryComer:
Summary:
The size calculation in `get_pixel_shuffle_out_target_size` can trigger signed integer overflow (UB) with very large upscale_factors. This manifested as an integer division by zero fault during fuzzing.

Specifically, the calculation of `upscale_factor * upscale_factor` ([source](https://github.com/pytorch/executorch/blob/04f1e4d22383ffcbc770acf5002348e3f95082a2/kernels/portable/cpu/util/copy_ops_util.cpp#L364)) can overflow. In the motivating case, SizesType is a signed 32-bit integer and upscale_factor = 2^17.

Since this is an impractically large upscale factor, I'm just adding a constraint that upscale_factor < 32768 (2^15). In theory, SizesType could be defined as less than 32 bits, but this seems unlikely in practice. I check this upper bound in `check_pixel_shuffle_args` and also assert in `get_pixel_shuffle_out_target_size` to ensure we don't hit UB.

Differential Revision: D88693324

### DIFF
--- a/kernels/portable/cpu/util/copy_ops_util.cpp
+++ b/kernels/portable/cpu/util/copy_ops_util.cpp
@@ -346,11 +346,15 @@ bool check_pixel_unshuffle_args(
   return true;
 }
 
-void get_pixel_shuffle_out_target_size(
+bool get_pixel_shuffle_out_target_size(
     const Tensor& in,
     int64_t upscale_factor,
     executorch::aten::SizesType* out_sizes,
     size_t* out_ndim) {
+  // Prevent signed integer overflow when computing upscale_factor ^ 2.
+  ET_CHECK_OR_RETURN_FALSE(
+      upscale_factor < 32768, "Upscale factor must be less than 32768.");
+
   *out_ndim = in.dim();
   const executorch::aten::SizesType casted_upscale_factor = upscale_factor;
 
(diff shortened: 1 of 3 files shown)


JacobSzwejbka (approved):

kernels/portable/cpu/util/copy_ops_util.cpp
```diff
@@ -351,6 +353,10 @@ void get_pixel_shuffle_out_target_size(
     int64_t upscale_factor,
     executorch::aten::SizesType* out_sizes,
     size_t* out_ndim) {
+  // Prevent signed integer overflow when computing upscale_factor ^ 2.
+  // This constraint is checked in check_pixel_shuffle_args.
+  ET_CHECK(upscale_factor < 32768);
```
JacobSzwejbka: Nit: Can you update this to not fatal error and instead just set the runtime context state

GregoryComer: Yeah, I did add a non-fatal check in the pixel shuffle args check, which will in practice handle it gracefully. I added this assert here just as a sanity check, but am open to changing it if you prefer. It seemed harder to bail out at this point but I wanted to catch any UB.

GregoryComer: Actually it does seems simpler to just refactor the get out size calculation to be able to fail non-fatally and move the check there. I'll push another revision.

manuelcandales (approved):

PR #16152, opened 2025-12-09
Cortex_m backend: Add softmax support for int8 via CMSIS NN (dim=-1)

xingguo01:
- integrate CMSIS softmax into Cortex-M backend
- add fusion pass/tests for quantized softmax

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai @Sebastian-Larsson @AdrianLundell @psiddh

### DIFF
--- a/backends/cortex_m/ops/op_softmax.cpp
+++ b/backends/cortex_m/ops/op_softmax.cpp
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2025 Arm Limited and/or its affiliates.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "cortex_m_ops_common.h"
+
+#include <cmath>
+#include <cstdint>
+#include <limits>
+
+// Include CMSIS-NN headers with C linkage
+extern "C" {
+#include "arm_nnfunctions.h"
+}
+
+namespace cortex_m {
+namespace native {
+
+namespace {
+
+constexpr int32_t kCmsisSoftmaxZeroPoint = -128;
+
+inline bool is_int8_tensor(const Tensor& tensor) {
+  return tensor.scalar_type() == ScalarType::Char;
+}
+
+inline bool is_last_dim(const Tensor& tensor, int64_t dim) {
+  const auto rank = tensor.dim();
+  const int64_t positive_dim = dim >= 0 ? dim : dim + rank;
+  return positive_dim == static_cast<int64_t>(rank - 1);
+}
+
+inline int64_t normalize_dim(const Tensor& tensor, int64_t dim) {
+  const auto rank = tensor.dim();
+  const int64_t positive_dim = dim >= 0 ? dim : dim + rank;
+  return positive_dim;
+}
+
+} // namespace
+
+Tensor& softmax_out(
+    KernelRuntimeContext& context,
+    const Tensor& input,
+    int64_t dim,
+    int64_t input_zero_point,
+    int64_t output_zero_point,
+    int64_t input_multiplier,
+    int64_t input_shift,
+    int64_t diff_min,
+    Tensor& out) {
+  if (!is_int8_tensor(input) || !is_int8_tensor(out)) {
+    ET_LOG(
+        Error,
+        "softmax_out: only int8 tensors are supported (input=%d, out=%d)",
+        static_cast<int>(input.scalar_type()),
+        static_cast<int>(out.scalar_type()));
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  if (!is_last_dim(input, dim)) {
+    ET_LOG(
+        Error,
+        "softmax_out: only last-dimension softmax is supported (dim=%lld, rank=%zu)",
+        static_cast<long long>(dim),
+        static_cast<size_t>(input.dim()));
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  const int32_t output_zp_val = static_cast<int32_t>(output_zero_point);
+  const int32_t input_multiplier_val = static_cast<int32_t>(input_multiplier);
+  const int32_t input_shift_val = static_cast<int32_t>(input_shift);
+  const int32_t diff_min_val = static_cast<int32_t>(diff_min);
+
+  validate_single_quant_params(
+      Scalar(static_cast<int32_t>(input_zero_point)),
+      Scalar(input_multiplier_val),
+      Scalar(input_shift_val),
+      "softmax input");
+
+  const auto positive_dim = normalize_dim(input, dim);
+  const int64_t row_size64 = input.size(positive_dim);
+  if (row_size64 <= 0 || row_size64 > std::numeric_limits<int32_t>::max()) {
+    ET_LOG(
+        Error,
+        "softmax_out: row size must fit in int32 (row_size=%lld)",
+        static_cast<long long>(row_size64));
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  const int32_t row_size = static_cast<int32_t>(row_size64);
+  const int64_t num_rows64 = input.numel() / row_size64;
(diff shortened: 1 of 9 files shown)


xingguo01:
@psiddh much appreciated if you can have a look

AdrianLundell:
I've reviewed this internally so I am happy with it, I'll leave it to the other reviewers to give their comments before approving!

backends/cortex_m/ops/op_softmax.cpp
```diff
@@ -0,0 +1,148 @@
...
+    ET_LOG(
+        Error,
+        "softmax_out: expected output zero_point=%d (got zero_point=%d)",
+        kCmsisSoftmaxZeroPoint,
+        output_zp_val);
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  arm_softmax_s8(
```
psiddh, line 132: Checking for status ?  something like

arm_status status = arm_softmax_s8(
    input_data,
    num_rows,
    row_size,
    input_multiplier_val,
    input_shift_val,
    diff_min_val,
    output_data);
if (status != ARM_MATH_SUCCESS) {
  ET_LOG(Error, "softmax_out: CMSIS NN softmax failed (status=%d)", status);
  context.fail(Error::InvalidArgument);
  return out;
}

backends/cortex_m/ops/op_softmax.cpp
```diff
@@ -0,0 +1,148 @@
...
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  const int32_t input_zp_val = static_cast<int32_t>(input_zero_point);
+  const int32_t output_zp_val = static_cast<int32_t>(output_zero_point);
+  (void)input_zp_val; // Zero-point difference cancels out during subtraction.
+
+  validate_single_quant_params(
+      Scalar(input_zp_val),
```
psiddh: Should this be 'output_zp_val' instead ?

backends/cortex_m/test/ops/test_softmax.py
```diff
@@ -0,0 +1,80 @@
...
+    ops_after_transforms = {
+        "executorch_exir_dialects_edge__ops_cortex_m_softmax_default": 1,
+    }
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return torch.softmax(x, dim=self.dim)
+
+
+test_cases = {
+    "rank1": McuTestCase(
```
psiddh, line 34: It'd be nice to add boundary test cases too here , something like

"empty_tensor": McuTestCase(
    CortexMSoftmax(dim=-1),
    (torch.empty((0, 4), dtype=torch.float32),),
),
"large_tensor": McuTestCase(
    CortexMSoftmax(dim=-1),
    (ramp_tensor(-10, 10, (8, 1024)),),  # Test CMSIS limits
),

backends/cortex_m/quantizer/quantizer.py
```diff
@@ -86,6 +88,76 @@ def nchw_filter(self, node: Optional[Node]) -> bool:
...
+            dim_arg = node.args[1]
+        elif "dim" in node.kwargs:
+            dim_arg = node.kwargs["dim"]
+
+        if dim_arg is None:
+            return -1
+
+        return self._resolve_int(dim_arg)
+
+    def softmax_memory_format_filter(self, node: Optional[Node]) -> bool:
```
psiddh, line 119: It'd be good to add test cases testing these various conditions (in future PR, if not in this one)

psiddh (commented):
lgtm, left few comments

backends/cortex_m/ops/op_softmax.cpp
rascani: Actually it looks like `validate_single_quant_params()` does not even check the zp value, so we could probably just remove it.

backends/cortex_m/ops/op_softmax.cpp
```diff
@@ -0,0 +1,148 @@
...
+        "softmax_out: only last-dimension softmax is supported (dim=%lld, rank=%zu)",
+        static_cast<long long>(dim),
+        static_cast<size_t>(input.dim()));
+    context.fail(Error::InvalidArgument);
+    return out;
+  }
+
+  const int32_t input_zp_val = static_cast<int32_t>(input_zero_point);
+  const int32_t output_zp_val = static_cast<int32_t>(output_zero_point);
+  (void)input_zp_val; // Zero-point difference cancels out during subtraction.
```
rascani: Since we only need the input multiplier and shift for `arm_softmax_s8`, is there any value in including the zero points in the signature? We could probably do the output zero point value == -128 check during AOT.

xingguo01: @rascani thanks for your comments. First point taken. For the second point, we still get both zero points from the AOT config because the Cortex-M codegen path treats all quantized ops uniformly—the generated kernels hand the same (zp, multiplier, shift) triple into each op. Even though CMSIS softmax only consumes the multiplier/shift, keeping the zero points in the signature lets us share the invocation plumbing with the rest of the operators without special-casing softmax in the serializer or wrappers. I’d also prefer to keep the runtime guard on the output zp: it’s a cheap check that catches misconfigured or hand-constructed calls (e.g. unit tests, future refactors) instead of relying solely on the offline config pass. So I’d lean toward leaving the signature as-is and just validating that the incoming zero point matches the CMSIS requirement.

backends/cortex_m/ops/op_softmax.cpp
xingguo01: As @rascani points out, it doesn't do anything. But I didn't completely remove it though.

backends/cortex_m/test/ops/test_softmax.py
xingguo01: empty_tensor blew up earlier in the pipeline: the dialect test hit TensorStatistics.from_tensor with a (0, 4) tensor and torch.quantile(flattened, 0.5) throws on empties (src/executorch/backends/test/harness/error_statistics.py (line 34)). Until we harden that helper (or skip stats when numel == 0), that case can’t live in the suite. The large_tensor case ran cleanly. Suggest either dropping empty_tensor or marking it TODO while we teach the harness to tolerate empty inputs.

backends/cortex_m/ops/op_softmax.cpp
xingguo01: CMSIS’ arm_softmax_s8 still has a void signature, so there’s no status code to capture.

backends/cortex_m/quantizer/quantizer.py
xingguo01: I’ll add targeted unit tests for the memory format gating logic in a future follow-up PR (either new cases in the existing quantizer suite or a small dedicated test module) so we can exercise each condition explicitly.

xingguo01:
Thanks @rascani and @psiddh for your feedback and comments. I will try to upload a new version shortly after some local/CI tests.

backends/cortex_m/quantizer/quantizer.py
psiddh: > I’ll add targeted unit tests for the memory format gating logic in a future follow-up PR (either new cases in the existing quantizer suite or a small dedicated test module) so we can exercise each condition explicitly.

Thanks

psiddh (approved):
lgtm

backends/cortex_m/ops/op_softmax.cpp
xingguo01: Just to be clear here: https://github.com/ARM-software/CMSIS-NN/blob/main/Include/arm_nnfunctions.h#L2307. Thanks for approving the PR, my first one to Excutorch

AdrianLundell:
Failures unrelated

backends/cortex_m/ops/op_softmax.cpp
psiddh: This is great ! Thanks for your contribution ! 
Hopefully many more contributions to come :)

PR #16174, opened 2025-12-10
Summary: Add context when LLMModule load fails to aid in debugging

psiddh:
Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### Summary

### Test plan

### DIFF
--- a/extension/android/jni/jni_layer_llama.cpp
+++ b/extension/android/jni/jni_layer_llama.cpp
@@ -23,6 +23,8 @@
 #include <executorch/runtime/platform/platform.h>
 #include <executorch/runtime/platform/runtime.h>
 
+#include <executorch/extension/android/jni/jni_helper.h>
+
 #if defined(ET_USE_THREADPOOL)
 #include <executorch/extension/threadpool/cpuinfo_utils.h>
 #include <executorch/extension/threadpool/threadpool.h>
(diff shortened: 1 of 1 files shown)


extension/android/jni/jni_layer_llama.cpp
```diff
@@ -404,12 +406,29 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
   }
 
   jint load() {
+    int result = -1;
+    std::stringstream ss;
+
     if (model_type_category_ == MODEL_TYPE_CATEGORY_MULTIMODAL) {
-      return static_cast<jint>(multi_modal_runner_->load());
+      result = static_cast<jint>(multi_modal_runner_->load());
```
abhinaykukkadapu, line 413: Thanks Sid for the changes, noob q, can we map and convert the int code to something readable here?

psiddh: Yes that is the idea, by integrating with the API: throwExecutorchException(..) (This is the support that was added few months back in the standard ET Android Modules), this carries rich context on the failure and propagates it all the way to the App in the form of ExecutotchRuntimeException .   We also log this error to logcat (just in case), we can use the same error message from exception to show alert dialogs or surface it to the user however the Apps wish.

kirklandsign (approved):

PR #16181, opened 2025-12-10
Arm backend: Match arg ranks for min/max ops

YufengShi-dudu:
- Int32 clamp is decomposed into min/max, but input_tensor and min_arg/max_arg can end up with different ranks (e.g. (1,) vs ()) after subsequent passes. This violates the TOSA requirements for TOSA.MINIMUM and TOSA.MAXIMUM in the node visitors.
 - Fix this by matching the input ranks in MatchArgRanksPass
 
Change-Id: I47a3bb73832ddc35553cd3e5b65085155eb9ca48


ArmRyan:
int32 clamp already does a full op to handle the rank difference. If this is not working properly you should add a test that triggers this error or if it is a replacement, you should remove the full op

YufengShi-dudu:
> int32 clamp already does a full op to handle the rank difference. If this is not working properly you should add a test that triggers this error or if it is a replacement, you should remove the full op

Hey Ryan, thanks for your comment. I realize `DecomposeInt32ClampPass` does match the input ranks of aten.clamp.default. However, some passes between `DecomposeInt32ClampPass` and `MatchArgRanksPass` cause the input ranks to become unmatched. We can rematch them in `MatchArgRanksPass`.

This issue was found by some internal model tests and cannot be easily reproduced with an op test, but we have verified the fix against those internal models.

Thank you!

zingo:
Arm tests might fail  on test_index_tensor_tosa @oscarandersson8218 is working on a fix 
https://github.com/pytorch/executorch/pull/16220

Sebastian-Larsson (approved):

PR #16184, opened 2025-12-10
Expose static llama in OSS

metascroy:
Summary:
This exposes a static llama model for CoreML.

We want to unify development behind one static model so that we do not need to apply fixes in multiple places (e.g., iOS 26 fixes).

Differential Revision: D88875745

### DIFF
--- a/examples/apple/coreml/llama/export_static_llm_coreml.py
+++ b/examples/apple/coreml/llama/export_static_llm_coreml.py
@@ -1,0 +1,504 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Export script for static attention LLM models to CoreML via ExecuTorch.
+
+Usage:
+    python export_static_llm_coreml.py \
+        --checkpoint /path/to/model.pth \
+        --params /path/to/params.json \
+        --output static_llm_coreml_model.pte \
+        --max_context_len 1024 \
+        --input_len 32 \
+        --embedding_quantize 4,32 \
+        --coreml_quantize c4w \
+        --target_split_size 1048
+"""
+
+import argparse
+import json
+
+import coremltools as ct
+import torch
+import torch.nn as nn
+import torch.utils._pytree as pytree
+
+from executorch.backends.apple.coreml.compiler import CoreMLBackend
+from executorch.backends.apple.coreml.partition import CoreMLPartitioner
+from executorch.examples.apple.coreml.llama.utils import (
+    replace_linear_with_split_linear,
+)
+from executorch.examples.models.llama.llama_transformer import construct_transformer
+from executorch.examples.models.llama.model_args import ModelArgs
+from executorch.examples.models.llama.rope import Rope
+from executorch.examples.models.llama.static_attention import StaticAttentionIOManager
+from executorch.exir import EdgeCompileConfig, to_edge_transform_and_lower
+from executorch.exir.backend.utils import format_delegated_graph
+from executorch.exir.capture._config import ExecutorchBackendConfig
+from executorch.exir.passes import MemoryPlanningPass
+from executorch.exir.passes.sym_shape_eval_pass import ConstraintBasedSymShapeEvalPass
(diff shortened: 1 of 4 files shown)


billmguo (approved):

examples/apple/coreml/llama/export_static_llm_coreml.py
```diff
@@ -0,0 +1,504 @@
...
+
+
+class ExecutorchGraphBreakModule(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, *args, **kwargs):
+        return tuple(
+            (
+                torch.ops.executorch_utils.graph_break.Tensor(a)
```
JacobSzwejbka, line 66: Oh we should make a tutorial for this op. I think it could be broadly useful to backends.

examples/apple/coreml/llama/export_static_llm_coreml.py
```diff
@@ -0,0 +1,504 @@
...
+
+    # Generate metadata for C++ runner
+    print("\nGenerating metadata for C++ runner...")
+    constant_methods = _get_metadata(
+        model_args, example_inputs, args.input_len, cache_len, float_dtype
+    )
+
+    # Setup CoreML partitioner
+    print("\nSetting up CoreML partitioner...")
+    compile_specs = CoreMLBackend.generate_compile_specs(
```
JacobSzwejbka, line 456: Do we have a recipe we could use? Could you add one?

examples/apple/coreml/llama/export_static_llm_coreml.py
```diff
@@ -0,0 +1,504 @@
...
+    print("\nDelegated program:")
+    print(format_delegated_graph(edge_manager.exported_program().graph_module))
+
+    # Convert to ExecuTorch
+    print("\nConverting to ExecuTorch...")
+    remove_graph_break_(edge_manager)
+    executorch_program = edge_manager.to_executorch(
+        ExecutorchBackendConfig(
+            extract_delegate_segments=True,
+            do_quant_fusion_and_const_prop=True,
```
JacobSzwejbka, line 490: Apple path leaves quant ops in the graph?

examples/apple/coreml/llama/export_static_llm_coreml.py
```diff
@@ -0,0 +1,504 @@
...
+            memory_planning_pass=MemoryPlanningPass(
+                alloc_graph_input=False, alloc_graph_output=False
+            ),
+            sym_shape_eval_pass=ConstraintBasedSymShapeEvalPass(),
```
JacobSzwejbka, line 494: This is the default now fwiw

examples/apple/coreml/llama/export_static_llm_coreml.py
```diff
@@ -0,0 +1,504 @@
...
+
+    print("\nDelegated program:")
+    print(format_delegated_graph(edge_manager.exported_program().graph_module))
+
+    # Convert to ExecuTorch
+    print("\nConverting to ExecuTorch...")
+    remove_graph_break_(edge_manager)
+    executorch_program = edge_manager.to_executorch(
+        ExecutorchBackendConfig(
+            extract_delegate_segments=True,
```
JacobSzwejbka, line 489: this one is also the default

examples/apple/coreml/llama/export_static_llm_coreml.py
metascroy: We could move it out of this file into executorch/exir?

examples/apple/coreml/llama/export_static_llm_coreml.py
metascroy: Core ML has a recipe.  I think the default recipe will likely work here, but I'm so used to using the partitioner.

examples/apple/coreml/llama/export_static_llm_coreml.py
metascroy: No, I guess this doesn't need to be specified.

examples/apple/coreml/llama/export_static_llm_coreml.py
metascroy: ConstraintBasedSymShapeEvalPass is default now?  Or the memory planning one?

PR #16186, opened 2025-12-10
[ez][ET-VK] Small fix for choose_qparams_affine_impl

SS-JIA:
* #16187
* __->__ #16186

It seems that `choose_qparams_affine` has recently appended some arguments to the schema. This causes newly exported models to break because at runtime, the output arg can no longer be found.

Fix by locating the output argument as the last entry in the args vector, rather than continuously incrementing the args index.

Update quantize/dequantize ops as well since it seems quantized_decomposed namespace ops are subject to change in the future.

Note that it would be good to do this for all operators in the Vulkan backend as a later refactor.

Differential Revision: [D88887463](https://our.internmc.facebook.com/intern/diff/D88887463/)

### DIFF
--- a/backends/vulkan/runtime/graph/ops/impl/QuantizeDequantize.cpp
+++ b/backends/vulkan/runtime/graph/ops/impl/QuantizeDequantize.cpp
@@ -369,7 +369,8 @@ void add_unpack_4w4c_and_dequantize_node(
 void quantize_per_tensor_impl(
     ComputeGraph& graph,
     const std::vector<ValueRef>& args) {
-  int32_t arg_idx = 0;
+  size_t arg_idx = 0;
+  size_t last_arg_idx = args.size() - 1;
   const ValueRef fp_input = args[arg_idx++];
   const ValueRef scale = args[arg_idx++];
   const ValueRef zero_point = args[arg_idx++];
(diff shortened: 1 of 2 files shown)


backends/vulkan/runtime/graph/ops/impl/QuantizeDequantize.cpp
```diff
@@ -392,7 +393,8 @@ void quantize_per_tensor_impl(
 void dequantize_per_tensor_impl(
     ComputeGraph& graph,
     const std::vector<ValueRef>& args) {
-  int32_t arg_idx = 0;
+  size_t arg_idx = 0;
+  size_t last_arg_idx = args.size() - 1;
```
JacobSzwejbka, line 397: For the ones in between do you just rely on default behavior? What if they are serialized with values != default? Shouldnt you error out?

trviv (approved):

backends/vulkan/runtime/graph/ops/impl/QuantizeDequantize.cpp
SS-JIA: yeah, that's a good point. I have a planned updated to improve arg checking for quantized_decomposed ops since there are currently a lot of unsupported input cases which are not accounted for - I will include this in that update.

The primary purpose of this PR as-is is to recover a currently broken CI signal, so I would prefer to keep it as simple as possible. In practice, not validating the args should be ok (for now) since the quantized_decomposed ops are inserted by a quantization workflow and Vulkan doesn't really work with non-supported quant workflows anyways 😛

PR #16199, opened 2025-12-11
Arm backend: Update vgf model tests

oscarandersson8218:
### Summary
Updates VgfTestPipeline to also take an optional argument quantize which
defaults to True. Also updates vgf tests in models/ to use the quantize flag
instead of passing a tosa_version to VgfTestPipeline. Also rename the tests
from vgf_INT/FP to vgf_quant and vgf_no_quant.


zingo (approved):
Approved, just make sure CI is green

PR #16215, opened 2025-12-11
Revert "Add resume for download function"

cccclai:
Reverts pytorch/executorch#15948

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -13,32 +13,32 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull.yml
```diff
@@ -13,32 +13,32 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-qnn-wheel-packages-linux:
```
larryliu0820, line 16: Can you just add 
```
if: false
```
Instead of commenting out the whole thing.

kirklandsign (approved):

PR #16219, opened 2025-12-12
fix(apple): Handle partial UTF-8 sequences in streaming LLM output

ShunL12324:
### Summary

ByteLevel BPE tokenizers (like GPT-2/SmolLM) can split multi-byte UTF-8 characters across token boundaries during streaming text generation. For example, the Chinese character "清" (UTF-8: `E6 B8 85`) might be split into two tokens: `æ¸` (`E6 B8`) and `ħ` (`85`).

This causes garbled output when displaying CJK characters and other multi-byte Unicode content on Apple platforms.

**Solution:** Add a `UTF8StreamingBuffer` class in `ExecuTorchLLMTextRunner.mm` that:
- Accumulates bytes until complete UTF-8 sequences are formed
- Validates continuation bytes for each sequence
- Rejects invalid bytes (overlong encodings `0xC0`, `0xC1`, out-of-range `0xF5+`)
- Skips lone continuation bytes to prevent buffer accumulation
- Flushes remaining valid bytes at generation end

### Test plan

- Tested with SmolLM model generating Chinese text - characters now display correctly instead of garbled output
- Tested with ASCII-only output - no regression
- Built debug framework (`./scripts/build_apple_frameworks.sh --Debug`) and verified fix on iOS device

### DIFF
--- a/extension/llm/apple/ExecuTorchLLM/Exported/ExecuTorchLLMTextRunner.mm
+++ b/extension/llm/apple/ExecuTorchLLM/Exported/ExecuTorchLLMTextRunner.mm
@@ -11,9 +11,140 @@
 #import "ExecuTorchLLMError.h"
 
 #import <executorch/extension/llm/runner/text_llm_runner.h>
+#import <memory>
 
 using namespace executorch::extension;
 using namespace executorch::runtime;
+
+namespace {
+
+/// A streaming UTF-8 buffer that accumulates bytes until complete UTF-8
+/// sequences are formed. This handles the case where BPE tokenizers output
+/// partial multi-byte UTF-8 sequences across token boundaries.
+///
+/// For example, the Chinese character "清" (UTF-8: E6 B8 85) might be split
+/// across two tokens: "æ¸" (E6 B8) and "ħ" (85). This buffer accumulates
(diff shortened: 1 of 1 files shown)


larryliu0820:
@shoumikhin can you take a look?

extension/llm/apple/ExecuTorchLLM/Exported/ExecuTorchLLMTextRunner.mm
```diff
@@ -11,10 +11,141 @@
...
 
+namespace {
+
+/// A streaming UTF-8 buffer that accumulates bytes until complete UTF-8
+/// sequences are formed. This handles the case where BPE tokenizers output
+/// partial multi-byte UTF-8 sequences across token boundaries.
+///
+/// For example, the Chinese character "清" (UTF-8: E6 B8 85) might be split
+/// across two tokens: "æ¸" (E6 B8) and "ħ" (85). This buffer accumulates
+/// bytes and only emits complete, valid UTF-8 strings.
```
larryliu0820, line 27: I'm actually curious if this belong to the realm of tokenizer. What tokenizer are you using?

PR #16221, opened 2025-12-12
Fix double-promotion in rsqrt for reduced-precision types

perheld:
Keep the bfloat16/fp16 rsqrt helper in float precision so Zephyr’s -Wdouble-promotion (treated as an error) no longer fires when ExecuTorch is built for Cortex-M targets.

Change-Id: I8fbcc99d8aebac67c4164f2ad502e9b3081386b4

### DIFF
--- a/runtime/core/portable_type/c10/c10/util/BFloat16-math.h
+++ b/runtime/core/portable_type/c10/c10/util/BFloat16-math.h
@@ -1,3 +1,10 @@
+/*
+ * Copyright 2025 Arm Limited and/or its affiliates.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
 #pragma once
 
 #include <c10/util/BFloat16.h>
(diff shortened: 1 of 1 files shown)


perheld:
@digantdesai is c10 something that is copied from pytorch and needs fixing upstream there if there is a needed change?

zingo:
Hi @SS-JIA / @swolchok is this something we want to do or if from PyTorch maybe we rather not touch it?

And if fixed maybe this can be  removed from the file:

> C10_CLANG_DIAGNOSTIC_PUSH()
> #if C10_CLANG_HAS_WARNING("-Wimplicit-float-conversion")
> C10_CLANG_DIAGNOSTIC_IGNORE("-Wimplicit-float-conversion")
> #endif

zingo:
@digantdesai / @SS-JIA Is this fix/update something we want to do? A side effect is that It would also remove some build warnings in the logs.

digantdesai (approved):
Good catch!

perheld:
> Good catch!

Thanks! But I guess this needs to go into pytorch and the c10 directory there. The check-c10-sync script is there for that reason. I shall try to fix a MR there instead.

perheld:
I guess when pytorch is bumped we get new c10 files.

PR #16222, opened 2025-12-12
Use default FakeTensorMode when calling module without input

Erik-Lundell:
Before, _ExportPassBase set self.tracer.fake_tensor_mode to a default value, but didn't use it when tracing. This caused operators that only have fake tensor implementations to crash.


Erik-Lundell:
It is hard for me to asses the side effects of this change but I think it makes sense. Can I get help getting the correct person to review this? @digantdesai

exir/pass_base.py
```diff
@@ -593,14 +594,13 @@ def call(self, graph_module: fx.GraphModule) -> PassResult:
                 ), "Multiple fake tensor mode detected."
                 fake_tensor_mode = i.fake_mode
         if fake_tensor_mode is None:
-            self.tracer.fake_tensor_mode = FakeTensorMode(allow_non_fake_inputs=True)
-            fake_tensor_mode = nullcontext()  # type: ignore[assignment]
```
JacobSzwejbka, line 636: @angelayi  @tugsbayasgalan  what is the difference between None and nullcontext() for fake tensor mode

Erik-Lundell: Bumping this @angelayi @tugsbayasgalan

angelayi (approved):

exir/pass_base.py
angelayi: @JacobSzwejbka fake tensor mode is a context manager so "nullcontext" is like the "None" version for fake tensor mode since we can't do "with None"

@Erik-Lundell thanks for the change (and bump!)

PR #16223, opened 2025-12-12
NXP backend: Add documentation for channels last dim order support in Neutron backend + example in `aot_neutron_compile.py`.

MartinPavella:
### Summary
This PR adds a markdown documentation for the channels last dim order support from the NXP backend, and an example use in the `aot_neutron_compile.py`.

### Test plan
N/A

cc @robert-kalmar

### DIFF
--- a/examples/nxp/aot_neutron_compile.py
+++ b/examples/nxp/aot_neutron_compile.py
@@ -190,6 +190,14 @@ if __name__ == "__main__":  # noqa C901
         help="Visualize the lowered program. `show` launches a browser tab with the visualization. `store` stores the "
         "visualization in a json file for later inspection. See `docs/source/visualize-with-clusters.md` for details.",
     )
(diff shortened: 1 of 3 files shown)


examples/nxp/aot_neutron_compile.py
```diff
@@ -206,6 +214,15 @@ def get_model_and_inputs_from_name(model_name: str):
     )
     model = model.eval()
 
+    if args.use_channels_last_dim_order:
+        # Turn the model to channels last.
+        model.to(memory_format=torch.channels_last)
```
jirioc, line 219: Probably the assignment is missing:  `model = model.to(...)`

MartinPavella: It is actually not necessary. The call `.to()` modifies the original instance.

jirioc (approved):

PR #16231, opened 2025-12-12
Update ScalarToTensorPass to correctly track the modified bit

DrJessop:
Summary: Update ScalarToTensorPass to correctly track the modified bit, only run export pass if there are changes to the graph.

Also updated test_export_scalar_to_tensor_pass. to_edge already implicitly ran the pass, so the additional pass was idempotent. The updated test manually constructs a graph, checks that the pass modified the graph via graph examination and running the interpreter to check for numerical correctness.

Differential Revision: D89085647

### DIFF
--- a/exir/passes/scalar_to_tensor_pass.py
+++ b/exir/passes/scalar_to_tensor_pass.py
@@ -7,10 +7,14 @@
 # pyre-strict
 
 import torch
-from executorch.exir.pass_base import ExportPass, map_args
+from executorch.exir.pass_base import ExportPass, map_args, PassResult
 
 
 class ScalarToTensorPass(ExportPass):
+    def __init__(self) -> None:
+        super().__init__()
+        self._modified: bool = False
+
     # pyre-ignore
     def call_operator(self, op, args, kwargs, meta):
         # pyre-ignore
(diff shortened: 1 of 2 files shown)


exir/passes/scalar_to_tensor_pass.py
```diff
@@ -7,26 +7,47 @@
...
+            for node in module.graph.nodes:
+                if not isinstance(
+                    node.target, (torch._ops.OpOverload, torch._ops.OpOverloadPacket)
+                ):
+                    continue
+                changed |= self.maybe_remove_or_replace(node)
+
+        if changed:
+            graph_module.recompile()
+            return super().call(graph_module)
```
JacobSzwejbka: Why bother calling super anymore? Wont this just walk the graph doing a bunch of nothing?

DrJessop: Sorry, at that point the diff should have been in draft state. I just updated the diff to make minimal changes to the pass, just track whether there are any modifications or not.

mcremon-meta:
@larryliu0820 @JacobSzwejbka could we get eyes on this please? :)

abeakkas (approved):
Approved with comments on Phabricator.

PR #16233, opened 2025-12-12
Cortex-M: Add depthwise conv2d operator

### LINKED ISSUE #16105: DepthwiseConv : Add support to CMSiS-NN
See here for full ops support : https://github.com/ARM-software/CMSIS-NN?tab=readme-ov-file#mve-extension

rascani:
### Summary
Add quantized depthwise convolution operator for the Cortex-M backend using CMSIS-NN's optimized arm_depthwise_conv_wrapper_s8 function.

Fixes #16105

### Test plan
```
./backends/cortex_m/test/build_test_runner.sh
pytest --config-file=backends/arm/test/pytest.ini backends/cortex_m/test/ops/test_conv.py
```


backends/cortex_m/passes/convert_to_cortex_m_pass.py
```diff
@@ -156,11 +156,24 @@ def _get_convolution_replacement(self, node) -> int:
...
-        # Permute the weight tensor to the OHWI layout expected by CMSIS-NN.
         weight_tensor = get_param_tensor(self.exported_program, weight)
-        weight_permuted = weight_tensor.permute(0, 2, 3, 1).contiguous(
-            memory_format=torch.channels_last
-        )
+
+        # Detect depthwise convolution:
+        # PyTorch depthwise weight is [out_ch, 1, H, W] where dimension 1 is 1
+        # and groups == input_channels (groups > 1)
+        is_depthwise = weight_tensor.shape[1] == 1 and groups > 1
```
mansnils: I think groups could be 1 for a DW conv?
So a better condition is then, is_depthwise = (in_channels == groups) and (out_channels % in_channels) == 0

backends/cortex_m/passes/convert_to_cortex_m_pass.py
```diff
@@ -156,11 +156,24 @@ def _get_convolution_replacement(self, node) -> int:
...
+
+        if is_depthwise:
```
mansnils: Here we actually have the benefit of choosing between a regular and DW conv. It is likely but not certain that the un-optimized CMSIS-NN DW conv or the one without any SIMD is less efficient that the corresponding CMSIS-NN conv. We don't know exactly until we measure. We could then add something like this for now with a TODO comment:
        optimal_dw_conv_constraints = (
            in_channels == out_channels and dilation == [1, 1]
        ) or in_channels == 1

mansnils (changes requested):
Thanks for this @rascani ! It looks good, just a couple of comments.

backends/cortex_m/ops/op_quantized_depthwise_conv2d.cpp
```diff
@@ -0,0 +1,274 @@
...
+      bias.has_value() ? bias.value().const_data_ptr<int32_t>() : nullptr;
+
+  cmsis_nn_context cmsis_context;
+  cmsis_context.buf = nullptr;
+  cmsis_context.size = 0;
+
+  const size_t buffer_bytes =
+      static_cast<size_t>(arm_depthwise_conv_wrapper_s8_get_buffer_size(
+          &dw_conv_params, &input_dims, &filter_dims, &output_dims));
+  if (buffer_bytes > 0) {
```
psiddh: If the return val of  'arm_depthwise_conv_wrapper_s8_get_buffer_size' returns 'signed int' , should we add an  explicit '< 0'  check before we cast to size_t, because your current > 0 check happens after the potentially dangerous cast from signed to unsigned. (Just thinking aloud here..)

const int32_t raw_buffer_size = arm_depthwise_conv_wrapper_s8_get_buffer_size(
    &dw_conv_params, &input_dims, &filter_dims, &output_dims);
if (raw_buffer_size < 0) {
    ET_LOG(Error, "quantized_depthwise_conv2d_out: CMSIS-NN buffer size calculation failed");
    context.fail(Error::Internal);
    return out;
}

backends/cortex_m/test/ops/test_conv.py
```diff
@@ -178,6 +222,73 @@ def forward(self, x):
             ramp_tensor(0, 10, (1, 3, 8, 8)).to(memory_format=torch.channels_last),
         ),
     ),
+    # Depthwise convolution tests (groups == in_channels)
```
psiddh, line 225: pretty comprehensive set of tests...just two edge cases come to my mind.

1. Single Channel Edge Case
2. Minimal Spatial Dimensions

"depthwise_conv2d_single_channel": McuTestCase(
    model=CortexMDepthwiseConv2D(1, 2, 3, groups=1),  # Minimal groups=1 case
    example_inputs=(
        ramp_tensor(0, 10, (1, 1, 4, 4)).to(memory_format=torch.channels_last),
    ),
),

"depthwise_conv2d_minimal_spatial": McuTestCase(
    model=CortexMDepthwiseConv2D(2, 2, 3, padding=1, groups=2),  # 3x3 kernel on 1x1 input
    example_inputs=(
        ramp_tensor(0, 10, (1, 2, 1, 1)).to(memory_format=torch.channels_last),
    ),
),

rascani: Done.

backends/cortex_m/passes/convert_to_cortex_m_pass.py
rascani: Done.

backends/cortex_m/ops/op_quantized_depthwise_conv2d.cpp
rascani: Good catch, done.

backends/cortex_m/passes/convert_to_cortex_m_pass.py
rascani: Good idea. I tried to add these constraints, which resulted in falling back to conv2d for some of the tests cases. Unfortunately, that also resulted in some output mismatches. I filed #16347 to revisit this and left a TODO to improve the smarts here.

rascani (commented):
Thanks for the reviews @mansnils and @psiddh!

psiddh (approved):
lgtm

psiddh (approved):

mansnils (approved):

zingo:
Hi this seem to make a cortex-m CI tests fails see https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=(arm-backend%7Ccortex)&useRegexFilter=true

rascani:
> Hi this seem to make a cortex-m CI tests fails see https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=(arm-backend%7Ccortex)&useRegexFilter=true

Thanks @zingo . I missed that cortex-m CI tests were not run by default on the PR. I fixed up the failing activation tests in #16379.

PR #16235, opened 2025-12-12
Avoid copying output from GPU to CPU for ASR runner

larryliu0820:
### Summary

This PR adds the ability to skip copying GPU outputs back to CPU for specific methods in the CUDA backend, and enables conditional CUDA compilation for the ASR runner.

#### CUDA Backend (`backends/cuda/runtime/cuda_backend.cpp`)

- Changed the `skip_copy_output_to_cpu_for_method` backend option from a boolean to a **string** that accepts a method name
- The option can be set via `set_option()` after `init()` is called, allowing runtime configuration
- During `execute()`, the backend compares the configured method name against the handle's method name to decide whether to skip the GPU→CPU output copy
- Thread-safe access to the skip-copy method name via mutex

**Usage example:**
```
BackendOptions<1> options;
options.set_option("skip_copy_output_to_cpu_for_method", "encode");
set_option("CudaBackend", options.view());
```
#### AOTI Delegate Handle (`backends/aoti/aoti_delegate_handle.h`)

- Added `method_name` field to `AOTIDelegateHandle` to track which method each delegate handle corresponds to

#### ASR Runner CMake (`extension/asr/runner/CMakeLists.txt`)

- Added conditional CUDA support: when `EXECUTORCH_BUILD_CUDA` is enabled and `CUDAToolkit` is found, the `CUDA_AVAILABLE` compile definition is added
- This allows ASR runner code to conditionally compile CUDA-aware code paths

### Motivation

When running multi-method models (e.g., prefill + decode for LLMs, or encoder + decoder for ASR), some methods benefit from keeping outputs on GPU to avoid unnecessary memory copies between methods. This change enables fine-grained control over which method(s) skip the GPU→CPU copy.

### Test Plan

- [ ] Build with CUDA enabled
- [ ] Verify `skip_copy_output_to_cpu_for_method` option works for specified method
- [ ] Verify other methods still copy outputs to CPU by default

Perf improvement comparing with main:

Screenshot 2025-12-15 at 10 32 13 AM


backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -365,6 +443,10 @@ class ET_EXPERIMENTAL CudaBackend final
     delete handle;
     clear_all_tensors();
   }
+
+ private:
+  mutable std::mutex skip_copy_method_mutex_;
```
mergennachin, line 448: Do you need mutex at all? 

We enforce the callers of ExecuTorch to be aware of thread safety, and don't guarantee any thread safety within the internals of ET.

mergennachin (commented):
Great -- simplifies quite a bit. 

Three comments: 

1. Can you bring RAII cleanup pattern from https://github.com/pytorch/executorch/pull/16060. We should ensure GPU tensors are properly cleaned up on error paths.
```
  struct TensorCleanup {
      std::vector<AOTITensorHandle>& tensors;
      ~TensorCleanup() {
          for (auto* handle : tensors) {
              if (handle != nullptr) {
                  aoti_torch_delete_tensor_object(handle);
              }
          }
      }
  };
```

2. Can you add a simple clear_gpu_outputs option to free GPU memory when done with encoder-decoder loops

3. Even in the non-error case, isn't it leaking memory? How/when are the GPU tensors deleted?

larryliu0820:
> 3\. Even in the non-error case, isn't it leaking memory? How/when are the GPU tensors deleted?

Each `CudaBackend` instance will keep references to the GPU tensors (both encoder and decoder) and will be freed when it's getting destroyed.

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: I see the pattern in xnnpack so would like to keep it the same way. https://github.com/pytorch/executorch/blob/1ea3907ffa0e71ca0e2c9e536c36b6b4b5fb84a3/backends/xnnpack/runtime/XNNPACKBackend.cpp#L252

larryliu0820:
> > 3. Even in the non-error case, isn't it leaking memory? How/when are the GPU tensors deleted?
> 
> Each `CudaBackend` instance will keep references to the GPU tensors (both encoder and decoder) and will be freed when it's getting destroyed.

@Gasoonjia can you comment on if this still holds true, after slimtensor migration?

larryliu0820:
> * Can you add a simple clear_gpu_outputs option to free GPU memory when done with encoder-decoder loops
> *

I don't think we need that, if the destroy() hook is reliable.

Gasoonjia:
> > > 3. Even in the non-error case, isn't it leaking memory? How/when are the GPU tensors deleted?
> > 
> > 
> > Each `CudaBackend` instance will keep references to the GPU tensors (both encoder and decoder) and will be freed when it's getting destroyed.
> 
> @Gasoonjia can you comment on if this still holds true, after slimtensor migration?


i think it is no longer the case if we migrate to slimtensor since at that time the memory will be controled in tensor level.

But for this case i think we can do something simliar, e..g making `cudabackend` keep the reference to the output slim tensor to make it still alive after preprocessing and freed in `destroy` function, to make sure the pipeline still works

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -222,9 +292,33 @@ class ET_EXPERIMENTAL CudaBackend final
...
+
+      ~TensorCleanup() {
+        // Clean up input tensors
+        for (auto* handle : inputs) {
+          if (handle != nullptr) {
+            aoti_torch_delete_tensor_object(handle);
+          }
+        }
+        // Clean up output tensors
+        for (auto* handle : outputs) {
```
Gasoonjia: I think we may need to keep the output tensors alive, otherwise the another round of input may failed on getting the data.

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -365,6 +467,10 @@ class ET_EXPERIMENTAL CudaBackend final
     delete handle;
     clear_all_tensors();
   }
+
+ private:
+  mutable std::mutex skip_copy_method_mutex_;
+  std::string skip_copy_method_;
```
Gasoonjia, line 449: shouldn't it be an arrary of string cus we may skip copy on multiple methods? Its ok to be update in the following PRs.

Gasoonjia (commented):
LGTM, only some minor feedback. thanks for optimize the data pipeline!

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: Yeah next PR will be supporting a comma separated string.

backends/cuda/runtime/cuda_backend.cpp
larryliu0820: ok I think that's fair, basically we don't want to cleanup outputs when exiting execute() and probably will rely on destroy() for cleaning it up.

Gasoonjia (approved):

requirements-examples.txt
```diff
@@ -4,4 +4,4 @@ datasets == 3.6.0 # 4.0.0 deprecates trust_remote_code and load scripts. For now
 timm == 1.0.7
 torchsr == 1.0.4
 torchtune >= 0.6.1
-transformers == 4.56.1
+transformers == 5.0.0rc1
```
zingo, line 7: This unfortunately broke some tests under trunk, we are looking into how to fix it. 

backends/arm/test/models/test_T5ForConditionalGeneration_arm.py::test_T5ForConditionalGeneration_tosa_FP
backends/arm/test/models/test_T5ForConditionalGeneration_arm.py::test_T5ForConditionalGeneration_tosa_INT

Just adding this comment so people can find it if they dig for the problem. before we have a fix :)

PR #16236, opened 2025-12-12
Add 'all' option for android flavor

kirklandsign:
### Summary
xnnpack + qnn + vulkan

### Test plan
CI


psiddh:
Nice, couple of questions
- what is the APK size increase as a result ?
- Also did you do any manual testing to confirm that no extra steps are required ?

kirklandsign:
> Nice, couple of questions
> 
> * what is the APK size increase as a result ?

Minimal increase because we it's really the sum of vulkan + qnn

> * Also did you do any manual testing to confirm that no extra steps are required ?

Will do

psiddh (approved):

PR #16245, opened 2025-12-15
Arm backend: Update vgf ops tests [Part 1]

oscarandersson8218:
Updates the the first half of vgf op tests to use the quantize flag to VgfTestPipeline instead of passing a tosa_version.


zingo (approved):

zingo (approved):

zingo:
Change many files but only tests internal stuff so should be fine from buck2 perspective, macos and openvino fails are unrelated.

PR #16249, opened 2025-12-15
Arm backend: Test partial quantization of models

martinlsm:
Run tests of partial quantization on MobileNetV2 and Llama using the INT+FP profile. The tests quantize only some layer types and runs the rest in floating point.


oscarandersson8218 (approved):
Unrelated failures

PR #16274, opened 2025-12-16
Arm backend: Fix two control flow bugs

Erik-Lundell:
Arm backend: Regularize submodules before processing them.
    
    Add a function _regularize_submodule that contains special
    handling needed for submodules. This mainly solves two problems:
    - Buffers in submodules are (currently) handled differently depending
      on whether they are from tracing or added in passes. The old solution
      tried avoiding having to add a new meta field, but was brittle.
      The new solution simply marks all all placeholders before passes.
    - The pass pipeline assumes the dim_order of inputs and outputs to match
      the actual dim_order (tosa_dim_order). We need to ensure this.

------------------------

    Arm backend: Fix while quantization
    
    The output of the while loop body can either re-enter
    the body, or exit the while loop. Therefore, A and B
    in the diagrambelow need to share the same quantization parameters.
    A -> while ( RESCALE -> ... -> RESCALE -> ) -> B
    
    Earlier tests happened to get equal qparams on the
    input and output, but this is not the general case.

### DIFF
--- a/backends/arm/process_node.py
+++ b/backends/arm/process_node.py
@@ -15,7 +15,6 @@ from executorch.backends.arm.operators.node_visitor import NodeVisitor
 from executorch.backends.arm.tosa.mapping import TosaArg, TosaSpecialDtype
 from executorch.backends.arm.tosa.specification import TosaSpecification
 from executorch.backends.arm.tosa.utils import tosa_shape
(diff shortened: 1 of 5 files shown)


Erik-Lundell:
Failures unrelated. Should be a safe merge.

zingo (approved):

PR #16275, opened 2025-12-16
Arm backend: Add processed module to global scope in aot_arm_compiler

Erik-Lundell:
When processing globals, torch._dynamo.symbolic_trace loads them by calling import_lib.import_module. For this to succeed, the module aot_arm_compiler loads from a source file needs to be added to sys.modules.


Erik-Lundell:
Failures unrelated. Should be a safe merge.

zingo (approved):

PR #16276, opened 2025-12-16
NXP backend: added aten.split support

novak-vaclav:
### Summary

adds support for aten.split operator

### Test plan

cc @robert-kalmar @MartinPavella


backends/nxp/aten_passes/decompose_split_to_slices_pass.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Sometimes chunks are in tuples
+        if isinstance(split_nodes_chunks, tuple):
+            split_nodes_chunks = list(split_nodes_chunks)
+
+        if not isinstance(split_nodes_chunks, list):
+            raise RuntimeError("Faulty split chunks")
+
+        # Get split dim
+        dim = -1
```
MartinPavella: Why are you guessing the `dim`? Is it not possible to read it from the arguments?

backends/nxp/aten_passes/decompose_split_to_slices_pass.py
```diff
@@ -0,0 +1,203 @@
...
+                getitem_node = getitem_nodes[i]
+                getitem_node.replace_all_uses_with(slice_node)
+
+                self.graph_module.graph.erase_node(getitem_node)
+
+        # Wire split node correctly to the input node
+        split_node.replace_all_uses_with(input_node)
+        self.graph_module.graph.erase_node(split_node)
+
+        return True
```
MartinPavella: Why `return True`? I don't see any other possible return value.

backends/nxp/aten_passes/decompose_split_to_slices_pass.py
```diff
@@ -0,0 +1,203 @@
...
+                continue
+
+            # Get split args
+            split_node = node
+            input_node = split_node.all_input_nodes[0]
+            split_nodes_chunks = split_node.meta["val"]
+
+            # Check if split is even necessary - if not, remove it
+            if len(split_nodes_chunks) == 1:
+                getitem_node = split_node.next
```
MartinPavella: The `split_node.next` is merely the node that is stored right after the `split_node` in the `graph.nodes`. The nodes don't have to be connected at all. Please use something like `list(split_node.users)[0]` instead.

backends/nxp/tests/test_decompose_split_to_slices.py
```diff
@@ -0,0 +1,243 @@
...
+@pytest.mark.parametrize(
+    "input_shape, split_size, dim",
+    [
+        pytest.param((8,), 3, 0, id="1D."),
+        pytest.param((4, 8), 5, 1, id="2D."),
+    ],
+)
+def test_decompose_split_with_size(mocker, input_shape, split_size, dim):
+    model = SplitWithSize(split_size, dim)
+    example_input = torch.rand(input_shape)
```
MartinPavella, line 51: Please set a seed to make the tests deterministic. Add something like
```python
@pytest.fixture(autouse=True)
def reseed_model_per_test_run():
    torch.manual_seed(23)
    np.random.seed(23)
```
to the top of the file.

backends/nxp/tests/test_decompose_split_to_slices.py
```diff
@@ -0,0 +1,243 @@
...
+    assert len(outputs_before) == len(outputs_after)
+
+    for i in range(len(outputs_before)):
+        assert np.allclose(outputs_before[i], outputs_after[i])
+
+
+@pytest.mark.parametrize(
+    "input_shape, size_or_sections, dim",
+    [
+        # pytest.param((8, 4), 4, 1, id="2D, one chunk using split size."),
```
MartinPavella: Why is this commented out?

backends/nxp/tests/test_decompose_split_to_slices.py
MartinPavella, line 1: Very nice tests!
Would it be possible to add some end to end test? I would like to see a case where you use our regular pipeline (`to_quantized_edge_program()`), where the `split` gets correctly decomposed (without calling the `NeutronAtenPassManager` manually like you did here), and the `slice` nodes then get delegated.

backends/nxp/aten_passes/decompose_split_to_slices_pass.py
novak-vaclav: Yeah you are right, it just did not appear in the arguments if it was = 0 (default). Fixed.

backends/nxp/tests/test_decompose_split_to_slices.py
novak-vaclav: I forgot to uncomment it, sorry.

backends/nxp/tests/test_decompose_split_to_slices.py
novak-vaclav: Thank you! :))
I added such test, however it works only with Neutron Software version 2.2.1+. Otherwise I made the code skip the test.

novak-vaclav:
Fixed the issues found by @MartinPavella, please review them. Thank you!

backends/nxp/aten_passes/neutron_aten_pass_manager.py
MartinPavella: nit: 2026

backends/nxp/tests/test_decompose_split_to_slices.py
```diff
@@ -241,3 +259,51 @@ def test_decompose_split_with_one_chunk(mocker, input_shape, size_or_sections, d
...
+        model = SplitWithSections(size_or_sections, dim)
+    else:
+        model = SplitWithSize(size_or_sections, dim)
+
+    # Run conversion
+    edge_program = to_quantized_edge_program(
+        model,
+        input_shape
+    ).exported_program()
+    x = converter_spy.spy_return
```
MartinPavella: I don't see this variable `x` ever used.

backends/nxp/tests/test_decompose_split_to_slices.py
```diff
@@ -241,3 +259,51 @@ def test_decompose_split_with_one_chunk(mocker, input_shape, size_or_sections, d
...
+        [
+            torch.ops.aten.split.Tensor,
+            torch.ops.aten.split.default,
+            torch.ops.aten.split_with_sizes.default,
+        ],
+    )
+
+    # Capture generated model
+    tflite_flatbuffers_model, _ = converter_spy.spy_return
+    exported_program: ExportedProgram = converter_spy.call_args.args[1]
```
MartinPavella, line 300: It would be useful to add a check that the `exported_program` actually contains `slice` operators.

backends/nxp/tests/models.py
MartinPavella: nit: 2026

backends/nxp/tests/test_decompose_split_to_slices.py
```diff
@@ -0,0 +1,309 @@
...
+        edge_program.graph,
+        [
+            torch.ops.aten.split.Tensor,
+            torch.ops.aten.split.default,
+            torch.ops.aten.split_with_sizes.default,
+        ],
+    )
+
+    # Capture generated model
+    tflite_flatbuffers_model, _ = converter_spy.spy_return
```
MartinPavella: nit: calling this `neutron_ir_model` would be more accurate.

novak-vaclav:
Issues fixed. Thanks for the review and please re-review.

MartinPavella (approved):

roman-janik-nxp:
@novak-vaclav  Please do not use `List` and `Tuple` from `typing`, use `list` and `tuple` from base Python. This is possible since Python 3.10 and we use it in out codebase.

PR #16286, opened 2025-12-17
Arm backend: Disable relu-fusing for sym. quant

oscarandersson8218:
Adds new checks to quantization annotation to not fuse conv+relu patterns when output_qspec is symmetric. Symmetric quantization would force the zero-point to be 0, and in order for us to fuse relu the zero-point must equal qmin. Also adds new tests to verify that it works.


zingo:
Got fail, retrigger failed test job to see if it's random

FAILED backends/arm/test/ops/test_sub.py::test_sub_tensor_tosa_INT_alpha[rand_4D_pi] - AssertionError: Output 0 does not match reference output.
	Given atol: 0.001, rtol: 0.001.
	Output tensor shape: torch.Size([1, 10, 20, 30]), dtype: torch.float32
	Difference: max: 0.24429035186767578, abs: 0.24429035186767578, mean abs error: 4.071505864461263e-05.
	-- Model vs. Reference --
	 Numel: 6000, 6000
	Median: 0.0, 0.0
	  Mean: 0.06314905179540316, 0.06310833673675854
	   Max: 32.00203323364258, 32.00203323364258
	   Min: -30.292001724243164, -30.292001724243164

zingo:
it worked, other fails are unrelated

zingo (approved):

PR #16287, opened 2025-12-17
Arm backend: Improve while support

Erik-Lundell:
- Support additional_args for while by normalizing them to carried_args in a pass.
- Support not using all while outputs.

### DIFF
--- a/backends/arm/_passes/__init__.py
+++ b/backends/arm/_passes/__init__.py
@@ -102,6 +102,7 @@ from .insert_table_ops import InsertTableOpsPass  # noqa
 from .match_arg_dtype_pass import MatchArgDtypePass  # noqa
 from .match_arg_ranks_pass import MatchArgRanksPass  # noqa
 from .mm_to_bmm_pass import ConvertMmToBmmPass  # noqa
+from .normalize_while_initial_args_pass import NormalizeWhileInitialArgsPass  # noqa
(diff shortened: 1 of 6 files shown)


zingo:
Oh no :(  But mean abs error: is really low so it seem like it's flakey and unrelated, could be worth to just retrigger when the other tests have runned all the way.

FAILED backends/arm/test/ops/test_sdpa.py::test_sdpa_vgf_quant - AssertionError: Output 0 does not match reference output.
	Given atol: 0.006777485202997923, rtol: 0.001.
	Output tensor shape: torch.Size([1, 3, 197, 64]), dtype: torch.float32
	Difference: max: 0.01155497133731842, abs: 0.01155497133731842, mean abs error: 5.8043679665079164e-06.
	-- Model vs. Reference --
	 Numel: 37824, 37824
	Median: 0.0, 0.0
	  Mean: 0.0018774077386055387, 0.0018777132326251095
	   Max: 0.7048531770706177, 0.7048531770706177
	   Min: -0.6932982206344604, -0.6932982206344604

Erik-Lundell:
Since this pr touches while and the failing test i spda, I would assume it's a flaky test. Let me retrigger

zingo (approved):

zingo:
Worked, other fails unrelated.

PR #16291, opened 2025-12-17
Arm backend: Update docs to mention partial quantization

martinlsm:
VGF has now support for partial quantization, i.e., having the model run in mixed numerical precision. Update the markdown documentation to include and explain this feature.


zingo (approved):

zingo:
@martinlsm does the docc "Preview" link under  "Helpful Links" above look like you expect ?

martinlsm:
> @martinlsm does the docc "Preview" link under "Helpful Links" above look like you expect ?

Thanks for pointing that out. Yes, I verified it and it works like I expect it to.

zingo:
> > @martinlsm does the docc "Preview" link under "Helpful Links" above look like you expect ?
> 
> Thanks for pointing that out. Yes, I verified it and it works like I expect it to.

That is kind of neat that it has been added to the CI like this :) 
Thanks for checking

PR #16297, opened 2025-12-17
Arm backend: Simple example application for image classification

gggekov:
The goal is to create a minimalistic application that can explain  the ExecuTorch APIs to a person who is not familiar with ExecuTorch. The key is to make something simple and easy to understand.


zingo (approved):
@gggekov good PR :) Please just comment and mark Copilot resolved or even fix stuff if needed then I can merge :)

PR #16299, opened 2025-12-17
Clean up various leftover platform arg usage around the repo that never hit validation

8Keep:
Summary:
Cleaning up platform args before the whole feature gets deleted. These missed our validation.

---
These are to force android builds of mwa for xplat diffs

Reviewed By: mzlee

Differential Revision: D89322661


GregoryComer (approved):
Looks good to me. I'm not an expert on buck, but as long as OSS CI is green, we should be good. You will need to ensure that the internal diff is in sync with the PR, which can be a pain with internal-only changes.

PR #16304, opened 2025-12-18
[slim tensor migration 1/n]  copy and paste slim tensor into ExecuTorch

Gasoonjia:
* #16330
* #16329
* #16327
* #16314
* #16313
* __->__ #16304

This stack aims to migrate slim tensor into ExecuTorch stack to make it as internal tensor representation of cudabackend.

This diff aims to copy and paste slimtensor class into ExecuTorch codebase, with ZERO change on logic, namespace, class, etc. The ONLY TWO changes are the include pathes and buck target files.

However below are things are blocking slimtensor to land:
1. Inconsistent namespace. Now slimtensor has its own namespace structure (standalone::, standalone::c10, etc) which not inline with ExecuTorch.
2. Self-supported Exception Handling Strategy. slimtensor adopts STANDALONE_CHECK, STANDALONE_INTERNAL_ASSERT, STANDALONE_INTERNAL_ASSERT_DEBUG_ONLY, or directly throw std::runtime_error to check and raise error. We should align SlimTensor's error handling strategy with ExecuTorch's  approach.
3. Duplicated c10 functions. Now slimtensor owns c10 it is using , some of them have be in executorch/runtime/core/portable_type/c10/ and we should deduplicate.
4. Duplicated data types. Slimtensor holds and uses several datatypes that et has and it should just reuse them, like ArrayRef, Span, irange, IntArrayRef, etc.

This diff will not be landed until all things above have been done. In the diffs above it I will gradually update the slim tensor class to make it suitable for ET, and to make the udpate reviewer friendly.
There're other stuffs not perfect, but i think we can gradually solve them after all migration has been done.
1. Move all c10 files under executorch/runtime/core/portable_type/c10/. It would be great if we can gather all c10 header together.
2. Make new introed c10 files sync with pytorch/pytorch. those c10 files has been decoupled with pytorch/pytorch for a while and need some extra work to make them sync.
3. Reuse executorch macro

Differential Revision: [D89417354](https://our.internmc.facebook.com/intern/diff/D89417354/)


JacobSzwejbka:
Instead of copy pasting all the c10 headers can you depend on the already copy pasted ones in executorch core

Gasoonjia:
> Instead of copy pasting all the c10 headers can you depend on the already copy pasted ones in executorch core

yes that's in my plan. You can refer to the diff summary or PR's top comment for what i;m gonna doi

PR #15735, opened 2025-11-11
Qualcomm AI Engine Direct - QNN ExecuTorch Intermediate Output Debugger

winskuo-quic:
### Summary

- Enabled ExecuTorch QNN Intermediate Tensor Debugger.
- Provide an API for users to define their own metrics
- Offers a variety of output format to visualize the debug results: svg, csv, raw files.
- A README file and tutorial script to guide users on how to debug a model. Example script: `python examples/qualcomm/util_scripts/qnn_intermediate_debugger_demo.py -b build-android -m SM8550 --device $DEVICE --dataset ../imagenet-mini/val/ --dump_intermediate_outputs`

#### An example use case
MobileVit V2 has significant drop in accuracy in certain QNN versions, while QNN 2.29 has good accuracy. With the help of accuracy debugger, we have targeted the node native_group_norm_default_1 in the model. As shown below, in QNN 2.29, this node has a cos_similarity (QNN V.S. CPU) of 0.997, while all other QNN versions has cos_similarity of 0, which provides us some hint it is possibly this group_norm node that is causing accuracy drop.
image

#### What's Coming Next?
- Currently, we dump CPU outputs by manually inserting observer nodes. However, ExecuTorch actually has built in methods (intermediate_output_capturer) that could dump intermediate output for us, in format of a dict{debug_handle : tensor_output}. We will enable `debug_handle` and reuse https://github.com/pytorch/executorch/blob/main/devtools/inspector/_intermediate_output_capturer.py in future instead.
- Support graph with partitions
- Support LLM models

### Test plan
- E2E example script test
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleUtilsScript.test_intermediate_debugger  -s $DEVICE --model SM8650 --build_folder build-android/ --executorch_root . --image_dataset ../imagenet-mini/val/ --artifact ./e2e_test_debug`
- Simple model test
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_dump_intermediate_outputs_simple_model --model SM8550 --device $DEVICE --build_folder build-android`
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_dump_intermediate_outputs_topk --model SM8550 --device $DEVICE --build_folder build-android`

PR #5310, opened 2024-09-12
Qualcomm AI Engine Direct - Intermediate Tensor Dump

winskuo-quic:
## Summary

- Intermediate Tensor Dump enablement
- Added UT for intermediate tensor dump

PR #16316, opened 2025-12-18
Qualcomm AI Engine Direct - Support Debug Handle and Integrate IntermediateOutputCapturer

winskuo-quic:
### Summary

- [x] AOT Debug Handle Enablement. Also supported debug_handle map, although debugger did not use it. Enable this because of community request.
- [x] Runtime Debug Handle Enablement: Parse debug handle in runtime and use debug_handle and tensors key when storing result into etdump.
- [x] Reuse ExecuTorch Debugger feature and reduce redundancy between QNN ExecuTorch Debugger and ExecuTorch Debugger Utils.

#### Additional Topics:

- What is the official way of retrieving an edge module that does not carry backend info?

### Test plan
- E2E example script test
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleUtilsScript.test_intermediate_debugger  -s $DEVICE --model SM8650 --build_folder build-android/ --executorch_root . --image_dataset ../imagenet-mini/val/ --artifact ./e2e_test_debug`
- Simple model test
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_dump_intermediate_outputs_simple_model --model SM8550 --device $DEVICE --build_folder build-android`
   - `python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedUtils.test_qnn_backend_dump_intermediate_outputs_topk --model SM8550 --device $DEVICE --build_folder build-android`


Gasoonjia:
Is the PR ready to be reviewed now?

backends/qualcomm/_passes/resolve_debug_handle.py
```diff
@@ -0,0 +1,44 @@
...
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class ResolveDebugHandle(ExportPass):
+    """
+    Caution: This pass is executed as the last of the edge_passes.
+    For any passes executed during qnn_preprocess, users will need to handle debug_handle ID themselves.
+
+    Description: During passes transformation, some passes might be copying some node's meta when creating a new node,
+    which means multiple nodes might be sharing the same debug_handle ID while it shouldn't.
```
Gasoonjia, line 19: im not super understand here: if several nodes comes from one acient node (e..g doing decomposition on some op), they should have the same debug handle for tracing.

winskuo-quic:
> Is the PR ready to be reviewed now?

Hi @Gasoonjia,
I was rebasing previously, so I set it to draft. This PR should now be ready for review. Thanks.

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: I think the idea is that if we decompose the node but never assign a new handle ID, we are just saving the information for the last decomposed node rather than all decomposed node. I have draw an example below. Since edge and QNN has 1 to 1 mapping in this case, I think it would be better to gather all possible information rather than the last node's debug info. Since we reassign graph_handle, instead of only getting the output of **node2**, we can also get info for **node1**.
image

winskuo-quic:
Hi @cccclai, @Gasoonjia, @kimishpatel,
I have supported the debug_handle in this PR, which has been mentioned in https://github.com/pytorch/executorch/pull/5310 and https://github.com/pytorch/executorch/pull/15735.
We will now be using the ExecuTorch's official `Intermediate_Output_Capturer` to capture CPU's intermediate result.

Also, I would also like to get some suggestions on the official API to retrieve an edge IR. The current way of retrieving an edge IR is through: https://github.com/pytorch/executorch/blob/0fb422f9c59e0e5526c0082352a583baf0510fb7/examples/qualcomm/utils.py#L499
However, I encountered following issues when retrieving edge IR using the above method.
1. If there are partitions, I'll be getting a graph that fuses supported nodes to delegate node(s). However, it would be helpful for debugging if we could get the edge IR graph that does not fuse the backend supported nodes to delegate node(s).
2. I noticed that by using the edge IR graph above, the input order might have changed. This can be easily reproduced by using model that has more than 1 input (e.g., Roberta). I would like to know how if there's any way I can get a graph that has the correct input order.

Thanks

backends/qualcomm/_passes/resolve_debug_handle.py
Gasoonjia: here im a little confused: when we see the qnn graph, how can we know that the qnn_node_1 and qnn_node2 here comes from a same super node? Or another q might be, which graph will play as the ground truth graph, when you doing intermediate comparsion?


> gather all possible information rather than the last node's debug info.

We won't gather only the last node debug info, but all info.

In ExecuTorch normally we follow this rule:
if we transform {old_node_1, old_node_2, ..., old_node_n} into {new_node_1, new_node_2, ..., new_node_m}, where n and m can be arbitrary number starting from 1, then: eery new_node should have same debug handle, and the debug handle will be set(old_node_1.debug_handle + old_node_2.debug_handle, ..., old_node_n.debug_handle)

 you can see if n is 1, this transform will be a operator decomposition; if m is 1, this transform will be a operator fusion, etc.
 
 In this way whenever we see an arbitrary new_node, we will know its ancestor.
 
 Not sure if that make sense to you?

Gasoonjia:
hi @winskuo-quic 

I think instead of using the edge graph IR as the ground truth for comparsion, it will be great if we can use the export program ET stack get at the first place (e.g. the export graph of `model` variable https://github.com/pytorch/executorch/blob/0fb422f9c59e0e5526c0082352a583baf0510fb7/examples/qualcomm/utils.py#L480 here), since it should be the source graph ET stack take and our job is making sure our intermediate output the same as the input graph as much as possible.

You can see how we calculate intermediate output numercal descrepancy: https://github.com/pytorch/executorch/blob/0fb422f9c59e0e5526c0082352a583baf0510fb7/devtools/inspector/_inspector.py#L1407

https://github.com/pytorch/executorch/blob/0fb422f9c59e0e5526c0082352a583baf0510fb7/exir/passes/debug_handle_generator_pass.py here's pass for debug handle generation, where the debug handle of a node is the same as the node sharing the same greatest ancestor node in the export flow.

Gasoonjia:
Here's an example of how our current API works on VIT model on xnnpack backend: https://gist.github.com/Gasoonjia/db6285ac39ad5759b95c7a92d37cd4f8

and below is the expected output. For some ops like layernorm there're still some issue i need to fix.


| idx | aot_ops                                           | aot_intermediate_output                                | runtime_ops                                   | runtime_intermediate_output                           | gap                      |
|-----|---------------------------------------------------|--------------------------------------------------------|-----------------------------------------------|--------------------------------------------------------|--------------------------|
| 0   | [conv2d]                                         | [[[tensor(0.0253, -0.0287, -0.0042, 0.0118, …)]]       | [DELEGATE_CALL]                               | [[[tensor(0.0253, -0.0287, -0.0042, 0.0118, …)]]       | 3.2825094945114346e-15   |
| 1   | [permute, cat, add, dropout]                     | [[[tensor(-0.0024), tensor(0.0054), tensor(0.0…), …]] | [DELEGATE_CALL]                               | [[[tensor(-0.0024), tensor(0.0054), tensor(0.0…), …]] | 3.281230918554512e-15    |
| 2   | [expand]                                         | [[[tensor(-0.0012), tensor(0.0027), tensor(0.0…), …]] | [native_call_expand_copy.out]                 | [[[tensor(-0.0012), tensor(0.0027), tensor(0.0…), …]] | 0.0                      |
| 3   | [layer_norm]                                     | [[[tensor(-0.0001), tensor(0.0009), tensor(-0.…), …]] | [native_call_native_layer_norm.out]           | [[[tensor(31.1172)], [tensor(4.3549)], [tensor(…)…]]  | 19.7299543374596         |
| 4   | [transpose, linear, unflatten, …, transpose]     | [[[tensor(0.0027), tensor(-0.0032), tensor(0.0…), …]] | [DELEGATE_CALL, DELEGATE_CALL, DELEGATE_…]    | [[[tensor(0.0027), tensor(-0.0032), tensor(0.0…), …]] | 9.381436078525961e-05    |
| …   | …                                                | …                                                      | …                                             | …                                                      | …                        |
| 61  | [layer_norm_23]                                  | [[[tensor(-0.8604), tensor(-0.1713), tensor(-0.…),…]] | [native_call_native_layer_norm.out]           | [[[tensor(2.2180)], [tensor(1.8462)], [tensor(…)…]]   | 2.8061147356332854       |
| 62  | [linear_46, gelu_11, dropout_35, …, dropout_36]  | [[[tensor(-0.6561), tensor(-0.0496), tensor(-0.…),…]] | [DELEGATE_CALL]                               | [[[tensor(-0.6561), tensor(-0.0496), tensor(-0.…),…]] | 1.0872256686587983e-11   |
| 63  | [layer_norm_24]                                  | [[[tensor(-0.9040), tensor(-0.1004), tensor(-0.…),…]] | [native_call_native_layer_norm.out]           | [[[tensor(1.9138)], [tensor(1.9031)], [tensor(…)…]]   | 3.104443617092582        |
| 64  | [select_36]                                      | [[tensor(-0.9040), tensor(-0.1004), tensor(-0.…),…]]  | [native_call_select_copy.int_out]             | [[tensor(-0.9040), tensor(-0.1004), tensor(-0.…),…]]  | 1.1178469901123618e-12   |
| 65  | [linear_48]                                      | [[tensor(-0.9624), tensor(0.7285), tensor(0.79…),…]]  | [DELEGATE_CALL]                               | [[tensor(-0.9624), tensor(0.7285), tensor(0.79…),…]]  | 1.7864835786911282e-12   |

I would love to chat with you regarding how we can make the pipeline works on qualcomm backend!
Hope it can help you!

winskuo-quic:
Hi @Gasoonjia,
I have turned this PR back to draft for now.
I would love to learn more about ExecuTorch's Debugger Framework and let's move this conversation to email first.
I would also love to chat with you to discuss more in details.

backends/qualcomm/_passes/resolve_debug_handle.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Qualcomm Innovation Center, Inc.
+# All rights reserved
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import operator
+
+import torch
+from executorch.backends.qualcomm.utils.constants import QCOM_DEBUG_HANDLE
```
Gasoonjia: I would love to directly reuse `DEBUG_HANDLE_KEY` from ExecuTorch https://github.com/pytorch/executorch/blob/main/exir/passes/debug_handle_generator_pass.py#L10 to make sure that we are working on same item.

backends/qualcomm/_passes/resolve_debug_handle.py
```diff
@@ -0,0 +1,44 @@
...
+    debug_handle ID must be resolved so each op gets its own set of debug_handle ID and intermediate output.
+    """
+
+    def __init__(self):
+        super(ResolveDebugHandle, self).__init__()
+
+    def call(self, graph_module: torch.fx.GraphModule):
+        handle_counter = 1
+        visited = set()
+        for node in graph_module.graph.nodes:
```
Gasoonjia, line 31: not sure if Qualcomm can handle conditional graph. If so i think the way you are adding debug handle might not able to equip debug handle to all branches. You can follow what i'm doing here: 
https://github.com/pytorch/executorch/blob/main/exir/passes/debug_handle_generator_pass.py#L14

backends/qualcomm/_passes/resolve_debug_handle.py
```diff
@@ -0,0 +1,44 @@
...
+            # Assume node is traversed in topological order, adding a check here to be safe.
+            if node.target == operator.getitem:
+                source_node = node.args[0]
+                assert (
+                    source_node.name in visited
+                ), "Graph is not traversed in topological order, unexpected behavior."
+                node.meta[QCOM_DEBUG_HANDLE] = source_node.meta[QCOM_DEBUG_HANDLE]
```
Gasoonjia: Curious why do we need to set the get_item node the same debug handle as the soruce node? Since it will introduce duplicate debug handle in the graph and im a little bit worried if it could cause any issue in the down stream,

backends/qualcomm/builders/node_visitor.py
```diff
@@ -377,6 +377,11 @@ def get_tensor_name(
         wrapper_idx: int = 0,
     ):
         tensor_name = f"{node.name}_{wrapper_idx}"
+
+        # Only append special namings when enable tensor dump, since longer name results bigger .pte
+        if (handle_id := node.meta.get(QCOM_DEBUG_HANDLE)) and self.enable_tensor_dump:
```
Gasoonjia: wondering if we still need this file since we will migrate to devtool infra?

backends/qualcomm/debugger/qnn_intermediate_debugger.py
```diff
@@ -33,46 +36,61 @@ class OutputFormat(IntEnum):
     DUMP_RAW = 2
 
 
-class IntermediateModule(torch.nn.Module):
-    """
-    This class serves as an intermediate point and is inserted right after the call_function node.
-    It also saves some metadata such as scale, offset, etc.
-    Since we just want to check the intermediate output, we will directly return the value during the forward call.
+class QNNIntermediateDebugger:
```
Gasoonjia, line 39: I think we can change or update the class target; from the comments it plays as the same role of `Inspector.calculate_nunmeric_gap()`

Gasoonjia (commented):
thx for the work!

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: Will set this as resolved for now after offline sync. Please feel free to unresolve the issue if there's more concern. Thanks

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: I am actually just following the same behavior prior to this pass. For node's that has multi-output, it will have `get_item` node as users. I noticed that the `debug_handle` for all `get_item` node and node itself all have the same `debug_handle`, as shown below.
image

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: Thanks for the catch. This would definitely be safer.

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: I would like to confirm on definition of conditional graph. Initially, I thought we can't have conditional graph (e.g., `if else`) in Aten. The only way to do it is through operations such as `where` op, which still makes it a single graph. We actually use a lot of `for loop` for graph traversing in our passes. If you find this being a concern, please let me know and I'll take a look at all our other passes.

backends/qualcomm/builders/node_visitor.py
winskuo-quic: I believe this can be eventually deprecated as we discussed offline. However, since currently we don't have an official API to bring `debug_handle` to runtime, we will need to stick with parsing tensor name during runtime to extract `debug_handle` ID.

backends/qualcomm/debugger/qnn_intermediate_debugger.py
winskuo-quic: I would like to confirm on this part. Are you suggesting that it is better to leave some comments stating that part of this class will eventually be deprecated and be replace by `Inspector.calculate_numeric_gap()`? As synced offline, currently we cannot directly use `calculate_numeric_gap` as QNN requires some output post processing prior to comparison.

backends/qualcomm/_passes/resolve_debug_handle.py
Gasoonjia: conditiion graph refers to some branch like `torch.cond`. https://github.com/pytorch/executorch/blob/main/exir/tests/test_passes.py#L1275 is an example. 
For conditional graph like `torch.cond` we will capture the graph for both branch.

backends/qualcomm/_passes/resolve_debug_handle.py
Gasoonjia: I feel like the reason of get_item and its user sharing the same debug handle is because of exporting in strict mode and both `aten_topk` and its `get_item` function are from same source.
For here i don't have strong preference of whether or not we should make get_item's debug handle the same as source node, as long as it can support Qualcomm's requirement.

backends/qualcomm/debugger/qnn_intermediate_debugger.py
Gasoonjia: thanks for the comment!

My idea is this class should exist, but should only focus on the advance analysis after we fetch the operator-level numerical descrepancy from `calculate_numeric_gap`. 

For essential post-process you mentioned, we can make that happen by either update our current `calculate_numeric_gap` function, or make it as part of custom_comparator.

Please let me know if that works for you

backends/qualcomm/builders/node_visitor.py
Gasoonjia: oh will the tensor name here be part of runtime?

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: Thanks for sharing the operation. I think we currently does not have this op. I think I can add a TODO or some checks for now and actually made the change if QNN supports this operation in future. This will probably require a bigger refactor since all our passes are now using for loops.

backends/qualcomm/builders/node_visitor.py
winskuo-quic: During runtime, we will get tensor name like `relu__debugID_1`, where 1 here is the debug handle. We use regex in runtime to parse the ID `1` and save it to ETDUMP as the key.
I agree we will migrate to devtool infra, but due to some of the limitations where we don't really have an official flow to pass debug_handle ID related info to runtime, we need to keep it this way so the debugger can still function after this PR is merged.

backends/qualcomm/debugger/qnn_intermediate_debugger.py
winskuo-quic: Yes. As discussed last time, since current `calculate_numeric_gap` does not support these post processing, we will use our own comparator to calculate numeric gap for now. Once mainline has post processing introduced, we can then migrate to the official `calculate_numeric_gap` flow.
If we directly use `calculate_numeric_gap` for now, we are comparing CPU (FP32/NCHW) with HTP(Quantized/NHWC), which we will get low accuracy for all operators.

backends/qualcomm/builders/node_visitor.py
Gasoonjia: Thx for detailed explanation! Mind share me the ptr of how you retreive the op name during runtime?

backends/qualcomm/runtime/QnnManager.cpp
```diff
@@ -403,13 +409,36 @@ Error QnnManager::Execute(
...
-          executorch::aten::Tensor>(
-          event_tracer,
-          QNN_TENSOR_VER_PTR(output_tensor)->name,
-          /*delegate_debug_id=*/
-          static_cast<executorch::runtime::DebugHandle>(-1),
-          *dump_tensor);
+      std::string qnn_tensor_name =
+          std::string(QNN_TENSOR_VER_PTR(output_tensor)->name);
+      if (std::regex_search(qnn_tensor_name, match, re)) {
+        debug_handle_id = static_cast<uint32_t>(std::stoul(match[1].str()));
```
winskuo-quic, line 415: Hi @Gasoonjia,
This is where we parse the `qnn_tensor_name` and get the `debug_handle` id.

backends/qualcomm/builders/node_visitor.py
winskuo-quic: Of course! Please refer to the message below.

backends/qualcomm/debugger/qnn_intermediate_debugger.py
Gasoonjia: Yes i totally understand your concern, is there any chance we can work on solving the gap first then directly using the et pipeline in your debugger class? I feel like there are much things in the current `Inspector` class we can reuse to expedite our development.

winskuo-quic: Thanks for the explanation. I would just like to confirm about the current status. Please correct me if I am having the wrong assumption or understanding.
I am currently using our own comparison metrics since HTP has a lot of difference from CPU. This causes us unable to call `calculate_numeric_gap` since it requires some post processing. On top of this, `calculate_numeric_gap` will be using `exported_program` from `torch.export.export` as golden, which we also discussed in our meeting last time that this probably isn't the best golden for Qualcomm, and we prefer to use edge dialect graph after running edge passes.
Are you suggesting that we should keep this PR on hold, and try to have the following issues resolved before merging this PR?
1. Supports post-processing for `calculate_numeric_gap`
2. Able to use edge graph as golden graph.
3. Migrate to use etrecord's edge graph instead of using AOT runtime's edge_manager's edge graph.

Thanks

Gasoonjia:
hi @winskuo-quic any related PR recently need my support or review?

winskuo-quic:
> hi @winskuo-quic any related PR recently need my support or review?

Hi @Gasoonjia,
It would be appreciated if you can have a look at this PR: https://github.com/pytorch/executorch/pull/18834
It is the 2nd part of the PR, which is reusing all dev tools feature. After this PR is approved, I will have a 3rd PR for claude SKILL.

PR #16317, opened 2025-12-18
FlatTensor alignment tests

lucylq:
Summary:
1. Pass segment_alignment from ExecutorchBackendConfig to flat tensor serializer
2. Set segment_alignment=128 as default to match ExecutorchBackendConfig
3. Remove tensor_alignment from the config (do not have multiple tensors per segment)
4. Additional tests for varying segment alignment.

Differential Revision: D89422691


larryliu0820 (approved):

PR #16318, opened 2025-12-18
Arm backend: Add log1p decomposition

robell:
* Decompose log1p and mark as supported in TOSA FP profile
 * Add log1p operator tests with data around the epsilon value
 * Add log1p to table-op LUT mapping for quantized path

Change-Id: I34695a95e76e1291f0ccd52df8b7254260e74240


zingo (approved):

zingo:
@robell  this got merged but have fails in Arm tests, are they know fails and unrelated? If so can you please add a note about it so other that checked for errors know, and if not do we need to revert?

EDIT: no problem fails are know, see next comment

zingo:
Checked the error it is knows fail and are unrelated to the PR and hopefully fixed now a few merged later then this was merged. https://github.com/pytorch/executorch/pull/16335
FAILED backends/arm/test/models/test_T5ForConditionalGeneration_arm.py::test_T5ForConditionalGeneration_tosa_FP
FAILED backends/arm/test/models/test_T5ForConditionalGeneration_arm.py::test_T5ForConditionalGeneration_tosa_INT

PR #16322, opened 2025-12-18
Cortex-M backend: Update Scalar types to be int

SaoirseARM:
### Summary
- Scalar should only used if the operator truly supports either type.


backends/cortex_m/ops/cortex_m_ops_common.h
```diff
@@ -171,8 +167,8 @@ inline bool is_channel_broadcast(const Tensor& tensor1, const Tensor& tensor2) {
 // multiplier: Range {ARM_NN_Q31_MIN + 1, Q32_MAX}
 // shift     : Range {-31, 30}
 inline bool validate_per_channel_quant_params(
-    const int32_t* multipliers,
-    const int32_t* shifts,
+    const IntArrayRef multipliers,
+    const IntArrayRef shifts,
```
psiddh: Overall LGTM, enhances the type safety!
One question though: IntArrayRef = ArrayRef<int64_t> ? 

Different element sizes (4 bytes for int32_t vs 8 bytes for int64_t in IntArrayRef). Can you clarify if this change  (Ln 170/ Ln 171) indeed is safe ?

rascani: Looks like `using IntArrayRef = executorch::aten::ArrayRef<int64_t>;` is up at line 32. Might be worth changing to `Int64ArrayRef` just for clarity (separately).

rascani (approved):

backends/cortex_m/ops/cortex_m_ops_common.h
SaoirseARM: Thanks for review! I have changed this to be Int64ArrayRef to be clearer.

psiddh (approved):

PR #16323, opened 2025-12-18
Update tokenizer pin

mergennachin:
Here are the new commits: 

https://github.com/meta-pytorch/tokenizers/compare/3aada3fe28c945d14d5ec62254eb56ccdf10eb11...37e1c7ed13fa04accd696c776c2f05b4b12fe61d

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -3,6 +3,7 @@
 
 [buildfile]
   name = TARGETS
+  name_v2 = TARGETS,BUCK
 
 [repositories]
   root = .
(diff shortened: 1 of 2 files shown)


lucylq (approved):

.buckconfig
```diff
@@ -3,6 +3,7 @@
 
 [buildfile]
   name = TARGETS
+  name_v2 = TARGETS,BUCK
```
mergennachin, line 6: @bigfootjon 

Is this correct?

One of the git submodules of ET, i.e., tokenizers, now contains BUCK file instead of TARGET or targets.bzl file.

lucylq: It looks good based on [D52947777](https://www.internalfb.com/diff/D52947777), from https://fb.workplace.com/groups/fbcode.fyi/permalink/7176797622355977/

bigfootjon: Yeah that's the correct value

PR #16331, opened 2025-12-19
add support for Qualcomm QCM6490

9rum:
### Summary
This PR adds support for Qualcomm QCM6490, which has been discussed in issue #7356.

### Fixes
Fixes #7356.

### Test plan
The changes in this PR have been tested with the [official Llama example](https://github.com/pytorch/executorch/blob/v0.7.0/examples/qualcomm/oss_scripts/llama/runner/runner.cpp).

### DIFF
--- a/backends/qualcomm/serialization/qc_compiler_spec.fbs
+++ b/backends/qualcomm/serialization/qc_compiler_spec.fbs
@@ -50,6 +50,7 @@ enum QcomChipset: int {
   SAR2230P = 95,
   SA8255 = 52,
   SW6100 = 96,
+  QCS6490 = 35,
 }
 
 /// Indicate the information of the specified SoC.
--- a/backends/qualcomm/serialization/qc_compiler_spec.fbs
+++ b/backends/qualcomm/serialization/qc_compiler_spec.fbs
@@ -50,7 +50,7 @@ enum QcomChipset: int {
   SAR2230P = 95,
   SA8255 = 52,
   SW6100 = 96,
-  QCM6490 = 35,
+  QCM6490 = 93,
 }
 
 /// Indicate the information of the specified SoC.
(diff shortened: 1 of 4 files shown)


nil-is-all:
Hi @shewu-quic @chenweng-quic @winskuo-quic @DannyYuyang-quic @haowhsu-quic if you take a look at this PR fixing [#7356](https://github.com/pytorch/executorch/issues/7356)

cccclai (approved):
Thank you for contributing! I wonder how you run llama on QCM6490? It seems like v68 and doesn't support block wise quantization, which means the accuracy won't be ideal

9rum:
@cccclai Thank you for your kind reply.
We first quantize Llama 3.2 1B using [SmoothQuant](https://proceedings.mlr.press/v202/xiao23c/xiao23c.pdf) then run in 16a8w.
In such settings, the MMLU accuracy was 51.16%.

BTW, I see there is an enum value conflict: `error: duplicate case value: 'QCM6490' and 'SM8350' both equal '35'`.
This error was not found in the previous release where SM8350 was not supported.
However, simply setting SoC number to SM8350 does not cover QCM6490 since their VTCM sizes are different.
Should I remove `QCM6490` and comment it out?

cccclai:
oh, that's a very good point. Feel like we need to fix it cc: @haowhsu-quic @shewu-quic @winskuo-quic @DannyYuyang-quic please allow us a bit longer time to iterate due to holiday season

backends/qualcomm/serialization/qc_compiler_spec.fbs
```diff
@@ -50,6 +50,7 @@ enum QcomChipset: int {
   SAR2230P = 95,
   SA8255 = 52,
   SW6100 = 96,
+  QCM6490 = 35,
```
shewu-quic: the soc_model of QCM6490 is 93

shewu-quic (commented):
Thanks for your contribution.
Please note that the soc_model of QCM6490 is 93
Thanks

shewu-quic (approved):
Thanks a lot :)

9rum:
@shewu-quic Thank you for your correction.
The SoC model number has been fixed.

PR #16334, opened 2025-12-19
Implemented the loader for eIQ Neutron SDK converter.

jirioc:
### Summary
Implemented the loader for eIQ Neutron SDK converter.

### Test plan
Tested using the backends/nxp/tests/test_neutron_converter_manager.py which was also updated.

### DIFF
--- a/backends/nxp/backend/neutron_converter_manager.py
+++ b/backends/nxp/backend/neutron_converter_manager.py
@@ -7,6 +7,7 @@ import importlib
 import logging
 import multiprocessing
 import pkgutil
+import warnings
 
 
 def convert_unsafe(neutron_converter, tflite_model, cctx, queue):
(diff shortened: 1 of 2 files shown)


backends/nxp/backend/neutron_converter_manager.py
```diff
@@ -35,19 +35,27 @@ def __init__(
             module.name
             for module in pkgutil.iter_modules()
             if module.name.startswith("neutron_converter")
+            or module.name == "eiq_neutron_sdk"
         ]
 
-        requested_module_name = f"neutron_converter_{neutron_converter_flavor}"
+        if neutron_converter_flavor:
+            requested_module_name = f"neutron_converter_{neutron_converter_flavor}"
+            print(
```
MartinPavella: Minor:
We usually use `logging.warning()` instead of print.

MartinPavella (commented):
Please update the name of the commit to start with "NXP backend: "

JakeStevens:
Errors seem pre-existing due to some missing Samsung key.

JakeStevens:
Does this mean going forward we don't have to keep bumping the version suffix? That is great, each bump breaks internal tests

JakeStevens (approved):

roman-janik-nxp:
> Does this mean going forward we don't have to keep bumping the version suffix? That is great, each bump breaks internal tests

Yes, we just had a discussion about it with @robert-kalmar. We will abandon the "flavor" feature and will use only eiq_neutron_sdk package. I will prepare it in another PR. This won't be merged.

PR #16339, opened 2025-12-19
Arm backend: Fix incorrect lifetimes in executor runner (#16225)

### LINKED ISSUE #16225: ARM: Major Object Lifetime Error in ExecuTorch ARM Runner Hidden by ARM FVP
### 🐛 Describe the bug

I am working on ExecuTorch integration for a custom SoC based on ARM Cortex cores with an Ethos-U NPU. I was able to run the provided ARM executor runner example on an FPGA simulation, but discovered a critical issue: **several essential ExecuTorch runtime objects have incorrect lifetimes**, causing invalid pointers during model execution.

### Root cause
In `arm_executor_runner.cpp`, the example creates multiple **critical ExecuTorch objects** as local variables inside `runner_init()`, and then calls `execute()` later from a separate function (`run_model()`), _after those objects have already been destroyed_.
```
328: std::vector<Span<uint8_t>> planned_spans;

592: Result<Program> program = Program::load(&loader);

653: HierarchicalAllocator planned_memory(
        {planned_spans.data(), planned_spans.size()});

658: MemoryManager memory_manager(
        &ctx.method_allocator.value(),
        &planned_memory,
        &ctx.temp_allocator.value());
```
### Detailed Breakdown of the Failure
ExecuTorch’s `Method` object stores pointers, not owned objects:

**From** `program.h`
```
296: const executorch_flatbuffer::Program* internal_program_;
```

**From** `method.h`
```
364: const Program* program_;
      MemoryManager* memory_manager_;
      executorch_flatbuffer::ExecutionPlan* serialization_plan_;
      Chain* chains_;
```
But in this example:
- **Program** is destroyed
- the **ExecutionPlan** pointer becomes invalid
- **MemoryManager** becomes invalid
- **Chains becomes** invalid because they depend on the above

This leads to failures when executing instructions.

### Where It Breaks During Execution
Execution begins with chain execution:
**From** `method.cpp`
```
1600: for (...) {
          Chain& chain = chains_[step_state_.chain_idx];
      }
```
But `chains_` was allocated based on `serialization_plan_`:
**From** `method.cpp`
```
924: const auto chains = serialization_plan_->chains();
...
932: auto s_chain = chains->Get(i);
...
1042: chains_[i] = Chain{
              s_chain,
              Span<InstructionArgs>(chain_instruction_arg_lists, num_instructions),
              chain_instruction_kernels,
          };
```
And `serialization_plan_` comes from a pointer passed to:
**From** `method.cpp`
```
846:	Error Method::init(
	    executorch_flatbuffer::ExecutionPlan* s_plan,
	    const NamedDataMap* external_data_map) {
        	...
        	serialization_plan_ = s_plan;
```
`s_plan` is:
**From** `program.cpp`
```
297:	Result<Method> Program::load_method(
	    const char* method_name,
	    MemoryManager* memory_manager,
	    EventTracer* event_tracer,
	    const NamedDataMap* named_data_map) const {

	    ...
	   	auto plan = get_execution_plan(internal_program_, method_name);
		if (!plan.ok()) {
		return plan.error();
		}
		return Method::load(
		  plan.get(), this, memory_manager, event_tracer, named_data_map);
```
We end up with the following pointer chain:

chains_ → serialization_plan_ → s_plan → plan → internal_program_

**This chain would be valid only if all underlying objects were still alive.** As a result, the `program_` pointer inside `Method` now points to invalid memory. **No valid program means no valid execution.**

In practice, execution fails even earlier. Before any instruction is run, the runtime queries method metadata:
**From** `method.cpp`
```
1595:  ET_LOG(Debug, "Executing method: %s.", method_meta().name());
...
1638:  MethodMeta Method::method_meta() const {
		auto name = serialization_plan_->name()->c_str();
		auto method_meta = program_->method_meta(name);
```
`program_` is a dangling pointer → **crash**

### Expected Behavior
The example should store in 'RunnerContext':
- `Program`
- `planned_spans`
- `planned_memory`
- `memory_manager`

A good reference is the Raspberry Pi ExecuTorch [example](https://github.com/pytorch/executorch/blob/cdc2701acb8779dccc2c956f7d2cacd41a183e07/examples/raspberry_pi/pico2/main.cpp#L348), which correctly declares all long-lived objects in `main()` so they remain valid across both `init()` and `execute()` calls.

## Notes About ARM FVP
My concern is that developing against FVP can hide serious lifecycle and memory-safety bugs like this one, because the model does not fully reproduce real hardware behavior or memory access patterns.

In my testing, I used only a small subset of the example’s features (no semihosting, no event tracer, no BundleIO, etc.), so there may be additional issues there as well.

### Versions

ExecuTorch: main (latest commit)

martinlsm:
Keep `BufferDataLoader`, `Program`, `HierarchicalAllocator`, `MemoryManager` and a vector of spans alive for the full run by storing them in `RunnerContext` instead of local variables in `runner_init`. This solves a problem where these objects would go out of scope and therefore destruct when they where still needed, directly or indirectly, by the `Method` object created by `Program::load_method`. This can easily be verified by noting that `Method` holds a pointer to both the `Program` and the `MemoryManager`; thus, these objects must outlive the method.

Fixes #16225


zingo (approved):
Good to merge if/when CI is green

PR #16340, opened 2025-12-19
Use std::isfinite() instead of isfinite() in portable ops.

zingo:
### Summary
This make it possible to build with a simpler libc and saves some flash space. When moving from NEWLIB_LIBC in ZephyrOS to the default one we saved almost 60 Kb in the flash memory (with a simple add model, larger models/code might save less as more libc stuff are probably used).

### Test plan
Test by building ExecuTorch from ZephyrOS with and without NEWLIB_LIBC by hand.

### DIFF
--- a/kernels/portable/cpu/op_allclose.cpp
+++ b/kernels/portable/cpu/op_allclose.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) Meta Platforms, Inc. and affiliates.
  * All rights reserved.
+ * Copyright 2025 Arm Limited and/or its affiliates.
  *
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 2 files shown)


kernels/portable/cpu/op_allclose.cpp
```diff
@@ -42,7 +43,7 @@ bool data_is_close(
     } else {
       auto allowed_error = atol + fabs(rtol * b[i]);
       auto actual_error = fabs(a[i] - b[i]);
```
mergennachin, line 45: std::fabs for consistency?

mergennachin (approved):
Nice, thank you for figuring this out

kernels/portable/cpu/op_allclose.cpp
zingo: I agree, Ill create a PR :)

PR #16344, opened 2025-12-19
Add device type validation in aoti_torch_create_tensor_from_blob_v2

larryliu0820:
Validate that the data pointer location matches the requested device_type.
This prevents silent data corruption when tensors are created with
mismatched data pointer and device type.

Changes:
- Add cudaPointerGetAttributes check to verify data pointer location
- Error out if CPU data is provided but CUDA device is requested
- Error out if CUDA data is provided but CPU device is requested
- Add unit tests for device type mismatch scenarios


larryliu0820:
* #16366
* #16345
* __->__ #16344

backends/cuda/runtime/shims/memory.cpp
```diff
@@ -308,6 +334,11 @@ void clear_all_tensors() {
   ET_LOG(Info, "Cleared all tensors");
 }
 
+void clear_memory_tracking() {
```
Gasoonjia: any reason why we want to have an indivudual `clear_memory_tracking ` function, instead of making it a part of `clear_all_tensors`?

Gasoonjia (approved):
Thanks for adding the support and the tests!

backends/cuda/runtime/shims/memory.cpp
larryliu0820: I can make this change. Soon it will be replaced by slimtensor so I'm not too worried

PR #16345, opened 2025-12-19
Add aoti_torch_item_bool and aoti_torch_assign_tensors_out shims

larryliu0820:
Add two new shim implementations for the CUDA AOTI backend:

1. aoti_torch_item_bool: Extracts a boolean value from a 0D boolean tensor.
   Handles both CPU and CUDA tensors by using cudaPointerGetAttributes to
   determine the memory location and copying from device if needed.

2. aoti_torch_assign_tensors_out: Creates a new tensor view that shares the
   same underlying data as the source tensor. The new tensor has the same
   shape, strides, and dtype as the source.

Also adds:
- Declaration of aoti_torch_dtype_bool() in common_shims.h
- Unit tests for both new functions
- Update CMakeLists.txt with new test targets
- Update targets.bzl with new test targets


larryliu0820:
* #16366
* __->__ #16345

Gasoonjia (approved):
LGTM!

PR #16349, opened 2025-12-19
Add parakeet to examples/models

JacobSzwejbka:
Specifically the parakeet-tdt-0.6b-v3 checkpoint

For the most part this was pretty fast and easy. 90% of the time was just me catching up on how the parakeet model was structured and what the decode loop looks like. An even better solution would be to merge this runner with the one we use for whisper asr. 

Manually verified against no backend targeting (so Its kinda slow).


examples/models/parakeet/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+set(_common_include_directories ${EXECUTORCH_ROOT}/..)
+
+# Need this for gflags
+set(gflags_DIR ${CMAKE_CURRENT_BINARY_DIR}/../../../third-party/gflags)
+find_package(gflags REQUIRED)
+
+# Find executorch libraries
+list(APPEND CMAKE_FIND_ROOT_PATH ${CMAKE_CURRENT_BINARY_DIR}/../../..)
+find_package(executorch CONFIG REQUIRED FIND_ROOT_PATH_BOTH)
+executorch_target_link_options_shared_lib(executorch)
```
larryliu0820, line 27: I don't think you need this

examples/models/parakeet/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+
+set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+
+include(${EXECUTORCH_ROOT}/tools/cmake/Utils.cmake)
+
+if(CMAKE_TOOLCHAIN_FILE MATCHES ".*(iOS|ios\.toolchain)\.cmake$")
+  set(CMAKE_TOOLCHAIN_IOS ON)
+else()
+  set(CMAKE_TOOLCHAIN_IOS OFF)
+endif()
```
larryliu0820: Is this necessary?

examples/models/parakeet/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+# Find executorch libraries
+list(APPEND CMAKE_FIND_ROOT_PATH ${CMAKE_CURRENT_BINARY_DIR}/../../..)
+find_package(executorch CONFIG REQUIRED FIND_ROOT_PATH_BOTH)
+executorch_target_link_options_shared_lib(executorch)
+
+set(link_libraries executorch gflags)
+
+# Common ops for all builds
+list(APPEND link_libraries optimized_native_cpu_ops_lib cpublas eigen_blas)
+executorch_target_link_options_shared_lib(optimized_native_cpu_ops_lib)
```
larryliu0820, line 33: Probably not needed

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -0,0 +1,467 @@
...
+        mel, mel_len = model.preprocessor(
+            input_signal=audio, length=torch.tensor([audio.shape[1]])
+        )
+        encoded, encoded_len = model.encoder(audio_signal=mel, length=mel_len)
+        tokens = greedy_decode_eager(encoded, encoded_len, model)
+        return model.tokenizer.ids_to_text(tokens)
+
+
+def load_model():
+    import nemo.collections.asr as nemo_asr
```
larryliu0820, line 181: For this I think you need to add a install_requirements.txt in the folder so that people can install it

examples/models/parakeet/main.cpp
```diff
@@ -0,0 +1,417 @@
...
+  int64_t time_steps = enc_sizes[2];
+
+  // Create transposed tensor
+  std::vector<float> transposed_data(batch * time_steps * enc_dim);
+  const float* src = encoder_output.const_data_ptr<float>();
+  for (int64_t t = 0; t < time_steps; t++) {
+    for (int64_t d = 0; d < enc_dim; d++) {
+      transposed_data[t * enc_dim + d] = src[d * time_steps + t];
+    }
+  }
```
larryliu0820, line 72: Interesting. Wondering why this can't be part of the model?

larryliu0820 (approved):

examples/models/parakeet/CMakeLists.txt
JacobSzwejbka: Its used in the preprocessor iiuc

examples/models/parakeet/main.cpp
JacobSzwejbka: In abstract it probably /can/ be. I didnt want to do any rewrites though so I only exported functions that were already complete.

examples/models/parakeet/CMakeLists.txt
JacobSzwejbka: I have ET as a dep which part are you saying I dont need?

PR #16378, opened 2025-12-23
Move torch.cond predicate non-persistent buffer to CPU

larryliu0820:
Avoid device-to-host memory copies when evaluating `torch.cond` predicates.

When a GPU buffer (e.g., a KV cache `initialized` flag) is used as a predicate for `torch.cond`, the runtime must synchronize and copy the predicate value from GPU to CPU on every forward pass to evaluate the condition. This adds latency and synchronization overhead.

`MoveCondPredicateToCpuPass` moves non-persistent buffer predicates to CPU at export time, eliminating per-inference D2H transfers. The predicate is typically a small scalar (e.g., a boolean flag), so keeping it on CPU has negligible memory impact.

- Add `MoveCondPredicateToCpuPass` in `backends/cuda/passes/`
- Add unit tests covering:
  - GPU buffer predicates moved to CPU
  - CPU buffer predicates unchanged
  - Computed predicates unaffected
  - Multiple `torch.cond` calls
  - Cross-attention cache pattern
  - Persistent buffers (state_dict) not moved
- Add Python tests to `unittest-cuda` CI job in `cuda.yml`


larryliu0820:
* __->__ #16378

.github/workflows/cuda.yml
```diff
@@ -97,23 +97,26 @@ jobs:
       timeout: 90
       runner: linux.g5.4xlarge.nvidia.gpu
       gpu-arch-type: cuda
-      gpu-arch-version: 12.6
+      gpu-arch-version: 12.9
```
Gasoonjia: why do we need to change gpu test env?

backends/cuda/passes/move_cond_predicate_to_cpu.py
```diff
@@ -0,0 +1,88 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch.export import ExportedProgram
+
+
+class MoveCondPredicateToCpuPass:
+    """
+    A pass that moves the predicate of torch.cond to CPU if the predicate is a constantbuffer.
```
Gasoonjia, line 13: Can you highlight the data transfer time saving here?

backends/cuda/passes/move_cond_predicate_to_cpu.py
```diff
@@ -0,0 +1,88 @@
...
+            return self.out_proj(attn_output)
+    ```
+    In this example if we keep `self.initialized` on GPU, we will need to copy it to CPU for every forward pass.
+    We move the predicate to CPU to avoid device to host copies.
+    This pass is only applicable to models that use torch.cond and its predicate is a constant buffer.
+    """
+
+    requires_exported_program = True
+
+    def __call__(self, exported_program: ExportedProgram):
```
Gasoonjia, line 62: comment deleted

Gasoonjia (approved):
gogogo!

PR #16382, opened 2025-12-24
[slimtensor] Introduce Device and ScalarType headers for SlimTensor minimal support

Gasoonjia:
* #16385
* #16384
* #16383
* __->__ #16382

This diff introduces the foundational c10 core headers for SlimTensor, a lightweight tensor implementation used by torchnative, to cuda backend runtime and further it will be used by all aoti-driven backends like MPS.

We add:

- DeviceType.h - Device type enum (CPU only for now)
- Device.h - Device class representing compute device location
- ScalarType.h - Scalar type enum with elementSize() helper (Float only for now)

These headers are modeled after PyTorch's c10 but simplified for our needs. The enum values are kept compatible with PyTorch for serialization compatibility.

This is the first step in migrating SlimTensor to replace ETensor as the internal tensor representation in CUDA backend. Future diffs will add Storage, SlimTensor class, and additional dtypes/devices incrementally.

Differential Revision: [D89747061](https://our.internmc.facebook.com/intern/diff/D89747061/)


larryliu0820 (approved):

backends/aoti/slim/c10/core/ScalarType.h
```diff
@@ -0,0 +1,83 @@
...
+#include <cstdint>
+#include <ostream>
+
+#include <executorch/runtime/platform/assert.h>
+
+namespace executorch::backends::aoti::slim::c10 {
+
+/// Enum representing the scalar type (dtype) of tensor elements.
+/// Note: Enum values must match PyTorch's c10::ScalarType for compatibility.
+enum class ScalarType : int8_t {
```
manuelcandales: why don't you enable other dtypes here as well? You will need them in order to swap the backend at the end, while maintaining support for currently supported models

PR #16386, opened 2025-12-24
Add CUDA argmax kernel for LLM sampler

larryliu0820:
Add a CUDA kernel for argmax operation to support GPU-based sampling:

- argmax.cuh: Template kernel using warp-level reductions with __shfl_xor_sync
  for efficient parallel max finding. Supports float, half, and bfloat16.
- argmax.cu: Wrapper function argmax_cuda() that launches the kernel,
  handles device-to-host copy, and synchronization.
- test_argmax.cu: Comprehensive unit tests covering various vocab sizes,
  data types, edge cases, and numerical precision.
- CMakeLists.txt: Build configuration for extension_llm_sampler_cuda library
  and GoogleTest-based unit tests.


larryliu0820:
* __->__ #16386

extension/llm/sampler/argmax.cu
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/extension/llm/sampler/argmax.cuh>
+#include <executorch/extension/llm/sampler/cuda_sampler.h>
+#include <executorch/runtime/platform/log.h>
+
+namespace executorch {
```
Gasoonjia, line 12: consider using nested namespace to follow c++ 17 standard

extension/llm/sampler/argmax.cuh
```diff
@@ -0,0 +1,171 @@
...
+    case ::executorch::aten::ScalarType::Half:
+      argmax_vocab_rows_kernel<half><<<grid, block, 0, stream>>>(
+          (const half*)logits, rows, vocab, out_token, out_maxlogit);
+      break;
+    case ::executorch::aten::ScalarType::BFloat16:
+      argmax_vocab_rows_kernel<nv_bfloat16><<<grid, block, 0, stream>>>(
+          (const nv_bfloat16*)logits, rows, vocab, out_token, out_maxlogit);
+      break;
+    default:
+      // Unsupported type, fall back to float
```
Gasoonjia, line 168: perhapes we need to raise error here to avoid silent error?

Gasoonjia (approved):

mergennachin (changes requested):
I don't understand why you had to do a completely custom code.

There's already reduce aten kernel for argmax in pytorch (https://docs.pytorch.org/docs/stable/generated/torch.argmax.html)

https://github.com/pytorch/pytorch/blob/e20b76caff04de2c451777ea55e6fa97075ccbbf/aten/src/ATen/native/cuda/ReduceArgMaxKernel.cu

The torch layer has libtorch dependency (e.g., depends on TensorIterator) so you want to implement argmax as a shim layer instead in ExecuTorch, similar to how int4mm is done. That way, any model that uses torch.argmax can be lowered to ExecuTorch cuda backend easily in the future.

You can add files in 
backends/cuda/runtime/shims/argmax.cu
backends/cuda/runtime/shims/argmax.cuh 
backends/cuda/runtime/shims/argmax.h - header file
update aoti_cuda_shims.lib with the new kernel


Side note: another midsize task we can tackle as a team is to decouple TensorIterator in pytorch/pytorch to be libtorch independent as much as possible -- perhaps part of the header only file.

larryliu0820:
> I don't understand why you had to do a completely custom code.
> 
> There's already reduce aten kernel for argmax in pytorch (https://docs.pytorch.org/docs/stable/generated/torch.argmax.html)
> 
> https://github.com/pytorch/pytorch/blob/e20b76caff04de2c451777ea55e6fa97075ccbbf/aten/src/ATen/native/cuda/ReduceArgMaxKernel.cu
> 
> The torch layer has libtorch dependency (e.g., depends on TensorIterator) so you want to implement argmax as a shim layer instead in ExecuTorch, similar to how int4mm is done. That way, any model that uses torch.argmax can be lowered to ExecuTorch cuda backend easily in the future.
> 
> You can add files in backends/cuda/runtime/shims/argmax.cu backends/cuda/runtime/shims/argmax.cuh backends/cuda/runtime/shims/argmax.h - header file update aoti_cuda_shims.lib with the new kernel
> 
> Side note: another midsize task we can tackle as a team is to decouple TensorIterator in pytorch/pytorch to be libtorch independent as much as possible -- perhaps part of the header only file.

@mergennachin Sampler is outside of AOTI, it's not part of the lowered model. I don't see why we need this in backends/cuda/ though.

larryliu0820:
Talked to @mergennachin We decided to put argmax inside the model for ASR models. We might need cuda sampler for LLM in the future and we can revive this PR.

PR #16387, opened 2025-12-24
Add CudaSampler class for GPU-based token sampling

larryliu0820:
Add CudaSampler class that provides a high-level interface for GPU sampling:

- cuda_sampler.h: Class declaration with sample_argmax() method.
  Pre-allocates GPU memory to avoid allocation in hot path.
- cuda_sampler.cu: Implementation using the default CUDA stream (nullptr)
  for implicit synchronization with the CUDA backend's stream.

The default stream approach ensures proper ordering between decoder
output and argmax without requiring explicit cross-stream synchronization
or access to the backend's internal stream.

### DIFF
--- a/extension/llm/sampler/cuda_sampler.cu
+++ b/extension/llm/sampler/cuda_sampler.cu
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/extension/llm/sampler/cuda_sampler.h>
+#include <executorch/runtime/platform/log.h>
+
+namespace executorch {
+namespace extension {
+namespace llm {
+
+// CudaSampler implementation
+//
+// IMPORTANT: Stream synchronization considerations
+// ------------------------------------------------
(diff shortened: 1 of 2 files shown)


larryliu0820:
* #16388
* __->__ #16387
* #16386

extension/llm/sampler/cuda_sampler.cu
```diff
@@ -0,0 +1,103 @@
...
+//    logits to be fully written before we run argmax on them. Using different
+//    streams without explicit synchronization could cause race conditions.
+//
+// 3. The legacy default stream (stream 0 / nullptr) has special synchronization
+//    semantics: operations on the default stream implicitly synchronize with
+//    operations on other streams in the same CUDA context. This means:
+//    - The argmax kernel will wait for the decoder to finish writing logits
+//    - No explicit cudaDeviceSynchronize() or cross-stream synchronization needed
+//
+// 4. Trade-off: Using the default stream prevents concurrent execution between
```
Gasoonjia, line 37: Im wondering if we really need to make sampler and cuda backend using same cuda stream, since the sampling and decoding should be able to work in parallel: the argmax process of logits_{i} should be able to work with the decoder generating logits_{i+1} since they do not have any dependency, and such parallelism may not happen if argmax and decoder share the same cudastream.

Gasoonjia (approved):

larryliu0820:
obsolete

PR #16388, opened 2025-12-24
Integrate CUDA sampler into ASR runner and enable skip_copy for decoder

larryliu0820:
- cuda_backend.cpp: Support comma-separated method names in
  skip_copy_output_to_cpu_for_method backend option.
- runner.cpp: Use CudaSampler for argmax when CUDA is available and
  temperature==0. Skip copy to CPU for both encoder and decoder methods.
- CMakeLists.txt updates: Link against extension_llm_sampler_cuda library
  and include the sampler subdirectory.

This optimization keeps decoder logits on GPU and performs argmax directly
on GPU memory, avoiding unnecessary device-to-host copies in the decode loop.

### DIFF
--- a/backends/cuda/runtime/cuda_backend.cpp
+++ b/backends/cuda/runtime/cuda_backend.cpp
@@ -83,7 +83,36 @@ class ET_EXPERIMENTAL CudaBackend final
       return false;
     }
     std::lock_guard<std::mutex> guard(skip_copy_method_mutex_);
-    return method_name == skip_copy_method_;
+    // Support comma-separated list of method names
+    if (skip_copy_method_.empty()) {
+      return false;
+    }
+    // Check if method_name matches any entry in the comma-separated list
+    size_t start = 0;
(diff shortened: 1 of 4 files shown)


larryliu0820:
* __->__ #16388
* #16387
* #16386

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -83,7 +83,36 @@ class ET_EXPERIMENTAL CudaBackend final
       return false;
     }
     std::lock_guard<std::mutex> guard(skip_copy_method_mutex_);
-    return method_name == skip_copy_method_;
+    // Support comma-separated list of method names
+    if (skip_copy_method_.empty()) {
```
Gasoonjia, line 87: hmm can we make the `skip_copy_method_` as an array? Maybe we can avoid the fancy comparsion here for supporting multiple method names

Gasoonjia (approved):

larryliu0820:
obsolete

PR #16389, opened 2025-12-24
Better python resolution

TareHimself:
### Summary
used find_package to resolve python

### DIFF
--- a/tools/cmake/Utils.cmake
+++ b/tools/cmake/Utils.cmake
@@ -72,7 +72,10 @@ endfunction()
 function(resolve_python_executable)
   if(NOT PYTHON_EXECUTABLE)
     find_package(Python3 REQUIRED COMPONENTS Interpreter)
-    set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE} PARENT_SCOPE)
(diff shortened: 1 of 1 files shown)


GregoryComer (approved):

GregoryComer:
Thanks! This change makes sense and should make it more robust. Can you update to resolve the one lingering lint issue? I'm happy to merge it once the mandatory lint job is green.

TareHimself:
How do I resolve the lint issue ? not sure how to run the linter locally

TareHimself:
Also, does the CI succeeding mean it works with conda? I expect it to, just want to make sure

GregoryComer:
@TareHimself Sure. For the local lintrunner, here's the setup and run commands: https://github.com/pytorch/executorch/blob/main/CONTRIBUTING.md#lintrunner. Regarding conda, we do use conda for most of the CI jobs, so if CI is passing, it should be good.

GregoryComer:
I pulled down the PR and did a quick sanity check - clean install works without issue on Mac + Conda. I'll go ahead and merge. If we run into any surprises or issues as a result, we should be able to layer them on top of this change. @TareHimself  thanks for the fix!

PR #16403, opened 2025-12-27
[ET-VK][refactor] Introduce PackedDimInfo struct for packed dimension metadata

SS-JIA:
* #16412
* __->__ #16403

## Context

With the introduction of block-packed memory layouts for quantized tensors, the metadata stored by `vTensor` used to describe the data layout within a texture/buffer was no longer sufficient to form a complete description of the data layout. This created an awkward pattern of needing to estimate the `GPUMemoryLayout` which would be needed to compute storage descriptors such as image extents.

This diff addresses the problem by introducing the `PackedDimInfo` struct to `vTensor`, which provides a complete description of how data in a tensor may be organized in the GPU buffer/texture used to store the tensor data, and allows simplification of the functions used to compute buffer numel or texture extents.

## `PackedDimInfo`

Introduced the `PackedDimInfo` struct that encapsulates all information about packed dimensions in GPU tensors. This improves code organization and makes the relationship between related metadata fields explicit.

The `PackedDimInfo` struct contains:
- `packed_dim`: Which dimension is tightly packed (WHCN index) / contiguous in memory
- `packed_dim_padded`: Whether the packed dimension is padded to multiple of 4; some layouts will do this to accomodate vectorized load/stores
- `outer_packed_dim`: Second-level packing for block-packed layouts (4W4C, 4H4W); for layouts with only a single level of packing, will be equal to `packed_dim`
- `outer_packed_dim_padded`: Whether outer packed dim is padded (tiled only)

## Changes

- Added PackedDimInfo struct with helper function calculate_packed_dim_info()
- Replaced packed_dim_ member with packed_dim_info_ in vTensor class
- Updated function signatures to accept PackedDimInfo& instead of packed_dim_:
  * create_hashed_layout
  * calculate_dim_order
  * calculate_padded_sizes
  * calculate_logical_limits
  * TextureMetadata constructor/update
  * vTensorStorage constructor
- Added packed_dim_info() accessor to vTensor and ComputeGraph classes
- Store an additional `padded_sizes_` member in vTensor which is now used to for strides/image extents/GPU buffer numel computation, instead of using `sizes_` directly
- Introduce memory layouts for kInt8x4 type that only use a single level of packing

Differential Revision: [D89832382](https://our.internmc.facebook.com/intern/diff/D89832382/)

### DIFF
--- a/backends/vulkan/runtime/api/containers/Tensor.cpp
+++ b/backends/vulkan/runtime/api/containers/Tensor.cpp
@@ -14,6 +14,38 @@
 namespace vkcompute {
 namespace api {
 
+PackedDimInfo calculate_packed_dim_info(
+    const utils::GPUMemoryLayout memory_layout,
+    const utils::StorageType storage_type) {
+  const int32_t packed_dim = utils::to_packed_dim<int32_t>(memory_layout);
+
+  // Determine if packed dimension is padded
+  const bool packed_dim_padded = storage_type != utils::kBuffer ||
+      memory_layout == utils::kPackedInt8_4W ||
+      memory_layout == utils::kPackedInt8_4C ||
+      memory_layout == utils::kPackedInt8_4H ||
+      memory_layout == utils::kPackedInt8_4W4C ||
+      memory_layout == utils::kPackedInt8_4H4W;
+
+  // Determine outer packed dimension (for tiled layouts)
+  int32_t outer_packed_dim;
+  if (memory_layout == utils::kPackedInt8_4W4C) {
+    outer_packed_dim = 0; // Width
+  } else if (memory_layout == utils::kPackedInt8_4H4W) {
+    outer_packed_dim = 1; // Height
+  } else {
+    outer_packed_dim = packed_dim; // No tiled packing
+  }
+
+  // Determine if outer packed dimension is padded (only for tiled layouts)
+  const bool outer_packed_dim_padded =
+      memory_layout == utils::kPackedInt8_4W4C ||
+      memory_layout == utils::kPackedInt8_4H4W;
+
+  return PackedDimInfo(
+      packed_dim, packed_dim_padded, outer_packed_dim, outer_packed_dim_padded);
+}
+
 /*
  * For PackedInt8 memory layouts, ensure that the scalar type used for the
  * tensor is kInt8x4. Otherwise, return the original scalar type.
@@ -35,24 +67,23 @@ vkapi::ScalarType get_effective_scalar_type(
  */
 std::vector<int64_t> calculate_sizes(
     const vkapi::VulkanImage& image,
-    const utils::GPUMemoryLayout memory_layout) {
+    const PackedDimInfo& packed_dim_info) {
   auto sizes = std::vector<int64_t>{
       image.extents().width, image.extents().height, image.extents().depth};
-  const auto packed_dim = utils::to_packed_dim<int32_t>(memory_layout);
-  sizes.at(packed_dim) *= 4;
+  sizes.at(packed_dim_info.packed_dim) *= 4;
   return sizes;
 }
 
 std::vector<int64_t> calculate_dim_order(
     const size_t ndim,
-    const int32_t packed_dim) {
+    const PackedDimInfo& packed_dim_info) {
   // Special case for zero dim tensors
   if (ndim == 0) {
     return {0};
   }
   std::vector<int64_t> dim_order(ndim);
   // Explicitly convert ndim to signed to prevent underflow
-  int64_t last_dim = int64_t(ndim) - 1 - packed_dim;
+  int64_t last_dim = int64_t(ndim) - 1 - packed_dim_info.packed_dim;
 
   int64_t cur_dim = 0;
   for (int d = 0; d < ndim; ++d) {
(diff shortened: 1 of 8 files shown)


backends/vulkan/runtime/api/containers/Tensor.h
```diff
@@ -21,21 +21,60 @@ namespace api {
...
+
+  PackedDimInfo(
+      int32_t dim,
+      bool dim_padded,
+      int32_t outer_dim,
+      bool outer_dim_padded)
+      : packed_dim(dim),
+        packed_dim_padded(dim_padded),
+        outer_packed_dim(outer_dim),
+        outer_packed_dim_padded(outer_dim_padded) {}
```
mergennachin: don't you want to add validation that outer_packed_dim_padded can be only true for tiled layout?

backends/vulkan/runtime/api/containers/Tensor.h
```diff
@@ -21,21 +21,60 @@ namespace api {
...
+  // Describes a second level of packing, if applicable (which will only apply
+  // to the 4W4C and 4H4W layouts). If there is no second level of packing,
+  // then this will be equal to packed_dim. Otherwise, it will represent the
+  // outer dim used to construct tiled packing. For example, 4W4C will have
+  // packed_dim = 2 and outer_packed_dim = 0.
+  int32_t outer_packed_dim;
+  // Whether the outer packed dim is padded to the next multiple of 4. This is
+  // true only for tiled layouts.
+  bool outer_packed_dim_padded;
+
```
mergennachin, line 52: don't you want another constructor that takes the "legacy" int32_t packed_dim and converts to PackedDimInfo? 

The reason is that current calculate_* functions are semi public(?) and it is BC breaking

Instead of explicitly typing "PackedDimInfo(packed_dim, false, packed_dim, false)" callers can just pass in PackedDimInfo(packed_dim) when they're upgrading the API

backends/vulkan/runtime/api/containers/Tensor.cpp
```diff
@@ -1191,10 +1242,10 @@ void vTensor::virtual_transpose(const int64_t dim0, const int64_t dim1) {
 
   const int dim0_whcn = sizes_.size() - 1 - dim0;
   const int dim1_whcn = sizes_.size() - 1 - dim1;
-  if (packed_dim_ == dim0_whcn) {
-    packed_dim_ = dim1_whcn;
-  } else if (packed_dim_ == dim1_whcn) {
-    packed_dim_ = dim0_whcn;
+  if (packed_dim_info_.packed_dim == dim0_whcn) {
```
mergennachin, line 1298: virtual_transpose() function updates packed_dim when one of the transposed dimensions is the packed dimension, but it doesn't update outer_packed_dim for tiled layouts.

Either disallow transpose for tiled layouts (assert explicitly) or support it.

backends/vulkan/test/vulkan_compute_api_test.cpp
```diff
@@ -294,9 +295,14 @@ TEST_F(VulkanComputeAPITest, calculate_tensor_strides_test) {
          {utils::kWidthPacked, utils::kHeightPacked, utils::kChannelsPacked}) {
       {
         const int32_t packed_dim = static_cast<int32_t>(layout);
+        api::PackedDimInfo packed_dim_info(
```
mergennachin: Expand test coverage for padded and tiled layout scenarios

mergennachin (changes requested):

backends/vulkan/runtime/api/containers/Tensor.cpp
SS-JIA: Yeah that's a great point. Will disallow for tiled layouts!

backends/vulkan/runtime/api/containers/Tensor.h
SS-JIA: the `calculate_*` functions were exposed for the sole purpose of testing, but on second thought it is probably preferable to update the test logic so that these functions don't have to be exposed.

SS-JIA:
Addressed comments (cc: @mergennachin)

* Introduced better enforcement of rules in `PackedDimInfo`
  * Also introduced utility field `is_block_packed` to make code more readable
* Fixed logic of `virtual_transpose()` - virtual transposition is not allowed if either packed dim is the one being transposed
* Introduced comprehensive testing for the metadata of constructed `vTensor` instances in `vulkan_compute_api_test`
* `calculate_*()` functions are no longer public - the only callsites were in testing code anyway

backends/vulkan/runtime/api/containers/Tensor.cpp
```diff
@@ -69,24 +121,33 @@ std::vector<int64_t> calculate_dim_order(
   return dim_order;
 }
 
+/*
+ * Given the sizes of a tensor and the dim order of the tensor (both in NCHW
+ * dimension order), calculate the strides of the tensor.
+ */
 std::vector<int64_t> calculate_strides(
-    const std::vector<int64_t>& sizes,
+    const vkapi::ScalarType dtype,
```
mergennachin: is dtype this used?

backends/vulkan/runtime/api/containers/Tensor.cpp
```diff
@@ -285,73 +389,43 @@ utils::uvec3 calculate_logical_limits(
...
+/*
+ * Calculate the number of elements that a GPU buffer would require to store the
+ * contents of a tensor.
+ */
 int64_t calculate_gpu_buffer_numel(
-    const std::vector<int64_t>& sizes,
-    const utils::GPUMemoryLayout memory_layout,
-    const vkapi::ScalarType dtype) {
+    const vkapi::ScalarType dtype,
+    const PackedDimInfo& packed_dim_info,
```
mergennachin: Is packed_dim_info used at all?

mergennachin (approved):
There are a few unused variables.

What are your thoughts on adding stricter flag (e.g., -Wunused-parameter) in backends/vulkan/CMakeLists.txt?

SS-JIA:
@mergennachin yes, let's do that. Is there a list of additional flags you recommend in addition to `-Wunused-parameter`? I'll put up a follow up PR to add these flags + relevant fixes

SS-JIA:
Changes after accept: small change to `Concat.cpp` so that it doesn't need to call `calculate_padded_sizes`; this allows all `calculate_*()` utility functions to be made private.

Also resolved unused parameters issues.

mergennachin:
> Is there a list of additional flags you recommend in addition to -Wunused-parameter

@SS-JIA What about just -Wall and -Wextra (which will include unused-parameter)? arm, samsung and qualcomm backends have these flags enabled fwiw

https://github.com/pytorch/executorch/blob/main/backends/arm/CMakeLists.txt
https://github.com/pytorch/executorch/blob/main/backends/samsung/CMakeLists.txt
https://github.com/pytorch/executorch/blob/main/backends/qualcomm/CMakeLists.txt

PR #16419, opened 2025-12-30
[ET-VK][ez] Fix NaN propagation in binary div operations for padded texels

SS-JIA:
* #16420
* __->__ #16419

When the packed dimension size is not a multiple of 4, texture-backed
tensors have padding elements in the last texel. For division operations,
these padding regions contain 0/0 = NaN, which propagates through
subsequent reduction operations and corrupts results.

This fix adds conditional padding masking logic to binary_op shaders:
- Introduced MASK_PADDING codegen variable to binary_op.yaml
- Enabled MASK_PADDING=1 for binary_div and binary_floor_divide ops
- Added GLSL preprocessor macro definition in binary_op.glsl
- Implemented padding masking logic using modulo arithmetic to correctly
  identify last texels in batch concatenation scenarios
- Padding elements are explicitly zeroed out to prevent NaN propagation

The implementation follows GLSL best practices by using Python
preprocessing only for macro definition, keeping core shader logic
as pure GLSL with standard #ifdef directives.

Differential Revision: [D89935220](https://our.internmc.facebook.com/intern/diff/D89935220/)

### DIFF
--- a/backends/vulkan/runtime/graph/ops/glsl/binary_op.glsl
+++ b/backends/vulkan/runtime/graph/ops/glsl/binary_op.glsl
@@ -64,6 +64,9 @@ $else:
 #include "broadcasting_utils.h"
 #include "indexing_utils.h"
 
+$if MASK_PADDING:
+  #define MASK_PADDING
+
 layout(local_size_x_id = 0, local_size_y_id = 1, local_size_z_id = 2) in;
 
 ${layout_declare_spec_const(C, "int", "out_layout", "DEFAULT_LAYOUT")}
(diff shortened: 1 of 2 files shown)


backends/vulkan/runtime/graph/ops/glsl/binary_op.glsl
```diff
@@ -140,11 +143,26 @@ void main() {
...
+  const int nspill = mod4(out_sizes[packed_dim]);
+  const int texels_per_batch = divup4(out_sizes[packed_dim]);
+  const bool is_last_texel = (lpos[packed_dim] % texels_per_batch) == (texels_per_batch - 1);
+
+  if (is_last_texel && nspill > 0) {
+    // Explicitly set padding elements to 0 to avoid NaN
+    [[unroll]] for (int i = nspill; i < 4; i++) {
+      out_texel[i] = 0;
+    }
+  }
```
mergennachin, line 165: ```const int texels_per_batch = divup4(out_sizes[packed_dim]); ``` // -> this could be 0, and you're dividing by zero on the next line

i suggest a safer code: 

```
    const int nspill = mod4(out_sizes[packed_dim]);
    if (nspill > 0) {
      const int texels_per_batch = divup4(out_sizes[packed_dim]); // this won't be 0 since we check nspill >0
      const bool is_last_texel = (lpos[packed_dim] % texels_per_batch) == (texels_per_batch - 1);
      if (is_last_texel) {
        [[unroll]] for (int i = nspill; i < 4; i++) {
          out_texel[i] = 0;
        }
      }
    }
```

mergennachin (commented):
Can you add this regression test somewhere? 

something like this: 

```
x = torch.ones(2, 5, 3, 3)
y = torch.ones(2, 5, 3, 3)
return (x / y).sum()
```

We should at least add in vulkan test, but also, I wonder if we can expand it to Backend Test harness test that @GregoryComer built

backends/vulkan/runtime/graph/ops/glsl/binary_op.glsl
SS-JIA: @mergennachin `divup4` divides and rounds up by 4, so it will always be at least 1. But the structure you've recommended makes more sense to me, so will implement it nonetheless 👍

SS-JIA:
Addressed comments cc: @mergennachin 

* Applied suggested code change
* Added test case that can be used to validate the fix

mergennachin (approved):

PR #16436, opened 2026-01-05
Add PointwiseAffineRewritePass for XNNPACK

mergennachin:
Converts pointwise Linear/MatMul to Conv2d(1x1) to eliminate transpose overhead in vision and transformer models. Pattern matches through layout ops (permute, reshape) and replaces entire chains with optimized ops.

How it works

- NCHW patterns (permute→reshape→linear→reshape→permute) → Conv2d(1x1)
- Other patterns (NHWC, transformer) → optimized MatMul with minimal permutes
- Safety-first: rejects ambiguous patterns rather than risk miscompilation

Key features

- Inherits from ExportPass for pipeline compatibility
- Unique channel axis validation (reject if multiple candidates)
- Single consumer path (avoid breaking shared subgraphs)
- Bias vs residual detection (only accept provably bias adds)
- Collision-proof naming for new constants
- Q/DQ extension points for future quantization support

### DIFF
--- a/backends/xnnpack/_passes/pointwise_affine_pass.py
+++ b/backends/xnnpack/_passes/pointwise_affine_pass.py
@@ -1,0 +1,1047 @@
+"""
+Pointwise Affine Rewrite Pass for XNNPACK.
+
+Converts pointwise Linear/MatMul operations to Conv2d(1x1) or optimized MatMul,
+reducing transpose overhead in vision and transformer models.
+
+A pointwise affine operation applies Y = X @ W^T + b independently at each spatial
+position. The channel dimension is the contraction dimension for the matrix multiply,
+while spatial dimensions are preserved (each spatial location is processed identically).
+
+Algorithm Overview
+------------------
+MATCHING (PointwiseAffineMatcher):
+1. Find Linear/MatMul ops with 2D input [flat_batch, cin]
+2. Validate the flatten preserves channel axis:
+   - Producer must have exactly ONE axis of size cin
+   - That axis must end up at the last position (via permute or already there)
+   - Product of other dims must equal flat_batch
+3. Trace BACKWARD through layout ops to find N-D origin tensor
+4. Find unique channel axis in origin shape (reject if ambiguous)
+5. Trace FORWARD through layout ops to find N-D output tensor
+   - Require single-user path (avoid breaking shared subgraphs)
+   - Accept bias adds only if provably bias (parameter, not activation)
+6. Verify output shape matches expected [origin_shape with cout replacing cin]
+
+LOWERING (PointwiseAffineLowering):
+- NCHW (rank 4, channel axis 1): Replace entire pattern with Conv2d(1x1)
+- Other patterns: Replace with permute -> reshape -> mm -> reshape -> permute
+
+Safety measures to avoid false positives:
+1. Unique channel axis - reject if multiple axes could be the channel
+2. Single consumer path - require linear has exactly one user
+3. Explicit op allowlist - no substring matching
+4. Channel axis validation - verify flatten preserves channel intact
+5. Bias validation - only accept provably bias adds, not residual connections
+6. Concrete shapes only - reject symbolic dimensions
+
+Supported patterns:
+- NCHW (rank 4, channel axis 1) -> Conv2d(1x1)
+- NHWC/Transformer (channel last) -> MatMul
+"""
+
+import copy
+import logging
+from dataclasses import dataclass
+from typing import Callable, Dict, List, Optional, Set, Tuple
+
+import torch
+import torch.fx as fx
+from executorch.backends.xnnpack._passes.xnnpack_pass import XNNPACKPass
+from executorch.backends.xnnpack.utils.utils import get_param_tensor, is_param_node
+from executorch.exir.backend.utils import is_shape_dynamic
+from executorch.exir.dialects._ops import ops as exir_ops
+from torch.export.graph_signature import InputKind
+from torch.fx.passes.infra.pass_base import PassResult
+
+logger = logging.getLogger(__name__)
+
+
+# Op allowlists - explicit sets, no substring matching
+LAYOUT_OPS = frozenset(
+    [
+        torch.ops.aten.permute.default,
+        torch.ops.aten.permute_copy.default,
+        torch.ops.aten.transpose.int,
+        torch.ops.aten.view.default,
+        torch.ops.aten.view_copy.default,
+        torch.ops.aten.reshape.default,
+        torch.ops.aten._unsafe_view.default,
+        torch.ops.aten.flatten.using_ints,
+        torch.ops.aten.unflatten.int,
+        torch.ops.aten.squeeze.dim,
+        torch.ops.aten.squeeze.dims,
+        torch.ops.aten.unsqueeze.default,
+        torch.ops.aten.contiguous.default,
+        torch.ops.aten.clone.default,
+    ]
+)
+LINEAR_OPS = frozenset(
+    [
+        torch.ops.aten.linear.default,
+        torch.ops.aten.mm.default,
+        torch.ops.aten.addmm.default,
+        torch.ops.aten.matmul.default,
+    ]
+)
+RESHAPE_OPS = frozenset(
+    [
+        torch.ops.aten.view.default,
+        torch.ops.aten.view_copy.default,
+        torch.ops.aten.reshape.default,
+        torch.ops.aten._unsafe_view.default,
+    ]
+)
+PERMUTE_OPS = frozenset(
+    [torch.ops.aten.permute.default, torch.ops.aten.permute_copy.default]
+)
+TRANSPARENT_OPS = frozenset(
+    [torch.ops.aten.contiguous.default, torch.ops.aten.clone.default]
+)
+
+# Activation ops that can appear between linear and layout ops.
+# We trace through these and recreate them after the rewritten conv2d/mm.
+ACTIVATION_OPS = frozenset(
+    [
+        torch.ops.aten.relu.default,
+        torch.ops.aten.gelu.default,
+        torch.ops.aten.silu.default,
+        torch.ops.aten.hardswish.default,
+        torch.ops.aten.hardsigmoid.default,
+    ]
+)
+
+# Maximum depth for tracing through layout ops (permute, reshape, view, etc.)
+# This limits loop iterations when walking the graph to find patterns.
+# 20 is generous - real patterns typically have 3-6 layout ops.
+_MAX_TRACE_DEPTH = 20
+
+
+def _underlying(target):
+    """Get underlying ATen op for edge dialect ops."""
+    return getattr(target, "_op", target)
+
+
+def _op_in(target, op_set: frozenset) -> bool:
+    """Check if target or its underlying op is in the set."""
+    return target in op_set or _underlying(target) in op_set
+
+
+def _shape(node: fx.Node) -> Optional[Tuple[int, ...]]:
+    """Get concrete shape from node, None if symbolic."""
+    val = node.meta.get("val")
+    if val is None or not hasattr(val, "shape"):
+        return None
+    # Use is_shape_dynamic to efficiently check for symbolic dimensions
+    if is_shape_dynamic(node):
+        return None
+    return tuple(int(s) for s in val.shape)
+
+
+def _dtype(node: fx.Node) -> torch.dtype:
+    """Get dtype from node. Raises if node lacks valid tensor metadata."""
+    val = node.meta.get("val")
+    if val is None or not hasattr(val, "dtype"):
+        raise ValueError(
+            f"Node {node.name} lacks valid tensor metadata; cannot extract dtype"
+        )
+    return val.dtype
+
+
+def _copy_meta(meta: Dict, new_val=None) -> Dict:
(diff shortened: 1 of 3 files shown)


backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+                return None, set()
+        # Depth limit reached - log for debugging
+        logger.debug(
+            "PointwiseAffineRewritePass: _trace_back hit depth limit %d at node %s",
+            _MAX_TRACE_DEPTH,
+            start.name,
+        )
+        return None, set()
+
+    def _trace_forward(self, start: fx.Node, target_rank: int, cout: int):
```
GregoryComer, line 534: Nit: Is it possible to use SubgraphMatcher and then do validation on the result to avoid writing custom matching logic?

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+                    return tensor, b
+        elif _op_in(t, {torch.ops.aten.addmm.default}) and node.args:
+            b = node.args[0]
+            if isinstance(b, fx.Node):
+                tensor = self._get_param(b)
+                if tensor is not None and tensor.shape == (cout,):
+                    return tensor, b
+        return None, None
+
+    def _extract_add_bias(self, add_node: fx.Node, pred: fx.Node, cout: int):
```
GregoryComer, line 683: It doesn't hurt to have this code, but is this a pattern that we actually see in models? Having an explicit bias add node instead of using linear/addmm?

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+    """
+    Copy node metadata, optionally overriding 'val'.
+    """
+    result = copy.copy(meta)
+    if new_val is not None:
+        result["val"] = new_val
+    return result
+
+
+def _is_dq_node(node: fx.Node) -> bool:
```
GregoryComer: Nit: Duplicates existing `is_dequant` function in backends/xnnpack/utils/quant_utils.py.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+    """Check if node is a dequantize operation."""
+    return node.op == "call_function" and _op_in(node.target, DQ_OPS)
+
+
+def _get_dq_source(node: fx.Node) -> Optional[fx.Node]:
```
GregoryComer: This function seems a little odd. It's only handling dq nodes that take a direct graph input (placeholder node)? The quantized source, is, in general, the output of some op node and not a placeholder.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+        a0, a1 = add_node.args[:2]
+        bias_arg = a1 if a0 is pred else (a0 if a1 is pred else None)
+        if not isinstance(bias_arg, fx.Node):
+            return None, None
+        t = self._get_param(bias_arg)
+        if t is not None and self._valid_bias_shape(t.shape, cout):
+            return t.reshape(cout), bias_arg
+        return None, None
+
+    def _get_param(self, node: fx.Node) -> Optional[torch.Tensor]:
```
GregoryComer, line 701: Nit: Duplicates existing get_param_tensor function in backends/xnnpack/utils/utils.py.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+            m.output.replace_all_uses_with(new_out)
+
+            # Erase unused nodes in reverse topological order (users before inputs).
+            # Sort match nodes by graph position, then reverse to get proper erase order.
+            nodes_to_erase = [
+                node
+                for node in graph_module.graph.nodes
+                if node in m.nodes and node != m.origin
+            ]
+            for node in reversed(nodes_to_erase):
```
GregoryComer, line 1032: Can you add a test to validate that the original weights aren't present in the new delegate blob?

backends/xnnpack/test/passes/test_pointwise_affine.py
```diff
@@ -0,0 +1,585 @@
...
+
+        model = NHWCPointwiseLinear(8, 16).eval()
+        inputs = (torch.randn(2, 4, 4, 8),)
+
+        (
+            Tester(model, inputs)
+            .export()
+            .to_edge()
+            .run_passes(self.PassStage)
+            .check_node_count({MM_OP: 1})  # mm exists - pass worked!
```
GregoryComer, line 136: Can you add an E2E test to verify that the linear/mm is delegated post-transform for this case? Also, what is the intent of this re-write?

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+    try:
+        return tuple(int(s) for s in val.shape)
+    except (TypeError, ValueError):
+        return None
+
+
+def _dtype(node: fx.Node) -> torch.dtype:
+    """Get dtype from node."""
+    val = node.meta.get("val")
+    return val.dtype if val is not None and hasattr(val, "dtype") else torch.float32
```
GregoryComer: Nit: This logic seems suspect - it defaults to float32? If it's actually a valid tensor, it should always have a dtype. Falling back to float32 silently seems error-prone.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+    return target in op_set or _underlying(target) in op_set
+
+
+def _shape(node: fx.Node) -> Optional[Tuple[int, ...]]:
+    """Get concrete shape from node, None if symbolic."""
+    val = node.meta.get("val")
+    if val is None or not hasattr(val, "shape"):
+        return None
+    try:
+        return tuple(int(s) for s in val.shape)
```
GregoryComer: Nit: We can use is_shape_dynamic in exir/backend/utils.py to check for symbolic dimensions, instead of relying on it throwing on dynamic dims, which is slow.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+    def _meta_tensor(self, shape: List[int], dtype: torch.dtype) -> torch.Tensor:
+        """
+        Create a tensor for metadata purposes only.
+
+        Uses torch.empty to avoid unnecessary memory initialization since only
+        shape and dtype are used, not the tensor values.
+        """
+        return torch.empty(shape, dtype=dtype)
+
+    def _maybe_rewrap_qdq(self, node: fx.Node, reference: fx.Node) -> fx.Node:
```
GregoryComer: Can we either fully implement handling for quantized subgraphs (w/ tests) or clean up this logic? I'm hesitant to include a half-implementation.

GregoryComer: Additionally, it would be good to at a minimum, add some tests to verify that the pass doesn't break the lowering when it receives a quantized graph.

backends/xnnpack/_passes/pointwise_affine_pass.py
```diff
@@ -0,0 +1,1040 @@
...
+reducing transpose overhead in vision and transformer models.
+
+A pointwise affine operation applies Y = W @ X + b independently at each spatial
+position, where only the channel dimension participates in the matrix multiply.
+
+Algorithm Overview
+------------------
+MATCHING (PointwiseAffineMatcher):
+1. Find Linear/MatMul ops with 2D input [flat_batch, cin]
+2. Validate the flatten preserves channel axis:
```
GregoryComer, line 15: I'm not completely sold on this being sound in all cases by just looking at the input and output shapes. Consider the following case:

x: [1, 2, 3, 4]
y: [12, 2] = x.reshape(12, 2)
y: [12, 4] = linear(x)
y: [1, 4, 3, 4] = x.reshappe(1, 4, 3, 4)

If I'm understanding correctly, the logic is just looking at shape before and after any reshape/permute transformations and doesn't consider the actual op semantics. At a minimum, can we add tests for the above case?

In general, I think it would be safer to pattern-match on specific, concrete patterns than try to handle arbitrary view and permute ops by only looking at the input and output shapes. That would also allow re-using existing fx graph APIs to do subgraph matching instead of using a custom matching algorithm here.

If needed, we can canonicalize the view and reshape-stype ops down to just view and permute.

GregoryComer (changes requested):
Thanks for putting this up. I have a few concerns with the soundness of the matching logic, as it currently stands. I'm good to merge it if we can add tests to verify the cases called out in the comments.

As a nit, there are a lot of duplicated utility functions in this pass. It would be good to consolidate on the existing utils, where possible. I've left comments on most of them to mention the existing util.

Do you have performance numbers for a motivating case? I believe that this should be a net benefit, mostly just curious on the magnitude. In the long-term, zero cost views + transpose optimizations should give us this "for free" without this pass, but I'm happy to merge this for now.

GregoryComer:
Also, are we confident that the LLM-related failures are unrelated to these changes? I see some are non-CPU, but there are some CPU failures. The Phi3 job, for example, shows wrong outputs. It could be a flaky job, but it would be good to confirm this.

backends/xnnpack/_passes/pointwise_affine_pass.py
mergennachin: SubgraphMatcher is unfortunately not suitable. Variable-length chains can't be expressed as fixed subgraphs. 

We need to enumerate 

- origin -> reshape -> linear -> reshape -> output
- origin -> permute -> reshape -> linear -> reshape -> permute -> output
- origin -> flatten -> linear -> unflatten -> output

...

backends/xnnpack/_passes/pointwise_affine_pass.py
mergennachin: The pass doesn't just look at input/output shapes. It validates that the channel dim is preserved correctly through the layout ops. Specifically, _valid_reshape requires that:
  - Either the channel axis is already at the last position in the producer's shape
  - OR the producer is a permute operation that moves the channel axis to last

I added a test that to demonstrate that it rejects the example pattern above. **test_channel_axis_not_at_last_position_should_not_match**

backends/xnnpack/_passes/pointwise_affine_pass.py
mergennachin: I will remove quantization logic for now.

backends/xnnpack/_passes/pointwise_affine_pass.py
mergennachin: I remove the quantization code, and also added a test that quantized graph lowering is not broken (the pass doesn't recognize).

mergennachin:
Addressing @GregoryComer 's feedbacks 

1. Re-using utility functions 

2. Removed Quantization-Related Code

- Removed is_dequant import from quant_utils
- Removed _get_dq_source function (unused, had incorrect placeholder-only logic)
- Removed _maybe_rewrap_qdq method (was a no-op with TODO comment)
- Simplified _create_node to return node directly without wrapper call

3. Added Test for Reviewer's Example (lines 570-609)

- Added test_channel_axis_not_at_last_position_should_not_match in TestMatcherNegative
- Tests the specific pattern: [1, 2, 3, 4] -> reshape(12, 2) -> linear -> reshape(1, 4, 3, 4)
- Verifies the pass correctly rejects patterns where channel axis is not at last position without a permute

4. Added Quantization Compatibility Tests (lines 898-1048)

- New TestQuantizationCompatibility class with 4 tests:
  - test_quantized_nchw_linear_not_matched - Explicitly verifies pass does NOT create conv2d/mm for quantized graphs
  - test_quantized_nhwc_linear_not_matched - Same for NHWC pattern
  - test_quantized_nchw_linear_full_pipeline - Verifies full pipeline still works with quantization
  - test_quantized_linear_relu_full_pipeline - Verifies linear+relu with quantization

5. Fixed _dtype Function (lines 141-148)

- Changed from silently defaulting to float32 to raising ValueError if node lacks valid tensor metadata
- Prevents masking bugs from silent fallback

6. Added NHWC E2E Delegation Test (lines 732-782)

- Added test_nhwc_delegates_to_xnnpack in TestXNNPACKIntegration
- Verifies linear -> mm rewrite and XNNPACK delegation for NHWC patterns
- Includes detailed docstring explaining the intent of the NHWC rewrite

7. Added Weight Cleanup Verification Test (lines 784-850)

- Added test_original_weights_not_in_transformed_graph in TestXNNPACKIntegration
- Verifies original linear weight/bias placeholders are removed after transformation
- Confirms new conv weight placeholders are present instead

backends/xnnpack/_passes/pointwise_affine_pass.py
GregoryComer: Sounds good. I still think there may be cases that aren't handled here, but I'm okay to wait and see / take it as a follow-up as long as CI is green. Thanks.

GregoryComer (approved):

PR #16456, opened 2026-01-05
Remove .forward for call to attention block

kimishpatel:
Summary: We have been explicitly calling forward method on attention block from transformer block. This prevents use from registering various hooks on the attention module. I found this out while trying to implement hook that records the input to the attention block.

Differential Revision: D90126170


JacobSzwejbka (approved):

JacobSzwejbka:
can you check that we dont do this with anything else like the kvcache

kimishpatel:
> can you check that we dont do this with anything else like the kvcache

Will do. I dont know why we did this with attention.

PR #16458, opened 2026-01-05
Add eps parameter to QAT quantization config functions

psiddh:
Summary:
D90093073 introduced an `eps` parameter to PTQ quantization config functions but missed adding it to the QAT equivalents (`get_8a8w_qnn_qat_config`, `get_16a4w_qnn_qat_config`, `get_qat_per_channel_quant_config`). This caused `test_qnn_backend_linear_qat` to fail because `ModuleQConfig.__post_init__` now passes `eps=self.eps` to all config functions, but the QAT functions didn't accept this parameter.

This change adds the `eps` parameter to these QAT config functions and properly applies it to all observers (activation, weight, and bias) to maintain consistency with the PTQ functions.

Reviewed By: SS-JIA

Differential Revision: D90123723

cc @cccclai @winskuo-quic @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

### DIFF
--- a/backends/qualcomm/quantizer/qconfig.py
+++ b/backends/qualcomm/quantizer/qconfig.py
@@ -287,64 +287,6 @@ def get_16a8w_qnn_ptq_config(
     return quantization_config
 
 
-def get_16a8w_qnn_qat_config(
-    act_observer=MovingAverageMinMaxObserver, eps: float = None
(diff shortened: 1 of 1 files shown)


SS-JIA (approved):

PR #16461, opened 2026-01-05
Fix deprecated importlib.resources and exir.capture() API usage

mergennachin:
- Replace deprecated `read_binary()` with `files().joinpath().read_bytes()`
  in xnnpack_graph_serialize.py
- Replace deprecated `is_resource()` with `files().joinpath().is_file()`
  in _flatbuffer.py
- Replace deprecated `exir.capture()` with `torch.export.export()` + `to_edge()`
  in bilinear_2d.py

### DIFF
--- a/backends/xnnpack/partition/graphs/bilinear_2d.py
+++ b/backends/xnnpack/partition/graphs/bilinear_2d.py
@@ -7,10 +7,10 @@
 from functools import lru_cache
 from typing import Dict, List
 
-import executorch.exir as exir
 import torch
 
 from executorch.backends.xnnpack.utils.configs import get_xnnpack_edge_compile_config
+from executorch.exir import to_edge
 
 
 @lru_cache(maxsize=None)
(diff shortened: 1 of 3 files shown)


backends/xnnpack/partition/graphs/bilinear_2d.py
```diff
@@ -31,21 +31,19 @@ def forward(self, x):
...
-        for config in capture_configs:
-            for skip_dim_order_flag in [True, False]:
-                edge = exir.capture(
-                    bilinear2d(align_corners), sample_inputs, config
-                ).to_edge(
-                    config=get_xnnpack_edge_compile_config(
-                        skip_dim_order=skip_dim_order_flag
-                    )
-                )
-                _bilinear2d_graphs[edge.exported_program.graph_module] = align_corners
```
mergennachin, line 48: @tugsbayasgalan, @angelayi  - what's the correct one?

backends/xnnpack/partition/graphs/bilinear_2d.py
```diff
@@ -31,21 +31,19 @@ def forward(self, x):
 
     sample_inputs = (torch.randn(1, 3, 4, 4),)
     _bilinear2d_graphs = {}
-    capture_configs = [
```
GregoryComer, line 34: FYI I actually intend to remove this file as it's not currently used. I landed upstream changes in pytorch core a while back to prevent decomposition of bilinear2d, as it's a core aten op. I have an old PR to do this - https://github.com/pytorch/executorch/pull/8910 - I'll see if I can get it landed. Feel free to land these changes, though, if you get to it first.

JacobSzwejbka (approved):

backends/xnnpack/partition/graphs/bilinear_2d.py
JacobSzwejbka: enable_aot=True, _unlift=False is the one that should dispatch to torch.export under the hood.

PR #16489, opened 2026-01-07
NXP backend: Add edge dialect pass to remove useless `as_strided_copy` nodes.

MartinPavella:
### Summary
This PR integrates the `RemoveUselessOpPass` edge dialect pass from the Samsung backend into the NXP backend. This feature is required to support an internal model.

### Test plan
Tested internally by NXP on a private model.

cc @robert-kalmar

### DIFF
--- a/backends/nxp/edge_passes/neutron_edge_pass_manager.py
+++ b/backends/nxp/edge_passes/neutron_edge_pass_manager.py
@@ -1,4 +1,4 @@
-# Copyright 2025 NXP
+# Copyright 2025-2026 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -8,6 +8,9 @@ from executorch.backends.nxp.edge_passes.move_auxiliary_operator_into_separate_q
     MoveTrailingAuxiliaryOperatorIntoSeparateQDQClusterPass,
 )
 from executorch.backends.nxp.edge_passes.neutron_edge_pass import NeutronEdgePass
+from executorch.backends.nxp.edge_passes.remove_as_strided_copy_nodes import (
+    RemoveUselessAsStridedCopyNodes,
+)
 from torch.fx.passes.infra.pass_manager import PassManager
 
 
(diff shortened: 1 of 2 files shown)


backends/nxp/edge_passes/neutron_edge_pass_manager.py
```diff
@@ -17,6 +18,7 @@ def __init__(self, passes: list[NeutronEdgePass] = None):
         passes: list[NeutronEdgePass] = passes or [
             MoveLeadingAuxiliaryOperatorIntoSeparateQDQClusterPass(),
             MoveTrailingAuxiliaryOperatorIntoSeparateQDQClusterPass(),
+            RemoveUselessOpPass(),
```
robert-kalmar: @MartinPavella , you mentioned the `RemoveUselessOpPass` handled specifically the case of `mean` followed by the `as_strided_copy`, but seems it handles more "useless" op cases - https://github.com/pytorch/executorch/blob/main/backends/samsung/_passes/remove_useless_ops.py#L15

robert-kalmar:
@Jiseong-oh, we also need a pass to remove the `as_strided_copy` preceeded by the `mean`, what handled by the `executorch.backends.samsung._passes.remove_useless_ops.RemoveUselessOpPass`. Do you mind to reuse it in NXP backend?

robert-kalmar (approved):

backends/nxp/edge_passes/neutron_edge_pass_manager.py
MartinPavella: Now that you mention it, I thought about it a bit more. For our use case we actually mustn't remove the other "useless ops". Since we need to run this pass before partitioning, some of the ops (like `dim_order_ops._to_dim_order_copy`) are not useless at all, and are actually vital in case they don't get delegated.

Therefore, I will create our own pass based on this one, to only remove the `as_strided_copy` for now.

robert-kalmar:
> @Jiseong-oh, we also need a pass to remove the `as_strided_copy` preceeded by the `mean`, what handled by the `executorch.backends.samsung._passes.remove_useless_ops.RemoveUselessOpPass`. Do you mind to reuse it in NXP backend?

@Jiseong-oh , based on the feedback from Martin, the question is not anymore relevant.

backends/nxp/edge_passes/neutron_edge_pass_manager.py
MartinPavella: Done

robert-kalmar (approved):
Code looks fine to me. 
Let the CI's finish.

PR #16490, opened 2026-01-07
NXP backend: added aten.split support 

novak-vaclav:
### Summary

adds support for aten.split operator

### Test plan

cc @robert-kalmar @JakeStevens @digantdesai @MartinPavella

### DIFF
--- a/backends/nxp/aten_passes/neutron_aten_pass_manager.py
+++ b/backends/nxp/aten_passes/neutron_aten_pass_manager.py
@@ -1,4 +1,4 @@
-# Copyright 2026 NXP
+# Copyright 2025-2026 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -9,6 +9,9 @@ import torch
 
 from executorch.backends.nxp.aten_passes.convert_unsqueeze_to_view import (
     ConvertUnsqueezeToViewPass,
+)
+from executorch.backends.nxp.aten_passes.decompose_split_to_slices_pass import (
+    DecomposeSplitToSlicesPass,
 )
 from executorch.backends.nxp.aten_passes.fuse_batch_norm_with_conv_pass import (
     FuseBatchNormWithConvPass,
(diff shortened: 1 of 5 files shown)


novak-vaclav:
Fixed all issues mentioned in this PR (https://github.com/pytorch/executorch/pull/16276), including the comment of @roman-janik-nxp 
But due to my error I had to create a new PR.

backends/nxp/tests/test_split_group_convolution.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 NXP
+# Copyright 2026 NXP
```
MartinPavella: When updating an old file, please keep the original copyright year as well. In this case, please use:
`# Copyright 2025-2026 NXP`

backends/nxp/tests/test_split_group_convolution.py
```diff
@@ -246,7 +246,15 @@ def test_split_group_convolution__applied_by_default(self, _, is_qat: bool):
         graph_module = torch.export.export(module, example_input).module()
         original_module = deepcopy(graph_module)
 
-        modified_module = NeutronAtenPassManager(neutron_target_spec)(
+        neutron_pass_manager = NeutronAtenPassManager(neutron_target_spec)
+        # Provisional solution to manually remove DecomposeSplitToSlicesPass until NeutronAtenPassManager is fixed
```
MartinPavella: This is not true.
Regardless of how the NeutronAtenPassManager  works, the pass is removed here not because of any bugs, but only because it would (correctly) modify the graph, in a way that would require changes to this test, and it would reduce clarity.

**So the pass is removed only for testing purposes.**

Even when we update the NeutronAtenPassManager  as we discussed, this test will still have the pass removed.

backends/nxp/tests/test_split_group_convolution.py
```diff
@@ -246,7 +246,15 @@ def test_split_group_convolution__applied_by_default(self, _, is_qat: bool):
...
+        neutron_pass_manager.passes = [
+            a_pass
+            for a_pass in neutron_pass_manager.passes
+            if a_pass.__name__ != "DecomposeSplitToSlicesPass"
+        ]
```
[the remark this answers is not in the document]
MartinPavella: Not applicable

backends/nxp/aten_passes/neutron_aten_pass_manager.py
MartinPavella: `# Copyright 2025-2026 NXP`

MartinPavella (changes requested):

MartinPavella (approved):

PR #16496, opened 2026-01-07
llm release with logging

JacobSzwejbka:


CMakePresets.json
```diff
@@ -365,6 +385,20 @@
         }
       ]
     },
+    {
+      "name": "llm-release-logging",
```
Gasoonjia, line 435: what's the difference between this and the llm-release-logging above?

CMakePresets.json
```diff
@@ -145,6 +145,17 @@
         "CMAKE_INSTALL_PREFIX": "${sourceDir}/cmake-out"
       }
     },
+    {
+      "name": "llm-release-logging",
```
Gasoonjia, line 149: can we just concentrate all things inside debug mode? Now we have too much different modes

CMakePresets.json
JacobSzwejbka: The other stuff lets you do 

cmake --workflow --preset llm-release-logging

vs

cmake --preset llm-release-logging
cmake --build --preset llm-release-logging-install

CMakePresets.json
JacobSzwejbka: Right now we have logging coupled with optimizations (debug mode). Since we are for some reason using ET logs to report useful step information to the user in the examples binaries it would be useful to have these be independent decisions. Debug mode is at least twice as slow as release for cpu portable ops

manuelcandales (approved):

CMakePresets.json
```diff
@@ -193,7 +228,9 @@
         "llm-debug"
       ],
       "cacheVariables": {
-        "EXECUTORCH_BUILD_CUDA": "ON"
+        "EXECUTORCH_BUILD_CUDA": "ON",
+        "EXECUTORCH_ENABLE_LOGGING": "ON",
+        "ET_MIN_LOG_LEVEL": "Debug"
```
manuelcandales, line 233: does this actually work? did you try it and it printed Debug logs?
in my experience, I have to pass `"EXECUTORCH_LOG_LEVEL": "Debug"` to make it work

CMakePresets.json
```diff
@@ -145,6 +145,17 @@
...
     },
+    {
+      "name": "llm-release-logging",
+      "displayName": "LLM release build with logging enabled",
+      "inherits": [
+        "llm-release"
+      ],
+      "cacheVariables": {
+        "EXECUTORCH_ENABLE_LOGGING": "ON",
+        "ET_MIN_LOG_LEVEL": "Info"
```
manuelcandales, line 156: Info level is the default. You can delete this ET_MIN_LOG_LEVEL line, and it should still work

JacobSzwejbka: I prefer being verbose especially since I expect mostly LLMs will be reading/editing this file from now on.

manuelcandales: then use "EXECUTORCH_LOG_LEVEL": "Info"

manuelcandales:
i think you also need to enable logging in examples/models/parakeet/CMakePresets.json

PR #16516, opened 2026-01-09
Fix DecomposeLayerNormPass to handle 6-arg layer_norm

apullin:
Summary:
## Problem

When using `nn.LayerNorm` in models that go through the ARM backend's quantization flow, the `DecomposeLayerNormPass` fails with:

```
ValueError: DecomposeLayerNormPass: too many values to unpack (expected 2)
```

This happens because `torch.ops.aten.layer_norm.default` has **6 arguments**:
```
layer_norm(input, normalized_shape, weight, bias, eps, cudnn_enable)
```

But `DecomposeLayerNormPass` only handled up to 5 arguments (for `native_layer_norm`).

The error occurs during `transform_for_annotation_pipeline` in the ARM quantizer, which runs before edge transformation when the op is still `aten.layer_norm.default`.

## Solution

Add `case 6:` to the `match len(args)` block in `DecomposeLayerNormPass.call()` to handle the 6th argument (`cudnn_enable`). This argument is simply ignored during decomposition since it's only relevant for cuDNN GPU optimization.

## Testing

Added a new test file `test_layernorm_modai_compat.py` that:
1. Creates a simple Linear -> LayerNorm -> Linear model
2. Exports it via `torch.export`
3. Runs it through `transform_for_annotation_pipeline` (the exact path that was failing)
4. Verifies LayerNorm is decomposed correctly through the full TOSA pipelines

### DIFF
--- a/backends/arm/_passes/decompose_layernorm_pass.py
+++ b/backends/arm/_passes/decompose_layernorm_pass.py
@@ -90,6 +90,11 @@ class DecomposeLayerNormPass(ArmPass):
             args = node.args
             meta = node.meta
             match len(args):
+                case 6:
+                    # torch.ops.aten.layer_norm.default has 6 args:
+                    # (input, normalized_shape, weight, bias, eps, cudnn_enable)
+                    # cudnn_enable is not used in the decomposition
+                    x, normalized_shape, weights, bias, epsilon, _cudnn_enable = args
                 case 5:
                     x, normalized_shape, weights, bias, epsilon = args
                 case 4:
--- a/backends/arm/test/targets.bzl
+++ b/backends/arm/test/targets.bzl
@@ -19,7 +19,7 @@ def define_arm_tests():
         "ops/test_avg_pool2d.py",
         "ops/test_cat.py",
         "ops/test_conv2d.py",
-        "ops/test_linear.py", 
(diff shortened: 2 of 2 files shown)


backends/arm/_passes/decompose_layernorm_pass.py
```diff
@@ -90,6 +90,11 @@ def call(self, graph_module: torch.fx.GraphModule):
             args = node.args
             meta = node.meta
             match len(args):
+                case 6:
```
JacobSzwejbka, line 93: Is it safe to explicitly rely on the number of args like this? Its not considered a BC breaking change to add a new arg with a default value to an ATen op, so it could happen at any time.

JacobSzwejbka (approved):

backends/arm/_passes/decompose_layernorm_pass.py
apullin: It definitely is a code smell.

BUT I don't know enough about either end of this interaction to know how to cleverly control it, or make it correct-by-construction. Which would be better, yes.

Notably, the existing code already has hard-coded arg lengths below, anyway, so this is only an extension of the same.
At least this amendment has a comment for WHY it is 6.

PR #16520, opened 2026-01-09
Explicitly construct Span<EValue*> from EValue*[] in tests

rascani:
### Summary

GCC exhibits inconsistent behavior where it applies array-to-pointer decay when calling through function pointers, but not for direct function calls, even though both should invoke the same conversion to the parameter type.

Minimal reproduction:

    template <int N>
    struct S { S(int (&)[N]) {} };

    void f(S<3>) {}

    int main() {
        int a[3];
        f(a);       // OK on both GCC and Clang
        (&f)(a);    // OK on Clang, ERROR on GCC
    }

A similar issue is tracked at: [GCC Bugzilla 114812](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114812)

In ExecuTorch, OpFunction is defined as:

    using OpFunction = void (*)(KernelRuntimeContext&, Span<EValue*>);

When tests call `(*func)(context, stack)` where `stack` is an `EValue*[]` array, GCC decays it to `EValue**` before considering Span's array constructor `template<size_t N> Span(T (&)[N])`, causing compilation to fail.

This PR updates call sites to work around this issue by explicitly constructing the Span. An alternative option would be to convert the C arrays to std::arrays, which would not be subject to array to pointer decay.

### Test plan

```
# Using GCC 11.5
./test/run_oss_cpp_tests.sh
```


JacobSzwejbka (approved):

lucylq:
Wow thanks for the detailed explanation, TIL.

PR #16531, opened 2026-01-10
Add in_progress() state check to Method::execute()

rascani:
### Summary
Adds in_progress() to detect when execution is mid-way through the instruction sequence. Method::execute() now returns InvalidState if called while in_progress() is true.

execute() auto-resets step_state_ on error to allow retry attempts. step() does not auto-reset (requires manual reset_execution() call).

### Test plan
```
./test/run_oss_cpp_tests.sh
```

### DIFF
--- a/runtime/executor/method.cpp
+++ b/runtime/executor/method.cpp
@@ -1574,6 +1574,11 @@ Error Method::experimental_step() {
   return step();
 }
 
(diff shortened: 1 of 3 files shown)


GregoryComer (approved):

rascani:
CI failures are unrelated.

PR #16533, opened 2026-01-10
Support un-fused batchnorm1d/2d on XNNPACK via decomposition

GregoryComer:
Summary: Add a new pass - DecomposeBatchNorm - which converts standalone (non-fused) batch norm operators to 1x1 depthwise convolution. This prevents delegation graph breaks when batch norm operators can't be fused.

Differential Revision: D90422630

cc @digantdesai @cbilgin

### DIFF
--- a/backends/xnnpack/_passes/decompose_batch_norm.py
+++ b/backends/xnnpack/_passes/decompose_batch_norm.py
@@ -1,0 +1,247 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import logging
+import operator
+
+import torch
+from executorch.backends.transforms.utils import create_constant_placeholder
+from executorch.backends.xnnpack._passes.xnnpack_pass import XNNPACKPass
+from executorch.backends.xnnpack.utils.utils import (
+    check_or_raise,
+    get_param_tensor,
+    get_tensor_name,
+    is_param_node,
+)
+from executorch.exir.backend.utils import WhyNoPartition
+from executorch.exir.dialects._ops import ops as exir_ops
+from torch.export.graph_signature import InputKind
+from torch.fx.passes.infra.pass_base import PassResult
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+
+class DecomposeBatchNorm(XNNPACKPass):
+    """
+    Decompose batchnorm operators into 1x1 depthwise convolution.
+    """
+
+    BATCH_NORM_OPS = {
+        exir_ops.edge.aten.native_batch_norm.default,
+        exir_ops.edge.aten._native_batch_norm_legit_no_training.default,
+    }
+
+    @staticmethod
+    def can_decompose_batch_norm(
+        node: torch.fx.Node,
+        exported_program: torch.export.ExportedProgram,
+        why: WhyNoPartition | None = None,
+    ) -> bool:
+        """
+        Determine whether the given batch norm node can be decomposed by this pass.
+        """
+
+        if node.op != "call_function" or node.target not in DecomposeBatchNorm.BATCH_NORM_OPS:
+            return False
+
+        input_meta = node.args[0].meta["val"]
+
+        # Since we're converting to conv and XNNPACK doesn't support conv3d, we can't
+        # handle BatchNorm3d. Validate the input dimension. We'll take NC, NCL, or NCHW.
+        if input_meta.dim() not in (2, 3, 4):
+            if why:
+                why(node, f"Unsupported input rank {input_meta.dim()} for XNN batch norm operator.")
+            return False
+
+        # The batch norm node returns a tuple of output and other stuff we don't care about.
+        # All users must be getitem nodes that fetch the output (index 0).
+        # The partitioner should enforce this, but we'll check it here too.
+        for user in node.users:
+            if user.target != operator.getitem or user.args[1] != 0:
+                if why:
+                    why(node, "Batch norm users must only access the output tensor.")
+                return False
+
+        # Channel dimension and non-input args must be statically known.
+        if not isinstance(input_meta.shape[1], int):
+            if why:
+                why(node, f"Channel dimension must be statically known, but was {input_meta.shape[1]}.")
+            return False
+
+        if not is_param_node(exported_program, node.args[1]) or not is_param_node(exported_program, node.args[2]):
+            if why:
+                why(node, "Batch norm affine weight and bias must be static.")
+            return False
+
+        if not is_param_node(exported_program, node.args[3]) or not is_param_node(exported_program, node.args[4]):
+            if why:
+                why(node, "Batch norm running mean and variance must be static.")
+            return False
+
+        if isinstance(node.args[-1], torch.fx.Node):
+            if why:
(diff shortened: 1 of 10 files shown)


backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Since we're converting to conv and XNNPACK doesn't support conv3d, we can't
+        # handle BatchNorm3d. Validate the input dimension. We'll take NC, NCL, or NCHW.
+        if input_meta.dim() not in (2, 3, 4):
+            if why:
+                why(node, f"Unsupported input rank {input_meta.dim()} for XNN batch norm operator.")
+            return False
+
+        # The batch norm node returns a tuple of output and other stuff we don't care about.
+        # All users must be getitem nodes that fetches the output (index 0).
```
GregoryComer: Nit: Typo

backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,226 @@
...
+            # BatchNorm weights have shape [in_c].
+            # So we just need to unsqueeze the [in_c] to to [in_c, 1, 1, [1]].
+            input_meta = bn_node.args[0].meta["val"]
+            channel_count = input_meta.shape[1]
+            spatial_dims = max(input_meta.dim() - 2, 1) # Min of 1 since 1d can be NC or NCL.
+            new_weight_shape = [weight.shape[0], 1] + [1] * spatial_dims
+            weight = weight.reshape(new_weight_shape)
+
+            # Insert the new weight and biases in the graph.
+            # TODO?
```
GregoryComer: TODO: Make sure the new weights are properly serialized and the old weights are gone.

backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,226 @@
...
+                    exir_ops.edge.aten.squeeze_copy.dim,
+                    args=(conv_node, 2))
+                conv_node.replace_all_uses_with(squeeze_node)
+                # This gets overwritten by replace_all_uses_with. Maybe there's
+                # a better solution?
+                squeeze_node.args = (conv_node, *squeeze_node.args[1:])
+
+    # override
+    def call(self, graph_module: torch.fx.GraphModule):
+        # fall back to linear transform
```
GregoryComer: Nit: Extra comment copy-pasted.

backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,247 @@
...
+        gamma: torch.Tensor,        # [C], learned weight
+        beta: torch.Tensor,         # [C], learned bias
+    ) -> (torch.Tensor, torch.Tensor):
+        """
+        Compute equivalent per-channel weight and bias to match the batch norm
+        computation with frozen values.
+        """
+
+        # See https://docs.pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html
+        denom = torch.sqrt(running_var + torch.Tensor([eps]))
```
mergennachin: `torch.Tensor([eps])` won't preserve the dtype of running_var. You can do something like: 

```
eps_t = running_var.new_tensor(eps)
denom = torch.sqrt(running_var + eps_t)
```

backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,247 @@
...
+        graph_module: torch.fx.GraphModule,
+    ) -> torch.fx.Node:
+        """
+        Replace a BatchNorm with NCL or NCHW input with an equivalent depthwise
+        convolution.
+        """
+
+        # Compute the equivalent per-channel weights and biases.
+        # Note that the batch norm node args are
+        #   (input, gamma, beta, running_mean, running_var, [training], momentum, eps).
```
mergennachin, line 164: check that training is False

backends/xnnpack/_passes/decompose_batch_norm.py
```diff
@@ -0,0 +1,247 @@
...
+            beta=get_param_tensor(self.exported_program, bn_node.args[2]),
+        )
+
+        # Conv weights have shape [out_c, in_c/g, spatial...].
+        # For dw, in_c = g. The kernel is also 1x1 (or just 1, for 1d).
+        #
+        # BatchNorm weights have shape [in_c].
+        # So we just need to unsqueeze the [in_c] to to [in_c, 1, 1, [1]].
+        input_meta = bn_node.args[0].meta["val"]
+        channel_count = input_meta.shape[1]
```
mergennachin, line 180: if it's channels-last, we should also return early?

backends/xnnpack/test/ops/test_batch_norm.py
```diff
@@ -0,0 +1,279 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+
+import torch
+from executorch.backends.xnnpack.test.test_xnnpack_utils import randomize_bn
+from executorch.backends.xnnpack.test.tester import Tester
+
+
+class TestBatchNorm(unittest.TestCase):
```
mergennachin, line 14: - Looks like you're testing the no_training variant. Test the other variant too
- Add negative tests too

backends/xnnpack/test/passes/test_decompose_batch_norm.py
```diff
@@ -0,0 +1,104 @@
...
+        (
+            Tester(
+                self.BatchNorm1dNC(3).eval(),
+                (torch.randn(2, 3),),
+            )
+            .export()
+            .to_edge()
+            .check_count({self.bn_name: 1})
+            .run_passes(self.PassStage)
+            .check_count({self.conv_name: 1})
```
mergennachin, line 72: can you check that conv is in fact depthwise conv with 1×1 kernel?

GregoryComer: Done.

backends/xnnpack/test/ops/test_batch_norm.py
GregoryComer: Done.

backends/xnnpack/_passes/decompose_batch_norm.py
GregoryComer: It should be fine here, as the canonical shapes are still NCHW. I added a few more tests to cover the channels_last memory format case to make sure.

backends/xnnpack/_passes/decompose_batch_norm.py
GregoryComer: Done.

backends/xnnpack/_passes/decompose_batch_norm.py
GregoryComer: Done. I checked and the ATen kernels do the math in the compute dtype, which is higher precision for float16/bfloat16. I've just bumped it up to do this simplification in f64 and then cast back down to the target dtype. That should closely match the aten semantics, albeit aten uses compute dtype=f32 for in/out dtype=f32.

GregoryComer:
Updated to add test coverage around dtype, affine=False, memory_format/dim_order, and more comprehensive checks on the conv node created by the pass. Will re-open and request review once CI finshes.

mergennachin (approved):

GregoryComer:
Test failure is on a core ml test, so it appears to be a flake.

PR #16543, opened 2026-01-12
Add death test matchers to verify assertion messages

rascani:
### Summary
Added message matchers to death tests in 2 test files to verify
tests fail with the expected error messages, not just that they fail.

evalue_test.cpp (18 matchers):
- Type checks: "EValue is not an int", "EValue is not a"
- Null pointer checks: "Pointer is null", "pointer cannot be null"
- List pointer checks: "string/int/bool/double/tensor list pointer is null"
- BoxedEvalueList checks: "wrapped_vals/unwrapped_vals cannot be null"

tensor_util_test.cpp (29 matchers):
- Shape/dtype mismatches: "Tensors do not match"
- Dimension validation: "Ending/Starting dimension.*should be in the range"
- Empty matchers for stride checks (Windows regex limitations)

Note: Matchers use only cross-platform compatible regex features
(no brackets, unions, or grouping which fail on Windows).

### Test plan
```
./test/run_oss_cpp_tests.sh
```


larryliu0820 (approved):

rascani:
Failures unrelated.

PR #16552, opened 2026-01-13
Add prefill/decode multifunction support in ET

metascroy:
## Summary

This diff adds multifunction export support for static Llama models on CoreML. Multifunction models export separate prefill and decode graphs with weight sharing, enabling more efficient autoregressive generation compared to the single-method approach.

### Key Changes

**CoreML Backend Compiler (`coreml_preprocess.py`)**
- Added `MULTIMETHOD_WEIGHT_SHARING_STRATEGY` enum with `NONE` and `POSITIONAL` strategies
- Added `generate_multimethod_weight_sharing_strategy_compile_spec()` to enable weight sharing across methods
- Implemented multifunction CoreML model compilation using `ct.utils.MultiFunctionDescriptor`
- When weight sharing is enabled, weights from the first method are shared positionally with subsequent methods

**Model Metadata (`model_metadata.h`, `serde_json.mm`)**
- Added `MethodMetadata` struct to store per-method input/output names for multifunction models
- Extended `ModelMetadata` with `methods` map and `default_method` field
- Added `is_multifunction()` helper to detect multifunction models
- Updated JSON serialization to handle the new multifunction metadata format

**Runtime Changes (`ETCoreMLModelManager.mm`, `backend_delegate.mm`, `coreml_backend_delegate.mm`)**
- Updated `ETCoreMLModelManager` to set `functionName` on `MLModelConfiguration` only for multifunction models (based on `metadata.is_multifunction()`)
- Legacy single-function models continue to work with `functionName=nil`
- Added method name propagation through the delegate initialization path
- Updated model loading to use per-method input/output names when available

**Export Script (`export_static_llm_coreml.py`)**
- Added `--multifunction` flag to export models with separate prefill (seqlen=input_len) and decode (seqlen=1) methods
- Multifunction mode uses `generate_full_logits=False` for efficiency (only outputs last token logits)
- Single method mode (default) retains `generate_full_logits=True` for lookahead decoding support
- Generates combined metadata with method-specific prefixes (e.g., `decode_input_len`, `prefill_input_len`)

**New Runner (`run_static_llm_multifunction.py`)**
- Added dedicated runner for multifunction models
- Handles separate prefill and decode method execution
- Manages cache state transfer between prefill and decode phases
- Supports both 2D (generate_full_logits=False) and 3D (generate_full_logits=True) logits output

**Build System (`CMakeLists.txt`)**
- Fixed installation of CoreML backend headers

**Utilities (`extract_coreml_models.py`)**
- Updated model extraction script to handle multifunction models

**Documentation (`README.md`)**
- Added documentation for both export modes (single method and multifunction)
- Added comprehensive export options reference table
- Added usage examples for both modes

### Usage Examples

**Single Method Export (for lookahead decoding):**
```bash
python examples/apple/coreml/llama/export_static_llm_coreml.py \
    --checkpoint $HOME/models/llama1b/llama1b.pth \
    --params $HOME/models/llama1b/params.json \
    --output static_llm_coreml_model.pte \
    --input_len 32 \
    --max_context_len 1024
```

**Multifunction Export (separate prefill/decode):**
```bash
python examples/apple/coreml/llama/export_static_llm_coreml.py \
    --checkpoint $HOME/models/llama1b/llama1b.pth \
    --params $HOME/models/llama1b/params.json \
    --output static_llm_coreml_multifunction.pte \
    --input_len 64 \
    --max_context_len 1024 \
    --multifunction
```

**Run Single Method Model (with lookahead):**
```bash
python examples/apple/coreml/llama/run_static_llm.py \
    --model static_llm_coreml_model.pte \
    --params $HOME/models/llama1b/params.json \
    --tokenizer $HOME/models/llama1b/tokenizer.model \
    --prompt "Once upon a time" \
    --max_new_tokens 100 \
    --lookahead
```

**Run Multifunction Model:**
```bash
python examples/apple/coreml/llama/run_static_llm_multifunction.py \
    --model static_llm_coreml_multifunction.pte \
    --params $HOME/models/llama1b/params.json \
    --tokenizer $HOME/models/llama1b/tokenizer.model \
    --prompt "Once upon a time" \
    --max_new_tokens 100 \
    --input_len 64 \
    --max_context_len 1024
```

### Mode Comparison

| Feature | Single Method | Multifunction |
|---------|---------------|---------------|
| Sequence length | Fixed (input_len for both prefill & decode) | Separate (input_len for prefill, 1 for decode) |
| Logits output | Full (all tokens) | Last token only |
| Lookahead decoding | ✅ Supported | ❌ Not supported |
| Weight sharing | N/A | ✅ Enabled |
| Generation efficiency | Good with lookahead | Optimized decode step |

## Test Plan
New unit test +

Tested both export modes on Llama 1B:
1. Exported single method model with `--input_len 32 --max_context_len 1024`
2. Exported multifunction model with `--input_len 64 --max_context_len 1024 --multifunction`
3. Ran single method model with `--lookahead` flag
4. Ran multifunction model with matching input_len and max_context_len
5. Verified text generation produces coherent output for both modes

### DIFF
--- a/backends/apple/coreml/runtime/delegate/ETCoreMLModel.mm
+++ b/backends/apple/coreml/runtime/delegate/ETCoreMLModel.mm
@@ -200,12 +200,27 @@ void reset_state_for_feature_name(NSString *feature_name, MLState *state) {
         return nil;
     }
     
+    if (orderedInputNames == nil) {
+        ETCoreMLLogErrorAndSetNSError(error,
+                                      ETCoreMLErrorCorruptedModel,
+                                      "orderedInputNames must not be nil");
+        return nil;
+    }
+    
+    if (orderedOutputNames == nil) {
+        ETCoreMLLogErrorAndSetNSError(error,
+                                      ETCoreMLErrorCorruptedModel,
+                                      "orderedOutputNames must not be nil");
+        return nil;
+    }
+    
     self = [super init];
     if (self) {
         _mlModel = mlModel;
         _asset = asset;
         _orderedInputNames = [orderedInputNames copy];
         _orderedOutputNames = [orderedOutputNames copy];
+        
         _cache = [[NSCache alloc] init];
         _inputConstraintsByName = get_multi_array_input_constraints_by_name(mlModel.modelDescription);
         _outputConstraintsByName = get_multi_array_output_constraints_by_name(mlModel.modelDescription);
@@ -234,6 +249,15 @@ void reset_state_for_feature_name(NSString *feature_name, MLState *state) {
         BOOL lCopyData = copyData;
         NSString *argName = [nameEnumerator nextObject];
         MLMultiArrayConstraint *constraint = argConstraintsByName[argName];
+        
+        if (constraint == nil) {
+            ETCoreMLLogErrorAndSetNSError(error,
+                                          ETCoreMLErrorCorruptedModel,
+                                          "No constraint found for arg '%@'. Model may have mismatched input/output names.",
+                                          argName);
+            return nil;
+        }
+        
         const auto& layout = arg.layout();
         auto dataType = to_ml_multiarray_data_type(layout.dataType());
         MLMultiArray *multiArrayArg = nil;
--- a/backends/apple/coreml/runtime/delegate/ETCoreMLModelLoader.mm
+++ b/backends/apple/coreml/runtime/delegate/ETCoreMLModelLoader.mm
@@ -26,20 +26,6 @@ namespace {
         
         return result;
     }
-
-    ETCoreMLModel * _Nullable get_model_from_asset(ETCoreMLAsset *asset,
-                                                   MLModelConfiguration *configuration,
-                                                   const executorchcoreml::ModelMetadata& metadata,
-                                                   NSError * __autoreleasing *error) {
-        NSOrderedSet<NSString *> *orderedInputNames = ::get_ordered_set(metadata.input_names);
-        NSOrderedSet<NSString *> *orderedOutputNames = ::get_ordered_set(metadata.output_names);
-        ETCoreMLModel *model = [[ETCoreMLModel alloc] initWithAsset:asset
-                                                      configuration:configuration
-                                                  orderedInputNames:orderedInputNames
-                                                 orderedOutputNames:orderedOutputNames
-                                                              error:error];
-        return model;
-    }
 } // namespace
 
 @implementation ETCoreMLModelLoader
@@ -48,14 +34,30 @@ namespace {
                                           configuration:(MLModelConfiguration *)configuration
                                                metadata:(const executorchcoreml::ModelMetadata&)metadata
                                                 error:(NSError * __autoreleasing *)error {
+    if (compiledAsset == nil) {
+        return nil;
+    }
+    
+    // Use the metadata's ordered input/output names.
+    // For multifunction models, the caller should load the per-method metadata
+    // which contains the correct input/output names for that method.
+    NSOrderedSet<NSString *> *orderedInputNames = ::get_ordered_set(metadata.input_names);
+    NSOrderedSet<NSString *> *orderedOutputNames = ::get_ordered_set(metadata.output_names);
+    
     NSError *localError = nil;
-    ETCoreMLModel *model = (compiledAsset != nil) ? get_model_from_asset(compiledAsset, configuration, metadata, &localError) : nil;
+    ETCoreMLModel *model = [[ETCoreMLModel alloc] initWithAsset:compiledAsset
+                                                  configuration:configuration
+                                              orderedInputNames:orderedInputNames
+                                             orderedOutputNames:orderedOutputNames
+                                                          error:&localError];
     if (model) {
         return model;
     }
+    
     if (error) {
         *error = localError;
     }
+    
     return nil;
 }
                                         
--- a/backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
+++ b/backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
@@ -56,7 +56,7 @@ std::vector<std::string> canonical_path(NSString *path) {
     for (NSString *component in components) {
         result.emplace_back(component.UTF8String);
     }
-    
+
     return result;
 }
 
@@ -69,7 +69,7 @@ id<MLFeatureProvider> _Nullable get_feature_provider(NSArray<MLMultiArray *> *in
         NSString *input_name = [enumerator nextObject];
         features[input_name] = [MLFeatureValue featureValueWithMultiArray:input];
     }
-    
+
     return [[MLDictionaryFeatureProvider alloc] initWithDictionary:features error:error];
 }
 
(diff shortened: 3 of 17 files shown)


backends/apple/coreml/runtime/delegate/coreml_backend_delegate.mm
```diff
@@ -50,6 +52,63 @@
 using executorch::aten::Tensor;
 using executorch::runtime::kTensorDimensionLimit;
 
+/// Checks if the processed bytes represent a JSON reference to NamedDataStore.
+/// The JSON format is: {"version": 1, "key": "...", "method": "..."}
```
lucylq: I think we discussed this offline, but I'm wondering if we could have one multifunction with weights shared across all method+partitions, instead of the positional weight sharing (which may not be generalizable across graphs that aren't identical/highly similar?). 

I think the additions would be
1. In the JSON, save the method as {method_name}_{partition_idx}
2. At AoT set the key into multifunction blob as {method_name}_{partition_idx}

examples/apple/coreml/llama/run_static_llm_multifunction.py
```diff
@@ -0,0 +1,562 @@
...
+import torch
+import torch.utils._pytree as pytree
+
+from executorch.examples.models.llama.model_args import ModelArgs
+from executorch.examples.models.llama.runner.generation import next_token
+from executorch.examples.models.llama.static_attention import StaticAttentionIOManager
+from executorch.runtime import Runtime
+
+
+class Tokenizer:
```
lucylq: Use pytorch_tokenizers here

backends/apple/coreml/runtime/delegate/coreml_backend_delegate.mm
metascroy: I did think about this, and it's the reason I made weight sharing strategy an enum (to make this a future extension).

My primary concern here is CoreML failing at runtime if we make the multifunction too complex, but let me see if I can get this to work.

backends/apple/coreml/runtime/delegate/ETCoreMLModel.mm
```diff
@@ -204,8 +204,26 @@ - (nullable instancetype)initWithAsset:(ETCoreMLAsset *)asset
...
         _mlModel = mlModel;
         _asset = asset;
-        _orderedInputNames = [orderedInputNames copy];
-        _orderedOutputNames = [orderedOutputNames copy];
+        
+        // Use provided ordered names, or derive from model description as fallback
+        if (orderedInputNames != nil) {
+            _orderedInputNames = [orderedInputNames copy];
+        } else {
+            // Derive input names from the model's description in sorted order for determinism
```
metascroy: I think orderedInputNames should never be nil, so this can probably be deleted.

backends/apple/coreml/runtime/delegate/ETCoreMLModelLoader.mm
```diff
@@ -48,15 +34,22 @@ + (nullable ETCoreMLModel *)loadModelWithCompiledAsset:(ETCoreMLAsset *)compiled
...
-    return nil;
+    
+    // Use the metadata's ordered input/output names.
+    // For multifunction models, the caller should load the per-method metadata
+    // which contains the correct input/output names for that method.
+    NSOrderedSet<NSString *> *orderedInputNames = ::get_ordered_set(metadata.input_names);
+    NSOrderedSet<NSString *> *orderedOutputNames = ::get_ordered_set(metadata.output_names);
+    
+    ETCoreMLModel *model = [[ETCoreMLModel alloc] initWithAsset:compiledAsset
+                                                  configuration:configuration
```
metascroy, line 49: TODO: need to understand why we are no longer setting localError or calling get_model_from_asset

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
```diff
@@ -322,6 +360,29 @@ void add_compute_unit(std::string& identifier, MLComputeUnits compute_units) {
     identifier.append(to_string(compute_units));
 }
 
+void add_function_name(std::string& identifier, MLModelConfiguration *configuration) {
+    // NOTE: For multifunction CoreML models, we intentionally do NOT include the
```
metascroy: This node should be removed.

backends/apple/coreml/runtime/delegate/ETCoreMLModelManager.mm
```diff
@@ -322,6 +360,29 @@ void add_compute_unit(std::string& identifier, MLComputeUnits compute_units) {
...
+    // MLModelConfiguration determines which function is invoked at runtime when
+    // creating the MLModel from the cached compiled files.
+    //
+    // Previously this added "_func_{name}" to the identifier, which caused
+    // redundant compilations (once per function). Now we compile once and reuse.
+    (void)identifier;
+    (void)configuration;
+}
+
+void add_method_name(std::string& identifier, NSString *methodName) {
```
metascroy: Why do we have both both method_name and function_name?

backends/apple/coreml/runtime/delegate/backend_delegate.h
```diff
@@ -72,9 +72,12 @@ class BackendDelegate {
     ///
     /// @param processed The AOT blob.
     /// @param specs The specs at the time of compilation.
+    /// @param method_name The method name for multifunction model support (optional, may be nullptr).
     /// @retval An opaque handle to the initialized blob or `nullptr` if the
     /// initialization failed.
-    virtual Handle* init(Buffer processed, const std::unordered_map<std::string, Buffer>& specs) const noexcept = 0;
+    virtual Handle* init(Buffer processed,
```
metascroy, line 80: Do we need method_name on init?

It should be passed via backend context.

backends/apple/coreml/runtime/delegate/coreml_backend_delegate.mm
```diff
@@ -50,6 +52,63 @@
...
+///
+/// @param data Pointer to the processed bytes.
+/// @param size Size of the processed bytes.
+/// @return true if the bytes appear to be a JSON reference, false otherwise.
+bool isNamedDataReference(const void* data, size_t size) {
```
metascroy: I don't like how hacky this is.

backends/apple/coreml/runtime/delegate/model_metadata.h
```diff
@@ -29,7 +46,20 @@ struct ModelMetadata {
     inline ModelMetadata() noexcept { }
 
     /// Returns `true` if the metadata is valid otherwise `false`.
-    inline bool is_valid() const noexcept { return !identifier.empty() && !output_names.empty(); }
+    inline bool is_valid() const noexcept { return !identifier.empty() && (!output_names.empty() || !methods.empty()); }
```
metascroy: Why || !method.empty?

billmguo (approved):
this feature has been verified with 1b and 750MB model

backends/apple/coreml/runtime/delegate/coreml_backend_delegate.mm
metascroy: I've changed the runtime to accomodate this with (hopefully) only python changes in future.

The runtime looks at functionName in the json to decide what function to call, and it looks at method_name to look up metadata by input/output names.

So you can have multiple method/partition pairs all referring to the same key in NDM, but with different functionName.  Each will come with their own metadata as well.

billmguo (approved):

metascroy:
Landed as https://github.com/pytorch/executorch/pull/16791

PR #16499, opened 2026-01-07
WIP: Enable Parakeet on Metal

mergennachin:

PR #16562, opened 2026-01-13
Metal backend: enable Parakeet

manuelcandales:
This pull request builds on top of #16499 to introduce support for the Parakeet model in the Metal backend. The most important changes are grouped below:

### Parakeet export/lowering:

* Added support for Metal lowering to the Parakeet export/lowering script
* Provided custom linear decomposition. This achieved two objectives: avoided call to addmm, and avoided call to reinterpret_tensor_wrapper with 0 stride

### Operator updates:

* Added implementation for `aoti_torch_mps_bmm_out` to support batched matrix multiplication (bmm) in the Metal backend
* Fixed input channel dimension handling for grouped convolutions in `aoti_torch_mps_convolution` by reading the correct dimension from the weight tensor.

### Shim layer updates:

* Added implementation for `aoti_torch_new_tensor_handle`
* Enabled non-zero tensor storage offsets in `aoti_torch__reinterpret_tensor` by adjusting the data pointer instead of rejecting non-zero offsets, and updating memory tracking and Metal buffer mapping logic accordingly.
* Added the `metal_buffer_nocopy` function to map arbitrary memory pointers into Metal buffers, supporting cases where the data pointer is offset.
* Improved error messages in several stubbed shim functions by including the function name in the exception message for easier debugging.

### DIFF
--- a/backends/apple/metal/runtime/shims/et_metal.mm
+++ b/backends/apple/metal/runtime/shims/et_metal.mm
@@ -113,6 +113,18 @@ void metal_cleanup_resources() {
     }
 }
 
+void metal_buffer_nocopy(void* ptr, size_t nbytes, bool map_ptr_to_buffer) {
+    id<MTLDevice> device = get_metal_device();
+    id<MTLBuffer> subBuffer = [device newBufferWithBytesNoCopy:ptr
+                                                        length:nbytes
+                                                        options:MTLResourceCPUCacheModeWriteCombined | MTLResourceStorageModeShared
+                                                    deallocator:nil];
+
+    if (map_ptr_to_buffer) {
+        ptr_to_mtl_buffer[ptr] = subBuffer;  // Map contents to buffer
+    }
+}
+
 bool metal_is_device_pointer(void* ptr) {
     return ptr_to_mtl_buffer.find(ptr) != ptr_to_mtl_buffer.end();
 }
--- a/backends/apple/metal/runtime/shims/et_metal.mm
+++ b/backends/apple/metal/runtime/shims/et_metal.mm
@@ -113,16 +113,22 @@ void metal_cleanup_resources() {
     }
 }
 
-void metal_buffer_nocopy(void* ptr, size_t nbytes, bool map_ptr_to_buffer) {
+bool metal_buffer_nocopy(void* ptr, size_t nbytes, bool map_ptr_to_buffer) {
     id<MTLDevice> device = get_metal_device();
     id<MTLBuffer> subBuffer = [device newBufferWithBytesNoCopy:ptr
                                                         length:nbytes
                                                         options:MTLResourceCPUCacheModeWriteCombined | MTLResourceStorageModeShared
                                                     deallocator:nil];
+    if (!subBuffer) {
+        ET_LOG(Error, "metal_buffer_nocopy: Failed to create no-copy buffer (ptr=%p, nbytes=%zu)", ptr, nbytes);
+        return false;
+    }
 
     if (map_ptr_to_buffer) {
         ptr_to_mtl_buffer[ptr] = subBuffer;  // Map contents to buffer
     }
+
+    return true;
 }
 
 bool metal_is_device_pointer(void* ptr) {
--- a/backends/apple/metal/runtime/shims/et_metal_ops.mm
+++ b/backends/apple/metal/runtime/shims/et_metal_ops.mm
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
   }
 }
 
+AOTITorchError aoti_torch_mps_bmm_out(
+    AOTITensorHandle out,
+    AOTITensorHandle self,
+    AOTITensorHandle mat2) {
+
+  // Validate non-null handles
+  if (!out || !self || !mat2) {
+    ET_LOG(Error, "aoti_torch_mps_bmm_out: null tensor handles");
+    return Error::InvalidArgument;
+  }
+
+  @autoreleasepool {
+    try {
+      // Convert AOTITensorHandle to ExecutorTorch tensors
+      auto out_tensor = reinterpret_cast<Tensor*>(out);
+      auto self_tensor = reinterpret_cast<Tensor*>(self);
+      auto mat2_tensor = reinterpret_cast<Tensor*>(mat2);
+
+      // Validate tensor dimensions - bmm requires 3-D tensors
+      if (self_tensor->dim() != 3 || mat2_tensor->dim() != 3 || out_tensor->dim() != 3) {
+        ET_LOG(Error, "aoti_torch_mps_bmm_out: tensors must be 3-D. "
+               "Got self.dim=%zd (shape=[%d,%d,%d]), "
+               "mat2.dim=%zd (shape=[%d,%d,%d]), "
+               "out.dim=%zd (shape=[%d,%d,%d])",
+               self_tensor->dim(),
+               self_tensor->dim() > 0 ? (int)self_tensor->sizes()[0] : 0,
+               self_tensor->dim() > 1 ? (int)self_tensor->sizes()[1] : 0,
+               self_tensor->dim() > 2 ? (int)self_tensor->sizes()[2] : 0,
+               mat2_tensor->dim(),
+               mat2_tensor->dim() > 0 ? (int)mat2_tensor->sizes()[0] : 0,
+               mat2_tensor->dim() > 1 ? (int)mat2_tensor->sizes()[1] : 0,
+               mat2_tensor->dim() > 2 ? (int)mat2_tensor->sizes()[2] : 0,
+               out_tensor->dim(),
+               out_tensor->dim() > 0 ? (int)out_tensor->sizes()[0] : 0,
+               out_tensor->dim() > 1 ? (int)out_tensor->sizes()[1] : 0,
+               out_tensor->dim() > 2 ? (int)out_tensor->sizes()[2] : 0);
+        return Error::InvalidArgument;
+      }
+
+      int64_t B = self_tensor->sizes()[0];  // batch size
+      int64_t M = self_tensor->sizes()[1];  // rows of self
+      int64_t K = self_tensor->sizes()[2];  // cols of self / rows of mat2
+      int64_t N = mat2_tensor->sizes()[2];  // cols of mat2
+
+      // Validate shape constraints
+      // self: [B, M, K], mat2: [B, K, N], out: [B, M, N]
+      if (mat2_tensor->sizes()[0] != B) {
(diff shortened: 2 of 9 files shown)


examples/models/parakeet/README.md
```diff
@@ -25,26 +25,47 @@ python export_parakeet_tdt.py --audio /path/to/audio.wav
...
 
-First, build ExecuTorch with the LLM preset from the executorch root directory:
+First, build ExecuTorch with the appropriate preset from the executorch root directory:
 
 ```bash
+# For CPU/XNNPACK
 cmake --workflow --preset llm-release
+
+# For Metal (macOS)
+cmake --workflow --preset llm-debug-metal
```
mergennachin: Let's switch back to `llm-release-metal` in the instructions for consistency

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -363,48 +364,102 @@ def export_all(model):
...
     partitioner = {}
+    for key in programs.keys():
+        if key == "preprocessor":
+            partitioner[key] = []
+        else:
+            partitioner[key] = [XnnpackPartitioner()]
+    return partitioner, programs
+
+
+def _linear_bias_decomposition(input, weight, bias=None):
```
JacobSzwejbka, line 396: we naturally decompose this, if its not being decomposed its because the aoti backend probably specifies not to decompose it. So we should probably just disable that for metal

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
...
+  // Validate non-null handles
+  if (!out || !self || !mat2) {
+    ET_LOG(Error, "aoti_torch_mps_bmm_out: null tensor handles");
+    return Error::InvalidArgument;
+  }
+
+  @autoreleasepool {
+    try {
+      // Convert AOTITensorHandle to ExecutorTorch tensors
+      auto out_tensor = reinterpret_cast<Tensor*>(out);
```
JacobSzwejbka, line 643: Nit: It would be useful to wrap this cast in a helper function since we plan on changing the tensor definition backing the shim.

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
...
+      }
+
+      // Define shapes for graph placeholders and tensor data
+      NSArray<NSNumber*>* selfShape = @[@(B), @(M), @(K)];
+      NSArray<NSNumber*>* mat2Shape = @[@(B), @(K), @(N)];
+      NSArray<NSNumber*>* outShape = @[@(B), @(M), @(N)];
+
+      // Create cache key for this batched matrix multiplication
+      // Cache key includes: op_name, shape params {B, M, K, N}, dtype, transpose_flag
+      // This allows reuse when same BMM shape/dtype is called repeatedly
```
JacobSzwejbka: Does this happen often with transformers? I would think the shape mostly changes each iteration with growing context length

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
...
+
+      id<MTLDevice> device = get_metal_device();
+      if (!device) {
+        ET_LOG(Error, "aoti_torch_mps_bmm_out: Failed to get Metal device");
+        return Error::Internal;
+      }
+      (void)device;  // Used for validation, consistent with other ops
+
+      // Get Metal buffers for input and output tensors
+      id<MTLBuffer> self_buffer = get_mtl_buffer(self_tensor, "aoti_torch_mps_bmm_out", "self");
```
JacobSzwejbka, line 763: Just sort of curious why do you pass the fn name and arg name, are these keys to a dict?

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
...
+        ET_LOG(Error, "aoti_torch_mps_bmm_out: Failed to get current Metal stream");
+        return Error::Internal;
+      }
+
+      id<MTLDevice> device = get_metal_device();
+      if (!device) {
+        ET_LOG(Error, "aoti_torch_mps_bmm_out: Failed to get Metal device");
+        return Error::Internal;
+      }
+      (void)device;  // Used for validation, consistent with other ops
```
JacobSzwejbka: Whats going on here?

backends/apple/metal/runtime/shims/et_metal_ops.mm
```diff
@@ -626,6 +626,316 @@ AOTITorchError aoti_torch_mps_mm_out(
...
+  @autoreleasepool {
+    try {
+      // Convert AOTITensorHandle to ExecutorTorch tensors
+      auto out_tensor = reinterpret_cast<Tensor*>(out);
+      auto self_tensor = reinterpret_cast<Tensor*>(self);
+      auto mat2_tensor = reinterpret_cast<Tensor*>(mat2);
+
+      // Validate tensor dimensions - bmm requires 3-D tensors
+      if (self_tensor->dim() != 3 || mat2_tensor->dim() != 3 || out_tensor->dim() != 3) {
+        ET_LOG(Error, "aoti_torch_mps_bmm_out: tensors must be 3-D. "
```
JacobSzwejbka, line 649: If the failure is that dims is wrong why is it useful to print the shape. Especially since dims could be too big?

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -113,6 +113,18 @@ void metal_cleanup_resources() {
     }
 }
 
+void metal_buffer_nocopy(void* ptr, size_t nbytes, bool map_ptr_to_buffer) {
+    id<MTLDevice> device = get_metal_device();
+    id<MTLBuffer> subBuffer = [device newBufferWithBytesNoCopy:ptr
+                                                        length:nbytes
+                                                        options:MTLResourceCPUCacheModeWriteCombined | MTLResourceStorageModeShared
+                                                    deallocator:nil];
+
```
mergennachin, line 126: check for subBuffer for nullity/false-y

backends/apple/metal/runtime/shims/memory.cpp
```diff
@@ -496,11 +497,24 @@ AOTITorchError aoti_torch__reinterpret_tensor(
...
+  }
+
   // Increment the reference count for this memory address only if it is owned
   // by tensor
-  memory_to_n_tensor[data_ptr] = memory_to_n_tensor[data_ptr] == NOT_OWN
+  memory_to_n_tensor[adjusted_data] =
+      memory_to_n_tensor[adjusted_data] == NOT_OWN
       ? NOT_OWN
-      : memory_to_n_tensor[data_ptr] + 1;
+      : memory_to_n_tensor[adjusted_data] + 1;
```
mergennachin: Co-pilot review looks like legit: 

```
if (adjusted_data != data_ptr) {
      ET_LOG(Debug, "...");
      metal_buffer_nocopy(adjusted_data, tensor->nbytes(), true);
      memory_to_n_tensor[adjusted_data] = NOT_OWN;
  }
  if (memory_to_n_tensor[data_ptr] != NOT_OWN) {
      memory_to_n_tensor[data_ptr] += 1;
  }
```

backends/apple/metal/runtime/shims/memory.cpp
```diff
@@ -509,10 +523,92 @@ AOTITorchError aoti_torch__reinterpret_tensor(
 AOTITorchError aoti_torch_new_tensor_handle(
     Tensor* orig_handle,
     Tensor** new_handle) {
-  (void)orig_handle;
-  (void)new_handle;
-  throw std::runtime_error("Not implemented");
-  return Error::Internal;
+  ET_LOG(Debug, "aoti_torch_new_tensor_handle: entered");
```
mergennachin, line 532: Should you handle both zero and non-zero offset in this method?

mergennachin (commented):
See inline comments

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: just for logging

backends/apple/metal/runtime/shims/memory.cpp
manuelcandales: what do you mean? aoti_torch_new_tensor_handle doesn't take an offset

examples/models/parakeet/export_parakeet_tdt.py
manuelcandales: Your comment made me realize I should add a comment explaining this. This decomposition is the key to making things work. Linear does get decomposed, but not in a way that works for us. When bias is present, and tensors are 2D, this gets decomposed into addmm. This requires us to implement addmm in the Metal backend, but more importantly, reinterpret_tensor_wrapper gets called on the bias, to make it look like a 2D. That eventually makes its way to ExecuTorch, as a call to executorch::extension::from_blob with a 0 stride (to view a 1D tensor as a 2D tensor). ExecuTorch doesn't currently support that, and raises and error.
This decomposition avoids that problem, and also avoids having to implement addmm.

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: I am seeing bmm being called on encode, not on decode. I am seeing 72 calls to bmm on the inductor-generated code for encoding, with what seems to be 2 different shape combinations. So, this caching would be avoiding the redundant creation of 70 MPSGraphs.

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: unnecessary call to `get_metal_device`, cleaned up

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: cleaned up (originally introduced while debugging)

backends/apple/metal/runtime/shims/et_metal.mm
manuelcandales: done!

backends/apple/metal/runtime/shims/et_metal_ops.mm
manuelcandales: that would be better suited for a separate PR, since all other ops do it this way. Maybe just leave for when I do the migration to SlimTensor.

mergennachin (approved):
Thank you!

PR #16565, opened 2026-01-13
[slimtensor] integration into backend

Gasoonjia:
This diff makes cuda backend actually use slimtensor.
It:
updates cuda_backends to create slimtensor from given etensor
removed duplicate etensor-driven shim layers under cuda_backend
update cmake logic in both cuda backend and aoti backend
Perf maintains the same. Shows as before.

image

Worth to notice that we are still keeping two sets of common shims, one is etensor-based and for metal backend, the other is slimtensor-based which used by cuda backend, to not impact metal backend work.
When Metal backend finishs the migration, we should delete the duplicate common shims and only keep slimtensor-based one.

* __->__ #16565

Differential Revision: [D90606409](https://our.internmc.facebook.com/intern/diff/D90606409/)


larryliu0820 (approved):

larryliu0820 (approved):

PR #16570, opened 2026-01-13
add cuda doc

Gasoonjia:
Summary: Add doc for recent introed cuda backend

Differential Revision: D90635397

### DIFF
--- a/examples/cuda/scripts/export.py
+++ b/examples/cuda/scripts/export.py
@@ -21,8 +21,6 @@ from executorch.examples.models.model_factory import EagerModelFactory
 from executorch.exir import EdgeCompileConfig, to_edge_transform_and_lower
 
 from executorch.extension.export_util.utils import save_pte_program
-from torch._inductor.decomposition import conv1d_to_conv2d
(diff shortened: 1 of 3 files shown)


examples/cuda/scripts/export.py
```diff
@@ -88,24 +86,17 @@ def main():
...
-    print(exported_programs)
 
     partitioner = CudaPartitioner(
         [CudaBackend.generate_method_name_compile_spec(args.model_name)]
     )
-    # Add decompositions for triton to generate kernels.
-    exported_programs = exported_programs.run_decompositions(
-        {
-            torch.ops.aten.conv1d.default: conv1d_to_conv2d,
-        }
```
larryliu0820, line 100: why remove this?

larryliu0820 (approved):

examples/cuda/scripts/export.py
Gasoonjia: its already in cuda_backend (https://github.com/pytorch/executorch/blob/main/backends/cuda/cuda_backend.py#L137)

PR #16573, opened 2026-01-14
Cortex-M: Add transpose conv2d operator

rascani:
### Summary
Add quantized transpose convolution operator for the Cortex-M backend using CMSIS-NN's arm_transpose_conv_wrapper_s8 function.

Fixes #16106

### Test plan
```
./backends/cortex_m/test/build_test_runner.sh
pytest --config-file=backends/arm/test/pytest.ini backends/cortex_m/test/ops/test_conv_transpose.py
```


backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
```diff
@@ -0,0 +1,310 @@
...
+    const Tensor& requantize_multipliers,
+    const Tensor& requantize_shifts) {
+  if (input.dim() != kConvTransposeDim || weight.dim() != kConvTransposeDim ||
+      output.dim() != kConvTransposeDim) {
+    ET_LOG(Error, "quantized_transpose_conv2d_out: tensors must be 4-D");
+    context.fail(Error::InvalidArgument);
+    return false;
+  }
+
+  // Check for channels_last dim_order (NHWC: 0, 2, 3, 1)
```
AdrianLundell: Use this helper instead? https://github.com/pytorch/executorch/blob/73b0734524f216e1fc6add6cff776af4ea5855fe/backends/cortex_m/ops/cortex_m_ops_common.h#L132C13-L132C36

backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
```diff
@@ -0,0 +1,310 @@
...
+  if (stride.size() != 2 || padding.size() != 2 || output_padding.size() != 2 ||
+      dilation.size() != 2) {
+    ET_LOG(
+        Error,
+        "quantized_transpose_conv2d_out: stride, padding, output_padding, and dilation must have length 2");
+    context.fail(Error::InvalidArgument);
+    return false;
+  }
+
+  // Reject non-zero output_padding - CMSIS-NN does not support this parameter
```
AdrianLundell: My philosophy has been to reject these type of constraints in the AOT flow instead and keep the API to the cortex-m op 1:1 to the CMSIS-NN api. If there is no reason not to do that I would prefer that to keep it consistent.

backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
```diff
@@ -0,0 +1,310 @@
...
+    context.fail(buffer_or_error.error());
+    return out;
+  }
+  cmsis_context.buf = buffer_or_error.get();
+  cmsis_context.size = buffer_bytes;
+
+  const int32_t output_buffer_bytes =
+      arm_transpose_conv_s8_get_reverse_conv_buffer_size(
+          &transpose_conv_params, &input_dims, &filter_dims);
+  auto output_buffer_or_error = context.allocate_temp(
```
AdrianLundell, line 211: This will likely change with @mansnils work on the scratch buffer allocation, but it is fine to get something working for now.

AdrianLundell (changes requested):
Looks good!

backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
rascani: SGTM. Once the scratch buffer allocation PR lands, we can do a pass on all the operators to remove the temp allocator usage.

backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
rascani: I agree, I think that's a good call. I've moved these to AOT checks.

backends/cortex_m/ops/op_quantized_transpose_conv2d.cpp
rascani: Done.

AdrianLundell (approved):

PR #16577, opened 2026-01-14
Arm backend: Support a8w4 for convolution and linear.

Erik-Lundell:
Int4 is added as a TosaSpecialDtype. Since torch.dtype doesn't
have int4, the dtype is carried in an int8, quantized with
qmin=-7, qmax=7. This can then be detected when folding qparams
to add the tosa_special_dtype meta.
Some additional changes are needed to make sure the meta survives
all the way to TOSA.

Tests are added to conv2d, conv3d, depthwise conv, and linear.
One conv2d test case was actually dw conv, so it was modified.


zingo (approved):
OK to merge if/when all Arm backend and cortex-m tests passes!

PR #16579, opened 2026-01-14
Arm backend: Remove code duplication in Ethos-U pipelines

martinlsm:
- EthosU55PipelineINT and EthosU85PipelineINT contained copied code in their respective `__init__` functions. Introduce a common super class EthosUPipelineINTBase that runs the common code upon initialization.
- Some pipeline class names have the suffix "Maker". Remove this suffix to make naming of the classes consistent.


zingo (approved):
OK to marge if /when all Arm backend tests pass :)

Issue #16041, opened 2025-12-02
CORTEX_M: Memory optimization


AdrianLundell:
No work has been done looking into optimizing memory of the runtime. This ticket covers a broad investigation into what can be done in this space:
1. Can we optimize scratch buffer allocation (e.g. is it reused between kernels currently?)
2. Can we strip away anything from the elf to minimize runtime size?
3. Any other ideas to optimize performance related to memory

mansnils:
@AdrianLundell - I think we can close this now unless any other ideas?

AdrianLundell:
Agree!

PR #16580, opened 2026-01-14
PoC Cortex_m backend: Add support for CMSIS-NN scratch buffers

mansnils:
Use exir.memory.alloc for CMSIS-NN scratch buffers, which is ideal since it has a TensorSpec and gets memory planned but creates no additional operator overhead.
Use CMSIS-NN pybind wrapper to get correct buffer size.

Improves: https://github.com/pytorch/executorch/issues/16041

Comparing w/wo patch memory consumption for conv2d_x3 unit test case. Without patch:

I [executorch:arm_executor_runner.cpp:842 log_mem_status()] model_pte_program_size: 10208 bytes.
I [executorch:arm_executor_runner.cpp:843 log_mem_status()] model_pte_loaded_size: 10208 bytes.
I [executorch:arm_executor_runner.cpp:848 log_mem_status()] input_file_allocator_used: 10976 / 62914560 free: 62903584 ( used: 0 % )
I [executorch:arm_executor_runner.cpp:860 log_mem_status()] method_allocator_used: 5433 / 62914560 free: 62909127 ( used: 0 % )
I [executorch:arm_executor_runner.cpp:867 log_mem_status()] method_allocator_planned: 2560 bytes
I [executorch:arm_executor_runner.cpp:871 log_mem_status()] method_allocator_loaded: 2857 bytes
I [executorch:arm_executor_runner.cpp:875 log_mem_status()] method_allocator_input: 16 bytes
I [executorch:arm_executor_runner.cpp:876 log_mem_status()] method_allocator_executor: 0 bytes
I [executorch:arm_executor_runner.cpp:879 log_mem_status()] temp_allocator: 2097152

With patch:
I [executorch:arm_executor_runner.cpp:846 log_mem_status()] model_pte_program_size: 10336 bytes.
I [executorch:arm_executor_runner.cpp:847 log_mem_status()] model_pte_loaded_size: 10336 bytes.
I [executorch:arm_executor_runner.cpp:852 log_mem_status()] input_file_allocator_used: 11104 / 62914560 free: 62903456 ( used: 0 % )
I [executorch:arm_executor_runner.cpp:864 log_mem_status()] method_allocator_used: 6414 / 62914560 free: 62908146 ( used: 0 % )
I [executorch:arm_executor_runner.cpp:871 log_mem_status()] method_allocator_planned: 2560 bytes
I [executorch:arm_executor_runner.cpp:875 log_mem_status()] method_allocator_loaded: 3838 bytes
I [executorch:arm_executor_runner.cpp:879 log_mem_status()] method_allocator_input: 16 bytes
I [executorch:arm_executor_runner.cpp:880 log_mem_status()] method_allocator_executor: 0 bytes

Summary:
* The big temp_allocator used for scratch is removed in patch and no longer used except for Linear/FC but this is a PoC/Draft-PR anway
* method_allocator_planned: 2560 bytes is the same in both cases => This means there is 100% reuse of the scratch buffers
 *  increased model size and method_allocator_loaded delta => This can be explained by increased number of planned objects that describe scratch reuse (more meta data). This meta data should stay consistent if scratch buffer sizes increased so I think this acceptable

cc @digantdesai @freddan80 @per @zingo @oscarandersson8218 @Sebastian-Larsson @robell

### DIFF
--- a/backends/cortex_m/passes/convert_to_cortex_m_pass.py
+++ b/backends/cortex_m/passes/convert_to_cortex_m_pass.py
@@ -1,15 +1,21 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
-# Copyright 2025 Arm Limited and/or its affiliates.
+# Copyright 2025-2026 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
 
+from typing import Sequence
+
 import executorch.backends.cortex_m.ops.operators  # noqa
+import executorch.exir as exir
 
 import torch
 import torch.fx
+
+from cmsisnn_sizes import convolve_wrapper_s8_buffer_size_mve
+
 from executorch.backends.arm._passes.arm_pass_utils import get_first_fake_tensor
 from executorch.backends.cortex_m.passes.passes_utils import quantize_multiplier_aot
 
@@ -20,8 +26,78 @@ from executorch.backends.transforms.utils import (
 
 from executorch.backends.xnnpack._passes.xnnpack_pass import XNNPACKPass
 from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.tensor import TensorSpec
 from torch.export.graph_signature import InputKind
 from torch.fx.passes.infra.pass_manager import PassResult
+
+
+def shape_from_node(n: torch.fx.Node) -> list[int]:
+    spec = n.meta.get("spec", None)
+    if spec is not None and getattr(spec, "shape", None) is not None:
+        return [int(s) for s in spec.shape]
+
+    v = n.meta.get("val", None)
+    if v is not None:
+        if isinstance(v, (tuple, list)):
+            v = v[0]
+        return [int(s) for s in v.shape]
+
+    raise KeyError(f"No shape meta on node {n.format_node()} (need spec or val)")
+
+
+def cmsisnn_conv_s8_required_bytes_mve(
+    *,
+    x: torch.fx.Node,
+    conv_node: torch.fx.Node,
+    weight_shape: Sequence[int],
+    stride: Sequence[int],
+    padding: Sequence[int],
+    dilation: Sequence[int],
+    input_zero_point: int,
+    output_zero_point: int,
+    output_qmin: int,
+    output_qmax: int,
+) -> int:
+    # Input is NCHW (PyTorch); CMSIS-NN wants NHWC dims.
+    N, C_in, H, W = shape_from_node(x)
+
+    # Weight is (C_out, C_in/groups, kH, kW) in PyTorch
+    C_out, _, kH, kW = map(int, weight_shape)
+
+    # Output is NCHW; convert to NHWC dims.
+    N2, C_out2, H_out, W_out = shape_from_node(conv_node)
+
+    input_nhwc = [N, H, W, C_in]
+    filter_nhwc = [
+        C_out,
+        kH,
+        kW,
+        C_in,
+    ]  # CMSIS-NN convention: n=out_ch, h=kH, w=kW, c=in_ch
+    output_nhwc = [N2, H_out, W_out, C_out2]
+
+    stride_hw = [int(stride[0]), int(stride[1])]
+    padding_hw = [int(padding[0]), int(padding[1])]
+    dilation_hw = [int(dilation[0]), int(dilation[1])]
+
+    # CMSIS-NN conv_params offsets are "negative of zero point"
+    input_offset = -int(input_zero_point)
+    output_offset = -int(output_zero_point)
+
+    return int(
+        convolve_wrapper_s8_buffer_size_mve(
+            input_nhwc=input_nhwc,
+            filter_nhwc=filter_nhwc,
+            output_nhwc=output_nhwc,
+            padding_hw=padding_hw,
+            stride_hw=stride_hw,
+            dilation_hw=dilation_hw,
+            input_offset=input_offset,
+            output_offset=output_offset,
+            activation_min=int(output_qmin),
+            activation_max=int(output_qmax),
+        )
+    )
 
 
 class ConvertToCortexMPass(XNNPACKPass):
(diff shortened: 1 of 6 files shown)


backends/cortex_m/passes/convert_to_cortex_m_pass.py
```diff
@@ -1,15 +1,21 @@
...
 
+from typing import Sequence
+
 import executorch.backends.cortex_m.ops.operators  # noqa
+import executorch.exir as exir
 
 import torch
 import torch.fx
+
+from cmsisnn_sizes import convolve_wrapper_s8_buffer_size_mve
```
rascani, line 17: Where does cmsisnn_sizes come from? I don't see it in the upstream repo, but from the summary it sounds like this is a pybinding around the underlying `get_buffer_size` functions from [here](https://github.com/ARM-software/CMSIS-NN/blob/main/Source/ConvolutionFunctions/arm_convolve_get_buffer_sizes_s8.c).

If so, how portable are those get_buffer_size functions? We'll obviously be running those functions on x86 and 64-bit archs now. I would imagine the sizeof() usage would be the riskiest, but could be okay if we always stick to fixed width integers. 

Maybe it would be good to have some compile time checks that verify certain shapes result in certain buffer sizes?

rascani:
Thanks for putting this together @mansnils. I think the general approach makes sense to me, but I have a couple additional questions:

1. The buffer sizes required will likely need to vary based on the specific cortex-m configuration (ie, dsp and/or mve or neither). I don't think we've currently plumbed that sort of knowledge into the AOT stack, but it is appealing (albeit making the model more specialized). In the interim, were you thinking of always allocating based on the worst case size or only focus on MVE optimizations for now?

2. Another potential risk is drift between the CMSIS-NN version the model was compiled against vs the CMSIS-NN version the runtime uses. Model artifacts tend to have long lifetimes. Since we are also doing a runtime check that the compiled scratch buffer size can still fit, I think we're generally okay here as well. The only downside is that working models could start failing if the runtime is upgraded to a new version that wants more memory than the model was compiled with.

backends/cortex_m/passes/convert_to_cortex_m_pass.py
mansnils: The "cmsis_nn_sizes" are just local changes for now, and not yet upstreamed. So if/when committing to this approach we need official pybinding support in CMSIS-NN.

 The actual underlying get_buffer_size function in this case is this one: https://github.com/ARM-software/CMSIS-NN/blob/fc08374121353a41076389de7007f49a487bbdb6/Include/arm_nnfunctions.h#L216
 So it is actually intended for compiling on host and we have an equivalent for DSP. We only have this kind of host friendly functions for the get_buffer_size functions, which is fine since that is what we need here.

mansnils:
> Thanks for putting this together @mansnils. I think the general approach makes sense to me, but I have a couple additional questions:
> 
>     1. The buffer sizes required will likely need to vary based on the specific cortex-m configuration (ie, dsp and/or mve or neither). I don't think we've currently plumbed that sort of knowledge into the AOT stack, but it is appealing (albeit making the model more specialized). In the interim, were you thinking of always allocating based on the worst case size or only focus on MVE optimizations for now?
> 
>     2. Another potential risk is drift between the CMSIS-NN version the model was compiled against vs the CMSIS-NN version the runtime uses. Model artifacts tend to have long lifetimes. Since we are also doing a runtime check that the compiled scratch buffer size can still fit, I think we're generally okay here as well. The only downside is that working models could start failing if the runtime is upgraded to a new version that wants more memory than the model was compiled with.

1. A very good question. The interim plan is to focus on MVE for now. However as you say for this approach we need a way to signal AOT stack, what kind of processor we are targeting - scalar, DSP or MVE based.

2. Also a good question. To solve this I imagine we document carefully relation between actual CMSIS-NN release version and CMSIS-NN pybinding version. So that a user would be responsible for installing the right pip package version given the runtime version (the same version probably). We could also have runtime checks but would be nice to not rely on it as it affects memory footprint (maybe just in debug builds).

rascani:
> 1. A very good question. The interim plan is to focus on MVE for now. However as you say for this approach we need a way to signal AOT stack, what kind of processor we are targeting - scalar, DSP or MVE based.

SGTM, I'll prioritize looking at how to plumb those details through the AOT stack. Let me know if you or @AdrianLundell  already had something in mind.

> 2. Also a good question. To solve this I imagine we document carefully relation between actual CMSIS-NN release version and CMSIS-NN pybinding version. So that a user would be responsible for installing the right pip package version given the runtime version (the same version probably). We could also have runtime checks but would be nice to not rely on it as it affects memory footprint (maybe just in debug builds).

I think it makes sense to have some version warning checks where we can. My biggest concern is the runtime compatibility check because model artifacts have a tendency to stick around a long time. Since the compatibility is scoped to the buffer size, I do think its pretty limited risk. The two scenarios seem to be:

* If the runtime wants a bigger buffer than what was memory planned, we reject the model and tell the user to regenerate the model (or downgrade the runtime). 
* If the runtime wants a smaller buffer, we can still run the model (and maybe emit a debug log that a regenerated model could potentially save memory space).

Regardless, I think this is a pretty good plan.

mansnils:
> > 1. A very good question. The interim plan is to focus on MVE for now. However as you say for this approach we need a way to signal AOT stack, what kind of processor we are targeting - scalar, DSP or MVE based.
> 
> SGTM, I'll prioritize looking at how to plumb those details through the AOT stack. Let me know if you or @AdrianLundell already had something in mind.
> 
> > 2. Also a good question. To solve this I imagine we document carefully relation between actual CMSIS-NN release version and CMSIS-NN pybinding version. So that a user would be responsible for installing the right pip package version given the runtime version (the same version probably). We could also have runtime checks but would be nice to not rely on it as it affects memory footprint (maybe just in debug builds).
> 
> I think it makes sense to have some version warning checks where we can. My biggest concern is the runtime compatibility check because model artifacts have a tendency to stick around a long time. Since the compatibility is scoped to the buffer size, I do think its pretty limited risk. The two scenarios seem to be:
> 
>     * If the runtime wants a bigger buffer than what was memory planned, we reject the model and tell the user to regenerate the model (or downgrade the runtime).
> 
>     * If the runtime wants a smaller buffer, we can still run the model (and maybe emit a debug log that a regenerated model could potentially save memory space).
> 
> 
> Regardless, I think this is a pretty good plan.

SGTM, we will proceed with CMSIS-NN pybindings then.

rascani:
Any update here @mansnils?

mansnils:
> Any update here @mansnils?

Sorry for delay. Bindings are now merged: https://github.com/ARM-software/CMSIS-NN/commit/ad95bdf8460a68a88dfcff54b626cab221b9b570

mansnils:
More updates: https://github.com/pytorch/executorch/pull/18940

mansnils:
Closing now as this is up: https://github.com/pytorch/executorch/pull/19636

PR #16589, opened 2026-01-14
Add xnnpack-no-sme2 flavor to Android release workflow

kirklandsign:
### Summary
Add option to disable ARM SME2 for Android emulator compatibility, which may crash when SME2 kernels are used.

### Test plan
CI


GregoryComer (approved):
It's odd that it's crashing on emulator. I can try to take a look and we can fix the underlying issue, but I'm fine to just disable it for now.

kirklandsign:
> It's odd that it's crashing on emulator. I can try to take a look and we can fix the underlying issue, but I'm fine to just disable it for now.

GPT says something to do with emulator support on Apple Silicon.

PR #16604, opened 2026-01-14
Use custom rmsnorm in static attention

lucylq:
### Summary
Use custom RMSNorm for static attention. This is what's used in mha attention as well.

https://github.com/pytorch/executorch/blob/9cbe754f6ea66d99e9a7977c956faec686a96f03/examples/models/llama/attention.py#L365-L366

### Test plan
Exporting qwen with coreml. Qwen config sets `use_qk_norm` to true in the config, and then fails because of CoreML fp16 conversion error. Looks like the custom RMSNorm explicitly casts to fp32. 

https://github.com/pytorch/executorch/blob/9cbe754f6ea66d99e9a7977c956faec686a96f03/examples/models/llama/norm.py#L54

### DIFF
--- a/examples/models/llama/static_attention.py
+++ b/examples/models/llama/static_attention.py
@@ -14,6 +14,7 @@ from executorch.examples.models.llama.attention import (
     register_attention,
 )
 from executorch.examples.models.llama.model_args import ModelArgs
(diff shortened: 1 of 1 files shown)


metascroy:
@sxu can you make sure these changes are OK?

PR #16611, opened 2026-01-15
Add lora to static attention

lucylq:
### Summary
Add lora modules to static attention

1. Pass in adapter checkpoint & config
2. Convert checkpoint to meta format
3. Filter for LoraLinears in quantization
4. Update linears to be LoraLinear, if adapter checkpoint/config are passed in

### Test plan
Export & run llama1b lora model. See test plan in https://github.com/pytorch/executorch/pull/16606

File sizes of regular llama1b and lora llama1b. 
```
-rw-r--r--@ 1 lfq  staff   884356621 Jan 14 15:50 coreml_llama1b.pte
-rw-r--r--  1 lfq  staff   907086740 Jan 15 10:50 coreml-llama1b-lora.pte
```

### DIFF
--- a/examples/apple/coreml/llama/export_static_llm_coreml.py
+++ b/examples/apple/coreml/llama/export_static_llm_coreml.py
@@ -21,6 +21,7 @@ Usage:
 
 import argparse
 import json
+from typing import Optional
 
 import coremltools as ct
 import torch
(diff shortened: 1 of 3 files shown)


metascroy:
@lucylq does this export the base model and the lora adapter as separate models?

How does that work with weight sharing?  It seems that the lora model would be as big as the base model in the flow in this PR?

lucylq:
> @lucylq does this export the base model and the lora adapter as separate models?
> 
> How does that work with weight sharing? It seems that the lora model would be as big as the base model in the flow in this PR?

@metascroy This PR just adds support for lora in static attention and the export/runner scripts.

There are changes to the export script that we need to make to construct a multi-method model, before applying weight-sharing.

PR #16620, opened 2026-01-15
Allow symints to be created for arguments

per:
### Summary
Add test for creating args of SymInt type to be able to use them in view_copy nodes together with the fix to make the test pass.

### Test plan
Tested through CI tests.

### DIFF
--- a/exir/pass_base.py
+++ b/exir/pass_base.py
@@ -1,6 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
-# Copyright 2025 Arm Limited and/or its affiliates.
+# Copyright 2025-2026 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
@@ -191,6 +191,11 @@ class _ExportPassBase(PassBase):
                 if not hasattr(a, "constant") or a.constant is None:
                     raise ExportPassBaseError(f"Cannot add {a} to graph.")
                 a = a.constant
+            elif isinstance(a, torch.SymInt):
+                if a.node.constant is not None:
+                    return a.node.constant
+                else:
+                    return a
             node = super().create_arg(a)
             if (
                 isinstance(a, torch.Tensor)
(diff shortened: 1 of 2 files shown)


zingo:
Hi @SS-JIA / @metascroy this is touching files outside Arm code and need a review if possible

exir/pass_base.py
```diff
@@ -191,6 +191,11 @@ def create_arg(self, a: Argument) -> torch.fx.Node:
                 if not hasattr(a, "constant") or a.constant is None:
                     raise ExportPassBaseError(f"Cannot add {a} to graph.")
                 a = a.constant
+            elif isinstance(a, torch.SymInt):
```
mergennachin, line 194:
```
elif isinstance(a, (torch.SymInt, torch.SymFloat, torch.SymBool)):
```

and add corresponding unit test for symfloat and symbool please

thank you for finding this bug btw @per

SS-JIA:
Other than Mergen's comment, LGTM 👍

exir/pass_base.py
per: Not sure how to trigger the SymFloat and SymBool paths here, since it comes from the dynamic shape export, which implies SymInts only, right?

mergennachin: @per 

SymFloat - https://gist.github.com/mergennachin/db34d1272f06b3be61ed1910b5dc6eaa
SymBool - https://gist.github.com/mergennachin/48326573d23293865d6447ab3dad622f

per: Sorry, the proposed tests doesn't trigger the bug, since the symbool/symfloat aren't part of a list argument (as the shape argument is in view_copy). The tests you linked to are properly handled already (scalars and tensors).
The only alternative I've come up with is to manually construct a symbol with a shape_env, but that feels like a really constructed way to trigger the bug:
```
        shape_env = ShapeEnv()
        sym_bool = shape_env.create_unbacked_symbool()
        tracer_owner = ExportPass()
        tracer = tracer_owner.tracer
        tracer.create_arg([sym_bool])
```
I've haven't been able to track down any operator that takes a list of bools or list of floats as argument, so IMHO it makes sense to keep the test as is, since that is what is exercised through the normal export flow. Or am I missing something?

mergennachin (approved):

exir/pass_base.py
mergennachin: Sounds good, thank you for explaining and walking me through the code

PR #16622, opened 2026-01-15
[ET-VK] Do not partition unsupported dtypes

pssrawat:
Summary: For nodes with torch.complex dtypes, do not partition them. Currently the partitioner tries to partition and fails with unsupported dtype error.

Differential Revision: D90764480


SS-JIA (changes requested):

SS-JIA (approved):

PR #16623, opened 2026-01-15
NXP backend: Linear + BatchNorm QAT fusing

StrycekSimon:
### Summary

Adds two passes for inserting/removing simulated BatchNorm fusion for QAT training, similarly to how `_fuse_conv_bn_qat` adds simulated Conv+BatchNorm fusion in `prepare_qat_pt2e` function from TorchAO.

### Test plan

Added integration tests that covers newly added implementation,

cc @robert-kalmar


backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+            target_node = node.args[0]
+        else:
+            raise ValueError(
+                f"FakeQuantize node '{node}' should have at least one argument, but has {len(node.args)}."
+            )
+
+    return target_node
+
+
+def _get_input_nodes(graph_module: GraphModule) -> list[Node]:
```
roman-janik-nxp: Return value is different from type hint.

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+
+def _is_batch_norm(node: Node) -> bool:
+    return (
+        hasattr(node, "op")
+        and node.op == "call_function"
+        and node.target == torch.ops.aten.batch_norm.default
+    )
+
+
+def _is_linear(node: Node):
```
roman-janik-nxp: Inconsistent return type hints, please add them everywhere.

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+        return scale_factor
+
+    return WrapperModule(_compute_scale)
+
+
+def _is_batch_norm(node: Node) -> bool:
+    return (
+        hasattr(node, "op")
+        and node.op == "call_function"
+        and node.target == torch.ops.aten.batch_norm.default
```
roman-janik-nxp: In previous PR for Conv + BN during QAT you have this code:

```
 return node_.op == "call_function" and node_.target in [
        torch.ops.aten.batch_norm.default,
        torch.ops.aten.native_batch_norm.default,
        torch.ops.aten._native_batch_norm_legit_no_training.default,
    ]
```
Is omitting the two other variants intended?

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+def _get_input_nodes(graph_module: GraphModule) -> list[Node]:
+    input_nodes = []
+    for node in graph_module.graph.nodes:
+        if node.op == "placeholder":
+            input_nodes.append(node)
+
+    return tuple(input_nodes)
+
+
+def _get_tensor_shape(node: Node) -> Optional[torch.Size]:
```
roman-janik-nxp: Minor: 
```suggestion
def _get_tensor_shape(node: Node) -> torch.Size | None:
```

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+    Batch norm computation can be in some cases fused with preceding linear transformation like conv or linear (see fuse_batch_norm_with_linear_pass.py).
+    We cannot do this before the QAT training because we would lose the ability to update the batch norm statistics during QAT training.
+    This pass takes inspiration from already existing mechanics of simulated Conv+BN folding present in TorchAO implementation, and applies it to a torch.nn.Linear.
+    The implementation can be found in _fuse_conv_bn_qat function [1] that is being used in _prepare_qat_pt2e [2].
+
+    We simulate the fusion by:
+    1. Adding computation of the fused scale factor from batch norm parameters to the graph
+    2. Adding pre-processing of the linear weights with this scale factor and applying inverse to it's output
+    3. (Optionally) Moving bias application after the output denormalization to work in the same space it was originally pre-trained in
+    3. Keeping the batch norm operator in the graph for statistics tracking during QAT
```
roman-janik-nxp: step 4?

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+        bias_add_node.insert_arg(1, reshape_node)
+
+
+class AddSimulatedLinearBatchNormFusionQATPass(PassBase):
+    """
+    Batch norm computation can be in some cases fused with preceding linear transformation like conv or linear (see fuse_batch_norm_with_linear_pass.py).
+    We cannot do this before the QAT training because we would lose the ability to update the batch norm statistics during QAT training.
+    This pass takes inspiration from already existing mechanics of simulated Conv+BN folding present in TorchAO implementation, and applies it to a torch.nn.Linear.
+    The implementation can be found in _fuse_conv_bn_qat function [1] that is being used in _prepare_qat_pt2e [2].
+
```
roman-janik-nxp, line 171: Could graph illustration like in some other our passes be added?

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
```diff
@@ -40,6 +40,27 @@ class FuseBatchNormWithLinearPass(PassBase):
...
 
+    def _is_fq_node(self, node: Node):
+        return (
+            node is not None
+            and hasattr(node, "target")
+            and isinstance(node.target, str)
+            and node.target.startswith("activation_post_process")
+        )
+
+    def _unwrap_if_fq(self, node: Node):
```
roman-janik-nxp: Redefinition of the same function from [add_simulated_linear_bn_fusion_qat_pass.py](https://github.com/pytorch/executorch/pull/16623/changes#diff-8d506bd5d3a99d0f1d957e66bed5e5035efd591631442c04fe19d620c7cb14c9)?

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,346 @@
...
+
+    [1] https://github.com/pytorch/ao/blob/main/torchao/quantization/pt2e/quantize_pt2e.py
+    [2] https://github.com/pytorch/ao/blob/main/torchao/quantization/pt2e/qat_utils.py
+    """
+
+    def call(self, graph_module: GraphModule) -> Optional[PassResult]:
```
roman-janik-nxp: `Optional` can be replaced by `<type> | None`.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,175 @@
+import executorch.backends.nxp.tests.models as models
```
roman-janik-nxp, line 6: license missing

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,175 @@
...
+        ),
+        None,
+    )
+
+    assert len(graph_nodes) == expected_number_of_nodes
+
+    # Assert Linear weight being quantized and "normalized"
+    assert linear_node is not None
+    # "activation_post_process" is not a standard operator so we need to check by name.
+    assert all(n.target.startswith("activation_post_process") for n in linear_node.args)
```
roman-janik-nxp: There can still be an option to check it by its target class.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
robert-kalmar: Agree with Roman. The implementation of the functions is the same (except class method vs. function). Move them to separate module and import them to avoid code redundancy. These functions can be moved: `_unwrap_if_fq`, `_is_fq_node`, but even `_is_batch_norm` and `_is_linear`, or import them from `add_simulated_linear_bn_fussion_qat_pass` as you do in `remove_simulated_linear_bn_fusion_qat_pass.py`

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,167 @@
...
+
+        last_matched_subgraph_node.replace_all_uses_with(weight)
+
+        for node in match.nodes_map.values():
+            if node not in match.placeholder_nodes:
+                graph_module.graph.erase_node(node)
+
+
+def _is_div_node(node: Node) -> bool:
+    return hasattr(node, "target") and node.target == torch.ops.aten.div.Tensor
```
robert-kalmar: We are checking for node type in various way, either by checking the `node.op` is call function and target is specific aten op, or like here checking only the  op.target. 
Is there any difference? Can we unite?

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/add_simulated_linear_bn_fusion_qat_pass.py
StrycekSimon: I added it.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
StrycekSimon: I didn't want to introduce dependency on the Add/Remove passes as the `FuseBatchNormWithLinearPass` is not direcly related. I found a better solution.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: This was a bit tricky as the node itself does not contain any other information based on which you can decide. Anyways, I found a way... 😄

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,180 @@
...
```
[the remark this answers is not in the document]
StrycekSimon: This is correct behavior. The condition is a bit more excessive than necessary but in case anything changes in the produced graph, it should fail.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,395 @@
...
+):
+    """
+    Moves bias application from the normalized space to after simulated denormalization.
+    """
+    _ = _insert_disabled_linear_bias(graph_module, linear_b_node)
+
+    bias_add_node = _insert_bias_add_after_denorm(graph_module, denorm_output_node)
+
+    with graph_module.graph.inserting_before(bias_add_node):
+        linear_output_shape = linear_node.meta.get("tensor_meta").shape
```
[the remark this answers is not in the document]
StrycekSimon: `None` should not be there so let it just fail the test.

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,395 @@
...
+
+    return WrapperModule(_preprocess)
+
+
+def _get_compute_scale_factor_pattern() -> Callable:
+    def _compute_scale(
+        bn_weight: torch.Tensor,
+        bn_running_var: torch.Tensor,
+    ) -> torch.Tensor:
+        bn_eps = 1e-5
```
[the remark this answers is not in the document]
StrycekSimon: I forgot to put TODO there earlier. Thanks for reminding me. :-)

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
```diff
@@ -10,6 +10,21 @@
...
+
+    if _is_activation_post_process_node(node, named_modules):
+        if len(node.args) >= 1:
+            target_node = node.args[0]
+        else:
+            raise ValueError(
+                f"FakeQuantize node '{node}' should have at least one argument, but has {len(node.args)}."
+            )
+
+    return target_node
```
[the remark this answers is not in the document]
StrycekSimon: I do not see a reason for adding explicit None as it is already explicit enough.

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,173 @@
...
+    return hasattr(node, "target") and node.target == torch.ops.aten.zeros_like
+
+
+def _remove_late_bias_pattern(graph_module: GraphModule, bias_node: Node):
+    linear_b_users = list(bias_node.users.keys())
+
+    if len(linear_b_users) != 2:
+        return
+
+    zeros_node, reshape_node = linear_b_users
```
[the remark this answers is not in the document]
StrycekSimon: I added more robust checks.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/add_simulated_linear_bn_fusion_qat_pass.py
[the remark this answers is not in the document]
StrycekSimon: The reshape is not necessary in this case. I added that to both the Add and Remove pass.

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
StrycekSimon: Sure, I unified it.

backends/nxp/tests/models.py
```diff
@@ -481,6 +481,41 @@ def forward(self, x):
...
+        if num_dims == 3:
+            self.bn = torch.nn.BatchNorm1d(bn_features)
+        elif num_dims == 4:
+            self.bn = torch.nn.BatchNorm2d(bn_features)
+        elif num_dims == 5:
+            self.bn = torch.nn.BatchNorm3d(bn_features)
+        else:
+            raise ValueError(
+                f"Unknown input_dim: {len(input_shape)}, supported values are 1, 2 or 3."
+            )
```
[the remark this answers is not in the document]
StrycekSimon: Makes sense, I fixed it.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,186 @@
...
+        processed_model
+    ).graph_module
+
+    assert list(processed_model.graph.nodes)[8].args[1] == bn_eps
+
+    processed_model = RemoveSimulatedLinearBatchNormFusionQATPass()(
+        processed_model
+    ).graph_module
+
+    assert list(processed_model.graph.nodes)[-2].args[7] == bn_eps
```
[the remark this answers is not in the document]
StrycekSimon: Let's stick with the hard-coded index. The layout change would probably suggest something is wrong anyway...

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,200 @@
...
+        linear_w_fq_or_zeros = _unwrap_if_fq(
+            linear_node.args[2], named_modules=named_modules
+        )
+        has_late_bias = _is_zeros_like(linear_w_fq_or_zeros)
+
+        if has_late_bias:
+            _remove_late_bias_pattern(
+                graph_module, bias_node=linear_w_fq_or_zeros.args[0]
+            )
+
```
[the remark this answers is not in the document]
StrycekSimon: I will add it just for the consistency with the `fuse_batch_norm_with_linear_pass.py`. The linear op should always have the bias argument even if `bias=False`.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,186 @@
...
+    linear_node = graph_nodes[-4]
+
+    def _is_bn(node_: Node) -> bool:
+        return (
+            hasattr(node_, "target")
+            and node_.target == torch.ops.aten.batch_norm.default
+        )
+
+    assert len(graph_nodes) == 11
+
```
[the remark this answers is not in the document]
StrycekSimon: Let's keep it hard-coded.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,186 @@
...
+    exported_model = export(model, (random_input,), strict=True)
+    prepared_model = prepare_qat_pt2e(
+        exported_model.module(), NeutronQuantizer(neutron_target_spec, is_qat=True)
+    )
+
+    prepared_model = AddSimulatedLinearBatchNormFusionQATPass()(
+        prepared_model
+    ).graph_module
+    for data in (random_input,):
+        prepared_model(*data)
```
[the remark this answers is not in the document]
StrycekSimon: I fixed it.

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+def _get_bn_bias(bias_node: Node) -> float:
+    if "eps" in bias_node.kwargs:
+        return bias_node.kwargs["eps"]
+    elif len(bias_node.args) > 7:
+        return bias_node.args[7]
+    return 1e-5
+
+
```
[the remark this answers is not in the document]
StrycekSimon: Yep, there was a typoo. I fixed it.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
roman-janik-nxp: Where is the way? Code is still the same.

StrycekSimon: I fixed it on the most critical place - the passes itself, and forgot about the tests. I am adding it now.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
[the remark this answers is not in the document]
StrycekSimon: The case is handled by the the `_is_activation_post_process_node` being `False` when passing `None`.

backends/nxp/aten_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,410 @@
...
+        return node.meta["tensor_meta"].shape
+
+    return None
+
+
+def _get_bn_eps(bias_node: Node) -> float:
+    if "eps" in bias_node.kwargs:
+        return bias_node.kwargs["eps"]
+    elif len(bias_node.args) > 7:
+        return bias_node.args[7]
```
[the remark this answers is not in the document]
StrycekSimon: I fixed it.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,188 @@
...
+from torch.export import export
+from torch.fx import Node
+from torchao.quantization.pt2e.prepare import _is_activation_post_process_node
+from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_qat_pt2e
+
+
+@pytest.mark.parametrize("input_shape", [(1, 5, 5), (1, 5, 5, 5)])
+@pytest.mark.parametrize("linear_bias", [True, False])
+def test_add_simulated_linear_bn_fusing(input_shape, linear_bias):
+    random_input = torch.randn(*input_shape)
```
roman-janik-nxp: Please use this code to align with other tests like `test_clone_converter.py`:
```
calibration_inputs = get_random_calibration_inputs(
            to_model_input_spec(input_shape)
        )

        example_input = calibration_inputs[0]
```

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,188 @@
...
+        prepared_model
+    ).graph_module
+    prepared_model = FuseBatchNormWithLinearPass()(prepared_model).graph_module
+    converted_model = convert_pt2e(prepared_model)
+
+    quantized_output = converted_model(random_input)
+    graph_nodes = list(converted_model.graph.nodes)
+    linear_node = graph_nodes[-4]
+
+    def _is_bn(node_: Node) -> bool:
```
roman-janik-nxp: Could `def _is_batch_norm(node: Node) -> bool:` from `[add_simulated_linear_bn_fusion_qat_pass.py](https://github.com/pytorch/executorch/pull/16623/changes#diff-8d506bd5d3a99d0f1d957e66bed5e5035efd591631442c04fe19d620c7cb14c9) ` be used instead of local function?

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        assert graph_nodes[-3].args[0].target == torch.ops.aten.div.Tensor
+
+    assert raw_output.shape == fake_quantize_output.shape
+
+
+@pytest.mark.parametrize("input_shape", [(1, 5, 5), (1, 5, 5, 5)])
+@pytest.mark.parametrize("linear_bias", [True, False])
+def test_full_linear_bn_fusing(input_shape, linear_bias):
+    # TODO: Add pass for quantizing bias node when Linear has bias=False
```
roman-janik-nxp, line 104: Docs in [add_simulated_linear_bn_fusion_qat_pass.py](https://github.com/pytorch/executorch/pull/16623/changes#diff-8d506bd5d3a99d0f1d957e66bed5e5035efd591631442c04fe19d620c7cb14c9R234-R253) illustrates bias, why is it turned off?

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
roman-janik-nxp: Please also change Optional to `<type> | None` notation in this file.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2026 NXP
```
roman-janik-nxp, line 1: There should also be tests for using `executorch_pipeline.py` and the new passes should be added to pipeline on the right places. Now its only called from these tests. There is no verification that the modified model passes through pipeline conversions. I got errors trying to export the module to `ExportedProgram`.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,188 @@
...
+
+    exported_model = export(model, (random_input,), strict=True)
+    prepared_model = prepare_qat_pt2e(
+        exported_model.module(), NeutronQuantizer(neutron_target_spec, is_qat=True)
+    )
+    prepared_model = AddSimulatedLinearBatchNormFusionQATPass()(
+        prepared_model
+    ).graph_module
+
+    graph_nodes = list(prepared_model.graph.nodes)
```
roman-janik-nxp, line 64: There is a bug. `zeros_like` node is there twice and forms a loop, most probably causes failure of export.

StrycekSimon: Can you provide more context? I tried debugging it and found no issues.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: I will add it to the `executorch_pipeline.py` but not to the `NeutronAtenPassManager` as it should be called before the `NeutronAtenPassManager` in our pipeline.

Anyways, the error with exporting the module is indeed problematic and is most probably a limitation in dynamo. Found an open issue in pytorch repo that describes seemingly the same problem: https://github.com/pytorch/pytorch/issues/167063 The issue is with trying to call `export(graph_module, ...)` on a module that has `FakeQuantize` or `Observer` inserted by `prepare_pt2e`/`prepare_qat_pt2e`.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: This is currently a limitation and will be resolved by one of the future PRs.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: Do we need it here? I see eg. in `test_batch_norm_fusion.py` that we use torch to generate inputs.

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,203 @@
...
+from executorch.backends.nxp.aten_passes.fuse_batch_norm_with_linear_pass import (
+    _unwrap_if_fq,
+)
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch.fx.passes.utils.matcher_utils import InternalMatch, SubgraphMatcher
+from torchao.quantization.pt2e.qat_utils import _get_aten_graph_module_for_pattern
+
+
+def _is_op_node(node: Node, op) -> bool:
```
robert-kalmar: NIT: I love this helper function. 
And you have `_is_linear` and `_is_batch_norm`,  in `add_simulated_linear_bn_fusion_qat_pass.py` that can be defined using this helper function too, as you do few lines later. 
If you will do any other changes consider to create a `util.py` module and move all the _is_whatever` functions there. 
Alternatively feel free to create in in different PR cleaning up stuffs, to not spin around this PR. It is a detail.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
robert-kalmar, line 1: NIT: Consider adding both `add_simulated_linera_bn_fusion_qat_pass` and `remove_simulated_linera_bn_fusion_qat_pass` into one file. Because essentially they are usable only together in QAT training mode - transform and then inverse transformation.
This way you will:
- get rid of some duplicated, 
- importing from each other, 
- having the documentation in one place

But, as marked - nice to have.

robert-kalmar:
Only change I really would like to see is integration into the `executorch_pipeline.py` and to the `aot_neutron_convert.py` if it make sense (at this moment). Other my comments are nice to have.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
roman-janik-nxp: Cannot reproduce it anymore.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
StrycekSimon: Both files are pretty complex by themselves and contains a lot of pass-specific helper functions. I prefer sticking to separate files to clearly separate the context. Let's make a compromise and put them into a dedicated directory with `__init__.py` so both passes can be imported from the same module.

StrycekSimon:
> Only change I really would like to see is integration into the `executorch_pipeline.py` and to the `aot_neutron_convert.py` if it make sense (at this moment). Other my comments are nice to have.

I will add it to the `calibrate_and_quantize` function just to include it in tests and cover that there are no conflicts (failing tests) with the rest of the implementation. Just note that the only model in `aot_neutron_compile.py` that uses QAT so far is CifarNet on which it is pretty pointless to explicitly use it as it does not contain the Linear+BN pattern.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
robert-kalmar: It was a just a nit-pick. We do not need any compromise. I am fine with the current solution also.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,194 @@
...
+        return False
+
+    div_user_ops = [
+        user.target for user in node.users.keys() if hasattr(user, "target")
+    ]
+    if len(list(div_user_ops)) < 1:
+        return False
+
+    if torch.ops.aten.batch_norm.default in div_user_ops:
+        return True
```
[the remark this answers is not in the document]
StrycekSimon: Good one. 👍

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
StrycekSimon: I already added it. Both are in a dedicated subdir which makes them visibly related.

backends/nxp/aten_passes/remove_simulated_linear_bn_fusion_qat_pass.py
StrycekSimon: I added it into a separate `graph_utils.py` file and replaced it in the code introduced in this PR. The cleaning might be a part of some future PR...

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
StrycekSimon: Done ✅

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: After further in-person conversation I decided to add it. I am resolving this comment.

StrycekSimon:
After the recent push, one of PTQ test cases for mm op conversion fails. I will investigate.

Update: It's a mismatch by 1 in one of the output values, much like what can be seen in some other tests (eg. `test_clone_converter.py`) probably caused by float arithmetics as it does not fail when selecting different random seed value. Increasing `atol` to 1 like in similar tests resolved this issue.

backends/nxp/tests/models.py
```diff
@@ -481,6 +481,42 @@ def forward(self, x):
...
+            raise ValueError(
+                f"Unsupported dimension {len(input_shape)} of the input_shape "
+                + f"({input_shape}). Only 3, 4 and 5 are supported."
+            )
+
+        self.act = act
+
+    def forward(self, x):
+        x = self.linear(x)
+        x = self.bn(x)
```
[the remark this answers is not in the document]
StrycekSimon: Fixed

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: Fixed ✅

backends/nxp/tests/ir/converter/node_converter/test_mm_converter.py
```diff
@@ -60,6 +60,7 @@ def test_mm_conversion(self, _, use_qat: bool):
                 exported_program,
                 input_data,
                 tfl_model=tflite_flatbuffers_model,
+                atol=1.0,
```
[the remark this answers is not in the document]
StrycekSimon: We are working with integers here so 1.0 is the lowest possible non-zero error.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
```diff
@@ -0,0 +1,189 @@
...
+        None,
+    )
+
+    assert len(graph_nodes) == expected_number_of_nodes
+
+    # Assert Linear weight being quantized and "normalized"
+    assert linear_node is not None
+    assert all(
+        _is_activation_post_process_node(n, named_modules) for n in linear_node.args
+    )
```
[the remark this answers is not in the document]
StrycekSimon: For cases covered in this test, checking for non-quantized is not necessary. All parameters should be quantized and in case of `bias=False` we expect the bias parameter not being mentioned in the arguments.

backends/nxp/tests/models.py
```diff
@@ -481,6 +481,43 @@ def forward(self, x):
...
+        return self.act(x) if self.act is not None else x
```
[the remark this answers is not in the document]
StrycekSimon: Thanks, this is pretty on point and reviles possibly a serious problem. We are in both QAT and PTQ fusing the Linear+BatchNorm no matter the case which might be mathematically incorrect. If I understand it correctly, the only case when Linear+BN fusion is possible is when there is `nn.BatchNorm1d` with `(N, C)` inputs.

I will add a check to our passes in a different PR and will fix only tests and the model here.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
roman-janik-nxp: It's not, sorry.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
roman-janik-nxp: I see add it in `prepare_qat_pt2e()`, but there is no full pipeline test with `to_quantized_edge_program()`.

roman-janik-nxp (commented):
I miss the full pipeline test. Moving `is_batch_norm()` to `graph_utils.py` is great idea.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: Good point. I will add it.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
StrycekSimon: Don't know where it disappeared but it was there. 😀 I will fix it in a new PR, let's close this one.

Edit: I added it now.

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
robert-kalmar: @StrycekSimon , I am ok to add this in a separate PR. Internally we already test it with the anomaly detection model.

robert-kalmar (approved):

backends/nxp/tests/ir/edge_passes/test_linear_bn_fusing.py
StrycekSimon: I added it.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/remove_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,190 @@
...
+from executorch.backends.nxp.backend.graph_utils import is_batch_norm, is_op_node
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+from torch.fx.passes.utils.matcher_utils import InternalMatch, SubgraphMatcher
+from torchao.quantization.pt2e.qat_utils import _get_aten_graph_module_for_pattern
+
+_is_add = partial(is_op_node, target_op=torch.ops.aten.add.Tensor)
+_is_div = partial(is_op_node, target_op=torch.ops.aten.div.Tensor)
+_is_linear = partial(is_op_node, target_op=torch.ops.aten.linear.default)
+_is_reshape = partial(is_op_node, target_op=torch.ops.aten.reshape)
```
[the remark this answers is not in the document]
StrycekSimon: The `add_simulated_linear_bn_fusion_qat_pass.py` adds the reshape without overload so we can expect it.

backends/nxp/aten_passes/simulated_linear_bn_fusion_passes/add_simulated_linear_bn_fusion_qat_pass.py
```diff
@@ -0,0 +1,384 @@
...
+        reshape_target_shape = [1] * len(linear_output_shape)
+
+        bias_redirect_node = linear_b_node
+
+        if len(reshape_target_shape) > 1:
+            reshape_target_shape[1] = -1
+            bias_redirect_node = graph_module.graph.create_node(
+                op="call_function",
+                target=torch.ops.aten.reshape,
```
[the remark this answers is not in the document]
StrycekSimon: This should be consistent with the remove pass. I prefer using more broader op here as long as it works.

backends/nxp/aten_passes/fuse_batch_norm_with_linear_pass.py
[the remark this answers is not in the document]
StrycekSimon: This is a minor issue present in already existing implementation. Let's skip it and merge this PR.

StrycekSimon:
@roman-janik-nxp I added the pipeline test and fixed other minor issues you mentioned. Can we merge?

roman-janik-nxp (approved):

PR #16625, opened 2026-01-15
Add debug mode for parakeet

mergennachin:


CMakePresets.json
```diff
@@ -183,7 +183,9 @@
       ],
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Debug",
-        "CMAKE_INSTALL_PREFIX": "${sourceDir}/cmake-out"
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/cmake-out",
+        "EXECUTORCH_ENABLE_LOGGING": "ON",
+        "ET_MIN_LOG_LEVEL": "Info"
```
manuelcandales, line 188: why Info on llm-debug? why not Debug?

manuelcandales: also, based on my experience, if you want to see debug logs, you need to add "EXECUTORCH_LOG_LEVEL": "Debug"

manuelcandales: "ET_MIN_LOG_LEVEL": "Info" is redundant. See here:
https://github.com/pytorch/executorch/blob/00b39fd2faac6b4a72f7863173576d28749e6f63/runtime/platform/log.h#L23-L26

PR #16630, opened 2026-01-15
Optimize Parakeet export by combining components to reduce CPU-GPU round trips

mergennachin:
Summary:

Reduces CPU↔GPU data transfer overhead in Parakeet TDT inference by combining related neural network components into fewer ExecuTorch methods.

Previously, the encoder output required a separate joint_project_encoder call, and each token emission required separate decoder_predict and joint_project_decoder calls. This caused unnecessary round trips between CPU and GPU memory.

Now, encoder includes the transpose and projection operations, and decoder_step combines the RNN step with projection. This reduces exported methods from 6 to 4 and eliminates ~51 CPU↔GPU round trips per utterance (~1 for encoder, ~50 for decoder tokens).

Also do argmax in directly in GPU

Test plan

 - Export with --backend metal succeeds
 - C++ runner produces correct transcription
 - Timestamps match expected output
 - Optimized the whole inference from 650 downto 550ms. Thanks @manuelcandales for helping with the profiling.

### DIFF
--- a/examples/models/parakeet/main.cpp
+++ b/examples/models/parakeet/main.cpp
@@ -115,7 +115,6 @@ std::vector<Token> greedy_decode_executorch(
     int64_t pred_hidden = 640,
     int64_t max_symbols_per_step = 10) {
   std::vector<Token> hypothesis;
-  int64_t num_token_classes = vocab_size + 1;
 
   // f_proj is already transposed and projected by the encoder
   // Shape: [1, time_steps, joint_hidden]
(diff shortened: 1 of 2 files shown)


mergennachin:
cc @mattjcly

examples/models/parakeet/main.cpp
```diff
@@ -117,37 +117,11 @@ std::vector<Token> greedy_decode_executorch(
...
-  // Project encoder output
-  auto proj_enc_result = model.execute(
-      "joint_project_encoder",
-      std::vector<::executorch::runtime::EValue>{transposed_tensor});
-  if (!proj_enc_result.ok()) {
-    ET_LOG(Error, "joint_project_encoder failed");
-    return hypothesis;
-  }
-  auto f_proj = proj_enc_result.get()[0].toTensor();
+  // f_proj is already transposed and projected by the encoder
```
JacobSzwejbka: nit: drop this line

JacobSzwejbka (approved):
Looks good to me, can you clean up some of claude's comments

mergennachin:
cc @mattjcly - this is BC breaking change. You'll need to re-export the model again.

PR #16635, opened 2026-01-15
Add CoreML pte BC test

metascroy:
Introduces a CI test to verify backward compatibility of the CoreML runtime with older .pte files. This ensures that models exported with previous ExecuTorch releases continue to work correctly with the current runtime.

The test:

* Checks out ET 1.0.0 release and exports a simple model to .pte format using CoreML delegation
* Checks out the current version and runs the old .pte file with the current runtime
* Verifies the output is correct

The test runs on both macos-m1-stable and macos-m2-stable Apple Silicon runners for broader coverage.

### DIFF
--- a/.ci/scripts/test_coreml_bc.sh
+++ b/.ci/scripts/test_coreml_bc.sh
@@ -50,6 +50,8 @@ cleanup() {
     cd "${WORKING_DIR}"
     git checkout "${CURRENT_HEAD}" --quiet || true
     rm -f "${PTE_FILE}"
+    # Remove the conda environment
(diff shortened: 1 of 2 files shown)


GregoryComer (approved):
Nice. I should do something similar for XNNPACK.

Does it make sense to use a more complex model, like one from torchvision? Feel free to ignore this comment if add is sufficient.

metascroy:
> Nice. I should do something similar for XNNPACK.
> 
> Does it make sense to use a more complex model, like one from torchvision? Feel free to ignore this comment if add is sufficient.

For now I'll stick with add, but might extend in future.

PR #16636, opened 2026-01-15
[emitter] Ensure dim order and physical tensor memory match

### LINKED ISSUE #16464: Lowered parameter with channels last dim order is actually stored as contiguous
### 🐛 Describe the bug

I have played around with the channels last dim order (memory format) and discovered that trained parameters with this dim order are not actually stored as channels last in the `.pte` file. When I parse the `.pte` file (raw data), I can find the parameter's data stored in the contiguous dim order. Also, when I execute the model using `executor_runner` and debug the `kernels/portable/cpu/op_add.cpp` (which is the first operator to use the parameter in my model), I can see that the data is still contiguous, even though the parameter's `.dim_order()` returns `(0, 2, 3, 1)` (channels last dim order).

I have determined that the parameter data is contiguous by comparing it with the data from the PyTorch model before converting to channels last memory format and lowering to edge dialect. (`print()` statement in the code below)

I believe this is not the expected behavior and is probably the cause why my model is producing incorrect results.

Code to generate my model:
```python
import torch
from executorch.exir.capture._config import ExecutorchBackendConfig
from executorch.exir.program._program import to_edge_transform_and_lower


class AddModel(torch.nn.Module):

    def __init__(self):
        super().__init__()
        self.param = torch.nn.Parameter(torch.randn(1, 2, 3, 4))
        self.param.requires_grad = True

    def forward(self, x):
        return x + self.param


def to_executorch_program(model, example_input):
    exir_program_aten = torch.export.export(model, example_input, strict=True)

    partitioners = []
    edge_program_manager = to_edge_transform_and_lower(
        export(exir_program_aten.module(), example_input, strict=True),
        partitioner=partitioners,
    )

    return edge_program_manager.to_executorch(
        config=ExecutorchBackendConfig(extract_delegate_segments=False)
    )


if __name__ == "__main__":
    model = AddModel().eval()

    print("Contiguous flat data", model.param.data.flatten())
    model = model.to(memory_format=torch.channels_last)

    input_tensor = (torch.randn(1, 2, 3, 4).to(memory_format=torch.channels_last),)
    executorch_program_manager = to_executorch_program(model, input_tensor)

    executorch_program_manager.save("channels_last_dim_order_bug.pte")

```

The stored model is [attached](https://github.com/user-attachments/files/24449481/channels_last_dequantize_bug.zip).

### Versions

PyTorch version: 2.10.0.dev20251025+cpu
Is debug build: False
CUDA used to build PyTorch: None

OS: Ubuntu 20.04.6 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-2ubuntu1~20.04) 11.4.0
Clang version: Could not collect
CMake version: version 3.31.10
Libc version: glibc-2.31

Python version: 3.10.12 (main, Oct  1 2025, 11:13:47) [GCC 9.4.0] (64-bit runtime)
Python platform: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.31
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
Caching allocator config: N/A

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Byte Order:                           Little Endian
Address sizes:                        46 bits physical, 48 bits virtual
CPU(s):                               16
On-line CPU(s) list:                  0-15
Thread(s) per core:                   2
Core(s) per socket:                   8
Socket(s):                            1
NUMA node(s):                         1
Vendor ID:                            GenuineIntel
CPU family:                           6
Model:                                186
Model name:                           13th Gen Intel(R) Core(TM) i5-1350P
Stepping:                             2
CPU MHz:                              2188.810
BogoMIPS:                             4377.62
Virtualization:                       VT-x
Hypervisor vendor:                    Microsoft
Virtualization type:                  full
L1d cache:                            384 KiB
L1i cache:                            256 KiB
L2 cache:                             10 MiB
L3 cache:                             12 MiB
NUMA node0 CPU(s):                    0-15
Vulnerability Gather data sampling:   Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Not affected
Vulnerability Reg file data sampling: Mitigation; Clear Register File
Vulnerability Retbleed:               Mitigation; Enhanced IBRS
Vulnerability Spec rstack overflow:   Not affected
Vulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; RSB filling; PBRSB-eIBRS SW sequence; BHI BHI_DIS_S
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Not affected
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology tsc_reliable nonstop_tsc cpuid tsc_known_freq pni pclmulqdq vmx ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt clwb sha_ni xsaveopt xsavec xgetbv1 xsaves avx_vnni vnmi umip waitpkg gfni vaes vpclmulqdq rdpid movdiri movdir64b fsrm md_clear serialize flush_l1d arch_capabilities

[pip3] executorch==1.1.0a0+a069bba
[pip3] flake8==6.1.0
[pip3] flake8-breakpoint==1.1.0
[pip3] flake8-bugbear==24.4.26
[pip3] flake8-comprehensions==3.14.0
[pip3] flake8-plugin-utils==1.3.3
[pip3] flake8-pyi==23.5.0
[pip3] mypy==1.14.1
[pip3] mypy_extensions==1.1.0
[pip3] numpy==2.0.0
[pip3] optree==0.18.0
[pip3] pytorch_tokenizers==1.0.1
[pip3] torch==2.10.0.dev20251025+cpu
[pip3] torchao==0.14.0+git01849b2b1
[pip3] torchaudio==2.10.0.dev20251025+cpu
[pip3] torchdata==0.11.0
[pip3] torchsr==1.0.4
[pip3] torchtune==0.6.1
[pip3] torchvision==0.25.0.dev20251025+cpu
[conda] Could not collect

lucylq:
### Summary
At the emitter, resolved tensors should be in contiguous (NCHW) or channels-last (NHWC) format, with the correct corresponding physical memory. 

This PR checks that
1. If a tensor is not contiguous or channels-last, we call .contiguous on it.
2. After calling .contiguous, make sure the spec is updated with the new dim order.

Fixes a bug where channels-last tensors were made into NCHW, with the dim order not updated.

### Test plan
python -m unittest executorch.exir.emit.test.test_emit.TestEmit.test_emit_channels_last_constant

### DIFF
--- a/exir/emit/_emitter.py
+++ b/exir/emit/_emitter.py
@@ -83,6 +83,7 @@ from executorch.exir.schema import (
 )
 from executorch.exir.tensor import (
     AddressSpaceOverflowException,
+    dim_order_from_stride,
     layout_enum,
     make_allocation_info,
     make_tensor_value,
@@ -92,11 +93,9 @@ from executorch.exir.tensor import (
 )
 from executorch.exir.types import LeafValueSpec, ValueSpec
 from torch._subclasses.fake_tensor import FakeTensor
-
 from torch.export.exported_program import ExportedProgram, ExportGraphSignature
 from torch.fx.node import Node
 from torch.utils import _pytree as pytree
-
 from typing_extensions import TypeAlias
 
 
@@ -1994,14 +1993,20 @@ class _TopLevelEmitter(_Emitter):
 
             # assign the storage of the placeholder spec to the storage of the real tensor if there is one
             if real_tensor is not None:
-                # for non-contigous tensors, convert to a contiguous one
-                real_tensor = real_tensor.contiguous()
+                # For tensors that are neither contiguous nor channels-last, convert to contiguous format.
+                if not (
+                    real_tensor.is_contiguous()
+                    or real_tensor.is_contiguous(memory_format=torch.channels_last)
+                ):
+                    real_tensor = real_tensor.contiguous()
+
                 # Weights cannot be views during emission or serialization
                 if real_tensor.nbytes != real_tensor.untyped_storage().nbytes():
                     real_tensor = real_tensor.clone()
 
                 spec.storage = real_tensor.untyped_storage()
-
+                spec.stride = real_tensor.stride()
+                spec.dim_order = dim_order_from_stride(spec.stride)
             # User inputs and mutable buffers are not constants, other buffers or parameters are.
             if initialize_buffer and is_mutable_buffer:
                 spec.const = True
--- a/exir/emit/test/test_emit.py
+++ b/exir/emit/test/test_emit.py
@@ -13,7 +13,6 @@ from copy import deepcopy
 from typing import List, Optional, Tuple
 
 import executorch.exir as exir
-
 import executorch.exir.schema as schema
 import executorch.exir.tests.models as models
 import pytest
@@ -63,9 +62,7 @@ from executorch.extension.pybindings.portable_lib import (
 )
 from executorch.runtime import Runtime
 from torch import nn
-
 from torch._higher_order_ops import cond as torch_cond, map as torch_map
-
 from torch.export import Dim, export
 from torch.export.experimental import _export_forward_backward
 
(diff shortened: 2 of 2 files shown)


exir/emit/_emitter.py
```diff
@@ -1994,14 +1995,17 @@ def _is_buffer(node: Node, graph_signature: ExportGraphSignature) -> bool:
...
+                if not (real_tensor.is_contiguous() or real_tensor.is_contiguous(memory_format=torch.channels_last)):
+                    real_tensor = real_tensor.contiguous()
+
                 # Weights cannot be views during emission or serialization
                 if real_tensor.nbytes != real_tensor.untyped_storage().nbytes():
                     real_tensor = real_tensor.clone()
 
                 spec.storage = real_tensor.untyped_storage()
-
+                spec.stride = real_tensor.stride()
```
Gasoonjia, line 2008: Why didn't we reset stride before?

lucylq: edit: set it here instead of after .contiguous() call to make sure it's correct.

Gasoonjia: ahh no no i mean before this PR there should be some incontiguous tensor being replaced by contiguous one by calling `.contiguous()`, and for those tensors they should also update spec. Wondering why we didn't do that before?

lucylq: Yeah not sure. We do update in tensor.py after calling contiguous though:

https://github.com/pytorch/executorch/blob/d58c8eec8e071b9ecff4ec9be43ac34187364e23/exir/tensor.py#L189-L209

GregoryComer (approved):
Thanks for fixing this! It might be worth adding one test for a constant tensor that has neither contiguous nor channels-last dim order. One motivating case was something like torch.randn(2, 16).transpose(1, 0).

lucylq:
> Thanks for fixing this! It might be worth adding one test for a constant tensor that has neither contiguous nor channels-last dim order. One motivating case was something like torch.randn(2, 16).transpose(1, 0).

Will do. For this, we'd expect it to be converted to contiguous, as the portable ops etc. don't handle non-standard dim orders, and constants at this point aren't delegated. Is that what you're expecting too?

GregoryComer:
> > Thanks for fixing this! It might be worth adding one test for a constant tensor that has neither contiguous nor channels-last dim order. One motivating case was something like torch.randn(2, 16).transpose(1, 0).
> 
> Will do. For this, we'd expect it to be converted to contiguous, as the portable ops etc. don't handle non-standard dim orders, and constants at this point aren't delegated. Is that what you're expecting too?

Yeah, conversion to continuous would be fine, I think. For this case, it doesn't really matter which dim order the constant has, so long as the metadata matches the storage. I will need to convert it back in the delegate if it's (1, 0), though we might be able to optimize this in the future. I believe the portable ops can largely handle arbitrary dim orders, so long as the dim orders match for all inputs. I'm not sure how often this comes up, though, so I don't have a strong opinion on serialization of non-contig/non-channels-last constants.

The motivating case I ran into is from Yolo ([source link](https://github.com/ultralytics/ultralytics/blob/016fc41f401c3c246a05b3fd31e9162d1212e5ab/ultralytics/nn/modules/head.py#L174)). In this case, there is a constant tensor returned from make_anchors, which is transposed and ends up with dim order (0, 1). It's fed into an add, which is delegated to XNNPACK. The constant is fed into the CALL_DELEGATE subgraph as a USER_INPUT tensor (not constant - it's not consumed).

```py
def __init__(...):
    self.anchors = some_constant.transpose(1, 0)

def forward(self, x):
    y = x + self.anchors
```

exir/emit/_emitter.py
```diff
@@ -1994,14 +1995,17 @@ def _is_buffer(node: Node, graph_signature: ExportGraphSignature) -> bool:
 
             # assign the storage of the placeholder spec to the storage of the real tensor if there is one
             if real_tensor is not None:
-                # for non-contigous tensors, convert to a contiguous one
-                real_tensor = real_tensor.contiguous()
+                # for non-contiguous or channels-last tensors, convert to contiguous.
+                if not (real_tensor.is_contiguous() or real_tensor.is_contiguous(memory_format=torch.channels_last)):
```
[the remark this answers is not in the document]
lucylq: Sounds like only channels last up to 4d is supported here, cc @Gasoonjia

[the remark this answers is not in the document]
lucylq: See: https://github.com/pytorch/executorch/blob/b46f6b50a2f915fcda1469de8628d5df63784aa7/exir/dim_order_utils.py#L25-L31

PR #16638, opened 2026-01-15
Arm backend: Add decomposition pass and test for tan

AdrianLundell:


zingo (approved):
OK to merge if Arm test passed

PR #16642, opened 2026-01-15
solve norm layer false negtive gap

Gasoonjia:
* __->__ #16642

When comparing AOT intermediate outputs with runtime, we believed that AOT and runtime should have same output for same operator. But if there're multiple intermediate outputs from a single operator / single operator blob, the statement may not correct. Like drop out, which only record output tensor during AOT, but in runtime we record both mask and output tensor.

To support that, for 1 to many scenerio, instead of only take the last element for comparsion, we compare the runtime output sharing the same size and dtype with the aot one to have the best comparsion.

Differential Revision: [D90790256](https://our.internmc.facebook.com/intern/diff/D90790256/)


devtools/inspector/_inspector_utils.py
```diff
@@ -751,6 +751,168 @@ def dfs(node_id, component):
...
+    runtime_intermediate_output: Sequence,
+) -> Any:
+    """
+    Find the runtime output that matches the AOT output shape and dtype.
+    Used for multi-output operations (like native_layer_norm.out, native_dropout.out).
+
+    Returns:
+        The matching runtime output, or runtime_intermediate_output[-1] as fallback.
+    """
+    # Find all runtime outputs that match the AOT shape
```
GregoryComer, line 802: Is it practical to capture all of the outputs AOT so that we don't have to do this?

GregoryComer (approved):
I don't fully follow why we can't capture all of the reference outputs for each op to avoid needing to heuristically match, as both portable ops and delegate calls should have matching signatures AOT and and at runtime, but I'm assuming we have a good reason. I'll approve to unblock.

devtools/inspector/_inspector_utils.py
Gasoonjia: We are capturing all outputs from both AOT and runtime, but their outputs do not always match each other. 
Take dropout as example, AOT only generate the function output, but runtime generates two: mask and real output.

PR #16645, opened 2026-01-16
Arm backend: Use .module() in minimal examples

Erik-Lundell:
.graph_module doesn't work for models with weights. Also set check_guards=False, guards don't work well with et. This is addressed in et in a rather hacky way, see https://github.com/pytorch/pytorch/blob/19d8ad08cd8c6ee15c34117daaaea04d929778e4/torch/export/_unlift.py#L734


zingo (approved):
OK to merge!

PR #16663, opened 2026-01-16
Fix ARM test resource loading for TOSA schema and flatc binary

rascani:
Summary:
The `dbg_tosa_fb_to_json` function was failing in buck tests because it relied on `__file__`-relative paths to find the TOSA schema file, which doesn't work in buck-out link-trees. Additionally, the `flatc` binary wasn't available in the test environment.

This fix uses `importlib.resources` to properly load the schema file and flatc binary as buck resources, following the same pattern used in `executorch/exir/_serialize/`.

Differential Revision: D90879503

### DIFF
--- a/backends/arm/test/TARGETS
+++ b/backends/arm/test/TARGETS
@@ -22,11 +22,15 @@ runtime.python_library(
 runtime.python_library(
     name = "runner_utils",
     srcs = ["runner_utils.py"],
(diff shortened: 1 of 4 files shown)


mergennachin (approved):

zingo:
@rascani thanks for fixing this!

PR #16674, opened 2026-01-16
parakeet cuda mode works 

Gasoonjia:
When exporting with Dim.AUTO, GPU-enabled PyTorch adds CUDA-specific guards for convolution kernel based on cuDNN workspace calculations, which constrain the dynamic dimension based on the sample tensor size. With the original miscalculated 100-frame sample, this created a guard limiting inputs to ~160 mel frames (~1.6 sec), causing runtime failures for longer audio. 
On Mac/CPU-only PyTorch, these cuDNN guards are never added since no CUDA backend is selected, so NeMo's internal limit of 5000 frames is preserved so that we could have correct inference.
This PR fix the issue by making sample tensor matching the desired max audio duration (max_mel_frames), ensuring the cuDNN guard accommodates the full input range.

Also introduces ci for parakeet running on cuda backend.


JacobSzwejbka:
It seems like this PR just makes the max audio length user settable. Its not clear to me why this specifically was impacting your ability to run with cuda acceleration.

Gasoonjia:
> It seems like this PR just makes the max audio length user settable. Its not clear to me why this specifically was impacting your ability to run with cuda acceleration.

The thing is we miscalculated the num of mel we need to use for export from 1000 to 100; therefore when I working on portable mode on cuda server it can not support it due to the input is too large for the pte
not super sure why it can work on Mac (Claude said cpu PyTorch is not that strict so it can work but gpu PyTorch is stricter)
This pr makes the export file use the correct tensor for export, and make sure the miscalculation will not happen again

.github/workflows/cuda.yml
```diff
@@ -214,3 +214,108 @@ jobs:
...
+            --output-dir "${RUNNER_ARTIFACT_DIR}"
+
+        # Verify exported artifacts exist
+        test -f "${RUNNER_ARTIFACT_DIR}/parakeet_tdt.pte"
+        test -f "${RUNNER_ARTIFACT_DIR}/aoti_cuda_blob.ptd"
+        test -f "${RUNNER_ARTIFACT_DIR}/tokenizer.model"
+        ls -al "${RUNNER_ARTIFACT_DIR}"
+        echo "::endgroup::"
+
+  test-parakeet-cuda-e2e:
```
larryliu0820: Can you merge this into `test-model-cuda-e2e`

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -311,18 +327,22 @@ def export_all(model):
...
         },
         strict=False,
     )
     torch.cuda.is_available = old_cuda_is_available
 
     feat_in = getattr(model.encoder, "_feat_in", 128)
-    audio_signal = torch.randn(1, feat_in, 100)
-    length = torch.tensor([100], dtype=torch.int64)
+    # Use max_mel_frames for the sample input to help Dim.AUTO infer appropriate bounds
+    # max_mel_frames = max_audio_sec / window_stride (e.g., 60 sec / 0.01 = 6000 frames)
```
mergennachin: No need for this comment anymore if we're explicit about setting

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -296,12 +295,29 @@ def forward(
...
+    # Calculate sizes based on max_audio_sec
+    # max_audio_samples: raw audio samples for preprocessor
+    # max_mel_frames: mel spectrogram frames for encoder (~100 frames/sec with 10ms window_stride)
+    max_audio_samples = int(sample_rate * max_audio_sec)
+    max_mel_frames = int(max_audio_sec / window_stride)
+
     preprocessor_wrapper = PreprocessorWrapper(model.preprocessor)
     preprocessor_wrapper.eval()
-    sample_audio = torch.randn(16000 * 10)
+    # Use max_audio_samples for the sample input to help Dim.AUTO infer appropriate bounds
```
mergennachin: Also, no need for this comment.

mergennachin (commented):
<Deleted>

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -526,6 +546,12 @@ def main():
         choices=["fp32", "fp16", "bf16"],
         help="Model dtype for Metal/CUDA backends (default: fp32)",
     )
+    parser.add_argument(
+        "--max-audio-sec",
+        type=int,
+        default=30,
+        help="Maximum audio duration in seconds the model will support (default: 30)",
+    )
```
mergennachin: remove this

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -296,12 +295,29 @@ def forward(
...
+    Args:
+        model: The Parakeet model to export.
+        max_audio_sec: Maximum audio duration in seconds the model will support.
+    """
     programs = {}
 
+    # Get audio parameters from model config
+    sample_rate = model.preprocessor._cfg.sample_rate
+    window_stride = float(model.preprocessor._cfg.window_stride)
+
```
mergennachin, line 309: calculate directly

```
max_audio_sec = int(model.encoder.max_audio_length * window_stride)
```

mergennachin (approved):
See inline, land if CI is green

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -296,12 +295,29 @@ def forward(
         return mel, mel_len
 
 
-def export_all(model):
+def export_all(model, max_audio_sec: int = 60):
```
mergennachin: Remove this parameter altogether and calculate directly as below

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -553,7 +576,7 @@ def main():
 
     et = lower_to_executorch(programs, metadata=metadata, backend=args.backend)
 
-    pte_path = os.path.join(args.output_dir, "parakeet_tdt.pte")
+    pte_path = os.path.join(args.output_dir, "model.pte")
```
JacobSzwejbka, line 579: why change the default?

Gasoonjia: To in line with et-optimum for better ci workflow.

.ci/scripts/export_model_artifact.sh
```diff
@@ -101,9 +103,21 @@ case "$HF_MODEL" in
     PREPROCESSOR_FEATURE_SIZE=""
     PREPROCESSOR_OUTPUT=""
     ;;
+  nvidia/parakeet-tdt)
+    if [ "$DEVICE" = "metal" ]; then
+      echo "Error: Export for device 'metal' is not yet tested for model '$HF_MODEL'"
```
Gasoonjia, line 108: @manuelcandales we can introduce metal ci starting from here.

Issue #14644, opened 2025-09-28
Direct conversion of Ultralytics YOLO11n model to Executorch


LASTsue:
When attempting to convert a standard Ultralytics YOLO model (yolo11n.pt) directly to the Executorch .pte format using the torch.export and to_edge flow, the process fails with the error 'float' object has no attribute 'node'. This error occurs even when the model and inputs are explicitly placed on the CPU, ruling out device mismatches with the XNNPACK partitioner. 

Versions of the libraries：

- PyTorch version: 2.8.0
- Executorch version: 0.7.0
- Ultralytics YOLO version: 8.3.203
- Python version: 3.10
- OS: Ubuntu 22.04

Code:

```
import torch
from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
from executorch.exir import to_edge_transform_and_lower
from ultralytics import YOLO

yolo_model = YOLO("yolo11n.pt")
yolo_model=yolo_model.model
yolo_model.eval()

device = 'cpu'
yolo_model = yolo_model.to(device)

sample_inputs = (torch.randn(1, 3, 640, 640).to(device),)

exported=torch.export.export(yolo_model,args=sample_inputs)

et_program = to_edge_transform_and_lower(
    exported,
    partitioner=[XnnpackPartitioner()],
).to_executorch()

with open("yolo11n.pte", "wb") as file:
    et_program.write_to_file(file)
```

ERROR TRACEBACK:
```
'float' object has no attribute 'node'

While executing %item : [num_users=1] = call_function[target=torch.ops.aten.item.default](args = (%getitem_21,), kwargs = {})
GraphModule: class GraphModule(torch.nn.Module):
    ...

Original traceback:
File "/home/zll/miniconda3/envs/yolo/lib/python3.10/site-packages/ultralytics/nn/tasks.py", line 139, in forward
    return self.predict(x, *args, **kwargs)
  File "/home/zll/miniconda3/envs/yolo/lib/python3.10/site-packages/ultralytics/nn/modules/head.py", line 125, in forward
    y = self._inference(x)
```

SS-JIA:
cc: @digantdesai since this issue appears to be related to Xnnpack partitioning.

On my side, I will try to reproduce the issue as well. @LASTsue is there more to the stack trace? Specifically, I want to know where the failure happens:

```python
# Here?
exported=torch.export.export(yolo_model,args=sample_inputs)

# Here?
et_program = to_edge_transform_and_lower(
[...]

PR #16681, opened 2026-01-17
Normalize constant and buffer dim order in to_edge

GregoryComer:
Summary:
Add a pass in to_edge to ensure all constant tensors have contiguous or channels-last dim order. Tensors with dim orders not present in this list get converted to contiguous. This solves a class of bugs related to strided constant tensors, such as the one causing https://github.com/pytorch/executorch/issues/14644.

Context for this change:
* Some models, such as Yolo, include strided constant tensors.
* PT2E quantization can convert constant tensors to buffers.
* Prior to @lucylq's fix in https://github.com/pytorch/executorch/pull/16636, we would serialize these with contiguous data layout but strided metadata, causing correctness issues and/or runtime shape mismatch.
* As per discussion with @Gasoonjia, EXIR already assumes that tensors have either contiguous or channels last dim order.
* If we modify tensor dim order in a backend-visible manner, it has to happen before delegation. Otherwise, the backend will see a different dim order at runtime compared to lowering. 
   * Constants and buffers can be passed into delegate subgraphs as inputs. 
   * XNNPACK, for example, will conditionally insert transposes based on input dim order.
* Correspondingly, if we're going to convert dim order in EXIR, we need to do so before delegation.

Test Plan:
I've added tests for constants and buffers with a variety of dim orders and arbitrary strides in test_passes.py. I've also added some end to end test coverage under the XNNPACK backend that also covers the scenario we originally encountered with Yolo and XNNPACK.

Differential Revision: D90898806

### DIFF
--- a/backends/xnnpack/test/TARGETS
+++ b/backends/xnnpack/test/TARGETS
@@ -28,6 +28,28 @@ runtime.python_test(
 )
 
 runtime.python_test(
+    name = "test_xnnpack_fragments",
+    srcs = glob([
+        "fragments/*.py",
+    ]) + [
+        "test_xnnpack_utils.py",
+    ],
+    deps = [
+        "//executorch/backends/xnnpack/partition:xnnpack_partitioner",
+        "//executorch/backends/xnnpack/quantizer:xnnpack_quantizer",
+        "//executorch/backends/xnnpack/test/tester:tester",
+        "//executorch/devtools:lib",
+        "//executorch/devtools/bundled_program:config",
+        "//executorch/devtools/bundled_program/serialize:lib",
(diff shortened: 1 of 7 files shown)


GregoryComer:
Note to reviewers: There is some ongoing design discussion around dim order handling. We are actually already enforcing that all constants get converted to continuous or channels last during serialization in EXIR. So this PR is not adding any new restrictions, just moving the conversion from serialize -> to_edge so backends get the correct metadata AOT. I'm open to alternative approaches, though it may be worth landing this as is fix the bug while we discuss the long-term design.

Gasoonjia (approved):
Approve for unblocking dim order mismatch issue happen in delegate asap

PR #16688, opened 2026-01-19
Arm backend: Improve configuration of skipping quantization for nodes

martinlsm:
Prior to this change, nodes excluded for quantization based on the global quantization config were not tagged properly. This resulted in affected nodes got transformed for annotation (and possibly decomposed) by mistake when were not quantized. This patch fixes this issue.

Also, `TosaQuantizer._set_disallow_tfa_for_nodes` now checks `self.module_name_config` such that nodes can be excluded for "transformfor annotation" when they are referred to via their owning modules' names.


zingo (approved):
OK the marge after **the other comments hare handled** (marked resolved) and **Arm tests pass** (OK to re-trigger 1-2 times if it's flakye to prove it's works)

PR #16689, opened 2026-01-19
Arm backend: Simplify using a16w8 quantizer in tests

martinlsm:


zingo (approved):
OK to merge if/when all Arm tests pass

PR #16696, opened 2026-01-20
Arm backend: Update ai_ml_* pip packages to 0.8.0

mansnils:
Updates ai_ml_* pip packages (VGF/MLSDK) to 0.8.0. Makes pip installing VGF/MLSDK default.
Adds new option to build from source instead.

### DIFF
--- a/examples/arm/setup.sh
+++ b/examples/arm/setup.sh
@@ -53,7 +53,7 @@ OPTION_LIST=(
   "--disable-ethos-u-deps Do not setup what is needed for Ethos-U"
   "--enable-mlsdk-deps Setup what is needed for MLSDK"
   "--install-mlsdk-deps-with-pip (default) Use MLSDK PyPi package. This flag will be removed."
-  "--install-mlsdk-deps-from-src Build from source intead of using MLSDK PyPi packages"
+  "--install-mlsdk-deps-from-src Build from source instead of using MLSDK PyPi packages"
   "--mlsdk-manifest-url URL to the MLSDK manifest for vulkan."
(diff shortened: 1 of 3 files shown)


examples/arm/setup.sh
```diff
@@ -52,7 +52,8 @@ OPTION_LIST=(
   "--enable-emulation-layer Enable MLSDK Vulkan emulation layer"
   "--disable-ethos-u-deps Do not setup what is needed for Ethos-U"
   "--enable-mlsdk-deps Setup what is needed for MLSDK"
-  "--install-mlsdk-deps-with-pip Use MLSDK PyPi package instead of building from source"
+  "--install-mlsdk-deps-with-pip (default) Use MLSDK PyPi package. This flag will be removed."
+  "--install-mlsdk-deps-from-src Build from source intead of using MLSDK PyPi packages"
```
zingo: nit: intead -> instead

zingo:
@SS-JIA or @swolchok  this updated some packages and might need some internal updates/testing.
It would also be nice to get this into 1.1 so Ill tag it :)

examples/arm/setup.sh
mansnils: Done

SS-JIA (approved):
This PR appears to be updating dependencies only for Github CI tests, which will not impact internal tests (i.e. none of the files touched will impact internal CI testing).

I also cannot find an internal version of ai_ml_* packages in Meta internal repository; therefore there is no need for internal upgrade either like what was needed for the tosa upgrade.

LGTM!

PR #16703, opened 2026-01-20
Arm backend: Extend fp64 partition check to more dtypes

Erik-Lundell:
Check both inputs and outputs.
Specifically add check for bf16 operators.


zingo (approved):
OK to merge  when and if all Arm backend tests pass and the other tests is consistent with main.

PR #16709, opened 2026-01-20
Enable program_test in CI and fix flaky LoadFromMutableSegment test

mergennachin:
Enable program_test target in CMakeLists.txt which was previously
disabled (TODO T191569140).

Fix the LoadFromMutableSegment test which was fragile due to checking
for a hardcoded byte value (232) that varies across platforms and
PyTorch versions. The test now:

1. Loads the full 36-byte weight tensor instead of a single byte
2. Verifies data consistency by comparing two loads with memcmp
3. Checks that loaded data is non-trivial (contains non-zero bytes)
4. Validates source immutability by mutating local buffer and
   reloading to verify original data is preserved

This maintains the test's purpose of validating mutable segment
loading while being robust across different environments.

### DIFF
--- a/runtime/executor/test/program_test.cpp
+++ b/runtime/executor/test/program_test.cpp
@@ -187,6 +187,14 @@ TEST_F(ProgramTest, BadMagicFailsToLoad) {
   }
 }
 
+// These tests require ET_ENABLE_PROGRAM_VERIFICATION to be enabled.
+// In Release builds, verification is disabled by default to save binary size.
+#ifndef ET_ENABLE_PROGRAM_VERIFICATION
+#define ET_ENABLE_PROGRAM_VERIFICATION 1
+#endif
+
+#if ET_ENABLE_PROGRAM_VERIFICATION
+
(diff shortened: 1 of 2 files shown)


runtime/executor/test/program_test.cpp
```diff
@@ -508,32 +508,49 @@ TEST_F(ProgramTest, LoadFromMutableSegment) {
...
+  err = ProgramTestFriend::load_mutable_subsegment_into(
+      &program.get(), 0, 1, kWeightSize, buffer2);
+  EXPECT_EQ(err, Error::Ok);
+
+  // Verify both loads return identical data
+  EXPECT_EQ(std::memcmp(buffer, buffer2, kWeightSize), 0)
+      << "Loading the same segment twice should return identical data";
+
+  // Verify that the loaded data is non-trivial (not all zeros).
+  // Neural network weights initialized with torch.manual_seed(0) should
```
JacobSzwejbka: Nit: remove this comment

JacobSzwejbka (approved):

PR #16718, opened 2026-01-20
ExecuTorch MLX delegate

metascroy:
# Summary

This PR adds an MLX backend for ExecuTorch, enabling Metal-accelerated inference on Apple Silicon. It runs Llama, Qwen, Gemma, Whisper, Voxtral, and Parakeet models end-to-end, with 637 passing op tests and multithreaded execution support. For many models, it offers best performance among all ExecuTorch backends on Apple Silicon, offering 2-6x speedups over what was previously possible with ExecuTorch, and up to 30% smaller model sizes compared to XNNPACK due to BF16 support and tied quantized embedding support.

The PR is large due to extensive op coverage, testing, and documentation, but almost all changes are confined to `backends/mlx/`. The design is described in `backends/mlx/README.md`.

**Suggested review approach:**

1. **Review files outside `backends/mlx/` carefully** — these integrate with ExecuTorch's build system and are the most likely to need changes.
2. **For `backends/mlx/`**, focus on structural design (see README) and test coverage (CI job is .github/workflows/mlx.yml)

## Prerequisite PRs

These fixes were developed alongside the MLX backend. Once merged, this PR can be rebased to remove the duplicated changes:

- [#17257](https://github.com/pytorch/executorch/pull/17257) — Improve lowering time with NamedDataMap
- [#17679](https://github.com/pytorch/executorch/pull/17679) — Allow transform passes in etLLM
- [#17678](https://github.com/pytorch/executorch/pull/17678) — Fix dynamic shape bug in `remove_noop_pass`
- [#17378](https://github.com/pytorch/executorch/pull/17378) — Fix pocketfft intermittent bus errors on macOS ([upstream fix](https://github.com/mreineck/pocketfft/pull/28))

## Tests

CI is defined in `.github/workflows/mlx.yml`:

- **test_ops.py**: 637 passing op tests
- **Multithreading**: launches models on 50 threads, verifies correctness
- **GenAI E2E**: parakeet, voxtral, etLLM (stories110m), HF LLMs (llama, qwen, gemma)
- **backend-tester**: 380 passed, 1 failed, 86 skipped operator tests; 34 passed, 3 failed, 5 skipped model tests

The 1 failing operator test is a test-side issue being fixed in [#17539](https://github.com/pytorch/executorch/pull/17539). The 3 failing model tests will be addressed in follow-ups — they are not an initial focus compared to the GenAI models above.


extension/llm/export/builder.py
```diff
@@ -451,7 +451,9 @@ def to_backend(self, partitioners: Optional[List[Partitioner]]) -> "LLMEdgeManag
         return self
 
     def to_edge_transform_and_lower(
-        self, partitioners: Optional[List[Partitioner]]
+        self,
```
metascroy, line 454: Will remove once https://github.com/pytorch/executorch/pull/17679 lands

exir/passes/remove_noop_pass.py
```diff
@@ -77,9 +77,15 @@ def call(self, graph_module: GraphModule) -> PassResult:
                 continue
 
             if node.target == torch.ops.aten.slice_copy.Tensor:
-                # Only do this check if all the dims are static.
-                if all(isinstance(dim, int) for dim in orig_tensor.size()):
-                    if orig_tensor.shape == node.meta["val"].shape:
+                output_tensor = node.meta["val"]
```
metascroy, line 80: Will remove once https://github.com/pytorch/executorch/pull/17678 lands

backends/mlx/examples/llm/export_llm_hf.py
```diff
@@ -0,0 +1,464 @@
...
+    linear_group_size: Optional[int] = None,
+    embeddings_group_size: Optional[int] = None,
+) -> None:
+    import executorch.exir as exir
+    from executorch.backends.mlx import MLXPartitioner
+    from executorch.backends.mlx.passes import get_default_passes
+    from executorch.exir import EdgeCompileConfig
+    from executorch.exir.capture._config import ExecutorchBackendConfig
+    from executorch.exir.passes import MemoryPlanningPass
+    from optimum.exporters.executorch.tasks.causal_lm import load_causal_lm_model
```
larryliu0820, line 68: hm how does this work? I guess we don't want to depend on optimum-executorch from executorch? Or this is just for testing

metascroy: It requires you install optimum-executorch after installing executorch, like we do in our CI jobs.

This is primarily for testing (in examples folder), and meant to show we can export generic LLMs from optimum with good perf. Ideally MLX would be a recipe in optimum and this could be deleted, but MLX needs to be in executorch before I can add an MLX recipe to optimum.

CMakePresets.json
```diff
@@ -110,7 +110,7 @@
       "inherits": ["common"],
       "cacheVariables": {
         "EXECUTORCH_BUILD_PRESET_FILE": "${sourceDir}/tools/cmake/preset/pybind.cmake",
-        "CMAKE_OSX_DEPLOYMENT_TARGET": "12.0"
+        "CMAKE_OSX_DEPLOYMENT_TARGET": "14.0"
```
mergennachin, line 113: Upgrade is intentional? It's probably fine as 12 and 13 is EOL now.

kernels/optimized/patches/pocketfft_aligned_alloc.patch
```diff
@@ -0,0 +1,36 @@
+diff --git a/pocketfft_hdronly.h b/pocketfft_hdronly.h
```
mergennachin: Remove this now that you upstreamed?

exir/passes/remove_noop_pass.py
mergennachin: Approved that PR

backends/mlx/partitioner.py
```diff
@@ -0,0 +1,314 @@
...
+            return False
+
+        # Check if builder determined this node is supported
+        info = self._builder.node_info.get(node)
+        if info is not None and info.supported:
+            logger.debug(f"[SUPPORTED] Node {node.target}")
+            return True
+
+        logger.debug(f"[UNSUPPORTED] Node {node.target}")
+        return False
```
mergennachin, line 74: Why do you need to do MLXProgramBuilder yet again? Can we just call check_support_only?

backends/mlx/CMakeLists.txt
```diff
@@ -0,0 +1,311 @@
...
+
+# Add logging flag if enabled
+if(ET_MLX_ENABLE_OP_LOGGING)
+  target_compile_definitions(mlxdelegate PRIVATE ET_MLX_ENABLE_OP_LOGGING=1)
+  message(STATUS "MLX delegate op logging ENABLED")
+endif()
+
+target_include_directories(
+  mlxdelegate PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/runtime
+                      ${_mlx_schema__include_dir} ${MLX_SOURCE_DIR}
```
mergennachin: Is _mlx_schema__include_dir defined?

backends/mlx/serialization/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+            print(f"Error generating MLX code: {result.stderr}", file=sys.stderr)
+            raise RuntimeError(
+                f"Failed to generate MLX delegate code. "
+                f"Run 'python {generate_script}' manually to see the error."
+            )
+
+        print("MLX delegate: Code generation complete.", file=sys.stderr)
+
+
+# Auto-generate files if they don't exist
```
mergennachin: Why auto generate? This would be surprising for __init__.py load

backends/mlx/builder/program_builder.py
```diff
@@ -0,0 +1,1578 @@
...
+
+This module is responsible for:
+1. Walking the FX graph from an ExportedProgram
+2. Converting each node to the corresponding MLX op
+3. Managing tensor and value slots
+4. Building the final MLXGraph dataclass for serialization
+
+Op handlers are registered in ops.py.
+Pattern handlers are registered in patterns.py.
+"""
```
mergennachin, line 20: This file is too large, thoughts on splitting into multiple

backends/mlx/builder/program_builder.py
```diff
@@ -0,0 +1,1578 @@
...
+
+            try:
+                handler(self, n)
+                self._mark_supported(n, handler=handler)
+            except Exception as e:
+                trace_str = traceback.format_exc()
+                msg = f"{handler} failed for {n.target}: {e}.\n{trace_str}"
+                if n.meta.get("val", None) is not None:
+                    self.slot_manager.make_or_get_slots(n)
+                self._mark_unsupported(n, msg)
```
mergennachin, line 542: rethrow the caught exception?

backends/mlx/builder/program_builder.py
```diff
@@ -0,0 +1,1578 @@
...
+        )
+        slot_to_vid = {s: idx for idx, s in enumerate(slot_to_vid)}
+
+        # Remap all Tid/Vid values in instructions to use global indices
+        if hasattr(self, "_tid_slot_map"):
+            for tid, slot in self._tid_slot_map:
+                if slot in slot_to_tid:
+                    tid.idx = slot_to_tid[slot]
+                else:
+                    logger.warning(f"Slot {slot} not found in slot_to_tid mapping")
```
mergennachin, line 713: don't you wanna error out loudly?

backends/mlx/builder/program_builder.py
```diff
@@ -0,0 +1,1578 @@
...
+
+        if hasattr(self, "_vid_slot_map"):
+            for vid, slot in self._vid_slot_map:
+                if slot in slot_to_vid:
+                    vid.idx = slot_to_vid[slot]
+                else:
+                    logger.warning(f"Slot {slot} not found in slot_to_vid mapping")
```
mergennachin, line 720: .

backends/mlx/runtime/MLXBackend.cpp
```diff
@@ -0,0 +1,372 @@
...
+// MLX is not thread-safe: its computation graph is global shared state.
+// A global mutex serializes graph construction and command submission
+// across all handles. GPU execution and output copies can proceed
+// without the lock (see execute() for the async pipeline design).
+static std::mutex& mlx_global_mutex() {
+  static std::mutex m;
+  return m;
+}
+
+class MLXBackend final : public ::executorch::runtime::BackendInterface {
```
mergennachin, line 192: You should probably check the MLX schema version in the delegate blog?

backends/mlx/program_builder.py
```diff
@@ -0,0 +1,1578 @@
...
+        return self.free.pop() if self.free else self._bump()
+
+    def _bump(self):
+        idx = self.next_new_id
+        self.next_new_id += 1
+        return idx
+
+    def return_id(self, idx):
+        if self.free and self.free[-1] == idx:
+            return
```
mergennachin: why only check for the last entry instead of the whole list? also, should you change to a set instead of list, if you change to checking all entry

backends/mlx/patterns.py
```diff
@@ -0,0 +1,936 @@
...
+                update=P.slot_to_tid(rest_chunk_slot),
+                axis=IntOrVid.from_literal(2),
+                start=IntOrVid.from_literal(0),
+                stop=P.to_int_or_vid(overflow_slot),
+            )
+        )
+
+
+@REGISTRY.register_pattern(name="SDPA")
+class SDPAHandler(PatternHandler):
```
mergennachin, line 393: Would it find Metal SDPA? 

https://github.com/pytorch/executorch/pull/17720

backends/mlx/passes.py
```diff
@@ -0,0 +1,696 @@
...
+                modified = True
+
+        if modified:
+            graph.eliminate_dead_code()
+            graph.lint()
+
+        return PassResult(graph_module, modified)
+
+
+class CSEPass(ExportPass):
```
mergennachin: Consider reusing CSEPass from torch fx 

See an example here: 
https://github.com/pytorch/executorch/pull/15813

backends/mlx/patches/mlx_json.patch
```diff
@@ -0,0 +1,29 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
mergennachin, line 1: Can we do without having any patches?

backends/mlx/runtime/MLXExecutor.h
```diff
@@ -0,0 +1,790 @@
...
+//   program: The loaded MLXProgram containing tensor metadata and named_slots
+//   named_data_map: ExecuTorch's interface for accessing named data
+//   store: Output storage for loaded constant tensors
+//   constant_buffers: Vector to store FreeableBuffers (must outlive store for
+//   zero-copy)
+inline void load_constants(
+    const MLXProgram& program,
+    const runtime::NamedDataMap* named_data_map,
+    ConstantData& store,
+    std::vector<runtime::FreeableBuffer>& constant_buffers) {
```
mergennachin, line 715: instead of linear scans for each constant, thoughts on having a map datastructure?

backends/mlx/runtime/MLXExecutor.h
```diff
@@ -0,0 +1,790 @@
...
+      throw std::runtime_error(
+          "tensor_index: called with constant tensor id " +
+          std::to_string(id.idx));
+    }
+    uint32_t idx = id.idx - num_constants;
+    // If this ID is after mutable buffer range, subtract mutable buffer count
+    if (id.idx >= mutable_buffer_end) {
+      idx -= program->num_mutable_buffer_tensors;
+    }
+    return idx;
```
mergennachin, line 240: id.idx - num_constants and conditional subtraction of num_mutable_buffer_tensors. These are unsigned arithmetic. If Tid values from a malformed payload are larger than expected, underflow could produce valid-looking indices

backends/mlx/serialization/MLXLoader.cpp.tmpl
```diff
@@ -0,0 +1,188 @@
...
+// =============================================================================
+
+MLXProgram load_program(const void* data, size_t size) {
+  MLXHeader header;
+  if (!parse_header(data, size, header)) {
+    throw std::runtime_error("Invalid MLX header");
+  }
+
+  const uint8_t* fb_data = static_cast<const uint8_t*>(data) + kHeaderSize;
+  size_t fb_size = header.data_offset - kHeaderSize;
```
mergennachin, line 105: If data_offset < kHeaderSize, this underflows.

metascroy:
I've rebased onto main now that all "prerequisite" PRs have landed.

mergennachin (changes requested):
- In the runtime unittest, can we compile with -Wall, -Werror, and -Wconversion, -Wsign-conversion, -Wshorten-64-to-32, asan, ubsan flag to uncover any security issues

-  Use overflow-safe arithmetic for all bounds checks. Use __builtin_add_overflow / __builtin_mul_overflow or c10's safe arithmetic utilities.

- Check for null pointer deferences

backends/mlx/CMakeLists.txt
```diff
@@ -0,0 +1,311 @@
+#
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+#
+
+cmake_minimum_required(VERSION 3.19)
+
```
mergennachin, line 10: Consider adding -Wall -Werror to the MLX backend

backends/mlx/examples/llm/README.md
```diff
@@ -0,0 +1,100 @@
...
+|--------|---------|-------------|
+| `--model-id` | `unsloth/Llama-3.2-1B-Instruct` | HuggingFace model ID |
+| `--output` | *(required)* | Output .pte file path |
+| `--max-seq-len` | `1024` | Maximum sequence length for KV cache |
+| `--dtype` | `bf16` | Model dtype (`fp32`, `fp16`, `bf16`) |
+| `--quantize-linear` | None | Quantization for linear layers (`int4`, `int8`) |
+| `--quantize-embeddings` | None | Quantization for embedding layers (`int4`, `int8`) |
+| `--no-tie-word-embeddings` | `False` | Disable re-tying lm_head to embedding after quantization |
+| `--use-custom-sdpa` | `False` | Use MLX custom SDPA (`mlx::custom_sdpa`) |
+| `--use-custom-kv-cache` | `False` | Use MLX custom KV cache (`mlx::kv_cache_update`) |
```
JacobSzwejbka, line 69: we have inplace index_put_ I believe. Could you just use that or do you need slice_scatter_

backends/mlx/examples/llm/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+---
+
+## `run_llm_hf`
+
+Run models exported with `export_llm_hf`. Supports both full-prompt prefill (dynamic seq len exports) and token-by-token prefill (fixed seq len exports).
```
JacobSzwejbka, line 75: why bother supporting fixed if dynamic works? Is there a big perf trade off in which case the reverse question.

backends/mlx/pte_inspector.py
```diff
@@ -0,0 +1,1114 @@
...
+
+    except ImportError:
+        result["program_parse_error"] = "ExecuTorch FlatBuffer parsing not available"
+    except Exception as e:
+        result["program_parse_error"] = str(e)
+
+    return result
+
+
+def extract_delegate_payload(  # noqa: C901
```
JacobSzwejbka: can you put the pte stuff in exir? Though some things like the PTE header already exist I thought

backends/mlx/README.md
```diff
@@ -0,0 +1,493 @@
...
+  primitives.
+
+> **Adding a new op?** Jump to [How to Add a New Op](#how-to-add-a-new-op).
+
+## Getting Started
+
+The MLX delegate requires **Apple Silicon** (M1 or later) and the **Metal
+compiler**, which ships with Xcode (not the standalone Command Line Tools).
+
+**Check if Metal is available:**
```
JacobSzwejbka, line 18: Do we do this automatically if someone tries to use the partitioner. Do we surface the recommendations below if they do that and dont have the metal compiler installed?

backends/mlx/README.md
```diff
@@ -0,0 +1,493 @@
...
+│   ├── mlx_graph_schema.py     # [GENERATED] Python dataclasses for IR nodes
+│   ├── mlx_graph_serialize.py  # Serialization to FlatBuffer binary
+│   ├── _generated_serializers.py # [GENERATED] Per-op FlatBuffer builders
+│   └── _generated/             # [GENERATED] FlatBuffer Python bindings (flatc)
+├── runtime/                    # C++ runtime (loaded at inference time)
+│   ├── MLXBackend.cpp          # BackendInterface (init / execute / destroy)
+│   ├── MLXLoader.h/.cpp        # [GENERATED] FlatBuffer → C++ structs
+│   ├── MLXExecutor.h           # ExecutionState, constant loading, helpers
+│   ├── MLXInterpreter.h        # Op dispatch loop + per-op exec_* functions
+│   └── schema_generated.h      # [GENERATED] FlatBuffer C++ bindings (flatc)
```
JacobSzwejbka, line 109: why check this in?

backends/mlx/serialization/schema.fbs
```diff
@@ -0,0 +1,995 @@
...
+}
+
+// Value slot identifier - indexes into values array
+// Values are stored as variant<int64, double, bool> at runtime
+struct Vid {
+    idx: uint32;
+}
+
+// For fields that can be either a literal int or a runtime Vid
+table IntOrVid {
```
JacobSzwejbka, line 41: Why not just use a union type?

backends/mlx/serialization/schema.fbs
```diff
@@ -0,0 +1,995 @@
...
+    out: Tid (required);
+    dims: int32;
+    offset: TidOrVid (required);  // Position offset: scalar (Vid) or tensor of positions (Tid)
+    freqs: Tid;  // optional
+    traditional: bool = false;
+    base: float = 500000.0;  // Llama 3 default
+    scale: float = 1.0;
+}
+
+table SdpaNode {
```
JacobSzwejbka, line 148: Whats the story for out of tree custom ops? Just not possible?

backends/mlx/serialization/schema.fbs
```diff
@@ -0,0 +1,995 @@
...
+    q: Tid (required);
+    k: Tid (required);
+    v: Tid (required);
+    out: Tid (required);
+    scale: float;
+    mask: Tid;  // optional
+    causal: bool = false;
+}
+
+table AddNode {
```
JacobSzwejbka, line 158: Similarly what happens if MLX changes the schema of any of these operators in a BC breaking way, or even adds a new optional arg with default values. Seems like there would be a lot of op versioning issues by explicitly encoding every ops schema in the serialization format. We also have a lot less influence on their op versioning story compared with ATen (where its already a big pita).

backends/mlx/ops.py
```diff
@@ -0,0 +1,4061 @@
...
+            P.emit(
+                ContiguousNode(
+                    x=P.slot_to_tid(x),
+                    out=P.slot_to_tid(out),
+                )
+            )
+        return out
+
+except ImportError:
+    # Edge IR ops not available (e.g., when building from ATen dialect)
```
JacobSzwejbka: When are you doing this?

backends/mlx/patterns.py
```diff
@@ -0,0 +1,936 @@
...
+                )
+            )
+            return existing_slot
+        else:
+            P.set_slot(n, dst)
+            return dst
+
+
+@REGISTRY.register_pattern(name="ET_KV_CACHE_UPDATE")
+class ETKVCacheUpdateHandler(PatternHandler):
```
JacobSzwejbka, line 132: Why is both a custom mutator op supported, but also the slice_update ops earlier?

backends/mlx/serialization/schema.fbs
metascroy: AddNode is an abstraction in the MLX delegate that we control (not part of MLX itself).

The runtime takes AddNode and translates it to an MLX call (add op) by parsing the args. So even if mlx changes the signature of their add op, it doesn't mean we break AddNode in the graph. We would just update our runtime appropriately. (With that said, MLX did say they try to maintain BC of their underlying ops).

Maybe I'm missing the concern here, though?

backends/mlx/pattern_utils.py
```diff
@@ -0,0 +1,360 @@
...
+            return None
+
+    if not isinstance(cur, Node):
+        return None
+
+    return cur, entries
+
+
+@dataclass
+class PatternMatch:
```
JacobSzwejbka, line 299: XNNPack has their own pattern matching util iirc. Could we just offer this in Exir cc @GregoryComer

CMakePresets.json
metascroy: Yes, it was intentional. It also makes our iOS default (17) inline with our macos one (14).

backends/mlx/examples/llm/README.md
metascroy: I support both index_put_ (running on MLX, not ET) and mlx::kv_cache_update.

mlx::kv_cache_update is slightly more efficient, though.

backends/mlx/examples/llm/README.md
metascroy: It's because optimum currently exports gemma with static shapes if you don't use custom sdpa/kv-cache.

So if you pass the custom args, you'll get a model with dynamic shape, and it will use that, but if you don't use custom args, optimum will give you a static shape model (and without this logic, it doesn't run)

backends/mlx/README.md
metascroy: It's automatically checked when installing pybindings. If not found, there is a warning about MLX not being installed because metal compiler is not available, asking users to install xcode if they want MLX.

backends/mlx/README.md
metascroy: The generated files are not checked in. This is the directory structure after running the code generator.

backends/mlx/patches/mlx_json.patch
metascroy: I'll see.

backends/mlx/serialization/__init__.py
metascroy: I'll move to install process

backends/mlx/serialization/schema.fbs
metascroy: FlatBuffer unions can only contain tables, not scalar primitives. Using a union here would require wrapping the int64 literal in a table (table IntLiteral { value: int64; }), adding pointer indirection and a table allocation for every shape dimension. The is_vid bool keeps the common case (literal) as a direct inline field read.

backends/mlx/serialization/schema.fbs
metascroy: What do you mean by out of tree custom ops?

From MLX perspective, they allow custom ops using raw shader code. I guess we could add a node for that to the schema.

backends/mlx/ops.py
metascroy: I'm not. I'll remove this

backends/mlx/partitioner.py
metascroy: Yes, we can just call check_support_only on the second call as well. I'll change it

backends/mlx/pattern_utils.py
metascroy: Happy to reuse any pre-existing utilities if they're in exir.

If the XNNPACK one is based on the torch.fx pattern finding util, I don't know if it'll work, though. I recall it being very bad at fuzzy matching of args/kwargs (like you needed exact matches, so you have to write like 3-4 pattern matchers for each op, and it's a bit error prone)

backends/mlx/patterns.py
metascroy: I guess you mean the difference between the IndexUpdateNode pattern above and this one?

They are similar, but ET_KV_CACHE_UPDATE has slightly better perf and is tied to a KVCache building block in mlx/llm/cache.py.

In terms of the difference:
* INDEX_UPDATE matches a functionalized copy_ pattern (used in many HF models) and maps it onto IndexUpdateNode, which is an in-place update op in the MLX backend. IndexUpdateNode works by finding contiguous ranges in indices and repeatedly calling slice update. So in practice it will be 1 slice update call for regular kv cache, but up to 2 calls for ring buffer stuff, but it does require the indices be materialized in order to parse the index ranges (which can slow down MLX computation).

* ET_KV_CACHE_UPDATE emits as one slice update call, and does not require materialization of the indices tensor.

At this point, the benefit is maybe 10-15% extra perf.

backends/mlx/patterns.py
metascroy: Metal SDPA is just standard sdpa with repeat interleave in front?

If so, then yes, it will find it and remove the repeat interleave. It also recognizes the analogous pattern in HF transformers (which don't use repeat interleave to express GQA).

The underlying MLX kernel supports GQA without the expansion (and I think Metal copied the MLX one?), so this pass is to just auto recognize this without a model re-write.

backends/mlx/builder/program_builder.py
metascroy: This is intentional.

Just having a op/pattern handler isn't enough for the partitioner to mark support. The op/pattern handlers are allowed to throw exceptions when processing a node if they encounter an unsupported argument or something they cannot handle, and this gets bubbled up to the partitioner as an unsupported node (with the unsupported reason), and falls back to portable.

If we rethrow, then lowering would fail (CoreML's current behavior) instead of falling back to portable.

backends/mlx/runtime/MLXExecutor.h
metascroy: Added check id.idx < num_constants

backends/mlx/runtime/MLXExecutor.h
metascroy: Added

backends/mlx/serialization/MLXLoader.cpp.tmpl
metascroy: Added check

backends/mlx/CMakeLists.txt
metascroy: Added -Wall -Werror -Wno-deprecated-declarations

backends/mlx/examples/voxtral/export_voxtral_hf.py
manuelcandales, line 1: I think you should add an MLX recipe in optimum-executorch [here](https://github.com/huggingface/optimum-executorch/tree/main/optimum/exporters/executorch/recipes), instead of adding this file

metascroy: I will do that as follow-up after this lands. I need MLX to land before editing optimum

backends/mlx/examples/voxtral/export_voxtral_hf.py
```diff
@@ -0,0 +1,264 @@
...
+    from executorch.backends.mlx.passes import get_default_passes
+    from executorch.exir import EdgeCompileConfig
+    from executorch.exir.capture._config import ExecutorchBackendConfig
+    from executorch.exir.passes import MemoryPlanningPass
+    from executorch.extension.audio.mel_spectrogram import WhisperAudioProcessor
+    from torch.export import Dim
+
+    logger.info("Exporting audio preprocessor with MLX backend...")
+
+    model = WhisperAudioProcessor(
```
manuelcandales, line 63: Rename if this is used in models other than Whisper (like here in Voxtral)

backends/mlx/examples/voxtral/export_voxtral_hf.py
manuelcandales: Ah, true, I see you answered this earlier: "MLX needs to be in executorch before I can add an MLX recipe to optimum."

backends/mlx/serialization/schema.fbs
metascroy: The more important reason is flatbutter doesn't allow vector of union

metascroy:
> * In the runtime unittest, can we compile with -Wall, -Werror, and -Wconversion, -Wsign-conversion, -Wshorten-64-to-32, asan, ubsan flag to uncover any security issues
> * Use overflow-safe arithmetic for all bounds checks. Use __builtin_add_overflow / __builtin_mul_overflow or c10's safe arithmetic utilities.
> * Check for null pointer deferences

I added a strict compile test, addressed security issues, and addressed null ptr references.

metascroy:
Continuing development here: https://github.com/pytorch/executorch/pull/17803

PR #16725, opened 2026-01-21
cortex-m dim order

lucylq:
### Summary
https://github.com/pytorch/executorch/pull/16636 preserves the memory layout of tensors that are channels-last - previously, they were being converted to .contiguous(). 

test-mcu-cortex-m-backend test is failing after that PR. So potentially cortex-m requires that the tensors be contiguous after the permute, and not contiguous with memory_format=torch.channels_last? 

This PR calls .contiguous() in `convert_to_cortex_m_pass.py`, instead of .contiguous(memory_format=torch.channels_last).

### Test plan
CI test [test-mcu-cortex-m-backend / linux-job](https://github.com/pytorch/executorch/actions/runs/21184785023/job/60936271927#logs)

### DIFF
--- a/backends/cortex_m/passes/convert_to_cortex_m_pass.py
+++ b/backends/cortex_m/passes/convert_to_cortex_m_pass.py
@@ -199,14 +199,16 @@ class ConvertToCortexMPass(XNNPACKPass):
         if use_depthwise_conv:
             # For depthwise: OIHW -> IHWO which gives [1, H, W, C_OUT] for CMSIS-NN
             # PyTorch depthwise weight is [out_ch, 1, H, W], permute to [1, H, W, out_ch]
-            weight_permuted = weight_tensor.permute(1, 2, 3, 0).contiguous(
-                memory_format=torch.channels_last
-            )
+            # The permute achieves the desired logical layout (IHWO). CMSIS-NN expects
+            # weights in physically contiguous memory after the permute (not in channels-last)
+            # so we use contiguous() here.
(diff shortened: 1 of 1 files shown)


AdrianLundell (approved):

backends/cortex_m/passes/convert_to_cortex_m_pass.py
```diff
@@ -199,14 +199,10 @@ def _get_convolution_replacement(self, node) -> int:
         if use_depthwise_conv:
             # For depthwise: OIHW -> IHWO which gives [1, H, W, C_OUT] for CMSIS-NN
             # PyTorch depthwise weight is [out_ch, 1, H, W], permute to [1, H, W, out_ch]
-            weight_permuted = weight_tensor.permute(1, 2, 3, 0).contiguous(
-                memory_format=torch.channels_last
-            )
+            weight_permuted = weight_tensor.permute(1, 2, 3, 0).contiguous()
```
psiddh, line 205: nits: Adding a clarifying comment explaining why plain .contiguous() is used
Something like this ?
"CMSIS-NN expects weights in physically contiguous memory after permute.
Do NOT use memory_format=torch.channels_last here - the permute already achieved the desired logical layout (OHWI for regular conv, IHWO for depthwise)"

As a follow up PR, (I can take this AI)
- Add explicit unit test for weight tensor layout verification

psiddh (approved):

rascani (approved):

PR #16727, opened 2026-01-21
Use fp32 accumulation in portable/optimized norms

GregoryComer:
Summary:
When running in fp16, our layer norm implementation diverges from ATen, especially for larger variances. This is because ATen internally accumulates mean and variance in fp32 for fp16/bf16, whereas we accumulate in the input/output dtype.

To resolve this, I've matched the ATen behavior by using acc_t (similar to OpMathType from ATen) in optimized. In portable, I've just removed the cast back to CTYPE, as the accumulation logic was already in float.

There was also a straight up bug where the element count was being cast to CTYPE in group norm, which tends to overflow or lose precision.

I've added an additional test case with high variance. In combination with the existing coverage, this should show that the behavior is unchanged for existing cases and no longer overflows or diverges for larger inputs.

Thanks Jason Zhu for spotting this!

Differential Revision: D91096147


kernels/optimized/utils/math_utils.h
```diff
@@ -37,6 +39,15 @@ template <>
 struct ComputeDTypeTraits<int8_t> {
   using type = int32_t;
 };
+// For 16 bit float types, ops should perform internal math in float32.
```
GregoryComer, line 42: This matches the ATen convention from [OpMathType.h](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/OpMathType.h).

kernels/portable/cpu/op_native_group_norm.cpp
```diff
@@ -77,8 +77,8 @@ void group_norm(
     const CTYPE* x = input_data + i * inner_size;
 
     // compute E[X] and Var[x] = E[x^2] - E[x]^2
-    CTYPE sum = reduce_add(x, static_cast<CTYPE>(inner_size));
-    CTYPE sq_sum = vec_powerf(x, static_cast<CTYPE>(inner_size));
+    float sum = reduce_add(x, inner_size);
```
GregoryComer, line 80: Note: We should ideally use float64 if CTYPE is double, but we're already forcing float32 in the various utility functions - reduce_add and vec_powerf return floats and use floats internally. I'm just removing the implicit cast to CTYPE on the return value.

I'm inclined to leave this for a follow-up on an as-needed basis.

GregoryComer:
Note that unit tests are failing on trunk. Checking the job logs on this PR shows it's an unrelated failure.

larryliu0820 (approved):

mergennachin (approved):

PR #16735, opened 2026-01-21
tutorial for devtool debugibility

Gasoonjia:
Summary: This diff introduces a tutorial for executorch devtool debugibility

Reviewed By: GregoryComer

Differential Revision: D90809255

### DIFF
--- a/docs/source/tutorials_source/devtools-debugging-tutorial.py
+++ b/docs/source/tutorials_source/devtools-debugging-tutorial.py
@@ -1,0 +1,487 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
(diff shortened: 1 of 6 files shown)


docs/source/tutorials_source/devtools-debugging-tutorial.py
```diff
@@ -0,0 +1,482 @@
...
+import tempfile
+
+import torch
+
+from executorch.backends.xnnpack.partition.xnnpack_partitioner import XnnpackPartitioner
+from executorch.backends.xnnpack.utils.configs import get_xnnpack_edge_compile_config
+
+from executorch.exir import ExecutorchProgramManager, to_edge_transform_and_lower
+from torch.export import export, ExportedProgram
+from torchvision import models
```
psiddh: Looks like this lint failure needs to be addressed

Gasoonjia: that's a wierd lint issue cuz; `models` is used by the code indeed.
Let me try to suppress it

psiddh (approved):
Approving it to unblock, (there are lint errors to be addressed)

PR #16740, opened 2026-01-21
Arm backend: Initial bfloat16 support.

Erik-Lundell:
Add the necessary infrastructure and add support for some initial ops. Currently, only the TOSA backend supports the dtype.


zingo:
Rebasing to verify that the cortex-m fails goes away

backends/arm/test/runner_utils.py
```diff
@@ -60,7 +60,7 @@
     torch.float16: np.float16,
     torch.float32: np.float32,
     torch.float64: np.float64,
-    torch.bfloat16: np.float32,
+    torch.bfloat16: np.uint16,
```
zingo, line 63: Is this to make sure it fits in 16bit? And actually fixes a bug?

zingo: Spotted the code later in the file!

zingo (approved):
OK to merge!

backends/arm/test/runner_utils.py
Erik-Lundell: I hope bfloat16 fits in 16 bits!

PR #16763, opened 2026-01-22
solve false negtive gap for multiple output operators (#16729)

Gasoonjia:
Summary:

When comparing AOT intermediate outputs with runtime, we believed that AOT and runtime should have same output for same operator. But if there're multiple intermediate outputs from a single operator / single operator blob, the statement may not correct. Like drop out, which only record output tensor during AOT, but in runtime we record both mask and output tensor.

To support that, for 1 to many scenerio, instead of only take the last element for comparsion, we compare the runtime output sharing the same size and dtype with the aot one to have the best comparsion.

Differential Revision: D91201882

### DIFF
--- a/devtools/inspector/_inspector_utils.py
+++ b/devtools/inspector/_inspector_utils.py
@@ -751,6 +751,168 @@ def _find_connected_components(
     return connected_components
 
 
+def _map_sequence_aot_output(
+    aot_intermediate_output: Sequence,
+    runtime_intermediate_output: Any,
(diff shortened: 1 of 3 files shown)


GregoryComer (approved):
Can you make sure the Windows test passes before landing? Approving to unblock.

Gasoonjia:
> Can you make sure the Windows test passes before landing? Approving to unblock.

will take a look that, but i believe that's not related to this PR.

PR #16768, opened 2026-01-22
Arm backend: Ignore _debug_passes.py for code coverage

martinlsm:
_debug_passes.py does not contain production code and is therefore not relevant to measure as a part of overall test coverage. For this reason, ignore this file for code coverage reports.


zingo (approved):
OK to merge

PR #16773, opened 2026-01-22
Cortex-M backend: Introduce NodeFinders

AdrianLundell:
NodeFinders is one step in generalizing the OperatorConfigQuantizer and making the CortexMQuantizer more flexible, by separating the selection of nodes/QuantizationConfig applied by one single OperatorConfigQuantizer from backend specific selection logic.

This aims to allow better configuration of the quantization + reuse of a similar quantizer structure in other backends.


AdrianLundell:
@psiddh Any comments on this?

rascani (approved):

PR #16777, opened 2026-01-22
Arm backend: Improve pre-push hook checks

martinlsm:
Update the pre-push hook:
- Only check if the 'Arm backend: ' prefix is present in case all file
  modifications are in backends/arm or examples/arm; otherwise, skip the
  check. However, missing this prefix when it should be in place is now
  a fatal error for the pre-push hook.
- The imperative mood check is still in place but it is now able to
  check patches no matter if they have the 'Arm backend: ' or not.
  Effectively it is now independent from the prefix check. When the
  imperative mood check fails, it is still just a warning.
- Add a non-zero exit code (1) in case parse_test_names.py finds any
  test name violations. Otherwise the pre-push hook will not fail in
  case the script finds any violations.


zingo (approved):
OK to merge

PR #16790, opened 2026-01-22
[ET-VK][ez] Add support for auto_functionalized_v2

SS-JIA:
Summary:
Update the Vulkan partitioner to recognize torch.ops.higher_order.auto_functionalized_v2
in addition to the existing auto_functionalized op when determining operator support
and extracting target names for custom operators.


SS-JIA:
Fixes https://github.com/pytorch/executorch/issues/16647

kirklandsign (approved):

PR #16809, opened 2026-01-23
Remove aoti_cuda/metal_blob.ptd file from being generated for AOTI

larryliu0820:
As titled

### DIFF
--- a/backends/aoti/aoti_backend.py
+++ b/backends/aoti/aoti_backend.py
@@ -75,6 +75,19 @@ class AotiBackend(ABC):
         pass
 
     @classmethod
+    def save_data_externally(cls) -> bool:
+        """
+        Return whether to save the named data map to an external .ptd file.
+
+        If True, the SO blob and weights blob will be saved to a separate .ptd file
+        (e.g., aoti_cuda_blob.ptd) that must be provided at runtime.
+        If False, the data will be merged into the .pte file.
+
+        Default is False (merge with .pte file). Subclasses can override this.
+        """
+        return False
+
+    @classmethod
     def get_extra_aoti_compile_context_manager(cls):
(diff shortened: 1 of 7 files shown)


backends/aoti/aoti_backend.py
```diff
@@ -227,10 +227,7 @@ def preprocess(
 
         # Add SO and weights blob separately
         named_data_store.add_named_data(method_name + "_so_blob", so_data, 1, None)
-        weights_blob_data_type = f"aoti_{device_name}_blob"
-        named_data_store.add_named_data(
-            method_name + "_weights_blob", blob_data, 1, weights_blob_data_type
-        )
+        named_data_store.add_named_data(method_name + "_weights_blob", blob_data, 1)
```
lucylq: is this being stored inside the pte file now?

lucylq (approved):

Gasoonjia:
i thought we only focus on getting rid of aoti_metal_blob.ptd.

why do we want to remove aoti_cuda_blob.ptd? I remember the reason we introduce that is for running across differnet gpus: each gpu will have one ptd file. 

Is that still the goal?

backends/aoti/aoti_backend.py
larryliu0820: yeah. I changed this to only work for Metal. so no `aoti_metal_blob.ptd` after my change, but only `aoti_cuda_blob.ptd`.

larryliu0820:
> i thought we only focus on getting rid of aoti_metal_blob.ptd.
> 
> why do we want to remove aoti_cuda_blob.ptd? I remember the reason we introduce that is for running across differnet gpus: each gpu will have one ptd file.
> 
> Is that still the goal?

we never need multiple ptd file. If anything it should be multiple graphs sharing the same weight. But yeah keeping the file for cuda right now.

PR #16821, opened 2026-01-23
Add bfloat16 support for Parakeet

larryliu0820:
Enable both export script and runner to support bfloat16. Changing CI to run on bfloat16 by default.

Also added a `parakeet-cuda-debug` mode

### DIFF
--- a/examples/models/parakeet/export_parakeet_tdt.py
+++ b/examples/models/parakeet/export_parakeet_tdt.py
@@ -295,11 +295,15 @@ class PreprocessorWrapper(torch.nn.Module):
         return mel, mel_len
 
 
-def export_all(model):
+def export_all(model, dtype=torch.float):
     """Export all model components.
 
     The maximum audio duration is determined by the model's internal
     max_audio_length (~50 seconds for Parakeet with max_audio_length=5000).
+
+    Args:
+        model: The NeMo ASR model to export.
+        dtype: Data type for floating-point tensors (default: torch.float).
     """
     programs = {}
 
@@ -316,7 +320,8 @@ def export_all(model):
 
     preprocessor_wrapper = PreprocessorWrapper(model.preprocessor)
     preprocessor_wrapper.eval()
-    sample_audio = torch.randn(max_audio_samples)
+
+    sample_audio = torch.randn(max_audio_samples, dtype=torch.float)
     sample_length = torch.tensor([sample_audio.shape[0]], dtype=torch.int64)
     # The preprocessor uses different code paths when CUDA is available, which include
     # data-dependent conditionals that torch.export cannot handle. Force CPU path.
@@ -337,7 +342,7 @@ def export_all(model):
     feat_in = getattr(model.encoder, "_feat_in", 128)
     # Use max_mel_frames as example to ensure Dim.AUTO infers the full range.
     # Smaller examples cause Dim.AUTO to infer narrow bounds.
-    audio_signal = torch.randn(1, feat_in, max_mel_frames)
+    audio_signal = torch.randn(1, feat_in, max_mel_frames, dtype=dtype)
     length = torch.tensor([max_mel_frames], dtype=torch.int64)
     encoder_with_proj = EncoderWithProjection(model.encoder, model.joint)
     encoder_with_proj.eval()
@@ -359,8 +364,8 @@ def export_all(model):
     decoder_step = DecoderStep(model.decoder, model.joint)
     decoder_step.eval()
     token = torch.tensor([[0]], dtype=torch.long)
-    h = torch.zeros(num_layers, 1, pred_hidden)
-    c = torch.zeros(num_layers, 1, pred_hidden)
+    h = torch.zeros(num_layers, 1, pred_hidden, dtype=dtype)
+    c = torch.zeros(num_layers, 1, pred_hidden, dtype=dtype)
     programs["decoder_step"] = export(
         decoder_step,
         (token, h, c),
@@ -371,8 +376,8 @@ def export_all(model):
     joint_hidden = model.joint.joint_hidden
     num_token_classes = model.tokenizer.vocab_size + 1  # +1 for blank
 
-    f_proj = torch.randn(1, 1, joint_hidden)
-    g_proj = torch.randn(1, 1, joint_hidden)
+    f_proj = torch.randn(1, 1, joint_hidden, dtype=dtype)
+    g_proj = torch.randn(1, 1, joint_hidden, dtype=dtype)
     programs["joint"] = export(
         JointWithArgmax(model.joint, num_token_classes),
         (f_proj, g_proj),
@@ -551,9 +556,9 @@ def main():
     )
     args = parser.parse_args()
 
-    # Validate dtype for Metal backend
-    if args.backend == "metal" and args.dtype == "fp16":
-        parser.error("Metal backend only supports fp32 and bf16, not fp16")
+    # Validate dtype
+    if args.dtype == "fp16":
+        parser.error("fp16 is not yet supported")
 
     os.makedirs(args.output_dir, exist_ok=True)
 
@@ -572,7 +577,8 @@ def main():
         model = model.to(torch.float16)
 
     print("\nExporting components...")
-    programs, metadata = export_all(model)
+    export_dtype = torch.bfloat16 if args.dtype == "bf16" else torch.float
+    programs, metadata = export_all(model, dtype=export_dtype)
 
     et = lower_to_executorch(programs, metadata=metadata, backend=args.backend)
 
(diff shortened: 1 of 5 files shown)


examples/models/parakeet/main.cpp
```diff
@@ -118,27 +153,49 @@ std::vector<Token> greedy_decode_executorch(
...
+  // Shape: [1, T, joint_hidden]
+  size_t proj_dim = static_cast<size_t>(f_proj.sizes()[2]);
 
-  // Initialize LSTM state
-  std::vector<float> h_data(num_rnn_layers * 1 * pred_hidden, 0.0f);
-  std::vector<float> c_data(num_rnn_layers * 1 * pred_hidden, 0.0f);
+  // Get expected dtype for decoder_step h and c inputs (indices 1 and 2)
+  auto h_dtype_result = get_input_scalar_type(model, "decoder_step", 1);
+  if (!h_dtype_result.ok()) {
+    return hypothesis;
```
Gasoonjia, line 162: Why should we return hypothesis here? If there's any issue happen during get_input_scalar_type, maybe we should return error?

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -572,7 +577,12 @@ def main():
         model = model.to(torch.float16)
 
     print("\nExporting components...")
-    programs, metadata = export_all(model)
+    export_dtype = (
+        torch.bfloat16
+        if args.dtype == "bf16"
+        else torch.float16 if args.dtype == "fp16" else torch.float
```
Gasoonjia: have you checked fp16? I don't think cuda backend can work on fp16 now.

examples/models/parakeet/main.cpp
```diff
@@ -108,6 +110,39 @@ TimestampOutputMode parse_timestamp_output_mode(const std::string& raw_arg) {
...
+        Error,
+        "Failed to get input tensor metadata for %s[%zu]",
+        method_name,
+        input_index);
+    return input_meta_result.error();
+  }
+  return input_meta_result.get().scalar_type();
+}
+
 std::vector<Token> greedy_decode_executorch(
```
Gasoonjia, line 146: im a little bit lost in the logic here: based on my understanding, greedy_decode_executorch should only for joint and decoder, but not encoder. and encoder takes preprocessor's output as input, which is always fp32 now. Since we exported encoder as target dtype (e.g. bf16), how do we makes bf16 encoder consume fp32 preprocessor's outpu?

Gasoonjia (commented):
Thx man for such efficient support!

examples/models/parakeet/main.cpp
larryliu0820: The rest of the function is returning hypothesis when there's error so I'm trying to follow the pattern. Also I logged the error message when the result is not ok.

examples/models/parakeet/export_parakeet_tdt.py
larryliu0820: Haven't checked fp16

examples/models/parakeet/main.cpp
larryliu0820: > and encoder takes preprocessor's output as input, which is always fp32 now

Not true, if we pass dtype == bfloat16, preprocessor will take in a float and gives bfloat16 result

mergennachin:
@manuelcandales @larryliu0820 it would be good to test bf16 for metal too

examples/models/parakeet/export_parakeet_tdt.py
Gasoonjia: maybe we should raise error when user gives cuda + fp16?

examples/models/parakeet/main.cpp
Gasoonjia: hmm i'm thinking rest of function is returning hypothesis because of there's no error being raised.
Maybe we can apply `ET_CHECK_MSG ` here?

examples/models/parakeet/main.cpp
Gasoonjia: hmm not sure if i misunderstood something,  but export script says:
```# Preprocessor always uses float32 - runner converts output to encoder's dtype```
but i didn't find the code in runner for the type conversion.
if preprocessor will give bfloat16 output directly:
1.  perhapes updating the export doc?
2. how does the preprocessor know that? Seems the export configs for preprocessor are the same across different dtype.

Comment (larryliu0820) [reply examples/models/parakeet/main.cpp:146]: Let me dive a bit deeper into the model. I can take a look at the graph to see if it changes the dtype of the output

File: examples/models/parakeet/main.cpp (hunk 1 above)
Comment (Gasoonjia) [reply examples/models/parakeet/main.cpp:162]: after sync offline we have use this pattern for error out in the other places in the runner (e.g. https://github.com/pytorch/executorch/blob/74ea5848ad8bde7d7867be445291e6722114d89c/examples/models/parakeet/main.cpp#L154-L157) i believe we can keep the current style -- maybe another pr for better error out.

File: examples/models/parakeet/main.cpp (hunk 3 above)
Comment (larryliu0820) [reply examples/models/parakeet/main.cpp:146]: Graph:
```
graph():
    %b_preprocessor_dtype_sentinel_tensor : [num_users=0] = placeholder[target=b_preprocessor_dtype_sentinel_tensor]
    %b_preprocessor_featurizer_window : [num_users=1] = placeholder[target=b_preprocessor_featurizer_window]
    %b_preprocessor_featurizer_fb : [num_users=1] = placeholder[target=b_preprocessor_featurizer_fb]
    %audio : [num_users=2] = placeholder[target=audio]
    %length : [num_users=1] = placeholder[target=length]
    %sym_size_int_4 : [num_users=1] = call_function[target=torch.ops.aten.sym_size.int](args = (%audio, 0), kwargs = {})
    %unsqueeze : [num_users=1] = call_function[target=torch.ops.aten.unsqueeze.default](args = (%audio, 0), kwargs = {})
    %submod_5 : [num_users=1] = get_attr[target=submod_1]
    %to : [num_users=1] = call_function[target=torch.ops.higher_order.wrap_with_set_grad_enabled](args = (False, %submod_5, %unsqueeze), kwargs = {})
    %getitem_4 : [num_users=1] = call_function[target=operator.getitem](args = (%to, 0), kwargs = {})
    %submod_6 : [num_users=1] = get_attr[target=submod_2]
    %wrap_with_set_grad_enabled : [num_users=2] = call_function[target=torch.ops.higher_order.wrap_with_set_grad_enabled](args = (False, %submod_6, %length, %sym_size_int_4, %getitem_4, %b_preprocessor_featurizer_window, %b_preprocessor_featurizer_fb), kwargs = {})
    %masked_fill_2 : [num_users=1] = call_function[target=operator.getitem](args = (%wrap_with_set_grad_enabled, 0), kwargs = {})
    %where : [num_users=1] = call_function[target=operator.getitem](args = (%wrap_with_set_grad_enabled, 1), kwargs = {})
    %submod_7 : [num_users=1] = get_attr[target=submod_3]
    %to_6 : [num_users=1] = call_function[target=torch.ops.higher_order.wrap_with_set_grad_enabled](args = (False, %submod_7, %masked_fill_2), kwargs = {})
    %getitem_5 : [num_users=1] = call_function[target=operator.getitem](args = (%to_6, 0), kwargs = {})
    return (getitem_5, where)
```

Looking at the graph, the key is in those placeholder buffers:
* `b_preprocessor_featurizer_window` - the window function (e.g., Hann window)
* `b_preprocessor_featurizer_fb` - the mel filterbank matrix

When we call `model.to(torch.bfloat16)`, it converts all parameters and buffers in the model, including the preprocessor's `window` and `filterbank` tensors.

The mel spectrogram computation does something like:
```python
# Simplified view of what happens inside:
windowed = audio_frame * window          # window is bf16 → result is bf16
spectrum = torch.fft.rfft(windowed)         # stays bf16
mel = spectrum @ filterbank                      # filterbank is bf16 → result is bf16
```

PyTorch's type promotion rules mean that when our bf16 audio interacts with bf16 buffers, the output stays bf16.

Gasoonjia (approved):
Thanks for detailed update! LGTM now!
nit: mind changing PR title to remove float16?

examples/models/parakeet/export_parakeet_tdt.py
```diff
@@ -316,7 +320,8 @@ def export_all(model):
 
     preprocessor_wrapper = PreprocessorWrapper(model.preprocessor)
     preprocessor_wrapper.eval()
-    sample_audio = torch.randn(max_audio_samples)
+    # Preprocessor always uses float32 - runner converts output to encoder's dtype
```
Gasoonjia: Maybe remove this comment? looks like torch.export handle the dtype changing instead of runner,.

PR #16822, opened 2026-01-23
Arm backend: Add partial 3d support for gather

oscarandersson8218:
Add support for 3D gathers when input_size[-1] == index_size[-1]. In this case we can permute and reshape input [N,K,C] and index [N,W,C] to [N*C,K,1] and [N*C,W], apply a tosa.GATHER, and then reshape and permute back to [N,W,C].


zingo (approved):
OK to merge

PR #16825, opened 2026-01-23
Arm backend: Add "normalized op name" -> "edge op name mapping"

martinlsm:
To produce exir operator test coverage reports for the Arm backend, it requires mapping from normalized op name to full edge op name. Add this to collect_testname_resources.py.


zingo (approved):
OK to merge

PR #16826, opened 2026-01-23
Arm backend: Fix LeakyReLU to handle the inplace attribute

gggekov:


zingo (approved):
OK to merge when tests has runned if the Arm-backend and Cortex-M test pass and the rest matches latest on main :)

PR #16828, opened 2026-01-23
Arm backend: Fix replace_inf pass test names

martinlsm:
Pre-push naming checks require test_PASS_TARGET_* in passes. Rename the replace_inf tests to use the pass token (replace_inf_and_limit_values) and add a target suffix so the hook accepts them.


zingo (approved):
OK to merge when you can get the Arm Backend tests to PASS 

I spot 

FAILED backends/arm/test/ops/test_sub.py::test_sub_tensor_tosa_INT_2[rand_3d_Scalar] - AssertionError: Output 0 does not match reference output.

So you might just need to re-run the failed tests.

PR #16829, opened 2026-01-23
Arm backend: Fix tosa_spec issue in op_gather

Erik-Lundell:
A recent commit removed it from TosaArg, preferring instead to use the node_visitors self.tosa_spec.


zingo (approved):
OK to merge

PR #16833, opened 2026-01-23
Qualcomm arange

JacobSzwejbka:


cccclai:
@haowhsu-quic @winskuo-quic @shewu-quic @DannyYuyang-quic any chance you know a good way to handle this patch?

backends/qualcomm/builders/op_arange.py
```diff
@@ -27,7 +26,8 @@ def define_node(
     ) -> PyQnnManager.PyQnnOpWrapper:
         start, end = node.args[0:2]
         step = node.args[2] if len(node.args) > 2 else 1
-        out_tensor = torch.arange(start, end, step)
+        dtype = node.kwargs.get("dtype")
```
shewu-quic, line 29: How about use `node.meta["val"].dtype`? I think it will not be None.

backends/qualcomm/tests/test_passes.py
```diff
@@ -149,6 +148,91 @@ def test_mha_to_sha(self):
                 f"Output {i} mismatch: got {out}, expected {ref}",
             )
 
+    def test_arange_dtype_from_kwargs(self):
```
shewu-quic, line 151: Could you reuse original unit test for arange op?
Such as add more dtype coverage.
https://github.com/pytorch/executorch/blob/ac0a201450d65ce90f54c5eea202384f84106d10/backends/qualcomm/tests/test_qnn_delegate.py#L207
https://github.com/pytorch/executorch/blob/ac0a201450d65ce90f54c5eea202384f84106d10/backends/qualcomm/tests/test_qnn_delegate.py#L2335

shewu-quic (commented):
Thanks for fixing it.

PR #16841, opened 2026-01-23
Handle constants without placeholder nodes in convert_constant_dim_order_pass

GregoryComer:
Summary: After recent changes to convert_constant_dim_order_pass, there are a few lowering breakages when constant nodes don't have associated placeholders. I didn't realize this could happen, but apparently it does. So just skip placeholder meta updates in this case.

Differential Revision: D91334285

### DIFF
--- a/exir/passes/convert_constant_dim_order_pass.py
+++ b/exir/passes/convert_constant_dim_order_pass.py
@@ -1,6 +1,10 @@
+import logging
+
 import torch
 from torch.export import ExportedProgram
 from torch.export.graph_signature import InputKind
+
+logger = logging.getLogger(__name__)
 
 
 def _should_transform(tensor: torch.Tensor) -> bool:
@@ -17,7 +21,7 @@ def _update_placeholder_meta(
     exported_program: ExportedProgram,
     target: str,
     kind: InputKind,
-):
+) -> None:
     input_spec = next(
         (
             spec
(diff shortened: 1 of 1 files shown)


Gasoonjia (commented):
Thanks for update!
My major concern is whether or not that's the target behavior we want to have. In `convert_constant_dim_order_pass` we make const in contiguous first, then update the placeholder metadata. If in the _update_placeholder_meta we skip the update, will it introduce metadata-mismatch and lead to more issue?

exir/passes/convert_constant_dim_order_pass.py
```diff
@@ -27,7 +31,8 @@ def _update_placeholder_meta(
         None,
     )
     if input_spec is None:
-        raise RuntimeError(f"Missing input spec for lifted tensor {target}.")
+        logger.warning(f"Missing input spec for constant {target}")
```
Gasoonjia, line 34: Do you know when the export program may not have input_spec? Doesn't every program have that?

GregoryComer: I can track down a better repro. This showed up with an internal CV model. My guess is that there is some pass that is modifying the constant and it hasn't been fully re-lifted at this point. I'll see if I can find an exact case. There are several broken internal tests for export currently.

PR #16844, opened 2026-01-23
Fix backends toctree

rascani:
### Summary
Update the toctree links to match the backend table links.


psiddh (approved):
Accepting it unblock. Once the doc build is done, lets verify all the links

rascani:
Fixed in #16861

PR #16848, opened 2026-01-24
Support tuple outputs in aten bridge

GregoryComer:
Summary: Update the aten bridge logic in make_aten_functor_from_et... to handle kernels that return tuples of tensors. This motivating use case is the box_with_nms_limit custom op, which returns a tuple of 6 output tensors.

Differential Revision: D91375148

### DIFF
--- a/extension/aten_util/make_aten_functor_from_et_functor.h
+++ b/extension/aten_util/make_aten_functor_from_et_functor.h
@@ -249,28 +249,29 @@ struct type_convert<std::tuple<F...>, std::tuple<T...>> final {
  public:
   // We need to remove references from the output type because converters expect
   // to return by value. For example, the tensor converter creates a temporary
-  // tensor and returns it by value - so we can't return an lvalue reference to it.
+  // tensor and returns it by value - so we can't return an lvalue reference to
+  // it.
   using InType = std::tuple<F...>;
   using OutType = std::tuple<std::remove_reference_t<T>...>;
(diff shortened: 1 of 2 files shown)


extension/aten_util/make_aten_functor_from_et_functor.h
```diff
@@ -244,17 +326,23 @@ struct wrapper_impl<R (*)(Args...), f, int, N> {
...
   static_assert(
-      (N < num_args &&
-       std::is_same_v<
-           executorch::extension::kernel_util_internal::element_t<
-               N,
-               executorch::extension::kernel_util_internal::typelist<Args...>>,
-           R>) ||
+      N < num_args,
+      "The index of the out tensor can't be greater or equal to num_args.");
+  static_assert(
```
GregoryComer, line 336: The verification for tuple outputs could be nicer here. I didn't do it, mainly because I'll need to add even more template utility to conditionally work with tuples. I'll do it if anyone wants it.

JacobSzwejbka (approved):
Only Tuple<Tensor> is supported in aten right?

GregoryComer:
> Only Tuple is supported in aten right?

@JacobSzwejbka Can you clarify? With this change, we technically support bidirectional conversion of tuples as it recurses into its arguments. Though, in practice, we just need tuple<etensor...> -> tuple<atensor...> for returns.

PR #16865, opened 2026-01-26
Qualcomm AI Engine Direct - Optimize UT execution time

chenweng-quic:
### Summary
- Currently all tests will push same libs, which is redundant. With this PR it only push once, and reduce execution time from two to three times for operator's tests.

### Test plan
```
python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNFloatingPointOperator --device <device> --host <host> --model SM8650 --build_folder build-android --executorch_root . --artifact all_artifact
```
without optimization
image

with optimization
image

cc @cccclai @cbilgin


cccclai (approved):
Looks good

PR #16868, opened 2026-01-26
Arm backend Split up ZephyrOS tests per target and add SDK to docker

zingo:
### Summary
Split ZephyrOS tests per target so tests can run separate if wanted
and add support to run multipe tests after setup if you want to run
them all localy.

This removed the allowpattern as it would not stop bad code anyway
but gives the false impression that it does, removing the check also
make the code more easy to understant.

This also introduces invisible tags in the [README.md](http://readme.md/) in the form of

To make the parsing a bit more machine friendly and to make it easier
to change the readable text without strange test side effects.

This also only set git user in CI scripts if not already set and backup test folder so you don't destroy existing work or test data. This allows the script to be run locally more easy for testing.

### Test plan
Test is part of this PR


zingo:
Cortex-M fails: test_implementation_activation (12 fails) and test_implementation_linners (1 fail) are also failing on main.

mansnils (approved):

zingo:
Hi @SS-JIA or @rascani  anything you want to check/look at with this ZephyrOS tests update before merging?

rascani (approved):

rascani:
> Hi @SS-JIA or @rascani anything you want to check/look at with this ZephyrOS tests update before merging?

This looks good to me. Thank you for setting this up!

PR #16870, opened 2026-01-26
Cortex-M backend: Fix quantizer edge-cases

AdrianLundell:
- Make sure not to annotate non-float tensors
- Handles the possibility of non-annotated nodes in the SharedQspecQuantizer


psiddh:
Nice fixes. 
-  Should we also add test case for #1 non-float input tensors.
The CI for cortex m tests are broken because of my earlier changes and not this PR ( am looking into  it)

psiddh (approved):

AdrianLundell:
Test fails unrelated

PR #16878, opened 2026-01-26
[RELEASE ONLY CHANGES] Use torch 2.10 release packages

rascani:
### Summary
This change switches to torch 2.10 release packages for pytorch and domain libraries for 1.1 release. See #15168 as example from 1.0 release. One notable difference is that we will allow newer versions of torch (ie, 2.11) to be used with 1.1 release.

### DIFF
--- a/requirements-examples.txt
+++ b/requirements-examples.txt
@@ -5,3 +5,5 @@ timm == 1.0.7
 torchsr == 1.0.4
 torchtune >= 0.6.1
 transformers == 5.0.0rc1
+torchaudio == 2.10.0
+torchvision == 0.25.0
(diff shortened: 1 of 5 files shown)


requirements-examples.txt
```diff
@@ -5,3 +5,5 @@ timm == 1.0.7
 torchsr == 1.0.4
 torchtune >= 0.6.1
 transformers == 5.0.0rc1
+torchaudio == 2.10.0
+torchvision == 0.25.0
```
mergennachin: Also inequality?

rascani: Done.

mergennachin (approved):

PR #16887, opened 2026-01-26
Out of bounds heap write due to an incorrect bounds check in Program::get_constant_buffer_data

lucylq:
### Summary
Fix inequality and add test. We need to make sure the requested buffer is within the constant_buffer size and doesn't read out of bounds (constant_buffer having more data than requested is fine).

### Test plan
```
cmake --build . --target program_test
ctest -R program_test -V
```

### DIFF
--- a/runtime/executor/program.cpp
+++ b/runtime/executor/program.cpp
@@ -402,12 +402,16 @@ Result<const void*> Program::get_constant_buffer_data(
     const auto& constant_buffer = *constant_buffer_ptr;
     const auto* storage = constant_buffer[buffer_index]->storage();
     auto storage_size = storage == nullptr ? 0 : storage->size();
+    // nbytes (requested from the program) should be less than storage_size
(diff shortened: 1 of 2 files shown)


runtime/executor/program.cpp
```diff
@@ -403,11 +403,11 @@ Result<const void*> Program::get_constant_buffer_data(
     const auto* storage = constant_buffer[buffer_index]->storage();
     auto storage_size = storage == nullptr ? 0 : storage->size();
     ET_CHECK_OR_RETURN_ERROR(
-        storage_size <= nbytes,
+        nbytes <= storage_size,
```
larryliu0820, line 410: We allow nbytes to be smaller than storage_size, because we can potentially add padding when storing constant buffers. Add a comment?

lucylq: Thanks - updated

larryliu0820 (approved):

PR #16888, opened 2026-01-27
Run sampler as a method if available

larryliu0820:
With this PR: https://github.com/huggingface/optimum-executorch/pull/207 we are adding a new method "sampler" to ASR models, alongside with "encoder" and "text_decoder". The flow becomes: if temperature is 0 and sampler method is available, run that method. Otherwise still go with the old path. This change should largely improve the performance on CUDA since we don't have to copy logits from device to CPU for sampling purpose.

Benchmark result on RTX 5080:

```

======================================================================
BENCHMARK SUMMARY
======================================================================
Total runs: 30
Generated tokens per run: 104

THROUGHPUT (tokens/sec):
  Min:    793.89 t/s
  Max:    845.53 t/s
  Mean:   820.35 t/s
  Stdev:  11.86 t/s

MODEL LOAD TIME (ms):
  Min:    620 ms
  Max:    2170 ms
  Mean:   700 ms
  Stdev:  279 ms

ENCODE TIME (ms, inference_start to prompt_eval_end):
  Min:    36 ms
  Max:    38 ms
  Mean:   37 ms
  Stdev:  1 ms

DECODE TIME (ms, prompt_eval_end to inference_end):
  Min:    123 ms
  Max:    131 ms
  Mean:   127 ms
  Stdev:  2 ms

======================================================================
```


extension/asr/runner/runner.cpp
```diff
@@ -276,15 +289,37 @@ Result<std::vector<int64_t>> AsrRunner::transcribe(
...
-        std::move(decoder_outputs[0]).toTensor();
-    const int64_t vocab_size = logits_tensor.numel();
-    ET_CHECK_OR_RETURN_ERROR(
-        vocab_size > 0, Internal, "Decoder logits tensor is empty.");
 
-    const int64_t next_token =
-        static_cast<int64_t>(::executorch::extension::llm::logits_to_token(
-            logits_tensor, config.temperature));
+    int64_t next_token = 0;
+    if (!use_sampler_method || config.temperature == 0.0f) {
```
Gasoonjia: Here should be `config.temperature != 0.0f` right? We will fallback to original way when temperature != 0.

Gasoonjia (approved):
Perf is crazy!! THanks for the great work!

And we now is still load generate token from gpu to cpu one by one right iiuc. if so the perf should be boost again after we use batch transfer!

extension/asr/runner/runner.cpp
larryliu0820: ah you are right! Thank you!

PR #16892, opened 2026-01-27
[RELEASE ONLY CHANGES] Remove trailing slash from TORCH_URL_BASE

rascani:
The trailing slash on TORCH_URL_BASE caused the install_requirements.py to pick up the torchvision from PyPI rather than the pytorch index url.


mergennachin (approved):
Tentatively accepted

PR #16901, opened 2026-01-27
NXP backend: Update partitioning to avoid delegating no-op partitions.

MartinPavella:
### Summary
If a partition delegated no Neutron only contained no-ops, it would cause a crash. This PR identifies such cases and prohibits their delegation.

### Test plan
Unit-tests provided.

### DIFF
--- a/backends/nxp/backend/edge_helper.py
+++ b/backends/nxp/backend/edge_helper.py
@@ -1,7 +1,9 @@
-# Copyright 2024-2025 NXP
+# Copyright 2024-2026 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
+
+import logging
 
 import torch
 
@@ -18,6 +20,14 @@ DEQUANTIZE_OPERATORS = [
     exir_ops.edge.quantized_decomposed.dequantize_per_channel.default,
     exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
 ]
+
+# A set of operators which could possibly be no-ops in certain conditions. The operators in this set will be proclaimed
+#  as no-ops (and potentially not delegated), if their input and output tensors are equal (when run on random data).
+no_op_candidates = {
+    exir_ops.edge.aten.add.Tensor,
+    exir_ops.edge.aten.mul.Tensor,
+    exir_ops.edge.aten.sub.Tensor,
+}
 
 
 def input_tensor(node: Node, input_index: int) -> torch.Tensor:
@@ -220,3 +230,127 @@ def get_non_qdq_parent(node: Node, input_index: int = 0) -> Node | None:
         return None
 
     return quant_node.args[0]
+
+
+def try_get_dequantized_data(
+    dequantize_node: Node, parameters_mapping: dict[str, Parameter]
+) -> Parameter | None:
+    """Get the dequantized data from the following pattern. The dequantization formula is `r = (q - Z) * S`, where `q`
+        represents the static quantized data.
+
+         ┌─────────────────────────┐
+         │ <static_quantized_data> │
+         └────────────┬────────────┘
+                      │
+                ┌─────▼──────┐
+                │ Dequantize │
+                └─────┬──────┘
+                      ▼
+
+
+    :param dequantize_node: The Dequantize node from the pattern, which dequantizes the static quantized data.
+    :param parameters_mapping: Dict mapping tensor names to their static data. Should be inferred from the
+                                `state_dict` attribute of an edge program.
+    :return: The dequantized static parameter, or `None` if the data is not available.
+    """
+    if not _is_dequantize(dequantize_node):
+        return None
+
+    if not node_is_static_tensor(param := dequantize_node.args[0], parameters_mapping):
+        return None
+
+    # The pattern is correct. Dequantize the static data and return it.
+    scale, zp = get_quantization_parameters_for(dequantize_node)
+    quantized_data = parameters_mapping[param.name]
+
+    dequantized_data = (quantized_data - zp) * scale
+    return dequantized_data
+
+
+def is_no_op_on_neutron(node: Node, parameters_mapping: dict[str, Parameter]) -> bool:
+    """Check if a node is a no-op operation from the perspective of Neutron."""
+    if node.op != "call_function":
+        raise ValueError(
+            f"is_no_op_on_neutron(): Expected call_function node, got {node.op}."
+        )
+
+    if node.target in [
+        exir_ops.edge.aten.view_copy.default,
+        exir_ops.edge.dim_order_ops._clone_dim_order.default,
+        exir_ops.edge.aten.clone.default,
+    ]:
+        # Known operators which are always no-ops on Neutron.
+        return True
+
+    if node.target == exir_ops.edge.aten.cat.default and len(node.args[0]) == 1:
+        # Concatenation with 1 input is a no-op.
+        return True
+
+    # For any other operators, run them with random data and see if the output is identical to the input.
+    torch.manual_seed(42)
+    # noinspection PyBroadException
+    try:
+        input_data = None
+        args_with_random_data = []
+        for arg in node.args:
+            match arg:
+                case Node():
+                    # `arg` is either another operator, a model input, or a static parameter.
+
+                    if (
+                        data := try_get_dequantized_data(arg, parameters_mapping)
+                    ) is not None:
+                        # The `arg` is a static parameter. Use it's actual static data during the no-op test.
+                        args_with_random_data.append(data)
+
+                    else:
+                        # The `arg` is a compute node or a model input. Replace it with random data for the no-op test.
+                        if input_data is not None:
+                            # Some random input data for `node` has already been stored, which means that the node has
+                            #  more than 1 dynamic input node. Therefore, it cannot be a no-op.
+                            return False
+
+                        # Generate the random data. Use the range [-5, 5) to avoid proclaiming operations like Relu as
+                        #  no-ops.
+                        val = arg.meta["val"]
+                        input_data = torch.rand(val.shape, dtype=val.dtype) * 10 - 5
+                        args_with_random_data.append(input_data)
+
+                case list():
+                    # Lists of input nodes are not supported to keep the code simple. It is not crucial to support this
+                    #  case as the affected operators are either not supported on Neutron, or are extremely unlikely to
+                    #  be no-ops (e.g. GRU). One exception is `aten.cat`, which is explicitly supported above.
+                    return False
+
+                case _:
+                    # Generic argument (value). Not an input from a previous node. Store it in the arguments for the
+                    #  no-op test.
+                    args_with_random_data.append(arg)
+
+        # Run the operator with the random data. If the input equals the output, the node is considered a no-op.
+        output_data = node.target(*args_with_random_data)
+
+        val = node.meta["val"]
+        if (
+            output_data.dtype == val.dtype
+            and output_data.shape == val.shape
+            and torch.all(input_data == output_data)
+        ):
+            # The operator preserves the shape, data type, and data. Therefore, it is a no-op from the perspective of
+            #  Neutron.
+            if node.target in no_op_candidates:
+                return True
+            else:
+                logging.info(
+                    f"Found the operator `{node.target}`, which appears to be a no-op, but is not in the "
+                    "known no-op list. Please report this issue."
+                )
+                return False
+
+        else:
+            # Type, shape, or data doesn't match.
+            return False
+
+    except Exception:
+        # If execution fails, assume it's not a no-op.
+        return False
--- a/backends/nxp/backend/edge_program_converter.py
+++ b/backends/nxp/backend/edge_program_converter.py
@@ -13,6 +13,7 @@ from executorch.backends.nxp.backend.ir.converter.builder.aten_model_builder_dir
 from executorch.backends.nxp.backend.ir.converter.node_converter import (
     CustomDelegationOptions,
 )
+from torch._subclasses import FakeTensor
 from torch.export import ExportedProgram
 from torch.export.graph_signature import InputKind
 from torch.fx import Node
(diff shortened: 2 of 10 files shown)


backends/nxp/neutron_partitioner.py
```diff
@@ -314,12 +316,40 @@ def __init__(
...
+        non_q_dq_partition_nodes = list(filter(is_not_qdq_node, partition.nodes))
+
+        if all(
+            is_no_op_on_neutron(node, parameters_mapping)
+            for node in non_q_dq_partition_nodes
+        ):
+            # All operators in the partition are no-ops from the perspective of Neutron.
+            return False
+
+        return True
```
jirioc, line 351: What if non_q_dq_partition_nodes is empty (if the partition contains only QDQ). Should not this routine return False?

jirioc: The code would be faster with the reverse logic: if any node is valid on neutron, return true, otherwise false.

MartinPavella: > The code would be faster with the reverse logic: if any node is valid on neutron, return true, otherwise false.


That is actually not true, as the `all(...)` call does not go through all the nodes if it doesn't have to. If the first node returns False from the `is_no_op_on_neutron(node, parameters_mapping)` call, no other nodes are checked, and the `all(...)` returns False.

jirioc: Correct. Still my first argument holds.

MartinPavella: > What if non_q_dq_partition_nodes is empty (if the partition contains only QDQ). Should not this routine return False?

I believe that it is not possible in our current codebase for a partition with only Quantize/Dequantize nodes to exist. Either way, in this case I think the function should return False (as it is doing now), so I think the code is correct.

jirioc: I meant the code would be faster because you don't have to check the emptiness of the filtered partition. The ```all()``` function returns true on empty input. However, I understand that this situation is not possible. I am resolving this part.

backends/nxp/backend/edge_helper.py
```diff
@@ -19,6 +21,14 @@
     exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
 ]
 
+# A set of operators which could possibly be no-ops in certain conditions. The operators in this set will be proclaimed
+#  as no-ops (and potentially not delegated), if when run with random data their output matches their input.
```
roman-janik-nxp: "if when run with random data their output matches their input." - not clear language.

backends/nxp/backend/edge_helper.py
```diff
@@ -19,6 +21,14 @@
...
+operators_which_may_be_no_ops = {
```
roman-janik-nxp: Minor: what about `no_ops_candidates`?

backends/nxp/backend/edge_helper.py
```diff
@@ -220,3 +230,124 @@ def get_non_qdq_parent(node: Node, input_index: int = 0) -> Node | None:
...
+        exir_ops.edge.aten.clone.default,
+    ]:
+        # Known operators which are always no-ops on Neutron.
+        return True
+
+    if node.target == exir_ops.edge.aten.cat.default and len(node.args[0]) == 1:
+        # Concatenation with 1 input is a no-op.
+        return True
+
+    # For any other operators, run them with random data ad see if the output is identical to the input.
```
roman-janik-nxp:
```suggestion
    # For any other operators, run them with random data and see if the output is identical to the input.
```

backends/nxp/backend/edge_helper.py
```diff
@@ -220,3 +230,124 @@ def get_non_qdq_parent(node: Node, input_index: int = 0) -> Node | None:
...
+                            return False
+
+                        # Generate the random data. Use the range [-5, 5) to avoid proclaiming operations like Relu as
+                        #  no-ops.
+                        val = arg.meta["val"]
+                        input_data = torch.rand(val.shape, dtype=val.dtype) * 10 - 5
+                        args_with_random_data.append(input_data)
+
+                case list():
+                    # Multiple input nodes are not supported. `aten.cat` is explicitly supported above.
```
roman-janik-nxp: Please elaborate why multiple input nodes are not supported. If there is a list of args in node, it's always supported on Neutron? There are other ops with list in args, for example GRU.

backends/nxp/backend/edge_program_converter.py
```diff
@@ -161,20 +163,49 @@ def _process_nodes(self, nodes: list[Node], conversion_context: ConversionContex
...
-    def map_inputs_to_parameters(edge_program: ExportedProgram) -> dict[str, Parameter]:
+    def map_inputs_to_parameters(
+        edge_program: ExportedProgram,
+        post_quantization_state_dict: dict[str, Parameter] | None = None,
+    ) -> dict[str, Parameter]:
         """
         Create mapping between program parameters (input nodes & static data nodes) and their names.
 
         :param edge_program: EdgeProgram instance.
+        :param post_quantization_state_dict: State-dict of the model right after quantization. During partitioning, the
```
roman-janik-nxp, line 173: Is this the same dict as:
```
:param parameters_mapping: Dictionary mapping static parameter names to Parameter objects containing their data
                                    (if they have any). During partitioning, this data is extracted from the model right
                                    after quantization and before edge dialect passes. Therefore, it could potentially
                                    be outdated.
```
from `node_converter.py`?

backends/nxp/tests/ir/converter/node_converter/test_view_copy_converter.py
```diff
@@ -507,3 +507,67 @@ def test__view_copy__channels_first_to_channels_first__transpose_not_supported()
...
+    new_shape2 = [6, 4]
+    module = ViewViewModel(new_shape1, new_shape2)
+
+    ep = to_quantized_edge_program(
+        module,
+        input_shape,
+    ).exported_program()
+
+    # Make sure neither `view_copy` was delegated.
+    assert not any(n.name == "executorch_call_delegate" for n in ep.graph.nodes)
```
roman-janik-nxp: Please use comparisons against `target` attribute instead of str names:
`n.target == torch._higher_order_ops.executorch_call_delegate`.

Applies to all occurrences.

backends/nxp/tests/ir/converter/node_converter/test_view_copy_converter.py
```diff
@@ -507,3 +507,67 @@ def test__view_copy__channels_first_to_channels_first__transpose_not_supported()
...
+    input_shape = (1, 2, 3, 4)
+    new_shape = [2, 12]
+    module = ViewAddZeroModel(new_shape)
+
+    ep = to_quantized_edge_program(
+        module,
+        input_shape,
+    ).exported_program()
+
+    # Make sure neither the `view` nor the `add` were delegated
```
roman-janik-nxp:
```suggestion
    # Make sure neither the `view` nor the `add` was delegated
```
grammar - neither is singular

backends/nxp/tests/test_context_sensitive_delegation.py
```diff
@@ -25,47 +27,150 @@ def forward(self, x):
...
+def test_noop_partitions__concatenate_one_tensor_and_add_zeros():
+    input_shape = (1, 2, 3, 4)
+    module = ConcatAddNoOpModel(input_shape)
+
+    ep = to_quantized_edge_program(
+        module,
+        input_shape,
+    ).exported_program()
+
+    # Make sure neither the `cat` nor the `add` were delegated
```
roman-janik-nxp:
```suggestion
    # Make sure neither the `cat` nor the `add` was delegated
```
grammar - neither is singular

backends/nxp/neutron_partitioner.py
```diff
@@ -328,10 +358,33 @@ def validate_partitioning_result(
...
+        """Verify that the result of the partitioning can be safely used for delegation.
+
+        :param graph: Original graph that was partitioned.
+        :param partition_list: List of the resulting partitions.
+        :param custom_delegation_options: Custom user options which affect node delegation.
+        :param parameters_mapping: Dict mapping tensor names to their static data. Should be inferred from the
+                                   `state_dict` attribute of an edge program.
+        :return: True, if the partitioning result is valid.
+        """
+        partitioning_valid = True
```
roman-janik-nxp, line 373: Not related to this feature, but could be added in this PR: if the `partition_list` is empty we can return `True` immediately and skip the rest of the function.

backends/nxp/backend/edge_helper.py
MartinPavella: I added a comment explaining the motivation.

backends/nxp/backend/edge_program_converter.py
MartinPavella: Yes

Edit: Sorry, I misread your comment. These are 2 different dictionaries.

backends/nxp/neutron_partitioner.py
MartinPavella: Good point, I will add it.

backends/nxp/backend/edge_helper.py
```diff
@@ -220,3 +230,124 @@ def get_non_qdq_parent(node: Node, input_index: int = 0) -> Node | None:
...
+            # The operator preserves the shape, data type, and data. Therefore, it is a no-op from the perspective of
+            #  Neutron.
+            if node.target in operators_which_may_be_no_ops:
+                return True
+            else:
+                logging.info(
+                    f"Found the operator `{node.target}`, which appears to be a no-op, but is not in the "
+                    "known no-op list. Please report this issue."
+                )
+
```
jirioc, line 349: Should any value be returned in this case? Logging message might not be sufficient in this case.

MartinPavella: Of course. Thanks for spotting that. I will add a `return False`.

robert-kalmar:
@MartinPavella  the internal build failed https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH64-6

backends/nxp/backend/edge_program_converter.py
roman-janik-nxp: Ok, consider unification of param name and doc across files.

backends/nxp/tests/test_context_sensitive_delegation.py
```diff
@@ -25,47 +27,156 @@ def forward(self, x):
...
+
+def test_single_view_copy_partition():
+    input_shape = (2, 10)
+    module = SingleViewCopyModule([1, 20])
+
+    ep = to_quantized_edge_program(module, input_shape).exported_program()
+
+    # Make sure the `view_copy` was not delegated.
+    assert graph_contains_any_of_ops(ep.graph, [exir_ops.edge.aten.view_copy.default])
+    assert not any("delegate" in n.name for n in ep.graph.nodes)
```
roman-janik-nxp: Please use comparisons against target attribute instead of str names:
n.target == torch._higher_order_ops.executorch_call_delegate.

Applies to all occurrences.

backends/nxp/neutron_partitioner.py
```diff
@@ -314,12 +316,40 @@ def __init__(
...
         neutron_target_spec: NeutronTargetSpec,
         custom_delegation_options: CustomDelegationOptions | None = None,
+        post_quantization_state_dict: dict[str, Parameter] | None = None,
     ) -> None:
+        """Class responsible for identifying partitions suited for delegation to the eIQ Neutron NPU.
+
+        :param compile_spec: A list of compile specifications for Neutron.
+        :param neutron_target_spec: Object for querying the target platform to retrieve its properties.
+        :param custom_delegation_options: Custom user options which affect node delegation.
+        :param post_quantization_state_dict: Optional dictionary mapping static parameter names to Parameter
```
roman-janik-nxp: https://github.com/pytorch/executorch/pull/16901#discussion_r2759605076

MartinPavella: I unified the `post_quantization_state_dict` doc-strings.

roman-janik-nxp:
Code looks ok now, good job! Just fix the internal build and I can approve.

MartinPavella:
> Code looks ok now, good job! Just fix the internal build and I can approve.

Thank you. The internal build is passing: https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH64-9

roman-janik-nxp (approved):

JakeStevens:
Does this break the MobileNetv2 example?

It broke our internal model lowering.

The reason: 

Depthwise needs to be static, so it does that check for the post quant map, which we were not providing (and to be honest, do not really want to because it would be pain)

SO

it needs

    cdo = CustomDelegationOptions(allow_no_op_partitions=True)
    
I suspect the example requires the same OR update to pass it the post quant params, it currently does neither.

Perhaps depthwise should be excluded from this? It is exceedingly unlikely that a depthwise is a no-op in a proper model.

I actually don't understand the point of this generally.

Avoiding something like:

ADD(x, 0) that is not able to be address in AOT?

MartinPavella:
@JakeStevens thanks for your comment and for spotting this issue.

---
> Does this break the MobileNetv2 example?

You are right, I didn't notice that the depth-wise convolutions were no longer selected for delegation in the example. I have a fix ready in [another PR](https://github.com/pytorch/executorch/pull/17479).

---
> it needs
> `cdo = CustomDelegationOptions(allow_no_op_partitions=True)`

I have no idea how this would fix the depth-wise delegation. Are you sure it really helps?

---
> Perhaps depthwise should be excluded from this? It is exceedingly unlikely that a depthwise is a no-op in a proper model.

I see no connection between the depth-wise conv requiring static data during partitioning, and the no-op checks. This PR added the no-op checks, which required access to the actual static data of the model during partitioning. Providing this access (via the `post_quantization_state_dict`) had an effect on the whole partitioning process. Since some operators need actual static data to determine Neutron support, providing the state_dict was inevitable, and it was going to happen regardless of the "no-op" stuff.
But, since I wanted to restrict the mapping to real data and avoid the FakeTensors, I didn't add the fallback to that you have added in [your PR](https://github.com/pytorch/executorch/pull/17452), which is what caused the depth-wise convolution issue in your internal tests. But the "checking for no-ops" is not relevant here.

---
>I actually don't understand the point of this generally.
>
>Avoiding something like:
>
>ADD(x, 0) that is not able to be address in AOT?

There are good reasons why we introduced this change. The main one is that some operators are necessary in ExecuTorch, but are no-ops in Neutron (for example `dim_order_ops._clone_dim_order.default`). They cannot be removed from the model beforehand, because if the surrounding nodes were not selected for delegation, the resulting model would be invalid. That is why it needs to be addressed during partitioning like this.

You are right that cases such as `ADD(x, 0)` could be removed before the partitioning stage. But following the robustness principle, I wanted to have a powerful check for not delegating no-ops, to minimize the risk of future failures. Plus, it was faster to implement these partitioning restrictions than to remove the nodes beforehand. That's why I did it first. In the future, I plan to add a pre-processing pass to remove these cases from the model.

JakeStevens:
> Are you sure it really helps?

My apologies, I forgot to update this comment after I opened my PR. 

The PR is the true resolution to my issue, the no op delegation is not.

PR #16905, opened 2026-01-27
Arm backend: Add support for transpose conv

mansnils:
Adds:
* transpose‑conv operator registration/TOSA-serialization
* conv pass rewrite support for transpose conv
* quantization handling
* weight dim_order update for correct TOSA layout
* unit tests


zingo (approved):

mansnils:
Fails unrelated

PR #16906, opened 2026-01-27
Arm backend: Fix Windows incompatibility in vgf_compile()

sarah-blades:


zingo (approved):
OK to merge if tests PASS

zingo:
Cortex-m tests test_implementation_activation and test_implementation_linear is a know problem and unrelated fail that fail on main also. Fix is ongoing

PR #16908, opened 2026-01-27
Fix docs upload job being canceled mid-operation

svekars:
Add workflow-level concurrency that only allows cancellation for PR builds, ensuring push/schedule builds always complete.

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -13,6 +13,10 @@ on:
   schedule:
     - cron: '0 0 * * *'
 
+concurrency:
+  group: docs-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main


.github/workflows/doc-build.yml
```diff
@@ -13,6 +13,10 @@ on:
   schedule:
     - cron: '0 0 * * *'
 
+concurrency:
+  group: docs-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
JacobSzwejbka, line 18: Will it reschedule it after the doc build completes?

JacobSzwejbka (approved):

.github/workflows/doc-build.yml
```diff
@@ -13,6 +13,10 @@ on:
   schedule:
     - cron: '0 0 * * *'
 
+concurrency:
+  group: docs-${{ github.workflow }}-${{ github.ref }}
```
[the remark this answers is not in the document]
svekars: For docs upload jobs that push to gh-pages, queuing is the desired behavior. Using `github.sha` would put each run in a separate group, allowing them to run in parallel — which could cause conflicts and doesn't protect against cancellation. The `github.ref` pattern ensures uploads run sequentially, which is correct for this use case.

.github/workflows/doc-build.yml
svekars: with this fix the job shouldn't be canceled

PR #16915, opened 2026-01-27
Fix heap-buffer-overflow vulnerability in WAV file loader

larryliu0820:
This PR fixes a security vulnerability in the WAV file loader where a malicious WAV file could trigger an out-of-bounds heap read by providing an inflated `Subchunk2Size` value in the file header.

The `load_wav_audio_data` function in `extension/llm/runner/wav_loader.h` reads `data_offset` and `data_size` directly from the WAV file header without validating that the claimed data size falls within the bounds of the actual file buffer. When `Subchunk2Size` exceeds the actual file size, the copy operation reads beyond the allocated heap buffer, potentially exposing sensitive data.

This vulnerability affects all code paths that process user-supplied WAV files through the ExecuTorch audio loading API, including the Whisper, Parakeet, and Voxtral example applications.

Added bounds validation in two locations:

1. **`load_wav_header`**: Validates that the data chunk header can be safely read before accessing `Subchunk2Size`
2. **`load_wav_audio_data`**: Validates that `data_offset + data_size <= buffer.size()` before performing any copy operations

Malicious WAV files with inflated size values are now rejected with a clear error message before any memory access beyond the buffer bounds can occur.

Added three new test cases:
- `LoadAudioDataRejectsInflatedSubchunk2Size`: Tests rejection of a WAV file claiming 1MB of data but containing only 4 bytes
- `LoadAudioDataRejectsDataOffsetBeyondBounds`: Tests rejection when data offset points beyond file bounds
- `LoadAudioDataAcceptsExactSizeMatch`: Verifies legitimate WAV files continue to work

Also updated `ET_EXPECT_DEATH` macro in `test/utils/DeathTest.h` to handle builds with logging disabled (`!ET_LOG_ENABLED`).

- `extension/llm/runner/wav_loader.h` - Added bounds validation
- `extension/llm/runner/test/test_wav_loader.cpp` - Added security tests
- `test/utils/DeathTest.h` - Updated macro to handle logging-disabled builds

### DIFF
--- a/extension/llm/runner/wav_loader.h
+++ b/extension/llm/runner/wav_loader.h
@@ -127,8 +127,19 @@ inline std::unique_ptr<WavHeader> load_wav_header(const std::string& fp) {
       ET_LOG(Info, "Bits per Sample: %d", header->bitsPerSample);
 
       if (data_offset != 0) {
+        // Validate that we can safely read the Subchunk2Size (4 bytes at
+        // data_offset + 4) and that the data starts at data_offset + 8
+        if (data_offset + 8 > data_size) {
+          ET_LOG(
+              Error,
+              "WAV file structure is invalid: data chunk header extends beyond file bounds (offset %zu, file size %zu)",
+              data_offset,
+              data_size);
+          throw std::runtime_error(
+              "Invalid WAV file: data chunk header extends beyond file bounds");
+        }
         header->Subchunk2Size =
-            *reinterpret_cast<const int32_t*>(data + data_offset + 4);
+            *reinterpret_cast<const uint32_t*>(data + data_offset + 4);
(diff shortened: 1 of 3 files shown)


extension/llm/runner/wav_loader.h
lucylq, line 89: noob q: does this mean the text "data" can be anywhere in the file/the header can be arbitrarily large?

lucylq (approved):

extension/llm/runner/wav_loader.h
larryliu0820: Yeah the thing is we don't know what's between the file start and "data". The wav spec didn't force this.

See an example here:

Image

"data" shows up really late in the file.

lucylq: Wow, interesting. Thanks for the explanation!

PR #16928, opened 2026-01-28
Arm backend: Make run_method_and_compare_outputs more flexible

Erik-Lundell:
Make it possible to configure the reference_stage in run_method_and_compare_outputs.
Make it possible to configure the comparison function in run_method_and_compare_outputs. Use this
to implement full tensor tests, instead of elementwise.

Use this to implement run_and_compare_to_initial_model in testing pipelines. This function (by default)
compares the initial model to the quantized, in eager mode. This is meant to be used to make sure that nothing bad happens when quantizing, most importantly in
transform_for_annotation.

Breaks out some logic from run_method_and_compare_outputs to reduce complexity.


zingo (approved):

Erik-Lundell:
Unrelated issues, merging.

PR #16930, opened 2026-01-28
Arm backend: Unmark Transformer as xfail

oscarandersson8218:
Transformer model was previously marked as expected failure due to numerical differences between reference and lowered model. At closer inspection, this difference appear to be within an acceptable range.


zingo (approved):

oscarandersson8218:
Test failure is unrelated.

PR #16931, opened 2026-01-28
Arm backend: Stop removing kwargs in quantizer

Erik-Lundell:
torchao now allows non-tensor kwargs when quantizing. Currently, I am not aware of any tensor kwargs so it is probably better to crash and handle it if we run into them instead of silently not quantizing them.

Add a test to make sure kwargs actually do survive quantization.


zingo (approved):

zingo (approved):

Erik-Lundell:
Unrelated issues, merging.

PR #16932, opened 2026-01-28
Fixed ci device allocation

Jiseong-oh:
### Summary
Enable device queue for exynos ci
Fixes https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=samsung

### Test plan
This pr will be tested on exynos device farm.


SS-JIA (changes requested):
LGTM, pending fixing the issues with SAMSUNG_AI_LITECORE_KEY and observing that the signals pass. Thanks for the fix!

PR #16933, opened 2026-01-28
[llm][runner] Respect num_bos in MultimodalRunner

kamalkraj:
### Summary
This PR addresses an issue where `MultimodalRunner` was ignoring the `num_bos` setting in `GenerationConfig`. This resulted in the BOS (Beginning of Sequence) token not being prepended to the prompt, which caused some models, specifically **MedGemma** (e.g., `medgemma-1.5-4b-it`), to generate garbage output or lose context.

The following changes were made:
1.  Updated `MultimodalPrefiller::prefill` signature to accept optional `bos` and `eos` parameters.
2.  Modified `MultimodalPrefiller` implementation to pass these parameters to the tokenizer's `encode` method.
3.  Updated `MultimodalRunner::generate` to read `config.num_bos` and pass it to the prefiller for the first text input.

This aligns `MultimodalRunner` behavior with `TextLLMRunner` and ensures correct prompt formatting for models that rely on explicit BOS tokens.

**Fixes:** https://github.com/meta-pytorch/executorch-examples/issues/185

### Test Plan
Verified manually using a **MedGemma 1.5** model (which requires a BOS token).

1.  **Before fix:** Running inference produced repetitive/garbage output (e.g., "modelmodelmodel...").
2.  **After fix:** Running inference produces coherent text output.

### Verified Script (`model_infer.py`)
```python
from executorch.extension.llm.custom_ops import custom_ops
from executorch.kernels import quantized
from executorch.extension.llm.runner import (
    GenerationConfig, MultimodalRunner, 
    make_text_input, make_image_input, make_audio_input
)
import torch

# Create a multimodal runner
runner = MultimodalRunner(
    model_path="medgemma-1.5-4b-it-8da4w-executorch/model.pte",
    tokenizer_path="medgemma-1.5-4b-it-8da4w-executorch/tokenizer.model"
)

# Create multimodal inputs
inputs = []
inputs.append(make_text_input("<start_of_turn>user\nHi<end_of_turn>\n<start_of_turn>model\n"))

# # Add image from torch tensor (supports both CHW and HWC formats)
# image_tensor = torch.randint(0, 255, (3, 896, 896), dtype=torch.float32)  # CHW format
# inputs.append(make_image_input(image_tensor))

# Configure generation
config = GenerationConfig(
    max_new_tokens=100,
    temperature=0.0,
    echo=False,
    num_bos=1,
)

# Generate with streaming output
def token_callback(token: str):
    # The runner already prints the token to stdout
    pass

def stats_callback(stats):
    print(f"\n[Stats] Generated {stats.num_generated_tokens} tokens")
    inference_time = stats.inference_end_ms - stats.inference_start_ms
    if inference_time > 0:
        tokens_per_sec = stats.num_generated_tokens * 1000 / inference_time
        print(f"[Stats] Speed: {tokens_per_sec:.1f} tokens/sec")

runner.generate(inputs, config, token_callback, stats_callback)
```


extension/android/jni/jni_layer_llama.cpp
```diff
@@ -143,22 +145,30 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
...
       facebook::jni::alias_ref<jstring> model_path,
       facebook::jni::alias_ref<jstring> tokenizer_path,
       jfloat temperature,
-      facebook::jni::alias_ref<jobject> data_files = nullptr) {
+      facebook::jni::alias_ref<jobject> data_files = nullptr,
+      jint num_bos = 0,
+      jint num_eos = 0) {
     temperature_ = temperature;
+    num_bos_ = num_bos;
+    num_eos_ = num_eos;
```
larryliu0820, line 171: If we are storing `num_bos_` and `num_eos_`, we should use it in the rest of this method

extension/llm/runner/multimodal_prefiller.cpp
```diff
@@ -176,7 +178,7 @@ Result<uint64_t> MultimodalPrefiller::prefill(
     std::vector<uint64_t> tokens;
     if (input.is_text()) {
       auto& text = input.get_text();
-      auto encode_result = tokenizer_->encode(text);
+      auto encode_result = tokenizer_->encode(text, bos, eos);
```
larryliu0820, line 181: Ideally tokenizer should be opinionated about how many bos/eos are needed. This is slightly annoying since only HF tokenizer has this info.

larryliu0820 (approved):

extension/android/jni/jni_layer_llama.cpp
kirklandsign: @larryliu0820 should this be per model or per generate()? Should be determined by model? but we do need the flexibility for per generate()?

This makes sense to me:

The member variables num_bos_ and num_eos_ are stored but never used. This is inconsistent with how temperature_ is handled (which has a fallback at line 245: temperature >= 0 ? temperature : temperature_).

Adding fallback logic similar to temperature: use the member variables when the generate parameters are negative or a special sentinel value

nil-is-all:
Hi @kamalkraj, feel free to merge the approved PR if it looks good to you.

kirklandsign:
> Task :executorch_android:compileDebugJavaWithJavac FAILED
/pytorch/executorch/extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/llm/LlmModule.java:238: error: no suitable method found for generate(String,int,LlmCallback,boolean,float)
    return generate(prompt, seqLen, llmCallback, echo, DEFAULT_TEMPERATURE);
           ^
    method LlmModule.generate(String,LlmCallback) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(String,int,LlmCallback) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(String,LlmCallback,boolean) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(String,int,LlmCallback,boolean) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(String,int,LlmCallback,boolean,float,int,int) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(String,LlmGenerationConfig,LlmCallback) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(int[],int,int,int,String,int,LlmCallback,boolean) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(int[],int,int,int,String,int,LlmCallback,boolean,float) is not applicable
      (actual and formal argument lists differ in length)
    method LlmModule.generate(int[],int,int,int,String,int,LlmCallback,boolean,float,int,int) is not applicable
      (actual and formal argument lists differ in length)
1 error

PR #16936, opened 2026-01-28
Arm backend: Fix VGF backend calling

per:
### Summary
Setting shell=True, makes run only use the first command in the list, causing model-converter to read from stdin instead of the supplied flatbuffer argument. Switch to shell=False to run the commands as expected.

### Test plan
Tested through ci tests.


zingo (approved):
Ok to merge

PR #16937, opened 2026-01-28
Arm backend: Add SharedQspec tests

AdrianLundell:
Adds Quantizer edge-case tests for operators using SharedQspec to lock down the behaviour and notice any changes.

Checks both qspecs and qparams to be extra rigorous.


zingo (approved):

zingo (approved):

AdrianLundell:
Fails not related

PR #16986, opened 2026-01-29
New exported program pass manager and exported program passes (#16986)

DrJessop:
Summary:

Adds support to run passes on ExportedPrograms, as well as a pass manager for exported programs. Now, users can provide graph module passes or exported program passes as an interleaved list, and the pass manager will run the appropriate logic depending on the pass type. The new pass manager also runs program validation before and after running all passes, and also includes support to run program-level validation after each pass. Users should use this pass manager in favor of the old exir.PassManager.

Reviewed By: angelayi, larryliu0820, ethansfng

Differential Revision: D91725222

### DIFF
--- a/backends/arm/test/tester/test_pipeline.py
+++ b/backends/arm/test/tester/test_pipeline.py
@@ -48,7 +48,7 @@ from executorch.backends.arm.util._factory import create_quantizer
 from executorch.backends.arm.vgf.compile_spec import VgfCompileSpec
 from executorch.backends.test.harness.stages import StageType
 from executorch.exir.pass_base import ExportPass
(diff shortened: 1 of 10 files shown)


robert-kalmar:
CC @roman-janik-nxp , @MartinPavella

larryliu0820 (approved):

robert-kalmar:
@DrJessop, is this PR still active?

DrJessop:
> @DrJessop, is this PR still active?

Yes it is, had to put it on the backburner for a bit, but coming back to it. Essentially, after further discussion, we would like to support not just passes on ExportedPrograms, but on EdgeProgramManagers themselves.

PR #17002, opened 2026-01-29
NXP backend: Fix Neutron IR node support check.

MartinPavella:
### Summary
The check to make sure a node is supported in Neutron IR was incomplete, so potential partitioning issues could have gone unnoticed. Now, this check is complete.

### Test plan
Tested by most of the existing tests.

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converter.py
@@ -1,4 +1,4 @@
-# Copyright 2024-2025 NXP
+# Copyright 2024-2026 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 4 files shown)


backends/nxp/backend/ir/converter/node_converter.py
robert-kalmar: Detail: Update the copyright

backends/nxp/tests/executors.py
```diff
@@ -392,10 +393,16 @@ def __init__(
     ):
         self._converter_class = converter_class
         self.new_target_support_check = new_target_support_check
+
+        # Store the original target check method. NOTE: The staticmethod must be unwrapped to get the underlying
+        #  function, therefore the stored function is not a staticmethod anymore.
         self.old_target_support_check = converter_class._is_supported_on_target
```
robert-kalmar: Nit: You can avoid the unwrapping by copying the method from the object/class dictionary:

```
self.old_target_support_check = converter_class.__dict__["_is_supported_on_target"]
```

This way you restore exactly what was originally assigned (i.e. the static method) and you don't need to wrap it back with `staticmethod()`

robert-kalmar (approved):

backends/nxp/tests/executors.py
MartinPavella: Done. Thanks for the tip.

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -173,13 +173,22 @@ def assert_convertible(self, node):
...
+        supported_on_target = self._is_supported_on_target(
+            node,
+            self.neutron_target_spec,
+            self.context.parameters_mapping,
+            self.context.custom_delegation_options,
+        )
+
+        assert supported_in_ir and supported_on_target, (
+            f"Node `{node}` was selected for delegation to Neutron, but it is not convertible to the intermediate "
+            "representation. There is an error in the NXP partitioner,]. Please report this."
```
roman-janik-nxp:
```suggestion
            "representation. There is an error in the Neutron partitioner. Please report this."
```

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -173,13 +173,22 @@ def assert_convertible(self, node):
         """Assert that the call `_is_supported_in_IR()` returns `True`. Otherwise, raise an exception and print an
```
roman-janik-nxp: Should change as well.

backends/nxp/tests/executors.py
```diff
@@ -392,10 +400,24 @@ def __init__(
     ):
         self._converter_class = converter_class
         self.new_target_support_check = new_target_support_check
-        self.old_target_support_check = converter_class._is_supported_on_target
+
+        # Retrieve the method exactly as defined on the class, not the bound attribute.
```
roman-janik-nxp:
```suggestion
        # Retrieve the method exactly as defined in the class, not the bound attribute.
```

MartinPavella:
The failing checks seem unrelated, and the internal build passed. Merging.

PR #17004, opened 2026-01-29
Arm backend: Add matmul decomposition pass

oscarandersson8218:
Adds pass that performs aten to core-aten decomposition of torch.matmul. Pass can be extended to other targets as well. Also updates test_at as tests had incorrect names.


zingo (approved):

oscarandersson8218:
Failures seem unrelated.

zingo:

PR #17011, opened 2026-01-29
CI: Remove docker-builds from viable/strict requirements and add debugging

mergennachin:
- Remove docker-builds from the required jobs for viable/strict
  - docker is run only once a week so it's too restrictive to have this as part of the viable/strict movement
- Add set -x and stdout logging to "Summarize CI failures" step for
  better debugging visibility


mergennachin:
docker is run only once a week so it's too restrictive to have this as part of the viable/strict movement

GregoryComer (approved):

rascani (approved):

PR #17012, opened 2026-01-29
Arm backend: Use fixed random seed by default in ARM test suite

psiddh:
Summary:
Preventive improvement: Change the default behavior of the ARM test suite's random seed fixture from RANDOM to a fixed seed (0). This ensures reproducible tests by default.

Previously, the fixture used random seeds by default, which could lead to non-deterministic test behavior. While no specific regression was confirmed, using fixed seeds is a best practice for test reproducibility.

With this change:
- Default: Fixed seed (0) for reproducible tests
- ARM_TEST_SEED=RANDOM: Opt-in to random seed behavior
- ARM_TEST_SEED=<number>: Use a specific seed

This aligns with standard testing practices where deterministic behavior is preferred over random initialization.

Differential Revision: D91752426


martinlsm:
Will this work given that pytests run multiprocessed? Just wondering if there is some randomness involved there? I guess test inputs will vary at least depending on which host that runs it because the number of cores determines the number of processes to spawn for the tests.

zingo:
> Will this work given that pytests run multiprocessed? Just wondering if there is some randomness involved there? I guess test inputs will vary at least depending on which host that runs it because the number of cores determines the number of processes to spawn for the tests.

Update: We did some testing and the SEED get set at every test so this should not be a problem.

zingo (approved):

psiddh:
For full context, this PR does not fix any broken issues , I ended with this fix, chasing an internal broken CI , but later figured this fix is not going to solve any broken CI, but thought , it could be a good enough preventive fix.

PR #17019, opened 2026-01-29
CI: Use _upload_docs.yml for gh-pages push

mergennachin:
Switch to using pytorch/test-infra's _upload_docs.yml reusable workflow.

Also fix artifact structure to put docs directly in RUNNER_ARTIFACT_DIR (not in html/ subfolder) to match _upload_docs.yml expectations.

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -87,10 +87,12 @@ jobs:
 
         cp -rf docs/_build/html/* "${RUNNER_DOCS_DIR}"
 
-        mv docs/_build/html "${RUNNER_ARTIFACT_DIR}"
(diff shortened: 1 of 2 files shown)


GregoryComer (approved):
Looks good to me. I did some more debugging and the token is missing in the new job, likely as a result of the checkout changes. If this resolves it before dev infra, I'm happy to take that path.

mergennachin:
cc @zklaus @atalman

mergennachin:
cc @clee2000 do you know if there are any users of .github/workflows/_upload_docs.yml in test-infra? 

The context is that our doc upload job starting failing recently, we're yet to figure out the root cause. We're trying to see if migrating to this job will help at all. But I don't find any usage of it in prod.

PR #17066, opened 2026-01-30
NXP backend: Unify ExecuTorch and Neutron data format enums.

MartinPavella:
### Summary
Originally, the NXP backend used the `NodeFormat` enum to mark data formats (memory layouts) of ExecuTorch nodes, and the `TensorFormat` enum was used to mark the formats for Neutron IR tensors. However, these two classes fundamentally represent the same thing, and there is no need to keep them separated. Additionally, sometimes it was confusing as for which class to use which could result (and has resulted) in bugs.

This PR unifies these 2 enums into 1.

### Test plan
N/A


MartinPavella:
This PR depends on https://github.com/pytorch/executorch/pull/17002, so only the last commit is relevant.

robert-kalmar (approved):

jirioc (approved):

MartinPavella:
The failing checks seem unrelated, and the internal build passed. Merging.

PR #17068, opened 2026-01-30
Arm backend: Use TosaSpecification list method

per:
### Summary
Move added TOSA dialect operator to use the class method to be registered for all supported TOSA versions.

### Test plan
Through upstream CI.


zingo (approved):

zingo:

PR #17075, opened 2026-01-30
Arm backend: Add Cortex-M as a first-class target in aot_arm_compiler

psiddh:
Add a dedicated to_edge_cortex_m() path selected via --target=cortex-m that
owns the full pipeline: CortexMQuantizer for INT8 quantization, correct
EdgeCompileConfig with preserve_ops to prevent premature decomposition, and
CortexMPassManager.pass_list for op conversion. Remove the old scattered
transform_for_cortex_m_backend() function.

Verified all ops fully lowered to cortex_m::quantized_* operators for both
MobileNetV2 (70 nodes) and MobileNetV3 (122 nodes). E2E inference tested
on Alif E8 board.

Test Plan:
    - python3 -m examples.arm.aot_arm_compiler -m mv2 --target=cortex-m55+int8 --quantize --intermediates=./mv2_intermediates --output=./mv2_cortex_m.pte
    - python3 -m examples.arm.aot_arm_compiler -m mv3 --target=cortex-m55+int8 --quantize --intermediates=./mv3_intermediates --output=./mv3_cortex_m.pte

    Also ran E2E inference on Alif E8 board

### DIFF
--- a/examples/arm/aot_arm_compiler.py
+++ b/examples/arm/aot_arm_compiler.py
@@ -36,11 +36,12 @@ from executorch.backends.arm.tosa.compile_spec import TosaCompileSpec
 from executorch.backends.arm.util._factory import create_partitioner, create_quantizer
 
 from executorch.backends.arm.vgf import VgfCompileSpec
+from executorch.backends.cortex_m.passes.cortex_m_pass_manager import CortexMPassManager
 
-# To use Cortex-M backend
 from executorch.backends.cortex_m.passes.replace_quant_nodes_pass import (
     ReplaceQuantNodesPass,
 )
+from executorch.backends.cortex_m.quantizer.quantizer import CortexMQuantizer
 
 from executorch.devtools import generate_etrecord
 from executorch.devtools.backend_debug import get_delegation_info
@@ -396,6 +397,7 @@ TARGETS = [
     "TOSA-1.0+INT",
     "TOSA-1.0+FP",
     "TOSA-1.0+INT+int16",
+    "cortex-m55+int8",
 ]
 
 
@@ -528,7 +530,7 @@ def get_args():
         required=False,
         default="ethos-u55-128",
         choices=TARGETS,
-        help=f"For ArmBackend delegated models, pick the target, and therefore the instruction set generated. valid targets are {TARGETS}",
+        help=f"Target backend. For delegated models: Ethos-U/VGF/TOSA variants. For non-delegated: cortex-m55+int8 (CMSIS-NN portable kernels). Valid targets: {TARGETS}",
     )
     parser.add_argument(
         "-e",
@@ -795,6 +797,75 @@ def to_edge_TOSA_delegate(
     return model_quant, edge
 
 
+def to_edge_cortex_m(
+    exported_program: ExportedProgram,
+    args,
+    model: GraphModule,
+    example_inputs: Tuple[torch.Tensor],
+):
+    """Cortex-M/CMSIS-NN compilation path with no delegation."""
+    logging.info("Using Cortex-M/CMSIS-NN compilation path (no delegation)")
+
+    def _to_channels_last(x):
+        if isinstance(x, torch.Tensor):
+            if x.dim() == 4 and not x.is_contiguous(memory_format=torch.channels_last):
+                logging.warning(
+                    "Converting input tensor with shape %s to channels_last",
+                    list(x.shape),
+                )
+                return x.to(memory_format=torch.channels_last)
+            return x
+        elif isinstance(x, tuple):
+            return tuple(_to_channels_last(t) for t in x)
+        return x
+
+    if not args.quantize:
+        logging.warning(
+            "Quantization is DISABLED. Cortex-M typically requires quantization."
+        )
+    else:
+        model = model.to(memory_format=torch.channels_last)
+        example_inputs = tuple(_to_channels_last(x) for x in example_inputs)
+
+        quantizer = CortexMQuantizer()
+        prepared = prepare_pt2e(model, quantizer)
+
+        dataset = get_calibration_data(
+            args.model_name, example_inputs, args.evaluate, args.evaluate_config
+        )
+
+        if isinstance(dataset, DataLoader):
+            for sample, _ in dataset:
+                prepared(_to_channels_last(sample))
+        else:
+            prepared(*tuple(_to_channels_last(x) for x in dataset))
+
+        model_quant = convert_pt2e(prepared)
+
+        exported_program = torch.export.export(
+            model_quant, example_inputs, strict=args.strict_export
+        )
+
+    edge = to_edge_transform_and_lower(
+        exported_program,
+        compile_config=EdgeCompileConfig(
+            preserve_ops=[
+                torch.ops.aten.linear.default,
+                torch.ops.aten.hardsigmoid.default,
+                torch.ops.aten.hardsigmoid_.default,
+                torch.ops.aten.hardswish.default,
+                torch.ops.aten.hardswish_.default,
+            ],
+            _check_ir_validity=False,
+        ),
+    )
+
+    pass_manager = CortexMPassManager(edge.exported_program())
+    edge._edge_programs["forward"] = pass_manager.transform()
+
+    return model_quant if args.quantize else None, edge
+
+
 def to_edge_no_delegate(
     exported_program: ExportedProgram,
     args,
@@ -873,7 +944,24 @@ if __name__ == "__main__":  # noqa: C901
 
     # Quantize if required
     model_quant = None
-    if args.delegate:
+    if args.target == "cortex-m55+int8":
+        # Cortex-M path: CMSIS-NN portable kernels, no delegation
+        if getattr(args, "evaluate", False):
+            logging.error(
+                "--evaluate is not supported for target 'cortex-m55+int8' "
+                "because this path does not use a TOSA delegate."
+            )
+            sys.exit(1)
+        if args.delegate:
+            logging.warning(
+                "--delegate is ignored for target 'cortex-m55+int8' "
+                "(this target does not use delegated ops)."
+            )
+            args.delegate = False
+        model_quant, edge = to_edge_cortex_m(
+            exported_program, args, model, example_inputs
+        )
+    elif args.delegate:
         model_quant, edge = to_edge_TOSA_delegate(
             exported_program, args, model, example_inputs
         )
--- a/backends/cortex_m/passes/convert_to_cortex_m_pass.py
+++ b/backends/cortex_m/passes/convert_to_cortex_m_pass.py
@@ -71,7 +71,7 @@ class ConvertToCortexMPass(XNNPACKPass):
 
     def _get_linear_replacement(self, node):
         """
-         Let
+        Let
         - yi be the output activations (y1, ... yn)
         - xj be the input activations (x1, ... xm)
         - wij be the weights (w11, ... wnm)


examples/arm/aot_arm_compiler.py
```diff
@@ -890,11 +975,6 @@ def transform_for_cortex_m_backend(edge_program_manager, args):
             exported_program, args, model, example_inputs
         )
 
-    # Cortex-m ops are never included in vgf or direct-drive
-    if args.target != "vgf" and not args.direct_drive:
```
psiddh: Should TOSA targets even have CortexM fallback ?   ( --target=u55/u85 → TOSA delegation)

AdrianLundell (changes requested):
Hi, this PR needs major changes I'm afraid. 

1. The changes to fold_qdq_with_annotated_qparams_pass and propagate_qparams_pass are very likely not needed, rather they are masking a faulty implementation either of the add_mm or the integration in the aot_arm_compiler.
2. The addition of the add_mm is a significant change which should be made in a separate PR properly tested with unittests as is done with all other ops. 
3. It would be great to add mv2 also as a pytest similar to mv3, in fact I would suggesting starting to get that working before adding support to the aot_arm_compiler since the compilation pipeline is guaranteed to be working there.

psiddh:
> Hi, this PR needs major changes I'm afraid.
> 
> 1. The changes to fold_qdq_with_annotated_qparams_pass and propagate_qparams_pass are very likely not needed, rather they are masking a faulty implementation either of the add_mm or the integration in the aot_arm_compiler.
> 2. The addition of the add_mm is a significant change which should be made in a separate PR properly tested with unittests as is done with all other ops.
> 3. It would be great to add mv2 also as a pytest similar to mv3, in fact I would suggesting starting to get that working before adding support to the aot_arm_compiler since the compilation pipeline is guaranteed to be working there.

Sure - I agree with the approach. I just wanted to share the work I've been up to recently so    
  that we can have exactly this kind of discussion.                                                                   
                                                                                                                      
  **Context on the design choice:**                                                                                       
                                                                                                                      
  The Cortex-M backend keeps addmm directly (vs ARM's decomposition to Conv2D) to leverage CMSIS-NN's optimized linear
   kernels. This creates a qparam propagation challenge:                                                              
                                                                                                                      
  When PyTorch decomposes nn.Linear to edge dialect:                                                                  
  linear(input, weight, bias) → addmm(bias, input, weight.T)                                                          
                                                                                                                      
  The weight flows through a transpose before reaching addmm:                                                         
  weight → permute_copy → addmm                                                                                       
                                                                                                                      
  FoldAndAnnotateQParamsPass folds the DQ into permute, but output_qparams remains empty (no Q node after permute).   
  The addmm node expects weight qparams at input_qparams[2], hence PropagateQParamsPass bridges this gap.             
                                                                                                                      
  Proposed approach:                                                                                                  
                                                                                                                      
  1. I'll first add test_addmm.py and test_mobilenet_v2.py unit tests following the existing patterns                 
  2. Once those pass and validate the pipeline, we can review whether PropagateQParamsPass is the right solution or if
   there's a cleaner approach                                                                                         
  3. The aot_arm_compiler integration can follow in a subsequent PR                                                   
                                                                                                                      
  This way we have proper test coverage before discussing the implementation details. Let me get the unit tests       
  working first.

AdrianLundell:
Sounds good!

> When PyTorch decomposes nn.Linear to edge dialect:
> linear(input, weight, bias) → addmm(bias, input, weight.T)
> 
> The weight flows through a transpose before reaching addmm:
> weight → permute_copy → addmm
> 

I think the issue here is that you are not using the EdgeCompileConfig used in the tester:
```
 config = EdgeCompileConfig(
            preserve_ops=[
                torch.ops.aten.linear.default,
                torch.ops.aten.hardsigmoid.default,
                torch.ops.aten.hardsigmoid_.default,
                torch.ops.aten.hardswish.default,
                torch.ops.aten.hardswish_.default,
            ],
            _check_ir_validity=False,
            _core_aten_ops_exception_list=[torch.ops.aten.max_pool2d.default],
        ) 
```

When linear is not decomposed you avoid the issues around q/dq folding. In general the design philosophy is that we want to make the decompositions and annotations to get correct q/dq values directly rather than handling special cases in the folding, as that gets complex very quickly from our previous experience in the arm backend.

examples/arm/aot_arm_compiler.py
```diff
@@ -793,6 +784,83 @@ def to_edge_TOSA_delegate(
...
+    )
+
+    pass_instances = []
+    for pass_cls in CortexMPassManager.pass_list:
+        sig = inspect.signature(pass_cls.__init__)
+        if "exported_program" in sig.parameters:
+            pass_instances.append(pass_cls(edge.exported_program()))
+        else:
+            pass_instances.append(pass_cls())
+    edge = edge.transform(pass_instances)
```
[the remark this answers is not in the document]
psiddh: Already tried the CortexMPassManager approach, and it broke things — CortexMPassManager.transform() returns an ExportedProgram, not an EdgeProgramManager. The edge object was left untransformed, resulting in 351 raw aten ops. 

The cleanest approach: add an instantiate_passes method to CortexMPassManager that extracts the inspect logic into a reusable method. Then both transform() and to_edge_cortex_m() can use it without duplication, which can be a follow up PR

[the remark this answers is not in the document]
AdrianLundell: If you want a one-liner to do this in the meanwhile take a look in the cortex-m jupyter notebook example and see how it is done there!

examples/arm/aot_arm_compiler.py
AdrianLundell: We want the Q/DQ fallbacks when enabled, other fallbacks would be nice at sometime but I don't think we are ready for it yet

examples/arm/aot_arm_compiler.py
```diff
@@ -793,6 +784,83 @@ def to_edge_TOSA_delegate(
...
+    """Cortex-M/CMSIS-NN compilation path with no delegation."""
+    logging.info("Using Cortex-M/CMSIS-NN compilation path (no delegation)")
+
+    model_quant = None
+
+    if args.quantize:
+        model_channels_last = model.to(memory_format=torch.channels_last)
+        example_inputs_cl = tuple(
+            x.to(memory_format=torch.channels_last) if x.dim() == 4 else x
+            for x in example_inputs
```
AdrianLundell: Should we also check for non-tensor inputs here?

AdrianLundell: Also maybe it is enough to warn if the dataset is not channels last but let the user convert it, always better to have less undocumented magic behaviour in my opinion.

examples/arm/aot_arm_compiler.py
```diff
@@ -793,6 +784,83 @@ def to_edge_TOSA_delegate(
...
+
+        exported_program = torch.export.export(
+            model_quant, example_inputs_cl, strict=args.strict_export
+        )
+    else:
+        logging.warning(
+            "Quantization is DISABLED. Cortex-M typically requires quantization."
+        )
+
+    edge = to_edge_transform_and_lower(
```
AdrianLundell, line 849: I think now might be a good time to create a default CortexMEdgeCompileConfig with these options which is used in both the pytests, jupyter notebook and here, to avoid code duplication.

examples/arm/aot_arm_compiler.py
```diff
@@ -793,6 +784,83 @@ def to_edge_TOSA_delegate(
...
+    args,
+    model: GraphModule,
+    example_inputs: Tuple[torch.Tensor],
+):
+    """Cortex-M/CMSIS-NN compilation path with no delegation."""
+    logging.info("Using Cortex-M/CMSIS-NN compilation path (no delegation)")
+
+    model_quant = None
+
+    if args.quantize:
```
AdrianLundell: Nit: I prefer to not use global state variables inside functions

examples/arm/aot_arm_compiler.py
```diff
@@ -396,6 +386,7 @@ def forward(self, x):
     "TOSA-1.0+INT",
     "TOSA-1.0+FP",
     "TOSA-1.0+INT+int16",
+    "cortex-m",
```
AdrianLundell: cortex-m55? To let us add support other cortex-m cpus cleanly in the future.

AdrianLundell (commented):
Looks good overall, just a few comments

examples/arm/aot_arm_compiler.py
rascani: Or maybe `cortex-mve` or `cortex-m55-mve` since MVE is optional?

examples/arm/aot_arm_compiler.py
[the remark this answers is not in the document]
psiddh: Replaced with notebook-style 2-liner as suggested. Removed import inspect.

examples/arm/aot_arm_compiler.py
psiddh: > Should we also check for non-tensor inputs here?

_to_channels_last now recurses into nested tuples, handling models like LSTM that return (tensor, (tensor,    
  tensor)) as example inputs.  



> Also maybe it is enough to warn if the dataset is not channels last but let the user convert it, always better to have less undocumented magic behaviour in my opinion.

 Added non-tensor input handling and logging.warning when conversion happens. 

 aot_arm_compiler , a CLI tool where the user has no control over say, get_example_inputs() format, so this converts and warn rather than complete fail. What do you think ?

examples/arm/aot_arm_compiler.py
psiddh: Agree,  Will do in follow-up focussed PR — the shared config needs to be created in backends/cortex_m/, then updated in three places (aot_arm_compiler, tester.py, notebook).

examples/arm/aot_arm_compiler.py
psiddh: Changed so model_quant is only assigned where used, no pre-initialization.

examples/arm/aot_arm_compiler.py
psiddh: **Target naming** (cortex-m → cortex-m-int8):

cortex-m-int8 follows the existing  TOSA-1.0+INT / TOSA-1.0+FP pattern — backend + data type. The .pte , I think is, processor-agnostic (same file runs on M55, M85, M4, M0 via CMSIS-NN), so a processor-specific suffix like cortex-m55 would be very specific , isn't ?  

- m55 excludes M85, which has the same MVE capability
- mve excludes M4/M7/M33, which run the same .pte fine using DSP or scalar fallback
- This cortex-m-int8 suffix leaves room for cortex-m-int16 later

What do you both think ?

examples/arm/aot_arm_compiler.py
```diff
@@ -795,6 +785,75 @@ def to_edge_TOSA_delegate(
...
+                torch.ops.aten.hardsigmoid_.default,
+                torch.ops.aten.hardswish.default,
+                torch.ops.aten.hardswish_.default,
+            ],
+            _check_ir_validity=False,
+        ),
+    )
+
+    pass_manager = CortexMPassManager(edge.exported_program())
+    edge._edge_programs["forward"] = pass_manager.transform()
```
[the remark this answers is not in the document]
psiddh: This suggestion  would pass the already-transformed     
  ExportedProgram back through to_edge_transform_and_lower(), which re-runs edge decomposition on a graph that's already been lowered.  Leaving the current impl as-is to be consistent with Cortex-M 
  notebook, tester.py

examples/arm/aot_arm_compiler.py
AdrianLundell: I thought that the .pte file will not be processor-specific since we have slightly different support for different M-classes and since we allocate scratch buffers in the pte which also varies. We can have one target for generating a processor-agnostic .pte as well although that will be less optimized.

examples/arm/aot_arm_compiler.py
```diff
@@ -396,6 +388,7 @@ def forward(self, x):
     "TOSA-1.0+INT",
     "TOSA-1.0+FP",
     "TOSA-1.0+INT+int16",
+    "cortex-m-int8",
```
zingo: Thanks for adding this, I think the flag might need to be different for different cortex-m as the memory planning might differentiate depending on impementation, so maybe it should be cortex-m55* for now?  Also the other target use +<TYPE> so it might be more consistent with "cortex-m55+int8" ?

Also we need to update  the two M55 examples (Corstone-300 and NXP N6) in 
https://github.com/pytorch/executorch/tree/main/zephyr
to use this new nice flag :)  We can maybe also do that in a separate PR also.

psiddh: Updated the flag, and will update the M55 examples in follow up PR

AdrianLundell (approved):

zingo:
Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not  :pray:  .

psiddh:
> Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .

Are you referring to the CI failure on this PR: https://github.com/pytorch/executorch/actions/runs/22315191858/job/64557957144?pr=17075   . If so, it seems unrelated and it is failing due to size check.

> Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .

Looked at the failing test-arm-backend* fail jobs, Looks unrelated to me. I will keep a close eye on the CI post landing.
I think if CI tests are using 'qdq' flag which was introduced at the very beginning of CortexM development 6 months ago, they will fail. But I don't think there are any CI jobs atm, that use this flag afaia. Will keep a close eye on CI

The failed job for Zephyr :
- 2026-02-23T16:50:39.0579191Z E [executorch:operator_registry.cpp:256 get_op_function_from_registry()] kernel 'quantized_decomposed::dequantize_per_tensor.out' not found.

zingo:
> > Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .
> 
> Are you referring to the CI failure on this PR: https://github.com/pytorch/executorch/actions/runs/22315191858/job/64557957144?pr=17075 . If so, it seems unrelated and it is failing due to size check.
> 
> > Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .
> 
> Looked at the failing test-arm-backend* fail jobs, Looks unrelated to me. I will keep a close eye on the CI post landing. I think if CI tests are using 'qdq' flag which was introduced at the very beginning of CortexM development 6 months ago, they will fail. But I don't think there are any CI jobs atm, that use this flag afaia. Will keep a close eye on CI
> 
> The failed job for Zephyr :
> 
> * 2026-02-23T16:50:39.0579191Z E [executorch:operator_registry.cpp:256 get_op_function_from_registry()] kernel 'quantized_decomposed::dequantize_per_tensor.out' not found.

I agree I also think they are unrelated, usually I would just rebase the PR to retrigger a rerun on the latest to make sure but I think it might mess up your internal-external Meta PR sync. Thats why I just added a comment :)
We just need to keep an eye the hud after it is merge to make sure the errors does not start to appare  there :)  e.g. https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=test-arm%7Ccortex-m&useRegexFilter=true

I can try to trigger a re-test without a rebase, that might work.

psiddh:
> > > Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .
> > 
> > 
> > Are you referring to the CI failure on this PR: https://github.com/pytorch/executorch/actions/runs/22315191858/job/64557957144?pr=17075 . If so, it seems unrelated and it is failing due to size check.
> > > Hi @psiddh this seem to give some errors is test-arm-backed-* test runners it might be (probably is) unrelated but take care before merging as it will break CI if not 🙏 .
> > 
> > 
> > Looked at the failing test-arm-backend* fail jobs, Looks unrelated to me. I will keep a close eye on the CI post landing. I think if CI tests are using 'qdq' flag which was introduced at the very beginning of CortexM development 6 months ago, they will fail. But I don't think there are any CI jobs atm, that use this flag afaia. Will keep a close eye on CI
> > The failed job for Zephyr :
> > 
> > * 2026-02-23T16:50:39.0579191Z E [executorch:operator_registry.cpp:256 get_op_function_from_registry()] kernel 'quantized_decomposed::dequantize_per_tensor.out' not found.
> 
> I agrre I also think they are unrelated, usually I would just rebase the PR to retrigger a rerun on the latest to make sure but I think it might mess up your internal-external Meta PR sync. Thats why I just added a comment :) We just need to keep an eye the hud after it is merge to make sure the errors does not start to appare there :) e.g. https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=test-arm%7Ccortex-m&useRegexFilter=true
> 
> I can try to trigger a re-test without a rebase, that might work.

Rebased to the HEAD

zingo:
PTE size seems to still fail and might actually be bigger, maybe strings are longer, orse some other metadata bigger.  it seems the number in the test need to be bumpt to not fail CI.

psiddh:
> PTE size seems to still fail and might actually be bigger, maybe strings are longer, orse some other metadata bigger. it seems the number in the test need to be bumpt to not fail CI.

https://github.com/pytorch/executorch/pull/17725

examples/arm/aot_arm_compiler.py
```diff
@@ -885,11 +929,6 @@ def transform_for_cortex_m_backend(edge_program_manager, args):
             exported_program, args, model, example_inputs
         )
 
-    # Cortex-m ops are never included in vgf or direct-drive
```
rascani: I wonder if this removal increased the size

zingo:
Hi there also seem to be a real problem with ZephyrOS tests:

I see this in the logs in 2 of them
https://github.com/pytorch/executorch/actions/runs/22425903988/job/64933843809?pr=17075 and https://github.com/pytorch/executorch/actions/runs/22425903988/job/64933843807?pr=17075

```
...
Thu, 26 Feb 2026 03:22:53 GMT E [executorch:method.cpp:784 resolve_operator()] Missing operator: [0] quantized_decomposed::quantize_per_tensor.out
Thu, 26 Feb 2026 03:22:53 GMT E [executorch:operator_registry.cpp:256 get_op_function_from_registry()] kernel 'quantized_decomposed::dequantize_per_tensor.out' not found.
...
Thu, 26 Feb 2026 03:22:53 GMT I [executorch:operator_registry.cpp:257 get_op_function_from_registry()] 0,
Thu, 26 Feb 2026 05:03:31 GMT Error: The operation was canceled.
```
As the Corstone FVP does not quite on this errors the jobs time times out instead.

So maybe something needs to be added to the cmake/link rules now for cortex-m?

In the last Zephyr failed job https://github.com/pytorch/executorch/actions/runs/22425903988/job/64933843797?pr=17075

I see 
```
Thu, 26 Feb 2026 03:22:08 GMT                   from /pytorch/executorch/zephyr_scratch/modules/lib/executorch/../executorch/runtime/core/exec_aten/exec_aten.h:36,
Thu, 26 Feb 2026 03:22:08 GMT                   from /pytorch/executorch/zephyr_scratch/modules/lib/executorch/../executorch/runtime/core/evalue.h:10,
Thu, 26 Feb 2026 03:22:08 GMT                   from /pytorch/executorch/zephyr_scratch/modules/lib/executorch/../executorch/runtime/executor/method.h:18,
Thu, 26 Feb 2026 05:03:31 GMT  Error: The operation was canceled.
```
e.g. time out like the other 2 BUT in a strange place, I suspect it could be the same but logs are not flushed. Lets see itfthis also goes away with the same fix.

zingo:
> I agree I also think they are unrelated...

Sorry that I was wrong and maybe caused hope and maybe more work and unneeded delays for this PR .

psiddh:
I think with Zephyr test failures with (thinking aloud here), 
  elif args.delegate → to_edge_TOSA_delegate(), though100% of ops are delegated to the NPU, the input/output boundary QDQ nodes are still in the graph as quantized_decomposed::*. No ReplaceQuantNodesPass to convert them. and hence Missing kernel at runtime ?

psiddh:
> I think with Zephyr test failures with (thinking aloud here), elif args.delegate → to_edge_TOSA_delegate(), though100% of ops are delegated to the NPU, the input/output boundary QDQ nodes are still in the graph as quantized_decomposed::*. No ReplaceQuantNodesPass to convert them. and hence Missing kernel at runtime ?

Created this clean PR to remove all legacy flow:  #17740 , Lets see if the CI passes clean on this PR. If it does, then I will rebase this ongoing PR on top of it

psiddh:
Unrelated CI failures, no Zephyr or arm failures , Merging the PR now

PR #17076, opened 2026-01-31
Metal backend: test modules

manuelcandales:
This pull request adds new support for building and testing the Metal backend modules on macOS, including automation in the CI workflow and a dedicated test runner script. The main changes are the addition of a reusable shell script for building and running Metal backend tests, and the integration of a new GitHub Actions job to automate module testing for the Metal backend.

**CI/CD and Testing Automation:**

* Added a new `test-metal-backend-modules` job to the `.github/workflows/metal.yml` workflow, which runs on a macOS M2 runner, sets up the environment, builds the Metal runtime, and runs Python unit tests for the Metal backend modules.

**Tooling and Scripts:**

* Introduced a new script `backends/apple/metal/tests/run_metal_test.sh` that provides commands to build the Metal runtime, check if it is built, and run inference with given model files. This script is now used in the CI workflow to automate Metal backend testing.


manuelcandales:
* #17237
* #17236
* #17235
* #17188
* #17187
* #17186
* #17127
* #17118
* #17117
* #17116
* #17115
* __->__ #17076

mergennachin (commented):
What are your thoughts on reusing Backend Test Harness? 

https://github.com/pytorch/executorch/tree/main/backends/test/suite

test_modules.py almost look like this: 

https://github.com/pytorch/executorch/tree/main/backends/test/suite/operators

I think we should try to reuse as much as possible, and if possible contribute back to test harness

cc @GregoryComer

manuelcandales:
> What are your thoughts on reusing Backend Test Harness?

Every backend has its own tests, independent of our more comprehensive testing frameworks. E.g. [xnnpack](https://github.com/pytorch/executorch/tree/main/backends/xnnpack/test), [vulkan](https://github.com/pytorch/executorch/tree/main/backends/vulkan/test), [qualcomm](https://github.com/pytorch/executorch/tree/main/backends/qualcomm/tests), [arm](https://github.com/pytorch/executorch/tree/main/backends/arm/test)

The intent of frameworks like FACTO or Backend Test Harness is to find unknown issues, on mature operators/backends where certain level of coverage and correctness is expected.

On the other hand, having the ability to add specific tests remains useful, as we develop and want to verify that a given PR intended to enable certain feature, is indeed providing the desired functionality.

It is also a minimum bar that the backend/operator should pass, and that will be enforced on CI.

mergennachin:
Discussed offline with @manuelcandales 

Agreed that we need both approaches (backend specific unit testing, and backend test harness)

As discussed, let's remove unnecessary tests that's not testing what you're developing right now and could be enabled by backend test harness in the future.

mergennachin (approved):

backends/apple/metal/tests/run_metal_test.sh
```diff
@@ -0,0 +1,119 @@
...
+    CMAKE_ARGS="-DEXECUTORCH_BUILD_METAL=ON \
+                -DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
+                -DEXECUTORCH_BUILD_EXECUTOR_RUNNER=ON \
+                -DEXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR=ON \
+                -DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
+                -DEXECUTORCH_BUILD_EXTENSION_MODULE=ON \
+                -DEXECUTORCH_BUILD_EXTENSION_NAMED_DATA_MAP=ON \
+                -DAOTI_METAL=ON \
+                -DEXECUTORCH_LOG_LEVEL=Info \
+                -DCMAKE_BUILD_TYPE=Release"
```
larryliu0820, line 56: why not use preset? `cmake --workflow --preset llm-metal-stats`

larryliu0820: If not llm specific, also strongly recommend to add a base preset `metal-release` and let `llm-metal-stats` extend it.

backends/apple/metal/tests/run_metal_test.sh
```diff
@@ -0,0 +1,119 @@
...
+
+    if [[ ! -f "$pte_path" ]]; then
+        echo "Error: PTE file not found: $pte_path"
+        exit 1
+    fi
+
+    echo "Running inference..."
+    echo "  PTE: $pte_path"
+
+    "$EXECUTOR_RUNNER" --model_path "$pte_path"
```
larryliu0820, line 92: it would be much better to add pybind and run tests in pure python environment. This way you don't need this script.

manuelcandales: I am using this script in CI as well (for building and updating ao).

PR #17014, opened 2026-01-29
Revert "Qualcomm AI Engine Direct - Optimize UT execution time (#16865)"

GregoryComer:
This reverts commit b6b940a17faac1e642f899f1f7a6b885c390314d.

The change is breaking internal simulator tests. Happy to help re-land and investigate, but reverting to fix internal CI.

PR #17086, opened 2026-02-02
Qualcomm AI Engine Direct - Optimize UT execution time re-submit

chenweng-quic:
### Summary
- Currently all tests will push same libs, which is redundant. With this PR it only push once, and reduce execution time from two to three times for operator's tests.

### Test plan
```
python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNFloatingPointOperator --device <device> --host <host> --model SM8650 --build_folder build-android --executorch_root . --artifact all_artifact
```
without optimization
image

with optimization
image

cc @cccclai @cbilgin

### DIFF
--- a/backends/qualcomm/tests/test_qnn_delegate.py
+++ b/backends/qualcomm/tests/test_qnn_delegate.py
@@ -8183,8 +8183,15 @@ class TestUtilsScript(TestQNN):
                 self.build_folder,
                 "--input_list",
                 f"{tmp_dir}/input_list",
+                "--model",
+                self.model,
+                "--host",
+                self.host,
+                "--target",
+                self.target,
+                "--device",
+                self.device,
             ]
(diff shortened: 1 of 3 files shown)


mergennachin:
@lucylq 

Can you import internally and see if it fixed the internal CI tests. 

Last week, @GregoryComer reverted (https://github.com/pytorch/executorch/pull/17014) due to internal CI failures

mergennachin:
@GregoryComer when reverting, it would be good to provide more information about what the error was in internal CI tests. We can't share everything but providing something would be good than nothing.

cccclai:
can you rebase to main?

cccclai (approved):
I don't see the internal CI failure and will try to merge again. @GregoryComer feel free to chime in

PR #17088, opened 2026-02-02
Arm backend: Add FP16 support to operators pt.2

martinlsm:
 - conv2d
 - conv3d
 - cos
 - slice
 - pad

Also correct the test name violations seen in test_conv2d.py.


zingo (approved):

zingo:

PR #17091, opened 2026-02-02
Arm backend: Use %lu instead of %zu in logs

zingo:
### Summary
Usuing %zu have different problem depending on compiler flags and clib used, as this
is just example code it better to stick to %lu and make it easier to port as 
it seem to work better between different compiler flags and different clib setups.


oscarandersson8218 (approved):

zingo:
This fail on test_nss_model after merged  this is expected and unrelated, a separate fix is on it's way https://github.com/pytorch/executorch/pull/17200

PR #17092, opened 2026-02-02
Use cmath functions in kernels/portable

zingo:


zingo:
Hi @SS-JIA here is a generic PR that is the result of a review comment from @mergennachin in https://github.com/pytorch/executorch/pull/16340 to handle more cases where std:: was missing :)

SS-JIA (approved):
LGTM; cc: @manuelcandales since this is related to kernel libraries

mergennachin (approved):
Thank you for following up

PR #17093, opened 2026-02-02
Cortex-M backend: Break out pattern_matcher from quantizer

AdrianLundell:
This change completes the split between quantization configuration, e.g. which ops to quantize in which way and what is supported by the backend.

The CortexMQuantizer now completely specifies which patterns it supports for quantization in the CORTEX_M_QUANTIZER_SUPPORT_DICT. Each key in this dict specifies one pattern which is supported, and a PatternChecker which filters out certain configurations of the pattern which are not supported. (E.g. dim_orders, parameters values)

This dict is consumed by a new pattern-matcher which partitions all nodes given by the node_finder into supported patterns.


backends/cortex_m/quantizer/pattern_checkers.py
```diff
@@ -0,0 +1,250 @@
...
+class CortexMSoftmaxCheck(PatternCheck):
+
+    @classmethod
+    def check_pattern(cls, pattern):
+        """
+        Checks that given the tensor must either
+        - be contiguous (default layout) with softmax dim == last logical dim, or
+        - be channels_last with softmax dim == channel dim.
+        """
+        assert len(pattern) == 1
```
psiddh, line 147: This is good for debug builds, should we retunr some false or something in release builds explicitly ?

backends/cortex_m/quantizer/pattern_checkers.py
```diff
@@ -0,0 +1,250 @@
...
+                return False
+
+        return True
+
+    @classmethod
+    def check_quantization_config(cls, quantization_config):
+        """
+        Checks that the quantization config uses per-tensor int8 quantization.
+        """
+        is_int8 = (
```
psiddh: Every checker class has nearly identical int8 validation:                                                                                                                                                          
  is_int8 = (                                                                                                                                                                                                        
      quantization_config.input_activation.dtype == torch.int8                                                                                                                                                       
      and quantization_config.output_activation.dtype == torch.int8                                                                                                                                                  
  )                                                                                                                                                                                                                  
  This could be extracted to a helper method in the PatternCheck base class.

backends/cortex_m/quantizer/node_finders.py
```diff
@@ -35,7 +35,7 @@ class GlobalNodeFinder(NodeFinder):
     """
 
     def find_nodes(self, model: GraphModule) -> Iterator[Node]:
-        return model.graph.nodes
+        return (n for n in model.graph.nodes)
```
psiddh, line 38: All callers of this api / calling site, should ensure  that no existing code iterates over the result more than once , since it returns a generator now instead of list

psiddh (approved):
Overall solid refactoring PR, +1 left few minor comments .

backends/cortex_m/quantizer/pattern_checkers.py
AdrianLundell: In my view asserts is a nice way to check internal logic, in release builds a softmax pattern of length != 1 should never appear and so we don't need to handle it. But we can consider it and handle in a later patch if we find it useful!

backends/cortex_m/quantizer/pattern_checkers.py
AdrianLundell: Good idea, done

backends/cortex_m/quantizer/node_finders.py
AdrianLundell: Yes, I think it is an improvement to have the same behaviour from all NodeFinders, we just have to make sure to keep this in mind.

psiddh (approved):

PR #17095, opened 2026-02-02
Arm backend: Add FP16 support to operators pt.3

martinlsm:
 - eq
 - erf
 - exp
 - floor
 - depthwise conv


zingo (approved):

zingo:

PR #17109, opened 2026-02-02
Arm backend: Add STABLE softmax decomposition config for Ethos-U

JakeStevens:
Summary:
Current behavior for U55 defaults to UNSTABLE. It seems like this is because mask is not supported on U55, but there does not seem to be an inherent need to couple this. Defaulting to UNSTABLE negatively impacts quantization performance.

This PR adds a new `STABLE` option to `SoftmaxDecompositionConfig` that provides numerically stable softmax decomposition without masked fill decomposition.

The three softmax configs now behave as follows:
- `MASKED`: Stable softmax (with amax subtraction) + masked fill decomposition
- `UNSTABLE`: Unstable softmax (no amax subtraction), no masked fill decomposition
- `STABLE`: Stable softmax (with amax subtraction), no masked fill decomposition

For Ethos-U55 targets, `disable_masked_softmax()` now sets the config to `STABLE` instead of `UNSTABLE`, providing numerically stable softmax while avoiding masked fill decomposition which is not needed for these targets.

Differential Revision: D92058235

### DIFF
--- a/backends/arm/_passes/arm_pass_manager.py
+++ b/backends/arm/_passes/arm_pass_manager.py
@@ -174,6 +174,9 @@ class ArmPassManager(PassManager):
             case SoftmaxDecompositionConfig.UNSTABLE:
                 skip_set.add(DecomposeSoftmaxPass)
                 skip_set.add(DecomposeMaskedFillPass)
+            case SoftmaxDecompositionConfig.STABLE:
+                skip_set.add(DecomposeSoftmaxUnstablePass)
+                skip_set.add(DecomposeMaskedFillPass)
 
         if config.fuse_duplicate_users is FuseDuplicateUsersConfig.DISABLED:
             skip_set.add(FuseDuplicateUsersPass)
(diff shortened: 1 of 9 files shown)


JakeStevens:
I see amax is not supported on u55. Is just max supported? Is there really no support for a stable softmax?

AdrianLundell:
Hi, thanks for your PR. Amax support was added for Ethos-U55 just recently with the latest Vela update, so it is just a matter of enabling it in our backend and then the stable softmax decomposition should be available also for U55. I'll start working on it.

JakeStevens:
> Hi, thanks for your PR. Amax support was added for Ethos-U55 just recently with the latest Vela update, so it is just a matter of enabling it in our backend and then the stable softmax decomposition should be available also for U55. I'll start working on it.

Great, I have created:

https://github.com/pytorch/executorch/issues/17211

to help track.

I will wait on this PR until then.

Thanks!

JakeStevens:
@AdrianLundell  @zingo  what do you think?

I've had teammates confirm this solved issues with QAT and the unstable version

AdrianLundell (approved):
I like it, thanks for creating tests as well.

PR #17123, opened 2026-02-02
Shim LI op defs in OSS

GregoryComer:
### Summary
Buck unit test jobs are failing since the TARGETS file migration due to xnnpack/buck/test/BUCK referencing an internal target. The specific dependency (ATen quant ops) should be available by default in OSS, so I'm just gating this and adding a no-op definition of get_pt_ops_deps in shim_et.

### Test Plan
Buck tests are passing on OSS CI.

### DIFF
--- a/backends/xnnpack/test/BUCK
+++ b/backends/xnnpack/test/BUCK
@@ -8,6 +8,7 @@ load(
     "@fbsource//xplat/caffe2:pt_defs.bzl",
     "get_pt_ops_deps",
 )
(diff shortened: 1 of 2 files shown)


JacobSzwejbka (approved):

lucylq (approved):

kirklandsign (approved):

GregoryComer:
I'm testing the internal build to make sure I didn't break anything. Will land once it's green.

PR #17124, opened 2026-02-02
Fix kernel tests to respect build configuration options

mergennachin:
The kernel tests in kernels/test/CMakeLists.txt unconditionally depended
on targets that only exist when certain build options are enabled:
- optimized_native_cpu_ops_lib requires EXECUTORCH_BUILD_KERNELS_OPTIMIZED
- optimized_portable_kernels requires EXECUTORCH_BUILD_PTHREADPOOL and
  EXECUTORCH_BUILD_KERNELS_OPTIMIZED
- quantized_kernels requires EXECUTORCH_BUILD_KERNELS_QUANTIZED

The fix makes the kernel list and test targets conditional on the
corresponding build options being enabled.


SS-JIA (approved):
LGTM

PR #17130, opened 2026-02-03
Disable smollm2 on ethos-u

GregoryComer:
### Summary
This model is too large to fit on Ethos-U. After discussion with @psiddh, sounds like the best call is to disable it for now. It was previously working only because the weights were all zeros, allowing it to be compressed and fit. With recent changes to not zero weights, the job is failing on trunk.


JacobSzwejbka (approved):

psiddh:
Long-term fix is to use a tiny LLM-like model designed for CI that fits on Ethos-U with realistic weights.

psiddh (approved):

PR #17137, opened 2026-02-03
Add DecomposeGruPass for ARM backend (#17137)

apullin:
Summary:

Adds a decomposition pass that transforms aten.gru.input into elementary
ops supported by TOSA (matmul, sigmoid, tanh, mul, add, slice, cat).

GRU cell equations per timestep:
    r_t = sigmoid(x_t @ W_ir.T + b_ir + h_{t-1} @ W_hr.T + b_hr)
    z_t = sigmoid(x_t @ W_iz.T + b_iz + h_{t-1} @ W_hz.T + b_hz)
    n_t = tanh(x_t @ W_in.T + b_in + r_t * (h_{t-1} @ W_hn.T + b_hn))
    h_t = n_t + z_t * (h_{t-1} - n_t)

Features:
- Multi-layer GRU support
- Bidirectional GRU support
- With/without bias
- batch_first support
- Batched gate computation (2 mm ops per timestep instead of 6)

Differential Revision: D92058313

### DIFF
--- a/backends/arm/_passes/__init__.py
+++ b/backends/arm/_passes/__init__.py
@@ -53,6 +53,7 @@ from .decompose_gelu_pass import DecomposeGeluPass  # noqa
 from .decompose_glu_pass import DecomposeGluPass  # noqa
 from .decompose_grouped_conv_pass import DecomposeGroupedConvPass  # noqa
 from .decompose_groupnorm_pass import DecomposeGroupNormPass  # noqa
+from .decompose_gru_pass import DecomposeGruPass  # noqa
 from .decompose_index_copy_pass import DecomposeIndexCopyPass  # noqa
 from .decompose_index_select_to_gather_pass import (  # noqa
     DecomposeIndexSelectToGatherPass,
--- a/backends/arm/_passes/arm_pass_manager.py
+++ b/backends/arm/_passes/arm_pass_manager.py
@@ -61,6 +61,7 @@ from executorch.backends.arm._passes import (
     DecomposeGluPass,
     DecomposeGroupedConvPass,
     DecomposeGroupNormPass,
+    DecomposeGruPass,
     DecomposeIndexCopyPass,
     DecomposeIndexSelectToGatherPass,
     DecomposeIndexTensorToGatherPass,
@@ -365,6 +366,7 @@ class ArmPassManager(PassManager):
                 ConvertToClampPass(),
                 DecomposeTOSAUnsupportedClampPass(),
                 DecomposeGroupNormPass(),
+                DecomposeGruPass(),
                 DecomposeLayerNormPass(),
                 DecomposeVarPass(),
                 DecomposeMeanDimPass(exported_program.graph_module, self.tosa_spec),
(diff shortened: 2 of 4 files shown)


backends/arm/test/models/test_nss.py
```diff
@@ -16,9 +15,11 @@
     TosaPipelineINT,
     VgfPipeline,
 )
-
 from huggingface_hub import hf_hub_download
 
+# Skip entire module if ng_model_gym is not installed (requires pydantic)
+pytest.importorskip("ng_model_gym")
```
mansnils: I think this issue has been resolved, so hopefully this skip should not be needed. Maybe rebase?

JakeStevens:
We should use something like

run_and_compare_to_initial_model

to make sure this decomposition is numerically equivalent

AdrianLundell:
> We should use something like run_and_compare_to_initial_model to make sure this decomposition is numerically equivalent

We are working on adding this, I can tag you when the PR is up.

AdrianLundell:
Hi @JakeStevens, this PR adds something similar what you describe to the tosa_INT test pipeline:  #17473

JakeStevens:
@apullin  can you update to use the new utility functions?

the decompose LGTM and seemed to work internally but I did not verify numerical closeness

JakeStevens:
also have some lint issues

apullin:
@AdrianLundell @martinlsm I think we re seeking an approval from someone inside ARM for a PR like this. Please have a look.

AdrianLundell:
I'm a bit confused it seems like this is added also in #17140? And did you look into what @gggekov said about using torch._decomp.get_decompositions?

apullin:
> added also in #17140
@AdrianLundell 
This is my mistake - I have several changes in a stack, and PR's got exported for each one, although they are 3 instances of the same technique, adding explicte decomp passes for common RNN layers.

See comment here: 
https://github.com/pytorch/executorch/pull/17140#issuecomment-3973910869

If we want to hit all 3, yes, #17140 would have that. instead of #17137 (this) and #17139 separately. Please advise.

JakeStevens:
per @gggekov comment on a different pull request:

> apologies for the radio silence over the last weeks. Happy to go with your approach of manually decomposing rather than the get_decompositions route. I see you have a few failing tests and a merge conflict, could you resolve these ? Your patch looks good to me, I haven't verified the numerics of the decomposition, I assume they must be alright since you pass the unit tests. It would be great to have the LSTM, GRU & RNN available in the Arm backend.


based on this and internal testing LGTM

JakeStevens (approved):
you need to rebase and resolve merge conflicts

apullin:
> you need to rebase and resolve merge conflicts

Done now, and the other failed test is expected to be green now. However - do we want to approve this one, or #17140, which includes the RNN & LSTM support, following the same pattern?

AdrianLundell:
I would say go with #17140 and get everything at the same time

apullin:
Closing. Changes will be done all together in #17140.

PR #17138, opened 2026-02-03
Arm backend: Disable _sort_outputs in TOSA backend (MLETORCH-1266)

apullin:
Summary:
Disables the _sort_outputs call in the TOSA backend lowering.

The _sort_outputs function reorders TOSA flatbuffer outputs, but the
runner reads output shapes from original_module (a pre-preprocess copy),
creating a mismatch between expected and actual output order.

The TOSA reference model preserves flatbuffer order, so no sorting is
actually needed. This change prevents the mismatch issue.

When output_order_workaround is enabled, the code now logs a debug
message indicating the option is set but _sort_outputs is disabled.
---
> Generated by [Confucius Code Assist (CCA)](https://www.internalfb.com/wiki/Confucius/Analect/Shared_Analects/Confucius_Code_Assist_(CCA)/)
[Confucius Session](https://www.internalfb.com/confucius?host=62602.od.fbinfra.net&port=8086&tab=Chat&session_id=e1d1ac52-0014-11f1-9d55-75b7d4e71d8a&entry_name=Code+Assist), [Trace](https://www.internalfb.com/confucius?session_id=e1d1ac52-0014-11f1-9d55-75b7d4e71d8a&tab=Trace)

Differential Revision: D92059044

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai

PR #17140, opened 2026-02-03
Add DecomposeLstmPass for ARM backend (#17140)

apullin:
Summary:

Adds a decomposition pass that transforms aten.lstm.input into elementary
ops supported by TOSA (matmul, sigmoid, tanh, mul, add, slice, cat).

LSTM cell equations per timestep:
    i_t = sigmoid(x_t @ W_ii.T + b_ii + h_{t-1} @ W_hi.T + b_hi)
    f_t = sigmoid(x_t @ W_if.T + b_if + h_{t-1} @ W_hf.T + b_hf)
    g_t = tanh(x_t @ W_ig.T + b_ig + h_{t-1} @ W_hg.T + b_hg)
    o_t = sigmoid(x_t @ W_io.T + b_io + h_{t-1} @ W_ho.T + b_ho)
    c_t = f_t * c_{t-1} + i_t * g_t
    h_t = o_t * tanh(c_t)

Features:
- Multi-layer LSTM support
- Bidirectional LSTM support
- With/without bias
- batch_first support
- Batched gate computation (2 mm ops per timestep instead of 8 )

Differential Revision: D92059277

### DIFF
--- a/backends/arm/_passes/decompose_gru_pass.py
+++ b/backends/arm/_passes/decompose_gru_pass.py
@@ -1,0 +1,335 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import operator
+from typing import List, Set, Tuple, Type
+
+import torch
+from executorch.backends.arm._passes.arm_pass import ArmPass
+from executorch.backends.arm._passes.arm_pass_utils import create_node
+from executorch.backends.arm._passes.insert_table_ops import InsertTableOpsPass
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class DecomposeGruPass(ArmPass):
+    """Decomposes aten.gru.input into elementary ops supported by TOSA.
+
+    GRU cell equations per timestep:
+        r_t = sigmoid(x_t @ W_ir.T + b_ir + h_{t-1} @ W_hr.T + b_hr)
+        z_t = sigmoid(x_t @ W_iz.T + b_iz + h_{t-1} @ W_hz.T + b_hz)
+        n_t = tanh(x_t @ W_in.T + b_in + r_t * (h_{t-1} @ W_hn.T + b_hn))
+        h_t = n_t + z_t * (h_{t-1} - n_t)
+
+    The weights are batched: one mm computes all three gates at once, then the
+    result is sliced into r/z/n components. This yields 2 mm ops per timestep
+    instead of 6.
+
+    Supports multi-layer, bidirectional, with/without bias, and batch_first.
+
+    """
+
+    _passes_required_after: Set[Type[ExportPass]] = {InsertTableOpsPass}
+
+    _TARGET = torch.ops.aten.gru.input
+
+    _linear = torch.ops.aten.linear.default
+    _add = torch.ops.aten.add.Tensor
+    _sub = torch.ops.aten.sub.Tensor
+    _mul = torch.ops.aten.mul.Tensor
+    _sigmoid = torch.ops.aten.sigmoid.default
+    _tanh = torch.ops.aten.tanh.default
+    _slice = torch.ops.aten.slice_copy.Tensor
+    _unsqueeze = torch.ops.aten.unsqueeze.default
+    _cat = torch.ops.aten.cat.default
+    _select = torch.ops.aten.select_copy.int
+
+    def _build_direction(
+        self,
+        graph: torch.fx.Graph,
+        node: torch.fx.Node,
+        current_input: torch.fx.Node,
+        h_prev: torch.fx.Node,
+        weight_ih: torch.fx.Node,
+        weight_hh: torch.fx.Node,
+        bias_ih,
+        bias_hh,
+        hidden_size: int,
+        seq_len: int,
+        time_dim: int,
+        reverse: bool,
+    ) -> Tuple[List[torch.fx.Node], torch.fx.Node]:
+        """Build GRU cell computation for one direction.
+
+        Returns (timestep_outputs, h_final) where timestep_outputs are
+        unsqueezed hidden states in forward time order.
+
+        """
+        time_indices = range(seq_len - 1, -1, -1) if reverse else range(seq_len)
+        timestep_outputs = []
+
+        for t_idx in time_indices:
+            x_t = create_node(
+                graph,
+                self._select,
+                args=(current_input, time_dim, t_idx),
+                from_node=node,
+            )
+
+            gates_x = create_node(
+                graph, self._linear, args=(x_t, weight_ih, bias_ih), from_node=node
+            )
+            gates_h = create_node(
+                graph, self._linear, args=(h_prev, weight_hh, bias_hh), from_node=node
+            )
+
+            H = hidden_size
+            r_x = create_node(
+                graph, self._slice, args=(gates_x, 1, 0, H), from_node=node
+            )
+            z_x = create_node(
+                graph, self._slice, args=(gates_x, 1, H, 2 * H), from_node=node
+            )
+            n_x = create_node(
+                graph,
+                self._slice,
+                args=(gates_x, 1, 2 * H, 3 * H),
+                from_node=node,
+            )
+            r_h = create_node(
+                graph, self._slice, args=(gates_h, 1, 0, H), from_node=node
+            )
+            z_h = create_node(
+                graph, self._slice, args=(gates_h, 1, H, 2 * H), from_node=node
+            )
+            n_h = create_node(
+                graph,
+                self._slice,
+                args=(gates_h, 1, 2 * H, 3 * H),
+                from_node=node,
+            )
+
+            r_pre = create_node(graph, self._add, args=(r_x, r_h), from_node=node)
+            r_t = create_node(graph, self._sigmoid, args=(r_pre,), from_node=node)
+
+            z_pre = create_node(graph, self._add, args=(z_x, z_h), from_node=node)
+            z_t = create_node(graph, self._sigmoid, args=(z_pre,), from_node=node)
+
+            r_n_h = create_node(graph, self._mul, args=(r_t, n_h), from_node=node)
+            n_pre = create_node(graph, self._add, args=(n_x, r_n_h), from_node=node)
+            n_t = create_node(graph, self._tanh, args=(n_pre,), from_node=node)
+
+            diff = create_node(graph, self._sub, args=(h_prev, n_t), from_node=node)
+            z_diff = create_node(graph, self._mul, args=(z_t, diff), from_node=node)
+            h_t = create_node(graph, self._add, args=(n_t, z_diff), from_node=node)
+            h_prev = h_t
+
+            h_t_expanded = create_node(
+                graph, self._unsqueeze, args=(h_t, time_dim), from_node=node
+            )
+            timestep_outputs.append(h_t_expanded)
+
+        # Backward outputs were appended in reverse time order; flip to
+        # forward order so they align with the forward direction for concat.
+        if reverse:
+            timestep_outputs.reverse()
+
+        return timestep_outputs, h_prev
+
+    def call(self, graph_module: torch.fx.GraphModule):  # noqa: C901
+        graph = graph_module.graph
+        made_changes = False
+
+        for node in list(graph.nodes):
+            if (
+                node.op != "call_function"
+                or node.target != self._TARGET
+                or not self.allowed_to_transform(node.meta)
+            ):
+                continue
+
+            args = node.args
+            input_node = args[0]
+            hx_node = args[1]
+            params = args[2]
+            has_biases = args[3]
+            num_layers = args[4]
+            # dropout (args[5]) and train (args[6]) are unused at inference
+            bidirectional = args[7]
+            batch_first = args[8]
+
+            input_val = input_node.meta["val"]
+            hx_val = hx_node.meta["val"]
+
+            if batch_first:
+                seq_len = input_val.shape[1]
+                time_dim = 1
+            else:
+                seq_len = input_val.shape[0]
+                time_dim = 0
+
+            hidden_size = hx_val.shape[-1]
+            num_directions = 2 if bidirectional else 1
+            # Params per layer: (w_ih, w_hh[, b_ih, b_hh]) * num_directions
+            dir_step = 4 if has_biases else 2
+            layer_step = dir_step * num_directions
+
+            with graph.inserting_before(node):
+                current_input = input_node
+                layer_final_hiddens = []
+
+                for layer_idx in range(num_layers):
+                    layer_offset = layer_idx * layer_step
+
+                    # Forward direction
+                    fw_off = layer_offset
+                    fw_w_ih = params[fw_off]
+                    fw_w_hh = params[fw_off + 1]
+                    fw_b_ih = params[fw_off + 2] if has_biases else None
+                    fw_b_hh = params[fw_off + 3] if has_biases else None
+
+                    fw_h0 = create_node(
+                        graph,
+                        self._select,
+                        args=(hx_node, 0, num_directions * layer_idx),
+                        from_node=node,
+                    )
+
+                    fw_outputs, fw_h_final = self._build_direction(
+                        graph,
+                        node,
+                        current_input,
+                        fw_h0,
+                        fw_w_ih,
+                        fw_w_hh,
+                        fw_b_ih,
+                        fw_b_hh,
+                        hidden_size,
+                        seq_len,
+                        time_dim,
+                        reverse=False,
+                    )
+
+                    if bidirectional:
+                        bw_off = layer_offset + dir_step
+                        bw_w_ih = params[bw_off]
+                        bw_w_hh = params[bw_off + 1]
+                        bw_b_ih = params[bw_off + 2] if has_biases else None
+                        bw_b_hh = params[bw_off + 3] if has_biases else None
+
+                        bw_h0 = create_node(
+                            graph,
+                            self._select,
+                            args=(hx_node, 0, 2 * layer_idx + 1),
+                            from_node=node,
+                        )
+
+                        bw_outputs, bw_h_final = self._build_direction(
+                            graph,
+                            node,
+                            current_input,
+                            bw_h0,
+                            bw_w_ih,
+                            bw_w_hh,
+                            bw_b_ih,
+                            bw_b_hh,
+                            hidden_size,
+                            seq_len,
+                            time_dim,
+                            reverse=True,
+                        )
+
+                        # Concatenate fw + bw at each timestep along feature dim
+                        merged = []
+                        for fw_out, bw_out in zip(fw_outputs, bw_outputs):
+                            merged.append(
+                                create_node(
+                                    graph,
+                                    self._cat,
+                                    args=([fw_out, bw_out], -1),
+                                    from_node=node,
+                                )
+                            )
+
+                        layer_output = create_node(
+                            graph,
+                            self._cat,
+                            args=(merged, time_dim),
+                            from_node=node,
+                        )
+
+                        layer_final_hiddens.append(
+                            create_node(
+                                graph,
+                                self._unsqueeze,
+                                args=(fw_h_final, 0),
+                                from_node=node,
+                            )
+                        )
+                        layer_final_hiddens.append(
+                            create_node(
+                                graph,
+                                self._unsqueeze,
+                                args=(bw_h_final, 0),
+                                from_node=node,
+                            )
+                        )
+                    else:
+                        layer_output = create_node(
+                            graph,
+                            self._cat,
+                            args=(fw_outputs, time_dim),
+                            from_node=node,
+                        )
+
+                        layer_final_hiddens.append(
+                            create_node(
+                                graph,
+                                self._unsqueeze,
+                                args=(fw_h_final, 0),
+                                from_node=node,
+                            )
+                        )
+
+                    current_input = layer_output
+
+                # Build h_n
+                if len(layer_final_hiddens) == 1:
+                    h_n = layer_final_hiddens[0]
+                else:
+                    h_n = create_node(
+                        graph,
+                        self._cat,
+                        args=(layer_final_hiddens, 0),
+                        from_node=node,
+                    )
+
+                output_node = current_input
+
+            # Replace getitem users: GRU returns (output, h_n)
+            getitem_nodes = []
+            for user in list(node.users.keys()):
+                if user.target == operator.getitem:
+                    idx = user.args[1]
+                    if idx == 0:
+                        user.replace_all_uses_with(output_node)
+                    elif idx == 1:
+                        user.replace_all_uses_with(h_n)
+                    getitem_nodes.append(user)
+
+            # Erase getitem nodes then the GRU node explicitly;
+            # eliminate_dead_code does not remove GRU because aten
+            # considers it impure (may have dropout side-effects).
+            for gi in getitem_nodes:
+                graph.erase_node(gi)
+            graph.erase_node(node)
+            made_changes = True
+
+        if not made_changes:
+            return PassResult(graph_module, False)
+
+        graph_module.recompile()
+        graph_module = super().call(graph_module).graph_module
+        return PassResult(graph_module, True)
--- a/backends/arm/_passes/decompose_lstm_pass.py
+++ b/backends/arm/_passes/decompose_lstm_pass.py
@@ -1,0 +1,376 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import operator
+from typing import List, Set, Tuple, Type
+
+import torch
+from executorch.backends.arm._passes.arm_pass import ArmPass
+from executorch.backends.arm._passes.arm_pass_utils import create_node
+from executorch.backends.arm._passes.insert_table_ops import InsertTableOpsPass
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class DecomposeLstmPass(ArmPass):
+    """Decomposes aten.lstm.input into elementary ops supported by TOSA.
+
+    LSTM cell equations per timestep:
+        i_t = sigmoid(x_t @ W_ii.T + b_ii + h_{t-1} @ W_hi.T + b_hi)
+        f_t = sigmoid(x_t @ W_if.T + b_if + h_{t-1} @ W_hf.T + b_hf)
+        g_t = tanh(x_t @ W_ig.T + b_ig + h_{t-1} @ W_hg.T + b_hg)
+        o_t = sigmoid(x_t @ W_io.T + b_io + h_{t-1} @ W_ho.T + b_ho)
+        c_t = f_t * c_{t-1} + i_t * g_t
+        h_t = o_t * tanh(c_t)
+
+    The weights are batched: one mm computes all four gates at once, then
+    the result is sliced into i/f/g/o components. This yields 2 mm ops per
+    timestep instead of 8.
+
+    Supports multi-layer, bidirectional, with/without bias, and batch_first.
+
+    """
+
+    _passes_required_after: Set[Type[ExportPass]] = {InsertTableOpsPass}
+
+    _TARGET = torch.ops.aten.lstm.input
+
+    _linear = torch.ops.aten.linear.default
+    _add = torch.ops.aten.add.Tensor
+    _mul = torch.ops.aten.mul.Tensor
+    _sigmoid = torch.ops.aten.sigmoid.default
+    _tanh = torch.ops.aten.tanh.default
+    _slice = torch.ops.aten.slice_copy.Tensor
+    _unsqueeze = torch.ops.aten.unsqueeze.default
+    _cat = torch.ops.aten.cat.default
+    _select = torch.ops.aten.select_copy.int
+
+    def _build_direction(
+        self,
+        graph: torch.fx.Graph,
+        node: torch.fx.Node,
+        current_input: torch.fx.Node,
+        h_prev: torch.fx.Node,
+        c_prev: torch.fx.Node,
+        weight_ih: torch.fx.Node,
+        weight_hh: torch.fx.Node,
+        bias_ih,
+        bias_hh,
+        hidden_size: int,
+        seq_len: int,
+        time_dim: int,
+        reverse: bool,
+    ) -> Tuple[List[torch.fx.Node], torch.fx.Node, torch.fx.Node]:
+        """Build LSTM cell computation for one direction.
+
+        Returns (timestep_outputs, h_final, c_final).
+
+        """
+        time_indices = range(seq_len - 1, -1, -1) if reverse else range(seq_len)
+        timestep_outputs = []
+
+        for t_idx in time_indices:
+            x_t = create_node(
+                graph,
+                self._select,
+                args=(current_input, time_dim, t_idx),
+                from_node=node,
+            )
+
+            gates_x = create_node(
+                graph, self._linear, args=(x_t, weight_ih, bias_ih), from_node=node
+            )
+            gates_h = create_node(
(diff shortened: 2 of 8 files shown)


backends/arm/tosa/backend.py
```diff
@@ -350,11 +350,16 @@ def _preprocess_module(  # noqa: C901
...
+        # original_module (a pre-preprocess copy), creating a mismatch.
+        # The TOSA reference model preserves flatbuffer order, so no
+        # sorting is needed.
         if output_order_workaround:
-            logger.debug("Re-sorting outputs during TOSA lowering.")
-            graph_module = _sort_outputs(graph_module, node_to_id_map)
-        else:
-            logger.debug("No re-sorting outputs (workaround) during TOSA lowering.")
+            logger.debug(
+                "output_order_workaround is set but _sort_outputs is "
```
mansnils: See comment: https://github.com/pytorch/executorch/pull/17138#pullrequestreview-3761694279

mansnils (changes requested):

gggekov:
What is the reason to decompose LSTM in the Arm backend rather than let `torch.export.export` decompose the LSTM ?

gggekov:
Never mind- i see the `torch.nn.LSTM` is not decomposed in the `torch.export.export` as I thought initially.

apullin:
> Never mind- i see the `torch.nn.LSTM` is not decomposed in the `torch.export.export` as I thought initially.

If I have understood ET correctly - LSTM currently has quant support in torch.ao itself. Instead of amending torch (big job, long horizon, wait for release, etc), this stack of commits should make it possible to swap out LSTM either by swapping the whole module on ET front end _or_ doing a full decomposition on the backend. Same for GRU and RNN (newly added in both places).

gggekov:
You are right - the `torch.nn.quantizable.modules.rnn.LSTM` comes decomposed(that's the LSTM we were supporting) whereas the `torch.nn.LSTM` is not decomposed out of the box, hence either we need to decompose it manually(as your patch is doing) or we need to let ExecuTorch decompose it. 

Inside the `DecomposeMatmulPass(ArmPass)`, we've introduced a new promising way to decompose complex operators - we use the `torch._decomp.get_decompositions`. The advantage is that we don't need to decompose operators ourselves, we let PyTorch do the decomposition. The risk is that the PyTorch decomposition may introduce operators we don't support, hence this needs to be done with care. Could you decompose LSTM in similar way ? If you add the LSTM here(https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/decompose_matmul.py#L37), you may need some extra logic to manage the various attributes, but at a high level, i would expect this approach should work for all the operators you are trying to add support for.

apullin:
> torch._decomp.get_decompositions

@gggekov This was tried with a vanilla `torch.nn.GRU`, which produces a graph with:
- addmm                       
- split_with_sizes_copy       
- view_copy                   
- squeeze_copy                
- permute_copy                
- add, sub, mul, sigmoid, tanh
- cat, unsqueeze_copy       

By comparison, the added frontend decomposition pass (#17137 for GRU) produces a graph with:
- mm + add (written out directly in the new code)
- slice_copy (again, direct)
- add, sub, mul, sigmoid, tanh
- cat, unsqueeze_copy

The advantage of using `get_decompositions` is that pytorch controls the implementation, as opposed to us bringing an alternate re-telling of the GRU/LSTM/RNN story to the table.

Beyond that, there's also #17141 , which adds quantizable whole modules, rather than decomposing.

The most correct way (I think) would be to add GRU and RNN to torch.ao, but that requires upstreaming.

This approach follows the same pattern as the existing `DecomposeLinearPass` in the ARM backend, which manually decomposes `aten.linear` into `mm` + `add` rather than relying on PyTorch's decomposition to `addmm`. And gives us a touchpoint for direct control over which ops reach the TOSA partitioner (which has quirks, and we can thus tailor for it here).


There are 3 PR's up for commits (stacked together) for GRU, LSTM, and RNN decompose passes, using the same technique. *If* we move forward, I could squash those into one commit? Or one PR w/ 3 commits.
(sorry, internal->github export is a little weird, that's why this came out as several separate PR's)

gggekov:
@apullin apologies for the radio silence over the last weeks. Happy to go with your approach of manually decomposing rather than the `get_decompositions` route. I see you have a few failing tests and a merge conflict, could you resolve these ? Your patch looks good to me, I haven't verified the numerics of the decomposition,  I assume they must be alright since you pass the unit tests. It would be great to have the LSTM, GRU & RNN available in the Arm backend.

gggekov:
Thanks for updating ! I aim to run locally your patch before the end of the week and understand if there are any numerics issues. In the meantime, I see the pull / unittest-arm-backend-with-no-deps (test_pytest_ops_tosa) / linux-job (pull_request) CI job is still failing, will you update the PR to fix it ?

apullin:
>  pull / unittest-arm-backend-with-no-deps (test_pytest_ops_tosa) / linux-job (pull_request) CI job is still failing

This one is being extremely elusive for me, partly because I cannot run CI test suite locally on my Meta machine - not Linux and Docker is locked out. That is compounded by the difficulty of syncing with the internal repo to this PR. 

You'll see a number of PR updates, but I'll @ you when I am sure it is ready.

apullin:
@gggekov OK, should be as green as it is going to get, now.

apullin (commented):
@mansnils Trying to clear out the previous requested changes related to MLETORCH-1266, but that has long since been settled by others' PRs. Do I have to do anything from my end?

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+            x = torch.randn(batch_size, seq_len, input_size)
+        else:
+            x = torch.randn(seq_len, batch_size, input_size)
+        h = torch.randn(
+            self.num_layers * self.num_directions, batch_size, self.hidden_size
+        )
+        return (x, h)
+
+
+def test_decompose_rnn_tanh_tosa_FP():
```
gggekov: Could you please name the test case "test_decompose_rnn_tosa_FP_tanh" ? 

That is because we only run tests named "test_PASS_TARGET_*".  We have ongoing work to loosen up that in the future so that any test is run. I imported your PR internally with test names following the convention and the test case passed, so i'm just asking if you can change the name.

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+    """Test basic RNN with tanh (single layer, unidirectional, with bias)."""
+    module = RNNTanh()
+    pipeline = PassPipeline[input_t](
```
gggekov: Rather than testing via PassPipeline, could you please write the test to use TosaPipelineINT and TosaPipelineFP (and also add EthosU55PipelineINT / EthosU85PipelineINT if you want to run the test on the Corstone-300/Corstone-320 FVP to be sure it works on the Ethos-U55 or Ethos-U85.) . Since you decompose in transform_for_annotation(meaning you first decompose and then quantize), i'd say TosaPipelineINT and TosaPipelineFP are really useful to have and EthosU55PipelineINT/EthosU85PipelineINT is at your discretion.

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+            # RNN tanh: 1 tanh per timestep
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 5,
+        },
+        pass_list=[DecomposeRnnPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_rnn_relu_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_rnn_tosa_FP_relu" ?

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+            # RNN relu: 1 relu per timestep
+            "executorch_exir_dialects_edge__ops_aten_relu_default": 5,
+        },
+        pass_list=[DecomposeRnnPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_rnn_tanh_bidirectional_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_rnn_tosa_FP_tanh_bidirectional" ?

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+            # Bidirectional: 2x the ops
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 10,
+        },
+        pass_list=[DecomposeRnnPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_rnn_tanh_no_bias_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_rnn_tosa_FP_tanh_no_bias" ?

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,200 @@
...
+        ops_after_pass={
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 5,
+        },
+        pass_list=[DecomposeRnnPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_rnn_tanh_multilayer_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_rnn_tosa_FP_tanh_multilayer" ?

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+            "executorch_exir_dialects_edge__ops_aten_sigmoid_default": 10,
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 5,
+        },
+        pass_list=[DecomposeGruPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_gru_bidirectional_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_gru_tosa_FP" ?

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+            "executorch_exir_dialects_edge__ops_aten_sigmoid_default": 20,
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 10,
+        },
+        pass_list=[DecomposeGruPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_gru_no_bias_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_gru_tosa_FP_no_bias"

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+            "executorch_exir_dialects_edge__ops_aten_sigmoid_default": 10,
+            "executorch_exir_dialects_edge__ops_aten_tanh_default": 5,
+        },
+        pass_list=[DecomposeGruPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_gru_multilayer_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_gru_tosa_FP_multilayer" ?

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+        pass_list=[DecomposeGruPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_gru_no_bias_tosa_FP():
+    """Test GRU without bias."""
+    module = GRU(bias=False)
+    pipeline = PassPipeline[input_t](
```
gggekov: Same comment as for the RNN pass for testing via TosaPipelineINT + TosaPipelineFP rather than via the PassPipeline

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+        pass_list=[DecomposeGruPass],
+    )
+    pipeline.run_and_compare_to_initial_model()
+    pipeline.run()
+
+
+def test_decompose_gru_bidirectional_tosa_FP():
+    """Test bidirectional GRU."""
+    module = GRU(bidirectional=True)
+    pipeline = PassPipeline[input_t](
```
gggekov: Same comment as for the RNN pass for testing via TosaPipelineINT + TosaPipelineFP rather than via the PassPipeline

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,142 @@
...
+        h = torch.randn(
+            self.num_layers * self.num_directions, batch_size, self.hidden_size
+        )
+        return (x, h)
+
+
+def test_decompose_gru_tosa_FP():
+    """Test basic GRU (single layer, unidirectional, with bias)."""
+    module = GRU()
+    pipeline = PassPipeline[input_t](
```
gggekov: Same comment as for the RNN pass for testing via TosaPipelineINT + TosaPipelineFP rather than via the PassPipeline

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,119 @@
...
+    with TosaLoweringContext(tosa_spec):
+        DecomposeLstmPass()(ep.graph_module)
+    ep.graph_module.recompile()
+
+    with torch.no_grad():
+        test_output, test_h_n, test_c_n = ep.module()(*test_data)
+
+    torch.testing.assert_close(test_output, ref_output, atol=1e-5, rtol=1e-5)
+    torch.testing.assert_close(test_h_n, ref_h_n, atol=1e-5, rtol=1e-5)
+    torch.testing.assert_close(test_c_n, ref_c_n, atol=1e-5, rtol=1e-5)
```
gggekov: Do you mind to instead rewrite the test to use TosaPipelineFP ? The advantage is that it will be consistent with the rest of the backend, you call all passes in order rather than just invoking a single pass ?  
Since LSTM is decomposed in transform_for_annotation(meaning before quantization), I would vote to also add test cases for TosaPipelineINT to make sure the decompositions work numerically for the 8x8 case.

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_decompose_lstm_tosa_FP():
+    """Test basic LSTM (single layer, unidirectional, with bias)."""
+    _export_run_pass_and_compare(LSTM())
+
+
+def test_decompose_lstm_bidirectional_tosa_FP():
```
gggekov: If you keep the test in backends/arm/test/_passes, could you please rename the test to "test_decompose_lstm_tosa_FP_bidirectional" so that it gets picked up by our automated tooling and actually gets tested?

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,119 @@
...
+    """Test bidirectional LSTM."""
+    _export_run_pass_and_compare(LSTM(bidirectional=True))
+
+
+def test_decompose_lstm_no_bias_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_lstm_tosa_FP_no_bias" ?

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,119 @@
...
+    """Test LSTM without bias."""
+    _export_run_pass_and_compare(LSTM(bias=False))
+
+
+def test_decompose_lstm_multilayer_tosa_FP():
```
gggekov: Could you please rename to "test_decompose_lstm_tosa_FP_multilayer" ?

gggekov (commented):
Thanks for updating the PR and fixing the CI ! It's a great PR and would be really nice to add support for torch.nn.lstm/rnn/gru! I have added comments in the unit tests, I have two overall comments. 
1) Could you please use TosaPipelineFP and TosaPipelineINT rather than the PassPipeline to test the decompositions for the three operators ? 

2) We have very strict naming convention for how to name the unit tests. We are working on loosening the convention, but right now, if a unit test does not respect our naming convention, it is not picked up by the CI. I've added comments how to name the various tests if you keep them in backends/arm/test/passes. 
IMO, it would make more sense to move the tests under `backends/arm/test/ops`, that's where we test most operators that we decompose in transform_for_annotationn(e.g. test_adaptive_avg_pool2d.py, test_gelu.py, test_layer_norm.py, etc). In the case of LSTM, we also already have a unit test in executorch/backends/arm/test/models/test_lstm_arm.py for the `rnn.LSTM`, you can create another model with `torch.nn.lstm` and amend that file if you go down the route of moving the tests.

backends/arm/test/passes/test_decompose_rnn_pass.py
apullin: Done, renamed:
```
test_decompose_rnn_tosa_FP_relu
test_decompose_rnn_tosa_FP_tanh_bidirectional
test_decompose_rnn_tosa_FP_tanh_no_bias
test_decompose_rnn_tosa_FP_tanh_multilayer
```
to follow the pattern.

Same for GRU and LSTM functions.

apullin:
@gggekov The renaming is done.

And: I redid the tests to use TosaPipelineFP & TosaPipelineINT for all all three layer types (RNN, GRU, LSTM).

BUT: The LSTM TosaPipeline tests are currently skipped due to a TOSA backend shape mismatch on h_n/c_n (our decomposition pass outputs correct shapes, verified by the unit tests — the extra dimension is introduced during TOSA lowering). I'm pretty sure it's a TOSA backend issue. I can attempt to fix this directly as another commit, but that is some creep. Unless I am wrong about that as the failure origin.

I also tried adding EthosU55PipelineINT / EthosU85PipelineINT tests, and they ran fine locally, but GH CI doesn't seem to have the ethos-u-vela package installed?? So they fail there despite working locally.

backends/arm/test/passes/test_decompose_rnn_pass.py
apullin: Tests updated, but please check. EthosU55PipelineINT / EthosU85PipelineINT work locally but seemed to fail in CI due to lack of `etho-u-vela`. I should have left it on that version and asked for feedback,a ctually.

backends/arm/test/passes/test_decompose_lstm_pass.py
apullin: TosaPipeline tests are added but skipped due to the TOSA backend shape issue mentioned below, and the manual tests are kept as a fallback until that's resolved.

gggekov:
Hi @apullin ,
I am also getting a failure for the test_decompose_lstm_tosa_FP_e2e test case
```
FAILED backends/arm/test/passes/test_decompose_lstm_pass.py::test_decompose_lstm_tosa_FP_e2e - ValueError: Output needs to be of same shape: torch.Size([1, 1, 2, 20]) != torch.Size([1, 2, 20])
```

What is the reason for the shape mismatch for TOSA FP? Normally, when lowering an operator, the hard part is to get the numerics right for TOSA INT. In your case, the numerics for TOSA INT seem fine, but you get shape mismatch for TOSA FP. I wonder if something in another pass, run before the DecomposeLstmPass, doesn't cause the shape mismatch. That's why it's important to test via `TosaPipelineINT` & `TosaPipelineFP` because you run the full pipeline rather than a single pass in isolation. 

Could you also add `TosaPipelineINT` & `TosaPipelineFP` tests for the other cases ( `LSTM(bidirectional=True)` , `LSTM(bias=False)`, `LSTM(num_layers=2)`) ?

gggekov:
>Could you also add TosaPipelineINT & TosaPipelineFP tests for the other cases ( LSTM(bidirectional=True) , LSTM(bias=False), LSTM(num_layers=2)) ?

This applies also to the RNN & GRU test cases as well. It would be useful if you test the various attributes of RNN & GRU with `TosaPipelineFP` and `TosaPipelineINT` in order to make sure we run the full pipeline rather than just one pass. That's how we test all the operators we support via `transform_for_annotation`. 

One idea about the shape mismatch for LSTM FP- it's possible that you may need to run the pass at a slightly different place in the `arm_pass_manager.py` - i did a few tests locally running the DecomposeLstmPass earlier in _tosa_pipeline, although i kept getting shape mismatch for TOSA FP :|

apullin:
@gggekov ok, I THINK I have this down, now. Updated all three PRs with the following changes:

  All three passes (GRU, RNN, LSTM):
  - Added `graph_module = super().call(graph_module).graph_module after recompile()` to re-propagate FakeTensor shapes after decomposition. This was the root cause of the TOSA shape mismatch on LSTM `h_n`/`c_n` outputs.

  Tests:
  - All PassPipeline tests replaced with TosaPipelineFP + TosaPipelineINT across all three layer types.
  - LSTM tests are no longer skipped — the shape fix plus a workaround (`run_transform_for_annotation_pipeline` before to_edge) resolves the extra-dimension
  issue. All 8 LSTM tests (4 FP + 4 INT) now run end-to-end through TosaPipeline.
  - GRU: 8 tests (4 configs × FP + INT)
  - RNN: 10 tests (5 configs × FP + INT)
  - LSTM: 8 tests (4 configs × FP + INT)

All tests pass in a linux environment. But we will have to see if CI agrees with that.

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,157 @@
...
+    Workaround for a torch.export bug where FakeTensor propagation for
+    aten.lstm.input records h_n/c_n with an extra leading dimension (4D instead
+    of 3D). When to_edge() decomposes the LSTM, it inserts view_copy nodes to
+    match the wrong metadata. Decomposing LSTM before to_edge avoids this.
+
+    TODO(upstream): Remove once pytorch/pytorch fixes aten.lstm.input metadata.
+    Verify by checking whether torch.export.export(LSTMModule, ...) records 3D
+    (not 4D) shapes for h_n/c_n in the aten.lstm.input node metadata.
+    """
+    pipeline.add_stage_after(
```
gggekov: Can you please remove that workaround and run the test properly, like all other ops that are supported in transform_for_annotation. I'll explain in the comment history what is wrong and how to remove that.

gggekov (commented):
Hello @apullin ,
Thank you for updating the unit tests to use `TosaPipelineFP`/`TosaPipelineINT`! I think the PR is going in the right direction. 

My main comment is that currently, your 3 unit tests `test_decompose_<rnn/lstm/gru>_pass.py` only work for TOSA INT(not sure if that is intentional?). However, for the `TosaPipelineFP` unit tests, your decomposition is never called. You are currently working around that in a slightly strange way by manually calling transform_for_annotation after the export from the ArmTester, but that is masking the underlying problem.

I think the proper fix is to make the pass decomposition include both EdgeIR ops(currently missing in your PR) and the ATen ops(that's what you currently do). See for example here (https://github.com/pytorch/executorch/blob/main/backends/arm/_passes/decompose_adaptive_avg_pool2d_pass.py#L23). In this way, for the TOSA FP case, you would use the EdgeIR decomposition and for the TOSA INT case, you'd use the ATen decomposition. 

Why do you need the EdgeIR decomposition for the TOSA FP case ? Have a look at https://github.com/pytorch/executorch/blob/main/exir/program/_program.py#L1384, insert a breakpoint/print statement before and after the call to `edge_manager = _gen_edge_manager_for_partitioners(....)` 
Do `print(aten_programs['forward'])` on line 1383 and `print(edge_manager.exported_program("forward"))` on line 1387. On line 1383, you still have the `torch.ops.aten.lstm.input` that you decompose. However, on line 1387, you no longer have the torch.ops.aten.lstm, you have EdgeIR ops instead. Note that this code is called before our backend has been called. Under the hood, the `edge_manager = _gen_edge_manager_for_partitioners(....)` API is calling the `from torch._decomp import get_decompositions` API to do the decomposition from ATen to EdgeIR(i mentioned that API at the very start of the code review a few weeks ago). I think for `TosaPipelineFP`, because the `torch.ops.aten.lstm.input` has been replaced with EdgeIR ops, your pipeline is never called and the unit test is passing relying on the decompositions taking place before our backend has been reached. We'd like to have consistency between TOSA FP & TOSA INT - have the same decomposition for both profiles. For other passes decomposed in transform_for_annotation, the TOSA FP case is managed via the EdgeIR operators. 

Two more comments:
- I notice in the first commit for the GRU, you add test_decompose_recurrent_tosa_pipelines.py and then in the last commit(LSTM), you remove that file. Is it not possible to remove it altogether from the first commit ? 
- Do you mind to do `$ pip install -r requirements-lintrunner.txt &  lintrunner init` and then run the ` ./backends/arm/scripts/pre-commit` script to run the lintrunner ? I had to manually fix a few minor errors to get your PR in the internal PR. 

Well done, thanks for pushing with that!

gggekov:
Hi @apullin , 
I had a second thought - i think for the TOSA FP case, it will be hard to do your own decomposition given that the op is already decomposed by ExecuTorch via the `from torch._decomp import get_decompositions` API before our backend is reached as per my comment above. I suggest to just remove the the `_add_lstm_workaround` function, and instead just mark the TOSA FP tests as expected failures with 
```
@pytest.mark.skip(reason="Shape mismatch for TOSA FP")
```
and add appropriate comment that your pass is only handling the TOSA INT case. For TOSA INT, you shouldn't need the `_add_lstm_workaround` function. 

It would be interesting to understand where is the shape mismatch stemming from for the TOSA FP case, maybe the `from torch._decomp import get_decompositions` API decomposes the op in a way that results in a slightly different shape compared to running the torch.nn.LSTM in eager mode, or maybe we are introducing an extra dimension somewhere in our FP passes by accident. I am not sure. I suggest to proceed the PR just managing the TOSA INT case, which should make it possible to run the LSTM on the Ethos-U85.

apullin:
@gggekov done, updated across all three commits:

- Removed _add_lstm_workaround and test_decompose_recurrent_tosa_pipelines.py
- All tests use TosaPipelineFP / TosaPipelineINT (basic, bidirectional, no_bias, multilayer)
- LSTM FP tests skipped but present, INT is still faithfully tested
- No feature smear between commits

apullin (commented):
Changed related to MLETORCH-1266 have been removed, and MLETORCH-1266 has been settled by another PR.

backends/arm/tosa/partitioner.py
```diff
@@ -349,6 +349,13 @@ def ops_to_not_decompose(  # noqa: C901
         }
         ops_to_not_decompose_always = {
             torch.ops.aten.logit.default,
+            # Preserve RNN-family ops so DecomposeLstmPass/GruPass/RnnPass
```
gggekov: What is the rational for adding the 3 ops here in the `partitioner.py` ? For TOSA INT, you decompose the ops before they are quantized so that check won't be reached. And for TOSA FP, the ops are decomposed by ExecuTorch before our backend is reached so I don't understand what is achieved by adding the operators here.

gggekov:
Thank you @apullin ! 
I think you are on the final straight - do you mind to fix the lintrunner issue from the CI ? 
I think if you do
```
$ pip install -r requirements-lintrunner.txt
$ lintrunner init 
$ lintrunner --revision 'HEAD^' 
```
on the latest commit, you should be able to reproduce the CI failure and fix it(should be quite easy fix). 
I am also not sure I fully understand the rational for adding the ops in the `backends/arm/tosa/partitioner.py` - i believe you shouldn't need that. 

We are very close to merging that!

backends/arm/_passes/decompose_gru_pass.py
```diff
@@ -0,0 +1,345 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: One more minor comment - on the new files you've created, I don't think you should use the Arm copyright since you don't work in Arm. See example here for another external company copyright contributing to our backend https://github.com/pytorch/executorch/blob/main/backends/arm/test/ops/test_abs.py#L1

backends/arm/_passes/decompose_lstm_pass.py
```diff
@@ -0,0 +1,386 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: Not sure if you need to use Arm copyright here

backends/arm/_passes/decompose_rnn_pass.py
```diff
@@ -0,0 +1,300 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: Not sure if you need Arm copyright here

backends/arm/test/passes/test_decompose_rnn_pass.py
```diff
@@ -0,0 +1,198 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: Not sure if you need Arm copyright

backends/arm/test/passes/test_decompose_gru_pass.py
```diff
@@ -0,0 +1,140 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: Not sure if you need Arm copyright

backends/arm/test/passes/test_decompose_lstm_pass.py
```diff
@@ -0,0 +1,146 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
```
gggekov: Not sure if you need Arm copyright

gggekov (commented):
Added one more comment about the copyright mention at the top of each new file

backends/arm/_passes/decompose_gru_pass.py
apullin: copy paste error, resolved throughout

backends/arm/tosa/partitioner.py
apullin: Removed. Left over from before understanding how our passes decompose before this is checked (INT) or before the backend is reached (FP)

oscarandersson8218 (approved):

gggekov (commented):
Looks good to me to merge

mansnils (approved):

gggekov:
The merge button is greyed out, i believe a maintainer of ExecuTorch such as @digantdesai needs to approve in order to merge

apullin:
@gggekov Ah, sadly, there's one final thing: It looks like there is something wrong with the decomposition where it uses BMM resulting in a failure in `ConvertMmToBmmPass`, but everything passes public CI and only fails on the internal Meta test stack. Really not sure why that is.

The workaround I tried & pushed is to replace `mm(x, transpose(W))` with `linear(x, W, bias)` ... a little unfortunate since I know people love BMM. But that gets this passing without needing to add another commit for BMM.
Changes are located entirely in `_build_direction`.

Commits are updated with the changes that pass internally.

zingo:
> @gggekov Ah, sadly, there's one final thing: It looks like there is something wrong with the decomposition where it uses BMM resulting in a failure in `ConvertMmToBmmPass`, but everything passes public CI and only fails on the internal Meta test stack. Really not sure why that is.
> 
> The workaround I tried & pushed is to replace `mm(x, transpose(W))` with `linear(x, W, bias)` ... a little unfortunate since I know people love BMM. But that gets this passing without needing to add another commit for BMM. Changes are located entirely in `_build_direction`.
> 
> Commits are updated with the changes that pass internally.

Thanks for making sure is really good. Would it possible/ok to get some version of that test inte to tests on GitHub

gggekov:
Restarting the CI as there was one failure for our vkml test cases.

gggekov:
Think you need a stamp from @digantdesai and can then merge

digantdesai:
Rebase?

gggekov:
Thanks a lot for the contribution, @apullin !

PR #17141, opened 2026-02-03
Add quantizable GRU and RNN modules for ARM backend (#17141)

apullin:
Summary:

Adds quantizable GRU and RNN modules for quantization-aware training (QAT). Standard `nn.GRU` and `nn.RNN` modules are opaque composite operations, so these modules expose their arithmetic as `nn.Linear` and `FloatFunctional` operations that can be observed and quantized.

The implementation supports:

- Conversion from `nn.GRU` and `nn.RNN` via `from_float()`
- Multi-layer and bidirectional recurrent networks
- `batch_first`, optional bias, and explicit hidden state
- Tanh and ReLU RNN nonlinearities
- Dropout between recurrent layers during training
- GRU gate arithmetic represented through quantization-aware functional operations

Conversion creates independent parameter storage, implicit hidden states preserve the input dtype and device, and each converted layer derives its input width from its actual weights.

Differential Revision: D92059608


backends/arm/tosa/backend.py
```diff
@@ -350,11 +350,16 @@ def _preprocess_module(  # noqa: C901
...
+        # flatbuffer outputs but the runner reads output shapes from
+        # original_module (a pre-preprocess copy), creating a mismatch.
+        # The TOSA reference model preserves flatbuffer order, so no
+        # sorting is needed.
         if output_order_workaround:
-            logger.debug("Re-sorting outputs during TOSA lowering.")
-            graph_module = _sort_outputs(graph_module, node_to_id_map)
-        else:
-            logger.debug("No re-sorting outputs (workaround) during TOSA lowering.")
+            logger.debug(
```
mansnils: See comment: https://github.com/pytorch/executorch/pull/17138#pullrequestreview-3761694279

mansnils (changes requested):

backends/arm/tosa/backend.py
```diff
@@ -350,11 +350,16 @@ def _preprocess_module(  # noqa: C901
...
+        # _sort_outputs is intentionally disabled: it reorders the TOSA
+        # flatbuffer outputs but the runner reads output shapes from
+        # original_module (a pre-preprocess copy), creating a mismatch.
+        # The TOSA reference model preserves flatbuffer order, so no
+        # sorting is needed.
         if output_order_workaround:
-            logger.debug("Re-sorting outputs during TOSA lowering.")
-            graph_module = _sort_outputs(graph_module, node_to_id_map)
-        else:
-            logger.debug("No re-sorting outputs (workaround) during TOSA lowering.")
```
mansnils: See comment: https://github.com/pytorch/executorch/pull/17138#pullrequestreview-3761694279

mansnils (changes requested):

backends/arm/tosa/backend.py
apullin: I think I may have done the "export" wrong, and it created 5 separate PR's where then should have been only one, which had all the commits in it. Sorry for the mess! But that's why this is coming up in each PR.

Once rebased and refreshed with the disablement for MLETORCH-1266 as requested, all these should go away.

apullin: this is superseded by someone else fixing MLETORCH-1266

digantdesai:
@apullin whats next here?

apullin:
> @apullin whats next here?

Not sure. I rebased the diff, resolved conflict in internal source, but now GH is showing conflicts. 
This might have been caused by different changes going to internal vs external.

PR #17147, opened 2026-02-03
NXP backend: Add support for `aten.upsample_nearest2d.vec`.

MartinPavella:
### Summary
Add quantization and NeutronIR conversion of `aten.upsample_nearest2d.vec`.

### Test plan
Unit-tests provided in `backends/nxp/tests/test_convert_upsample_nearest2d.py`.

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/upsample_nearest2d_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/upsample_nearest2d_converter.py
@@ -1,0 +1,107 @@
+# Copyright 2026 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import numpy as np
+
+from executorch.backends.nxp.backend.data_format import DataFormat, NXP_NODE_FORMAT
+from executorch.backends.nxp.backend.edge_helper import node_has_well_defined_shape
+from executorch.backends.nxp.backend.ir.converter.node_converter import (
+    CustomDelegationOptions,
+    NodeConverter,
+)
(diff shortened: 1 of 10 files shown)


MartinPavella:
This PR is based on https://github.com/pytorch/executorch/pull/17066, so only the last commit is relevant.

backends/nxp/tests/test_convert_upsample_nearest2d.py
robert-kalmar, line 1: Can we support the channel last dim order with upsample_nearest2d? If so, please add a corresponding unit tests.

backends/nxp/backend/ir/converter/node_converters/ops_converters/upsample_nearest2d_converter.py
robert-kalmar, line 1: Also update the documentation https://github.com/pytorch/executorch/blob/main/docs/source/backends/nxp/op-support.csv.

backends/nxp/tests/test_convert_upsample_nearest2d.py
MartinPavella: Actually, as the `upsample` distinguishes between channels and spatial dimensions, it always requires the channels last format, which is reflected in the `node_format_inference.py`. So all `upsample` tests will always be with channels last.
And regarding the dim order, nothing operator specific needs to be done. Everything is handled by the `node_format_inference` and by the `model_builder`.

backends/nxp/backend/ir/converter/node_converters/ops_converters/upsample_nearest2d_converter.py
MartinPavella: Done.

robert-kalmar (approved):

PR #17155, opened 2026-02-03
Arm backend: Conv + INT16: check DTYPE with qparams rather than runtime's DTYPE

elpdumont:
Summary:
## Overview

- Modifying `DecomposeConvWithInt16ActivationPass` to check the quantization parameters' dtype (meta.data['input_qparams'][0].dtype) instead of the raw tensor dtype (args[0].data.dtype).

## Problem before this PULL request
- The original code checked args[0].data.dtype (the FakeTensor's runtime dtype), which may differ from the quantization target.

Specifically:
- Added a lookup for input_qparams from meta.data
- Added a guard to return early if input qparams are not present (index 0 not in qparams)
- Changed the dtype check from activation_tensor.dtype != torch.int16 to activation_dtype != torch.int16 where activation_dtype comes from the quantization parameters
- Added tests.

## Note:
- Other passes (`insert_rescales_pass.py`) use the same approach to check for INT16 activations. ```shift_bits = 12 if input_qparams[keys[0]].dtype == torch.int16 else 20```

Differential Revision: D92054367

### DIFF
--- a/backends/arm/_passes/decompose_int16_activation_conv_pass.py
+++ b/backends/arm/_passes/decompose_int16_activation_conv_pass.py
@@ -52,7 +52,15 @@ class DecomposeConvWithInt16ActivationPass(ArmPass):
         activation_tensor = args[0].data
         activation_rank = activation_tensor.dim()
 
-        if activation_rank not in (4, 5) or activation_tensor.dtype != torch.int16:
+        # Check input qparams dtype instead of raw tensor dtype, since the tensor
+        # may have been rescaled to int32 by earlier passes while the quantization
+        # parameters still indicate the original int16 dtype.
+        input_qparams = meta.data.get("input_qparams", {})
+        if 0 not in input_qparams:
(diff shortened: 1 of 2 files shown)


zingo (approved):
Thanks for the help, This is OK to merge as is but it would also be really helpful/nice if you add a test for this if you have the time. 
If you want to make a test it can also be in a separate PR, depending on if you need this merged fast. 
Thanks!

zingo:
The error 
**ERROR backends/arm/test/models/test_nss.py** 
Is knows fail and has been fixed on latest and should go away on merge (or rebase)

elpdumont:
@zingo i added test functions as requested. What is the next step for merging?

zingo:
Hi code is approved :) and tests looks fine so feel free to merge when you like, thanks for the help! 
If you don't have access/premissions to merge just add a comment and Ill do it next time I pop in here :)

PR #17156, opened 2026-02-03
NXP backend: added PReLU support

novak-vaclav:
### Summary

adds support for aten.prelu operator

### Test plan

cc @robert-kalmar @JakeStevens @digantdesai @roman-janik-nxp @MartinPavella


backends/nxp/backend/ir/converter/node_converters/ops_converters/prelu_converter.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2024-2026 NXP
```
MartinPavella: Nit: Just 2026?

backends/nxp/neutron_partitioner.py
```diff
@@ -441,4 +442,25 @@ def ops_to_not_decompose(
...
+            node_converter = aten_op_to_converter.get(node.target)
+            if node_converter is None:
+                return False
+
+            delegable = node_converter._is_supported_on_target(
+                node,
+                self.neutron_target_spec,
+                parameters_mapping,
+                self.custom_delegation_options,
+            )
```
MartinPavella, line 532: Why are you only checking target support? Why not call `is_supported()`?

novak-vaclav:
(hopefully) fixed issues from tests

backends/nxp/neutron_partitioner.py
```diff
@@ -419,3 +424,46 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
...
+            nodes that should be continued to be decomposed even though the op they correspond to is
+            in the list returned by ops_to_not_decompose.
+        """
+        parameters_mapping = EdgeProgramToIRConverter.map_inputs_to_parameters(ep)
+        aten_op_to_converter = {}
+        for exir_op, converter in supported_ops.items():
+            aten_op_to_converter[exir_op._op] = converter
+
+        def check_op_support_extended(node: torch.fx.Node):
+            if node not in self.preserve_ops:
```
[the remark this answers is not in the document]
novak-vaclav: True, will fix.

novak-vaclav:
Fixed issues specified by Copilot

backends/nxp/neutron_partitioner.py
```diff
@@ -419,3 +424,46 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
...
+        registered and the `to_backend` is invoked through to_edge_transform_and_lower it will be
+        guaranteed that the program that the backend receives will not have any of these ops
+        decomposed.
+
+        Returns:
+            List[torch._ops.OpOverload]: a list of operator names that should not be decomposed.
+            Optional[Callable[[torch.fx.Node], bool]]]: an optional callable, acting as a filter, that users can provide
+            which will be called for each node in the graph that users can use as a filter for certain
+            nodes that should be continued to be decomposed even though the op they correspond to is
+            in the list returned by ops_to_not_decompose.
```
MartinPavella: Nit: That's a long sentence :D Perhaps it could be rephrased for more clarity. (I know Meta has the same doc-string in the parent class, but we can do better ;) )

Also, it is quite confusing to me how the returned list and function work together. 
I guess it's: Do not decompose anything in the list, unless the function says it should be decomposed anyway.
But then I don't understand why the function contains 
```
if node.target not in self.preserve_ops:
    return False
```
Or does the returned list get ignored if the function is returned as well?

Also, it is not clear if returning `True` (in `check_op_support_extended`) would cause the node to get decomposed or not. I think a comment would be useful.

backends/nxp/backend/ir/converter/node_converters/ops_converters/prelu_converter.py
```diff
@@ -0,0 +1,94 @@
...
+        for h_idx in possible_h_idx:
+            for w_idx in possible_w_idx:
+                if h_idx != w_idx and node_shape[w_idx] * node_shape[h_idx] > 4096:
+                    return False
+
+        return True
+
+    @staticmethod
+    def _get_dim_order(node: Node):
+        if node.meta.get(NXP_NODE_FORMAT) is None:
```
MartinPavella: This is not OK. We need to have the format known here.

In the problematic case, the `is_supported` methods are called by the function `check_op_support_extended()` which is returned by `NeutronConverter.ops_to_not_decompose()`.

`ops_to_not_decompose()` is called here: https://github.com/pytorch/executorch/blob/17adba19d0051b4a184f7ff7e7a0612cf8bd3000/exir/program/_program.py#L1385

and `check_op_support_extended()` is used a few lines later: https://github.com/pytorch/executorch/blob/17adba19d0051b4a184f7ff7e7a0612cf8bd3000/exir/program/_program.py#L1396

Perhaps we can call `NodeFormatInference(exported_program).identify_node_formats()` in the `ops_to_not_decompose()`. It is currently being called later, in the `NeutronPartitioner.partition()`.

Please give this a try. Hopefully it will give us access to the node formats.

novak-vaclav:
Fixed the issues as instructed, please review. Thank you! :)

backends/nxp/neutron_partitioner.py
```diff
@@ -477,3 +482,46 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
...
+            aten_op_to_converter[exir_op._op] = converter
+
+        # determine node format so it can be checked if the nodes are delegable
+        NodeFormatInference(ep, only_for_delegation_check=True).identify_node_formats()
+
+        def check_op_support_extended(node: torch.fx.Node):
+            # if the converter does not exist, the operator should not be decomposed
+            node_converter = aten_op_to_converter.get(node.target)
+            if node_converter is None:
+                return True
```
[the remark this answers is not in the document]
novak-vaclav: fixed

backends/nxp/neutron_partitioner.py
```diff
@@ -477,3 +482,46 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
...
+        If the operator is in the list and the evaluation of the callable is True (or the callable is not specified),
+        the operator should not decomposed (i.e. it truly belongs to the list). Otherwise, it should decomposed.
+        """
+        parameters_mapping = EdgeProgramToIRConverter.map_inputs_to_parameters(ep)
+        aten_op_to_converter = {}
+        for exir_op, converter in supported_ops.items():
+            aten_op_to_converter[exir_op._op] = converter
+
+        # determine node format so it can be checked if the nodes are delegable
+        NodeFormatInference(ep, only_for_delegation_check=True).identify_node_formats()
```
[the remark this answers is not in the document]
novak-vaclav: not applicable

backends/nxp/tests/executorch_pipeline.py
```diff
@@ -132,14 +135,18 @@ def to_quantized_edge_program(
...
     ]
 
     edge_program_manager = to_edge_transform_and_lower(
         export(exir_program_aten__module_quant, example_input, strict=True),
         transform_passes=NeutronEdgePassManager(),
         partitioner=partitioners,
-        compile_config=EdgeCompileConfig(_check_ir_validity=False),
+        compile_config=EdgeCompileConfig(
+            _check_ir_validity=False,
+            _core_aten_ops_exception_list=[exir_ops.edge.aten.prelu.default],
```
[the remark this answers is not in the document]
novak-vaclav: removed it all together - should work

backends/nxp/backend/ir/converter/node_converters/ops_converters/prelu_converter.py
```diff
@@ -0,0 +1,78 @@
...
+            return False
+
+        # According to Neutron spec., height * width cannot be greater than a given constant.
+        if node_shape[w_idx] * node_shape[h_idx] > 4096:
+            return False
+
+        return True
+
+    @staticmethod
+    def _get_dim_order(node: Node):
```
MartinPavella: Nit: 
The name is a bit misleading as the `node` actually has a dim order attribute `node.meta['val'].dim_order()`, which is not what this method returns. And your method only returns 3 value, whereas the actual `dim_order` would have 4 (+1 for the batch) . 
I would consider a different name. Perhaps something like `_get_channels_and_spatial_dimension_indices()` (or something shorter).

backends/nxp/backend/node_format_inference.py
```diff
@@ -218,6 +224,26 @@ def _handle_node_which_uses_channels_first_format(self, node: Node):
...
+        Determines if the node is one of the following:
+        - output
+        - lowered_module
+        - getitem
+        - executorch_call_delegate
+        I.e. making it non-delegable
+        """
+
+        if isinstance(node.target, str):
+            target_str = str(node.target)
```
MartinPavella: Nit: This has no effect, right?

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,126 @@
...
+@pytest.mark.parametrize(
+    "input_shape",
+    [
+        pytest.param((1, 8, 24, 32), id="4D."),
+    ],
+)
+def test_prelu_2_partitions_test(mocker, input_shape):
+    # Run conversion
+    edge_program = to_quantized_edge_program(
+        TwoPartitionPReLUModel(), [input_shape, input_shape]
```
MartinPavella, line 94: Hmm, I would like to see this test done with the `channels_last` dim order, but the support for that is not yet in our `executorch_pipeline.py` (Roman is adding it now).
Could you please add a `# TODO (Martin) Add a channels last dim order variant of this test to verify correct partitioning.` here?

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,126 @@
...
+    channels = input_shape[-1]
+    _ = to_quantized_edge_program(
+        LinearPReLUModule(in_features=channels, out_features=channels),
+        input_shape,
+    )
+
+    # Capture converted program
+    edge_program: ExportedProgram = converter_spy.call_args.args[1]
+
+    # Check that `prelu` was properly decomposed into simpler edge operators since it could not be delegated
```
MartinPavella: The `assert` you wrote does not check this.
You are checking the graph which was converted to NeutronIR and delegated. So it probably just contains the `Linear` node. You should keep the `assert` to make sure the `PRelu` was not delegated, but you should add another one, where you check that the edge program returned by `to_quantized_edge_program()` contains the ops that the `PRelu` was decomposed into.

backends/nxp/neutron_partitioner.py
```diff
@@ -487,16 +488,40 @@ def ops_to_not_decompose(
...
-            which will be called for each node in the graph that users can use as a filter for certain
-            nodes that should be continued to be decomposed even though the op they correspond to is
-            in the list returned by ops_to_not_decompose.
+        Method to determine which operators SHOULD NOT be decomposed to edge dialect.
+
+        The method returns:
+            a list of operators NOT to decompose
+            optional callable - filter
+        If the operator is in the list and the evaluation of the callable is True (or the callable is not specified),
+        the operator should not decomposed (i.e. it truly belongs to the list). Otherwise, it should decomposed.
```
MartinPavella: Well explained, thank you 👍🏻 

Nit: ... should not **be** decomposed ... should **be** decomposed.

backends/nxp/neutron_partitioner.py
```diff
@@ -487,16 +488,40 @@ def ops_to_not_decompose(
...
         """
-        return self.preserve_ops, self.check_op_support
+        parameters_mapping = EdgeProgramToIRConverter.map_inputs_to_parameters(ep)
```
MartinPavella, line 505: Hmm, it is possible that in the future, we will want to prevent the decomposition of nodes which require static data for their `is_supported_*` methods. Since this is called during partitioning, the model only contains `FakeTensors`. 
Please add a comment to this line, with something like:
```
# The parameters_mapping will only contain `FakeTensor` objects since the partitioning is done with a "Fake model".
# If a node is selected to not be decomposed and it requires the real static parameters in its 'is_supported_*' methods, the state dict from the quantized model will need to be provided here.
```

backends/nxp/neutron_partitioner.py
```diff
@@ -487,16 +488,40 @@ def ops_to_not_decompose(
...
+            aten_op_to_converter[exir_op._op] = converter
+
+        # determine node format so it can be checked if the nodes are delegable
+        NodeFormatInference(ep, only_for_delegation_check=True).identify_node_formats()
+
+        def check_op_support_extended(node: torch.fx.Node):
+            # if the converter for the node does not exist, the operator should be decomposed
+            node_converter = aten_op_to_converter.get(node.target)
+            if node_converter is None:
+                return False
```
MartinPavella, line 525: I understand that this is done to avoid the exception (that an unexpected aten node was left in the edge model), but can lead to unintuitive cases, where someone will try to not decompose some node by adding it to the list, but it will silently get decomposed anyway.
Please consider printing a warning message in this rare case.

backends/nxp/backend/node_format_inference.py
```diff
@@ -226,10 +252,14 @@ def _handle_node_which_can_use_any_node_format(self, node: Node):
...
         #  Therefore, it's format must be marked as `channels first` here, which will cause it to be converted to
         #  channels last later during the conversion stage.
-        if hasattr(val := node.meta["val"], "dim_order") and is_channels_last_dim_order(
-            list(val.dim_order())
-        ):
-            self._assign_format_to_node(node, DataFormat.CHANNELS_FIRST)
+
+        # For the purpose of checking delegation compatibility, node format must be inferred. Nodes that cannot be delegated
+        #  may not always have the format specified. Thus the node format inference is skipped.
+        if not self.only_for_delegation_check or not self._is_non_delegable_node(node):
```
robert-kalmar: This condition is hard to understand, You have 2 negations (` not self.only_for_delegation_check`, `not self._is_non_delegable_node(node)` ) and the `self._is_non_delegable_node(node)` is negation on its own, effectivelly nesting the negations. 

What about refactoring: 
`_is_non_delegable_node` --> `_is_potentially_delegable_node`, which will return true for standard nodes, and false for getitem, lowered_module, output and executorch_call_delegate. 

Then the condition become: 
`if not self.only_for_delegation_check or self._is_potentially_delegable_node(node):` 
What makes it clearer that the nested code runs if: 
- we run in "Neutron Backend" (i.e. not during the partitioning)
- or we run in partitioning but the node is "potentially_delegable"

backends/nxp/backend/node_format_inference.py
```diff
@@ -49,7 +51,11 @@ class NodeFormatInference:
     # List of all edge operations, which are supported by the converter.
     _known_targets: list[EdgeOpOverload]
 
-    def __init__(self, edge_program: ExportedProgram):
+    def __init__(self, edge_program: ExportedProgram, only_for_delegation_check=False):
+        # If true, the format is inferred only for nodes that are eligible for delegation,
+        # excluding nodes like `output`, `getitem`, etc.
+        self.only_for_delegation_check = only_for_delegation_check
```
robert-kalmar: Parameter name is misleading. [Delegation](https://docs.pytorch.org/executorch/0.3/concepts.html#delegation) per ExecuTorch glosary is the process of running part of the graph on accelerator, in the AoT case converting the subgraph for a particular backend. 

What you are introducing here is the check for the [partitioning](https://docs.pytorch.org/executorch/0.3/concepts.html#partitioner) phase. 

```suggestion
        self.only_for_partitioning_check = only_for_partitioning_check
```
or just
```suggestion
        self.check_for_partitioning = check_for_partitioning
```

backends/nxp/tests/executorch_pipeline.py
```diff
@@ -120,6 +120,8 @@ def to_quantized_edge_program(
         is_qat=use_qat,
     )
 
+    # List of operators to not decompose during the lowering.
+    preserve_ops = [torch.ops.aten.gru.input, torch.ops.aten.prelu.default]
```
robert-kalmar: Remove the GRU as we are not yet converting it. 
Otherwise a model with GRU node will fail to convert to executorch.

backends/nxp/nxp_backend.py
```diff
@@ -187,7 +187,11 @@ def preprocess(  # noqa C901
             edge_program._verifiers = [
                 EXIREdgeDialectVerifier(
                     class_only=True,
-                    core_aten_ops_exception_list=[torch.ops.aten.max_pool2d.default],
+                    core_aten_ops_exception_list=[
+                        torch.ops.aten.max_pool2d.default,
+                        torch.ops.aten.gru.input,
```
robert-kalmar: Remove GRU

backends/nxp/backend/ir/converter/node_converters/ops_converters/prelu_converter.py
novak-vaclav: fixed

backends/nxp/backend/node_format_inference.py
novak-vaclav: I refactored this part. However conversion to string is still necessary for checking using `startswith`

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
novak-vaclav: done.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
novak-vaclav: fixed. Is it okay now?

backends/nxp/neutron_partitioner.py
novak-vaclav: thank you. Fixed.

backends/nxp/neutron_partitioner.py
novak-vaclav: done.

backends/nxp/neutron_partitioner.py
novak-vaclav: done.

backends/nxp/backend/node_format_inference.py
novak-vaclav: Nice point, thank you! Fixed.

backends/nxp/backend/node_format_inference.py
novak-vaclav: If `only_for_delegation_check = True`, then `NodeFormatInference` determines `NodeFormat` only for the nodes that can be delegable (i.e. not `output` etc.). That was the idea.
However the name `only_for_partitioning_check` is quite unintuitive for me. May I suggest `only_for_op_support_check`/`check_for_op_support`?

backends/nxp/tests/executorch_pipeline.py
novak-vaclav: done.

backends/nxp/nxp_backend.py
novak-vaclav: done.

backends/nxp/backend/node_format_inference.py
robert-kalmar: Understand the idea behind this addition. 
Both `only_for_op_support_check` or `check_for_op_support`  sounds good to me.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    input_data = (
+        (2 * np.random.random(input_shape).astype(np.float32) - 1) * 50
+    ).astype(np.int8)
+
+    convert_run_compare(
+        exported_program,
+        input_data,
+    )
```
MartinPavella: Why not use the `neutron_ir_model` here?
If you don't provide it, the function `convert_run_compare` will convert the `exported_program` to NeutronIR again, which should result in exactly `neutron_ir_model` again. But I guess there is some risk of producing a different result (I can't imagine how though, but you never know), and it slows down the tests.
I would just pass in the `neutron_ir_model` here.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,128 @@
...
+    # Run conversion
+    channels = input_shape[-1]
+    edge_program = to_quantized_edge_program(
+        LinearPReLUModule(in_features=channels, out_features=channels),
+        input_shape,
+    ).exported_program()
+
+    # Check `prelu` was not delegated (only `linear` was)
+    edge_nodes = list(edge_program.graph.nodes)
+    assert sum(str(n.target).startswith("lowered_module") for n in edge_nodes) == 1
```
MartinPavella: Well done, with these checks I am confident the behavior is as we expect 👍🏻 

Just 1 thing, you can look  for `executorch_call_delegate` nodes instead of  `lowered_module` nodes. These always come together to represent 1 Neutron delegated partition, and the  `executorch_call_delegate` has the target `torch._higher_order_ops.executorch_call_delegate`, so you don't have to compare strings. Lately we have been making a switch from comparing against string targets where possible.

```python
assert len(n.target = torch._higher_order_ops.executorch_call_delegate for n in edge_nodes) == 1
```

backends/nxp/backend/node_format_inference.py
novak-vaclav: I decided to go with `only_for_op_support_check`. So this is fixed.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
novak-vaclav: Fixed.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
novak-vaclav: Fixed.

MartinPavella: I can still see it here 😃 
You fixed it in a different place (which is good), but please fix it here too.

novak-vaclav: Yeah, sorry. Should be fixed everywhere now.

backends/nxp/tests/executorch_pipeline.py
roman-janik-nxp, line 1: The commit [Add aten.gru.input to preserve ops list to prevent decomposition](https://github.com/pytorch/executorch/pull/17156/changes/6507bc0cd9e1f11aced4bba1018b90718c517cec) should not be a part of this PR. In the second commit, you removed the gru.input, but it is still in the name of the commit. I don't mind 2 commits, but the first should be named "Add aten.prelu.default to preserve ops list to prevent decomposition".

backends/nxp/backend/node_format_inference.py
```diff
@@ -218,6 +224,23 @@ def _handle_node_which_uses_channels_first_format(self, node: Node):
...
+        - output
+        - lowered_module
+        - getitem
+        - executorch_call_delegate
+        The nodes listed above are non-delegable.
+        """
+        return (
+            node.target != "output"
+            and node.target != operator.getitem
+            and node.target != torch.ops.higher_order.executorch_call_delegate
```
roman-janik-nxp, line 240: Please switch to `torch._higher_order_ops.executorch_call_delegate` with underscore.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,137 @@
...
+    ).exported_program()
+
+    # Capture generated entities
+    neutron_ir_model, _ = converter_spy.spy_return
+    exported_program: ExportedProgram = converter_spy.call_args.args[1]
+
+    # Check `prelu` was not decomposed into simpler edge operators
+    assert graph_contains_any_of_ops(
+        exported_program.graph,
+        [torch.ops.aten.prelu.default, exir_ops.edge.aten.prelu.default],
```
roman-janik-nxp: This and next assert are suspicious. Are you sure both aten and edge variants should be included in the check?

backends/nxp/tests/executorch_pipeline.py
robert-kalmar: Once merged it will be squashed to 1 commit. So at this stage removing a commit is minor update, not having impact on final result.

backends/nxp/backend/node_format_inference.py
robert-kalmar: For curiosity, why you suggest using torch's private function `._higher_order_ops.executorch_call_delegate` instead of public `torch.ops.higher_order.executorch_call_delegate`?

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
novak-vaclav: That's true, the assert checked the wrong thing. Fixed it to check for decomposition into edge `edge.aten.where`, `edge.aten.mul` etc. Thank you

backends/nxp/backend/node_format_inference.py
roman-janik-nxp: I suggest it because the delegate node target is exactly `torch._higher_order_ops.executorch_call_delegate`  when you check during runtime.

Still not done.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,146 @@
...
+    assert (
+        sum(
+            n.target == torch._higher_order_ops.executorch_call_delegate
+            for n in edge_nodes
+        )
+        == 1
+    )
+
+    # Check `prelu` was decomposed into simpler edge operators
+    assert graph_contains_any_of_ops(edge_program.graph, [exir_ops.edge.aten.gt.Scalar])
```
roman-janik-nxp: Please put these 3 checks into one, the function accepts list of node targets.

backends/nxp/backend/node_format_inference.py
novak-vaclav: When both options work, I feel like it is better to use the public alternative (ie. without `_`) since it is designed to be used outside of the module it was imported from.

novak-vaclav: What do you think @robert-kalmar ?

novak-vaclav:
Rebased onto `main` to fix issues with internal tests.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -0,0 +1,146 @@
...
+    [
+        pytest.param((1,), id="1D not supported."),
+        pytest.param((1, 8), id="2D not supported."),
+        pytest.param((1, 8, 16), id="3D not supported."),
+        pytest.param((1, 8, 16, 32, 64), id="5D not supported."),
+        pytest.param((1, 8, 16, 31), id="channels must be divisible by NUM_MACS"),
+        pytest.param((1, 8, 1024, 8), id="width*height is too big (limit 4096)"),
+    ],
+)
+def test_invalid_prelu_conversion(mocker, input_shape):
```
roman-janik-nxp:
```suggestion
def test_prelu__no_delegation__unsupported_conversion(mocker, input_shape):
```
Current name suggest the conversion is invalid. The suggestion is inspired by naming in our other tests and is more explicit.

backends/nxp/backend/node_format_inference.py
novak-vaclav: Or @MartinPavella ??

robert-kalmar: Agree with @novak-vaclav , this is the implementation in [torch](https://github.com/pytorch/pytorch/blob/main/torch/_ops.py#L1419):
```
class _Ops(types.ModuleType):
    __file__ = "_ops.py"

    def __init__(self):
        super().__init__("torch.ops")
        self.loaded_libraries = set()
   -->self.higher_order = _HigherOrderNamespace()
        self._dir = []
```
--> 
```
class _HigherOrderNamespace(types.ModuleType):
    __file__ = "torch.ops"

    def __init__(self) -> None:
        super().__init__("torch.ops.higher_order")
        self._dir: list[str] = []

    def __iter__(self) -> Iterator[str]:
        return iter(self._dir)

    def __getattr__(self, name: str) -> HigherOrderOperator:
        # Following _OpNamespace.__getattr__, we cache the op on this object.
        op = _higher_order_ops.get(name)
        if op is None:
            raise AttributeError(
                f"'_HigherOrderNamespace' 'torch.ops.higher_order' object has no attribute '{name}'"
            )
        setattr(self, name, op)
        self._dir.append(name)
  --> return op
```
Once you query it (`__getattr__`) it looks up in the ` _higher_order_ops` dictionary `, and returns that operator. So in the condition it becomes the same object: 
```
node.target (=> torch._higher_order_ops.executorch_call_delegate)  != torch.ops.higher_order.executorch_call_delegate (=>torch._higher_order_ops.executorch_call_delegate)
```

The internal implementation (e.g. change of the dictionary name) might change quietly, and it is not considered as API change, but the `ops.higher_order` is considered as API, and if someone changes it, one shall apply the change  with all the usage or mark it `@deprecated` to let users to notice the upcoming change.

novak-vaclav:
Fixed all the issues mentioned by @roman-janik-nxp and Copilot.

Added another commit to fix checking for `executorch_call_delegate` target using `torch.ops.higher_order.executorch_call_delegate` and not `torch._higher_order_ops.executorch_call_delegate` throughout the NXP backend codebase.

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
MartinPavella: That is not correct.

Your suggestion would make the test less strict as `graph_contains_any_of_ops()` returns `True` if at least 1 of the listed operators is in the model. The original implementation was better at identifying potential bugs, which is preferred in tests in my opinion.

robert-kalmar (approved):
Looks good to me, 
just please update the test_prelu_converter.py as Martin pointed out here https://github.com/pytorch/executorch/pull/17156#discussion_r2817354245

novak-vaclav:
> Looks good to me, just please update the test_prelu_converter.py as Martin pointed out here [#17156 (comment)](https://github.com/pytorch/executorch/pull/17156#discussion_r2817354245)

Fixed.

backends/nxp/backend/node_format_inference.py
novak-vaclav: Fixed for all instances in the `NXP` codebase, commit dd17cd78c7bd99eab8a75025068eaeda22e1dc6b

backends/nxp/tests/ir/converter/node_converter/test_prelu_converter.py
```diff
@@ -29,6 +29,10 @@ def reseed_model_per_test_run():
     np.random.seed(23)
 
 
+# noinspection PyProtectedMember
```
MartinPavella, line 32: Nit: This is not necessary. Same in `test_neg_converter.py`, `test_context_sensitive_delegation.py`, `test_view_copy_converter.py` and `test_convert_upsample_*.py` modules.

MartinPavella (approved):

MartinPavella:
Once the internal build passes, we can merge.

roman-janik-nxp (approved):

roman-janik-nxp:
The failing checks seem unrelated and the internal build is passing. Merging.

PR #17181, opened 2026-02-03
Infinite loop in XNNExecutor::resize_outputs

lucylq:
### Summary
size_t --> ssize_t so it's signed and won't loop indefinitely.

### Test plan
```
cd build
cmake .. -DEXECUTORCH_BUILD_TESTS=ON -DEXECUTORCH_BUILD_XNNPACK=ON
cmake --build . --target backends_xnnpack_test
ctest -R  backends_xnnpack_test -V
```


backends/xnnpack/runtime/XNNExecutor.cpp
```diff
@@ -246,7 +246,7 @@ ET_NODISCARD Error XNNExecutor::resize_outputs(Span<EValue*> args) const {
     if (out_tensor->scalar_type() == ScalarType::Long) {
       int64_t* data_64 = out_tensor->mutable_data_ptr<int64_t>();
       const int32_t* data_32 = out_tensor->const_data_ptr<int32_t>();
-      for (size_t j = out_tensor->numel() - 1; j >= 0; --j) {
```
GregoryComer, line 249: Wow, I'm surprised this has wasn't caught before.

GregoryComer (approved):

PR #17184, opened 2026-02-03
Arm backend: Fix test_nss.py import error when ng_model_gym is not installed

apullin:
Summary:
Fixes test collection failure in CI when `ng_model_gym` is not installed.

The `test_nss.py` file imports `ng_model_gym` at module level, which requires `pydantic`. In CI environments where these dependencies are not installed, pytest fails during test collection with `ModuleNotFoundError: No module named 'pydantic'`.

This fix adds `pytest.importorskip("ng_model_gym")` before the import, which gracefully skips the entire test module when the dependency is unavailable instead of failing test collection.

Differential Revision: D92206283


mergennachin (approved):

zingo:
Hi, thanks for the fix.
We have already merged a separate PR to fix the issue do we want to merge this anyway?

apullin:
> Hi, thanks for the fix. We have already merged a separate PR to fix the issue do we want to merge this anyway?

@zingo I was just trying to unblock myself. I have some additions in a diff stack, which appeared to be failing because of this. If this is known fixed, we can decline?

zingo:
> > Hi, thanks for the fix. We have already merged a separate PR to fix the issue do we want to merge this anyway?
> 
> @zingo I was just trying to unblock myself. I have some additions in a diff stack, which appeared to be failing because of this. If this is known fixed, we can decline?

Ah OK then I understand,  this is hopefully working now without this change now as the other PR is merged. Can you tests if you are unblocked without this change?

PR #17199, opened 2026-02-04
NXP Backend: Add option for prefetching weights from external memory

jirioc:
### Summary
Models that do not fit into SRAM can be placed into external memory. Since the Neutron has no access to external FLASH, the firmware needs to prefetch the weights from FLASH to SRAM in advance. This commit instructs the Neutron Converter to create extra instructions for such prefetching. Each prefetching is started in parallel with the compute, but can take longer, so this feature needs to be used only when necessary. 

This commit changes both the AoT part and runtime.

### Test plan
Unit test added to backends/nxp/tests/test_neutron_converter_manager.py

### DIFF
--- a/backends/nxp/backend/neutron_converter_manager.py
+++ b/backends/nxp/backend/neutron_converter_manager.py
@@ -72,7 +72,19 @@ class NeutronConverterManager:
                 f"Target `{target}` is not a valid target. Must be one of `{valid_targets}`."
             )
 
-    def convert(self, tflite_model: bytes, target: str) -> bytes:
+    def convert(
+        self, tflite_model: bytes, target: str, fetch_constants_to_sram: bool = False
+    ) -> bytes:
+        """
+        Call Neutron Converter.
+
+        :param tflite_model: A generic TFLite model to be converted.
+        :param target: The target platform.
+        :param fetch_constants_to_sram: Add microcode that fetches weights from external memory.
+        This allows running models which do not fit into SRAM. Applies to Neutron-C only (microcontrollers).
+
+        :return: TFLite model with Neutron microcode as bytes.
+        """
         # Neutron converter crashes if we provide invalid target -> verify.
         self.verify_target(target)
 
@@ -82,6 +94,7 @@ class NeutronConverterManager:
         cctx.compilationOpts.excludeGraphPasses = (
             "HoistSliceAboveTranspose,MergeTranspose"
         )
+        cctx.compilationOpts.fetchConstantsToSRAM = fetch_constants_to_sram
 
         # Try to use multiprocessing for isolation, but fall back to direct execution
         # if the environment doesn't support it (e.g., in sandcastle/build environments)
--- a/backends/nxp/nxp_backend.py
+++ b/backends/nxp/nxp_backend.py
@@ -46,6 +46,7 @@ class NeutronCompileSpecBuilder:
         self.operators_not_to_delegate: List[str] = []
         self.neutron_converter_flavor = None
         self.use_neutron_for_format_conversion = True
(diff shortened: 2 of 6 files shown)


backends/nxp/nxp_backend.py
```diff
@@ -73,6 +75,8 @@ def neutron_compile_spec(
             use_neutron_for_format_conversion: If True, the EdgeProgramToIRConverter will insert `Transpose` ops to
                                                 ensure that the IO matches the executorch partition, which will be
                                                 delegated to Neutron.
+            fetch_constants_to_sram: If True, the Neutron Converter will insert microinstructions to prefetch weights
+                                     from FLASH to SRAM. This should be used when the whole model does not fit into SRAM.
```
MartinPavella, line 79: Nit:
Perhaps such a doc-string could be useful elsewhere as well. For example in `NeutronConverterManager.convert()`.

MartinPavella (approved):

backends/nxp/nxp_backend.py
jirioc: Done in a new commit.

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -79,7 +80,18 @@ typedef struct {
   const uint8_t* outputTranspositionFlags;
   const uint8_t* inputMap;
   const uint8_t* outputMap;
-} NeutronConfig;
+} NeutronExecutorchConfig;
+
+#ifdef EXTERNAL_MEM
```
robert-kalmar, line 87: This needs to be conditionally enabled? If needed also update the CMakeLists.txt to enable setting this value.

backends/nxp/runtime/NeutronBackend.cpp
```diff
@@ -70,6 +70,7 @@ typedef struct {
   int numOutputs = 0;
   int numInputArgs = 0;
   uint32_t scratchSize = 0;
+  uint32_t sramScratchSize = 0;
```
robert-kalmar, line 74: The sramScratchSize is not used when the `EXTERNAL_MEM` is not defined.
Can we add covert it with ifdef too? Now it is inconsistent as some part of the code is ifdefed some not.

backends/nxp/backend/neutron_converter_manager.py
```diff
@@ -72,7 +72,19 @@ def verify_target(self, target: str):
                 f"Target `{target}` is not a valid target. Must be one of `{valid_targets}`."
             )
 
-    def convert(self, tflite_model: bytes, target: str) -> bytes:
+    def convert(
+        self, tflite_model: bytes, target: str, fetch_constants_to_sram: bool
```
robert-kalmar: NIT: You can define default value for the fetch_constants_to_sram.

backends/nxp/runtime/NeutronBackend.cpp
jirioc: This needs to be enabled only in examples that are storing models in flash, I think it should stay off by default to save memory (otherwise memcpy is added by the linker unnecessarily). In MCUXpresso it can be selectively enabled per-example in flags.cmake file.

backends/nxp/runtime/NeutronBackend.cpp
jirioc: Added in the new commit.

backends/nxp/backend/neutron_converter_manager.py
jirioc: Added in the new commit.

robert-kalmar (approved):

jirioc:
Failing tests are unrelated. Merging.

PR #17201, opened 2026-02-04
Optimization enn executor Runner

Jiseong-oh:
### Summary
 - Runtime update and appended runtime only build option
 - Build Runtime with NDK
 - Append required lines for enn-types
 - Thread-safe code appended to api_impl
 - support iteration in enn_executor 
 - Separate Enn init from model load

### Test plan
$ enn_executor_runner -model <model> .... -iteration 100

cc @SS-JIA @digantdesai @kimishpatel @kirklandsign @larryliu0820


SS-JIA:
Changes LTM - however, the samsung CI signals are currently failing

SS-JIA:
Samsung tests seem to have been cancelled this time around.

Failing with

```
The job has exceeded the maximum execution time of 3h0m0s
```

Is it because it's waiting to reserve a device and none are available?

Jiseong-oh:
> Samsung tests seem to have been cancelled this time around.
> 
> Failing with
> 
> ```
> The job has exceeded the maximum execution time of 3h0m0s
> ```
> 
> Is it because it's waiting to reserve a device and none are available?

I found one peculiarity. Even after CI was over, Samsung's devices continued to connect due to ping from CI server. When the CI was completed, I thought there was no need for connection anymore. I will make a PR to solve the the issue.

SS-JIA:
Got it. My preference would be to merge that fix first, then rebase this PR on top of the fix, so that we can get signal on this PR. Is that ok?

Jiseong-oh:
> My preference would be to merge that fix first, then rebase this PR on top of the fix, so that we can get signal on this PR. Is that ok?

I agree with you.

SS-JIA (approved):
Samsung tests pass, thus LGTM.

PR #17206, opened 2026-02-04
Arm backend: Make setting artifact_path optional for tests

Erik-Lundell:
dump_artifact dir is meant for getting artifacts,
not testing infra. This caused issues in preprocess, where it was impossible to separate the cases
 - user has set dump_artifact path, wants artifacts
 - test uses dump_artifact, doesn't want artifacts

A hack was used where debug info was output unless the artifact path was in a temporary directory.
However, this is a confusing if a user *intends*
to output debug artifacts into a temporary directory.


zingo (approved):

zingo:
This was reverted by https://github.com/pytorch/executorch/pull/17288

PR #17210, opened 2026-02-04
Neutron backend: Add NXP Executor Runner

robert-kalmar:
### Summary
Add NXP executor runner demonstrating the use of Neutron Backend in executorch on eIQ Neutron NPU Simulator (eIQ NSYS). 

### Test plan
The `run.sh` e2e example is part of the unittest-nxp-neutron job. 

cc @JakeStevens @digantdesai

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -966,15 +966,17 @@ jobs:
         # The generic Linux job chooses to use base env, not the one setup by the image
         CONDA_ENV=$(conda env list --json | jq -r ".envs | .[-1]")
         conda activate "${CONDA_ENV}"
+        
+        # Install eIQ packages
+        pip install -r backends/nxp/requirements-eiq.txt
 
         # Build and install Executorch
         PYTHON_EXECUTABLE=python \
-        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON" \
+        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON -DEXECUTORCH_BUILD_NXP_NEUTRON_RUNNER=ON " \
         .ci/scripts/setup-linux.sh --build-tool "cmake"
 
         # Install test requirements
         pip install -r backends/nxp/requirements-tests-pypi.txt
-        pip install -r backends/nxp/requirements-tests-eiq.txt
         PYTHON_EXECUTABLE=python bash examples/nxp/setup.sh
 
         # Run pytest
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -628,6 +628,10 @@ if(EXECUTORCH_BUILD_NXP_NEUTRON)
   list(APPEND _executorch_backends executorch_delegate_neutron)
 endif()
 
+if(EXECUTORCH_BUILD_NXP_NEUTRON_RUNNER)
```
robert-kalmar, line 631: @kirklandsign , @larryliu0820 we are adding a `nxp_executor_runner` application working with our simulator for the eIQ Neutron NPU. As prerequisity it need the `eiq_neutron_sdk` package to be installed, otherwise the build fails. 

This way, I letting you to consider if you want to have this option available or rather remove it. The nxp_executor_runner can be build also separately via `examples/nxp/executor_runner/CMakeLists.txt`.

.github/workflows/pull.yml
```diff
@@ -933,15 +933,17 @@ jobs:
...
+        pip install -r backends/nxp/requirements-eiq.txt
 
         # Build and install Executorch
         PYTHON_EXECUTABLE=python \
-        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON" \
+        CMAKE_ARGS="-DEXECUTORCH_BUILD_NXP_NEUTRON=ON -DEXECUTORCH_BUILD_NXP_NEUTRON_RUNNER=ON " \
         .ci/scripts/setup-linux.sh --build-tool "cmake"
 
         # Install test requirements
         pip install -r backends/nxp/requirements-tests-pypi.txt
```
MartinPavella, line 979: Nit:
Can this be moved up as well to keep it combined with
```
# Install test requirements
pip install -r backends/nxp/requirements-eiq.txt
```
a few lines above.

Perhaps it would be a bit clearer to not have `# Install test requirements` twice like this.

tools/cmake/preset/default.cmake
MartinPavella, line 5: Nit:
Should the NXP copyright be added here?

MartinPavella (approved):

.github/workflows/pull.yml
robert-kalmar: Renamed the comment. 
I will keep the order as in to let executorch to install python packages in preferred version, and rest of the packages for testing to accomodate.

robert-kalmar:
The only related failure is the [Lint / link-check / lint-urls / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/21979519630/job/63498468794?pr=17210), which fails unreachable URL. This URL targets the  folder introduced by this change - https://github.com/pytorch/executorch/blob/main/examples/nxp/executor_runner/ .

Merging, the lint-urls check shall pass once merged.

roman-janik-nxp (approved):

PR #17213, opened 2026-02-04
Pad channels to 4 for xnnpack test

lucylq:
### Summary
Seems like XNNPACK reads 16 bytes at a time, causing reads into uninitialized memory unless channels=4.

### Test plan
```
cmake --build . --target backends_xnnpack_test
ctest -R  backends_xnnpack_test -V
```

Internally:
```
buck test @fbcode//mode/dev fbcode//executorch/backends/xnnpack/test:xnnexecutor_test -- --exact 'fbcode//executorch/backends/xnnpack/test:xnnexecutor_test - XNNExecutorTest.ResizeOutputsWithLongTensorConvertsInt32ToInt64'
```


GregoryComer (approved):

GregoryComer:
Yeah, XNNPACK assumes that any tensor memory extends 16 bytes past the end of data. This is handled normally during memory planning (ET + XNN), but has to be handled manually for tests.

PR #17217, opened 2026-02-04
Add slimtensor INTERFACE target and refactor CMake dependencies

Gasoonjia:
Summary:
This change refactors the CMake build system for AOTI backends to properly structure SlimTensor dependencies:

1. **Created `slimtensor` INTERFACE target**: SlimTensor is a header-only library, so it should be an INTERFACE target that provides include directories and conditionally links CUDA runtime when building with CUDA support.

2. **Simplified `aoti_common_shims_slim`**:
   - Now properly depends on the `slimtensor` INTERFACE target
   - Removed redundant include directories (inherited from slimtensor)
   - Removed redundant CUDA configuration (inherited from slimtensor)
   - Removed `extension_tensor` dependency (not used by shim functions)
   - Added `slim/cuda/guard.cpp` to sources

3. **Refactored `aoti_cuda_shims`**:
   - Now links to `aoti_common_shims_slim` instead of compiling `common_shims_slim.cpp` and `slim/cuda/guard.cpp` directly
   - This creates a cleaner dependency graph and avoids code duplication

4. **Updated `aoti_cuda_backend`**:
   - Moved `extension_tensor` dependency here (where it's actually used for ETensor operations in cuda_backend.cpp)

The new dependency graph is:
```
aoti_cuda_backend
    ├── extension_tensor
    └── aoti_cuda_shims
            └── aoti_common_shims_slim
                    └── slimtensor (INTERFACE)
```

Also remove a unused header file.

Differential Revision: D92309483


backends/cuda/runtime/shims/memory_slim.h
```diff
@@ -1,175 +0,0 @@
-/*
```
larryliu0820, line 1: why is this ok to delete?

larryliu0820 (approved):

backends/cuda/runtime/shims/memory_slim.h
Gasoonjia: `memory_slim.h` is a temporary header file used during slimtensor migration. Now memory.h has been migrated to slimtensor-based shim layers, `memory_slim.h` is no longer needed.

PR #17222, opened 2026-02-04
Runner to utilize ring buffer

kirklandsign:
### Summary
Allow exceeding context window

### Test plan
CI

### DIFF
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -1322,6 +1322,7 @@ def _load_llama_model(llm_config: LlmConfig) -> "LLMEdgeManager":
             #  Module]`.
             model.vocab_size,
             llm_config.base.metadata,
+            use_ring_buffer=llm_config.model.local_global_attention is not None,
         ),
     )
 
--- a/examples/models/llama/export_llama_lib.py
+++ b/examples/models/llama/export_llama_lib.py
@@ -1231,6 +1231,7 @@ def _load_llama_model_metadata(
     n_layers: int,
     vocab_size: int,
     metadata_str: Optional[str] = None,
+    use_ring_buffer: bool = True,
 ):
     metadata = {
         "get_max_seq_len": max_seq_len,
@@ -1240,6 +1241,7 @@ def _load_llama_model_metadata(
         "use_kv_cache": use_kv_cache,
         "use_sdpa_with_kv_cache": use_sdpa_with_kv_cache,
         "enable_dynamic_shape": enable_dynamic_shape,
+        "use_ring_buffer": use_ring_buffer,
     }
     if metadata_str:
         try:
--- a/examples/models/llama/main.cpp
+++ b/examples/models/llama/main.cpp
@@ -184,6 +184,7 @@ int32_t main(int32_t argc, char** argv) {
   }
 
   auto error = runner->generate(prompt, config);
+  auto error2 = runner->generate(prompt, config);
   if (error != executorch::runtime::Error::Ok) {
     ET_LOG(Error, "Failed to run llama runner");
     return 1;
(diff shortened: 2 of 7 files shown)


examples/models/llama/export_llama_lib.py
```diff
@@ -1320,6 +1322,7 @@ def _load_llama_model(llm_config: LlmConfig) -> "LLMEdgeManager":
             #  Module]`.
             model.vocab_size,
             llm_config.base.metadata,
+            use_ring_buffer=llm_config.model.local_global_attention is not None,
```
larryliu0820, line 1325: I don't think this is right. local global attention uses a sliding window which may or may not relevant to ring buffer.

larryliu0820: I think the high level seems right, we need to have this metadata serialized into .pte.

examples/models/llama/main.cpp
```diff
@@ -184,6 +184,7 @@ int32_t main(int32_t argc, char** argv) {
   }
 
   auto error = runner->generate(prompt, config);
+  auto error2 = runner->generate(prompt, config);
```
larryliu0820, line 187: ?

extension/llm/runner/multimodal_runner.cpp
```diff
@@ -183,16 +183,25 @@ Error MultimodalRunner::generate(
       get_rss_bytes() / 1024.0 / 1024.0);
 
   // Resolve max_new_tokens based on config
+  // Check if ring buffer is enabled - if so, we can exceed context length
+  bool use_ring_buffer = metadata_.at(kUseRingBuffer);
```
larryliu0820, line 187: I think we need to set this in llm_runner_helper not here.

extension/llm/runner/multimodal_runner.cpp
```diff
@@ -183,16 +183,25 @@ Error MultimodalRunner::generate(
...
   int64_t max_context_len =
       metadata_.at(kMaxContextLen) - 0; // No start_pos offset
-  int32_t max_new_tokens = config.resolve_max_new_tokens(max_context_len, pos_);
+  // When ring buffer is enabled, use a large context length to allow unlimited
+  // generation.
+  int64_t effective_context_len =
+      use_ring_buffer ? INT64_MAX : max_context_len;
+  int32_t max_new_tokens =
+      config.resolve_max_new_tokens(effective_context_len, pos_);
```
larryliu0820, line 195: This logic needs to be applied to text_llm_runner as well.

extension/llm/runner/text_llm_runner.cpp
```diff
@@ -129,33 +129,45 @@ Error TextLLMRunner::generate(
...
+  }
+
+  // Determine max_new_tokens using the GenerationConfig's resolve method.
+  // When ring buffer is enabled, use a large context length to allow unlimited
+  // generation.
+  int64_t effective_context_len =
+      use_ring_buffer ? INT64_MAX : max_context_len;
   int max_new_tokens =
-      config.resolve_max_new_tokens(max_context_len, num_prompt_tokens);
+      config.resolve_max_new_tokens(effective_context_len, num_prompt_tokens);
```
larryliu0820, line 159: Seems like duplicate code to me.

extension/llm/runner/llm_runner_helper.cpp
```diff
@@ -96,6 +96,7 @@ get_llm_metadata(tokenizers::Tokenizer* tokenizer, Module* module) {
       {llm::kMaxContextLen, 128},
       {llm::kUseKVCache, true},
       {llm::kUseSDPAWithKVCache, false},
+      {llm::kUseRingBuffer, true},
```
larryliu0820, line 99: I think we need to set max_context_len to INT MAX here

PR #17238, opened 2026-02-05
transforms: Fix bug in SDPA decomposition

oscarandersson8218:
When SDPA with causal=True was lowered, a bug in decompose_sdpa was discovered. It assumed that all args are nodes which caused a crash when this is not the case. The pass has been updated to instead access node.all_input_nodes instead of node.args. New test cases are added to backends/arm/test/ops/test_sdpa.py. 


zingo:
Hi @SS-JIA  here is a PR the needs non Arm review :) Thanks for all the help!

SS-JIA (approved):
LGTM - change to decompose_sdpa.py should be safe

zingo (approved):

oscarandersson8218:
Unrelated failures.

PR #17248, opened 2026-02-05
Integrating multiple operators (iDMA + Cache) onto cadence/vision on a chosen, stable, ExecuTorch branch.

cad-rlc:
### Summary
Following operators integrated:

1. Softmax (iDMA + Cache)
2. Maxpool (Cache based)
3. Add (w/o broadcasting) (iDMA + Cache)
4. ReLu (iDMA + Cache)
5. Quantize (float to int8) (iDMA + Cache)
6. Dequantize (int8 to float) (iDMA + Cache)

### Test plan
Commands used for testing can be found in backends/cadence/build_cadence_vision.sh

P.S. - This is based on an older version of ExecuTorch, which we have treated as a stable branch. 

@mcremon-meta @hsharma35  @zonglinpengmeta


1
```diff
@@ -0,0 +1,25 @@
+Collecting matplotlib
```
JakeStevens, line 1: seems like this was added accidentally

backends/cadence/aot/ref_implementations.py
```diff
@@ -6,29 +6,17 @@
...
-
-    custom_libs: list[Path] = list(
-        Path(__file__)
-        .parent.parent.parent.resolve()
-        .glob("**/kernels/quantized/**/*custom_ops_generated_lib.*")
-    )
-    if custom_libs:
-        torch.ops.load_library(str(custom_libs[0]))
-    del Path
+torch.ops.load_library("./pip-out/lib.linux-x86_64-cpython-311/executorch/kernels/quantized/libquantized_ops_aot_lib.so")
```
JakeStevens, line 19: not sure this change should be upstreamed

mcremon-meta:
@cad-rlc  is this ready for review, or still under development? Just let us know when it is!

PR #17255, opened 2026-02-05
Expose hooks to load ET pybindings with your own data loader

JacobSzwejbka:
Summary: Allow using out of tree data loaders with the pybindings.

Differential Revision: D92431122

### DIFF
--- a/extension/pybindings/pybindings_data_loader.h
+++ b/extension/pybindings/pybindings_data_loader.h
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <memory>
+
+#include <executorch/runtime/core/data_loader.h>
+#include <executorch/runtime/core/freeable_buffer.h>
+#include <executorch/runtime/core/result.h>
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch {
+namespace extension {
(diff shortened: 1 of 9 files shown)


extension/pybindings/pybindings_data_loader.h
```diff
@@ -0,0 +1,78 @@
...
+#include <executorch/runtime/core/freeable_buffer.h>
+#include <executorch/runtime/core/result.h>
+#include <executorch/runtime/platform/compiler.h>
+
+namespace executorch {
+namespace extension {
+namespace pybindings {
+
+/// DataLoader wrapper holding a shared_ptr, allowing sharing between Python
+/// and C++ while Module takes ownership via unique_ptr.
```
lucylq, line 23: @JacobSzwejbka bit confused, how is the DataLoader being used here?

JacobSzwejbka: Module requires a unique ptr for a dataloader but what the pybindings pass around is a shared pointer of one so we have to wrap it

JacobSzwejbka: We could pipe through a unique_ptr probably but its not very pythonic imo

lucylq: I'm wondering if we need a shared ptr here - it doesn't seem like this pattern is used anywhere else in the pybindings file, but maybe I am missing something.. 

Is the loader needed after being passed the Module? Otherwise feel like it makes sense for Module to just take ownership of it.

lucylq (approved):
Lgtm, have a question on the shared ptr still - what's the usecase of having it, and think the PR could be a bit simpler with unique_ptr without loss of functionality?

PR #17257, opened 2026-02-05
Speed up named data map + add uint32 support

metascroy:
Summary:
his diff significantly speeds up named data map insertion by switching from bytes(tensor.untyped_storage()) to tensor.numpy().tobytes() for converting tensors to bytes. The new _tensor_to_bytes() helper function provides a much faster conversion path, which should noticeably improve XNNPACK and Vulkan lowering times since these backends heavily use the named data store during serialization.

Additionally, this diff adds torch.uint32 to the ScalarType mapping, enabling support for unsigned 32-bit integer tensors in ExecuTorch.

Changes:
_named_data_store.py: Added _tensor_to_bytes() helper that uses numpy().tobytes() for fast tensor-to-bytes conversion, with special handling for bfloat16 (which numpy doesn't support natively)
tensor.py: Added torch.uint32 → ScalarType.UINT32 mapping

Differential Revision: D92447071

### DIFF
--- a/exir/_serialize/_named_data_store.py
+++ b/exir/_serialize/_named_data_store.py
@@ -7,14 +7,30 @@
 # pyre-strict
 
 import hashlib
-
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Union
 
 import torch
-
 from executorch.exir._serialize.data_serializer import DataEntry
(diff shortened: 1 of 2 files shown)


GregoryComer (approved):
Nice find.

PR #17258, opened 2026-02-05
Metal backend: use HQQ in Parakeet quantization

manuelcandales:
Bump ao pin to get https://github.com/pytorch/ao/pull/3829, and use HQQ in Parakeet quantization


mergennachin (approved):
Thanks. 

Observe the transcription correctness in the CI output before landing

https://github.com/pytorch/executorch/actions/runs/21731366078/job/62686508905?pr=17258

PR #17276, opened 2026-02-06
Arm backend: Default output_order_workaround=False

oscarandersson8218:
output_order_workaround was enabled by default. This can cause failures. Also addresses serialization issue of boolean in compile spec.


zingo (approved):

zingo (changes requested):
This seem to make the PTE a few bytes bigger and the size tests number also needs a bump

zingo (approved):

PR #17280, opened 2026-02-06
Arm backend: Improve noop handling

AdrianLundell:
- Removes the identity node_visitor in favour of handling all noops in the remove_noop_pass
- Adds detach_copy to the partitioner is_noop check to avoid partitioning graphs with single detach_copy ops
- Make noop check functions in the partitioner private and remove public documentation.

Instead handle alias_copy similar to other noops


zingo (approved):
OK to merge if all arm-tests pass and rest matches main

AdrianLundell:
Fails unrelated

PR #17300, opened 2026-02-08
Title: Add MobileNetV2 pytest for Cortex-M backend

psiddh:
Summary:
                                                                                        
  Added RemoveNoopPass (from ARM backend) to CortexMPassManager — it checks that              
  _clone_dim_order has matching input/output dtypes before removing it, so it won't strip any 
  actual type or dim_order conversions. This eliminates the identity clone from MV2's         
  flatten/view and unblocks the runtime.
                                                                                              
  Lowered qtol from 20 to 10 — passes reliably. Didn't go lower since calibration data is 
  random (torch.randn) so the numerical diff varies across runs. The argmax assertion provides
   an additional correctness check independent of tolerance.

  Removed xfail from test_dialect_mv2 (stable now), kept it on test_implementation_mv2 with
  strict=False for potential flakiness in the serialization/runtime path.

### Test plan
pytest -c backends/arm/test/pytest.ini backends/cortex_m/test/models/test_mobilenet_v2.py -v

### DIFF
--- a/backends/cortex_m/test/models/test_mobilenet_v2.py
+++ b/backends/cortex_m/test/models/test_mobilenet_v2.py
@@ -1,0 +1,89 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from executorch.backends.arm.test.common import parametrize
+
+from executorch.backends.cortex_m.test.tester import CortexMTester, McuTestCase
+from executorch.backends.test.harness.stages import StageType
+from torchvision import models
+
+
+ops_before_transforms: dict[str, int] = {
+    "executorch_exir_dialects_edge__ops_aten_add_Tensor": 10,
+    "executorch_exir_dialects_edge__ops_aten_avg_pool2d_default": 1,
+    "executorch_exir_dialects_edge__ops_aten_convolution_default": 52,
+    "executorch_exir_dialects_edge__ops_aten_hardtanh_default": 35,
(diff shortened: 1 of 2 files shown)


psiddh:
Note : test_implementation_mv2 is marked XFAIL because MV2's graph contains a dim_order_ops._clone_dim_order.default 
  op (from the flatten/view before the classifier) that MV3 doesn't have. The runtime doesn't include a kernel for    
  this op (use_portable_ops=False).                                                                                   
                                                                                                                      
Should we add ARM's RemoveNoopPass to CortexMPassManager to remove identity clone ops.

AdrianLundell:
Yes if it is a noop I think it makes sense to remove, just make sure that it does not do any dim_order/ type conversion first. Also see if you can lower the numerical tolerance compared to mv3 and still get it to pass, would be interesting to know.

backends/cortex_m/test/models/test_mobilenet_v2.py
```diff
@@ -0,0 +1,96 @@
...
+    ),
+}
+
+
+@parametrize(
+    "test_case",
+    test_cases,
+    xfails={
+        "mobilenet_v2": "MLETORCH-XXX - Investigate mobilenet_v2 flakiness"
+    },
```
rascani: Shouldn't this be on the test_implementation_mv2?

psiddh: Yes you're right, Removed xfail from test_dialect_mv2 (stable now), kept it on test_implementation_mv2 with
  strict=False for potential flakiness in the serialization/runtime path.

rascani (approved):

PR #17302, opened 2026-02-09
Qualcomm AI Engine Direct - Support SLC allocator feature

shewu-quic:
Summary:
- Support SLC allocator feature setting
- Support spill-fill buffer for hybrid mode
- Fixed redundant convert_op in 8w8a quantization config

Test Plan:
- Check  the debug log to confirm if the SLC allocator is enabled.
```
 I [Qnn ExecuTorch]: QnnDsp <V> SLC_ALLOCATOR is set to 1 for graph 0
```
- Check the debug log whether the spill-fill buffer is configured for the LLM in hybrid mode.
```
 python examples/qualcomm/oss_scripts/llama/llama.py -b build-android -H ${HOST} -s ${DEVICE} -m SM8750 --temperature 0 --model_mode hybrid --max_seq_len 128 --prefill_ar_len 32 --decoder_model smollm2_135m --prompt "I would like to learn python, could you teach me with a simple example?"  --artifact {ARTIFACTS}
```

### DIFF
--- a/backends/qualcomm/serialization/qc_compiler_spec.fbs
+++ b/backends/qualcomm/serialization/qc_compiler_spec.fbs
@@ -181,6 +181,11 @@ table QnnExecuTorchHtpBackendOptions {
   /// When multiple graphs appear inside the same context,
   /// weights could be reused across all graphs.
   use_weight_sharing:bool;
+
+  /// Allows user to enable the usage of the System Level Cache Allocator for a given graph. 
+  /// It will help the by reducing overall bandwith on the use case.
+  /// The feature is only supported by specific SOCs.
+  use_slc_allocator:bool;
 }
 
 /// Logging level of the delegate and QNN backend.
--- a/backends/qualcomm/aot/python/PyQnnManagerAdaptor.cpp
+++ b/backends/qualcomm/aot/python/PyQnnManagerAdaptor.cpp
@@ -207,6 +207,7 @@ PYBIND11_MODULE(PyQnnManagerAdaptor, m) {
       .def("Init", &PyQnnManager::Init)
       .def("InitBackend", &PyQnnManager::InitBackend)
       .def("InitContext", &PyQnnManager::InitContext)
+      .def("InitContextCache", &PyQnnManager::InitContextCache)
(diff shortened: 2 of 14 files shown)


shewu-quic:
Hi @cccclai, @billmguo,

This PR introduces the use_slc_allocator option, which enables users to utilize the System Level Cache Allocator for a specific graph. This feature can help save overall bandwidth in the use case.

When you have time, could you help review this PR?
Thanks

billmguo:
Look good to me.

billmguo (approved):

backends/qualcomm/serialization/qc_compiler_spec.fbs
```diff
@@ -181,6 +181,11 @@ table QnnExecuTorchHtpBackendOptions {
   /// When multiple graphs appear inside the same context,
   /// weights could be reused across all graphs.
   use_weight_sharing:bool;
+
+  /// Allows user to enable the usage of the System Level Cache Allocator for a given graph. 
+  /// It will help the by reducing overall bandwith on the use case.
+  /// The feature is only supported by specific SOCs.
+  use_slc_allocator:bool;
```
cccclai, line 188: looks like we're adding a new feature, can we update the read me regarding how to use it?

cccclai: Possibly a bit more explanation on System Level Cache Allocator

shewu-quic: Yes, this is a new feature. Users just need to set use_slc_allocator=True in compile_spec to enable it.
https://github.com/pytorch/executorch/pull/17302/changes#diff-0439f6a7c1a3a3cfb222cd6409b6754f17a1ce782dd231de1d12bbf957d588f7R1000

System Level Cache Allocator is a shared cache at the system level of a SoC, serving as the last caching layer before external DDR memory. Its primary purpose is to optimize memory bandwidth, thereby potentially improving performance and reducing power consumption. However, it is model-dependent, so it is not guaranteed to be effective in all cases.

cccclai: Thanks, this is helpful. Can we add this information somewhere

PR #17304, opened 2026-02-09
Arm backend: Reject ops with mixed fp type inputs + tests

Erik-Lundell:


zingo (approved):
OK to merge when you get Arm tests to pass I see:

FAILED backends/arm/test/models/test_mobilenet_v2_arm.py::test_mv2_tosa_FP_bf16 - AssertionError: Output 0 does not match reference output.
FAILED backends/arm/test/models/test_resnet18.py::test_resnet_18_tosa_FP_bf16 - AssertionError: Output 0 does not match reference output.

PR #17305, opened 2026-02-09
NXP backend: Add support for `aten.leaky_relu.default`.

MartinPavella:
### Summary
Add support for `aten.leaky_relu.default` to the NXP backend.

### Test plan
Unit-tests provided.


MartinPavella:
Internal build: https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH75-1

MartinPavella:

backends/nxp/backend/ir/converter/node_converters/ops_converters/leaky_relu_converter.py
```diff
@@ -0,0 +1,42 @@
...
+
+class LeakyReluConverter(NodeConverter):
+
+    @staticmethod
+    def _is_supported_in_IR(
+        node: Node,
+        parameters_mapping: dict[str, Parameter],
+        custom_delegation_options: CustomDelegationOptions,
+    ) -> bool:
+        return True
```
novak-vaclav, line 25: In the neutron converter supported operators documentation, I see that `leaky_relu` is only supported under these conditions:
```
inpN == 1
inpC % NUM_MACS == 0
```

maybe add a check for this?

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+@pytest.fixture(autouse=True)
+def reseed_model_per_test_run():
+    torch.manual_seed(42)
+    np.random.seed(23)
+
+
+# noinspection PyProtectedMember
+ExecutorchDelegateCall = torch._higher_order_ops.executorch_call_delegate
```
novak-vaclav: We have stumbled upon an interesting conversation about whether it is better to use the public target `torch.ops.higher_order.executorch_call_delegate`, or private `torch._higher_order_ops.executorch_call_delegate`, if both work fine.

I think it is better to use the public variant since it it designed to be used outside of the module it was imported from and assumes the user is agnostic to what private overload is used.

https://github.com/pytorch/executorch/pull/17156#discussion_r2794411685

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
```diff
@@ -0,0 +1,183 @@
...
+    # Make sure the `leaky_relu` was delegated.
+    assert graph_contains_any_of_ops(delegated_ep.graph, [ExecutorchDelegateCall])
+    assert not graph_contains_any_of_ops(delegated_ep.graph, [LeakyRelu2D])
+
+    # Verify correct behavior of the converted NeutronIR model.
+    intermediate_ep = converter_spy.call_args.args[1]
+    neutron_ir_model, _ = converter_spy.spy_return
+
+    input_data = (
+        np.random.random(input_shape).astype(np.float32) * 256.0 - 128.0
```
novak-vaclav, line 44: Just out of curiosity, why is there `-128.0`?

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
```diff
@@ -0,0 +1,183 @@
...
+    [
+        (5,),
+        (4, 5),
+        (3, 4, 5),
+        (2, 3, 4, 5),
+        (1, 2, 3, 4, 5),
+    ],
+    ids=lambda input_shape: f"{len(input_shape)}D",
+)
+def test_convert_leaky_relu__ranks(mocker, input_shape: tuple[int, ...]):
```
novak-vaclav, line 117: The code of all the tests seems to be almost identical. I would try to extract it into a separate function, where one of the arguments would be the `model`.

docs/source/backends/nxp/op-support.csv
```diff
@@ -9,6 +9,7 @@ aten.clone.default,int8,static int8,
 aten.constant_pad_nd.default,int8,static int8,"H or W padding only"
 aten.convolution.default,int8,static int8,"1D or 2D convolution, constant weights, groups=1 or groups=channels_count (depthwise)"
 aten.hardtanh.default,int8,static int8,"supported ranges: <0,6>, <-1, 1>, <0,1>, <0,inf>"
+aten.leaky_relu.default,int8,static int8,
```
novak-vaclav, line 13: If you decide to check for constraints imposed by Neutron Converter on what `leaky_relu` is supported (ie. `num_channels % 8 == 0`, ...), then it should be mentioned here.

novak-vaclav (changes requested):

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
MartinPavella: Just to cover the maximum `int8` range (`[-128, 127]`).

backends/nxp/backend/ir/converter/node_converters/ops_converters/leaky_relu_converter.py
MartinPavella: The `leaky_relu` is actually converted to Neutron `Activation` (https://bitbucket.sw.nxp.com/projects/AITEC/repos/neutron-software/browse/neutron-converter/docs/SupportedOperatorsC.md#Activation), which does not have these restrictions.

As you can see, my tests use a variety of input shapes (such as `(23,)` or `(3, 4, 5)`), which work just fine.

docs/source/backends/nxp/op-support.csv
MartinPavella: Not applicable. More details here: https://github.com/pytorch/executorch/pull/17305#discussion_r2817457432

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
MartinPavella: Good point. 
Personally I'm not a fan of extracting `assert` statements into separate functions in tests, as it makes it a bit harder to debug when something fails.

But I will do it here as these 4 tests are really nearly identical, and it makes it easier to review and avoid bugs in the tests.

backends/nxp/tests/ir/converter/node_converter/test_leaky_relu_converter.py
MartinPavella: I also prefer not to use the private version. I will update it, thanks.

novak-vaclav:
Thank you for the fixes, looks good to me 👍

roman-janik-nxp (approved):
👍

MartinPavella:
The failing `unittest-nxp-neutron` is unrelated and will be fixed in another PR. The other failing checks seem unrelated and the internal CI is passing. Mergning.

PR #17306, opened 2026-02-09
Arm backend: Add FP16 support to operators pt.4-6

martinlsm:
 - ge
 - gt
 - alias_copy
 - index_tensor
 - identity
 - mul
 - matmul
 - log
 - max_pool2d
 - view
 - pow
 - reciprocal
 - rsqrt
 - sigmoid

The index_tensor node visitor (in op_index_tensor.py) was missing dtype
validation. Add such validation while including the new fp16 dtype.

Clean up defining of test suites in test_matmul.py.


zingo (approved):
OK to merge if Arm and other tests are OK

PR #17288, opened 2026-02-07
Revert "Arm backend: Make setting artifact_path optional for tests"

lucylq:
This reverts commit 52bdacc16b40448a9f2993229474f4cf665347b2.

### Summary
This is breaking internal CI in https://www.internalfb.com/diff/D92366066

Specifically, breaking CI for all the _INT variant tests across multiple ops (avg_pool2d, tanh, rsqrt, conv2d, cat, addmm, add).

cc @psiddh

PR #17310, opened 2026-02-09
Reland  #17206

lucylq:
Reland https://github.com/pytorch/executorch/pull/17206

After reverting: https://github.com/pytorch/executorch/pull/17288

Reason - we have a fwd fix in internal now.


psiddh:
cc @zingo @Erik-Lundell

lucylq:
landed with: https://github.com/pytorch/executorch/pull/17313

PR #17314, opened 2026-02-09
Validate dim_order is a permutation in dim_order_to_stride

rascani:
### Summary
The validate_dim_order function only checked that values were in bounds, allowing invalid inputs like {0, 0, 0} to pass. This caused uninitialized memory access in dim_order_to_stride_nocheck.

Fix by using a bitmask to detect duplicates. Also adds test fixture with runtime_init() for error logging and removes duplicate include.

### Test plan
```
./test/run_oss_cpp_tests.sh
```

### DIFF
--- a/runtime/core/exec_aten/util/dim_order_util.h
+++ b/runtime/core/exec_aten/util/dim_order_util.h
@@ -8,13 +8,13 @@
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


lucylq (approved):
Thanks!

rascani:
Failures are unrelated.

PR #17315, opened 2026-02-09
Remove unnecessary cuda sync for better perf

Gasoonjia:
* __->__ #17315

Right now we always do cudasync before existing cudabackend.execution(). However we only need that when copying data from gpu to cpu; any actions happen inside a same stream do not need explicit sync.

Differential Revision: [D92193164](https://our.internmc.facebook.com/intern/diff/D92193164/)


backends/aoti/slim/core/storage.h
```diff
@@ -127,13 +127,13 @@ struct DeviceTraits<c10::DeviceType::CUDA> {
...
+    // TODO(gasoonjia): add cuda stream as a member of MaybeOwningStorage to
+    // support multiple devices.
     auto stream_result = executorch::backends::cuda::getCurrentCUDAStream(-1);
-    if (stream_result.ok()) {
-      ET_CUDA_LOG_WARN(cudaFreeAsync(ptr, stream_result.get()));
-    } else {
-      // Fallback to synchronous free if we can't get the stream
-      ET_CUDA_LOG_WARN(cudaFree(ptr));
-    }
+    ET_CHECK_MSG(stream_result.ok(), "Failed to get current CUDA stream");
```
larryliu0820, line 135: Just to be clear this is turning a warning into a fatal error. Just making sure this is your intention.

backends/cuda/runtime/cuda_backend.cpp
```diff
@@ -313,10 +366,27 @@ class ET_EXPERIMENTAL CudaBackend final
...
-    ET_CUDA_CHECK_OR_RETURN_ERROR(cudaStreamCreate(&cuda_stream));
-    handle->cuda_stream = static_cast<void*>(cuda_stream);
+
+    // Use shared CUDA stream if enabled via options, otherwise create one.
+    // A shared stream ensures proper ordering across multiple methods
+    // (e.g., encoder, decoder, sampler) when using skip-copy optimization.
+    if (is_using_shared_cuda_stream()) {
+      // Shared stream mode: set handle's stream to nullptr.
+      // The stream will be retrieved from backend in execute().
+      handle->cuda_stream = nullptr;
```
larryliu0820: I think it's better to set handle->cuda_stream to the only cuda stream.

backends/aoti/slim/core/storage.h
Gasoonjia: Yes that's my plan. We should guarantee that each tensor should get the correct cuda_stream

backends/cuda/runtime/cuda_backend.cpp
Gasoonjia: based on offline sync we tried to create a new CudaHandle class to inherit current aoti_handle, puting a shared_ptr<cuda_stream> inside cuda_handle, to make sure there's only one cuda stream in the whole pipeline.

larryliu0820 (approved):
Thank you for splitting aoti_delegate_handle and cuda_delegate_handle, it's much cleaner this way.

PR #17327, opened 2026-02-10
NXP backend: Add support for `aten.clamp.default`.

MartinPavella:
### Summary
This PR adds support for the `aten.clamp.default` operator no NXP backend.

### Test plan
Unit-tests provided.


MartinPavella:
Internal build: https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH76-1

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
```diff
@@ -0,0 +1,126 @@
...
+        node: Node,
+        partition_list: list[Partition],
+        custom_delegation_options: CustomDelegationOptions,
+        neutron_target_spec: NeutronTargetSpec,
+        parameters_mapping: dict[str, Parameter],
+    ) -> bool:
+        bounds = cls._get_clamp_bounds(node)
+
+        if bounds in [(0, 6), (0, None)]:  # Relu6  # Relu
+            # If this is the only operator in the partition, NeutronConverter will not create a NeutronNode for some
```
roman-janik-nxp, line 78: I think Neutron doesn't support single activation nodes in the partition. It's worth checking and if so adding this check to all affected activation converters.

MartinPavella:

roman-janik-nxp (approved):

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
MartinPavella: `ReluN1To1` and `Relu0To1` can be the only nodes in a partition. I have a test for this case (`test_convert_clamp__single_op__no_delegation__custom_bounds`).
I now see that the test name is not good, so I will rename it.

For the activations which cannot be alone in a partition, I have created a ticket: https://jira.sw.nxp.com/browse/EIEX-735

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
```diff
@@ -0,0 +1,126 @@
...
+    SUPPORTED_CLAMP_BOUNDS = [
+        (-1, 1),  # ReluN1To1
+        (0, 1),  # Relu0To1
+        (0, 6),  # Relu6
+        (0, None),  # Relu
+    ]
+
+    # noinspection PyShadowingBuiltins
+    @staticmethod
+    def _get_clamp_bounds(clamp_node: Node) -> tuple[float | None, float | None]:
```
novak-vaclav, line 36: Isn't it possible that clamp bound can be `int`? If so, then the type annotation should be modified.

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
```diff
@@ -0,0 +1,126 @@
...
+        cls,
+        node: Node,
+        partition_list: list[Partition],
+        custom_delegation_options: CustomDelegationOptions,
+        neutron_target_spec: NeutronTargetSpec,
+        parameters_mapping: dict[str, Parameter],
+    ) -> bool:
+        bounds = cls._get_clamp_bounds(node)
+
+        if bounds in [(0, 6), (0, None)]:  # Relu6  # Relu
```
novak-vaclav: I would convert `SUPPORTED_CLAMP_BOUNDS` to a `dict` so:
```
SUPPORTED_CLAMP_BOUNDS = {
    "ReluN1To1": (-1, 1),
    "Relu0To1": (0, 1),
    "Relu6": (0, 6),
    "Relu": (0, None),
}
```

then you could use it here like this:
```
if bound in [SUPPORTED_CLAMP_BOUNDS["Relu6"], SUPPORTED_CLAMP_BOUNDS["Relu"]]:
```

to make it more legible and not use the comments. It would also come in handy in the `convert` method.

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
```diff
@@ -0,0 +1,126 @@
...
+                Tensor self,
+                Scalar? min=None,
+                Scalar? max=None
+            ) -> Tensor
+        """
+        self.assert_convertible(node)
+
+        bounds = self._get_clamp_bounds(node)
+
+        match bounds:
```
novak-vaclav: Modifying `SUPPORTED_CLAMP_BOUNDS` as stated above would enable you to remove the comments and make the code more legible in my opinion:
```
match bounds:
    case SUPPORTED_CLAMP_BOUNDS["ReluN1To1"]:
        ....
    case SUPPORTED_CLAMP_BOUNDS["ReluN1To1"]:
```

renaming `SUPPORTED_CLAMP_BOUNDS` to something shorter may result in even more legible code, maybe just `SUPPORTED_CLAMPS` or `SUPPORTED_BOUNDS`.

backends/nxp/tests/ir/converter/node_converter/test_clamp_converter.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@pytest.fixture(autouse=True)
+def reseed_model_per_test_run():
+    torch.manual_seed(42)
+    np.random.seed(23)
+
+
+# noinspection PyProtectedMember
+ExecutorchDelegateCall = torch._higher_order_ops.executorch_call_delegate
```
novak-vaclav: Would switch to public version of the target.

docs/source/backends/nxp/op-support.csv
```diff
@@ -2,9 +2,10 @@ Operator,Compute DType,Quantization,Constraints
 aten.abs.default,int8,static int8,
 aten._adaptive_avg_pool2d.default,int8,static int8,"ceil_mode=False, count_include_pad=False, divisor_override=False"
 aten.addmm.default,int8,static int8,2D tensor only
-aten.add.Tensor,int8,static int8,"alpha = 1, input tensor of rame rank"
+aten.add.Tensor,int8,static int8,"alpha = 1, input tensor of name rank"
 aten.avg_pool2d.default,int8,static int8,"ceil_mode=False, count_include_pad=False, divisor_override=False"
 aten.cat.default,int8,static int8,"input_channels % 8 = 0, output_channels %8 = 0"
+aten.clamp.default,int8,static int8,"Bounds = (-1, 1) or (0, 1) or (0, 6) or (0, None)"
```
novak-vaclav, line 8: I would modify the string so it is clear the bounds correspond to `ReluN1To1`, `Relu0To1`, ...

backends/nxp/tests/ir/converter/node_converter/test_clamp_converter.py
```diff
@@ -0,0 +1,175 @@
...
+    [
+        pytest.param(0, 6, id="min = 0, max = 6 (Relu6)"),
+        pytest.param(0, 1, id="min = 0, max = 1 (Relu0To1)"),
+        pytest.param(-1, 1, id="min = -1, max = 1 (ReluN1To1)"),
+        pytest.param(0, None, id="min = 0, max = None (Relu)"),
+        # float bounds.
+        pytest.param(0.0, 6.0, id="min = 0.0, max = 6.0 (Relu6)"),
+        pytest.param(0.0, 1.0, id="min = 0.0, max = 1.0 (Relu0To1)"),
+        pytest.param(-1.0, 1.0, id="min = -1.0, max = 1.0 (ReluN1To1)"),
+        pytest.param(0.0, None, id="min = 0.0, max = None (Relu)"),
```
novak-vaclav, line 69: Nice touch testing both `int` and `float` bounds 👍

novak-vaclav (changes requested):

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
MartinPavella: Yes, the bounds can be `int`, but the type hint `float` can by used to annotate both `float` and `int` data types. (https://peps.python.org/pep-0484/#the-numeric-tower)
image

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
MartinPavella: Good point. I will use your idea :D

docs/source/backends/nxp/op-support.csv
MartinPavella: Is there a point in the user understanding that the `clamp` is converted to `relu` variants? 
From my point of view, this is just an implementation detail, and the user does not need to be confused with how we enable the op.

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
MartinPavella: Indexing the dictionary doesn't work like this in a `case` statement (https://stackoverflow.com/questions/66159432/how-to-use-values-stored-in-variables-as-case-patterns).
So, I refactored it a bit differently.

MartinPavella:
The failing checks are unrelated and the internal build is passing. Merging.

zingo:
Hi this might have got the size tests to fail I see you got 
[pull / test-binary-size-linux-gcc / linux-job](https://hud.pytorch.org/pr/pytorch/executorch/17327#63973193339) ([gh](https://github.com/pytorch/executorch/actions/runs/22131696831/job/63973193339))

And see this in the hud: https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=test-binary-size&useRegexFilter=true

Note sure why as this PR seem to only touch NXP backend stuff, so it might be unrelated.

MartinPavella:
> Hi this might have got the size tests to fail I see you got [pull / test-binary-size-linux-gcc / linux-job](https://hud.pytorch.org/pr/pytorch/executorch/17327#63973193339) ([gh](https://github.com/pytorch/executorch/actions/runs/22131696831/job/63973193339))
> 
> And see this in the hud: https://hud.pytorch.org/hud/pytorch/executorch/main/1?per_page=50&name_filter=test-binary-size&useRegexFilter=true
> 
> Note sure why as this PR seem to only touch NXP backend stuff, so it might be unrelated.

I have no idea what that test relates to, so I don't know how to investigate the failure. Can I rely on you to find out what's wrong? Or is there something I can do?

PR #17329, opened 2026-02-10
fix to allocate samsung device issue

Jiseong-oh:
### Summary
- Clean up operation is required to create an available device
- CI is complete, but device is not aware of the situation so it should be done by themselves


SS-JIA (approved):
Samsung tests passing, so LGTM

PR #16691, opened 2026-01-19
exir: add flatbuffer Program serialization for performance

chizkiyahu:
## **Title**
Introduce flatbuffer serializer as default for program serialization

## Summary

Add a FlatBuffers-based serializer at `exir/_serialize/_flatbuffer_program.py` and make it the default serializer for program artifacts. The new serializer achieves large efficiency gains across our model suite while retaining robustness via a JSON fallback and detailed error logging.

### Before
python -> json -> binary 

### After
python  -> binary 

## Performance

Measured across ~600 models:

* ~36.3× average speedup (serialize latency).
* ~61.6% average reduction in memory usage during serialization.

## Tests

Unit tests added at:

* `exir/_serialize/test/test_flatbuffer_program.py`

cc @freddan80 @per @zingo @oscarandersson8218 @digantdesai

PR #17333, opened 2026-02-10
exir: add flatbuffer Program serialization for performance

chizkiyahu:
## **Title**
Introduce flatbuffer serializer as default for program serialization

## Summary

Add a FlatBuffers-based serializer at `exir/_serialize/_flatbuffer_program.py` and make it the default serializer for program artifacts. The new serializer achieves large efficiency gains across our model suite while retaining robustness via a JSON fallback and detailed error logging.

### Before
python -> json -> binary 

### After
python  -> binary 

## Performance

Measured across ~600 models:

* ~36.3× average speedup (serialize latency).
* ~61.6% average reduction in memory usage during serialization.

## Tests

Unit tests added at:

* `exir/_serialize/test/test_flatbuffer_program.py`


chizkiyahu:
@mergennachin 

this is the updated PR of https://github.com/pytorch/executorch/pull/16691
thanks

exir/_serialize/_flatbuffer_program.py
```diff
@@ -0,0 +1,237 @@
...
+        data = _create_aligned_byte_vector(
+            builder, _coerce_bytes(self.data), _DELEGATE_ALIGNMENT.get()
+        )
+    _BackendDelegateInlineData.BackendDelegateInlineDataStart(builder)
+    if data:
+        _BackendDelegateInlineData.BackendDelegateInlineDataAddData(builder, data)
+    return _BackendDelegateInlineData.BackendDelegateInlineDataEnd(builder)
+
+
+def _install_fast_packers() -> None:
```
JacobSzwejbka, line 115: Does this make a big difference in serialization speed?

exir/_serialize/_flatbuffer_program.py
```diff
@@ -0,0 +1,237 @@
...
+    global _PACKERS_INSTALLED
```
JacobSzwejbka: @functools.lru_cache(maxsize=1) is maybe a little cleaner then manually maintaining the global

exir/_serialize/_flatbuffer_program.py
```diff
@@ -0,0 +1,237 @@
...
+        storage = _create_aligned_byte_vector(
+            builder, _coerce_bytes(self.storage), _BUFFER_ALIGNMENT.get()
+        )
+    _Buffer.BufferStart(builder)
+    if storage:
+        _Buffer.BufferAddStorage(builder, storage)
+    return _Buffer.BufferEnd(builder)
+
+
+def _pack_backend_delegate_inline_data(self: Any, builder: Any) -> int:
```
lucylq, line 102: Add a comment in the .fbs schema file, if `Buffer` or `BackendDelegateInlineData` tables are changed, to update these functions, as they override the generated packers.

exir/_serialize/_flatbuffer_program.py
chizkiyahu: yes it give very big improvement  at speed

exir/_serialize/generate_program.py
```diff
@@ -0,0 +1,159 @@
...
+        ]
+    )
+    fb_init.write_text("\n".join(fb_init_contents), encoding="utf-8")
+    return 2
+
+
+def main() -> int:
+    repo_root = _repo_root()
+    _run_flatc(repo_root)
+    updated = _rewrite_imports(repo_root)
```
lucylq, line 160: add a comment explaining that flatc generates imports relative `executorch_flatbuffer`, which is why this is necessary? It would be nice if flatc could generate them relative to a specific path...

exir/_serialize/_flatbuffer_program.py
chizkiyahu: quick test on `resnet18` show improve from ~3.54 sec to 0.07 sec  ~50x

lucylq: wow, that's a big speed up!

exir/_serialize/generate_program.py
chizkiyahu: >  It would be nice if flatc could generate them relative to a specific path...

The `flatc` derives the import path from the `namespace` declaration. 
Modifying the `namespace` results in widespread breaking changes across both the Python and C++ codebases. Given those constraints, this approach is the only viable solution I’ve identified that avoids large-scale refactoring.

https://github.com/pytorch/executorch/blob/429925d622479a82ffd2ce80ee1aac8f5a39abc7/schema/program.fbs#L9

exir/_serialize/_flatbuffer_program.py
chizkiyahu: I added

exir/_serialize/generate_program.py
chizkiyahu: comment added

zingo:
amax U55 was broken for a few hour I rebased this to be based after a merge was fixed.

mergennachin:
@lucylq we should import this PR internally and see if it's breaking any of the internal tests

JacobSzwejbka:
Getting a ton of failures on internal CI. Let me triage them some and get back to you.

mergennachin:
@chizkiyahu can you make the following changes? 

https://gist.github.com/mergennachin/6b2ee388f8f138b4dfb774cf4103a8dc

chizkiyahu:
> @chizkiyahu can you make the following changes?
> 
> https://gist.github.com/mergennachin/6b2ee388f8f138b4dfb774cf4103a8dc

@mergennachin 
done 
Thanks

exir/_serialize/_flatbuffer_program.py
```diff
@@ -0,0 +1,234 @@
...
+
+def _set_pack_alignments(tensor_alignment: int, delegate_alignment: int) -> None:
+    _BUFFER_ALIGNMENT.set(tensor_alignment)
+    _DELEGATE_ALIGNMENT.set(delegate_alignment)
+
+
+def _convert_double(val: Double) -> Any:
+    result = _flatbuffer_t_class(Double)()
+    double_val = val.double_val
+    if isinstance(double_val, str):
```
JacobSzwejbka, line 130: Why is this receiving strings?

exir/_serialize/_program.py
```diff
@@ -418,6 +423,35 @@ def _extract_named_data(
...
+    *,
+    constant_tensor_alignment: int,
+    delegate_alignment: Optional[int],
+) -> _FlatbufferResult:
+    """
+    Serializes the Program into a FlatBuffer, with a JSON fallback for robustness.
+
+    The FlatBuffer serialization path is the preferred fast path, offering
+    significantly better runtime performance and lower memory usage in benchmarks.
+    The JSON path is retained solely as a fallback to ensure robustness in cases
```
JacobSzwejbka: Do we know what these cases are? I would actually prefer to not silently fail here.

exir/_serialize/test/test_flatbuffer_program.py
```diff
@@ -0,0 +1,212 @@
...
+            program, constant_tensor_alignment=32, delegate_alignment=64
+        )
+        self.assertGreater(len(result.data), 8)
+        self.assertEqual(result.data[4:6], b"ET")
+        self.assertGreaterEqual(result.max_alignment, 64)
+
+        program2 = _json_to_program(_program_flatbuffer_to_json(result.data))
+        self.assertEqual(program2, program)
+
+    def test_flatbuffer_paths_match(self) -> None:
```
JacobSzwejbka, line 175: Can we add a test to see if they are bitwise accurate, if they are not do we know why?

JacobSzwejbka (changes requested):

JacobSzwejbka:
Internal breakages got fixed in the latest version so just the couple of comments

exir/_serialize/test/test_flatbuffer_program.py
chizkiyahu: I built a minimal test case showing that the outputs are not identical, even with a very small change.

My understanding is that the difference comes from how defaults/optionals are handled.

However, I assume it is not a good idea to push this test into the codebase, because changes in `flatc` could make the test flaky.

```python
    def test_flatbuffer_binary_equality_cases(self) -> None:
        program = Program(  # type: ignore[arg-type]
            version=0,
            execution_plan=None,
            constant_buffer=None,
            backend_delegate_data=None,
            segments=None,
            constant_segment=None,
            mutable_data_segments=None,
            named_data=None,
        )
        with self.subTest(case="minimal"):
            result = _program_to_flatbuffer(program)
            result2 = _program_json_to_flatbuffer(_program_to_json(program))
            self.assertEqual(result.data, result2.data)

        program.constant_segment = SubsegmentOffsets(segment_index=0, offsets=[])
        with self.subTest(case="nontrivial"):
            result = _program_to_flatbuffer(program)
            result2 = _program_json_to_flatbuffer(_program_to_json(program))
            self.assertNotEqual(result.data, result2.data)
            self.assertNotEqual(len(result.data), len(result2.data))
```

exir/_serialize/_flatbuffer_program.py
chizkiyahu: This behavior occurs because the Double definition in the code allows string representations such as "inf" and "-inf":

https://github.com/pytorch/executorch/blob/aebf42ea512d1360337e890e688a013b2a2aa50d/exir/schema.py#L93-L111

The same assumption is reflected in this comment:

https://github.com/pytorch/executorch/blob/aebf42ea512d1360337e890e688a013b2a2aa50d/exir/_serialize/_flatbuffer.py#L326-L330

I will update the comment accordingly and simplify the logic.

exir/_serialize/_program.py
chizkiyahu: I remove the fallback

JacobSzwejbka:
Pulling internally again. Will play around with the bitwise differences to make sure they all checkout.

JacobSzwejbka (approved):

exir/_serialize/_program.py
```diff
@@ -19,8 +20,8 @@
 from executorch.exir._serialize._flatbuffer import (
     _FlatbufferResult,
     _program_flatbuffer_to_json,
```
Gasoonjia, line 22: do we still need this?

chizkiyahu: This is used in the `deserialize_pte_binary` function, which still relies on JSON:
```
program: Program = _json_to_program(
    _program_flatbuffer_to_json(program_data[:program_size])
)
```
As I understand it, `deserialize_pte_binary` is used only for testing and tooling, not in production code. However, since it is part of the public API, I am not entirely sure whether that assumption is correct.

chizkiyahu: @Gasoonjia

Please let me know if you would like me to remove _program_flatbuffer_to_json from deserialize_pte_binary, or even remove it entirely from the codebase.

Alternatively, let me know if you think it is fine to keep it as it is.

lucylq: We should keep `deserialize_pte_binary` - is it possible to deserialize using the new python path instead of json?

JacobSzwejbka:
Its all good on my side. Just need the merge conflict to be resolved. Im happy if you want to take a look at deserialize in a followup I dont think its blocking for this one. @chizkiyahu

chizkiyahu:
>Just need the merge conflict to be resolved

done

exir/_serialize/_program.py
chizkiyahu: is a not a small change 
I will open another PR for this

JacobSzwejbka:
Just wanted to say thanks again @chizkiyahu. This is an awesome change!

PR #17348, opened 2026-02-10
Add quantize fused convbn bias pass

JakeStevens:
Summary:
When performing QAT with a model that has a conv layer with no bias followed by batch norm, the fusion process creates a bias. This is done *after* observers are attached so the resulting bias is kept as float.

This diff adds a pass which grabs the proper qparams and applies them to the non-quantized bias.

Differential Revision: D92733079

### DIFF
--- a/backends/transforms/quantize_fused_convbn_bias_pass.py
+++ b/backends/transforms/quantize_fused_convbn_bias_pass.py
@@ -1,0 +1,355 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import torch
+from torch import fx
+from torch._export.utils import (
+    get_buffer,
+    get_lifted_tensor_constant,
+    get_param,
+    is_lifted_tensor_constant,
+    is_param,
+)
+from torch._guards import detect_fake_mode
+from torch.export.exported_program import InputKind, InputSpec, TensorArgument
+from torch.fx.passes.infra.pass_base import PassBase, PassResult
+
+
+# --- ExportedProgram param helpers ---
+
+
+def _set_param_ep(exported_program, node_or_name, tensor, insert_before=None):
+    """Set or create a parameter in an exported program.
+
+    If node_or_name is a Node, updates the existing parameter or constant value.
+    If node_or_name is a string, creates a new parameter placeholder.
+    """
+    fake_mode = detect_fake_mode(
+        tuple(
+            node.meta["val"]
+            for node in exported_program.graph.nodes
+            if node.op == "placeholder"
+        )
+    )
+
+    if isinstance(node_or_name, fx.Node):
+        node = node_or_name
+        if node.name in exported_program.graph_signature.inputs_to_parameters:
+            name = exported_program.graph_signature.inputs_to_parameters[node.name]
+            exported_program.state_dict[name] = torch.nn.Parameter(
+                tensor, requires_grad=False
+            )
+        elif (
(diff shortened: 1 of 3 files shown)


robert-kalmar:
CC @StrycekSimon @roman-janik-nxp

backends/transforms/quantize_fused_convbn_bias_pass.py
```diff
@@ -0,0 +1,270 @@
...
+        super().__init__()
+        self.exported_program = exported_program
+        self.default_zero_bias = default_zero_bias
+
+    def _is_conv_node(self, node):
+        """Check if node is a convolution operation."""
+        return node.target in (
+            exir_ops.edge.aten.convolution.default,
+            torch.ops.aten.convolution.default,
+            torch.ops.aten.conv2d.default,
```
StrycekSimon: Can you add also transposed convs here? Or is there a reason for omitting them?

backends/transforms/test/test_quantize_fused_convbn_bias_pass.py
```diff
@@ -0,0 +1,208 @@
...
+            torch.ops.aten.convolution.default,
+            torch.ops.aten.conv2d.default,
+        )
+        dequant_targets = (
+            torch.ops.quantized_decomposed.dequantize_per_tensor.default,
+            torch.ops.quantized_decomposed.dequantize_per_channel.default,
+        )
+
+    pass_instance = QuantizeFusedConvBnBiasPass(exported_program)
+    result = pass_instance.call(exported_program.graph_module)
```
StrycekSimon: Is there some postprocessing needed after this line of code? I tried plugging it into our pipeline right after the `calibrate_and_quantize` call (same as here) and it fails on: "TypeError: missing a required argument: 'b__scale_0'" when trying to call `export(result.graph_module, ...)` which we do [here](https://github.com/pytorch/executorch/blob/20dec9d8ce2398bad8afdd7c59b37ecd671ec0f5/backends/nxp/tests/executorch_pipeline.py#L139).

StrycekSimon (changes requested):
I tried running it with our conversion pipeline but not successfully. Seems like the bias is being added as another input of the model. Can you take a look at it? Or is there some postprocessing step needed I am missing?

backends/transforms/test/test_quantize_fused_convbn_bias_pass.py
JakeStevens: Can you put up a draft pr to show how you tried plugging it in? Then I can literate on that and make sure it works appropriately

JakeStevens:
@StrycekSimon the NXP changes and test are now here:

https://github.com/pytorch/executorch/pull/17599

This diff is now "standalone" pass and the integration with your backend in the above

backends/transforms/quantize_fused_convbn_bias_pass.py
```diff
@@ -0,0 +1,270 @@
...
+        )
+
+    def _is_edge_dialect(self, node):
+        """Check if node uses edge dialect operators."""
+        return node.target == exir_ops.edge.aten.convolution.default
+
+    def _get_or_create_bias_node(self, node):
+        """Get existing bias node or create a default zero bias if enabled."""
+        input_dequant, weight_dequant, bias_node, *_ = node.args
```
StrycekSimon: In case there is a biasless convolution in the graph without the following `BatchNorm`, this thing might fail as convolution won't have the third (`bias_node`) argument.

JakeStevens: good catch, I've updated both the code and tests

larryliu0820 (approved):

PR #17365, opened 2026-02-11
Fix CUDA cross-compilation and Windows support

larryliu0820:
- Enable CUDA language in CMake for proper .cu file compilation
- Fix compiler flag conditions to use COMPILE_LANGUAGE for multi-language targets
- Use WIN32 instead of MSVC for MinGW cross-compilation dllexport symbols
- Add MSVC-specific object library linking to preserve AOTI symbols
- Implement proper Windows error formatting in platform.cpp
- Add PAL initialization in CUDA memory shims for Windows DLL isolation
- Implement missing AOTI tensor shim APIs (create_tensor_from_blob, clone)
- Add logging helper function for AOTI-generated code
- Add TORCH_CUDA_ARCH_LIST detection and warning for non-cross-compile scenarios
- Add explicit 'configuration' field to CMake build presets for consistency
- Fix ET_CHECK_TK_OK_OR_RETURN_ERROR macro to use do-while instead of statement expression

This enables CUDA backend compilation on Windows with proper symbol visibility and device architecture detection during cross-compilation scenarios.


mergennachin:
@larryliu0820 unittest-windows failure seems related

backends/cuda/runtime/shims/memory.cpp
```diff
@@ -162,6 +231,38 @@ AOTITorchError aoti_torch_new_tensor_handle(
...
+    SlimTensor** ret_new_tensor) {
+  ET_CHECK_OR_RETURN_ERROR(
+      self != nullptr, InvalidArgument, "aoti_torch_clone: self is null");
+
+  ET_CHECK_OR_RETURN_ERROR(
+      ret_new_tensor != nullptr,
+      InvalidArgument,
+      "aoti_torch_clone: ret_new_tensor is null");
+
+  *ret_new_tensor = new SlimTensor(self->clone());
```
mergennachin, line 260: should it be 

```
*ret_new_tensor = new SlimTensor(self->clone_contiguous());
```

JacobSzwejbka: Not with the "preserve strides" variant is my understanding

backends/cuda/CMakeLists.txt
```diff
@@ -33,6 +33,13 @@ set(CUDA_USE_STATIC_CUDA_RUNTIME OFF)
 
 find_package(CUDAToolkit REQUIRED)
 
+# Ensure CUDA language is enabled so .cu sources are compiled.
+# Without this, some Windows generator/toolchain combinations may silently skip
+# CUDA compilation units in mixed C++/CUDA targets.
+if(NOT CMAKE_CUDA_COMPILER)
```
JacobSzwejbka, line 40: is this for int4?

JacobSzwejbka (approved):

backends/cuda/CMakeLists.txt
larryliu0820: yeah. Need to compile int4mm.cu

backends/cuda/runtime/shims/memory.cpp
```diff
@@ -81,6 +95,61 @@ AOTITorchError aoti_torch_create_tensor_from_blob_v2(
   return Error::Ok;
 }
 
+AOTITorchError aoti_torch_create_tensor_from_blob(
```
Gasoonjia, line 98: why do we need new shim functions? Is the operator used by Parakeet different between linux and windows?

larryliu0820: For some reason the windows cross compiled AOTI shared library is asking for this symbol:

```
(executorch) PS C:\executorch> llvm-readobj --coff-imports C:\\Users\\larryliu\\AppData\\Local\\Temp\\encoder_so_blob7268.so | findstr /i aoti_torch_create_tensor_from_blob
  Symbol: aoti_torch_create_tensor_from_blob_v2 (8)
  Symbol: aoti_torch_create_tensor_from_blob (7)
```

PR #17369, opened 2026-02-11
Arm backend: Scalars must not be get_attr for control ops

SaoirseARM:
### Summary
When we lift out each control flow body as its own GraphModule, any scalar constants that were captured in Python become module attributes. When we get to the backend, we fail as we assume that any getattr operators are Graph Modules. 
This updates:

- Updates ScalarsToAttributePass can handle control flow nodes 
- Ensure that the constants are updated from getattr operators to call_function full operators.


zingo (approved):

zingo (approved):

SaoirseARM:
Failures on the tests are unrelated to this change.

PR #17371, opened 2026-02-11
Arm backend: Add Zephyr Cortex-M STM Nucleo n657x0_q example

zingo:
### Summary
Add information in the zephyr readme on how to build and run
a model on STM Nucleo n657x0_q board.

Moves the Ethos-U driver config to Corstone FVPs board files.
This enable smoother building different targets that will enable
or not enable the Ethos-U drivers/backend depending on board.

This makes it possible to build the example both for Ethos-U
and Cortex-M only targets without change.


rascani (approved):

zingo:
Faild tests are unrelated

PR #17378, opened 2026-02-11
Fix intermittent bus errors in portable pocketfft op on macOS

metascroy:
This PR applies pocketfft patch for ASAN-compatible aligned_alloc on POSIX systems. The portable aligned_alloc in pocketfft stores metadata at ptr[-1], which conflicts with ASAN's heap redzone and causes intermittent bus errors. This patch uses posix_memalign instead, which is ASAN-compatible.

### DIFF
--- a/kernels/optimized/CMakeLists.txt
+++ b/kernels/optimized/CMakeLists.txt
@@ -21,6 +21,36 @@ if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
+# Apply pocketfft patch for ASAN-compatible aligned_alloc on POSIX systems. The
+# portable aligned_alloc in pocketfft stores metadata at ptr[-1], which
+# conflicts with ASAN's heap redzone and causes intermittent bus errors. This
+# patch uses posix_memalign instead, which is ASAN-compatible.
+set(POCKETFFT_DIR "${EXECUTORCH_ROOT}/third-party/pocketfft")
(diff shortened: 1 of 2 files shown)


kernels/optimized/CMakeLists.txt
```diff
@@ -21,6 +21,36 @@ if(NOT EXECUTORCH_ROOT)
   set(EXECUTORCH_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
+# Apply pocketfft patch for ASAN-compatible aligned_alloc on POSIX systems. The
+# portable aligned_alloc in pocketfft stores metadata at ptr[-1], which
+# conflicts with ASAN's heap redzone and causes intermittent bus errors. This
+# patch uses posix_memalign instead, which is ASAN-compatible.
+set(POCKETFFT_DIR "${EXECUTORCH_ROOT}/third-party/pocketfft")
```
manuelcandales, line 28: should we land this upstream instead of having this patch, changing code on the fly?
cc @mergennachin

mergennachin: Yeah, let's try upstreaming to https://github.com/mreineck/pocketfft

metascroy: I've put up a PR here: https://github.com/mreineck/pocketfft/pull/28, but should it not land in pocketfft for some reason, should we not patch in ET given we're distributing the buggy software?

metascroy:
Upstreamed fix. Pin bump PR here: https://github.com/pytorch/executorch/pull/17746

PR #17383, opened 2026-02-11
Add aten::_conj_physical.out operator (#17207)

GeorgeTzoupis:
Summary:

Added `aten::_conj_physical.out` which implements https://docs.pytorch.org/docs/stable/generated/torch.conj_physical.html

Differential Revision: D92991877


JacobSzwejbka (approved):

PR #17396, opened 2026-02-11
Update assertRaisesRegex pattern for Python 3.12 compatibility

drinkmorewaterr:
Summary: Python 3.12 changed the error message for read-only property assignment from "can't set attribute 'spec'" to "property 'spec' of ... object has no setter". Update the regex pattern to match both versions.

Differential Revision: D92997414

### DIFF
--- a/exir/backend/test/test_partitioner.py
+++ b/exir/backend/test/test_partitioner.py
@@ -106,7 +106,7 @@ class TestPartitioner(unittest.TestCase):
 
         with self.assertRaisesRegex(
             AttributeError,
-            "can't set attribute 'spec'",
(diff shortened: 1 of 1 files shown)


exir/backend/test/test_partitioner.py
```diff
@@ -106,7 +106,7 @@ def partition(
 
         with self.assertRaisesRegex(
             AttributeError,
-            "can't set attribute 'spec'",
+            "(can't set attribute|has no setter)",
```
JacobSzwejbka: I believe we still test earlier python versions in oss (or we should be if not) can you make the regex pass on either string

JacobSzwejbka (changes requested):

PR #17403, opened 2026-02-11
add cuda backend to backend test infra

Gasoonjia:
Integrate cuda backend into backend test infra; skipped the unsupported test for now

### DIFF
--- a/backends/test/suite/flows/cuda.py
+++ b/backends/test/suite/flows/cuda.py
@@ -1,0 +1,22 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from executorch.backends.cuda.test.tester import CudaTester
+from executorch.backends.test.suite.flow import TestFlow
+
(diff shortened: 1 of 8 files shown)


backends/test/suite/flows/cuda.py
```diff
@@ -0,0 +1,62 @@
...
+    The test harness serialize stage has been updated to support loading external
+    data via the data_map_buffer parameter of _load_for_executorch_from_buffer.
+    """
+    # Tests to skip due to known failures in CUDA backend
+    # TODO: Fix these tests and remove from skip list
+    skip_patterns = [
+        # test_mean_output_dtype: float64 output dtype causes crash in AOTI compiled code
+        "test_mean_output_dtype",
+        # Adaptive pooling ops - not fully supported
+        "test_adaptive_avgpool1d",
```
GregoryComer: What do you think about leaving this tests enabled? The jobs aren't land blocking and don't currently show up as red in CI - they are more informational about what is supported on each backend. If we leave them enabled, then they will give that signal. Ideally, each backend can also just not partition unsupported ops.

GregoryComer (approved):
Thanks for adding this!

PR #17405, opened 2026-02-11
Fix Cortex-M MobileNetV2 dialect test expected op counts

rascani:
### Summary

The test expected 1 dequantize/quantize_per_tensor op each after transforms, but the graph actually produces 2. The extra pair is a requantization boundary between avg_pool2d and linear caused by an unannotated clone_dim_order node blocking FoldAndAnnotateQParamsPass from folding through the view_copy -> clone_dim_order -> linear chain.

This PR was authored with Claude.

### Test plan
```
pytest "backends/cortex_m/test/models/test_mobilenet_v2.py::test_dialect_mv2[mobilenet_v2]"
```


psiddh (approved):

rascani:
Failures unrelated.

PR #16540, opened 2026-01-12
NXP backend: added Squeeze support

novak-vaclav:
### Summary

adds support for "Squeeze" operator

### Test plan

tests can be manually run using `pytest -c /dev/null backends/nxp/tests/`

cc @robert-kalmar @MartinPavella

PR #17412, opened 2026-02-12
NXP backend: Add support for `aten.avg_pool1d.default`.

MartinPavella:
### Summary
During the conversion from aten dialect to edge dialect, ExecuTorch replaces the `avg_pool1d` with `avg_pool_2d` and adds an `aten,unsqueeze` before, to introduce the 4th dimension of size 1, and an `aten.squeeze` is added after to remove it. The `avg_pool1d` is then supported by the existing `AvgPool2DConverter`.

### Test plan
Unit-tests provided.


MartinPavella:
The changes are based on another PR (https://github.com/pytorch/executorch/pull/16540) so only the last commit is relevant.

MartinPavella:

novak-vaclav:
Looks good to me 👍

backends/nxp/edge_passes/move_auxiliary_operator_into_separate_qdq_cluster_pass.py
```diff
@@ -106,7 +109,11 @@ class MoveLeadingAuxiliaryOperatorIntoSeparateQDQClusterPass(NeutronEdgePass):
         ],
         ViewCopy: [Clone, CloneDimOrder],
         Conv: [
-            ViewCopy,  # For 1D conv.
+            ViewCopy,  # For 1D conv
+        ],
+        AvgPool2D: [
+            ViewCopy,  # For 1D AvgPool
```
novak-vaclav: Nit: Without prior knowledge of what this PR is about, I would find this comment confusing as why there is something about 1D AvgPool in a relation to AvgPool2D. I would modify the comment like this `For 1D AvgPool, since 1D version is converted to 2D to enable delegation`

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
```diff
@@ -219,3 +230,59 @@ def test_avg_pool_2d_quant_conversion__padded(mocker, use_qat):
...
+
+        self.avg_pool = torch.nn.AvgPool1d(
+            kernel_size=3,
+        )
+
+    def forward(self, x):
+        return self.avg_pool(x)
+
+
+def test_from_avg_pool_1d(mocker):
```
novak-vaclav, line 247: Nit: I'm a bit confused about why the 1D variant test is placed in `test_avg_pool2d_converter`, even though I understand the reasoning. Maybe adding a comment would help, or maybe find a way to move the test to edge_passes.

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
```diff
@@ -219,3 +230,59 @@ def test_avg_pool_2d_quant_conversion__padded(mocker, use_qat):
...
+    extended_shape = (1, 3, 1, 12)
+
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+    delegated_ep = to_quantized_edge_program(
+        model, input_shape, use_neutron_for_format_conversion=False
+    ).exported_program()
+
+    # Make sure the `avg_pool` was delegated.
+    assert graph_contains_any_of_ops(delegated_ep.graph, [ExecutorchDelegateCall])
+    assert not graph_contains_any_of_ops(delegated_ep.graph, [AvgPool2D])
```
novak-vaclav, line 267: Nit: I would check that there are also no `AvgPool1D` operators.

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
```diff
@@ -219,3 +230,59 @@ def test_avg_pool_2d_quant_conversion__padded(mocker, use_qat):
...
+    input_shape = (
+        1,
+        3,
+        12,
+    )  # Don't use multiples of `num_macs` so the `view_copy` nodes will NOT be delegated.
+    extended_shape = (1, 3, 1, 12)
+
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+    delegated_ep = to_quantized_edge_program(
+        model, input_shape, use_neutron_for_format_conversion=False
```
novak-vaclav, line 262: Out of curiosity, why is `use_neutron_for_format_conversion=False`? 😄

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
MartinPavella: There are 2 edge passes relevant to AvgPool1D but only indirectly. I think it makes the most sense to keep these tests in this module, but I will add the comment as you suggested.

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
MartinPavella: Good point 👍🏻 

Edit:
As there is no `aten.avg_pool1d.default` in the edge dialect, I added a check for AvgPool1D by string (in case it is ever added in the future).

backends/nxp/tests/ir/converter/node_converter/test_avg_pool2d_converter.py
MartinPavella: If it was `True` (default), our backend would insert `Transpose` operations to convert between input/output channels last <-> channels first on Neutron. But whether these `Transpose` ops are inserted depends on the shapes (due to Neutron `Transpose` restrictions). This often results in cases where for example the input `Transpose` is inserted, but the output isn't (or vice versa, or any other combination). Taking this into account in the tests makes them unnecessarily complicated. So I set `use_neutron_for_format_conversion=False`, and no `Transpose` ops are inserted, and we just always have to do the `ToChannelsLastPreProcess()` and stuff.

novak-vaclav: Thank you for the explanation 😄

novak-vaclav:
Thank you for incorporating my comments into the code @MartinPavella 
Everything looks good to me 👍

novak-vaclav (approved):

roman-janik-nxp:
@MartinPavella please rebase on main and check if internal build passes https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH80-10. Otherwise looks good.

roman-janik-nxp (approved):

PR #17416, opened 2026-02-12
Neutron Backend: Add model test (Cifar10) with eIQ NSYS

robert-kalmar:
### Summary
Add end-to-end model test with small CNN trained on Cifar10 dataset with eIQ NSYS (Neutron functional simulator).

### Test plan
Introducing end-to-end model test

cc @JakeStevens @digantdesai


backends/nxp/tests_models/test_cifarnet.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024-2026 NXP
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
```
MartinPavella, line 4: Nit: Missing blank line.

backends/nxp/tests_models/utils.py
```diff
@@ -0,0 +1,179 @@
...
+        m(*input_data)
+        input_data.clear()
+
+    exir_program_aten_quant = convert_pt2e(m)
+
+    # To ATen
+    core_aten_ep = _to_core_aten(exir_program_aten_quant, example_input, None, verbose=True)
+
+    partitioners = (
+        [NeutronPartitioner(generate_neutron_compile_spec("imxrt700", "SDK_25_12"),
```
MartinPavella: Nit: 
To avoid duplication, can we extract the target and the Neutron flavor from the
```python
neutron_target_spec = NeutronTargetSpec(
        target="imxrt700", neutron_converter_flavor="SDK_25_12"
    )
```
on line 65?

MartinPavella (approved):

backends/nxp/tests_models/utils.py
robert-kalmar: The `neutron_target_spec` does not stores the `target`, and the `neutron_converter_flavor`, but it translates it to neutron converter's params. 
But indeed the information provided to  NeutronPartitioner in `compile_spec` and `neutron_target_spec` is the same, hence those can be united.

PR #17419, opened 2026-02-12
Re-enable death test matchers on Windows

### LINKED ISSUE #16281: Support message matching in death tests

In #16162, we added a variety of null pointer asserts and a corresponding set of death tests to ensure they are caught. An earlier revision of the the PR included message matchers for those assertions, but these all failed on Windows. The reason why was not immediately clear, but the default matcher for string literals is regex-based and there is different gtest regex parsing on [posix systems than Windows & Mac](https://github.com/google/googletest/blob/main/googletest/include/gtest/gtest-death-test.h#L106-L117).

It would be preferable to bring back the message matching and investigate why Windows failed. Having message matching would allow us to ensure that the right assert is firing.

### Alternatives

_No response_

### Additional context

_No response_

### RFC (Optional)

_No response_

rascani:
### Summary

The Windows branch was added because some matchers used POSIX-specific regex features (bracket syntax [0-9]) that GTest doesn't support on Windows. With compatible matchers in place, we can re-enable matching to get better test diagnostics.

Fixes #16281

This PR was authored with Claude.

### Test plan
CI


lucylq (approved):
Thanks!

PR #17421, opened 2026-02-12
1/x: add LoadBackendOptionsMap for backend options routing

metascroy:
Summary:
This diff introduces `LoadBackendOptionsMap`, a container for routing backend-specific options to delegates at model load time. This enables users to configure multiple backends with options that are automatically routed to the appropriate delegate during `Module::load()`.

Key components:
- `LoadBackendOptionsMap` class in `backend_options_map.h` that maps backend IDs to their options
- Template `set_options(const Builder&)` method for builder pattern integration
- Added const overload to `BackendOptions::view()` to support const-correct builders
- `ExampleBackendOptions` reference implementation showing the recommended pattern for backend developers

The builder pattern enables fluent API usage:
```cpp
ExampleBackendOptions opts;
opts.setNumThreads(4).setEnableOptimization(true);

LoadBackendOptionsMap map;
map.set_options(opts);  // Uses builder's backend_id()
```

Reviewed By: larryliu0820

Differential Revision: D92358596

### DIFF
--- a/runtime/backend/backend_options_map.h
+++ b/runtime/backend/backend_options_map.h
@@ -1,0 +1,188 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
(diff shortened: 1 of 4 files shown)


larryliu0820 (approved):

PR #17423, opened 2026-02-12
Fix NHWC depthwise conv weight layout and dimension indexing for HiFi operators

JakeStevens:
Summary:
The NHWC depthwise conv HiFi operators had two bugs that caused incorrect inference results on the DSP:

  1. **Weight layout mismatch**: The operators passed weight tensors directly to NNLib's `xa_nn_conv2d_depthwise_per_chan_sym8sxasym8s` with `inp_data_format=0` (NHWC), but the weight data from the ExecuTorch graph is in `[OC, KH, KW, IC/G]` format (output-channel first), while NNLib's NHWC path expects `[KH, KW, OC]` (spatial-first / HWC). This caused every depthwise conv to compute with scrambled kernel values. Fixed by rearranging the weight buffer from `[OC, KH*KW]` to `[KH*KW, OC]` before calling NNLib. The NCHW operators were unaffected because NNLib's NCHW path (`inp_data_format=1`) expects CHW kernel format which matches the graph's weight layout.

  2. **NCHW dimension indexing in NHWC operators**: The operators extracted tensor dimensions using NCHW conventions (e.g., `input_channels = input.size(1)`) despite operating on NHWC-layout tensors. For 3D conv1d inputs `[N, W, C]`, `input.size(1)` is the spatial width, not channels. Fixed by adding a `conv1d` guard and using NHWC indexing: `input_channels = conv1d ? input.size(2) : input.size(3)`.

Reviewed By: mcremon-meta

Differential Revision: D93125592

cc @mcremon-meta

### DIFF
--- a/backends/cadence/hifi/operators/op_quantized_conv2d_nhwc_depthwise_asym8sxsym8s_asym8s_per_tensor_out.cpp
+++ b/backends/cadence/hifi/operators/op_quantized_conv2d_nhwc_depthwise_asym8sxsym8s_asym8s_per_tensor_out.cpp
@@ -48,14 +48,15 @@ void xa_opt_quantized_conv2d_nhwc_depthwise_asym8sxsym8s_asym8s(
   WORD32* __restrict__ p_bias =
       (WORD32* __restrict__)bias.const_data_ptr<int32_t>();
 
-  WORD32 input_height = conv1d ? 1 : input.size(2);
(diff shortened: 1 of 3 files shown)


mcremon-meta (approved):

mcremon-meta:
cc @dijopaul and @cad-audio

PR #17431, opened 2026-02-12
Enable Voxtral Realtime on XNNPACK (CPU)

mergennachin:
Adds Mistral's Voxtral-Mini-4B-Realtime-2602 (~4B parameter streaming
speech-to-text model) to ExecuTorch with XNNPACK backend support.

Phase 1: Self-contained eager model (model.py) with direct Mistral
checkpoint loading, multi-method export (audio_encoder, text_decoder,
token_embedding) to a single .pte, and TorchAO quantization (4bit blockwise, 8bit dynamic activation for linear layer and 8bit weight per-channel for embeddings).

Phase 2: C++ runner for offline transcription. Loads preprocessor.pte
for mel spectrogram computation, runs audio encoding, then autoregressive
decoding with element-wise audio+text embedding fusion.

Phase 3: Streaming support (follow-up PR: https://github.com/pytorch/executorch/pull/17440)

Phase 4: Enable on CUDA and Metal (follow-up PR)

Example output (8da4w quantized, 30s LibriSpeech audio):

```
$ cmake-out/examples/models/voxtral_realtime/voxtral_realtime_runner \
    --model_path voxtral_realtime.pte \
    --tokenizer_path tekken.json \
    --preprocessor_path preprocessor.pte \
    --audio_path output.wav

Mr. Quilter is the apostle of the middle classes, and we are glad to
welcome his gospel. Nor is Mr. Quilter's manner less interesting than
his matter. He tells us that at this festive season of the year, with
Christmas and roast beef looming before us, similes drawn from eating
and its results occur most readily to the mind. He has grave doubts
whether Sir Frederick Layton's work is really Greek after all, and...

Generated 392 tokens in 44s (~8.8 tok/s) on M1 Macbook
```

Test Plan: https://github.com/pytorch/executorch/actions/runs/21986674876/job/63522558208?pr=17431


extension/llm/export/quantize.py
```diff
@@ -0,0 +1,163 @@
+"""TorchAO source-transform quantization for ExecuTorch LLM export.
```
manuelcandales, line 1: Definitely agree with this kind of refactor. However, I am wondering if this is the right place? Why under llm? 4/8 bit quantization can be applied also outside llms.
There is also a quantizer_lib.py under this same directory. So, this might be confusing to a new contributor, why two quantization files? what is the purpose of quantizer_lib.py vs quantize.py

examples/models/voxtral_realtime/export_voxtral_rt.py
```diff
@@ -0,0 +1,296 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
manuelcandales, line 1: There is a PR adding this model to transformers (still open): https://github.com/huggingface/transformers/pull/43769

Do we plan to move this to optimum-executorch once that PR is landed?

mergennachin: @manuelcandales 

Yeah, when I first looked at it, the model wasn't in transformers a few days ago. FWIW, the vLLM has its own copy of the implementation in their repo, similar to what I'm doing. So, implementing directly seemed the most straightforward.

> Do we plan to move this to optimum-executorch once that PR is landed?

Maybe. Once they land in transformers, we might. There are a few variables such as upgrading the transformers pin in ET -- they recently had a major 5.0 update, so I assume there will be a few breakages that need foxing. Also, I'm fine keeping as it is, if it works already.

extension/llm/export/quantize.py
mergennachin: Yeah, I can mark the API as experimental, so that we can move things around (promote from llm to more generic location). 

At some point, I want to combine quantizer_lib.py and quantize.py but not in this PR. They're doing something very similar. quantizer_lib works with etLLM, there are some legacy stuff such as doing hybrid of PT2E quantization and torchao.

examples/models/voxtral_realtime/export_voxtral_rt.py
patrickvonplaten: (Voxtral author here) BTW the transformers implementation only supports "offline" streaming for now for which the whole audio file is encoded in one go. The arch and forward passing logic is def still the same, but I think what we're really interested in is the "true" online / realtime use case that we implemented via the realtime API inside vLLM (see: https://docs.vllm.ai/en/latest/examples/online_serving/openai_realtime_client/?h=realtime#openai-realtime-client)

examples/models/voxtral_realtime/main.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  voxtral_realtime::TranscribeConfig config;
+  config.temperature = static_cast<float>(FLAGS_temperature);
+  config.max_new_tokens = FLAGS_max_new_tokens;
+
+  stats.num_prompt_tokens = 0;
+  bool first_token = true;
+
+  int num_generated = runner.transcribe(
+      audio_data.data(),
```
patrickvonplaten, line 72: any chance that there is a way to feed in audio data iteratively via some kind of generator / iterator?

examples/models/voxtral_realtime/model.md
```diff
@@ -0,0 +1,218 @@
...
+
+```
+scale = 1 + Sequential(Linear(3072->32), GELU, Linear(32->3072))(t_cond)
+ffn_input = rms_norm(x) * scale
+```
+
+The `t_cond` is a sinusoidal embedding of `n_delay_tokens` (default 6 = 480ms),
+precomputed once and passed to each decoder layer as a constant.
+
+### Differences from original Voxtral (non-realtime)
```
patrickvonplaten, line 51: nice summary

examples/models/voxtral_realtime/voxtral_realtime_runner.cpp
```diff
@@ -0,0 +1,227 @@
...
+  return from_blob(
+      mel_ref.mutable_data_ptr<float>(),
+      {static_cast<int>(mel_ref.size(0)),
+       static_cast<int>(mel_ref.size(1)),
+       static_cast<int>(mel_ref.size(2))},
+      ::executorch::aten::ScalarType::Float);
+}
+
+int VoxtralRealtimeRunner::transcribe(
+    const float* audio_data,
```
patrickvonplaten, line 113: any chance to also provide a ::realtime interface?

examples/models/voxtral_realtime/main.cpp
mergennachin: Yeap, followup PR coming soon

manuelcandales (approved):

examples/models/voxtral_realtime/main.cpp
mergennachin: Here's the true streaming mode: 

https://github.com/pytorch/executorch/pull/17440

Issue #16381, opened 2025-12-24
Numeric Discrepancy detector for ExecuTorch Qualcomm backend.


Gasoonjia:
### 🚀 The feature, motivation and pitch

A Numeric Discrepancy Detector for Qualcomm backend is important for understanding the numerical issue happened in Qualcomm backend in operator-level.

Gasoonjia:
Shared proposal with Quallcomm team (@winskuo-quic and others) and we've aligned on both target the impl details. Starting et side impl this week.

Target for e2e pipeline work is q1.

Gasoonjia:
@winskuo-quic has stamped all PRs from ET side and confirm they should support Qualcomm requirement. Will land them asap.

Gasoonjia:
ExecuTorch updates have been done and confirmed by @winskuo-quic .

PR #17433, opened 2026-02-13
[devtools] Add preprocessing support to NumericalComparatorBase

Gasoonjia:
* #17459
* #17434
* __->__ #17433

This change extends NumericalComparatorBase to support custom preprocessing before numeric gap calculation, enabling backends like Qualcomm to apply necessary tensor transformations (e.g., dequantization, layout conversion) before comparison.

Key changes:
- Extended NumericalComparatorBase with:
  - `__init__(inspector)` to store optional Inspector reference for accessing graph metadata
  - `preprocessing(mapping)` method that can be overridden for custom tensor transformations (default: identity)
  - `element_compare(a, b)` abstract method for element-level tensor comparison
  - `compare(mapping, ...)` method that orchestrates the full pipeline: preprocessing → element-wise compare → aggregate to DataFrame

This is part of the operator-level numeric discrepancy detector project for ExecuTorch Qualcomm backend (https://github.com/pytorch/executorch/issues/16381).

Design doc: https://docs.google.com/document/d/1GaCHiy9InytOsUrl2BKEgOiP1iKTfpCVdWg6QDh0N2E/edit?tab=t.0#heading=h.fcrpnrtb6cud

Differential Revision: [D93169813](https://our.internmc.facebook.com/intern/diff/D93169813/)


devtools/inspector/tests/inspector_test.py
```diff
@@ -795,6 +795,235 @@ def compare(self, a, b):
...
+            )
+            inspector_instance._get_runtime_intermediate_outputs_and_op_names = (
+                lambda: (runtime_intermediate_outputs, runtime_debug_handle_to_op_name)
+            )
+
+            # Create custom comparator with 2x scaling
+            custom_comparator = ScalingComparator(scale_factor=2.0)
+
+            # Test with custom comparator
+            df = inspector_instance.calculate_numeric_gap(distance=custom_comparator)
```
winskuo-quic: Hi @Gasoonjia,
Thank you so much for the PR to support the preprocessing.
Sorry I am now having limited access to laptop and may not be able to try out this PR until next week.

For this unit test, could you add a case where we will be performing multiple comparisons? For example, perform custom_operator and then perform a snr comparator immediately, and both the comparator requires that same preprocessing function. I believe this would be a UT that can closely align with Qualcomm's use case, which also helps demonstrate how inspector expects users to handle the intermediate tensor map between comparison.
Thanks

larryliu0820 (approved):

winskuo-quic:
Hi @Gasoonjia,
Thanks again for the PR.
I have left a comment above related to unit test. Besides, this PR looks good to me.
Thanks

Gasoonjia:
thansk for comment @winskuo-quic 
I've updated the existing test to support two consecutive numeric comparator with same preprocessor. 
Plz let me know if there's any concern, or i will land them in recent days.

PR #17434, opened 2026-02-13
[devtools] Auto-record after-transform graph in ETRecord

Gasoonjia:
* #17459
* __->__ #17434
* #17433

When `to_edge_transform_and_lower` is called with `generate_etrecord=True` and custom `transform_passes` are applied, the after-transform graph is now automatically recorded in the ETRecord's `graph_map` under the key `"edge_after_transform"`. This enables backends like Qualcomm to use the post-custom-transform graph as the golden reference for numeric gap calculation, while have zero impact for regular exportation.

This is part of the operator-level numeric discrepancy detector project for ExecuTorch Qualcomm backend (https://github.com/pytorch/executorch/issues/16381).

Design doc: https://docs.google.com/document/d/1GaCHiy9InytOsUrl2BKEgOiP1iKTfpCVdWg6QDh0N2E/edit?tab=t.0#heading=h.fcrpnrtb6cud

Differential Revision: [D93176563](https://our.internmc.facebook.com/intern/diff/D93176563/)


devtools/etrecord/tests/etrecord_test.py
```diff
@@ -1819,3 +1819,122 @@ def test_multi_method_etrecord_generation(self):
...
+        )
+
+        # Verify that ETRecord was generated
+        self.assertIsNotNone(edge_manager._etrecord)
+        etrecord = edge_manager._etrecord
+
+        # Verify that 'edge_after_transform' is NOT in graph_map
+        if etrecord.graph_map is not None:
+            self.assertNotIn(
+                "edge_after_transform/forward",
```
winskuo-quic: Since this param will be used across UT, Qualcomm, and potentially other backend owners in the future, do you think we should make this a param and place it under `executorch.exir.debug_handle_utils`?

winskuo-quic (approved):
LGTM.
Thanks for having this PR. This allows us to retrieve valuable information for graphs with edge passes while not break the BC.

devtools/etrecord/tests/etrecord_test.py
winskuo-quic, line 1937: Just curious. 
I am studying the logic of the graph_map naming: https://github.com/pytorch/executorch/blob/43ee3f99e8957fcb633c36f87e45c546a87cf277/devtools/etrecord/_etrecord.py#L651.
This hardcoded `forward` is appened because an edge_after_transform could possibly consist of multimethods?

Gasoonjia: yes your are right. That's placeholder for multimethods supported.
Now we only support single forward name but in the future we will support multiple methods in the same program.

larryliu0820 (approved):

larryliu0820 (approved):

PR #17435, opened 2026-02-13
Enable building for metal with macOS SDK < 15

mattjcly:
### Summary
Fixes compilation errors on macOS w/ SDK < 15 due to non-existent properties:
```
executorch/backends/apple/metal/runtime/shims/et_metal.mm:239:21: error: property 'mathMode' not found on object of type 'MTLCompileOptions *'
            options.mathMode = MTLMathModeSafe;
                    ^
executorch/backends/apple/metal/runtime/shims/et_metal.mm:239:32: error: use of undeclared identifier 'MTLMathModeSafe'
            options.mathMode = MTLMathModeSafe;
                               ^
executorch/backends/apple/metal/runtime/shims/et_metal.mm:240:21: error: property 'mathFloatingPointFunctions' not found on object of type 'MTLCompileOptions *'
            options.mathFloatingPointFunctions = MTLMathFloatingPointFunctionsPrecise;
                    ^
executorch/backends/apple/metal/runtime/shims/et_metal.mm:240:50: error: use of undeclared identifier 'MTLMathFloatingPointFunctionsPrecise'
            options.mathFloatingPointFunctions = MTLMathFloatingPointFunctionsPrecise;
```
Keep runtime check

### Test plan
 Confirmed working build on machine with:
```
-> % xcodebuild -version
Xcode 15.2
Build version 15C500b
-> % xcrun --sdk macosx --show-sdk-version
14.2
```
That did not work before.

Also confirmed to still work on 26.1:
```
-> % xcodebuild -version
Xcode 26.1.1
Build version 17B100
-> % xcrun --sdk macosx --show-sdk-version
26.1
```
Including verification that inside of if/else def was triggered with a `#warning` (since removed) to see:
```
executorch/backends/apple/metal/runtime/shims/et_metal.mm:249:2: warning: "Compiling with MTLMathModeSafe and MTLMathFloatingPointFunctionsPrecise for improved precision (requires macOS 15/iOS 18 or later)" [-W#warnings]
  249 | #warning "Compiling with MTLMathModeSafe and MTLMathFloatingPointFunctionsPrecise for improved precision (requires macOS 15/iOS 18 or later)"
      |  ^
1 warning generated.
```
Confirmed Parakeet inference to still work on both machines.

### DIFF
--- a/backends/apple/metal/runtime/shims/et_metal.mm
+++ b/backends/apple/metal/runtime/shims/et_metal.mm
@@ -19,12 +19,32 @@
 #include <exception>
 
 #if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 150000) || \
-  (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 180000) || \
-  (defined(__TV_OS_VERSION_MAX_ALLOWED) && __TV_OS_VERSION_MAX_ALLOWED >= 180000) || \
-  (defined(__WATCH_OS_VERSION_MAX_ALLOWED) && __WATCH_OS_VERSION_MAX_ALLOWED >= 110000)
-#define ET_METAL_HAS_MTL_MATH_COMPILE_OPTIONS 1
+    (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 180000) || \
+    (defined(__TV_OS_VERSION_MAX_ALLOWED) && __TV_OS_VERSION_MAX_ALLOWED >= 180000) || \
+    (defined(__WATCH_OS_VERSION_MAX_ALLOWED) && __WATCH_OS_VERSION_MAX_ALLOWED >= 110000)
+#define ET_METAL_SDK_HAS_MTL_MATH_COMPILE_OPTIONS 1
 #else
-#define ET_METAL_HAS_MTL_MATH_COMPILE_OPTIONS 0
+#define ET_METAL_SDK_HAS_MTL_MATH_COMPILE_OPTIONS 0
+#endif
+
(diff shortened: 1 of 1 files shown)


mattjcly:
Only failing test that seems potentially relevant is [pull / unittest-editable / macos / macos-job](https://hud.pytorch.org/pr/pytorch/executorch/17435#63536908344) ... will try to understand exactly what this is testing and if this is a true failure/breakage or not

mergennachin:
@mattjcly Let me rerun the failing test again. It might some tolerance issue

malfet:
One more example where we could have shared code better between PT and ET, as we've solved this problem a while back :)

backends/apple/metal/runtime/shims/et_metal.mm
```diff
@@ -18,6 +18,15 @@
 #include <optional>
 #include <exception>
 
+#if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 150000) || \
```
malfet, line 21: IMO this is wrong, because it will require one to have two binaries, one for older and another one for newer SDK

malfet (commented):
IMO it'll be better to do it other way around:
define new enums if you compiling with the older SDK, and then indeed rely on `@available`...
See https://github.com/pytorch/pytorch/blob/4504c3dcee3c02886fae3340a8ee268717c6cb32/aten/src/ATen/native/mps/MPSGraphSequoiaOps.h as an example of accomplishing that

mattjcly:
> IMO it'll be better to do it other way around: define new enums if you compiling with the older SDK, and then indeed rely on `@available`... See https://github.com/pytorch/pytorch/blob/4504c3dcee3c02886fae3340a8ee268717c6cb32/aten/src/ATen/native/mps/MPSGraphSequoiaOps.h as an example of accomplishing that

Very smart! Thanks for taking a look. Just made the change and re-confirmed can still build and run parakeet metal on both SDK 14.2 and 26.1

shoumikhin:
@manuelcandales something known?

manuelcandales (approved):

PR #17440, opened 2026-02-13
Enable true streaming for Voxtral Realtime model on XNNPACK

mergennachin:
Follow-up to https://github.com/pytorch/executorch/pull/17431

Enable Voxtral Realtime streaming on XNNPACK
Adds true streaming support for the Voxtral-Mini-4B-Realtime-2602 model,
enabling real-time transcription from live audio input.

Key components:
- StreamingAudioEncoderExport: encoder with KV-cached attention and conv
  state for chunk-by-chunk processing (8 mel frames = 80ms per step)
- StreamingSession C++ API (feed_audio/flush) for incremental audio input
- Streaming mel preprocessor (WhisperAudioProcessor with --streaming flag)
- STFT overlap windowing (320 left + 40 right) for correct mel at chunk
  boundaries
- Per-component quantization (--qlinear-encoder / --qlinear / --qembedding)

Test Plan: 

https://github.com/pytorch/executorch/actions/runs/21993469202/job/63546903665?pr=17440

### DIFF
--- a/examples/models/voxtral_realtime/README.md
+++ b/examples/models/voxtral_realtime/README.md
@@ -31,6 +31,16 @@ This produces `preprocessor.pte` which takes a 1-D waveform tensor
 `(num_samples,)` and outputs a mel spectrogram `(1, 128, T_mel)`. The
 `--max_audio_len 300` flag supports audio up to 5 minutes.
 
+For streaming, add `--streaming` to skip the 30-second chunk padding so
+that 1280 samples (80ms) produces exactly 8 mel frames:
+
+```bash
+python -m executorch.extension.audio.mel_spectrogram \
+    --feature_size 128 \
+    --streaming \
+    --output_file ./voxtral_rt_exports/preprocessor.pte
+```
+
 ## Export
 
 Export produces a single `.pte` file with three methods:
@@ -46,12 +56,25 @@ python export_voxtral_rt.py \
     --model-path ~/models/Voxtral-Mini-4B-Realtime-2602 \
     --backend xnnpack \
     --output-dir ./voxtral_rt_exports \
+    --qlinear-encoder 8da4w \
     --qlinear 8da4w \
     --qembedding 8w
 ```
 
-This exports with XNNPACK backend acceleration and 8-bit dynamic activation /
-4-bit weight linear quantization + 8-bit embedding quantization.
+For streaming, add `--streaming` to export the encoder with KV caches for
+incremental processing. This replaces `audio_encoder` with
+`encode_audio_chunk` which processes 8 mel frames at a time:
+
+```bash
+python export_voxtral_rt.py \
+    --model-path ~/models/Voxtral-Mini-4B-Realtime-2602 \
+    --backend xnnpack \
+    --streaming \
+    --output-dir ./voxtral_rt_exports \
+    --qlinear-encoder 8da4w \
+    --qlinear 8da4w \
+    --qembedding 8w
+```
 
 ### Options
 
@@ -62,9 +85,13 @@ This exports with XNNPACK backend acceleration and 8-bit dynamic activation /
 | `--output-dir` | `./voxtral_rt_exports` | Output directory |
 | `--max-seq-len` | `4096` | KV cache length |
 | `--delay-tokens` | `6` | Transcription delay in tokens (6 = 480ms) |
-| `--qlinear` | (none) | Linear layer quantization (`4w`, `8w`, `8da4w`, `8da8w`) |
-| `--qlinear-group-size` | `32` | Group size for linear quantization |
+| `--qlinear` | (none) | Decoder linear layer quantization (`4w`, `8w`, `8da4w`, `8da8w`) |
+| `--qlinear-group-size` | `32` | Group size for decoder linear quantization |
+| `--qlinear-encoder` | (none) | Encoder linear layer quantization (`4w`, `8w`, `8da4w`, `8da8w`) |
+| `--qlinear-encoder-group-size` | `32` | Group size for encoder linear quantization |
 | `--qembedding` | (none) | Embedding layer quantization (`8w`) |
+| `--streaming` | off | Export streaming encoder with KV cache |
+| `--max-enc-len` | `750` | Max encoder KV cache length (~60s audio, streaming only) |
 
 ## Build
 
@@ -91,6 +118,21 @@ cmake-out/examples/models/voxtral_realtime/voxtral_realtime_runner \
     --audio_path input.wav
 ```
 
+For streaming, add `--streaming`. The runner feeds audio in 200ms chunks
+(simulating live microphone input), computing mel and running the
+encoder+decoder per 80ms step. The `StreamingSession` C++ API
+(`feed_audio` / `flush`) can be used directly for integration with live
+audio sources.
+
+```bash
+cmake-out/examples/models/voxtral_realtime/voxtral_realtime_runner \
+    --model_path voxtral_rt_exports/model.pte \
+    --tokenizer_path ~/models/Voxtral-Mini-4B-Realtime-2602/tekken.json \
+    --preprocessor_path voxtral_rt_exports/preprocessor.pte \
+    --audio_path input.wav \
+    --streaming
+```
+
 | Flag | Default | Description |
 |------|---------|-------------|
 | `--model_path` | `model.pte` | Path to exported model |
@@ -99,6 +141,7 @@ cmake-out/examples/models/voxtral_realtime/voxtral_realtime_runner \
 | `--audio_path` | (required) | Path to 16kHz mono WAV file |
 | `--temperature` | `0.0` | Sampling temperature (0 = greedy) |
 | `--max_new_tokens` | `500` | Maximum tokens to generate |
+| `--streaming` | off | Use streaming transcription |
 
 ### Example output
 
--- a/examples/models/voxtral_realtime/README.md
+++ b/examples/models/voxtral_realtime/README.md
@@ -91,7 +91,7 @@ python export_voxtral_rt.py \
 | `--qlinear-encoder-group-size` | `32` | Group size for encoder linear quantization |
 | `--qembedding` | (none) | Embedding layer quantization (`8w`) |
 | `--streaming` | off | Export streaming encoder with KV cache |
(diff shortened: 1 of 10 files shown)


mergennachin:
cc @patrickvonplaten

examples/models/voxtral_realtime/main.cpp
```diff
@@ -68,19 +70,39 @@ int main(int argc, char** argv) {
...
+    auto session = runner.create_streaming_session(config, token_cb);
+
+    // Feed audio in 200ms chunks (simulates live microphone input).
+    const int64_t chunk_size = 3200;
+    for (int64_t offset = 0; offset < static_cast<int64_t>(audio_data.size());
+         offset += chunk_size) {
+      int64_t n = std::min(
+          chunk_size, static_cast<int64_t>(audio_data.size()) - offset);
+      session->feed_audio(audio_data.data() + offset, n);
+    }
```
mergennachin, line 97: @patrickvonplaten 

This is "simulating" live microphone by chunking the audio and feeding into the stream.

The API supports being able to continuously read input stream and transcribe. It does the "right thing" by looking at a moving window of chunks.

The live microphone feature can be done in the application layer by end customers and demo builder (which is not super interesting in the context of model enablement)

examples/models/voxtral_realtime/export_voxtral_rt.py
```diff
@@ -159,6 +205,99 @@ def export_all(model, max_seq_len):
...
+    # mel_skip: number of overlap frames to skip at the start
+    hop_length = 160
+    n_fft = 400
+    sample_rate = 16000
+    frame_rate = 12.5
+    step_samples = int(sample_rate / frame_rate)
+    stft_left_overlap = ((n_fft // 2 + hop_length - 1) // hop_length) * hop_length
+    mel_skip_frames = stft_left_overlap // hop_length
+    chunk_mel_len = 8
+    stft_right_lookahead = (
```
patrickvonplaten, line 273: this should correspond to just like 2.5ms right?

examples/models/voxtral_realtime/export_voxtral_rt.py
```diff
@@ -236,20 +375,43 @@ def main():
...
+    parser.add_argument(
+        "--streaming",
+        action="store_true",
+        help="Export streaming encoder (encode_audio_chunk) instead of offline encoder.",
+    )
+    parser.add_argument(
+        "--max-enc-len",
+        type=int,
+        default=750,
+        help="Max encoder KV cache length for streaming (default: 750, ~60s audio).",
```
patrickvonplaten:
```suggestion
        help="Max encoder KV cache length for streaming (default: 750, ~15s audio).",
```

Note that 1 LLM token corresponds to 4 audio encoder tokens which corresponds to 80ms
=> 750ms corresponds to 750 / 4 * 0.08 s => 15s

examples/models/voxtral_realtime/main.cpp
```diff
@@ -68,19 +70,39 @@ int main(int argc, char** argv) {
...
+  };
+
+  int num_generated;
+  if (FLAGS_streaming) {
+    ET_CHECK_MSG(
+        runner.is_streaming(),
+        "Model was not exported with --streaming. Re-export with --streaming flag.");
+    auto session = runner.create_streaming_session(config, token_cb);
+
+    // Feed audio in 200ms chunks (simulates live microphone input).
```
patrickvonplaten: why exactly 200ms chunks? Couldn't it make sense to go a bit lower?

examples/models/voxtral_realtime/model.md
```diff
@@ -146,9 +146,92 @@ The encoder has no KV cache (processes full mel at once in offline mode) and
...
+-> audio_embeds, new_conv1_state, new_conv2_state
+```
+
+### Conv state management
+
+The causal convolutions need left context across chunk boundaries.
+Instead of zero-padding (offline) or recompute-with-overlap (vLLM),
+explicit conv state carries the tail of the previous chunk:
+
+- **Conv1** (kernel=3, stride=1): state = last 2 mel frames from previous
```
patrickvonplaten, line 183: Actually I think we need the last 4 mel frames instead of hte last 2. 

Our look-back is 52.5ms which AFAIU corresponds to

4 log mel frames (each 10ms) and then last frame looks back 12.5 because it has a window size of 25ms => 4 * 10ms + 12.5 = 52.5ms

examples/models/voxtral_realtime/voxtral_realtime_runner.h
```diff
@@ -61,6 +77,23 @@ class VoxtralRealtimeRunner {
...
+  int64_t max_enc_len_ = 750;
+  int64_t enc_dim_ = 1280;
+  int64_t conv1_pad_ = 2;
+  int64_t conv2_pad_ = 2;
+
+  // Raw audio samples per streaming step (sampling_rate / frame_rate = 1280)
+  int64_t step_samples_ = 1280;
+
+  // STFT overlap for streaming mel computation (read from model metadata).
+  int64_t stft_left_overlap_ = 320;
```
patrickvonplaten, line 93:
```suggestion
  int64_t stft_left_overlap_ = 840;
```

think our left overlap should be 52.5ms - see: 

```
    },
    "transcription_delay_ms": 480,
    "streaming_look_ahead_ms": 2.5,
    "streaming_look_back_ms": 52.5,
    "streaming_n_left_pad_tokens": 32,
    "transcription_format": "streaming"
  }
```

in the tekken.json

examples/models/voxtral_realtime/voxtral_realtime_runner.h
```diff
@@ -61,6 +77,23 @@ class VoxtralRealtimeRunner {
...
+  int64_t stft_right_lookahead_ = 40;
+  int64_t mel_skip_frames_ = 2;
```
patrickvonplaten, line 95:
```suggestion
  int64_t mel_skip_frames_ = 4;
```

think that this should actually be 4

examples/models/voxtral_realtime/export_voxtral_rt.py
mergennachin: Yes, that's correct. Fixing

examples/models/voxtral_realtime/export_voxtral_rt.py
mergennachin: Yes, it corresponds to 2.5ms, will add an inline comment

examples/models/voxtral_realtime/main.cpp
mergennachin: 200ms (3200 samples) is arbitrary for simulating microphone input. The internal step size is always 80ms (1280 samples), so can lower to 80ms

examples/models/voxtral_realtime/model.md
mergennachin: The conv state of 2 is correct. It's kernel_size - 1 for both CausalConv1d layers (kernel_size=3). This is the exact amount needed for a causal convolution to produce identical output to processing contiguous audio.

The streaming_look_back_ms = 52.5 in tekken.json serves a different purpose.  The look-back bundles STFT context, conv context, and encoder context into one value.

examples/models/voxtral_realtime/voxtral_realtime_runner.h
mergennachin: Our design handles these separately. Conv state carries the exact conv context (2 frames), and the encoder KV cache provides full attention history. So the left overlap only needs to cover the STFT window edge effect (n_fft/2 = 200 samples), and 320 is the aligned minimum for that.

mergennachin:
@patrickvonplaten 

Here's slight difference between vLLM and ExecuTorch, but they end up producing same result.
                                                                                                                                                                                                                                                                                                                                                                        
vLLM:                                                                                                                                                                                                                                                                                                                                                                  
Each step takes a large audio window — [840 look_back + 1280 step + 40 look_ahead] = 2160 samples — and processes the entire window through the full pipeline:                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                            
2160 raw audio → mel (~13 frames) → conv (~6 post-conv frames) → encoder (all frames, with KV cache) → downsample → audio tokens

The look_back region is re-processed every step. Its mel, conv output, and encoder output are all recomputed alongside the new frames.

Ours: State-based

Each step takes a smaller window — [320 overlap + 1280 step + 40 look_ahead] = 1640 samples — and processes only the new data, using explicit state for context:

1640 raw audio → mel (10 frames) → extract 8 aligned frames
8 mel frames + conv state (2 frames) → conv → 4 encoder frames
4 encoder frames + KV cache (all history) → encoder → downsample → 1 audio token

Why they produce the same result

Mel spectrogram: Both have ≥ n_fft/2 (200 samples) of left context for the step's first mel frame. Our 320 ≥ 200, so the 8 extracted mel frames have identical values to vLLM's corresponding frames. vLLM computes extra mel frames in the look_back region, but they're just for context.

I think you can implement similar approach in vLLM and make it more efficient too.

manuelcandales (approved):

examples/models/voxtral_realtime/model.md
patrickvonplaten: Makes sense!

examples/models/voxtral_realtime/voxtral_realtime_runner.h
patrickvonplaten: Make sense!

examples/models/voxtral_realtime/README.md
```diff
@@ -91,6 +118,21 @@ cmake-out/examples/models/voxtral_realtime/voxtral_realtime_runner \
     --audio_path input.wav
 ```
 
+For streaming, add `--streaming`. The runner feeds audio in 200ms chunks
```
patrickvonplaten, line 121: could it make sense to add an example for a microphone input demo here as well? Think that'd be super powerful :-)

mergennachin: @patrickvonplaten yeah, it makes sense. 

I'm gonna merge this PR, and will try to work on live demo, if time permits.

Also, in the meantime, do you wanna give it a shot as well? Might be good opportunity to test and validate too.

Thanks for reviewing the PR

PR #17452, opened 2026-02-13
Default to using edge_program state dict if quantization data not available

JakeStevens:
Summary:
PR #16901 added logic to replace FakeTensors with actually quantization data.

The docstring says:

```
        :param post_quantization_state_dict: State-dict of the model right after quantization. During partitioning, the
                                              `edge_program` only contains fake tensors without any data. In this case,
                                              this state dict is used instead (if provided). Notice: It may potentially
                                              contain outdated data,
```

with the key part being **if provided**.  However, this is not the actual functionality.

By not falling back, this introduces a regression for depthwise convolutions, which have a "no dynamic" constraint that is triggered by this missing information.

This PR restores the original behavior if quantization data is not provided.

Differential Revision: D93253828

### DIFF
--- a/backends/nxp/backend/edge_program_converter.py
+++ b/backends/nxp/backend/edge_program_converter.py
@@ -208,8 +208,12 @@ class EdgeProgramToIRConverter:
                         result_map[input_spec.arg.name] = param
 
                     else:
-                        # There is no data available.
-                        continue
+                        logger.w(
+                            f"No real or post-quantization data found for '{input_spec.target}'. "
+                            f"Using a FakeTensor."
+                        )
+                        param = edge_program.state_dict[input_spec.target]
+                        result_map[input_spec.arg.name] = param
 
         return result_map
 
--- a/backends/nxp/tests/executorch_pipeline.py
+++ b/backends/nxp/tests/executorch_pipeline.py
@@ -98,6 +98,7 @@ def to_quantized_edge_program(
     custom_delegation_options=CustomDelegationOptions(),  # noqa B008
     get_quantizer_fn=None,
     use_neutron_for_format_conversion=True,
+    use_quant_state_dict=True,
 ) -> EdgeProgramManager:
     _neutron_target_spec = NeutronTargetSpec(target, neutron_converter_flavor)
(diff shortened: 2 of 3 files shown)


backends/nxp/backend/edge_program_converter.py
```diff
@@ -206,8 +206,12 @@ def map_inputs_to_parameters(
                         result_map[input_spec.arg.name] = param
 
                     else:
-                        # There is no data available.
-                        continue
+                        logger.w(
+                            f"No real or post-quantization data found for '{input_spec.target}'. "
+                            f"Falling back to state_dict entry (may contain FakeTensor)."
```
MartinPavella: > (may contain FakeTensor)

Nit: It will always be a `FakeTensor` here. Consider changing the message.

```python
logger.w(
    f"No real or post-quantization data found for '{input_spec.target}'. "
    "Using a FakeTensor."
)
```

MartinPavella (approved):

backends/nxp/backend/edge_program_converter.py
```diff
@@ -206,8 +206,12 @@ def map_inputs_to_parameters(
                         result_map[input_spec.arg.name] = param
 
                     else:
-                        # There is no data available.
-                        continue
```
MartinPavella, line 212: I originally wanted to avoid returning `FakeTensors`, which is why there is no fallback here. But if you really can't provide the data in your internal tests, I guess we can accept it. We will need to keep in mind that the mapping created by this function may return `FakeTensors`.

StrycekSimon (approved):
Is the test supposed to fail without the changes in this PR (on current main)? Because it passed when I tried...

Anyways, makes sense to me and looks good. 👍

MartinPavella:
> Is the test supposed to fail without the changes in this PR (on current main)? Because it passed when I tried...
> 
> Anyways, makes sense to me and looks good. 👍

Our tests are unaffected, because we use the post-quantization state dict. Only Jake's internal test failed as he is not using the state-dict. And regarding our mobilenet example, I'm working on a different fix.

PR #17453, opened 2026-02-13
Fix out-of-bounds write in Cadence HiFi bmm bias buffer initialization

rascani:
### Summary
The temporary zero-bias buffer for xa_nn_matmul_f32xf32_f32 was being "initialized" by writing a single zero one element past the end of the allocation (`tmp[batch_size * m * p] = {0}`). Replace with memset to correctly zero the entire buffer, and use sizeof(FLOAT32) consistently.

This diff was authored with the help of Claude.

### Test plan
CI


mcremon-meta (approved):
cc @dijopaul and @cad-audio

rascani:
Failures unrelated.

Issue #17299, opened 2026-02-08
Vulkan backend produces all-zero outputs on PowerVR GPU (Pixel 10 Pro)


abdelaziz-mahdy:
## Summary

The Vulkan backend produces **all-zero outputs** on a **PowerVR D-Series GPU** (Google Pixel 10 Pro). The same model files work correctly on macOS via MoltenVK and on Android via XNNPACK.

## Environment

- **Device**: Google Pixel 10 Pro
- **GPU**: PowerVR D-Series DXT-48-1536 MC1 (`maxImageDimension3D = 2048`)
- **ExecuTorch**: Built from source, branch `fix/vulkan-texture-ubo-budget` (my UBO budget fix from PR #17294, rebased on upstream/main as of Feb 7 2026, commit `ba2516cefa`)
- **NDK**: 28.2.13676358 (Clang 19.0.1)
- **Build**: Release, arm64-v8a, XNNPACK + Vulkan backends

## What I Observe

| Platform | Backend | YOLO (v8) | MobileNet |
|----------|---------|-----------|-----------|
| macOS | XNNPACK | Correct | Correct |
| macOS | Vulkan (MoltenVK) | Correct | Correct |
| Android | XNNPACK | Correct | Correct |
| Android | Vulkan (PowerVR) | All zeros | NaN values |

- **YOLO**: Output tensor shape `[1, 84, 8400]` is correct, but all confidence values are exactly `0.0`
- **MobileNet**: Output contains NaN values
- Both `texture_limits: (2048, 2048, 2048)` and `storage_type_override: BUFFER` produce the same zero results

## What I Found by Adding Tracing

I added `__android_log_print` traces to `VulkanBackend.cpp`, `ComputeGraph.cpp`, and `StagingBuffer.cpp` to narrow things down. Key findings:

### 1. GPU is PowerVR — no support in ExecuTorch

```
GPU: name='powervr d-series dxt-48-1536 mc1', max3D=2048
```

ExecuTorch's Vulkan backend only handles Adreno, Mali, NVIDIA, and SwiftShader. There is zero PowerVR-specific handling.

### 2. Output texture extents exceed `maxImageDimension3D`

```
output_tensor[3] extents=(2100,64,1)  max3D=2048 EXCEEDS=1
output_tensor[4] extents=(2100,4,16)  max3D=2048 EXCEEDS=1
```

[...]

PR #17467, opened 2026-02-14
[ET-VK] Use direct buffer-to-image copy for 1D tensor prepacking

abdelaziz-mahdy:
## Summary

Fixes 1D tensor prepacking (e.g., conv2d bias) producing incorrect results on GPUs with strict texture bounds checking (PowerVR).

The `nchw_to_image` compute shader uses `axis_map` coordinate remapping designed for multi-dimensional tensors. For 1D tensors padded to 4D with three fake dimensions of size 1, the coordinate math produces out-of-bounds texture writes. PowerVR GPUs silently drop these writes, so bias values never reach the texture. Other GPUs (Adreno, Mali) wrap coordinates, masking the bug.

For 1D width-packed tensors where the element count is divisible by 4, the staging buffer data is already in the correct layout. This change bypasses the compute shader and uses `vkCmdCopyBufferToImage` to copy the data directly.

Changes:
- `PrepackNode::encode()` — detect 1D width-packed tensors and use direct copy path
- `CommandBuffer::copy_buffer_to_image()` — new helper for buffer-to-image DMA with post-copy barrier
- `Allocator::create_staging_buffer()` — add `TRANSFER_SRC` to staging buffer usage flags (harmless for existing paths, enables direct copy)

Related: #17299

## Test plan

- [ ] Tested on Pixel 10 Pro (PowerVR DXT-48-1536) — conv2d bias values now reach the texture correctly
- [ ] Verified no regression on other GPUs (Adreno, Mali) since the coordinate wrapping behavior is unchanged for multi-dimensional tensors

### DIFF
--- a/backends/vulkan/runtime/graph/ops/PrepackNode.cpp
+++ b/backends/vulkan/runtime/graph/ops/PrepackNode.cpp
@@ -118,6 +118,65 @@ void PrepackNode::encode(ComputeGraph* graph) {
 
   context->check_device_capabilities(shader_);
 
+  // For 1D width-packed tensors (e.g., conv2d bias), use a direct
+  // vkCmdCopyBufferToImage instead of the nchw_to_image compute shader.
+  //
+  // The nchw_to_image shader uses axis_map coordinate remapping that assumes
+  // multi-dimensional tensors. For 1D tensors padded to 4D with three fake
+  // dimensions of size 1, the coordinate math can produce out-of-bounds texture
+  // writes. Some GPU drivers (e.g., PowerVR) strictly validate texture bounds
+  // and silently drop these writes, causing bias values to never reach the
+  // texture. Other drivers wrap coordinates, masking the bug.
+  //
+  // The staging buffer already contains correctly ordered data for 1D
+  // width-packed tensors, so we can copy it directly without remapping.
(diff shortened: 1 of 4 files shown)


abdelaziz-mahdy:
## Closing - Superseded by Serialize Prepack Fix

I'm closing this PR. The 1D bias prepacking issue on PowerVR turned out to be a symptom of a broader problem, not a coordinate math issue.

### What I Found

The actual root cause is that PowerVR corrupts prepacked constant data when multiple prepack compute dispatches are batched in a single Vulkan command buffer. Only the first constant in the batch is correct; subsequent constants read as zero.

For conv2d bias, this meant the bias tensor (typically the second prepacked tensor after weights) read as zero, producing the incorrect output I attributed to out-of-bounds texture writes.

### The Real Fix

A change in `ComputeGraph::prepack()` that serializes prepack dispatches on PowerVR. This fixes bias prepacking, weight prepacking, scalar constant prepacking, and every other multi-constant model — without touching `PrepackNode`, `CommandBuffer`, or `Allocator`.

I tested MobileNet V3 Small on Pixel 10 Pro with only this fix (all changes from this PR reverted): FP32 and FP16 both produce correct classification results, matching XNNPACK output.

PR #17468 has the fix. Details in #17299.

PR #17473, opened 2026-02-16
Arm backend: Add fp-quantized testing to TosaPipelineINT

AdrianLundell:
Adds a comparison between the original and quantized graph to catch bugs in transform_for_annotation and strange quantization behaviors.

Tests are updated according to the following rules:
1. If the result is unexpectedly off, bug tickets have been created and tests are marked as xfail or using high tolerance limits.
2. If the result is slightly off or there is a reasonable explanation for diffs, tolerances are increased with a comment explaining why.
3. If the input data is impossible to reasonably quantize, the input data is changed.

Additionally lowers default epsilon value since 2**-12 for the arm quantization specs which reduces diffs in test cases with small value ranges and fixes a typo for layer norm which exposed a numerical issue for Ethos-U55/U85.

### DIFF
--- a/backends/arm/quantizer/arm_quantizer.py
+++ b/backends/arm/quantizer/arm_quantizer.py
@@ -92,7 +92,7 @@ def get_symmetric_quantization_config(
         bias.
 
     """
-    extra_args: Dict[str, Any] = {"eps": 2**-12}
+    extra_args: Dict[str, Any] = {"eps": 2**-16}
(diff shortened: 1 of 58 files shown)


zingo (approved):
Touched a lot of files but as all changes are within files I see no need for buck2 problems. Please merge when tests pass!

AdrianLundell:
Fails unrelated

PR #17478, opened 2026-02-16
Arm backend: Fold constants in transform_for_annotation

AdrianLundell:
This pass runs the same code as used in
convert_pt2e(m, fold_quantize=True), forcing that behaviour in the arm-backend. This ensures that linear/convolution ops have constant parameters as inputs, fixing a bug where weights/bias was not quantized according to the weight/bias_qspec defined for the node as expected.


zingo (approved):

AdrianLundell:
Fails unrelated

PR #17479, opened 2026-02-16
NXP backend: Add post-quantization data utilization to `aot_neutron_compile.py`.

MartinPavella:
### Summary
A recent PR added the option to use the post-quantization state dict to access static data during quantization. This PR adds this feature to the `aot_neutron_compile.py`.

### Test plan
Unit-test with the example MobileNetV2 is provided.


JakeStevens:
looks like it is failing to load from URL.

Maybe just use random weights + data instead of "real"?

robert-kalmar (approved):

MartinPavella:
The failing checks seem unrelated and the internal build passed. Merging.

PR #17484, opened 2026-02-17
Arm backend: Don't fall back to TOSA-1.0+INT in case of exception

martinlsm:
aot_arm_compiler.py silently fell back to TOSA-1.0+INT in case the user-provided TOSA spec could not be created. This may easily mislead the user by hiding actual errors with the input to the program. Change aot_arm_compiler.py to instead raising the exception in case the user-provided TOSA spec can not be created.


zingo (approved):
OK to merge after Arm tests are fixed/pass This errors are not related to this so this PR might need a rebase after fix to get proper testing.

martinlsm:
Arm failures are unrelated. They are seen on main also.

PR #17485, opened 2026-02-17
Arm backend: Configure logging at runtime in aot_arm_compiler

martinlsm:
Move logging setup out of module import and into `get_args()`, and always apply it with `force=True`. This prevents prior imports from locking the root logger at a different level.


zingo (approved):
OK to merge after Arm tests are fixed/pass This errors are not related to this PR so this PR might need a rebase after fix to get proper testing.

martinlsm:
Failures were unrelated

PR #17486, opened 2026-02-17
Arm backend: Pin the pybind11 version to fix tosa-tools build

tom-arm:
Change-Id: I542f6284b3c6d76f8c13ed177fcde266c3343f71

### DIFF
--- a/.ci/scripts/unittest-linux-cmake.sh
+++ b/.ci/scripts/unittest-linux-cmake.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
+# Copyright 2026 Arm Limited and/or its affiliates.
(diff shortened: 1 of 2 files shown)


zingo (approved):

zingo:
Hi @SS-JIA this is an emergency fix for a problem with tosa-tools and pybinding, we are trying to see if they can fix it without this but as this blocks main we would like to do this while waiting to see if it can be solved in a better way. 

As this touches non Arm file we need a Meta review, and if OK feel free to merge also to get it in faster, We (Arm) will try to monitor this PR and will also merge if I see you are OK with it but due to timediff/sleeping might miss a few hours :)

SS-JIA (approved):
I think this should be okay as a temporary measure. CI signals appear to be unaffected by this change.

cc: @mergennachin 

We should remember to revert this change, however once the issue is resolved @zingo

PR #17491, opened 2026-02-17
Skeleton out more robust program validation

JacobSzwejbka:
Summary:
We should catch errors as early as possible, a great place to do that would be at program init where we can check things once and then not necessarily have to check them every inference or just in general being able to soft fail at invalid programs.

Currently just testing a couple obvious bad things like invalid sizes, or inconsistent list contents. Can add more checks in follow ups.

Differential Revision: D92787645

### DIFF
--- a/runtime/executor/program.cpp
+++ b/runtime/executor/program.cpp
@@ -14,6 +14,7 @@
 #include <executorch/runtime/core/event_tracer_hooks.h>
 #include <executorch/runtime/executor/memory_manager.h>
 #include <executorch/runtime/executor/method.h>
+#include <executorch/runtime/executor/program_validation.h>
 #include <executorch/runtime/platform/profiler.h>
 #include <executorch/schema/extended_header.h>
 #include <executorch/schema/program_generated.h>
@@ -150,6 +151,13 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
         ok,
         InvalidProgram,
         "Verification failed; data may be truncated or corrupt");
+    const executorch_flatbuffer::Program* flatbuffer_program =
+        executorch_flatbuffer::GetProgram(program_data->data());
+    Error err = validate_program(flatbuffer_program);
+    ET_CHECK_OR_RETURN_ERROR(
+        err == Error::Ok,
+        InvalidProgram,
+        "Program validation failed: likely a corrupt file");
 #else
     ET_LOG(
         Info, "InternalConsistency verification requested but not available");
--- a/runtime/executor/program.h
+++ b/runtime/executor/program.h
@@ -8,7 +8,6 @@
 
 #pragma once
 
-#include <cinttypes>
 #include <cstdint>
 #include <optional>
 
@@ -58,10 +57,15 @@ class Program final {
      */
     Minimal,
     /**
+     * When ET_ENABLE_PROGRAM_VERIFICATION is enabled,
      * Do full verification of the data, ensuring that internal pointers are
      * self-consistent and that the data has not been truncated or obviously
      * corrupted. May not catch all types of corruption, but should guard
      * against illegal memory operations during parsing.
+     * Also performs additional semantic validation such as:
+     * - Tensor numel overflow checks (ensuring size calculations don't
+     * overflow)
+     * - List element type validation
      *
      * Will have higher runtime overhead, scaling with the complexity of the
      * proram data.
--- a/runtime/executor/program_validation.cpp
+++ b/runtime/executor/program_validation.cpp
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <executorch/runtime/executor/program_validation.h>
+
+#include <cstdint>
+
+#include <executorch/runtime/core/exec_aten/util/scalar_type_util.h>
+#include <executorch/runtime/platform/log.h>
+#include <executorch/schema/program_generated.h>
+
+#include <c10/util/safe_numerics.h>
+
+namespace executorch {
+namespace runtime {
+
+ET_NODISCARD Error
+validate_tensor(const executorch_flatbuffer::Tensor* tensor) {
+  if (tensor == nullptr) {
+    return Error::InvalidProgram;
+  }
+
+  const auto* sizes = tensor->sizes();
+  if (sizes == nullptr) {
+    return Error::InvalidProgram;
+  }
+
+  ssize_t numel = 1;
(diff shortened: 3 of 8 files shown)


runtime/executor/program_validation.h
```diff
@@ -0,0 +1,53 @@
...
+ * Validates that computing numel (number of elements) from the tensor's sizes
+ * will not overflow. This check should be performed before creating TensorImpl
+ * objects to prevent undefined behavior from integer overflow.
+ *
+ * @param[in] tensor The flatbuffer Tensor to validate.
+ * @return Error::Ok if the numel calculation is safe, Error::InvalidProgram
+ *     if computing numel would overflow.
+ */
+ET_NODISCARD Error
+validate_tensor_numel_no_overflow(const executorch_flatbuffer::Tensor* tensor);
```
larryliu0820: no need to add "_no_overflow" to method name I think.

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,196 @@
...
+          "Size must be non-negative, got %d at dimension %u",
+          size,
+          static_cast<unsigned>(i));
+      return Error::InvalidProgram;
+    }
+
+    bool overflow =
+        c10::mul_overflows(numel, static_cast<ssize_t>(size), &numel);
+    if (overflow) {
+      ET_LOG(Error, "numel overflowed");
```
larryliu0820: Would be good to print out index i here, also useful to print the current numel.

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,196 @@
...
+    return Error::InvalidProgram;
+  }
+
+  size_t nbytes;
+  bool nbytes_overflow = c10::mul_overflows(
+      static_cast<size_t>(numel),
+      executorch::runtime::elementSize(scalar_type),
+      &nbytes);
+  if (nbytes_overflow) {
+    ET_LOG(Error, "nbytes overflowed");
```
larryliu0820: Same here

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <executorch/runtime/platform/log.h>
+#include <executorch/schema/program_generated.h>
+
+#include <c10/util/safe_numerics.h>
+
+namespace executorch {
+namespace runtime {
+
+ET_NODISCARD Error
+validate_tensor_numel_no_overflow(const executorch_flatbuffer::Tensor* tensor) {
```
lucylq: Call this validate_tensor? Currently it has numel overflow and scalar type, and could also check dim order.

runtime/executor/program.cpp
```diff
@@ -150,6 +151,13 @@ Result<executorch_flatbuffer::ExecutionPlan*> get_execution_plan(
         ok,
         InvalidProgram,
         "Verification failed; data may be truncated or corrupt");
+    const executorch_flatbuffer::Program* flatbuffer_program =
+        executorch_flatbuffer::GetProgram(program_data->data());
+    Error err = validate_program(flatbuffer_program);
```
lucylq, line 156: If we centralize pte validation here there'll be duplicate checks, e.g. [tensor parser](https://github.com/pytorch/executorch/blob/main/runtime/executor/tensor_parser_portable.cpp#L45-L50)

Is the plan to remove those and rely on program_validation?

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (values == nullptr) {
+      continue;
+    }
+
+    for (flatbuffers::uoffset_t value_idx = 0; value_idx < values->size();
+         value_idx++) {
+      const auto* value = values->Get(value_idx);
+      if (value == nullptr) {
+        continue;
+      }
```
larryliu0820, line 118: why do we choose to continue? If we can't get a value that means the program is invalid right

runtime/executor/program_validation.cpp
lucylq: e.g. dim order check https://github.com/pytorch/executorch/pull/17314

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,196 @@
...
+    // A program with no execution plans is unusual but not necessarily invalid.
+    return Error::Ok;
+  }
+
+  for (flatbuffers::uoffset_t plan_idx = 0; plan_idx < execution_plans->size();
+       plan_idx++) {
+    const auto* plan = execution_plans->Get(plan_idx);
+    if (plan == nullptr) {
+      continue;
+    }
```
larryliu0820, line 101: if the program say it has 1 execution plan but we can't get it, I think it's an invalid program

runtime/executor/program_validation.cpp
```diff
@@ -0,0 +1,204 @@
...
+              Error,
+              "Tensor validation failed for value %u in execution plan %u",
+              static_cast<unsigned>(value_idx),
+              static_cast<unsigned>(plan_idx));
+          return err;
+        }
+      }
+
+      // Check if this value is a TensorList.
+      if (value->val_type() == executorch_flatbuffer::KernelTypes::TensorList) {
```
lucylq, line 137: Maybe roll in OptionalTensorList here as well?

runtime/executor/program.h
```diff
@@ -58,10 +57,15 @@ class Program final {
      */
     Minimal,
     /**
+     * When ET_ENABLE_PROGRAM_VERIFICATION is enabled,
      * Do full verification of the data, ensuring that internal pointers are
      * self-consistent and that the data has not been truncated or obviously
      * corrupted. May not catch all types of corruption, but should guard
      * against illegal memory operations during parsing.
+     * Also performs additional semantic validation such as:
```
lucylq, line 65: Oh, I think these checks should happen regardless of `ET_ENABLE_PROGRAM_VERIFICATION`? At least that is the current state.

runtime/executor/program.cpp
lucylq: Yeah commented earlier but also think these happen regardless of `ET_ENABLE_PROGRAM_VERIFICATION` ..

larryliu0820 (approved):

lucylq (approved):

runtime/executor/program_validation.cpp
JacobSzwejbka: Yeah we can do follow ups. Just want to get some initial framework in

runtime/executor/program.h
JacobSzwejbka: Ah yeah we have some just in time checks today. Theres code size concerns, so for now Im just going to place it behind the flag (which is on by default).

lucylq: I think we can raise the binary size limits here - the code size should go back down after deduplicating the checks.

If this layer is gated, I'd be wary about deduplicating and regressing security for non-InternalConsistency case, which means we have two copies of every check. And when we have more security checks, it's less useful to add it to program_validation..

PR #17498, opened 2026-02-17
2/x: Add runtime specs support to BackendInitContext

metascroy:
Summary:
This diff extends `BackendInitContext` to receive and expose runtime specs (load-time options) to backend delegates. This allows backends to access per-delegate configuration that was passed via `LoadBackendOptionsMap` at `Module::load()` time.

Key changes:
- Added `runtime_specs_` member to `BackendInitContext` storing `Span<const BackendOption>`
- Constructor now accepts optional `runtime_specs` parameter
- Added `runtime_specs()` accessor returning the full span
- Added type-safe getter methods:
  - `get_runtime_spec_bool(key)` - Returns `Result<bool>`
  - `get_runtime_spec_int(key)` - Returns `Result<int>`
  - `get_runtime_spec_string(key)` - Returns `Result<const char*>`

Backend delegates can now retrieve their configuration:
```cpp
Result<const char*> compute_unit = context.get_runtime_spec_string("compute_unit");
if (compute_unit.ok()) {
  // Use compute_unit.get()
}
```

Reviewed By: larryliu0820

Differential Revision: D92358600

### DIFF
--- a/runtime/backend/backend_init_context.h
+++ b/runtime/backend/backend_init_context.h
@@ -7,9 +7,15 @@
  */
 
 #pragma once
+#include <executorch/runtime/backend/options.h>
+#include <executorch/runtime/core/error.h>
 #include <executorch/runtime/core/event_tracer.h>
 #include <executorch/runtime/core/memory_allocator.h>
 #include <executorch/runtime/core/named_data_map.h>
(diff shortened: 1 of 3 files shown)


larryliu0820 (approved):

PR #17503, opened 2026-02-17
Fix uninitialized memory in ExecuTorch portable batch norm kernel.

khazaei:
Summary:
Fix uninitialized memory bug in ExecuTorch portable batch norm kernel.

The _native_batch_norm_legit_no_stats_out function was accumulating mean and variance values into buffers without initializing them to zero first. This caused incorrect batch normalization results due to garbage values in uninitialized memory.

Changes:
- Added #include <cstring> for std::memset
- Initialize mean_data and invstd_data buffers to zero before accumulation loop

Also includes a unit test that reproduces the failure.

Differential Revision: D93119990

### DIFF
--- a/kernels/portable/cpu/op_native_batch_norm.cpp
+++ b/kernels/portable/cpu/op_native_batch_norm.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <cmath>
+#include <cstring>
 #include <tuple>
 
 #include <executorch/kernels/portable/cpu/util/normalization_ops_util.h>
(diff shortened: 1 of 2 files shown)


hsharma35 (approved):

khazaei:
@manuelcandales how do you suggest i resolve the failing issues? It doesn't look like my changes are related to the failures.

khazaei:
@manuelcandales the failures seem to be because of some SAMSUNG_AI_LITECORE_KEY api key didn't make it into the job environment. this seems like more of a ci issue, and its not something my changes have caused. Can you let me know how we can proceed further?

hsharma35 (approved):

PR #17510, opened 2026-02-17
Use stable topological sort in backend partitioning to fix SymInt constraint ordering

SS-JIA:
Summary:
During backend partitioning, `create_submodule_from_nodes` uses a topological sort
to order nodes in extracted subgraphs. The standard `topo_sort` (FIFO queue-based)
does not preserve relative ordering of independent nodes, which can reorder
`_assert_scalar` constraint nodes after the operations they constrain.

When the reordered subgraph is later retraced, fresh unbacked SymInts are created
but constraints haven't been registered yet, causing `GuardOnDataDependentSymNode`
errors. This manifests in models like LLaMA that use data-dependent indexing
(e.g., `input_pos[0].item()` for KV-cache slicing with `torch._check` bounds).

The fix replaces `topo_sort` with `_stable_topo_sort` which preserves the original
node ordering for nodes with no data dependencies between them, ensuring constraint
nodes remain before the operations they guard.

This change was authored with Claude.

Test Plan:
- Added `exir/tests/test_symint_preservation.py` which verifies stable topo sort
  preserves constraint ordering during subgraph extraction

### DIFF
--- a/exir/lowered_backend_module.py
+++ b/exir/lowered_backend_module.py
@@ -47,8 +47,70 @@ from torch.fx.passes.utils.fuser_utils import (
     insert_subgm,
     legalize_graph,
     NodeList,
-    topo_sort,
+    topo_sort as _topo_sort,
 )
+
+
+def _stable_topo_sort(nodes: NodeList) -> NodeList:
+    """
+    Stable topological sort that preserves the relative order of independent nodes.
+
+    Unlike the standard topo_sort which uses a queue (FIFO), this implementation
+    preserves the original ordering of nodes when they have no data dependencies
+    between them. This is important for constraint nodes (like _assert_scalar)
+    which must execute before the operations they constrain, even though there's
+    no explicit data edge between them.
+
+    Args:
+        nodes: List of nodes to sort. Assumed to already be in a valid topological
+               order (as is the case for nodes from an FX graph).
+
+    Returns:
+        Topologically sorted nodes that preserve original relative ordering of
+        independent nodes.
+    """
+    if not nodes:
+        return nodes
+
+    # Build indegree map only considering nodes in the partition
+    node_set = set(nodes)
+    indegree_map = {node: 0 for node in nodes}
+
+    for node in nodes:
+        for input_node in node.all_input_nodes:
+            if input_node in node_set:
+                indegree_map[node] += 1
+
+    # Create position map to preserve original ordering
+    position = {node: i for i, node in enumerate(nodes)}
+
+    sorted_nodes: NodeList = []
+    remaining = set(nodes)
+
+    while remaining:
+        # Find all nodes with zero indegree (can be scheduled now)
+        ready = [n for n in remaining if indegree_map[n] == 0]
+
+        if not ready:
+            # No nodes ready means we have a cycle
+            raise RuntimeError(
+                f"Cycle detected in topological sort. " f"Remaining nodes: {remaining}"
+            )
+
+        # Sort ready nodes by their original position to maintain stability
+        ready.sort(key=lambda n: position[n])
+
+        # Take the first ready node (earliest in original order)
+        node = ready[0]
+        sorted_nodes.append(node)
+        remaining.remove(node)
+
+        # Update indegrees of nodes that depend on this one
+        for user in node.users:
+            if user in node_set:
+                indegree_map[user] -= 1
+
+    return sorted_nodes
 
 
 class LoweredBackendModule(torch.nn.Module):
@@ -764,7 +826,7 @@ def create_submodule_from_nodes(
         The submodule that has been partitioned, the call_module node in the
         toplevel graph module calling the submodule
     """
-    sorted_nodes = topo_sort(node_list)
+    sorted_nodes = _stable_topo_sort(node_list)
 
     submodule_name = "fused_" + tag
     sub_gm, orig_inputs, orig_outputs = fuse_as_graphmodule(
--- a/exir/tests/TARGETS
+++ b/exir/tests/TARGETS
@@ -501,6 +501,18 @@ python_unittest(
 )
 
 python_unittest(
+    name = "symint_preservation",
+    srcs = [
+        "test_symint_preservation.py",
+    ],
+    deps = [
+        "//caffe2:torch",
+        "//executorch/exir:lib",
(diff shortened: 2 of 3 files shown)


exir/lowered_backend_module.py
```diff
@@ -47,10 +47,73 @@
     insert_subgm,
     legalize_graph,
     NodeList,
-    topo_sort,
+    topo_sort as _topo_sort,
```
JacobSzwejbka, line 50: naively where does the reordering come from in topo_sort?

SS-JIA: From Claude 😛 

---

Here's the issue. The original topo_sort (line 15 of fuser_utils.py) uses a SimpleQueue (FIFO). The reordering happens through the interaction of two things:                                     
                                                                                                                                                                                                    
  1. Initial seeding order matters, but it's preserved                                                                                                                                              
                                                                                                                                                                                                    
  Lines 20-25: Nodes with indegree 0 are enqueued in their original order. So far, no reordering.                                                                                                   
                                                                                                                                                                                                    
  2. The problem is when dependents become ready                                                                                                                                                    

  Lines 32-36: When a node is dequeued and processed, its users whose indegree drops to 0 are immediately appended to the back of the queue. Consider this graph:

```
  A → B → D
  A → C
        E  (independent, no deps on A/B/C)
```

  If the original order is [A, B, C, D, E]:
  
```  ┌──────┬───────────────────────────────────────────────────┬───────────────┐
  │ Step │                    Queue state                    │     Emit      │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ Init │ [A, E] (indegree 0)                               │ —             │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ 1    │ [E] → process A, B and C become ready → [E, B, C] │ A             │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ 2    │ [B, C] → process E                                │ A, E          │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ 3    │ [C] → process B, D becomes ready → [C, D]         │ A, E, B       │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ 4    │ [D] → process C                                   │ A, E, B, C    │
  ├──────┼───────────────────────────────────────────────────┼───────────────┤
  │ 5    │ [] → process D                                    │ A, E, B, C, D │
  └──────┴───────────────────────────────────────────────────┴───────────────┘
```

  The FIFO queue interleaves E before B and C, even though in the original graph E appeared after D.

  In the SymInt case specifically: when the add(u0, u34) node is processed, it releases both _assert_scalar(end <= 2048) and slice_copy into the queue simultaneously. But slice_copy might already
  be in the queue (or get enqueued first) depending on which other nodes released it, so it can end up before the assert. The FIFO queue has no mechanism to prefer the original program order — it
  just serves nodes in the order they happened to become ready.

  The _stable_topo_sort fix addresses this by always picking the ready node with the smallest original position index, rather than whatever the FIFO happens to yield.

exir/lowered_backend_module.py
```diff
@@ -47,10 +47,73 @@
...
 )
 
 
+def _stable_topo_sort(nodes: NodeList) -> NodeList:
```
JacobSzwejbka, line 54: Why is it a given that the assert ops are in the proper order on the original node list?

exir/lowered_backend_module.py
```diff
@@ -47,10 +47,73 @@
...
+    """
+    Stable topological sort that preserves the relative order of independent nodes.
+
+    Unlike the standard topo_sort which uses a queue (FIFO), this implementation
```
JacobSzwejbka, line 58: can we delete the other one?

exir/lowered_backend_module.py
```diff
@@ -47,10 +47,73 @@
...
+    preserves the original ordering of nodes when they have no data dependencies
+    between them. This is important for constraint nodes (like _assert_scalar)
+    which must execute before the operations they constrain, even though there's
+    no explicit data edge between them.
+
+    Args:
+        nodes: List of nodes to sort. Assumed to already be in a valid topological
```
JacobSzwejbka, line 65: If the assumption is the inputs are already sorted what is the point of the function?

JacobSzwejbka: Assuming this comment is incorrect, how do we generate a list of nodes not in topographical ordering from an fx graph anyway? Seems fine to add a pass to compel it but also seems like all our infra should do it by default anyway?

exir/lowered_backend_module.py
SS-JIA: The other one is in core Pytorch (`torch.fx.passes.utils.fuser_utils`) so there is a much larger blast radius to replace the original; that's why I opted to just fix it locally for now. Can raise it with core Pytorch folks to see if this kind of change makes sense in the default topo_sort function.

For more context, I was hitting an error exporting the Stories model to Vulkan due to this issue

```shell
torch.fx.experimental.symbolic_shapes.GuardOnDataDependentSymNode: Could not guard on data-dependent expression u407 + 3 > 2048 (unhinted: s11 + u407 > 2048).  (Size-like symbols: none)
```

and this change fixes the error. So part of my motivation for putting up this PR is to unblock setting up stories export tests for Vulkan.

exir/lowered_backend_module.py
SS-JIA: > If the assumption is the inputs are already sorted what is the point of the function?

I think the issue is that the current `topo_sort` function may modify the order of nodes. So even if the inputs are sorted in the correct fashion, there's no guarantee that the outputs will be (and that's what's causing a data dependency error for me during export).

The point of the new function is therefore that if the input node list is in a correct order, then the correct ordering will be at least preserved.

exir/lowered_backend_module.py
SS-JIA: There isn't necessarily a guarantee that it's the case. The purpose of the function is more so that assuming the assert ops are ordered correctly in the input node list, that order will be preserved in the output node list. It seems that the current topo_sort function can order the nodes in a way where the assert op occurs after the value that it is guarding against is used, since there is no explicit dependency between the assert and the usage of the symbolic int it is guarding.

PR #17511, opened 2026-02-17
Bump torch pin to 2.11.0.dev20260215

larryliu0820:
Bump torch pin to 2.11.0.dev20260215 so that RTX 5080 works.


JacobSzwejbka (approved):

JacobSzwejbka:
@anzr299 the openvino failure here looks like the type of thing that would be resolved by https://github.com/openvinotoolkit/nncf/pull/3854 is that PR still alive?

manuelcandales (approved):

install_requirements.py
```diff
@@ -119,12 +119,12 @@ def install_optional_example_requirements(use_pytorch_nightly):
...
     DOMAIN_LIBRARIES = [
         (
-            f"torchvision==0.25.0.{NIGHTLY_VERSION}"
+            f"torchvision==0.26.0.{NIGHTLY_VERSION}"
             if use_pytorch_nightly
             else "torchvision"
         ),
         (
-            f"torchaudio==2.10.0.{NIGHTLY_VERSION}"
+            f"torchaudio==2.11.0.{NIGHTLY_VERSION}"
```
mergennachin, line 127: Do we still need these @larryliu0820

larryliu0820: yeah the versions are correct for dev20260216.

anzr299:
> @anzr299 the openvino failure here looks like the type of thing that would be resolved by [openvinotoolkit/nncf#3854](https://github.com/openvinotoolkit/nncf/pull/3854) is that PR still alive?

We are discussing offline if we really need to add torchao as a dependency or we can work around importing it. It is unclear from the torch.ao -> torchao migration plan(https://dev-discuss.pytorch.org/t/torch-ao-quantization-migration-plan/2810/4?u=anzr299) whether we can independently remove torchao and keep legacy functions from torch.ao instead.

Only case where we need to certain imports are necessary to pass signatures checks in pytorch and so on.

PR #17513, opened 2026-02-17
add buck and vtito model support

billmguo:
Summary: Add Andorid standalone oss runner to test vtito model

Differential Revision: D93530191


larryliu0820 (approved):

PR #17523, opened 2026-02-18
Draft: Integration of `QuantizeFusedConvBnBiasPass` to NXP conversion pipeline

StrycekSimon:
This is a placeholder PR used as demonstration of integration of changes applied in [this pull request](https://github.com/pytorch/executorch/pull/17348).

Includes testing function for reference.

### DIFF
--- a/backends/nxp/quantizer/utils.py
+++ b/backends/nxp/quantizer/utils.py
@@ -13,6 +13,9 @@ from collections.abc import Iterable
 from typing import Any, Dict, List, Tuple, Type
 
 import torch
+from executorch.backends.transforms.quantize_fused_convbn_bias_pass import (
+    QuantizeFusedConvBnBiasPass,
+)
 from torch import fx
 from torch._ops import OpOverload
 from torch.export import ExportedProgram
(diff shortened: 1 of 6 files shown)


backends/nxp/quantizer/utils.py
```diff
@@ -179,17 +182,21 @@ def calibrate_and_quantize(
...
+        module = prepare_pt2e(module, quantizer)
 
     for data in calibration_inputs:
-        m(*data)
-    m = convert_pt2e(m)
+        module(*data)
+    module = convert_pt2e(module)
+
+    # Without this export, conv bias is not in the graph_signature.
+    model = torch.export.export(module, calibration_inputs[0], strict=True)
```
JakeStevens, line 198: So you want to try to drop needing this?

backends/nxp/quantizer/utils.py
```diff
@@ -179,17 +182,21 @@ def calibrate_and_quantize(
...
+    bias_quant_pass = QuantizeFusedConvBnBiasPass(model)
+    model = bias_quant_pass(model.graph_module)
 
-    return m
+    return model.graph_module
```
JakeStevens, line 202: is the change to use graph_module desired outcome or currently just needed for the pass as is?

backends/nxp/quantizer/utils.py
StrycekSimon: I just wanted to be clear as export wasn't used before in this function. It is okay to keep it.

backends/nxp/quantizer/utils.py
StrycekSimon: The function previously returned `GraphModule` instance. Returning `model` would be inconsistent with the rest of our pipeline.

PR #17532, opened 2026-02-18
bump torch/torchvision/torchaudio pins

manuelcandales:
As titled


mergennachin:
You might need to run `.github/scripts/update_pytorch_pin.py` so that it does necessary syncing (aligns torch commit as well as core aten headers).

torch_pin.py
```diff
@@ -1,2 +1,4 @@
-TORCH_VERSION = "2.11.0"
-NIGHTLY_VERSION = "dev20251222"
+TORCH_VERSION = "2.12.0"
+TORCHVISION_VERSION = "0.26.0"
+TORCHAUDIO_VERSION = "2.11.0"
+NIGHTLY_VERSION = "dev20260218"
```
mergennachin: Let's increase monotically 

When we release ET 1.2, we want to pin to 2.11

In our release branch, we don't want to go back to 2.11 from 2.12

Can you pick 1/29 version instead -- looks like the most stable nightly in 2.10

https://hud.pytorch.org/hud/pytorch/pytorch/nightly/1?per_page=50&mergeEphemeralLF=true

manuelcandales: I agree with your logic. I don't understand why 1/29. Should be 02/16, the last day of PT 2.11, no?

manuelcandales:
Yielding to https://github.com/pytorch/executorch/pull/17511

PR #17542, opened 2026-02-18
Cortex-M: change parameter-reporting logs from Info to Debug

rascani:
### Summary
These logs report quantization parameters and sizes on every kernel invocation, which is noisy at Info level. Move them to Debug so they're still available for debugging but don't clutter normal output.

### Test plan
```
pytest backends/cortex_m/test/
```


psiddh (approved):

psiddh (approved):

rascani:
Failures unrelated.

PR #17543, opened 2026-02-18
Cortex-M: remove extractScalarToInt helpers, use static_cast directly

rascani:
### Summary
Since #16322 changed all Scalar parameters to int64_t, the extractScalarToInt32 and extractScalarToInt helpers in cortex_m_ops_common.h are unnecessary indirection — they called Scalar::to<int64_t>() then cast, but the values are already int64_t.

Replace all call sites with direct static_cast and remove the helpers, the unused Scalar type alias, and the scalar_utils.h include. Also update the quantized_mul Python schema in operators.py to use int instead of Scalar, matching the C++ kernel and operators.yaml.

Additionally remove a duplicate kernel_includes.h include from cortex_m_ops_common.h.

### Test plan
```
pytest backends/cortex_m/test/
```

cc @digantdesai @SS-JIA @freddan80 @per @zingo @oscarandersson8218 @mansnils @Sebastian-Larsson @robell @psiddh @AdrianLundell


rascani:
Bumping this.

psiddh (approved):

AdrianLundell (approved):

rascani:
Failures unrelated.

PR #17546, opened 2026-02-19
Auto-initialize PAL when functions are called before et_pal_init()

rascani:
### Summary
Replace the debug-only assertion behavior with automatic initialization. When et_pal_current_ticks() or et_pal_emit_log_message() are called before et_pal_init(), they now call et_pal_init() automatically instead of aborting in debug builds or producing undefined behavior in release.

Uses std::call_once with a function-local static flag for thread-safe one-time initialization.

Rename executor_pal_death_test to executor_pal_autoinit_test since we now test auto-initialization behavior rather than death on missing init.


JacobSzwejbka (approved):
<deleted>

larryliu0820 (approved):

runtime/platform/default/android.cpp
```diff
@@ -69,12 +37,10 @@ static bool initialized = false;
...
 #endif // _MSC_VER
 void et_pal_init(void) {
-  if (initialized) {
-    return;
-  }
-
-  systemStartTime = std::chrono::steady_clock::now();
-  initialized = true;
+  [[maybe_unused]] static const bool init = [] {
+    systemStartTime = std::chrono::steady_clock::now();
```
larryliu0820, line 43: the behavior changes? like if we call et_pal_init() multiple times, we will override the systemStartTime now, before we don't.

larryliu0820 (approved):

runtime/platform/default/android.cpp
rascani: As discussed OOB, this will effectively still be a std::call_once since it is using a local static variable for initialization.

PR #17548, opened 2026-02-19
NXP backend: switch from raw tflite to from ir.lib.tflite

JakeStevens:
Summary:
Following convention, do not use from tflite but rather the lib in Executorch backend, eg

from executorch.backends.nxp.backend.ir.lib.tflite.BuiltinOperator import BuiltinOperator

This allows for usage in environments where tflite is not installed.

Reviewed By: SS-JIA

Differential Revision: D93624349


SS-JIA (approved):

JakeStevens:
NXP failure pre existing and un related, see

https://github.com/pytorch/executorch/pull/17551

Has same.

MartinPavella:
> NXP failure pre existing and un related, see
> 
> #17551
> 
> Has same.

We're looking into it.

PR #17586, opened 2026-02-20
Arm backend: Add FP16 tests of models (mv3, ic3)

martinlsm:
Add testing of the following models executed in FP16:
 - MobileNetV3
 - InceptionV3

This patch verifies that the Arm backend is able to lower full models in FP16 to valid TOSA, and execute them with acceptable numerical accuracy.

### DIFF
--- a/backends/arm/test/models/test_inception_v3_arm.py
+++ b/backends/arm/test/models/test_inception_v3_arm.py
@@ -24,6 +24,9 @@ from torchvision import models, transforms  # type: ignore[import-untyped]
 ic3 = models.inception_v3(weights=models.Inception_V3_Weights)
 ic3 = ic3.eval()
 
+ic3_fp16 = models.inception_v3(weights=models.Inception_V3_Weights).to(torch.float16)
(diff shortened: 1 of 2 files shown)


zingo (approved):
OK to merge when atol/rtol is numped to tests pass.
e.g. This seem to get 

```
FAILED backends/arm/test/models/test_inception_v3_arm.py::test_ic3_tosa_FP_fp16 - AssertionError: Output 0 does not match reference output.
	Given atol: 0.01, rtol: 0.001.
	Output tensor shape: torch.Size([1, 1000]), dtype: torch.float16
	Difference: max: 0.03515625, abs: 0.03515625, mean abs error: 0.005782970428466797.
	-- Model vs. Reference --
	 Numel: 1000, 1000
	Median: -0.06884765625, -0.06573486328125
	  Mean: -0.02605916690826416, -0.026028093814849853
	   Max: 2.703125, 2.67578125
	   Min: -2.623046875, -2.62109375
= 1 failed, 91 passed, 3 skipped, 7 xfailed, 952 warnings in 963.15s (0:16:03) =
```

PR #17596, opened 2026-02-20
Add memory planning to oss tests

lucylq:
### Summary
add to pytest.ini, and add cmake path-finding import for OSS

### Test plan
python -m unittest exir.tests.test_memory_planning

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -34,7 +34,7 @@ addopts =
     --ignore=exir/backend/test/demos/test_delegate_aten_mode.py
     --ignore=exir/backend/test/demos/test_xnnpack_qnnpack.py
     --ignore=exir/tests/test_memory_format_ops_pass_aten.py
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -34,7 +34,7 @@ addopts =
     --ignore=exir/backend/test/demos/test_delegate_aten_mode.py
     --ignore=exir/backend/test/demos/test_xnnpack_qnnpack.py
     --ignore=exir/tests/test_memory_format_ops_pass_aten.py
-    --ignore=exir/tests/test_memory_planning.py
+
```
JacobSzwejbka, line 37: lol. Didnt realize we had so much stuff disabled

JacobSzwejbka (approved):

pytest.ini
lucylq: Yeah, I can enable the others with the same fix here, probably.

PR #17597, opened 2026-02-20
Make NXP batch norm tests runnable in BUCK

JakeStevens:
Summary:
Make the NXP test_batch_norm_fusion tests compatible with the BUCK build
system. The tflite import in executors.py is made optional since
tensorflow/tflite_runtime are not available in the BUCK environment. The tests enabled do not rely on the functionality enabled by tflite.

Differential Revision: D93880277

### DIFF
--- a/backends/nxp/tests/executors.py
+++ b/backends/nxp/tests/executors.py
@@ -3,6 +3,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from __future__ import annotations
+
 import warnings
 from typing import Callable, Dict, Union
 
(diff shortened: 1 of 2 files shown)


backends/nxp/tests/executors.py
```diff
@@ -3,6 +3,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from __future__ import annotations
```
MartinPavella, line 6: ~This import is never used. Are you internally using Python <3.10 and need this for type annotations (such as | (union))? Or what is the purpose of this?~

Edit: Now I see it is required for the type hint:
```python
class TFLiteExecutor:
    _interpreter: tflite.Interpreter
```

MartinPavella (approved):
I assume the first commit is just a dependency, and only the second commit is meant for review in this PR. Perhaps it would be useful to state that explicitly. Otherwise, LGTM

JakeStevens:
> I assume the first commit is just a dependency, and only the second commit is meant for review in this PR. Perhaps it would be useful to state that explicitly. Otherwise, LGTM

Yes, just a dependency but not strictly needed, rebased so it is clear only one commit here and merging.

PR #17608, opened 2026-02-21
Fix thread safety of LLM token buffer in Android JNI

kirklandsign:
### Summary
The UTF-8 token buffer was a global variable shared across all ExecuTorchLlmJni instances, causing token corruption when multiple LlmModule instances run generate() concurrently. Move the buffer to a local variable in generate() captured by the callback lambda, giving each call its own isolated buffer. This matches the pattern already used by the ASR JNI layer.

### Test plan
CI


lucylq (approved):
Is this only a problem for the lora demo? :o

kirklandsign:
> Is this only a problem for the lora demo? :o

Everything

PR #17613, opened 2026-02-22
Add MmapUseMlockIgnoreErrors as default load mode when creating creat…

psiddh:
…e_text_llm_runner

Users can still override the default mode

### DIFF
--- a/extension/llm/runner/llm_runner_helper.h
+++ b/extension/llm/runner/llm_runner_helper.h
@@ -104,7 +104,8 @@ ET_EXPERIMENTAL std::unique_ptr<TextLLMRunner> create_text_llm_runner(
     std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
     std::optional<const std::string> data_path,
     float temperature = -1.0f,
-    const std::string& method_name = "forward");
+    const std::string& method_name = "forward",
+    Module::LoadMode load_mode = Module::LoadMode::MmapUseMlockIgnoreErrors);
 
 /**
  * @brief Creates a TextLLMRunner instance with dependency injection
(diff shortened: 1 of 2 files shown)


extension/llm/runner/llm_runner_helper.h
```diff
@@ -104,7 +104,8 @@ ET_EXPERIMENTAL std::unique_ptr<TextLLMRunner> create_text_llm_runner(
     std::unique_ptr<::tokenizers::Tokenizer> tokenizer,
     std::optional<const std::string> data_path,
     float temperature = -1.0f,
-    const std::string& method_name = "forward");
+    const std::string& method_name = "forward",
+    Module::LoadMode load_mode = Module::LoadMode::MmapUseMlockIgnoreErrors);
```
mergennachin, line 112: Why ignore errors variant over MmapUseMlock?

mergennachin, line 112: update the docblock

psiddh: The codebase only uses MmapUseMlockIgnoreErrors over MmapUseMlock for LLM runners. I feel that is the right default. 
  - MmapUseMlock: mlock failure → logs error, unmaps the pages, returns Error::NotSupported.
  The model fails to load entirely. (Stricter!)
  - MmapUseMlockIgnoreErrors: mlock failure → logs at Debug level and continues. The model
  loads via normal mmap, just without pages pinned in RAM.

 For LLM runners, a hard failure is almost never the right behavior. Failing to load the model at all will be 
   bad UX / functionality experience for the End User.  

  By using mmap-based loading in our LLM runners, we avoid loading the entire model into RAM
  upfront, which reduces peak memory usage and OOM risk. The MmapUseMlockIgnoreErrors variant
  additionally attempts to pin pages in memory for better inference latency, but gracefully
  falls back to standard mmap if the system can't support it, giving us the best of both
  worlds without hard failures.

GregoryComer: Realistically, I think that we're unlikely to actually be able to lock the entire LLM PTE on most systems, so using the base mmap might be better? The mlock ignore errors variant seems functionally fine, though. It'll just fall through to mmap pretty much 100% of the time.

GregoryComer (approved):

PR #17617, opened 2026-02-23
Arm backend: Raise quantization testing thresholds

AdrianLundell:
Raises frobenius norm and cosine similarity thresholds for comparison ops and 8aw4 ops. These type of ops will always have potentially large diffs from the original model for random inputs when quantized and they have been causing flaky test failures.


Sebastian-Larsson (approved):

zingo (approved):

AdrianLundell:
Fails unrelated

PR #17622, opened 2026-02-23
Fix QNN runner KV cache bitwidth in Android JNI

infil00p:
### Summary
The QNN runner was hardcoded to use Runner<uint16_t>, but all current Llama quantization recipes use annotate_kv_8bit for 8-bit KV cache. This mismatch caused the KV cache data to be misinterpreted, resulting in degenerate output (repetitive real words like "Nigeria Nigeria...") while the model otherwise ran correctly on the HTP NPU.

Authored with Claude

### Test plan
I built the AAR and used the AAR in the LlamaDemo in executorch-examples.  This does require QAIRT 2.43.0 to test, which is later than the last QAIRT used to build the last Executorch release.  This was tested on a OnePlus 15 running a Snapdragon 8 Elite Gen 5 (SM8850).

cc @kirklandsign @cbilgin @cccclai

### DIFF
--- a/extension/android/jni/jni_layer_llama.cpp
+++ b/extension/android/jni/jni_layer_llama.cpp
@@ -216,7 +216,7 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
           data_files_vector,
           executorch::extension::Module::LoadMode::MmapUseMlockIgnoreErrors);
       std::string decoder_model = "llama3"; // use llama3 for now
-      runner_ = std::make_unique<example::Runner<uint16_t>>( // QNN runner
+      runner_ = std::make_unique<example::Runner<uint8_t>>( // QNN runner (8-bit KV cache)
           std::move(module),
           decoder_model.c_str(),
           model_path->toStdString().c_str(),


extension/android/jni/jni_layer_llama.cpp
```diff
@@ -216,7 +216,7 @@ class ExecuTorchLlmJni : public facebook::jni::HybridClass<ExecuTorchLlmJni> {
           data_files_vector,
           executorch::extension::Module::LoadMode::MmapUseMlockIgnoreErrors);
       std::string decoder_model = "llama3"; // use llama3 for now
-      runner_ = std::make_unique<example::Runner<uint16_t>>( // QNN runner
+      runner_ = std::make_unique<example::Runner<uint8_t>>( // QNN runner (8-bit KV cache)
```
cccclai, line 219: I think we need a better way to handle this..I remember llama uses 16 bit kv cache and qwen uses 8 bit kv cache cc: @haowhsu-quic

haowhsu-quic: Yes, probably need a branch here to dispatch runner correctly.

abhinaykukkadapu: @haowhsu-quic Ccan we detect it by dynamically querying `get_kv_io_bit_width` from the model if the method exists and do something like (default to 8 bit):

```
example::KvBitWidth kv_bitwidth = example::KvBitWidth::kWidth8;
if (module->method_names()->count("get_kv_io_bit_width") > 0) {
  kv_bitwidth = static_cast<example::KvBitWidth>(
      module->get("get_kv_io_bit_width")
          .get()
          .toScalar()
          .to<int64_t>());
}

if (kv_bitwidth == example::KvBitWidth::kWidth16) {
  runner_ = std::make_unique<example::Runner<uint16_t>>(...)
} else {
  runner_ = std::make_unique<example::Runner<uint8_t>>(...)
}
```

nil-is-all: @haowhsu-quic there an update on dynamically querying the bit-width? Is there anything to recommend for the PR's author to do?

haowhsu-quic: Hi @abhinaykukkadapu , we actually do this in our runner (https://github.com/pytorch/executorch/blob/main/examples/qualcomm/oss_scripts/llama/qnn_llama_runner.cpp).
```cpp
example::KvBitWidth kv_bitwidth = example::KvBitWidth::kWidth8;
if (module->method_names()->count("get_kv_io_bit_width") > 0) {
  kv_bitwidth = static_cast<example::KvBitWidth>(
    module->get("get_kv_io_bit_width").get().toScalar().to<int64_t>());
}
```
Maybe PR's author can try to follow this apporach.

nil-is-all: @infil00p could you try this approach and request review once done?

nil-is-all: @infil00p bringing this up again

PR #17623, opened 2026-02-23
NXP backend: Add utilities for QAT testing

robert-kalmar:
### Summary
Add utilities to QAT model testing. 

### Test plan
N/A

cc @JakeStevens @digantdesai

### DIFF
--- a/backends/nxp/tests_models/executors.py
+++ b/backends/nxp/tests_models/executors.py
@@ -67,7 +67,7 @@ def _run_delegated_executorch_program(
     npu_results_dir,
     mocker,
     use_qat: bool = False,
-    train_fn: Callable | None = None,
+    train_fn: Callable[[torch.fx.GraphModule], None] | None = None,
 ) -> ExportedProgram:
     if len(input_spec) == 1:
         # Single input, use --dataset
@@ -144,7 +144,7 @@ def _run_non_delegated_executorch_program(
     input_spec,
     cpu_results_dir,
     use_qat: bool = False,
-    train_fn: Callable | None = None,
+    train_fn: Callable[[torch.fx.GraphModule], None] | None = None,
 ) -> ExportedProgram:
     if len(input_spec) == 1:
         # Single input, use --dataset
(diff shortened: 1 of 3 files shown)


backends/nxp/tests_models/executors.py
```diff
@@ -474,9 +495,96 @@ def convert_run_compare(
...
+    model: torch.nn.Module,
+    input_spec: list[ModelInputSpec] | tuple,
+    dlg_model_verifier: GraphVerifier,
+    train_fn: Callable,
+    dataset_creator=None,
+    output_comparator=None,
+    mocker: MockerFixture = None,
+):
+    """
+    Run provided program twice and compared it's results.
```
StrycekSimon: In case of some future push...
```suggestion
    Run provided program twice and compare it's results.
```

StrycekSimon (commented):
Looks legit. 👍

robert-kalmar:
Internal build link: https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH96

StrycekSimon:
Can we merge? I have [another PR](https://github.com/pytorch/executorch/pull/18527) blocked by this one.

backends/nxp/tests_models/utils.py
```diff
@@ -55,6 +59,7 @@ def to_quantized_edge_program(
     dataset_dir,
     delegate_to_npu=True,
     use_qat: bool = False,
+    train_fn: Callable | None = None,
```
MartinPavella: Nit: It would be useful if the input type of the `train_fn` was annotated too.

MartinPavella (approved):

backends/nxp/tests_models/utils.py
StrycekSimon: I fixed that.

backends/nxp/tests_models/executors.py
StrycekSimon: It is now correct.

PR #17629, opened 2026-02-23
Add support for shared CustomKVCache

lucylq:
### Summary

CustomKVCache is a mutable custom op. This means:
- export does not add any copies at functionalization, because it's a custom op and is a black box to export.
- this means it passes `inplace_lineage`, which determines if an output only has inplace ops.
- at `SpecPropPass`, the same TensorSpec (same python object) is given to the placeholder for the kvcache and the output spec, as it is completely inplace. 
- At memory planning, the placeholder is skipped (because it is mutable), and the output spec is assigned a mem_id. Because they share the same tensor spec, the placeholder is also assigned a mem_id.
- When we attempt to share mutable buffers, the placeholder already has a mem_id assigned.

This PR checks all TensorSpecs for mutability and makes sure mutable specs are not yielded to memory planning. A change to the verifier is also required, as the output node would have some specs with mem_id assigned, and some without (the mutable ones) which trigger verification failure.

### Test plan
Added unittest, confirm failing before PR and passing afterwards.

```
python -m pytest exir/tests/test_memory_planning.py -k test_custom_kv_cache_shared_buffers -xvs
```

### DIFF
--- a/exir/memory_planning.py
+++ b/exir/memory_planning.py
@@ -33,7 +33,6 @@ from executorch.exir.error import internal_assert, InternalError
 from executorch.exir.operator.convert import is_inplace_variant, is_out_variant
 from executorch.exir.schema import TensorShapeDynamism
 from executorch.exir.tensor import TensorSpec
-
 from torch import fx
 from torch.export.exported_program import (
     ConstantArgument,
@@ -204,7 +203,7 @@ class Verifier:
 
     def verify_graph_input_output(self) -> None:
         r"""
-        alloc_graph_input / alloc_graph_output indicas if memory for graph
+        alloc_graph_input / alloc_graph_output indicates if memory for graph
         input/output is allocated by the compiler. If not, the runtime will
         set them using buffers provided by users.
         """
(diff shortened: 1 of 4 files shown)


exir/memory_planning.py
```diff
@@ -410,7 +425,6 @@ def collect_specs_from_nodes(  # noqa: C901
     ignore_graph_input: bool = False,
     ignore_graph_output: bool = False,
     ignore_mutable_buffers: bool = False,
-    share_mutable_buffers: bool = False,
```
lucylq, line 413: rely on `ignore_mutable_buffers`, `share_mutable_buffers` is unused

exir/memory_planning.py
```diff
@@ -236,14 +235,24 @@ def verify_graph_input_output(self) -> None:
...
+        mutable_buffer_specs = _get_mutable_buffer_specs(
+            graph_module.graph.nodes, self.graph_signature
+        )
         for nd in graph_module.graph.nodes:
             if nd.op in check_list:
                 if not (specs := get_node_tensor_specs(nd)):
                     continue
                 if _is_mutable_buffer(nd, self.graph_signature):
                     continue
-                assert len(specs) > 0, "Expect tensor specs"
```
lucylq, line 245: this is checked in line 245

kimishpatel:
if shared buffer are mem planned in multimethod export, then do we need any special logic in rutnime?

lucylq:
> if shared buffer are mem planned in multimethod export, then do we need any special logic in rutnime?

@kimishpatel ah that's a good point. I think we also need some module-level changes, to make sure the same memory buffer is provided to each method.

JacobSzwejbka (approved):

JacobSzwejbka:
> if shared buffer are mem planned in multimethod export, then do we need any special logic in rutnime?

the arenas must be the same pointer. Which means as Lucy said we need to make module api changes. I had an old PR she is looking to revive

PR #17665, opened 2026-02-24
Qualcomm AI Engine Direct - Backend awareness quantizer

shewu-quic:
#### Summary:
- Add a file `backend_opinfo_adapter.py` which adapts `BackendOpInfo` from the QNN SDK for use with ExecuTorch
  - The `BackendOpInfo` API is supported starting from QNN SDK 2.41 and above.
  - The `BackendOpInfo` which is a pybind library contains a list of quantization constraints for each operator. These quantization constraints refer to the [operator definitions in QNN documents](https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-10/operations.html#backend-supplements).
- Refactor QnnQuantizer
  - Add support for backend-specific annotation by implementing lazy loading of `xxx_rules.py` in `registry_loader.py`.
  - Enable validation for quantization annotation with `BackendOpInfo`
    - It will bypass validation for QNN SDK versions 2.41 and earlier because the `BackendOpInfo` API is not supported in those versions..
    - Add the `backend` and `soc_model` parameters to allow configuration of `BackendOpInfo`.
    - Introduce a `strict` parameter. By default, it is enabled, causing the validation stage to `raise ValueError` if quantization constraints are not met. In this mode, all quantization constraints must be satisfied to fully delegate to the QNN Backend. If disabled, the process will only log warnings instead.
    - Validation items include:
      - Verify `htp_arch` for LPBQ support such as conv2d op.
      - Verify `htp_arch` for 16a16w support such as matmul op. - Ensure `SharedQuantizationSpec` is used for `is_math_invariant` such as view op. - Check `scale` and `zero_point` constraints for certain ops, such as requiring `scale = 1 / (q_max - q_min + 1)` and `zero_point = 0` for sigmoid op.
      - Confirm `qscheme` meets symmetric constraints.
      - Validate that the `dtype` of input and output is supported.

### Test plan
Successfully tested `test_qnn_delegate.py` and `static llama` with QNN version 2.41 and above.


shewu-quic:
Hi @cccclai,
This PR is to introduce `BackendOpInfo` API used to validate quantization configuration for each operator from QNN SDK.
Could you please take a look?
Thanks

cccclai:
@abhinaykukkadapu can you help reviewing?

abhinaykukkadapu:
@shewu-quic thanks for putting up the PR, sorry that i couldn't get to it sooner as it is a large one, please allow few days and i'm prioritizing this. Thanks.

backends/qualcomm/quantizer/quantizer.py
```diff
@@ -252,11 +267,28 @@ def __post_init__(self):
...
+        strict:
+          When enabled (default), the validation stage raises a ValueError if quantization constraints are not met.
+          In this mode, all quantization constraints must be satisfied to fully delegate to the QNN Backend.
+          When disabled, only warnings will be logged.
 
     Example usage:
-        quantizer = QnnQuantizer()
+        quantizer = QnnQuantizer(
+            backend=QnnExecuTorchBackendType.kHtpBackend,
+            soc_model=QcomChipset.SM8750
```
abhinaykukkadapu, line 290: @shewu-quic, Harsh mentioned that there is a context binary compatibility workstream within QCOM, aiming to decouple SOC model in the context binary in the coming phases, to avoid generating multiple ptes each targeting an SOC.

Introducing here would mean even after the delivery of the portable context binary, we still couple it with the SOC. Any thoughts?

backends/qualcomm/quantizer/quantizer.py
```diff
@@ -341,7 +425,7 @@ def _get_quant_config(self, node: torch.fx.Node) -> Optional[QuantizationConfig]
         op = node.target
         if isinstance(op, str):
             return
-        config = self._get_submodule_qconfig(node)
+        config = self.default_quant_config
```
abhinaykukkadapu: I see `_get_submodule_qconfig` is still retained but default is used for the config, is this intentional?

backends/qualcomm/quantizer/validators.py
```diff
@@ -0,0 +1,418 @@
...
+      - outputs: per-port normalized constraints
+    """
+
+    inputs: List[PortDatatypeConstraints] = field(default_factory=list)
+    outputs: List[PortDatatypeConstraints] = field(default_factory=list)
+
+
+class ConstraintCache:
+    def __init__(self):
+        self._store: Dict[str, NormalizedConstraints] = {}
```
abhinaykukkadapu: Nit: update type.

backends/qualcomm/quantizer/annotators/htp_rules.py
```diff
@@ -0,0 +1,1716 @@
...
+            annotate_fn = op_def.annotate
+            validate_fn = op_def.validate
+            rule = OpQuantRule(
+                aten_op=aten_op,
+                qnn_op=qnn_op,
+                annotate_fn=annotate_fn,
+                validate_fn=validate_fn,
+            )
+            _RULES[rule.aten_op] = rule
+        return rule
```
abhinaykukkadapu: should this return `op_def` instead of the rule?

backends/qualcomm/quantizer/validators.py
```diff
@@ -0,0 +1,418 @@
...
+                        valid &= _check_scale_and_offset(
+                            resolved_input_qspec,
+                            quant_constraint.scale,
+                            quant_constraint.offset,
+                        )
+                    break
+
+            if not valid:
+                logging.warning(
+                    f"Node ({input_node.name})'s input_qspec {input_qspec} failed to match constraints {quant_constraint}"
```
abhinaykukkadapu, line 403: We can may be let the user know which check failed, might be helpful for quick modification?

backends/qualcomm/quantizer/quantizer.py
shewu-quic: The soc_model is used for quantization constraint validation. Its primary purpose is to identify FP16 support, since some Qualcomm SoCs do not support FP for HTP.

I imagine the context binary compatibility feature in the QNN ExecuTorch workflow is as follows:

1. Given the target soc_models for the QNN quantizer, the graph is quantized and validated to ensure it is compatible.
2. Given the target soc_models specified in the compile_spec for the QNN partitioner and preprocessing steps, the graph is validated, compiled for each SoC model, and then packaged accordingly.

This is just my current understanding; the exact design can be finalized later.

backends/qualcomm/quantizer/quantizer.py
shewu-quic: Thanks for pointing out!

backends/qualcomm/quantizer/validators.py
shewu-quic: Thanks!

backends/qualcomm/quantizer/quantizer.py
abhinaykukkadapu: Thanks for the clarification. Makes sense, if the pte -> context binary has multiple compiled graphs for each soc. Curious, how do you envision this to evolve, should we accept a list of socs in that case?

shewu-quic: Yes, the graph may differ for each SoC, but they should share the same weights. Once the internal design reaches phase 3 or release, we can move forward with a PR to add this feature.

backends/qualcomm/quantizer/validators.py
shewu-quic: Sure, I think it would be helpful. Thanks!

abhinaykukkadapu (approved):

abhinaykukkadapu:
@shewu-quic can you please rebase, i will reimport the diff internally there are some CI failures. Thanks.

shewu-quic:
> @shewu-quic can you please rebase, i will reimport the diff internally there are some CI failures. Thanks.

Sure. I have rebased. Thanks.

abhinaykukkadapu:
> > @shewu-quic can you please rebase, i will reimport the diff internally there are some CI failures. Thanks.
> 
> Sure. I have rebased. Thanks.

Few CI failures with:

```
Undefined import [21]: Could not find a name `Q_ANNOTATION_KEY` defined in module `executorch.backends.qualcomm.quantizer.annotators`.
```

Can you please re-export `Q_ANNOTATION_KEY` in executorch/backends/qualcomm/quantizer/annotators/__init__.py?

shewu-quic:
Could you update the import to `from torchao.quantization.pt2e.quantizer.quantizer import Q_ANNOTATION_KEY`? 
I believe it should not rely on the ExecuTorch codebase.

abhinaykukkadapu:
> Could you update the import to `from torchao.quantization.pt2e.quantizer.quantizer import Q_ANNOTATION_KEY`? I believe it should not rely on the ExecuTorch codebase.

Yeah i think it is the right thing to do, there are multiple internal imports, will put a forward fix and will let you know if CI is cleared.

abhinaykukkadapu:
@shewu-quic CI is green. Merging.

PR #17670, opened 2026-02-24
NXP backend: added support for `aten.bmm`

novak-vaclav:
### Summary

adds support for `aten.bmm` operator

### Test plan

cc @robert-kalmar @JakeStevens @digantdesai @roman-janik-nxp @MartinPavella @StrycekSimon @jirioc

### DIFF
--- a/backends/arm/test/ops/test_eq.py
+++ b/backends/arm/test/ops/test_eq.py
@@ -146,8 +146,8 @@ def test_eq_scalar_tosa_INT_tensor(test_module):
         test_module().get_inputs(),
         Equal.aten_op_Tensor,
         Equal.exir_op,
-        frobenius_threshold=0.6,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
-        cosine_threshold=0.8,
+        frobenius_threshold=None,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
+        cosine_threshold=None,
     )
     pipeline.run()
 
@@ -159,8 +159,8 @@ def test_eq_scalar_tosa_INT(test_module):
         test_module().get_inputs(),
         Equal.aten_op_Tensor,
         Equal.exir_op,
-        frobenius_threshold=0.6,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
-        cosine_threshold=0.8,
+        frobenius_threshold=None,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
+        cosine_threshold=None,
     )
     pipeline.run()
 
--- a/backends/arm/test/ops/test_floor.py
+++ b/backends/arm/test/ops/test_floor.py
@@ -87,7 +87,7 @@ def test_floor_tosa_INT(test_data: input_t1):
         module.exir_op,
         atol=0.06,
         rtol=0.01,
-        frobenius_threshold=0.2,
+        frobenius_threshold=0.3,
     )
     pipeline.run()
 
--- a/backends/arm/test/ops/test_ge.py
+++ b/backends/arm/test/ops/test_ge.py
@@ -148,8 +148,8 @@ def test_ge_tensor_tosa_INT(test_module):
         test_module().get_inputs(),
         GreaterEqual.aten_op_tensor,
         GreaterEqual.exir_op,
-        frobenius_threshold=0.6,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
-        cosine_threshold=0.8,
+        frobenius_threshold=None,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
+        cosine_threshold=None,
     )
     pipeline.run()
 
@@ -161,8 +161,8 @@ def test_ge_scalar_tosa_INT(test_module):
         test_module().get_inputs(),
         GreaterEqual.aten_op_tensor,
         GreaterEqual.exir_op,
-        frobenius_threshold=0.6,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
-        cosine_threshold=0.8,
+        frobenius_threshold=None,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
+        cosine_threshold=None,
     )
     pipeline.run()
 
--- a/backends/arm/test/ops/test_gt.py
+++ b/backends/arm/test/ops/test_gt.py
@@ -149,8 +149,8 @@ def test_gt_tensor_tosa_INT(test_module):
         test_module().get_inputs(),
         Greater.aten_op_tensor,
         Greater.exir_op,
-        frobenius_threshold=0.6,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
-        cosine_threshold=0.8,
+        frobenius_threshold=None,  # Quantized comparisons with small diffs can be inaccurate, leading to large errors in unlucky cases.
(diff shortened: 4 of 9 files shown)


novak-vaclav:
In the last force-push, I just added the brief operator support description to `op-support.csv`

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,141 @@
...
+        pytest.param((4, 8, 16), (4, 16, 24), id="3D, more batches."),
+    ],
+)
+def test_convert_bmm__supported(mocker, input_shape_x1, input_shape_x2):
+    model = BatchMatMulModel()
+
+    converter_spy = mocker.spy(EdgeProgramToIRConverter, "convert_program")
+    delegated_ep = to_quantized_edge_program(
+        model, [input_shape_x1, input_shape_x2]
+    ).exported_program()
```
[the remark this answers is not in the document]
novak-vaclav: True, fixed.

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,94 @@
...
+    def _get_channels_last_shape(node: Node) -> list[int]:
+        input_shape = node.meta["val"].shape
+
+        if node.meta[NXP_NODE_FORMAT].is_channels_first():
+            input_shape = translator.apply_permutation_to(
+                input_shape,
+                translator.create_channels_first_to_channels_last_permutation(
+                    len(input_shape)
+                ),
+            )
```
MartinPavella: Just a hint for the future, the `translator.py` module contains the function `dims_to_channels_last()` which does this in 1 line.

backends/nxp/quantizer/patterns.py
```diff
@@ -298,6 +300,51 @@ def get_anchors(
...
+
+    def partition_types(self) -> list[torch.nn.Module]:
+        return [torch.ops.aten.bmm.default]
+
+    def _make_qspec(self):
+        observer = (
+            FakeQuantize.with_args(observer=MovingAverageMinMaxObserver)
+            if self.is_qat
+            else MinMaxObserver
+        )
```
MartinPavella: Why is this necessary? Can you please add a comment in the code too?

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,144 @@
...
+from executorch.exir.dialects._ops import ops as exir_ops
+
+
+@pytest.fixture(autouse=True)
+def reseed_model_per_test_run():
+    torch.manual_seed(23)
+    np.random.seed(23)
+
+
+# noinspection PyProtectedMember
```
MartinPavella: Nit: Redundant.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@pytest.mark.parametrize(
+    "input_shape_x1, input_shape_x2",
+    [
+        pytest.param((1, 7, 16), (1, 16, 24), id="3D, x1_C not divisible by NUM_MACS."),
+        pytest.param(
+            (1, 8, 7), (1, 7, 24), id="3D, x1_W (and x2_C) not divisible by NUM_MACS."
+        ),
+        pytest.param((1, 8, 16), (1, 16, 7), id="3D, x2_W not divisible by NUM_MACS."),
```
MartinPavella: Nit:
Since your tested model only contains the `bmm`, the input/output tensors will not be transposed to channels last (because there is no need to), so the Neutron Converter will see the same shapes as ExecuTorch. Therefore, from the perspective of the `BMMConverter` restrictions, the shapes here are `NWC`, so the descriptions of your test cases are not quite correct.

Sine both C and W must be multiples of num_macs, it has no effect. But if Neutron suddenly only required the channels to be a multiple of num_macs, these descriptions could be very misleading. (a test described as `"3D, x2_W not divisible by NUM_MACS."` would fail, but `"3D, x1_C not divisible by NUM_MACS."` would pass)

MartinPavella (approved):
Well done 👍🏻 
Feel free to rebase onto main, so the linting checks can run.

backends/nxp/quantizer/patterns.py
novak-vaclav: I saw this construct in a different pattern but after consultation with @StrycekSimon I realized this might be a mistake. I used the default from `neutron_quantizer` but with different `qscheme` because of requirements of Neutron.
Added an informative comment.

novak-vaclav:
Fixed the issues mentioned by @MartinPavella and rebased onto main.

backends/nxp/quantizer/patterns.py
```diff
@@ -298,6 +301,58 @@ def get_anchors(
...
+        self.neutron_target_info = (
+            self.neutron_quantizer.neutron_target_spec.neutron_target_info
+        )
+
+    def partition_types(self) -> list[torch.nn.Module]:
+        return [torch.ops.aten.bmm.default]
+
+    def _make_qspec(self):
+        # Because of Neutron restrictions, a diffent `qscheme` needs to be used. Otherwise everything
+        # is a default taken from `neutron_quantizer`.
```
MartinPavella: What are these restrictions exactly? How come no other operator requires this?

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
MartinPavella: It's still here.

MartinPavella:
When you implement changes based on comments and you need to rebase onto main, please push separately. When you push just once, I cannot use the diff efficiently because it contains all the commits you rebased onto.

backends/nxp/quantizer/patterns.py
novak-vaclav: When I use the `BatchMatMul` from `batch_mat_mul_options`, I am asked to supply a `boolean` parameter `asymmetric_quantize_inputs`.
If I understood the parameter correctly, if set to `False`, the input tensors must be quantized symmetrically. Based on my research, if set to `False`, the computation of the operator is supposed to be faster. Neutron itself recomputes the tensors if they are not quantized symmetrically (`OperatorC/BatchMatMulPlugin.cpp:46`).
That is why I need to override the default `qpec` and set `qscheme` to be symmetric.

MartinPavella: OK, makes sense. I was unsure mainly because of the 3 new imports, since we already had operators what required a specific zero point. So if I understand correctly, this case is unique in that the nodes are **per-tensor** quantized (not per-channel).

Side note: Have you checked whether per-channels quantization is supported for BatchMatMul on Neutron? I believe only FullyConnected and Conv support it, but just in case.

novak-vaclav: Well for the `MovingAverageMinMaxObserver` observer, only `per_tensor_symmetric` and `per_tensor_affine` is supported so that is why I went with `per_tensor_symmetric`.
Where would I check if per-channels quantization is supported on Neutron?

novak-vaclav:
Currently on hold, inquiring about the meaning of the negative zero point @MartinPavella

novak-vaclav:
@MartinPavella please run the Github tests.
Internal checks [passed](https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH97-7)

novak-vaclav:
I pushed to the branch without committing the work first and the PR closed itself auto-magically.
The new PR is [here](https://github.com/pytorch/executorch/pull/17818)

PR #17671, opened 2026-02-24
Metal backend: commitAndContinue dispatch pipelining

mergennachin:
The high‑level idea is to keep the GPU busy by overlapping driver scheduling with GPU execution. Instead of encoding hundreds of tiny kernels into a single command buffer and then waiting, we periodically “flush” the command buffer with commitAndContinue. That lets the driver begin preparing the next batch of work while the GPU is still executing the current batch, reducing idle gaps between dispatches.

To do that safely, we count dispatches and commit every N of them. The commit interval is tuned to the GPU class (or overridden by an env var), and the stream only does this when commit‑and‑continue mode is enabled. This doesn’t change kernel code or math; it just improves command submission cadence so token latency drops without changing model outputs.

```

    CPU: [====== encode all ======]
    GPU:                            [====== execute all ======]
         |<--------------------- 125ms ---------------------->|


  With commitAndContinue (flush every N dispatches):

    CPU: [enc][enc][enc][enc]...
    GPU:      [gpu][gpu][gpu][gpu]...
              ↑ overlap
         |<------------ 85ms ------------->|
```

Implementation wise: 

Periodically call [commandBuffer commitAndContinue] every N dispatches
so the Metal driver can prepare batch N+1 while the GPU executes batch N.
This overlaps driver scheduling overhead with GPU execution, reducing
Voxtral per-token latency from ~125ms to ~85ms on M1 Pro (7.5 -> 10.3 tok/s).

Flush interval is set automatically based on GPU architecture suffix
(matching MLX's approach): iPhone=20, base/Pro=40, Max=50, Ultra=50.
Override via ET_METAL_FLUSH_INTERVAL env var.

### DIFF
--- a/backends/apple/metal/runtime/metal_backend.cpp
+++ b/backends/apple/metal/runtime/metal_backend.cpp
@@ -14,6 +14,7 @@
 #include <unistd.h>
 #include <chrono>
 #include <cstdio>
+#include <cstdlib>
 
 #include <filesystem>
 #include <fstream>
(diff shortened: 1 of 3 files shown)


backends/apple/metal/runtime/metal_backend.cpp
```diff
@@ -394,6 +395,23 @@ class ET_EXPERIMENTAL MetalBackend final
...
+    std::call_once(flush_interval_flag, [] {
+      const char* env = std::getenv("ET_METAL_FLUSH_INTERVAL");
+      if (env) {
+        try {
+          int val = std::stoi(env);
+          if (val >= 0) {
+            getCurrentMetalStream()->setFlushInterval(val);
+          }
+        } catch (const std::exception&) {
+          // Invalid ET_METAL_FLUSH_INTERVAL value; use default.
```
metascroy: Do you want to raise here instead of silently falling back?

metascroy (approved):
Nice!

manuelcandales (approved):

PR #17676, opened 2026-02-24
Update HiFi1 to use nnlib (#17676)

mcremon-meta:
Summary:

As titled. Was falling back on xtensa.
Adds a few fixes to some operators to enable the build for all cases.

Differential Revision: D94179206

### DIFF
--- a/backends/cadence/hifi/operators/op_quantized_linear_out.cpp
+++ b/backends/cadence/hifi/operators/op_quantized_linear_out.cpp
@@ -13,7 +13,6 @@
 #include <optional>
 
 #include <xa_nnlib_api.h>
-#include <xtensa/tie/xt_datacache.h>
 
 #include <executorch/backends/cadence/generic/operators/op_quantized_linear.h>
 #include <executorch/backends/cadence/hifi/kernels/kernels.h>


mcremon-meta:
@dijopaul @cad-audio can you take a look please? :)

backends/cadence/hifi/third-party/nnlib/xa_nn_elm_minimum_maximum_f32.c
```diff
@@ -23,6 +23,19 @@
 #include "xa_nnlib_common_fpu.h"
 #include "xa_nnlib_err_chk.h"
 
+// HiFi1 compatibility: int32_rtor_xtbool2 is not available on HiFi1
+#ifndef int32_rtor_xtbool2
```
dijopaul: I think this is not needed. Can remove this line
xtbool2 con = int32_rtor_xtbool2(0x00000003);
the vaiable not used

backends/cadence/hifi/third-party/nnlib/xa_nn_greater_lesser_equal_f32.c
```diff
@@ -4,6 +4,25 @@
 #include "xa_nnlib_err_chk.h"
 #include "xa_nnlib_kernels_api.h"
 
+// HiFi1 compatibility: AE_MOVAB is not available on HiFi1
+#ifndef AE_MOVAB
+static inline unsigned char AE_MOVAB(xtbool b) {
```
dijopaul: Can use the below code

#ifndef AE_MOVAB
#ifdef AE_MOVAB1
#define AE_MOVAB AE_MOVAB1
#else
#define AE_MOVAB(x)  (x)  
#endif
#endif


#ifndef AE_MOVAB2
#define AE_MOVAB2(x) (x)
#endif

dijopaul:
> @dijopaul @cad-audio can you take a look please? :)

The root cause here is that we are using local implementations outside nnlib (which is hifi4 specific). The latest nnlib has all these implementation inside including hifi1 support. So better solution is to upgrade to latest nnlib and avoid these local files. We will make a PR for the same. This may involve some changes in API calls as well due to minor changes in API names.

mcremon-meta:
> > @dijopaul @cad-audio can you take a look please? :)
> 
> The root cause here is that we are using local implementations outside nnlib (which is hifi4 specific). The latest nnlib has all these implementation inside including hifi1 support. So better solution is to upgrade to latest nnlib and avoid these local files. We will make a PR for the same. This may involve some changes in API calls as well due to minor changes in API names.

I see, makes sense. Are some of those changes on our side, or if you open a PR to update the lib, we get things for free on our end? If that's the case, we can wait a little bit

cad-audio:
> > > @dijopaul @cad-audio can you take a look please? :)
> > 
> > 
> > The root cause here is that we are using local implementations outside nnlib (which is hifi4 specific). The latest nnlib has all these implementation inside including hifi1 support. So better solution is to upgrade to latest nnlib and avoid these local files. We will make a PR for the same. This may involve some changes in API calls as well due to minor changes in API names.
> 
> I see, makes sense. Are some of those changes on our side, or if you open a PR to update the lib, we get things for free on our end? If that's the case, we can wait a little bit

We will create a PR for nnlib update after verifying the build on HiFi1. We will target this on this week

mcremon-meta:
> > > > @dijopaul @cad-audio can you take a look please? :)
> > > 
> > > 
> > > The root cause here is that we are using local implementations outside nnlib (which is hifi4 specific). The latest nnlib has all these implementation inside including hifi1 support. So better solution is to upgrade to latest nnlib and avoid these local files. We will make a PR for the same. This may involve some changes in API calls as well due to minor changes in API names.
> > 
> > 
> > I see, makes sense. Are some of those changes on our side, or if you open a PR to update the lib, we get things for free on our end? If that's the case, we can wait a little bit
> 
> We will create a PR for nnlib update after verifying the build on HiFi1. We will target this on this week

Sounds good! Thanks :)

aliafzal (approved):

PR #17710, opened 2026-02-25
[ET-VK] Add aten.index.Tensor op and tests

SS-JIA:
* __->__ #17710
* #17709
* #17708
* #17707

Implements `aten.index.Tensor` for the Vulkan backend, supporting 1D
self tensors with exactly one non-None index tensor. Includes buffer
and texture GLSL shaders, C++ operator registration, and correctness
tests.

Also extends the op test code generators to handle
`c10::List<::std::optional<at::Tensor>>` (`Tensor?[]` in ATen), which
is the C++ type for the `indices` argument of `aten.index.Tensor`.

Differential Revision: [D94364638](https://our.internmc.facebook.com/intern/diff/D94364638/)


backends/vulkan/runtime/graph/ops/impl/IndexTensor.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void index_tensor(ComputeGraph& graph, const std::vector<ValueRef>& args) {
+  ValueRef self = args[0];
+  ValueRef indices_list_ref = args[1];
+  ValueRef out = args[2];
+
+  ValueListPtr indices_list = graph.get_value_list(indices_list_ref);
+  VK_CHECK_COND(
+      indices_list->size() == 1,
+      "index.Tensor: only one index tensor is supported");
```
manuelcandales, line 69: How about Bool vs Int? I think you are implementing the integer case only, right? If that's the case, also add a `VK_CHECK_COND` saying that only integer index is supported.

manuelcandales (approved):

PR #17717, opened 2026-02-25
Download QNN SDK outside of CMake for wheel builds

larryliu0820:
Move the Qualcomm QNN SDK download from cmake configure time to the pre-build stage in CI. setup.py now auto-detects QNN_SDK_ROOT and enables EXECUTORCH_BUILD_QNN, matching the existing CUDA pattern. The cmake download logic is preserved as a last-resort fallback.

Re-enable the test-qnn-wheel-packages-linux CI job and add a QnnBackend registration check to the Linux wheel smoke test.

### DIFF
--- a/tools/cmake/preset/pybind.cmake
+++ b/tools/cmake/preset/pybind.cmake
@@ -37,7 +37,7 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set_overridable_option(EXECUTORCH_BUILD_EXTENSION_LLM_RUNNER ON)
   set_overridable_option(EXECUTORCH_BUILD_EXTENSION_LLM ON)
   if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|amd64|i.86)$")
-    set_overridable_option(EXECUTORCH_BUILD_QNN OFF)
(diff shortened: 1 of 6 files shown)


tools/cmake/preset/pybind.cmake
```diff
@@ -36,9 +36,8 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set_overridable_option(EXECUTORCH_BUILD_EXTENSION_TRAINING ON)
   set_overridable_option(EXECUTORCH_BUILD_EXTENSION_LLM_RUNNER ON)
   set_overridable_option(EXECUTORCH_BUILD_EXTENSION_LLM ON)
-  if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|amd64|i.86)$")
-    set_overridable_option(EXECUTORCH_BUILD_QNN OFF)
```
abhinaykukkadapu, line 40: For building wheels this is fine, but can we actually keep this option and turn this ON, it would help with building from source using `./install_executorch.sh`, AOT can just work without explicit SDK download ?

abhinaykukkadapu (approved):

PR #17719, opened 2026-02-25
Share state in module

lucylq:
### Summary

Taken from [D82329513](https://www.internalfb.com/diff/D82329513)

Allow Module to share activation memory when `share_memory_arenas=True`. 

In a PTE file, each method may have:
mem_id=1: activation memory
mem_id=2: shared memory (usually for shared state)
mem_id>2: custom memory: remains unshared

These are specified in `program.execution_plan[method_index].non_const_buffer_sizes`. Tensors with `mem_id=2` point to the same memory offset across different methods.

In Module.cpp, freshly allocated memory is provided to each method for activation space. This PR creates a shared_buffer and passes it into to each method instead of allocating fresh memory, when `share_memory_arenas=True`. This is only for mem_id=1,2. For mem_id>2, we allocate fresh memory, as this is likely a custom memory plan.

**NOTE**
1. Regular activation memory is also shared (mem_id=1). The largest activation memory for each method is used.

2. I believe this is safe with the existing concurrent execution. In `TestConcurrentExecutionWithSharedProgram`, it's the (immutable) program that is shared. N separate Modules hold a shared_ptr to the same program. So each module has its own memory buffers etc., and sharing is local to the Module and not across threads. I don't think concurrency within a single Module/between methods is supported.
3. Running multiple methods concurrently in the same Module is unsafe, and this PR makes it even more unsafe.

**RISKS**
1. Data invalidation across methods with shared memory
```
  auto m1 = module.execute("m1", {x1});
  auto m2 = module.execute("m2", {x2});
```
m1 and m2 will point to the same output memory buffer when sharing memory (activation memory is shared); m1 must be copied to a separate buffer, otherwise it will store the result of m2.

### Test plan
```
cmake .. \
-DEXECUTORCH_BUILD_TESTS=ON \
-DEXECUTORCH_BUILD_XNNPACK=ON \
-DEXECUTORCH_BUILD_EXTENSION_NAMED_DATA_MAP=ON \
-DEXECUTORCH_BUILD_EXTENSION_DATA_LOADER=ON \
-DEXECUTORCH_BUILD_EXTENSION_TENSOR=ON \
-DEXECUTORCH_BUILD_KERNELS_PORTABLE=ON

cmake --build . --target extension_module_test

ctest -R extension_module_test --output-on-failure 
```

### DIFF
--- a/extension/module/module.cpp
+++ b/extension/module/module.cpp
@@ -80,7 +80,8 @@ Module::Module(
     const LoadMode load_mode,
     std::unique_ptr<runtime::EventTracer> event_tracer,
     std::unique_ptr<runtime::MemoryAllocator> memory_allocator,
-    std::unique_ptr<runtime::MemoryAllocator> temp_allocator)
+    std::unique_ptr<runtime::MemoryAllocator> temp_allocator,
+    bool share_memory_arenas)
     : file_path_(file_path),
       load_mode_(load_mode),
       memory_allocator_(
@@ -89,7 +90,8 @@ Module::Module(
       temp_allocator_(
           temp_allocator ? std::move(temp_allocator)
                          : std::make_unique<MallocMemoryAllocator>()),
-      event_tracer_(std::move(event_tracer)) {
+      event_tracer_(std::move(event_tracer)),
+      share_memory_arenas_(share_memory_arenas) {
   runtime::runtime_init();
 }
 
@@ -99,7 +101,8 @@ Module::Module(
     const LoadMode load_mode,
     std::unique_ptr<runtime::EventTracer> event_tracer,
     std::unique_ptr<runtime::MemoryAllocator> memory_allocator,
-    std::unique_ptr<runtime::MemoryAllocator> temp_allocator)
+    std::unique_ptr<runtime::MemoryAllocator> temp_allocator,
+    bool share_memory_arenas)
     : file_path_(file_path),
       load_mode_(load_mode),
       memory_allocator_(
@@ -108,7 +111,8 @@ Module::Module(
       temp_allocator_(
           temp_allocator ? std::move(temp_allocator)
                          : std::make_unique<MallocMemoryAllocator>()),
-      event_tracer_(std::move(event_tracer)) {
+      event_tracer_(std::move(event_tracer)),
+      share_memory_arenas_(share_memory_arenas) {
   if (!data_map_path.empty()) {
     data_files_.push_back(data_map_path);
   }
@@ -121,7 +125,8 @@ Module::Module(
     const LoadMode load_mode,
     std::unique_ptr<runtime::EventTracer> event_tracer,
     std::unique_ptr<runtime::MemoryAllocator> memory_allocator,
-    std::unique_ptr<runtime::MemoryAllocator> temp_allocator)
+    std::unique_ptr<runtime::MemoryAllocator> temp_allocator,
+    bool share_memory_arenas)
     : file_path_(file_path),
       data_files_(std::move(data_files)),
       load_mode_(load_mode),
@@ -131,7 +136,8 @@ Module::Module(
       temp_allocator_(
           temp_allocator ? std::move(temp_allocator)
                          : std::make_unique<MallocMemoryAllocator>()),
-      event_tracer_(std::move(event_tracer)) {
+      event_tracer_(std::move(event_tracer)),
+      share_memory_arenas_(share_memory_arenas) {
   runtime::runtime_init();
 }
 
@@ -140,7 +146,8 @@ Module::Module(
     std::unique_ptr<runtime::MemoryAllocator> memory_allocator,
     std::unique_ptr<runtime::MemoryAllocator> temp_allocator,
     std::unique_ptr<runtime::EventTracer> event_tracer,
-    std::unique_ptr<runtime::DataLoader> data_map_loader)
+    std::unique_ptr<runtime::DataLoader> data_map_loader,
+    bool share_memory_arenas)
     : data_loader_(std::move(data_loader)),
       memory_allocator_(
           memory_allocator ? std::move(memory_allocator)
@@ -148,7 +155,8 @@ Module::Module(
       temp_allocator_(
           temp_allocator ? std::move(temp_allocator)
                          : std::make_unique<MallocMemoryAllocator>()),
-      event_tracer_(std::move(event_tracer)) {
+      event_tracer_(std::move(event_tracer)),
+      share_memory_arenas_(share_memory_arenas) {
   if (data_map_loader) {
     data_map_loaders_.push_back(std::move(data_map_loader));
   }
@@ -160,7 +168,8 @@ Module::Module(
     std::unique_ptr<runtime::MemoryAllocator> memory_allocator,
     std::unique_ptr<runtime::MemoryAllocator> temp_allocator,
     std::unique_ptr<runtime::EventTracer> event_tracer,
-    std::unique_ptr<runtime::DataLoader> data_map_loader)
+    std::unique_ptr<runtime::DataLoader> data_map_loader,
+    bool share_memory_arenas)
     : program_(std::move(program)),
       memory_allocator_(
           memory_allocator ? std::move(memory_allocator)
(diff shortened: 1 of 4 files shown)


extension/module/module.cpp
```diff
@@ -263,29 +323,21 @@ runtime::Error Module::load_method(
...
+        auto sizes_res = get_mem_planned_buffer_sizes(method_name);
+        ET_CHECK_OK_OR_RETURN_ERROR(sizes_res.error());
+        method_holder.planned_memory = make_planned_memory(sizes_res.get());
+      } else {
+        if (!shared_planned_memory_) {
+          auto max_res = get_max_mem_planned_buffer_sizes();
+          ET_CHECK_OK_OR_RETURN_ERROR(max_res.error());
+          shared_planned_memory_ = make_planned_memory(max_res.get());
+        }
+        method_holder.planned_memory = shared_planned_memory_;
```
kimishpatel: This does mean that methods cannot be invoked in parallel as they may overwrite each others' arena. Document that explicitly

kimishpatel: Also if AOT memory planning accounted for it then should we not assert that the planned memory has the same mem_id or something for them to be shareable?

lucylq: @kimishpatel 

> This does mean that methods cannot be invoked in parallel as they may overwrite each others' arena. Document that explicitly

Yeah that's right. This isn't a currently supported feature (unsafe to run multiple methods in separate threads within a Module, even without sharing), but sharing makes it harder to support. I'll add an extra comment in module.h

> Also if AOT memory planning accounted for it then should we not assert that the planned memory has the same mem_id or something for them to be shareable?

Initially I was thinking we only share physical buffers when mem_id=2 (buffer marked as shareable AoT).

However, @JacobSzwejbka's original diff [D82329513](https://www.internalfb.com/diff/D82329513), shares both and I don't see any issues with that besides output tensors needing to be copied into permanent memory after running each method, which I think is OK (users have to do this when running the same method twice already).

kimishpatel: i was just wondering if we should enforce that only buffers that are on the same mem_id can share memory

lucylq: Yeah I was wondering that too. It seems fine to share all of it, and we get some memory savings as well, wdyt?

kimishpatel: i think mem_id has specific meaning so I think asserting for that is good.

lucylq: `mem_id=2`: mutable buffers with the same fqn will point to the same memory offset. Mostly useful when methods share state (must be backed by same buffer).

`mem_id=1`: overwritten with each execute, no guarantees between different methods/executions. Does not need to be backed by the same buffer, but would provide memory savings if so.

@kimishpatel  do you mean assert as in, only share buffer for mem_id=2?

kimishpatel: no i didnt mean that. actually i wasnt aware of, if true, that mem_id 2 is used for mutable buffers.

What i simply meant was
method 1: mutable buffer->plans... [mem_id = 2, mem_id = 3]...
method 2: mutable buffer -> plans ..[mem_id = 2, med_id =4]...

In this case the second buffer should not share memory.

But you should correct me if my understanding is wrong. I just wanted to highlight that custom memory planning algorithms can assign memory differently. However, if for mutable buffers we always know that their mem_ids have to be the same, then ignore my comment

JacobSzwejbka: >that mem_id 2 is used for mutable buffers. 

Only with the default plan with sharing turned on

> In this case the second buffer should not share memory.

If you need this fine grained control over what arenas are being shared and what are not then you should not use the module api. Use the lower level apis. Module really should only expose a binary choice: one set of buffers shared over all entry points, or each entry point gets unique buffers. 


>This does mean that methods cannot be invoked in parallel as they may overwrite each others' arena. Document that explicitly

Module is not thread safe. You shouldn't be doing this anyway.

kimishpatel: > If you need this fine grained control over what arenas are being shared and what are not then you should not use the module api. Use the lower level apis. Module really should only expose a binary choice: one set of buffers shared over all entry points, or each entry point gets unique buffers.


@JacobSzwejbka I dont need the control. What i am trying to say whether as Module API you should assert against such behavior or not. Right now there is not such checking for mem_ids when sharing buffer memory. If you say that this part of the Module API, that it doesnt care for mem_ids, then thats fine.

JacobSzwejbka: The runtime has no idea if a given tensor is a mutable buffer or not so it cannot provide these guarantees. That information is lost at serialization

kimishpatel: mem_id information is lost at serialization? I have a feeling that either I am not understanding what you are saying or maybe you are not. lol

kimishpatel (approved):
was trying to get conclusion on med_id stuff, but doesnt seem to be. So approving pending that reconciliation.

extension/module/module.cpp
lucylq: Hmn, `mem_id` information isn't lost at serialization, it's part of the tensor object..

I guess we can have mutable buffers, and non-mutable buffers in mem_id=1, and the runtime doesn't know.

But @kimishpatel do you mean, if we have mem_id > 2, we shouldn't share them by default because it's likely from a custom memory plan? I think that's reasonable (and I can make that change), as we expect certain behavior for mem_id=1 and 2, but not beyond that.

extension/module/module.cpp
```diff
@@ -279,29 +364,29 @@ runtime::Error Module::load_method(
...
+      } else {
+        auto sizes_res = get_mem_planned_buffer_sizes(method_name);
+        ET_CHECK_OK_OR_RETURN_ERROR(sizes_res.error());
+        auto& sizes = sizes_res.get();
+        if (shared_arenas_.empty()) {
+          auto max_res = get_max_mem_planned_buffer_sizes();
+          ET_CHECK_OK_OR_RETURN_ERROR(max_res.error());
+          auto& max_sizes = max_res.get();
+          // Only share for mem_id=1,2.
+          for (size_t i = 0; i < std::min(max_sizes.size(), (size_t)2); i++) {
```
lucylq: cc @kimishpatel, @JacobSzwejbka 

Limit shared arena to mem_id=1,2

extension/module/module.cpp
```diff
@@ -264,6 +275,80 @@ runtime::Result<std::unordered_set<std::string>> Module::method_names() {
...
+    const std::vector<std::vector<uint8_t>>& shared_arenas) {
+  auto planned = std::make_unique<PlannedMemory>();
+  planned->planned_buffers.reserve(buffer_sizes.size());
+  planned->planned_spans.reserve(buffer_sizes.size());
+  for (size_t i = 0; i < buffer_sizes.size(); i++) {
+    if (i < shared_arenas.size()) {
+      planned->planned_buffers.emplace_back();
+      planned->planned_spans.emplace_back(
+          shared_arenas[i].data(), shared_arenas[i].size());
+    } else {
```
lucylq, line 305: Here, create fresh buffer if mem_id>2. shared_arenas is at most size 2.

extension/module/module.cpp
lucylq: Resolved; only share for mem_id=1,2, and provide fresh buffers for mem_id>2

extension/module/module.cpp
```diff
@@ -279,29 +366,30 @@ runtime::Error Module::load_method(
...
+      } else {
+        auto sizes_res = get_mem_planned_buffer_sizes(method_name);
+        ET_CHECK_OK_OR_RETURN_ERROR(sizes_res.error());
+        auto& sizes = sizes_res.get();
+        if (shared_arenas_.empty()) {
+          auto max_res = get_max_mem_planned_buffer_sizes();
+          ET_CHECK_OK_OR_RETURN_ERROR(max_res.error());
+          auto& max_sizes = max_res.get();
+          // Only share for mem_id=1,2.
+          size_t shared = (max_sizes.size() > 2) ? 2 : max_sizes.size();
```
lucylq, line 381: cap max_sizes to 2, for mem_id=1,2

extension/module/module.cpp
```diff
@@ -264,6 +275,82 @@ runtime::Result<std::unordered_set<std::string>> Module::method_names() {
...
+      std::make_unique<runtime::HierarchicalAllocator>(runtime::Span(
+          planned->planned_spans.data(), planned->planned_spans.size()));
+  return planned;
+}
+
+std::unique_ptr<Module::PlannedMemory>
+Module::make_planned_memory_with_shared_arenas(
+    const std::vector<size_t>& buffer_sizes,
+    std::vector<std::vector<uint8_t>>& shared_arenas) {
+  auto planned = std::make_unique<PlannedMemory>();
```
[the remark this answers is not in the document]
lucylq: The shared_arena itself is mutable, so we don't want to pass a const span in

PR #17721, opened 2026-02-25
Move large test assets from APK to adb push

kirklandsign:
### Summary
vit_b_16.pte (~330MB) and stories.pte (~29MB) are no longer bundled in the androidTest APK. Instead they are uploaded as a separate CI artifact and pushed to /data/local/tmp/executorch/ via adb before tests run. TestFileUtils.prepareTestFile() checks the push directory first and falls back to APK resources, so both CI and local workflows remain functional. This reduces the APK from ~361MB to ~2MB.

### Test plan
CI


GregoryComer (approved):

kirklandsign:
@claude review

PR #17725, opened 2026-02-26
Increase model_pte_program_size limit in test script (3200 b)

psiddh:


rascani (approved):
Please add a description about why this is needed.

psiddh:
> Please add a description about why this is needed.

Something doesn't add up, Not sure why the model pte size has increased, If anything the PR : https://github.com/pytorch/executorch/pull/17075  should reduce the pte size as we are removing this block entirely

```
if args.target != "vgf" and not args.direct_drive:
    edge = transform_for_cortex_m_backend(edge, args)
```

psiddh:
The most likely culprit is the removal of transform_for_cortex_m_backend() from the Ethos-U path. On main, ReplaceQuantNodesPass runs on the Ethos-U add model. Even if it doesn't find matching quant nodes to replace (because they're consumed by the delegate), the pass itself may modify the graph metadata ? for example, by running the pass infrastructure which can compact or normalize node metadata. Without that pass running, the graph serializes slightly differently (larger) --> This is one hypothesis.

Alternatively, the new imports (CortexMPassManager, CortexMQuantizer) could be registering additional custom ops in the global operator registry (via import executorch.backends.cortex_m.ops.operators inside those modules), which get serialized into the PTE's operator table even if they're not used. --> This is second hypothesis.

Either ways it does look like the other PR does cause the CI job to fail. I will land this for now to unblock the other PR, (Happy to revert this if the root cause is something else related)

PR #17750, opened 2026-02-26
Fix output ordering in lowered modules to prevent spec violation error

metascroy:
When the ExecuTorch partitioner creates submodules from a partitioned graph, _get_new_signature builds output specs by iterating the submodule's output tuple in graph order. This can produce output specs where BUFFER_MUTATION and USER_OUTPUT kinds are interleaved. The ExportedProgram verifier expects all mutation outputs to precede user outputs ([buffer_mutations..., user_outputs...]), and a misordered tuple causes a SpecViolationError.

This PR adds arrange_graph_outputs, which reorders the submodule's output tuple and output specs so that all mutation outputs (BUFFER_MUTATION, USER_INPUT_MUTATION) come before USER_OUTPUT outputs. It also remaps getitem indices on the parent graph's call_module node so the parent continues to extract the correct values after reordering.

Changes
exir/lowered_backend_module.py

Added arrange_graph_outputs() function that:

* Reorders the submodule graph's output node to put mutation outputs first
* Reorders output_specs in-place to match
* Remaps getitem indices in the parent graph

Test plan:

Existing CI + new unit test

### DIFF
--- a/exir/backend/test/test_lowered_backend_module.py
+++ b/exir/backend/test/test_lowered_backend_module.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import operator
 import unittest
 
 import executorch.exir.tests.models as models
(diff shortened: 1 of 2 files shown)


metascroy:
@JacobSzwejbka any concerns here?

JacobSzwejbka:
> @JacobSzwejbka any concerns here?

Does this reordering happen after partitioning but before preprocess? My thought is basically the preprocess blob must respect the order it saw from the partitioner so we cant change it after the blob is created.

JacobSzwejbka (approved):

PR #17752, opened 2026-02-26
Introduce CSE pass to ExecuTorch

metascroy:
This introduces a CSE pass to ExecuTorch, which eliminates common subexpressions that occur in exported programs.

This pass was first developed as part of the MLX delegate (https://github.com/pytorch/executorch/pull/16718) to optimize transformers, but I'm introducing it to ExecuTorch more generally because I believe it could benefit many other backends.

Examples of common subexpressions that occur in transformers:

* Repeated mask constructions per layer (only needs to be done once)
* Repeated extraction of symints from 1d tensors for cache position (emits .item calls, which cause tensor materialization)

This pass eliminates these inefficiencies without having to rewrite the model.

### DIFF
--- a/exir/passes/BUCK
+++ b/exir/passes/BUCK
@@ -419,6 +419,18 @@ fbcode_target(_kind = runtime.python_library,
 )
 
 fbcode_target(_kind = runtime.python_library,
+    name = "cse_pass",
+    srcs = [
+        "cse_pass.py",
+    ],
+    deps = [
+        "//caffe2:torch",
+        "//executorch/exir:pass_base",
+        "//executorch/exir/dialects/edge:lib",
+    ],
+)
+
+fbcode_target(_kind = runtime.python_library,
     name = "convert_constant_dim_order_pass",
     srcs = [
         "convert_constant_dim_order_pass.py",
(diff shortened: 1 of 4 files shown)


metascroy:
@manuelcandales @mergennachin I think you can use this in the metal delegate

larryliu0820 (approved):
This is great! Should we turn it on by default?

metascroy:
> This is great! Should we turn it on by default?

Could do as follow up

mergennachin:
@metascroy 

There's already CSE in torch fx that we can use ```from torch.fx.passes.dialect.common.cse_pass import CSEPass```

See a draft PR on how I using (before running another fusion pass)

https://github.com/pytorch/executorch/pull/15813/

metascroy:
> @metascroy
> 
> There's already CSE in torch fx that we can use `from torch.fx.passes.dialect.common.cse_pass import CSEPass`
> 
> See a draft PR on how I using (before running another fusion pass)
> 
> #15813

Oh sorry, I merged this already. I will check of the CSE pass in torch and unify with ExecuTorch pass as follow-up.

PR #17766, opened 2026-02-28
Cortex-M backend: Fix conv2d scratch buffer allocation to match CMSIS-NN wrapper dispatch

rascani:
### Summary
Use arm_convolve_wrapper_s8_get_buffer_size instead of arm_convolve_s8_get_buffer_size so the buffer size matches whichever specialized kernel arm_convolve_wrapper_s8 will actually dispatch to at runtime (1x1 fast, 1xN, or general).

Also remove the Error::NotFound carve-out that silently proceeded with a null scratch buffer — CMSIS-NN returns ARM_CMSIS_NN_ARG_ERROR when ctx->buf is NULL and a buffer is required, so fail immediately on any allocation error, consistent with the other cortex_m conv ops.

Update CMSIS-NN from v7.0.0 to 84303a51fd867c7ddbd23068b7ce930af1b6269d
and remove GIT_SHALLOW (incompatible with SHA-based FetchContent pins).

Fixes #18044


rascani:
Merge will be blocked until https://github.com/ARM-software/CMSIS-NN/pull/200 has merged and we can update the CMSIS version.

zingo (approved):

mansnils (approved):

rascani:
Failures unrelated.

PR #17776, opened 2026-03-02
Cortex_M backend: Improve PatternMatcher algorithm

AdrianLundell:
The previous algorithm failed for cases where a pattern matched outside of the nodes given by the node iterator.

Update to a queue based algorithm to keep the function as a generator and add testing of this case.

Additionally do not quantize operators with very large inputs and remove unused filter_fn for filtering out nodes.

### DIFF
--- a/backends/cortex_m/quantizer/pattern_matcher.py
+++ b/backends/cortex_m/quantizer/pattern_matcher.py
@@ -8,9 +8,12 @@ from collections import defaultdict
 from dataclasses import dataclass
 from typing import Iterator, List, Optional
 
+from executorch.backends.arm.quantizer.quantization_annotator import _is_large_scalar
+
 from executorch.backends.cortex_m.quantizer.pattern_checkers import PatternCheck
 from executorch.backends.cortex_m.quantizer.quantization_configs import (
     CortexMQuantizationConfig,
+    QuantizationConfig,
 )
 from torch._ops import OpOverload
 from torch.fx import Node
@@ -30,7 +33,7 @@ class PatternMatcher:
     Attributes:
         support_dict: A dictionary mapping patterns (tuples of operator overloads) to
                   PatternCheck instances that validate the patterns.
-        filter_fn: A global filter applied over all nodes to exclude them from matching.
+        support_dict_name: An optional name for the support dict, used for logging.
     """
 
     Q_PATTERN_MATCHED_KEY = "quantizer_matched"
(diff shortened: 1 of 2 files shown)


rascani (approved):

backends/cortex_m/quantizer/pattern_matcher.py
```diff
@@ -40,63 +43,47 @@ class PatternMatcher:
...
         support_dict_name: str | None = None,
     ):
         self.support_dict = support_dict
-        self.filter_fn = filter_fn
         self.support_dict_name = support_dict_name
 
         self.patterns_by_first = defaultdict(list)
-        for p in sorted(support_dict.keys(), key=len, reverse=True):
-            self.patterns_by_first[p[0]].append(p)
+        self.max_pattern_len = max(len(pattern) for pattern in support_dict.keys())
```
psiddh: If we pass 'empty iterable', say PatternMatcher({}), would it be a problem ?

something like this would be more defensive ? 
self.max_pattern_len = max((len(p) for p in support_dict.keys()), default=0)

backends/cortex_m/quantizer/pattern_matcher.py
```diff
@@ -40,63 +43,47 @@ class PatternMatcher:
...
         self,
         support_dict: dict[tuple[OpOverload, ...], PatternCheck],
-        filter_fn=lambda node: False,
         support_dict_name: str | None = None,
     ):
         self.support_dict = support_dict
-        self.filter_fn = filter_fn
         self.support_dict_name = support_dict_name
 
         self.patterns_by_first = defaultdict(list)
```
psiddh: self.patterns_by_first used anywhere now ?

backends/cortex_m/quantizer/pattern_matcher.py
```diff
@@ -40,63 +43,47 @@ class PatternMatcher:
...
-        Return true if the node is a valid match for the given target.
-        """
-        if node is None:
-            return False, None
-        if not node.target == target:
-            return False, None
-        if node.meta.get(self.Q_PATTERN_MATCHED_KEY, False):
-            return False, self.REJECT_PREVIOUSLY_ANNOTATED
-        if self.filter_fn(node):
-            return False, self.REJECT_FILTERED_OUT
```
psiddh, line 71: nits : looks like we can remove  REJECT_FILTERED_OUT, unused ?

backends/cortex_m/quantizer/pattern_matcher.py
```diff
@@ -107,8 +94,68 @@ def check_pattern(
...
+            )  # Look for shorter matches in the next iter if no accepted match found
+
+            if match:
+                validated_match = self._validate_match(match, quantization_config)
+                accepted = validated_match.accepted
+                matches.append(validated_match)
+
+        return matches
+
+    def _deque_and_get_matches(
```
psiddh: nits: Should this be more like ? _dequeue_and_get_matches

psiddh (approved):

backends/cortex_m/quantizer/pattern_matcher.py
AdrianLundell: No, removed

backends/cortex_m/quantizer/pattern_matcher.py
AdrianLundell: Good catch, removed

backends/cortex_m/quantizer/pattern_matcher.py
AdrianLundell: Done, it will not match anything but it is good that we don't crash.

backends/cortex_m/quantizer/pattern_matcher.py
AdrianLundell: Correct!

PR #17786, opened 2026-03-02
Skip autoregressive generation during QNN LLM calibration

abhinaykukkadapu:
Fixes #17785 (Optimization 1)

During calibration, after prefilling the prompt tokens into the KV cache, the pipeline previously ran `_generate` — an autoregressive loop producing hundreds of tokens until EOS. This generation is unnecessary because the quantization observers already have sufficient activation statistics from the prefill pass.

For qwen2_5-1_5b (hybrid, max_seq_len=1024), this eliminates ~188 min of wasted computation (from 189 min down to <1 min for prompt calibration), reducing total end-to-end compilation time from ~8.1h to ~3.9h.

The `skip_generate` flag is threaded through `graph_module_inference` → `GraphModuleCalibrationWrapper._model_call` → `kv_inference`, which gates the `_generate()` call. Normal (non-calibration) inference is unaffected since the default is `skip_generate=False`.

**Note: See the gh issue #17785 for profiling results**

This PR was co-authored with Claude.

### DIFF
--- a/examples/qualcomm/oss_scripts/llama/decoder_utils.py
+++ b/examples/qualcomm/oss_scripts/llama/decoder_utils.py
@@ -94,6 +94,7 @@ class GraphModuleCalibrationWrapper(EagerEvalWrapper):
         get_example_inputs: Callable,
         use_i64_token: bool,
         seq_mse_candidates: int,
+        skip_generate: bool = False,
     ):
         # n seq len = n-1 cache len, so we len(inps) = n-1 during _model_call
         assert max_seq_length is not None, "max_seq_length must be provided"
@@ -107,6 +108,7 @@ class GraphModuleCalibrationWrapper(EagerEvalWrapper):
         self.max_seq_length = max_seq_length
         self.use_i64_token = use_i64_token
         self.seq_mse_candidates = seq_mse_candidates
+        self.skip_generate = skip_generate
 
     def _model_call(self, inps):
         all_logits = None
@@ -114,6 +116,7 @@ class GraphModuleCalibrationWrapper(EagerEvalWrapper):
         if self._use_kv_cache:
             kwargs["ar_len"] = self.ar_len
             kwargs["seq_mse_candidates"] = self.seq_mse_candidates
+            kwargs["skip_generate"] = self.skip_generate
 
         all_logits = INFERENCE_REGISTRY[self._use_kv_cache](
             self.get_example_inputs,
@@ -675,6 +678,7 @@ def kv_inference(  # noqa: C901
     collect_logits=False,
     seq_mse_candidates=0,
     lookahead_config=None,
+    skip_generate=False,
 ):
     is_multimodal = all(
         [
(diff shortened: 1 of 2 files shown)


cccclai:
I think about it more and feel like the intention ideally should be passed in the actual prompt + output, instead of let it generate to a certain length. Then the model will be the same amount of data anyway instead of seeing data while generating (which can be expensive)

haowhsu-quic:
I think `_generate` might be still necessary for some special tokens like reasoning which may not appear in the tasks.
Thanks for the thoughts from Chen, will figure out some possible reworks for the calibration process.

abhinaykukkadapu:
Thanks @cccclai and @haowhsu-quic for the review.

> intention ideally should be passed in the actual prompt + output

@cccclai do you mean we should replicate the prompt to max length to avoid triggering _generate during prompt calibration?

@haowhsu-quic Good point about special tokens. The current `_generate` loop doesn't guarantee those tokens appear either (depends on prompt?). A more reliable approach would be to include calibration prompts that explicitly contain the special tokens we care about?

haowhsu-quic:
I think we will apply chat template on user prompts, like wrapping it with \<think> for qwen3. Maybe having some quality dataset where model specific tokens are well-covered in the calibration process might be better (like Chen's thought if I understand correctly).

cccclai:
I think the prompt is provided by users, and after apply chat template, the special tokens might show up (wouldn't it show up in wikitext too? did we only apply chat template to the prompt but not the task? ). My suggestion is mostly to have a prompt with generated output that covers the tokens, so calibration we see both prompt + generated output at once and we don't need to generate them while calibrating. It should at least improve the prefill calibration a lot.

abhinaykukkadapu:
> I think the prompt is provided by users, and after apply chat template, the special tokens might show up (wouldn't it show up in wikitext too? did we only apply chat template to the prompt but not the task? ). My suggestion is mostly to have a prompt with generated output that covers the tokens, so calibration we see both prompt + generated output at once and we don't need to generate them while calibrating. It should at least improve the prefill calibration a lot.

@cccclai great idea, do you think we can do this? "Generate first on the fp32 model (before observer insertion), then calibrate the full sequence via prefill only" then we don't have to gate `_generate` at all. But if we go ahead with skipping prefill calibration from #17805 this wouldn't yield much savings, as the decode loop will run eitherway?

CC: @haowhsu-quic

haowhsu-quic:
The task calibration does not include special tokens, but user prompt does.
I think maybe to have a scenario like:
1. Generate user prompts by vanilla fp model without observer attached.
2. Export model with `prepare_pt2e` in `AR-CL` or `batch AR-N` to perform maybe 1~2 inference to cover all the tokens (tasks + user prompts).
3. Override encoding for both `AR-1` & `AR-N`

This approach probably could save calibration time but have higher memory footprint as a tradeoff.

abhinaykukkadapu:
We don't need this after the PR: https://github.com/pytorch/executorch/pull/17805

PR #17789, opened 2026-03-02
Partially reverting #17741 to fix Windows CUDA CI job

larryliu0820:
## Summary

  The IRunner refactoring in #17741 split `MultimodalRunner::generate()` into
  separate `prefill()` and `decode_from_token()` method calls. On Windows with
  CUDA, this triggers `STATUS_STACK_BUFFER_OVERRUN` (0xC0000409) — confirmed
  via SSH debugging that inlining the logic works but calling sub-methods from
  `generate()` crashes, even with an 8 MB stack.

  - Restore the monolithic `generate(vector, ...)` implementation with all
    prefill and decode logic inline, matching the pre-#17741 pattern
  - The separate `prefill()` and `decode_from_token()` methods are retained
    for external callers and the prefill-then-generate workflow
  - Pass `std::function` params to `decode_from_token` by const ref
  - Increase `voxtral_runner` stack to 8 MB on Windows (`/STACK:8388608`)
  - Print runner stderr in the Windows E2E test for diagnostics

## Test plan

  - [x] Verified fix on Windows CUDA CI machine via SSH
  - [x] `test-model-cuda-windows-e2e (mistralai, Voxtral-Mini-3B-2507, non-quantized)` CI passes

### DIFF
--- a/.ci/scripts/test_model_e2e_windows.ps1
+++ b/.ci/scripts/test_model_e2e_windows.ps1
@@ -186,14 +186,17 @@ try {
             -RedirectStandardError $stderrFile
 
         $stdout = if (Test-Path -Path $stdoutFile -PathType Leaf) { Get-Content -Path $stdoutFile -Raw } else { "" }
+        $stderr = if (Test-Path -Path $stderrFile -PathType Leaf) { Get-Content -Path $stderrFile -Raw } else { "" }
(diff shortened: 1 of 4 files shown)


manuelcandales (approved):

kirklandsign:
Thanks!

PR #17799, opened 2026-03-02
Cortex-M backend: Add quantized int8 batch matmul (CMSIS-NN)

rascani:
### Summary
Add cortex_m::quantized_batch_matmul wrapping arm_batch_matmul_s8. The RHS is always pre-transposed: constant RHS (parameters) are transposed at AOT time in the pass, dynamic RHS get a cortex_m::transpose node inserted in the graph.

It would be preferable if we could pre-compute or cache the constant RHS kernel sums, but I could not find any public CMSIS-NN APIs that would allow us to do so.

Fixes #16109

Authored with Claude.

### Test plan
```
pytest backends/cortex_m/test/ops/test_batch_matmul.py
```

### DIFF
--- a/backends/cortex_m/CMakeLists.txt
+++ b/backends/cortex_m/CMakeLists.txt
@@ -53,21 +53,22 @@ endif()
 
 # Cortex-M ops kernel sources
 set(_cortex_m_kernels__srcs
-    ${CMAKE_CURRENT_SOURCE_DIR}/ops/op_quantize_per_tensor.cpp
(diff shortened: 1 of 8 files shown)


rascani:
Any thoughts @AdrianLundell? Not sure if you had an alternate approach to bmm in mind.

AdrianLundell (approved):

AdrianLundell:
Sorry I missed this, I agree this looks reasonable!

PR #17804, opened 2026-03-03
 Qualcomm AI Engine Direct - Debugger Converge Phase 1: Introduce debug handle

winskuo-quic:
### Summary
We will be replacing part of the features in Qualcomm Intermediate Debugger with  ExecuTorch's Debugger shortly.
This includes using [Intermediate_Output_Capturer](https://github.com/pytorch/executorch/blob/0a8b904d7facd05f20133093da6f7a2f6c45758a/devtools/inspector/_intermediate_output_capturer.py#L18) to capture CPU golden output and [calculate_numeric_gap](https://github.com/pytorch/executorch/blob/0a8b904d7facd05f20133093da6f7a2f6c45758a/devtools/inspector/_inspector.py#L1468) to perform tensor to tensor comparison.
To reuse these features, one key component is to enable debug_handle since ExecuTorch's tool saves per tensor result with mapping of {debug_handle : tensor_value}

This PR will be the 1st stage of debugger convergence, which is to enable debug_handle.

### Test plan
`python -m unittest executorch.backends.qualcomm.tests.test_passes.TestPasses.test_resolve_debug_handle`

### DIFF
--- a/backends/qualcomm/_passes/qnn_pass_manager.py
+++ b/backends/qualcomm/_passes/qnn_pass_manager.py
@@ -53,6 +53,7 @@ from executorch.backends.qualcomm._passes import (
     RemoveRedundancy,
     ReplaceArangeArgs,
     ReplaceInfValues,
+    ResolveDebugHandle,
     TagQuantIO,
 )
 from executorch.backends.qualcomm._passes.utils import (
@@ -107,6 +108,7 @@ def get_capture_program_passes():
         (Remove0DTensor, True),
         (RemoveRedundancy, True),
         (TagQuantIO, False),
+        # ResolveDebugHandle will be added to last, check sorting below
     ]
 
     passes = OrderedDict()
@@ -175,6 +177,7 @@ class QnnPassManager(PassManager):
             if "edge_program" in kwargs:
                 kwargs["edge_program"] = exported_program
             self.add_pass(p(**kwargs))
+        self.add_pass(ResolveDebugHandle())
         return self.passes
 
     def transform_for_to_edge_pipeline(
--- a/backends/qualcomm/_passes/qnn_pass_manager.py
+++ b/backends/qualcomm/_passes/qnn_pass_manager.py
@@ -108,7 +108,7 @@ def get_capture_program_passes():
         (Remove0DTensor, True),
         (RemoveRedundancy, True),
         (TagQuantIO, False),
-        # ResolveDebugHandle will be added to last, check sorting below
+        (ResolveDebugHandle, True),
     ]
 
     passes = OrderedDict()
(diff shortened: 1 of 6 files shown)


winskuo-quic:
Hi @Gasoonjia, @cccclai,
As discussed offline and in the email thread, we will start converging Qualcomm Intermediate Debugger with ExecuTorch's Debugger API.
We will break this feature down 2 to 3 PRs. This will be the first PR, focusing on enabling debug_handle for Qualcomm backend. I have added a unit test to ensure that the behavior is desired.
Please have a look. Thanks

backends/qualcomm/_passes/resolve_debug_handle.py
```diff
@@ -0,0 +1,47 @@
...
+import torch
+from executorch.exir.debug_handle_utils import DEBUG_HANDLE_KEY
+from executorch.exir.pass_base import ExportPass, PassResult
+
+
+class ResolveDebugHandle(ExportPass):
+    """
+    Caution: This pass is executed as the last of the edge_passes.
+    For any passes executed during qnn_preprocess, users will need to handle debug_handle ID themselves.
+
```
Gasoonjia, line 17: maybe we should add a comment here, highlighting that conditional branch like torch.cond is not supported right now.

backends/qualcomm/_passes/qnn_pass_manager.py
Gasoonjia, line 90: ok so the order in this list will be the order applying the corresponding pass after `to_edge` right?

Gasoonjia (approved):
LGTM, thanks for your work!

backends/qualcomm/_passes/resolve_debug_handle.py
winskuo-quic: Hi @Gasoonjia,
Thanks for reviewing the PR.
I do have a comment below in `line 31` for mentioning about conditional branch.
Do you think the comment looks good to you?

backends/qualcomm/_passes/qnn_pass_manager.py
winskuo-quic: So the order is determined using the dependency table under `backends/qualcomm/_passes/utils.py`.
I made `ResolveDebugHandle` as the last executed pass in the dependency table. I have also added an assertion check under `qnn_pass_manager.py` to ensure this is the last pass for edge_transform.

backends/qualcomm/_passes/resolve_debug_handle.py
Gasoonjia: hi @winskuo-quic, it would be great to highlight that at the beginning of the class. User will first and sometimes only read the class-level comment and it will be great if we can warn them potential issue here.

winskuo-quic: That makes sense. Thanks for clarification. I have changed to the beginning of the class.

Gasoonjia (approved):

Gasoonjia:
LGTM! thanks for the great work!

winskuo-quic:
Hi @Gasoonjia,
We are ready to start the implementation of phase 2 for Debugger Convergence. May I kindly ask if this PR could be merged?
Thanks

Gasoonjia:
oh yeah sure i thought you can merge PR by yourself. Merge it!

PR #17805, opened 2026-03-03
Qualcomm AI Engine Direct - remove prefill calibration

haowhsu-quic:
### Summary
- calibrate kv text decoder only to reduce calibration time
- deprecate outdated implementation & use deterministic example inputs for llm

**Total Quantization Time**
| Model | Before(s) | After(s) | Improvement |
| :---: | :---: |  :---: |  :---: |
| gemma-2b | 2203.399 | 999.512 | 54.64% |
| gemma2-2b | 2177.285 | 1001.248 | 54.01% |
| gemma3-1b | 1776.861 | 548.312 | 69.14% |
| glm-1_5b | 1434.780 | 677.257 | 52.8% |
| granite_3_3-2b | 59566.790 | 6165.443 | 89.65% |
| llama3_2-1b | 4528.620 | 2953.233 | 34.79% |
| llama3_2-3b | 5744.429 | 1652.157 | 71.24% |
| phi_4_mini | 7005.601 | 2071.634 | 84.56% |
| qwen2_5-0_5b | 480.508 | 372.076 | 22.57% |
| qwen2_5-1_5b | 2064.333 | 899.164 | 56.44% |
| qwen3-0_6b | 1673.150 | 1124.149 | 32.81% |
| qwen3-1_7b | 3253.723 | 1148.511 | 64.7% |
| smollm2_135m | 502.779 | 414.510 | 17.56% |
| smollm3-3b | 4663.057 | 1613.516 | 65.4% |
| smolvlm_500m_instruct | 288.246 | 170.829 | 40.73% |
| internvl3_1b | 256.624 | 170.811 | 33.44% |

### Test plan
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestExampleLLMScript / TestExampleMultimodalityScript`


haowhsu-quic:
Hi @cccclai, this PR is to reduce the calibration time of llms. Might be able to mitigate #17784 a bit, will keep working on other possible optimizations.
I also notice that #17718 is breaking torchtune package, could you help update torchtune repo? Thank you!

cccclai:
Thanks a lot! In addition to this, I noticed that SeqMSE grid searching is done in sequential instead of parallel, is there room to improve there?

haowhsu-quic:
> Thanks a lot! In addition to this, I noticed that SeqMSE grid searching is done in sequential instead of parallel, is there room to improve there?

Yes, will look into it.

cccclai:
> I also notice that https://github.com/pytorch/executorch/pull/17718 is breaking torchtune package, could you help update torchtune repo?

can you share what's broken? Is it the flow to consume torchtune in executorch repo?

examples/qualcomm/oss_scripts/llama/model/static_llama.py
```diff
@@ -776,15 +774,15 @@ def get_example_inputs(self):
             for _ in range(self.n_layers):
                 # transpose first to decrease the runtime efforts
                 k_cache.append(
-                    torch.zeros(
+                    torch.ones(
```
cccclai: Are we initialized kv cache with different values?

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
```diff
@@ -709,43 +644,118 @@ def __init__(
...
+            setattr(
+                prefill_model,
+                prefill_node.target,
+                getattr(decode_model, decode_node.target),
+            )
+            # scale / zero point are part of op's attributes
+            if list(decode_node.users)[0].target in ptq_target:
+                activation_override(decode_node, prefill_node)
+
+        # copy encoding for hybrid mode
```
cccclai, line 706: Are you copying over the quantization parameters from kv mode to prefill mode?

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
```diff
@@ -709,43 +644,118 @@ def __init__(
...
-                    else:
-                        raise RuntimeError("failed to override quantization attribute")
+        # perform encoding override for hybrid mode
+        # ---
+        # theoretically decode & prefill model should share the same encoding
+        # given that they are using the identical calibration dataset.
+        #
+        # however, pytorch will use different computaion kernels for different
+        # workloads (AR1 vs ARN) which will introduce some numerical discrepancy.
+        #
```
cccclai, line 748: what is the mechanism to make sure the encoding align correctly?

abhinaykukkadapu: I'm worried about the accuracy too if we get rid of prefill calibration, do you think if we generate prompt + output using fp32 model (pre observers) as discussed in PR #17786 and run prefill + decode as before with `skip_generate` might yield better accuracy, rather than getting rid of entire prefill calibration?

cccclai: prefill calibration ideally is not needed because decode see all the generated tokens too and prefill graph and decode graph should be the same. I remember @haowhsu-quic mentioned we insert the kv cache of the output of prefill and connect to the decode of the input to make sure those quant nodes are also calibrated. I did a comparison for quant params between prefill and decode in the past and they are very very close.  I'm trying to figure out if this PR handle kv cache differently than before.

abhinaykukkadapu: I guess my question is prefill "sees" previous tokens and as attention block will take these into consideration while generating kv-cache. 

For weights, what you said makes sense as they are idempotent from the math standpoint, maybe we should just check the PPL on-device?

CC: @metascroy, @kimishpatel if you have any thoughts on this.

haowhsu-quic:
> > I also notice that #17718 is breaking torchtune package, could you help update torchtune repo?
> 
> can you share what's broken? Is it the flow to consume torchtune in executorch repo?

I think we're using torchtune to convert parameter naming in some llms. Since torchao has deprecated the `Int8DynamicActivationInt4WeightConfig` and torchtune has corresponding fix in this [commit](https://github.com/meta-pytorch/torchtune/commit/6f2aa7254458145f99d7004cbd6ebc8e53a06404).
I wonder if it is possible to help update the pip package of torchtune to resolve this issue?

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
haowhsu-quic: Yes, since the current bottleneck of calibration time is using prefill mode for generating user prompts.
I think the future scenario we want to try is to generate all the special tokens for target model and merge them with task calibration data. Have prefill model to iterate over it and use prefill's encoding for kv. We could save the user prompt generation with this approach.

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
haowhsu-quic: The mechanism is to compare the topology order of ops in both graph where each op in the same order should have identical `nn_module_stack`. The number / type of QDQ pairs in nodes' (call_function / placeholder) users are required to be identical as well.

haowhsu-quic: > prefill calibration ideally is not needed because decode see all the generated tokens too and prefill graph and decode graph should be the same. I remember @haowhsu-quic mentioned we insert the kv cache of the output of prefill and connect to the decode of the input to make sure those quant nodes are also calibrated. I did a comparison for quant params between prefill and decode in the past and they are very very close. I'm trying to figure out if this PR handle kv cache differently than before.

We used to calibrate two graphs separately and have calibrated decode model's output encoding appear as FixedQParamObservers in prefill calibration's output. But models like VLM are super sensitive and will have bad accuracy under this scenario.
Theoretically both decode / prefill should share the same encoding, the numerical discrepancy is from pytorch kernels implementation.

haowhsu-quic: > I guess my question is prefill "sees" previous tokens and as attention block will take these into consideration while generating kv-cache.
> 
> For weights, what you said makes sense as they are idempotent from the math standpoint, maybe we should just check the PPL on-device?
> 
> CC: @metascroy, @kimishpatel if you have any thoughts on this.

I think both decode / prefill are doing calibration against the same task dataset & user prompt. They should see the identical passed tokens in my understanding.

cccclai: > pytorch kernels implementation

That's interesting, can you explain a bit more that - I understand it's orthogonal to the PR but it's good to understand the workaround approach we take here

examples/qualcomm/oss_scripts/llama/model/static_llama.py
haowhsu-quic: This is for checking the numerical value with deterministic input for validation purpose. I can revert them back to zeros.

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
haowhsu-quic: Maybe a simple snippet could help:
```python
import torch

def fp_matmul(batch, ar_len, n_heads, head_dim, dim):
  with torch.no_grad():
    # ar_n case
    hidden_state = torch.randn(batch, ar_len, n_heads, head_dim)
    wq = torch.randn(head_dim, dim)
    q_ar_n = hidden_state @ wq
    # ar_1 case
    q_ar_1 = hidden_state[:, 0, ...] @ wq
    # check numeric difference
    print(f"is all close: {torch.allclose(q_ar_n[:, 0, ...], q_ar_1)}")

fp_matmul(1, 128, 24, 128, 512)
fp_matmul(1, 32, 12, 32, 64)
```

I think different optimization strategy based on the tensor size will have different accumulation order where the rounding error of floating point could happen (just my guess and please correct me if I'm wrong).

cccclai: @abhinaykukkadapu let's work with the team on confirming this

abhinaykukkadapu: @haowhsu-quic thanks for the explanation. I could repro the theory on different accumulation order, although the dimensions in your example seems to be bit exact (upto the defaults in allclose) i could see the numerical differences in higher dimensions, which happens to be the real dims for example in llama models. See below one uses `aten:mm` vs `aten:bmm`.

All being said, i think @haowhsu-quic this should match the accuracy compared to previous `FixedQParamsObserver` approach, but can you please verify the PPL on-device just to close the loop. Thanks again for this work. This will greatly improve the user experience calibrating the model.

```
import torch
from torch.profiler import profile, ProfilerActivity

hidden_state = torch.randn(1, 128, 32, 128)
wq = torch.randn(128, 4096)

with profile(activities=[ProfilerActivity.CPU], with_stack=True) as prof_arn:
    _ = hidden_state @ wq  # ARN: [1, 128, 32, 128] x [128, 4096]

with profile(activities=[ProfilerActivity.CPU], with_stack=True) as prof_ar1:
    _ = hidden_state[:, 0, ...] @ wq  # AR1: [1, 32, 128] x [128, 4096]

print("=== ARN (batched) ===")
print(prof_arn.key_averages().table(sort_by="cpu_time_total", row_limit=10))

print("\n=== AR1 (single) ===")
print(prof_ar1.key_averages().table(sort_by="cpu_time_total", row_limit=10))
```

haowhsu-quic: Thank you for giving it a try. I think we evaluate PPL with decode graph only, they should be the same as before given that the identical task dataset is used.

abhinaykukkadapu: @haowhsu-quic i wanted to give a quick try with these changes but seems like there is some issue with the logic in `HybridTextDecoder._encoding_override`

Command to repro:

```
python examples/qualcomm/oss_scripts/llama/llama.py -b build-android --compile_only -m SM8750 --temperature 0 --model_mode hybrid --max_seq_len 64 --prefill_ar_len 64 --decoder_model qwen2_5-1_5b --prompt "I would like to learn python, could you teach me with a simple example?" --tasks wikitext --limit 1 2>&1 | tee logs.txt
```
Here are the logs: 

```
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/llama.py", line 683, in main
    export_llama(args)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/llama.py", line 636, in export_llama
    compile(
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/llama.py", line 163, in compile
    multi_modal_mgr.compile(compile_specs=compile_specs, pte_filenames=pte_filenames)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 60, in wrapper
    func(cls, *args, **kwargs)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py", line 1026, in compile
    self.process(compile_request)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py", line 1008, in process
    Processor.process(self, request)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 155, in process
    return self._next_handler.process(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 181, in process
    super().process(request)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 155, in process
    return self._next_handler.process(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 181, in process
    super().process(request)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 155, in process
    return self._next_handler.process(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 180, in process
    getattr(self, request.method_name)(request)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py", line 60, in wrapper
    func(cls, *args, **kwargs)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py", line 769, in compile
    self._encoding_override(
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py", line 742, in _encoding_override
    compare_nodes(act_decode, act_prefill)
  File "/data/users/abhinayk/executorch/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py", line 689, in compare_nodes
    assert info(decode_node) == info(
AssertionError: found unmatched order for ops: input_pos va embedding
```


This error seems to comes from HybridTextDecoder._encoding_override (line 670-672 in the PR's llm_wrappers.py):                                      
                                                                                                                                              
assert info(decode_node) == info(                                                                                                           
    prefill_node                                                                                                                            
), f"found unmatched order for ops: {decode_node} va {prefill_node}"

The _encoding_override method tries to copy quantization encodings from the DECODE model to the PREFILL model by iterating through their graphs in parallel. It expects both models to have the same topology. But the DECODE model has an `input_pos` placeholder node where the PREFILL model has an `embedding` node.

haowhsu-quic: I think the model definition would be a bit different in `static_llama.py` (e.g. no kv_caches show up as inputs) if [`prefill_ar_len` == `max_seq_len`](https://github.com/pytorch/executorch/blob/8b353877820694a0cf45774e33c13b340d4f1173/examples/qualcomm/oss_scripts/llama/wrappers/base_component.py#L132). We normally use configurations that `prefill_ar_len` < `max_seq_len`, could you give it a try?

haowhsu-quic: Will have a fix for condition `prefill_ar_len` == `max_seq_len`, thanks for identifying this.

abhinaykukkadapu: Ran comparison over the weekend, observed great speedups overall, calibration time is cut in half overall time is cut by ~33% for the qwen model i'm benchmarking. I will do a final PPL check on-device and we will be good to go. Thanks again for working on this.

I think we also should find ways to cut per-iteration decode timing.

```
┌─────────────────────┬──────────┬────────┬─────────────────────┐
│                     │ Baseline │   PR   │       Savings       │
├─────────────────────┼──────────┼────────┼─────────────────────┤
│ DECODE calibration  │ 2h 59m   │ 2h 49m │ ~10m (noise)        │
├─────────────────────┼──────────┼────────┼─────────────────────┤
│ PREFILL calibration │ 2h 34m   │ 15.5s  │ 2h 34m              │
├─────────────────────┼──────────┼────────┼─────────────────────┤
│ Quantization total  │ 5h 33m   │ 2h 50m │ 2h 43m (49% faster) │
├─────────────────────┼──────────┼────────┼─────────────────────┤
│ Compile total       │ 2h 35m   │ 2h 35m │ ~0 (noise)          │
├─────────────────────┼──────────┼────────┼─────────────────────┤
│ End-to-end          │ 8h 13m   │ 5h 30m │ 2h 43m (33% faster) │
└─────────────────────┴──────────┴────────┴─────────────────────┘
```

abhinaykukkadapu: @haowhsu-quic PPL looks fine.

Let's go with the change. Can you please rebase.

abhinaykukkadapu (approved):

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
cccclai: > The mechanism is to compare the topology order of ops in both graph where each op in the same order should have identical nn_module_stack. The number / type of QDQ pairs in nodes' (call_function / placeholder) users are required to be identical as well.

Is there any particular reason that we remove the custom annotation? I understand that the custom annotation caused prefill/decode have different graphs, but how about the quantization params for kv cache?

haowhsu-quic: I think quantization params for placeholders (kv cache) will also be shared between prefill & decode. They should be the same under identical calibration data. May I learn more about your concern?

cccclai: The concern was previously raised by @shewu-quic and I would like to see if it's resolved or skipped. The claim is that the kv cache is shared by prefill and decode, we need to get the token out from prefill and passed it to decode to properly calibrate this part. 

I did remove the custom annotation and copied the quantization params from decode and prefill, most of them are about 10^-3 close, while there are a few outliers, I'm not sure if it's due to this or the different PyTorch kernels (mm vs mv). Have you seen that?

haowhsu-quic: Thanks for the information. We used to have shared encoding in order to prevent runtime requantization when switch from prefill mode to decode mode. And this behavior (custom annotation) is not required with this change.

I've seen that phenomenon; I think it's due to the accumulated rounding difference will become significant as attention layer grows.

cccclai: hmm okay maybe my understanding about the intention for custom annotation is wrong. Can you shared why we have it at first place

abhinaykukkadapu:
@haowhsu-quic can we modify this check: `args.max_context_len >= args.prefill_ar_len` to `args.max_context_len > args.prefill_ar_len` in hybrid mode after the changes?

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
haowhsu-quic: The initial motivation was trying to speed up the inference process, so we removed the QDQ for kv cache and logits. But the encoding is quite different between prefill & decode after calibration. We decided to introduce aforementioned custom annotation to help resolve this issue.

examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
```diff
@@ -712,43 +647,118 @@ def __init__(
...
+        ptq_target = {
+            torch.ops.quantized_decomposed.dequantize_per_tensor.default,
+            torch.ops.quantized_decomposed.quantize_per_tensor.default,
+        }
+        qdq_target = pbq_target | pcq_target | ptq_target
+
+        def compare_nodes(decode_node, prefill_node):
+            def info(node):
+                return node.name + (
+                    str(decode_node.meta["nn_module_stack"].values())
```
abhinaykukkadapu: Will merge it tomorrow after the above change and possible this below one is in.

@haowhsu-quic sorry for the back and forth, i think this is capturing wrong variable? shouldn't this be

```
str(node.meta["nn_module_stack"].values()) if node.op == "call_function"
```

haowhsu-quic: Thank you for catching this, will modify it and test a few models locally.

abhinaykukkadapu (approved):

PR #17812, opened 2026-03-03
Cortex-M backend: Improve int8-portable operator support

AdrianLundell:
Using the int8 portable ops is an option for most ops not requiring rescales such as data move ops, max/min ops and logic ops. Even though they are not accelerated, quantizing them to int8 is more efficient than doing them in fp32+dequant/quantization.

This patch adds a large number of such ops to be quantized by the SharedQspecQuantizer together with tests. It also modifies the quantizer to increase the support:
- If multiple qspecs are found, use top one rather than falling back to fp32 since this is what users most likely want.
- Reject node with non-float inputs/outputs which would previously crash
- Let BFS algorithm search through non-float edges to support indexing ops.


rascani (approved):

AdrianLundell:
Unrelated fails

PR #17818, opened 2026-03-03
NXP backend: added support for `aten.bmm`

novak-vaclav:
### Summary

adds support for `aten.bmm` operator.

The original PR is [here](https://github.com/pytorch/executorch/pull/17670), however I pushed to the branch without committing the work first and the PR closed itself auto-magically.

### Test plan

cc @robert-kalmar @JakeStevens @digantdesai @MartinPavella


novak-vaclav:
@MartinPavella 
Internal tests passing [here](https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH97-8).

Please try to re-run the `periodic / gather models` Github test if possible, according to the log it was stopped because some other, higher priority task was scheduled.

After that, this can be merged ✔️

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,145 @@
...
+from executorch.backends.nxp.tests.models import BatchMatMulConvModel, BatchMatMulModel
+from executorch.exir.dialects._ops import ops as exir_ops
+
+
+@pytest.fixture(autouse=True)
+def reseed_model_per_test_run():
+    torch.manual_seed(23)
+
+
+# noinspection PyProtectedMember
```
MartinPavella: Nit: This is still here.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,145 @@
...
+    assert graph_contains_any_of_ops(delegated_ep.graph, [Bmm])
+
+
+@pytest.mark.parametrize(
+    "conv_input_shape, bmm_input_shape",
+    [
+        pytest.param((4, 8, 16), (4, 16, 16), id="3D with conv. quant"),
+    ],
+)
+def test_convert_bmm__conv_quant(mocker, conv_input_shape, bmm_input_shape):
```
MartinPavella: I am surprised that this test passes. Let me look into it a bit more before merging.

MartinPavella: There is indeed an issue in your implementation. I will try to explain it here:
Your model gets partitioned in the following way:

image

So the convolution is not delegated, and the `bmm` is the only delegated node.

In your test, you verify correctness by running the delegated edge partition (`bmm_intermediate_ep`), and its Neutron IR converted equivalent (`bmm_neutron_ir_model`). The NodeFormatInference identified that the input and output nodes are channels first (due to the conv), so transpositions are required. As the first dimensions of the IO shapes in the delegated partition are `4`, the `Transpose` operator would not be supported on Neutron, so the transposition must be done on the CPU. You should have identified this and permuted the input/output testing data using the `tflite_input_preprocess` and `tflite_output_preprocess`. Or (even better) you could have prohibited the insertions of the `Transpose` ops regardless of whether they are supported (`use_neutron_for_format_conversion=False`), and used the pre/post processing (as you asked me on a recent PR if you remember).

The delegated Neutron IR partition looks like this (notice there are no `Transpose` ops):
image


So both the edge `bmm` and the Neutron IR `BatchMatMul` were computing on the same channels FIRST data in your test (not permuted to channels LAST for the Neutron IR model), which is not what would happen in an end to end test (if the full model was run using the `nxp_executor_runner`). 
If you hadn't used equal height and channels (`16`), or if the batch size was 1, the test would have crashed. That's why it's a good idea sometimes to test with weird shapes.

---

### **What needs to be done?**

You need to implement proper conversion support for the channels first version of `aten.bmm`. That means, in the IR, permuting it's IO to channels first (because it will automatically be channels last because of the `convolution`) using `Transpose` ops (or the parameters of the `BatchMatMul` perhaps?). I don't believe any changes to the `NodeFormatInference` are required.

We can talk about this if you have any questions. It's not easy to describe the issue only text :D

Edit:
Feel free to draw some inspiration from [onnx2tflite](https://bitbucket.sw.nxp.com/projects/AITEC/repos/eiq-onnx2tflite/browse/onnx2tflite/src/converter/node_converters/mat_mul_converter.py#193)

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,113 @@
...
+    def convert(self, node: Node):
+        """Convert the `aten.bmm` operator to TFLite `BatchMatMul`."""
+        self.assert_convertible(node)
+
+        t_op = self._create_tflite_op_with_io_tensors(node)
+
+        # We set `adj_x = adj_y = True` when the inputs are in channels‑first format so
+        # that TFLite internally transposes them to channels‑last. In that case, the
+        # output also becomes channels‑last, so we need to transpose it back to
+        # channels‑first afterward.
```
MartinPavella: This comment is incorrect. It's the other way around

\# We set `adj_x = adj_y = True` when the inputs are in channels‑**last** format so
\# that TFLite internally transposes them to channels‑**first.** In that case, the
\# output also becomes channels‑**first**, so we need to transpose it back to
\# channels‑**last** afterward.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,154 @@
...
+    # Make sure the `bmm` was NOT delegated.
+    assert graph_contains_any_of_ops(delegated_ep.graph, [Bmm])
+
+
+def test_convert_bmm__conv_quant(mocker, use_qat):
+    # These must match:
+    # - `h1 = h2`
+    # - `w1 = c2`
+    # Otherwise it violates `bmm` constraints.
+    h1 = h2 = 5
```
MartinPavella: Is it really supposed to be "height"? Isn't the first dimension "batch"?

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,113 @@
...
+        # Assign the operator its TFLite inputs and outputs
+        t_op.tmp_inputs = [x1, x2]
+        t_op.tmp_outputs = [y]
+
+        ops = OpsList(middle_op=t_op)
+
+        # Transpose back to channels-last if needed.
+        if node.meta[NXP_NODE_FORMAT].is_channels_first():
+            tensor_rank = input_rank(node, 0)
+            perm = translator.create_channels_first_to_channels_last_permutation(
```
MartinPavella, line 112: You need to verify that this `Transpose` op is supported in the `_is_supported_on_target()` method.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,154 @@
...
+    delegated_ep = to_quantized_edge_program(
+        model,
+        [input_shape_x1, input_shape_x2],
+    ).exported_program()
+
+    # Make sure the `bmm` was NOT delegated.
+    assert graph_contains_any_of_ops(delegated_ep.graph, [Bmm])
+
+
+def test_convert_bmm__conv_quant(mocker, use_qat):
```
MartinPavella: The point of this test is to verify correct support for the channels first format. The fact that this is achieved by using a `convolution` is not important (you could have also used MaxPool, AveragePool, Upsample, ...). So please consider a different test name. Maybe something like:
`test_convert_bmm__channels_first`

MartinPavella (changes requested):

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: fixed

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
novak-vaclav: Thank you, fixed.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: In the Neutron Converter documentation, the first dimension in 3D tensors was referred to as `height`, but you are right that in executorch this is considered `batch`. Fixed.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: I took the naming from some other, already merged test suite, but you're right. Fixed the name.

novak-vaclav:
Fixed the issues mentioned by @MartinPavella
When rebasing and splitting the feature changes and the rebased changes into separate commits, and then pushing them in the wrong order, it automatically closes the PR. However I found out that it can be reopened.

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,133 @@
...
+        if input_rank(node, 0) != 3 or input_rank(node, 1) != 3:
+            return False
+
+        return True
+
+    @staticmethod
+    def _get_channels_first_shape(node: Node) -> list[int]:
+        input_shape = node.meta["val"].shape
+
+        if node.meta[NXP_NODE_FORMAT].is_channels_last():
```
MartinPavella: This will always be false.
If you remember, last week we talked about how all nodes in the edge model are either channels first or formatless, and then all tensors in the Neutron IR are either channels last or formatless. So no node in the edge model will ever have the channels last format.

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,133 @@
...
+        is_ch_first_2 = node.args[1].meta[NXP_NODE_FORMAT].is_channels_first()
+        # This combination of node formats is not supported on Neutron (`adj_x = True`, `adj_y = False`),
+        # but it should never happen because both input tensors are expected to share the same format.
+        if is_ch_first_1 and not is_ch_first_2:
+            return False
+
+        # In case we need to insert transpose after `BatchMatMul`, we also need to check if
+        # such transposition is supported.
+        if node.meta[NXP_NODE_FORMAT].is_channels_first():
+            tensor_rank = input_rank(node, 0)
```
MartinPavella: Nit: Why are you getting the rank from the input? I guess it should always equal the output rank, but it seems like a bad practice.
Same on line 127

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,133 @@
...
+            tensor_shape = node.meta["val"].shape
+            perm = translator.create_channels_first_to_channels_last_permutation(
+                tensor_rank, return_list=True
+            )
+
+            if not transposition_is_supported_on_neutron(
+                tensor_shape, perm, neutron_target_spec
```
MartinPavella: After conversion to Neutron IR, the input of the `Transpose` will be channels last, but here you are checking with a channels first input shape. `tensor_shape` must be permuted to channels last first.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,202 @@
...
+            1: input_data_2,
+        },
+        atol=1,
+    )
+
+
+@pytest.mark.parametrize(
+    "input_shape_x1, input_shape_x2",
+    [
+        pytest.param((1, 8, 16), (1, 16, 24), id="3D, x1_W not divisible by NUM_MACS."),
```
MartinPavella: > x1_W not divisible by NUM_MACS

Unless W refers to the first dimension, this is not true,

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,202 @@
...
+
+    perm = translator.create_channels_first_to_channels_last_permutation(
+        len(x_input_shape), return_list=True
+    )
+    transp_not_supported_1 = not transposition_is_supported_on_neutron(
+        x_input_shape, perm, neutron_target_spec
+    )
+    transp_not_supported_2 = not transposition_is_supported_on_neutron(
+        y_input_shape, perm, neutron_target_spec
+    )
```
MartinPavella, line 157: These exact transpositions you are checking for are not necessary for `bmm` support, so it doesn't really make sense to skip the test if they are not supported. The real limitation would be:
```python
output_shape = (n1, c2, w1)  # I'm not 100% sure about the order here
transp_not_supported = not transposition_is_supported_on_neutron(
        output_shape, perm, neutron_target_spec
    )
```

But realistically the 2 transpositions you are checking will be supported at the same time as the actual limiting transposition, so I guess it's not a big deal.

MartinPavella (changes requested):

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
novak-vaclav: Yes thats true. Fixed it. Thanks

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
novak-vaclav: Because I took it from some other part of codebase :D but you are right, it is bad practice, fixed it.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: During the implementation and review of the feedback, I forgot to also modify this test. The test cases were all a bit off, so I fixed them. Please have a look.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: This makes sense, fixed it and checked the `output_shape` correctness to be sure.

novak-vaclav:
Fixed all the issues, please have a look @MartinPavella

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
```diff
@@ -0,0 +1,123 @@
...
+
+            tensor_shape_channels_last = [tensor_shape[i] for i in perm]
+            if not transposition_is_supported_on_neutron(
+                tensor_shape_channels_last, perm, neutron_target_spec
+            ):
+                return False
+
+        # Nodes in Edge models are always channels first/formatless
+        _, c1, w1 = node.args[0].meta["val"].shape
+        _, c2, w2 = node.args[1].meta["val"].shape
```
MartinPavella: Is this correct?
Your statement `# Nodes in Edge models are always channels first/formatless` is 100% correct, but `formatless` tensors are not permuted to channels last, so in Neutron IR, they have the same shape as in the edge model. And Neutron considers the last dimension to be the channels, so I would expect `_, w*, c* = node.args[*].meta["val"].shape` for the formatless case.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
```diff
@@ -0,0 +1,205 @@
...
+
+ExecutorchDelegateCall = torch.ops.higher_order.executorch_call_delegate
+Bmm = exir_ops.edge.aten.bmm.default
+
+
+@pytest.mark.parametrize(
+    "input_shape_x1, input_shape_x2",
+    [
+        pytest.param((1, 24, 16), (1, 16, 24), id="3D, one batch."),
+        pytest.param((3, 8, 24), (3, 24, 8), id="3D, more batches."),
```
MartinPavella, line 44: Regarding my other comment, these "symmetric" test cases are not good practice as they conceal potential issues. Please add a test case like:
```python
pytest.param((1, 24, 8), (1, 16, 24), id="3D, one batch. c1 != w2"),
```

novak-vaclav:
Just rebased onto main, will look into your new review comments now @MartinPavella

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
novak-vaclav: Such "non-symmetric" test cases will fail.
The formal requirements of matrix multiplication, thus also of matrix multiplication done in batches is that `w1 = c2`.
There is also a restriction imposed by Neutron converter that states that `c1 = w2`.
Therefore "non-symmetric" cases are not supported.

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
novak-vaclav: So you think I should add `if` statement that extracts `w*` and `c*` in different order for `channel-first` or `formatless`?

MartinPavella: Probably yes. It depends how Neutron handles the permutation of the inputs, and whether it's restrictions for the dimensions are the same or flipped when the flags are used.

backends/nxp/tests/ir/converter/node_converter/test_bmm_converter.py
MartinPavella: Oh, OK. I thought it was the standard matrix multiplication restriction.

backends/nxp/backend/ir/converter/node_converters/ops_converters/bmm_converter.py
novak-vaclav: Fixed as discussed in the office.

novak-vaclav:
Fixed all the issues as discussed in the office, please review @MartinPavella

MartinPavella (approved):

novak-vaclav:
@MartinPavella
The unittest checks are still timing out. Should I wait before it is fixed (the issue with the checks timing out also appears in other PRs) or should it be merged?
The `lintrunner-mypy` check fails in other PRs too and does not seem to be related to my code.

MartinPavella:
@novak-vaclav as you didn't modify any non-NXP files and the Linux unittests are passing, I don't see the need to wait for the window unittest time-out fix.
The lintrunner fail is indeed unrelated.
I'm merging this PR.

PR #17828, opened 2026-03-03
Mlx delegate part2

metascroy:
This is MLX delegate part2, which focusses on adding ops.


manuelcandales (approved):
stamping, make sure CI is green before landing

PR #17829, opened 2026-03-03
MLX delegate part 3

metascroy:
This is part 3 of the MLX delegate, which focusses on adding support for many GenAI models.


manuelcandales (approved):
stamping, make sure CI is green before landing

PR #17830, opened 2026-03-03
Add FuseConsecutiveRescalesPass to fuse redundant RESCALE pairs (#17830)

Ninja91:
Summary:

Add `FuseConsecutiveRescalesPass` to eliminate redundant INT32→INT8→INT32 RESCALE round-trips between chained arithmetic ops in the TOSA lowering pipeline.

`InsertRescaleInt32Pass` wraps each add/sub/mul with input RESCALEs (INT8→INT32) and output RESCALEs (INT32→INT8). When ops are chained (e.g., add→add or mul→add), the output RESCALE of op1 feeds directly into the input RESCALE of op2, creating a wasteful round-trip. Each unnecessary RESCALE decomposes into Add+Mul NPU instructions (~1,130 cycles each on Ethos-U55-128), and in quantized models RESCALE overhead accounts for 25-50% of total NPU cycles.

The pass detects consecutive RESCALE pairs (R1: INT32→INT8/INT16, R2: INT8/INT16→INT32) and handles two cases:

- **Identity** (composed scale ≈ 1.0, matching zero points): Removes both RESCALEs and directly wires R1's input to R2's users. This eliminates the entire round-trip. Bypassing the intermediate INT8/INT16 clamp can cause up to ~120 INT8 steps of output difference, handled via `qtol=1` in tests.

- **Non-identity**: Leaves the pair unchanged. Creating a single INT32→INT32 RESCALE would be semantically correct (and the TOSA ref model handles it), but Vela's NPU compiler produces all-zero outputs for INT32→INT32 RESCALE. Root cause: `EthosU55Constraints::SupportsRescale()` returns `false` for dtypes > 16 bits, causing `RewriteRescale()` to convert the RESCALE into a MUL with aggressive right-shift that zeros out values.

Multi-user R1 nodes (e.g., residual connections, branching) are handled by fusing each R1→R2 pair individually while preserving R1 for non-RESCALE users.

## Context

This pass runs unconditionally in the TOSA pipeline immediately after `InsertRescaleInt32Pass` (see `arm_pass_manager.py`). Identity pairs are the most common case between chained ops with similar quantization scales, so this optimization still eliminates the majority of redundant RESCALEs.

The stacked diff D95243636 adds a follow-on `EliminateRescaleBeforeMulPass` that absorbs residual INT32→INT32 RESCALEs before MUL ops.

## Vela INT32→INT32 RESCALE Limitation (Follow-up)

The Vela NPU compiler (Ethos-U55) cannot handle INT32→INT32 RESCALE:
- `EthosU55Constraints::SupportsRescale()` rejects types > 16 bits
- `GraphIrOptimiser::RewriteRescale()` decomposes rejected RESCALEs into MUL ops with explicit OFM scaling
- For INT32→INT32, the MUL's right-shift (typically 20-40 bits) zeros out the result
- `EliminateTosaRescale()` only handles Conv/MatMul patterns, not standalone RESCALEs
- Python-side `rewrite_rescale()` has no code path for INT32→INT32 with non-Conv predecessor

A follow-up Vela patch can fix `RewriteRescale()` to properly handle INT32→INT32 RESCALE (likely after the INT16 conversion step). Once Vela is fixed, this pass can be updated to also fuse non-identity pairs.

## Numerical Analysis

| Source | Magnitude | Mitigation |
| ------ | --------- | ---------- |
| **A**: Fixed-point decomposition non-associativity | ~1 INT8 step | Handled via `qtol=1` in tests |
| **B**: INT8 clamping bypass on identity removal | up to 120 INT8 steps | Bounded; handled via `qtol=1` in tests |

Reviewed By: 3l1

Differential Revision: D94483331

### DIFF
--- a/backends/arm/_passes/fuse_consecutive_rescales_pass.py
+++ b/backends/arm/_passes/fuse_consecutive_rescales_pass.py
@@ -1,0 +1,175 @@
+# Copyright 2026 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import logging
+from typing import cast, Set, Type
+
+import torch
+from executorch.backends.arm._passes.arm_pass import ArmPass
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass, PassResult
+from torch.fx import GraphModule, Node
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+# TOSA RESCALE argument positions:
+#   args[0] = input tensor (Node)
+#   args[1] = output dtype (e.g., torch.int8, torch.int32)
+#   args[2] = scale list (List[float]; per-tensor when len == 1)
+#   args[3] = input zero point (int)
+#   args[4] = output zero point (int)
+_ARG_INPUT = 0
+_ARG_OUTPUT_DTYPE = 1
+_ARG_SCALE = 2
+_ARG_INPUT_ZP = 3
+_ARG_OUTPUT_ZP = 4
+
+
+class FuseConsecutiveRescalesPass(ArmPass):
+    """Fuse consecutive RESCALE(INT32->INT8/INT16) -> RESCALE(INT8/INT16->INT32)
+    pairs.
+
+    InsertRescaleInt32Pass wraps each quantized arithmetic and comparison
+    operator (add, sub, mul, abs, eq, ge, gt, le, lt, max, min, sum) with
+    input rescales (INT8/INT16->INT32) and an output rescale
+    (INT32->INT8/INT16). When two such ops are chained (e.g., add1 -> add2),
+    the output rescale of add1 feeds directly into an input rescale of add2,
+    creating a redundant INT32->INT8/INT16->INT32 round-trip that loses
+    precision.
+
+    This pass detects such pairs and handles two cases:
+
+    - **Identity** (composed scale ~1.0, matching zero points): Removes both
+      RESCALEs and directly wires R1's input to R2's users.  This eliminates
+      the entire round-trip.  Bypassing the intermediate INT8/INT16 clamp can
+      in theory cause up to ~120 INT8 steps of output difference when all
+      inputs are near the clamp boundary; in practice, observed differences
+      are 0-1 steps for typical distributions.  Tests use qtol=1.
+
+    - **Non-identity**: Leaves the pair unchanged.  The Vela NPU compiler
+      cannot correctly process INT32->INT32 RESCALE (produces all-zero NPU
+      outputs), so non-identity pairs retain their INT8/INT16 intermediate.
+
+    Handles multi-user R1 nodes: when R1 feeds both RESCALE and
+    non-RESCALE users, each R1->R2 RESCALE pair is fused individually
+    while preserving R1 for its non-RESCALE users.
+
+    """
+
+    _passes_required_after: Set[Type[ExportPass]] = set()
+
+    def call(self, graph_module: GraphModule) -> PassResult:
+        graph = graph_module.graph
+        modified = False
+        rescale_before = sum(1 for n in graph.nodes if _is_rescale(n))
+        identity_pairs_fused = 0
+
+        for node in list(graph.nodes):
+            node = cast(Node, node)
+            if not _is_fuseable_r1(node):
+                continue
+
+            r1_input = node.args[_ARG_INPUT]
+            r1_input_zp = node.args[_ARG_INPUT_ZP]
+            r1_scale = float(node.args[_ARG_SCALE][0])  # type: ignore[arg-type]
+
+            node_fused = False
+            for user in list(node.users):
+                if _try_fuse_identity_pair(node, user, r1_input, r1_input_zp, r1_scale):
+                    node_fused = True
+                    identity_pairs_fused += 1
+
+            if node_fused:
+                modified = True
+
+        if modified:
+            graph.eliminate_dead_code()
+            rescale_after = sum(1 for n in graph.nodes if _is_rescale(n))
+            removed = rescale_before - rescale_after
+            logger.info(
+                "FuseConsecutiveRescalesPass: removed %d identity pairs "
+                "(%d RESCALEs: %d -> %d)",
+                identity_pairs_fused,
+                removed,
+                rescale_before,
+                rescale_after,
+            )
+            graph_module.recompile()
+            graph.lint()
+            # Note: we deliberately skip super().call() — retracing is
+            # unnecessary since this pass only rewires edges and removes
+            # nodes without introducing new operations.
+
+        return PassResult(graph_module, modified)
+
+
+def _is_rescale(node: Node) -> bool:
+    return (
+        node.op == "call_function"
+        and node.target == exir_ops.backend.tosa.RESCALE.default
+    )
+
+
+def _is_fuseable_r1(node: Node) -> bool:
+    """Check if node is an R1 candidate.
+
+    R1 is RESCALE(INT32 -> INT8/INT16) with per-tensor scale.
+
+    """
+    if not _is_rescale(node):
+        return False
+    if node.args[_ARG_OUTPUT_DTYPE] not in (torch.int8, torch.int16):
+        return False
+    if len(node.args[_ARG_SCALE]) != 1:  # type: ignore[arg-type]
+        return False
+    r1_input = node.args[_ARG_INPUT]
+    if not isinstance(r1_input, Node) or "val" not in r1_input.meta:
+        return False
+    if r1_input.meta["val"].dtype != torch.int32:
+        return False
+    return True
+
+
+def _try_fuse_identity_pair(
+    r1: Node,
+    r2: Node,
+    r1_input: Node,
+    r1_input_zp: int,
+    r1_scale: float,
+) -> bool:
+    """Try to fuse an R1->R2 identity pair.
+
+    Returns True if fused.
+
+    """
+    if not _is_rescale(r2):
+        return False
+    if r2.args[_ARG_OUTPUT_DTYPE] != torch.int32:
+        return False
+    if r1.args[_ARG_OUTPUT_ZP] != r2.args[_ARG_INPUT_ZP]:
+        return False
+    if len(r2.args[_ARG_SCALE]) != 1:  # type: ignore[arg-type]
+        return False
+
+    r2_scale = float(r2.args[_ARG_SCALE][0])  # type: ignore[arg-type, index]
+    composed_scale = r1_scale * r2_scale
+    r2_output_zp = r2.args[_ARG_OUTPUT_ZP]
+
+    if abs(composed_scale - 1.0) < 1e-6 and r1_input_zp == r2_output_zp:
+        # Identity case: remove both RESCALEs and directly wire
+        # R1's input (INT32) to R2's users.  The composed scale
+        # is ~1.0 so the round-trip is a no-op modulo the INT8
+        # clamp.  Bypassing the clamp can in theory cause up to
+        # ~120 INT8 steps of difference near clamp boundaries;
+        # observed differences are 0-1 steps.  Tests use qtol=1.
+        r2.replace_all_uses_with(r1_input)
+        return True
+
+    # Non-identity: leave the pair unchanged.  Creating a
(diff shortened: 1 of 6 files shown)


Ninja91:
This test is already failing: https://hud.pytorch.org/pytorch/executorch/commit/cb90f6b9a70a077e9002e45831ff9a6d33069559#65571852848-box

backends/arm/_passes/fuse_consecutive_rescales_pass.py
```diff
@@ -0,0 +1,121 @@
...
+    def call(self, graph_module: GraphModule) -> PassResult:
+        graph = graph_module.graph
+        modified = False
+        nodes_to_erase = []
+
+        for node in list(graph.nodes):
+            node = cast(Node, node)
+            if not _is_rescale(node):
+                continue
+
```
gggekov, line 73: If that pass is applicable for ADD/SUB/MUL nodes, don't you need to test if you are dealing with one of these operators ? Right now, i believe your pass would run for any operator, not just ADD/SUB/MUL

backends/arm/test/passes/test_fuse_quantized_activation_pass.py
```diff
@@ -0,0 +1,49 @@
...
+        quantize=True,
+        ops_before_pass={
+            "executorch_exir_dialects_edge__ops_aten_relu_default": 1,
+        },
+        ops_not_after_pass=[
+            "executorch_exir_dialects_edge__ops_aten_relu_default",
+        ],
+        pass_list=[FuseQuantizedActivationPass],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Can you please remove the 
pipeline.pop_stage("run_method_and_compare_outputs")  and do the element-wise numerical test? 
When changing the rescaling, it's easy to introduce numerical diffs, hence it's important to run with element-wise tests enabled.

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+    model = AddChain()
+    pipeline = PassPipeline[AddChain.input_t](
+        model,
+        model.get_test_inputs(),
+        quantize=True,
+        ops_not_before_pass={RESCALE_OP},
+        ops_after_pass={RESCALE_OP: 6},
+        pass_list=[FoldAndAnnotateQParamsPass, InsertRescaleInt32Pass],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Can you please remove the 
pipeline.pop_stage("run_method_and_compare_outputs")  and do the element-wise numerical test? 
When changing the rescaling, it's easy to introduce numerical diffs, hence it's important to run with element-wise tests enabled.

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    model = AddChain()
+    pipeline = PassPipeline[AddChain.input_t](
+        model,
+        model.get_test_inputs(),
+        quantize=True,
+        pass_list=[FoldAndAnnotateQParamsPass, InsertRescaleInt32Pass],
+        pass_functions=[assert_consecutive_rescales_exist],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Can you please remove the 
pipeline.pop_stage("run_method_and_compare_outputs")  and do the element-wise numerical test? 
When changing the rescaling, it's easy to introduce numerical diffs, hence it's important to run with element-wise tests enabled.

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+        model.get_test_inputs(),
+        quantize=True,
+        pass_list=[
+            FoldAndAnnotateQParamsPass,
+            InsertRescaleInt32Pass,
+            FuseConsecutiveRescalesPass,
+        ],
+        pass_functions=[assert_no_consecutive_rescales],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Can you please remove the 
pipeline.pop_stage("run_method_and_compare_outputs")  and do the element-wise numerical test? 
When changing the rescaling, it's easy to introduce numerical diffs, hence it's important to run with element-wise tests enabled.

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+            FoldAndAnnotateQParamsPass,
+            InsertRescaleInt32Pass,
+            FuseConsecutiveRescalesPass,
+        ],
+        pass_functions=[
+            assert_rescale_count_reduced,
+            assert_no_int8_to_int32_via_int8,
+        ],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Same as above about running the numerics tests

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    model = BranchingAdd()
+    pipeline = PassPipeline[BranchingAdd.input_t](
+        model,
+        model.get_test_inputs(),
+        quantize=True,
+        pass_list=[FoldAndAnnotateQParamsPass, InsertRescaleInt32Pass],
+        pass_functions=[assert_consecutive_rescales_exist],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Same as above

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,311 @@
...
+            FoldAndAnnotateQParamsPass,
+            InsertRescaleInt32Pass,
+            FuseConsecutiveRescalesPass,
+        ],
+        pass_functions=[
+            assert_no_consecutive_rescales,
+            assert_no_int8_to_int32_via_int8,
+        ],
+    )
+    pipeline.pop_stage("run_method_and_compare_outputs")
```
gggekov: Same as above

gggekov:
I like the idea you try to implement. Note that for the case of RESCALE/RESCALE and a single ADD, the Vela compiler team is doing refactoring to fuse the two rescales with the ADD and that should improve the performance considerably on some models on U55. 

I added the `ciflow/trunk` label to trigger the Arm unit tests, please make sure you pass all the unit tests.

3l1 (approved):

gggekov:
I ran the patch locally on my machine and `$ pytest -s --log-level=DEBUG backends/arm/test/models/test_lstm_arm.py -k test_lstm_u55_INT` produces only 0s in the output tensor. 
For the random input tensor, the ref output tensor is
```
 Ref output: [tensor([[[ 0.1257, -0.0977, -0.2514,  0.0186,  0.2700,  0.1350,  0.0605,
           0.2793,  0.0047,  0.4329,  0.3026, -0.2374,  0.2234,  0.0047,
          -0.1955,  0.0000,  0.0093,  0.0791, -0.1955,  0.4562],
         [-0.1955, -0.3491,  0.1676,  0.2514,  0.2839,  0.1303,  0.3072,
          -0.3584,  0.2141,  0.4887,  0.1676,  0.1536,  0.4515, -0.4050,
          -0.1722,  0.0512,  0.3119,  0.5493, -0.2514, -0.0326],
         [ 0.0233, -0.4003,  0.2653,  0.0047,  0.0977, -0.2234,  0.2420,
          -0.0791,  0.1629, -0.0047, -0.1862,  0.2095,  0.4143, -0.4143,
          -0.3677,  0.0419,  0.1629,  0.2700, -0.2002,  0.0977]],
.....
.....
```
and the test output tensor is 0s
```
Test output: [tensor([[[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]],

        [[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
         [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]],
....
....
```
In other words, for LSTM, the TOSA graph produces only 0s when ran on the FVP. I can see a speed-up from the patch, but i would vote to make sure the output tensor is meaningful.

Ninja91:
@gggekov will fix lint issues once i see other suites pass.

backends/arm/_passes/fuse_consecutive_rescales_pass.py
```diff
@@ -0,0 +1,174 @@
...
+import logging
+from typing import cast, List, Set, Type
+
+import torch
+from executorch.backends.arm._passes.arm_pass import ArmPass
+from executorch.exir.dialects._ops import ops as exir_ops
+from executorch.exir.pass_base import ExportPass
+from torch.fx import GraphModule, Node
+from torch.fx.passes.infra.pass_base import PassResult
```
oscarandersson8218: We typically use PassResult from exir.pass_base. Not sure if it matters at all though...

backends/arm/_passes/fuse_consecutive_rescales_pass.py
```diff
@@ -0,0 +1,174 @@
...
+                    node_fused = True
+                    identity_pairs_fused += 1
+
+            if node_fused:
+                nodes_to_erase.append(node)
+                modified = True
+
+        for node in nodes_to_erase:
+            if len(node.users) == 0:
+                graph.erase_node(node)
```
oscarandersson8218: What's the difference between this and `graph_module.graph.eliminate_dead_code()`?

backends/arm/_passes/fuse_consecutive_rescales_pass.py
```diff
@@ -0,0 +1,174 @@
...
+def _is_fuseable_r1(node: Node) -> bool:
+    """Check if node is an R1 candidate.
+
+    R1 is RESCALE(INT32 -> INT8/INT16) with per-tensor scale.
+    """
+    if not _is_rescale(node):
+        return False
+    if node.args[1] not in (torch.int8, torch.int16):
+        return False
+    if len(node.args[2]) != 1:  # type: ignore[arg-type]
```
oscarandersson8218: Some comments explaining what the args[i] is would be helpful for all node.args[i] comparisons.

backends/arm/test/passes/test_rescale_optimization.py
```diff
@@ -0,0 +1,689 @@
...
+    """
+    model = BranchingAdd()
+    pipeline = PassPipeline[BranchingAdd.input_t](
+        model,
+        model.get_test_inputs(),
+        quantize=True,
+        pass_list=[FoldAndAnnotateQParamsPass, InsertRescaleInt32Pass],
+        pass_functions=[assert_consecutive_rescales_exist],
+    )
+    pipeline.run()
```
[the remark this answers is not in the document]
oscarandersson8218: Agree with this comment.

backends/arm/test/passes/test_rescale_pass.py
```diff
@@ -81,15 +79,20 @@ def test_insert_rescale_u55_INT(test_data: input_t):
...
+    (exp→log→add→sub→mul→ sigmoid) accumulates rounding beyond the
+    default qtol=1.
     """
     module = RescaleNetwork()
     pipeline = EthosU85PipelineINT(
         module=module,
         test_data=test_data,
         aten_ops=[],
         exir_ops=[],
+        qtol=2,
```
oscarandersson8218: Why is `qtol=2` needed for U85 but not tosa_INT and U55?

oscarandersson8218 (changes requested):
Looks good on a high level. Mainly worried about the change of qtol in rescale-test.

oscarandersson8218:
If you fix the linting issues, I'm happy to approve this.

oscarandersson8218 (approved):

PR #17832, opened 2026-03-03
CUDA backend: fix Triton SDPA NaN with sparse boolean masks

mergennachin:
The Triton SDPA kernel produced NaN when boolean attention masks had
tile blocks where ALL entries were masked (False). This occurs with
sliding window / ring buffer patterns where a contiguous region of
KV positions is valid and the rest is masked — at block granularity
(BLOCK_N=32/64/128/256), some blocks fall entirely in the masked
region.

Root cause (both pow2 and non-pow2 kernels): the online softmax
computes `m_ij = max(m_i, max(qk))` per block, then
`p = exp(qk - m_ij)`. When all qk entries in a block are -inf
(fully masked), m_ij is -inf, and `qk - m_ij = -inf - (-inf) = NaN`.
This NaN propagates through `p`, into `acc` via `dot(p, v)`, and
corrupts all subsequent blocks' running softmax.

Fix: guard the subtraction with `tl.where(m_ij > -inf, qk - m_ij,
-inf)`, producing `exp(-inf) = 0` instead of `exp(NaN) = NaN` for
all-masked blocks. Same guard on the rescale factor `alpha =
exp(m_i - m_ij)` to prevent NaN when transitioning from an
all-masked block.

Also fixes `other=True` → `other=False` in the non-pow2 kernel's
mask load (line 184), which caused out-of-bounds mask positions to
be treated as "attend" instead of "don't attend".

Test: test_triton_sdpa_nan.py exports a ring buffer SDPA with sparse
bool masks (750 valid out of 1500 KV positions, matching streaming
audio encoder patterns) at various start positions including the
NaN-triggering start_pos=812. Verifies output is finite via pybind.

### DIFF
--- a/backends/cuda/tests/test_triton_sdpa_nan.py
+++ b/backends/cuda/tests/test_triton_sdpa_nan.py
@@ -1,0 +1,255 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Test Triton SDPA kernel with sparse boolean masks.
+
+Reproduces NaN bug when all entries in a block are masked (-inf - (-inf) = NaN
+in softmax). The fix guards exp(qk - m_ij) against all-masked blocks.
+
+Before fix: NaN at start_pos=812 (sparse ring buffer mask).
+After fix:  Finite output at all positions.
+
+Usage:
+    python -m pytest backends/cuda/tests/test_triton_sdpa_nan.py -v
+"""
+
+import glob
+import os
+import shutil
+import tempfile
+import unittest
+
+import torch
+import torch.nn as nn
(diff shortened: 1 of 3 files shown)


Gasoonjia (approved):

backends/cuda/tests/test_triton_sdpa_nan.py
```diff
@@ -0,0 +1,196 @@
...
+        cpu_inputs = [t.cpu() for t in inputs]
+        return mod.run_method("forward", cpu_inputs)
+
+
+# ---------------------------------------------------------------------------
+# Test: is_causal (baseline, no bool mask)
+# ---------------------------------------------------------------------------
+
+
+def test_causal_vs_eager():
```
Gasoonjia: cab we concentrate the tensor creation into a helper function? for every test we need to recreate qkv seems reduntant? or we can have a template function for every test

PR #17837, opened 2026-03-04
Arm backend: Add TOSA dialect shape ops 

oscarandersson8218:
Adds TOSA dialect shape ops. All added shape-ops are currently not serialized and are only tested through isolated tests with FakeTensor-inputs. To enable serialization of shape ops in the future, dim-order propagation of these ops is added.

cc @digantdesai @SS-JIA @freddan80 @per @zingo @mansnils @Sebastian-Larsson @robell


zingo (approved):
Added file is under test so should not cause buck2 problem

oscarandersson8218:
Unrelated failures.

PR #17839, opened 2026-03-04
Arm backend: Update Vela to 5.0.0

Erik-Lundell:
- Remove now deprecated --debug-force-regor from docs
- Remove some now passing xfails

### DIFF
--- a/backends/arm/arm_vela.py
+++ b/backends/arm/arm_vela.py
@@ -88,12 +88,7 @@ def vela_compile(
             args.append("--verbose-all")
         vela.main(" ".join(args).split(" "))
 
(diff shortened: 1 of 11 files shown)


zingo:
Hi @SS-JIA bumping Vela might need internal testing/updates. I think you install vela differently internally and need to keep it in sync for tests to pass, but could be wrong.

SS-JIA:
Will import and if tests pass we can merge this now and we can prepare a vela update for after it lands

zingo (approved):
OK to merge from Arm point of view

SS-JIA (approved):
Internal tests look ok! Good to merge whenever.

PR #17840, opened 2026-03-04
fix(qnn): fix OOM and Dequantize type error when compiling sharded LLMs

### LINKED ISSUE #17782: Qwen3-1.7B to SA8295 fails in dockerized build: num_sharding=1 hits QNN graph memory limit, num_sharding=2 causes host OOM
I’m converting qwen3-1_7b for SA8295 in a dockerized ExecuTorch environment.

I see two different failures depending on num_sharding:

1. num_sharding=1: Fails during QNN graph finalization/serialization with limit error:
graph requires estimated allocation of 2309576 KB, limit is 2097152 KB
Failed to finalize Qnn Graph with error: 1002
AssertionError: Failed to generate Qnn context binary.

2. num_sharding=2
The process runs out of host memory (OOM), even on a large machine.

Environment:
Deployment style: Dockerized
Host OS: Linux (KVM VM)
CPU: Intel Xeon Platinum 8375C, 128 vCPUs
RAM: ~495 GiB total, no swap
Target SoC: SA8295
Model: Qwen3-1.7B
Executorch version: v1.1.0

local modification (static_llm_quant_recipe.py: 575-609):
- Set default_quant_dtype = QuantDtype.use_16a8w
- Updated the conv2d quantization to: QuantDtype.use_16a8w
- Updated granularity=QuantGranularity.PER_CHANNEL (instead of per-block)

Command used: 
python examples/qualcomm/oss_scripts/llama/llama.py \
  -b build-android \
  -m SA8295 \
  --decoder_model qwen3-1_7b \
  --model_mode hybrid \
  --prefill_ar_len 128 \
  --max_seq_len 256 \ # Tried 128, 256, 512, 1024 
  --temperature 0 \
  --prompt "I would like to learn python, could you teach me with a simple example?" \
  --tasks wikitext \ 
  --limit 1 \
  --compile_only

My current hypothesis is that the num_sharding=1 failure is likely a platform limitation on SA8295 (DSP arch v68), given the explicit QNN graph serialization cap (2309576 KB required vs 2097152 KB limit).

However, the num_sharding=2 behavior is confusing: instead of resolving the graph-size issue, it leads to host-side OOM even on a 500 GB RAM-class system. I’m not sure whether this is expected compile-time memory behavior, a tooling inefficiency, or a potential memory leak. Also, other model with num_shardings > 1 are showing the same OOM behavior (Llama3.2-3B).

[convert_qwen3-1_7b_SA8295_20260302_082259.txt](https://github.com/user-attachments/files/25684871/convert_qwen3-1_7b_SA8295_20260302_082259.txt)
[convert_qwen3-1_7b_SA8295_20260302_065355.txt](https://github.com/user-attachments/files/25684870/convert_qwen3-1_7b_SA8295_20260302_065355.txt)

cc @cccclai @winskuo-quic @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

FouadSakr:
This PR fixes two bugs in the QNN backend that caused compilation to fail for sharded LLMs (e.g. Qwen3-1.7B with `num_sharding=2`).

**Bug 1 — `insert_io_qdq.py`: infinite loop → OOM**

`_insert()` iterated over `graph_module.graph.nodes` live while inserting new DQ nodes. The newly inserted DQ nodes inherited `QCOM_QUANT_ATTRS` and still fed into `output`, causing them to be re-processed on the next iteration, leading to an infinite loop and OOM.

Fix: snapshot the node list before iterating with `list(graph_module.graph.nodes)`.

**Bug 2 — `llm_wrappers.py`: wrong op dialect in `_tag_ios`**

`TagQuantIO._tag_ios()` used `exir_ops.edge.aten.select.int` (ATen dialect) to detect freq sin/cos tensors at shard boundaries. After `to_edge()`, all ops are the `_copy` variant, so the check always failed. Affected `select_copy` nodes never received the `QCOM_QUANTIZED_IO` tag, causing `InsertIOQDQ` to wrongly insert a DQ node after them, which QNN then rejected with `q::Dequantize could not create op` on `QuantUint16_T`.

Fix: use `exir_ops.edge.aten.select_copy.int`.

Neither fix affects non-sharded models (`num_sharding=1`).

Fixes #17782 

cc @cccclai @winskuo-quic @shewu-quic @haowhsu-quic @DannyYuyang-quic @cbilgin

### DIFF
--- a/backends/qualcomm/_passes/insert_io_qdq.py
+++ b/backends/qualcomm/_passes/insert_io_qdq.py
@@ -118,7 +118,10 @@ class InsertIOQDQ(ExportPass):
                     user.replace_input_with(node, inserted_node)
 
     def _insert(self, graph_module: torch.fx.GraphModule) -> torch.fx.GraphModule:
-        for n in graph_module.graph.nodes:
+        # Snapshot the node list before any insertions to avoid visiting
+        # newly-inserted Q/DQ nodes and triggering infinite re-insertion.
+        nodes = list(graph_module.graph.nodes)
+        for n in nodes:
             # do nothing when a node is expected to output a quant tensor
             if n.meta.get(QCOM_QUANTIZED_IO):
                 continue
--- a/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
+++ b/examples/qualcomm/oss_scripts/llama/wrappers/llm_wrappers.py
@@ -425,7 +425,7 @@ class TextDecoder(Component):
         }
 
         freq_op = {
-            exir_ops.edge.aten.select.int,
+            exir_ops.edge.aten.select_copy.int,
         }
         quant_io_type = None
 


abhinaykukkadapu:
@FouadSakr thanks so much for debugging the bug and for the fixes. Although we triggered `num_sharding` paths before for gemma models we haven't hit this issue probably and this might be hitting because of the second one at the IO boundary which is causing infinite loop and is causing bug 1. Changes looks good to me.

CC: @shewu-quic @cccclai

abhinaykukkadapu (approved):

haowhsu-quic:
May I have a deeper look at this one? Will reply ASAP.

haowhsu-quic:
May I know the ET commit you're using? I cannot reproduce the infinite loop in my side with ET latest main.
I'm using the quantization recipe & sharding number as mentioned in #17782.

FouadSakr:
@haowhsu-quic, I’m using v1.1.0 tag for my experiments.

haowhsu-quic:
Thank you for providing the reference. I've tried v1.1.0 but still could not reproduce. 
Hi @abhinaykukkadapu, may I know if you observed the same issue?

abhinaykukkadapu:
> Thank you for providing the reference. I've tried v1.1.0 but still could not reproduce. Hi @abhinaykukkadapu, may I know if you observed the same issue?

Hi @haowhsu-quic, i haven't reproduced it myself, probably @FouadSakr can you please let us know what command you used and any specific config modifications?

FouadSakr:
> > Thank you for providing the reference. I've tried v1.1.0 but still could not reproduce. Hi @abhinaykukkadapu, may I know if you observed the same issue?
> 
> Hi @haowhsu-quic, i haven't reproduced it myself, probably @FouadSakr can you please let us know what command you used and any specific config modifications?

We can close this issue. It was actually another modification from my end which caused this edge case to happen. 

Modification in q_dq_map that caused the issue:
`q_dq_map = {
        # per tensor (quantize -> dequantize)
        exir_ops.edge.quantized_decomposed.quantize_per_tensor.default: exir_ops.edge.quantized_decomposed.dequantize_per_tensor.tensor,
        exir_ops.edge.quantized_decomposed.quantize_per_tensor.tensor: exir_ops.edge.quantized_decomposed.dequantize_per_tensor.tensor,
        # per tensor (dequantize -> dequantize, for nodes with dequantize encoding)
        exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default: exir_ops.edge.quantized_decomposed.dequantize_per_tensor.default,
        exir_ops.edge.quantized_decomposed.dequantize_per_tensor.tensor: exir_ops.edge.quantized_decomposed.dequantize_per_tensor.tensor,
        # per channel (quantize -> dequantize)
        exir_ops.edge.quantized_decomposed.quantize_per_channel.default: exir_ops.edge.quantized_decomposed.dequantize_per_channel.default,
        # per channel (dequantize -> dequantize, for nodes with dequantize encoding)
        exir_ops.edge.quantized_decomposed.dequantize_per_channel.default: exir_ops.edge.quantized_decomposed.dequantize_per_channel.default,
    }`

nil-is-all:
Thanks for letting us know about the cause, @FouadSakr. Appreciate you investigating and filing a PR nevertheless. Feel free to review the [good-first-issue](https://github.com/pytorch/executorch/issues?q=state%3Aopen%20label%3A%22good%20first%20issue%22) label for any contributions you'd like to take up!

PR #17841, opened 2026-03-04
Arm backend: add phi3-mini layer tests

xingguo01:
- F32/F16 pass
- INT8 xfails partially
- BF16 passes for TOSA, runtime not supported for VGF


zingo:
@SS-JIA this add some file but as they are tests I don't expect thet to give any buck2 problem, do you still wan to check?

SS-JIA (approved):
Should be ok!

PR #17856, opened 2026-03-04
Fix for Raspberry Pi 5 GPU

sneakyHulk:
### Summary
The current check for VK_IMAGE_TILING_LINEAR (row-major order) only verifies vkCreateImage(...) succeeds, but Vulkan requires an additional capability VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT. For images that will be sampled by shaders (e.g. for conv2d) VK_IMAGE_TILING_OPTIMAL should be used.

Fixes #17855

### Test plan
converted some models with the VulkanPartitioner() and let it run on the Raspberry Pi 5 -> using VK_IMAGE_TILING_LINEAR will run but produces wrong results, as the Debug info suggests. Also using VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT fixes the issue as before (see https://github.com/pytorch/executorch/pull/7615)

### DIFF
--- a/backends/vulkan/runtime/vk_api/Adapter.cpp
+++ b/backends/vulkan/runtime/vk_api/Adapter.cpp
@@ -123,12 +123,6 @@ VkDevice create_logical_device(
     defined(ETVK_INSPECT_PIPELINES)
       VK_KHR_PIPELINE_EXECUTABLE_PROPERTIES_EXTENSION_NAME,
 #endif /* VK_KHR_pipeline_executable_properties && ETVK_INSPECT_PIPELINES */
-#ifdef VK_KHR_cooperative_matrix
-      VK_KHR_COOPERATIVE_MATRIX_EXTENSION_NAME,
-#endif /* VK_KHR_cooperative_matrix */
-#ifdef VK_NV_cooperative_matrix2
-      VK_NV_COOPERATIVE_MATRIX_2_EXTENSION_NAME,
-#endif /* VK_NV_cooperative_matrix2 */
   };
 
   std::vector<const char*> enabled_device_extensions;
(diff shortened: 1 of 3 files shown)


backends/vulkan/runtime/vk_api/Adapter.cpp
```diff
@@ -123,12 +123,6 @@ VkDevice create_logical_device(
     defined(ETVK_INSPECT_PIPELINES)
       VK_KHR_PIPELINE_EXECUTABLE_PROPERTIES_EXTENSION_NAME,
 #endif /* VK_KHR_pipeline_executable_properties && ETVK_INSPECT_PIPELINES */
-#ifdef VK_KHR_cooperative_matrix
```
SS-JIA, line 126: why were these removed?

backends/vulkan/runtime/vk_api/Adapter.cpp
```diff
@@ -375,10 +365,6 @@ void Adapter::submit_cmd(
   VK_CHECK(vkQueueSubmit(device_queue.handle, 1u, &submit_info, fence));
 }
 
-void Adapter::override_device_name(const std::string& new_name) {
```
SS-JIA, line 378: ditto, this should be kept - it is used for some tests in Meta internal CI 😛

SS-JIA (changes requested):
Overall, changes look good - thanks for the fix!

RC only because it seems some things were removed that aren't relevant to the stated purpose of these changes.

backends/vulkan/runtime/vk_api/Adapter.cpp
sneakyHulk: sry was starting from the release version

SS-JIA (approved):
LGTM!

PR #17858, opened 2026-03-04
Sortformer cuda backend support

Gasoonjia:


.github/workflows/cuda.yml
```diff
@@ -134,40 +129,50 @@ jobs:
...
-              name: "Voxtral-Mini-4B-Realtime-2602"
-            quant: "quantized-int4-weight-only"
+            - repo: "Qwen"
+              name: "Qwen3-0.6B"
+            - repo: "nvidia"
+              name: "parakeet-tdt"
+            - repo: "nvidia"
+              name: "diar_streaming_sortformer_4spk-v2"
+            - repo: "nvidia"
+              name: "diar_streaming_sortformer_4spk-v2.1"
```
larryliu0820, line 148: Do we need to do both?

Gasoonjia: that's what readme covers and i think we should

Gasoonjia:
close to make https://github.com/pytorch/executorch/pull/17865 land

PR #17866, opened 2026-03-05
Return Result for toTensor(), toOptional()

lucylq:
Existing toTensor() and toOptional<T>() methods abort via ET_CHECK_MSG on type mismatch, causing a denial of service.

Introduce tryToTensor and tryToOptional methods that return Result. Use them at the following callsites
- parseListOptionalType (tensor_parser.h)
- FreeCall handler in Method::execute_instruction (method.cpp)
- parseTensorList (tensor_parser_exec_aten.cpp)

The follow-ons from this are a little cumbersome:
- Add tryTo for the scalars, so toOptional works for all types (not just Tensors)
- Update codegen to use tryTo methods. This likely causes a binary size regression, as we are generating more code for each kernel arg. We can potentially limit this to tensors only.

### DIFF
--- a/runtime/core/evalue.h
+++ b/runtime/core/evalue.h
@@ -8,6 +8,7 @@
 
 #pragma once
 #include <executorch/runtime/core/exec_aten/exec_aten.h>
+#include <executorch/runtime/core/result.h>
 #include <executorch/runtime/core/tag.h>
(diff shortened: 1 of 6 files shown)


lucylq:
@claude review

lucylq:
@claude  review

PR #17897, opened 2026-03-05
[executorch][nvidia][tensorrt][23/n] Add benchmark infrastructure

shoumikhin:
* #17899
* #17898
* __->__ #17897
* #17896
* #17895
* #17894
* #17893
* #17892
* #17891
* #17890
* #17889
* #17888
* #17887
* #17886
* #17885
* #17884
* #17883
* #17882
* #17881
* #17880
* #17879
* #17878
* #17877
* #17876
* #17875

Add benchmark runners and infrastructure for measuring TensorRT inference performance including latency and throughput metrics.

Differential Revision: [D93275038](https://our.internmc.facebook.com/intern/diff/D93275038/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -807,6 +807,12 @@ if(EXECUTORCH_BUILD_PTHREADPOOL AND EXECUTORCH_BUILD_CPUINFO)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/extension/threadpool)
 endif()
 
+# TensorRT examples (benchmark, etc.) need extension_module and extension_tensor,
+# so they must be included after those targets are defined above.
+if(EXECUTORCH_BUILD_TENSORRT)
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/examples/nvidia)
+endif()
+
 if(EXECUTORCH_BUILD_KERNELS_TORCHAO)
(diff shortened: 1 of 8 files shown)


shoumikhin:
Closing this. It targets a backend that does not live in this repository.

The benchmark it adds is built against TensorRT directly:

```cpp
#include <NvInfer.h>
#include <NvOnnxParser.h>
```

and its CMake whole-archives a `tensorrt_backend` target:

```cmake
"SHELL:LINKER:--whole-archive $<TARGET_FILE:tensorrt_backend> LINKER:--no-whole-archive"
add_dependencies(benchmark tensorrt_backend)
```

No such target exists here. There is no TensorRT backend anywhere in the tree:

```
$ git ls-tree -r --name-only main | grep -ic tensorrt
0
```

`backends/` contains `cuda` but no `tensorrt`. The TensorRT integration is developed in
pytorch/TensorRT, so a benchmark for it cannot build from this repository as written.

It also no longer merges: four conflicts against current `main`, in `CMakeLists.txt`,
`setup.py`, `install_utils.py`, and `examples/models/mobilenet_v2/model.py`.

If NVIDIA benchmarking is still wanted, it belongs next to the backend it measures. For the CUDA
half specifically, `examples/cuda/` already exists here and would be the natural home. Closing
this so the open list reflects what is actually in progress.

PR #17908, opened 2026-03-05
Use direct function pointers for et_copy_index and et_view in prim_ops

lucylq:
Replace unnecessary lambda wrappers with direct function pointers for
et_copy_index and et_view kernel registrations. Both functions already
match the OpFunction signature exactly.

No stripped binary size change (GCC already optimized the forwarding
lambdas into tail calls), but the code is cleaner.


lucylq:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #17909
* __->__ #17908

rascani (approved):

PR #17910, opened 2026-03-05
Arm backend: Move clamp to independent quantization in annotator

JakeStevens:
Summary:
The clamp operation was incorrectly placed in `_one_to_one_shared_input_qspec`,
which causes the input and output observers to be shared. This is problematic
because clamp explicitly modifies the value range by enforcing min/max bounds.

When using clamp to prevent undefined behavior (e.g., clamping inputs to rsqrt
to be positive), the pre-clamp and post-clamp ranges can be very different.
With shared observers, the pre-clamp (smaller) values dominate the min_val,
causing incorrect quantization parameters for the post-clamp tensor.

This fix moves clamp to `_one_to_one`, giving it independent input/output
quantization so each observer properly tracks its respective range.

Differential Revision: D92408418

### DIFF
--- a/backends/arm/_passes/fold_qdq_with_annotated_qparams_pass.py
+++ b/backends/arm/_passes/fold_qdq_with_annotated_qparams_pass.py
@@ -11,6 +11,7 @@ from typing import cast, Optional, Set, Type
 import torch
 from executorch.backends.arm._passes import ArmPass
 from executorch.backends.arm._passes.arm_pass_utils import (
+    create_node,
     get_param_tensor,
     is_param_node,
     set_node_arg,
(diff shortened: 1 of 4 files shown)


AdrianLundell:
Hi, looks like the clamp-unittests start failing with this change. To make this work you would need to insert a rescale after the clamp operator to rescale the output to the new scale, currently it only clamps the quantized values, i.e [0, 1, 2, 3, 4, 5] clamped to quantized values [2,4] becomes [2,2,2,3,4,4], no scaling. Would be great to support this behavior with some config as well though!

JakeStevens:
@AdrianLundell  I've updated the quantize clamp pass now to insert a rescale *before*. This is to ensure that the clamp value does not get "swamped" by the input being clamped.

AdrianLundell:
Looks good, just need to check this internally!

JakeStevens:
Looks like arm cortex m size test is failing, I'm guessing due to the new rescale. Is it okay to bump in this PR?

AdrianLundell:
I would say so yes. But thinking about this more, if you are doing a rescale to the output range does the clamp even do anything? I would think that the range gets clamped already after the rescale, similarly to how clamps are fused to convs for example.

zingo:
The size tests was broken for a short while last week due to a totally unrelated PR that bumped the code size slightly in the libs, so regarding the size tests I would just start we a "Update branch"/rebase to see if it goes away.

JakeStevens:
> I would think that the range gets clamped already after the rescale

This is a good point-- and it means that this changes the behavior from clamp to rescale.

That is fundamentally shifting intent.

So, I do not think this is correct.

If we want the behavior described, we should insert a rescale instead, or have our own pass to convert the clamp to a rescale.

Abandoning this PR

PR #17326, opened 2026-02-10
Qualcomm AI Engine Direct - CDSP Direct Mode

winskuo-quic:
### Summary
- Support CDSP direct mode by defining ExecuTorch's customized rpc protocol.

- We have validated this PR with the following spec:
  - QNN 2.42.0
  - Hexagon SDK 6.4.02
  - Hexagon Tools Root 19.0.04 (This is tool chain under Hexagon SDK 6.4.02)
  - V79 & v81 device

- Please refer to README file under `backends/qualcomm/runtime/backends/direct_mode/README.md` for setup. Please be aware of the Note section if you observe the total execution time is slower the traditional mode.

#### Example Script
To build:
`backends/qualcomm/scripts/build.sh --enable_hexagon`

To run traditional mode (Same as usual):
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedOperator.test_qnn_backend_adaptive_avg_pool2d --model SM8750  --device $DEVICE --build_folder build-android`

To run direct mode (add --direct_build_folder build-hexagon):
`python backends/qualcomm/tests/test_qnn_delegate.py -k TestQNNQuantizedOperator.test_qnn_backend_adaptive_avg_pool2d --model SM8750  --device $DEVICE --build_folder build-android --direct_build_folder build-hexagon/`

### Test plan
add `--direct_build_folder build-hexagon/` at end of any `TestQNNQuantizedUtils`, `TestQNNQuantizedModel`, `TestQNNFloatingPointModel`, `TestQNNFloatingPointOperator` 

Author: @haowhsu-quic, @winskuo-quic

PR #17940, opened 2026-03-05
Arm backend: Add TOSA PAD dialect op

oscarandersson8218:
Adds new TOSA dialect op for PAD. aten.constant_pad_nd nodes are replaced by tosa.PAD in RewritePadPass. op_constant_pad_nd node visitor is replaced by a simpler node vistior for tosa.PAD.

cc @digantdesai @SS-JIA @freddan80 @per @zingo @mansnils @Sebastian-Larsson @robell


zingo:
Hi @SS-JIA this adds a file but I also spot `srcs = glob(["op_*.py", "ops_*.py"]), ` so I think it should be fine.

oscarandersson8218:
cortex_m_size_test failures are not introduced by this patch. Probably introduced by https://github.com/pytorch/executorch/pull/17326.

SS-JIA (approved):
Should be ok!

PR #17943, opened 2026-03-05
NXP backend: Fix kernel selection for Neutron converter, add missing …

roman-janik-nxp:
…condition

### Summary
Hotfix for missing condition, code was executed when it shouldn't and added clutter in logs.

### Test plan
All NXP backend tests.

cc @robert-kalmar


robert-kalmar (approved):

MartinPavella:
Just to be clear, this PR only removes the error message, right? Because the build is still hanging.

MartinPavella (approved):

PR #17981, opened 2026-03-06
Add BackendCache interface for backend init caching

rascani:
### Summary
Backends can persist expensive init artifacts (packed weights, compiled graphs) and restore them on subsequent runs, skipping the processed data load entirely. On cache hit the runtime skips loading the processed data segment, avoiding both the I/O and the init computation.

The runtime calls init_from_cache() before init() when a BackendCache is provided. Backends that don't support caching return NotSupported (the default), and the runtime falls through to the normal init() path. Any other error is treated as fatal.

BackendCache (abstract, in runtime/) takes (backend_id, delegate_index, key) as separate parameters rather than concatenating them into a single string. This keeps string manipulation and heap allocation out of the runtime layer entirely. DelegateBackendCache is a non-inheriting wrapper that captures the scoping components at construction time and presents a simple key-only API to backends via BackendInitContext::get_cache().

FileBackendCache (in extension/) maps entries to filesystem paths:
  {cache_dir}/{backend_id}/{delegate_index}/{key}

It uses aligned allocation (operator new with align_val_t and nothrow) so backends can request SIMD-aligned buffers, atomic rename for crash safety, and std::remove for cache entry deletion.

The BackendCache* flows through Module::load_method -> Program::load_method -> Method::load -> Method::init, where each delegate gets its own DelegateBackendCache placement-new'd from the method allocator.

### Test plan
Integration tests in backend_integration_test.cpp verify the cache-first init protocol end-to-end using the StubBackend and real .pte fixtures.

### DIFF
--- a/docs/source/design/backend-cache.md
+++ b/docs/source/design/backend-cache.md
@@ -1,0 +1,324 @@
+# Backend Init Caching
+
+## Problem
+
+Backend initialization in ExecuTorch can be expensive. Backends like XNNPACK
+repack weights during `init()`, which involves parsing the processed data
+segment, allocating aligned buffers, and transforming tensor layouts. For large
+models this takes hundreds of milliseconds on every load — even though the
+output is identical across runs on the same device.
+
+Today there is no way to persist these artifacts. Every `Program::load_method()`
+call pays the full initialization cost.
+
+## Design
+
+### Goals
+
+- Let backends persist init artifacts (packed weights, compiled graphs) and
+  restore them on subsequent loads, skipping the processed data segment entirely.
+- Keep the runtime layer (`runtime/`) free of string manipulation, heap
+  allocation, and filesystem I/O.
+- Provide a concrete filesystem-backed implementation in `extension/` that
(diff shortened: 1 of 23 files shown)


docs/source/design/backend-cache.md
```diff
@@ -0,0 +1,324 @@
...
+                                                           ┌──────────────────┐
+                                                           │ FileBackendCache │
+                                                           │ (extension/)     │
+                                                           └──────────────────┘
+```
+
+### Call flow
+
+The runtime attempts `init_from_cache()` before loading the processed data
+segment. This means a cache hit avoids both the data load and the init
```
JacobSzwejbka, line 64: One pattern thats pretty common is having both weights and graph in the compiled blob. Where the weights are then preprocessed and the graph is used as is. In that case I would still need to load both the cache (preprocessed weights) and the original blob (to get the graph). I guess the pattern could just be anything you need put in the cache but it does seem a bit wasteful of space to keep the original things around.

PR #17982, opened 2026-03-06
Backend test suite: Wire skip_patterns into pytest collection

rascani:
### Summary

TestFlow already has skip_patterns and should_skip_test(), but they were dead code — never called from conftest.py. Add a pytest_collection_modifyitems hook that applies pytest.mark.skip when a flow's skip_patterns match the test name. This activates skip_patterns for all flows (CoreML, Vulkan, etc.).

### DIFF
--- a/backends/test/suite/conftest.py
+++ b/backends/test/suite/conftest.py
@@ -4,9 +4,24 @@ from typing import Any
 import pytest
 import torch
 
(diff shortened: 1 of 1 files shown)


GregoryComer (approved):
Thanks, I think this was an oversight when I migrated to pytest.

PR #18002, opened 2026-03-09
Qualcomm AI Engine Direct - Support EPS for cli and fix mv2 to work in QNN2.45

winskuo-quic:
### Summary
- Mobile_Vit_V2 HTP previously only works in QNN2.29. In next QNN release, this issue should be resolved with the eps value provided.
- Support parser eps value for cli.py. This feature is mainly for internal testing purpose.

### Test plan
- MV2 gets good accuracy.
- Passing cli.py 

cc @cccclai @cbilgin


abhinaykukkadapu (approved):
LGTM

PR #18003, opened 2026-03-09
Arm backend: Simplify get_compile_spec signature

martinlsm:
Reduce the number of args to `get_compile_spec` from eight to one by forwarding program args object instead of eight of its members.


zingo (approved):

martinlsm:
Failures are unrelated

PR #18008, opened 2026-03-09
Fix AttributeError when torchao ops are not registered

elpdumont:
Summary:
The `_QUANT_PRIMITIVES` list initialization in `util.py` catches `ImportError` when torchao is not installed, but fails to catch `AttributeError` when torchao is installed but the operators (`dequantize_affine`, `quantize_affine`, `choose_qparams_affine`) are not registered.

This causes import failures in environments where torchao is available but an older version is used that doesn't have these operators registered.

The fix adds `AttributeError` to the exception handler alongside `ImportError`.

Differential Revision: D95735505


digantdesai:
Abandoned

PR #18011, opened 2026-03-09
Arm backend: Create more general GetDecompositionPass

tom-arm:
* Add GetDecompositiobPass for passes that use get_decomposition()

Change-Id: Ifd1ec40dda3d225a84dc37c0c60bac7051d7e5bf

### DIFF
--- a/backends/arm/_passes/get_decomposition_pass.py
+++ b/backends/arm/_passes/get_decomposition_pass.py
@@ -1,0 +1,116 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+# Copyright 2026 Arm Limited and/or its affiliates.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from typing import Set, Type
+
+import torch
+from executorch.backends.arm._passes import ArmPass
+from executorch.exir.pass_base import ExportPass, PassResult
+from torch._decomp import get_decompositions
+from torch._ops import OpOverload
+from torch.fx.experimental.proxy_tensor import make_fx
+
+
+class GetDecompositionPass(ArmPass):
+
+    _passes_required_after: Set[Type[ExportPass]] = set()
+
+    targeted_ops: list[OpOverload] = []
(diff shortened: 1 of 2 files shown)


zingo (approved):

zingo:
@SS-JIA this adds a file :)

backends/arm/_passes/get_decomposition_pass.py
```diff
@@ -0,0 +1,116 @@
...
+
+            # refer to pytorch/test/test_decomp.py
+            decomposed_module = make_fx(
+                node.target,
+                decomposition_table=get_decompositions(self.targeted_ops),  # type: ignore[arg-type]
+                tracing_mode="fake",
+                _allow_non_fake_inputs=False,
+            )(*input_tensors)
+
+            with graph_module.graph.inserting_before(node):
```
digantdesai, line 72: I wonder if there is any existing fx utility which does such substitute.

digantdesai (approved):
Seems like a refactor. Stamping it to unblock you. Left a comment.

backends/arm/_passes/get_decomposition_pass.py
```diff
@@ -0,0 +1,116 @@
...
+            if self._skip_pass(input_tensors):
+                continue
+
+            # refer to pytorch/test/test_decomp.py
+            decomposed_module = make_fx(
+                node.target,
+                decomposition_table=get_decompositions(self.targeted_ops),  # type: ignore[arg-type]
+                tracing_mode="fake",
+                _allow_non_fake_inputs=False,
+            )(*input_tensors)
```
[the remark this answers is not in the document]
tom-arm: It is true it's not as generic as it could be, but I would rather this was made more generic as we add more passes. Rather than trying to make the perfect initial implementation.

backends/arm/_passes/get_decomposition_pass.py
tom-arm: After a bit of a look around, I could not find a utility function to do this unfortunately

tom-arm:
Failures appear unrelated

PR #18037, opened 2026-03-10
Arm backend: Remove evaluation from aot_arm_compiler.py

martinlsm:
Model evaluation, which means to compute and present a model's top-1/top-5 accuracy on ImageNet, as well as some other useful metrics, is to be moved into a new python program.

Remove the evaluation feature from aot_arm_compiler.py, which is part of the migration.


zingo (approved):
Nice

PR #18010, opened 2026-03-09
Arm backend: Fix to ensure nested control flow graphs are delegated

SaoirseARM:
### Summary

- Added function to ensure True/False come in the right order.
- Update to quantizer to ensure submodules are converted first.
- Update to scalar pass for nested conditions.

cc @digantdesai @SS-JIA @freddan80 @per @zingo @oscarandersson8218 @mansnils @Sebastian-Larsson @robell

PR #18038, opened 2026-03-10
Portable kernel fixes for sabaton tests

JakeStevens:
Summary:
When running in Sabaton Zephyr Posix tests, the compiler complains for things like

```
foo >= 0 && foo < bar
```

when foo is unsigned.

Also avoid bringing in aten

Reviewed By: rascani

Differential Revision: D95834771


rascani (approved):

JakeStevens:
failing tests present on other PR like #18010, merging

PR #18039, opened 2026-03-10
Arm backend: Make aot_arm_compiler.py functions importable

martinlsm:
Calibration was tied to the ImageNet evaluator, so when the evaluation
feature was removed from aot_arm_compiler.py, custom calibration of a
quantized model was no longer possible.

This patch reimplements quantization calibration into
aot_arm_compiler.py, but it is instead controlled by a flag called
--calibration_data. This flag can target a file that is either a common
image format, a serialized file (.pt), or a directory which will make
the program search for applicable files within the directory.


zingo (approved):

zingo:
Fail with [trunk / test-mcu-cortex-m-backend / linux-job (push)](https://github.com/pytorch/executorch/actions/runs/22951066799/job/66615578575?pr=18039) is unrelated, that test sometimes times out on main branch testing also.

PR #18040, opened 2026-03-10
Remove uses_dotslash label

IanChilds:
Summary: Doesn't seem to be needed, prevents it running on RE and being cached

Differential Revision: D95928348


IanChilds:
@SS-JIA, mind approving please?

PR #18067, opened 2026-03-10
Implement a portable adaptive_avgpool2d kernel

GregoryComer:
### Summary
Add a basic _adaptive_avgpool2d kernel to portable. It shows up in some CV models.


kernels/portable/functions.yaml
```diff
@@ -17,6 +17,11 @@
 # See the README.md file in this directory for a description of the syntax used
 # by this file.
 
+- op: _adaptive_avg_pool2d.out
```
Gasoonjia, line 20: why do we need a underscore at the front? is this kernel internal only and won;t expose to user?

Gasoonjia (approved):

kernels/portable/functions.yaml
GregoryComer: Good question. This is actually the ATen op name, so this is referencing the official schema here: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L12443. I'm not sure why the ATen op itself starts with an underscore.

PR #18075, opened 2026-03-10
Revise ethos doc links in CMakeLists.txt

psiddh:
Updated documentation links for Ethos-U memory modes.


GregoryComer (approved):
Does it make sense to add a redirect? I'll defer to your judgement, just thinking out loud.

zingo:
Hi @psiddh / @GregoryComer  I don't think this made it into 1.2 releaase branch as it was taken already the 10:th should it?

PR #18078, opened 2026-03-10
[executorch] Propagate device metadata from partitioner result onto TensorSpecs

Gasoonjia:
* #18080
* #18328
* #18079
* __->__ #18078

Add end-to-end device type annotation support from export to runtime. Currently we only support one device per graph

The overall pipeline is:
a. Partitioner use `compile_spec` to determine which device the partitoned blob is runing on
b. after lowered partitioned graph to backend, the new-introed propagate_device_pass will annotate the input and output tensors of delegate blob as target device.

Differential Revision: [D95842511](https://our.internmc.facebook.com/intern/diff/D95842511/)


exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
+        changed = False
+        for node in graph_module.graph.nodes:
+            if node.op == "call_function" and node.target == executorch_call_delegate:
+                lowered_module = _get_lowered_module(graph_module, node)
+                if lowered_module is None:
+                    continue
+
+                result = _get_target_device_from_compile_specs(lowered_module)
```
digantdesai, line 153: This effectively assumes that we know the device 'name' AoT. In theory, we can have a multi-device delegate then the runtime might interpret this name differently and that can cause some confusion i.e `cuda:0` device on Metal.

I am not sure about using generic names like 'gpu' but also not sure about following PyTorch's eager/jit style naming convention where you won't switch devices underneath.

Gasoonjia: May I have your suggestions on the executorch device name?

Currently we set up the device name AOT and intentionally decouple dour device attribute with pytorch/pytorch device concept; we created a enum in the etensor schema for all devices we are supporting right now. In this way we can support as much as device as we want. 

For the situaton you mentioned, if other backend like vulken need its own gpu device, they should add a new one to the enum. We should avoid using generic names like 'gpu'.

JacobSzwejbka: Multi-device graph serialization will necessitate multiple graphs. We can maybe make an exception for input tensors, but for any intermediate the runtime needs to know what the device its loading intermediates onto. 

Device is fixed at export aot. If you want to have some generic shader style lib where the gpu type is decided lazily then you will have to use a generic key like gpu.

lucylq (approved):

exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,210 @@
...
+        (dt for dt in schema.DeviceType if dt.name.lower() == type_str),
+        None,
+    )
+    if device_type is None:
+        valid = ", ".join(dt.name for dt in schema.DeviceType)
+        raise ValueError(f"Unknown device type '{type_str}'. Valid types: {valid}")
+    return device_type, device_index
+
+
+def _get_lowered_module(
```
JacobSzwejbka, line 53: nit: this type of util really should be placed in a single spot. There are other things like this in the passes. Lets take it as a follow up to have claude just search for generic utils like this and centralize them

exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,210 @@
...
+    return None
+
+
+def _set_device_on_spec(
+    spec: TensorSpec,
+    device_type: schema.DeviceType,
+    device_index: int = 0,
+) -> None:
+    """Set the device attribute on a TensorSpec."""
+    spec.device = device_type
```
JacobSzwejbka, line 93: Are these fields already in the TensorSpec class definition? Are they initialized to just cpu and 0?

exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,210 @@
...
+    3. Getitem nodes that extract from a delegate call inherit the device from
+       the delegate call's output spec at the corresponding index.
+    """
+
+    def call(self, graph_module: torch.fx.GraphModule) -> PassResult:
+        changed = False
+        for node in graph_module.graph.nodes:
+            if node.op == "call_function" and node.target == executorch_call_delegate:
+                lowered_module = _get_lowered_module(graph_module, node)
+                if lowered_module is None:
```
JacobSzwejbka, line 146: We should throw here no?

exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,210 @@
...
+                    continue
+
+                result = _get_target_device_from_compile_specs(lowered_module)
+                if result is None:
```
JacobSzwejbka, line 154: Why does it not return cpu by default

exir/passes/propagate_device_pass.py
```diff
@@ -0,0 +1,210 @@
...
+                    "PropagateDevicePass: set device=%s on delegate node %s "
+                    "(backend=%s)",
+                    target_device_type,
+                    node.name,
+                    lowered_module.backend_id,
+                )
+
+        # Second pass: propagate device through getitem nodes that extract
+        # individual outputs from a delegate call.
+        for node in graph_module.graph.nodes:
```
JacobSzwejbka, line 186: Can we just do 1 pass. You can look at users of the delegate node to find the getitem nodes.

exir/passes/propagate_device_pass.py
Gasoonjia: Yes and yes.

exir/passes/propagate_device_pass.py
Gasoonjia: The default value for every tensor is `cpu:0`. If backend author didn't set the device-related compile spec, every tensor should be remain default value and no need to reset as cpu

exir/passes/propagate_device_pass.py
Gasoonjia: Let me throw here. I don't think it will be None.

exir/passes/propagate_device_pass.py
Gasoonjia: Yes we can but i feel like two passes are more structural: one is specific for delegate input and the other is for delegate output

PR #18082, opened 2026-03-11
Qualcomm AI Engine Direct - SeqMSE coarse-to-fine grid search

abhinaykukkadapu:
Replace brute-force linear grid search in SeqMSE with a two-phase coarse-to-fine approach. The first phase sweeps 100 evenly-spaced steps over [0.01, 1.0], the second phase refines with remaining budget in a ±0.02 window around the coarse best.

The `num_candidates` parameter now acts as a total evaluation budget: the first min(num_candidates, 100) evaluations are coarse, the rest are fine. With the default `seq_mse_candidates=150`, this gives 100 coarse + 50 fine steps — same total evaluations, better allocation of compute toward the optimum.

See the gh issue: https://github.com/pytorch/executorch/issues/18065 for more info

cc @cccclai @cbilgin


backends/qualcomm/_passes/seq_mse.py
```diff
@@ -42,9 +42,10 @@ def __init__(
         self.nominal_weight = nominal_weight
         self.nominal_bias = nominal_bias
         self.observer = observer
-        self.steps = torch.linspace(
-            1 / num_candidates, 1, steps=num_candidates
-        ).tolist()
+        num_coarse = min(num_candidates, 100)
```
haowhsu-quic: I wonder if we could introduce extra arguments like the `lower_bound`, `upper_bound` and iterate `num_candidates` steps within the range. I think it would be more straightforward and prevent hardcoded constants.

abhinaykukkadapu: `num_candidates` is still the right terminology as this represents overall mse steps, 150 for example. We consistently use first 100 as the coarse sweep and the remaining as fine sweep around the selected candidate.

We can expose this as `num_coarse_mse_candidates`, `num_fine_mse_candidates` but not sure if anyone uses it, as it is deep into the mse algorithmic territory.

haowhsu-quic: I see, how about using `num_candidates` for both `coarse_steps` & `fine_steps` for simplicity?
And maybe have the interval value `0.02` as an argument in constructor?

abhinaykukkadapu: That makes sense, we can even do, 50 candidates and 0.02 uniform for both coarse and fine, there is nothing technically limiting other than the trade off of having more/less candidates. If we do 100 that would be 200 candidates total which i think is unnecessary, i think we can do 50 for coarse and 50 for fine.

haowhsu-quic (approved):
This is awesome, thank you.

cccclai (approved):
Stamp on behalf of @haowhsu-quic

PR #18088, opened 2026-03-11
Add recurrent gated delta rule custom op for Qwen3.5 attention

Phineas1500:
## Summary

This PR adds a fused `llama::recurrent_gated_delta_rule` custom op and wires Qwen3.5 GatedDeltaNet attention to use it instead of the Python per-token recurrence loop when the op is available.

It also tightens local custom-op loading so we no longer implicitly scan repo-local `cmake-out*` directories, and adds coverage for recurrent-state correctness, chunked prefill behavior, and export graph selection.

## What changed

- added `llama::recurrent_gated_delta_rule` runtime and AOT registrations
- updated Qwen3.5 GatedDeltaNet attention to use the fused op with Python fallback preserved
- tightened `custom_ops_aot_lib` discovery:
  - default to package-local discovery
  - allow explicit override via `EXECUTORCH_CUSTOM_OPS_AOT_LIB`
  - removed implicit repo-local `cmake-out*` scanning
- added tests for:
  - recurrent op parity vs reference
  - `.out` variant behavior
  - chunked-state parity vs full-sequence execution
  - custom-op vs fallback attention parity
  - tiny Qwen3.5 export selecting `llama.recurrent_gated_delta_rule`

### Linux CPU-only (aarch64)

Built `custom_ops_aot_lib` successfully and loaded it via `EXECUTORCH_CUSTOM_OPS_AOT_LIB`.

Passed:
- `pytest extension/llm/custom_ops/test_update_cache.py::RecurrentGatedDeltaRuleTest -q`
  - `3 passed`
- `pytest examples/models/llama/tests/test_qwen3_5_attention.py -q`
  - `7 passed`
- `pytest examples/models/llama/tests/test_export_llama_lib.py::ExportLlamaLibTest::test_tiny_qwen35_export_uses_recurrent_gated_delta_rule -q`
  - `1 passed`

### Real-model CPU validation

On a real `Qwen3.5-0.8B` CPU run, fused recurrence matched the fallback path on next-token selection with very small logit drift, and improved eager prefill latency on the tested prompt.

Observed on local CPU validation:
- same next token from fused path vs fallback
- max logit diff on the order of `1e-5`
- eager prefill speedup about `1.6x` on the tested prompt

### Windows note

A local Windows-only FFHT/MSVC workaround was used during development to keep the local build usable, but that workaround is intentionally **not** included in this PR.

## Non-goals / separate issues

I did not treat the local `program.fbs` serialization issue as part of this change.

This branch does not modify `exir/_serialize/*` or `schema/program.fbs`, and serialization-focused checks passed on both this branch and clean `main` once the local environment was set up correctly.

A separate end-to-end tiny Qwen3.5 `.pte` export probe hit:
- `RuntimeError: Missing out variants: {'aten::alias'}`

That appears to be a separate pre-existing export issue outside this change set.


Phineas1500:
@lucylq this is a PR for optimizing the recurrence in Qwen 3.5, which we discussed here: https://github.com/pytorch/executorch/pull/17801#issuecomment-4014796663

I'm next going to make a PR for quantization. I'll let you know once that's up.

extension/llm/custom_ops/op_tile_crop_aot.cpp
```diff
@@ -51,5 +78,5 @@ TORCH_LIBRARY_IMPL(preprocess, CompositeExplicitAutograd, m) {
   m.impl("tile_crop", torch::executor::native::tile_crop_aten);
   m.impl(
       "tile_crop.out",
-      WRAP_TO_ATEN(torch::executor::native::tile_crop_out_no_context, 2));
```
JacobSzwejbka, line 54: what problem did you run into with WRAP_TO_ATEN?

JacobSzwejbka:
Can you tell me a bit more about the serialization issue you ran into as well as the MSVC one?

examples/models/llama/attention.py
```diff
@@ -668,6 +701,18 @@ def _recurrent_gated_delta_rule(
...
+                query,
+                key,
+                value,
+                g,
+                beta,
+                self.recurrent_state[:batch_size],
+            )
+            return core_attn_out.transpose(1, 2).contiguous().to(initial_dtype)
+
         core_attn_out = torch.zeros(
```
JacobSzwejbka, line 801: can you put this logic in some function called like "naive_gated_delta_rule_op" and then just have the if statement switch between them to tidy this function up a bit.

extension/llm/custom_ops/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ endif()
 
 set(_common_compile_options
     $<$<CXX_COMPILER_ID:MSVC>:/wd4996>
+    $<$<CXX_COMPILER_ID:MSVC>:/Zc:__cplusplus>
```
JacobSzwejbka: What codepath are you doing down that isnt triggering properly without this? Typically the c10 pattern is to just have explicit msvc conditions and not rely on the c++ version on windows iirc. I could be wrong on that though.

Phineas1500: you're right, i didn't need it. changed make_aten_functor_from_et_functor.h to use MSVC instead, and now it works without /Zc:__cplusplus

examples/models/llama/attention.py
Phineas1500: fixed it so _recurrent_gated_delta_rule() switches between _gated_delta_rule_op() and _naive_gated_delta_rule_op()

extension/llm/custom_ops/op_tile_crop_aot.cpp
Phineas1500: MSVC was failing for the accessed .out wrappers when building custom_ops_aot_lib in the WRAP_TO_ATEN instantiation path. I made those wrappers have explicit conversions, and it compiled cleanly on Linux/Windows

Phineas1500:
> Can you tell me a bit more about the serialization issue you ran into as well as the MSVC one?

I believe the serialization issue is separate from this PR, as I reproduced it here and in main. _prepare_schema() wants program.fbs to be under exir/_serialize, but in a normal source checkout, it's under schema/. The error is that it can't find program.fbs.

I had two separate local build issues on MSVC:
1. fht_avx.c uses inline asm that cl.exe doesn't accept. I kept my local workaround out of the PR because it's unrelated.
2. WRAP_TO_ATEN was failing for the touched .out wrappers, so I switched them to explicit conversions (ET to ATen) in this recent commit

I also dropped the `/Zc:__cplusplus` change from this PR.

Happy to answer more questions @JacobSzwejbka

digantdesai:
Taking a look.

extension/llm/custom_ops/op_sdpa.cpp
```diff
@@ -610,6 +676,133 @@ Tensor& sdpa_with_kv_cache_out(
...
+  const auto* key_data = key.const_data_ptr<float>();
+  const auto* value_data = value.const_data_ptr<float>();
+  const auto* g_data = g.const_data_ptr<float>();
+  const auto* beta_data = beta.const_data_ptr<float>();
+  auto* recurrent_state_data = recurrent_state.mutable_data_ptr<float>();
+  auto* output_data = output.mutable_data_ptr<float>();
+  std::vector<float> kv_mem(v_head_dim);
+  std::vector<float> delta(v_head_dim);
+
+  for (int64_t batch = 0; batch < batch_size; ++batch) {
```
digantdesai, line 739: This is a naive implementation, in terms of performance with all the loops etc. Do you have plans to improve this or add a new version and leave this as a reference C++ implementation?

digantdesai (approved):
Ok with merging this as is. Love the tests. Thanks.

digantdesai:
Let me try to merge this change.

nil-is-all:
Hi @Phineas1500, could you check the lintrunner logs and resolve the issues brought up? Once they are fixed, it should be good to merge.
cc @digantdesai

Phineas1500:
> Hi @Phineas1500, could you check the lintrunner logs and resolve the issues brought up? Once they are fixed, it should be good to merge. cc @digantdesai

Will do @nil-is-all !

Phineas1500:
@digantdesai made a commit addressing the linter issues. Hopefully should be good to merge after the tests run.

Phineas1500:
Apologies if I clicked the wrong button. Could a maintainer approve the twelve workflows?

cc @digantdesai @mergennachin

nil-is-all:
Hi @Phineas1500, approved the workflows to run. Could you take a look at the lintrunner error logs and resolve them?

Phineas1500:
@nil-is-all resolved the issues from the lintrunner logs

Phineas1500:
@nil-is-all do you think it's good to merge? I don't believe the current errors are related to this PR's code

Phineas1500:
@digantdesai @mergennachin is there anything else you think I should do? after this PR is merged, I'm thinking of making PRs adding q8da4w support, dynamic-shape export for DeltaNet, and better multi-token prefill

digantdesai:
Hey sorry, this is good to go. Let me run Claude through to catch nit issues. If something major comes up we should fix it else this is good.

digantdesai:
@claude Review this PR

digantdesai (approved):
Thanks, and sorry for the delay. I will try to merge it.

PR #18092, opened 2026-03-11
Arm backend: SymInt fixes

oscarandersson8218:
Avoid computations that collapses the symint representation and materializes the symbolic values to its hints.


zingo (approved):
The added file is a tests file and those are not handled with buck2 today so no need for extra buck2 review.

oscarandersson8218:
Unrelated failures.

PR #18097, opened 2026-03-11
Move Graph/ProgramBuilder to backends/test

oscarandersson8218:
Move GraphBuilder and ProgramBuilder to backends/test to make it accessible for other backends.


oscarandersson8218:
Hi, @SS-JIA, we would like to start using Program/GraphBuilder in the backends/arm/test  so I propose we move it to backends/test to make it accessible for other backends. I'm a bit unsure about the BUCK and bazel parts. Please add appropriate labels and reviewers.

oscarandersson8218:
@digantdesai Any movement on this?

hsharma35:
@oscarandersson8218 I'm looking into migrating the library in a way that our CI doesn't break. I'll add you as a reviewer once I'm done.

hsharma35:
https://github.com/pytorch/executorch/pull/18483 should work with internal CI, PTAL

PR #18098, opened 2026-03-11
Qualcomm AI Engine Direct - AMD backend error

winskuo-quic:
### Summary
We noticed that when performing inference with AMD CPU, we will run into `Floating point exception (core dumped)`.
This can be easily reproduced with following lines of code:

**import torch.nn as nn**
**import torch**
**w2_conv = nn.Conv2d(1536, 32, 1, bias=False)**
**x = torch.randn(1,1536,1,32)**
**w2_conv(x)**

Temp solution is to set mkldnn.enabled=False:
**torch.backends.mkldnn.enabled = False**

### Test plan
NA

cc @cccclai @cbilgin

### DIFF
--- a/backends/qualcomm/__init__.py
+++ b/backends/qualcomm/__init__.py
@@ -1,7 +1,9 @@
 import os
 
-from .scripts.download_qnn_sdk import install_qnn_sdk, is_linux_x86, QNN_ZIP_URL
+import cpuinfo
+import torch
 
+from .scripts.download_qnn_sdk import install_qnn_sdk, is_linux_x86, QNN_ZIP_URL
 
 env_flag = os.getenv("EXECUTORCH_BUILDING_WHEEL", "0").lower()
 # If users have preinstalled QNN_SDK_ROOT, we will use it.
@@ -27,3 +29,8 @@ if env_flag not in ("1", "true", "yes"):
                 "       export LD_LIBRARY_PATH="
                 "$QNN_SDK_ROOT/lib/x86_64-linux-clang/:$LD_LIBRARY_PATH"
             )
+
+info = cpuinfo.get_cpu_info()
+vendor = info.get("vendor_id_raw", "").lower()
+if "amd" in vendor:
+    torch.backends.mkldnn.enabled = False
--- a/backends/qualcomm/__init__.py
+++ b/backends/qualcomm/__init__.py
@@ -1,9 +1,19 @@
 import os
 
-import cpuinfo
 import torch
 
 from .scripts.download_qnn_sdk import install_qnn_sdk, is_linux_x86, QNN_ZIP_URL
+
+try:
+    import cpuinfo
+
+    info = cpuinfo.get_cpu_info()
+    vendor = info.get("vendor_id_raw", "").lower()
+    if "amd" in vendor:
+        torch.backends.mkldnn.enabled = False
+except ImportError:
+    raise ImportError("Please install the cpuinfo with pip install py-cpuinfo.")
+
 
 env_flag = os.getenv("EXECUTORCH_BUILDING_WHEEL", "0").lower()
 # If users have preinstalled QNN_SDK_ROOT, we will use it.
@@ -29,8 +39,3 @@ if env_flag not in ("1", "true", "yes"):
                 "       export LD_LIBRARY_PATH="
                 "$QNN_SDK_ROOT/lib/x86_64-linux-clang/:$LD_LIBRARY_PATH"
             )
-
-info = cpuinfo.get_cpu_info()
-vendor = info.get("vendor_id_raw", "").lower()
-if "amd" in vendor:
-    torch.backends.mkldnn.enabled = False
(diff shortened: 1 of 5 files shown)


winskuo-quic:
Hi @cccclai, @abhinaykukkadapu,
We have noticed that AMD CPU during AOT will run into the error: `Floating point exception (core dumped)`. This happens during inference, including nn.Module.
There's a sample in summary section to reproduce.
This PR is a quick workaround to fix the issue, but I am assuming if this is a AMD or Torch issue, placing these logic under QNN probably isn't the best option. 
Please have a look.
Thanks

digantdesai (approved):
I assume this is during eager model runs?

digantdesai:
Would you mind creating a ticket on PyTorch/PyTorch?

winskuo-quic:
> I assume this is during eager model runs?

Hi @digantdesai,
This issue happened in both Eager Model and Exported Program when we are calibrating the model.
I have created an issue ticket under pytorch/pytorch: https://github.com/pytorch/pytorch/issues/177227

abhinaykukkadapu:
@winskuo-quic there are bunch of failures, can we rebase, i will monitor and push if the CI is green. Thanks

winskuo-quic:
> @winskuo-quic there are bunch of failures, can we rebase, i will monitor and push if the CI is green. Thanks

@abhinaykukkadapu,
Thanks for the help to monitor the CI. I have rebased.
Also just FYI, I included a new library under `backends/qualcomm/scripts/build.sh`: `pip install py-cpuinfo`.
So if some CI's doesn't run `build.sh`, python might not be able to find the library.

backends/qualcomm/__init__.py
```diff
@@ -1,7 +1,9 @@
 import os
 
-from .scripts.download_qnn_sdk import install_qnn_sdk, is_linux_x86, QNN_ZIP_URL
+import cpuinfo
```
abhinaykukkadapu: Move this to lazily import within the `if "amd" in vendor:`

backends/qualcomm/__init__.py
```diff
@@ -27,3 +29,8 @@
                 "       export LD_LIBRARY_PATH="
                 "$QNN_SDK_ROOT/lib/x86_64-linux-clang/:$LD_LIBRARY_PATH"
             )
+
+info = cpuinfo.get_cpu_info()
+vendor = info.get("vendor_id_raw", "").lower()
+if "amd" in vendor:
```
abhinaykukkadapu: Please also consider surrounding this in try/except if there are any ImportError with logging to let user know to install py-cpuinfo

abhinaykukkadapu:
> > @winskuo-quic there are bunch of failures, can we rebase, i will monitor and push if the CI is green. Thanks
> 
> @abhinaykukkadapu, Thanks for the help to monitor the CI. I have rebased. Also just FYI, I included a new library under `backends/qualcomm/scripts/build.sh`: `pip install py-cpuinfo`. So if some CI's doesn't run `build.sh`, python might not be able to find the library.

Thanks @winskuo-quic, provided few suggestions, also we would want something like this: https://github.com/pytorch/executorch/blob/28f3cf3f82a5050cc50356b33823eb4f1f0de6f5/.ci/scripts/setup-openvino.sh#L47

If you can add it in this pr, otherwise i can take it up.

backends/qualcomm/__init__.py
winskuo-quic: I have moved all these into the try catch statement.

backends/qualcomm/__init__.py
winskuo-quic: Added a try catch statement.

winskuo-quic:
Hi @abhinaykukkadapu,
Thanks for reviewing and providing some valuable suggestions. I have added `executorch/backends/qualcomm/requirements.txt` and place all python libraries under this file. Please take a look and let me know if it anything is missing.

abhinaykukkadapu:
> Hi @abhinaykukkadapu, Thanks for reviewing and providing some valuable suggestions. I have added `executorch/backends/qualcomm/requirements.txt` and place all python libraries under this file. Please take a look and let me know if it anything is missing.

Thanks @winskuo-quic for addressing the comments, i think we missed one more spot to run the pip install requirements.

```
--- a/.ci/scripts/test_wheel_package_qnn.sh                                                       
  +++ b/.ci/scripts/test_wheel_package_qnn.sh                                                       
  @@ -176,6 +176,9 @@ run_core_tests () {                                                           
     "$PIPBIN" install . --no-build-isolation                                                       
     popd > /dev/null                                         
                                                                                                    
  +  # Install qualcomm backend dependencies                                                        
  +  "$PIPBIN" install -r "$REPO_ROOT/backends/qualcomm/requirements.txt"
  +                                                                                                 
     echo "=== [$LABEL] Import smoke tests ==="
     "$PYBIN" -c "import executorch; print('executorch imported successfully')"
     "$PYBIN" -c "import executorch.backends.qualcomm; print('executorch.backends.qualcomm imported 
  successfully')"
```

abhinaykukkadapu (approved):

PR #18099, opened 2026-03-11
Arm backend: Add pass to create CONST_SHAPEs

oscarandersson8218:
Add pass that creates CONST_SHAPEs for view_copy and repeat based on their list arguments.


zingo (approved):
Added file in under _passes that uses glob() so no need for any buck2 check.

Issue #17701, opened 2026-02-25
Align Arm quantizers and CortexMQuantizer


AdrianLundell:
## Overview 
The idea is to change the implementation of the `TOSAQuantizer` and the subclasses `EthosUQuantizer`/`VgfQuantizer` to be based on the same modular system as used for the Cortex-M backend. This has the following benefits:

1. **Improved configurability** - `CortexMQuantizer` allows custom annotation filters or even custom quantizers to allow perfect tailoring of quantization parameters.
2. **Improved visibility into the annotation process** - The `CortexMQuantizer` comes with a
`QuantizerReporter` which helps debugging the annotation process, and a single `quantizer_support` file which clearly defines supported operators.
3. **Code sharing** - By aligning the two quantizers users get a more predictable behaviour and both backends  will continuously benefit from each others improvements. 

## API
The function call API will stay consistent with the previous implementation, together with new functions exposing the new configuration possibilities. Any other known behaviour changes will be listed here:
* Input/output nodes dtypes are now determined by `set_global`, rather than the closest annotated node.
* Nodes with `SharedQspecs` will by default inherit dtype from its input rather than being set, see `SharedQspecQuantizer` in the detailed breakdown.

Sketch of new API:
```
qconfig1 = get_symmetric_quantization_config()  # Old way of creating quantization configs intact
qconfig2 = TOSAQuantizationConfig()             # TOSAQuantizationConfigs can also be created directly 

# Old API still intact
quantizer = TOSAQuantizer()
quantizer.set_global(qconfig1)
quantizer.set_module_name("sigmoid", qconfig2)

# New API function using a NodeFinder to filter out nodes, does the same thing but more flexible
node_finder = ModuleNameNodeFinder("sigmoid")   # Many more available, or create your own implementing the NodeFinder interface
quantizer.set_node_finder(node_finder, qconfig2)

# Third way of doing the same thing, even more flexible
pattern_matcher = PatternMatcher(TOSA_QUANTIZER_SUPPORT_DICT)  # TOSA_QUANTIZER_DICT is defined in the arm backend.
pattern_quantizer = PatternQuantizer(qconfig2, node_finder, pattern_matcher)
quantizer.add_quantizer(pattern_quantizer)
```

## Detailed breakdown
[...]

PR #18100, opened 2026-03-11
Arm backend: Add experimental support for new TOSAQuantizer

AdrianLundell:
Allows initializing TOSA/EthosU/Vgf quantizers with use_composable_quantizer=True to use a new implementation of the quantizer following the Cortex-M. See
https://github.com/pytorch/executorch/issues/17701 for more details.

- Creates a new temporary TOSAQuantizer API layer for switching between the two versions
- Adds a TOSAQuantizationConfig encapturing TOSA-specific qspec requirements for certain ops.
- Adds quantizer_support.py for defining what operators are supported by the quantizer.
- Align mark_node_as_annotated in cortex-m backend to TOSAQuantizer behaviour.
- Update quantizer reporter to handle TOSA qspecs as they are dynamically created.


zingo:
Hi @SS-JIA / @digantdesai this adds a file, do you want/need to check this? 
This is also something we would like to get into 1.2 if possible.

AdrianLundell:
Fails unrelated

backends/arm/quantizer/__init__.py
```diff
@@ -11,17 +11,34 @@
...
-    TOSAQuantizer,
-    VgfQuantizer,
-)
 
 # Used in tests
 from .arm_quantizer_utils import is_annotated  # noqa
 
+# Lazily import heavy quantizer classes to avoid circular imports with
+# Cortex-M quantization configs.
+_LAZY_EXPORTS = {
```
rascani: _LAZY_IMPORTS?

backends/arm/quantizer/quantizer_support.py
```diff
@@ -0,0 +1,180 @@
...
+        (torch.ops.aten.floor_divide.default,),
+        (torch.ops.aten.logit.default,),
+        (torch.ops.aten.glu.default,),
+        (torch.ops.aten.addmm.default,),
+        (torch.ops.aten.layer_norm.default,),
+        (torch.ops.aten.group_norm.default,),
+        (torch.ops.aten.sqrt.default,),
+        (torch.ops.aten.silu.default,),
+        (torch.ops.aten.silu_.default,),
+        (torch.ops.aten.logit.default,),
```
rascani: `torch.ops.aten.logit.default` listed twice (also at line 126).

backends/cortex_m/quantizer/quantizer_reporter.py
```diff
@@ -31,31 +31,66 @@
...
+            if qspec.quant_min != key.quant_min:
+                continue
+            if qspec.quant_max != key.quant_max:
+                continue
+            if qspec.qscheme != key.qscheme:
+                continue
+            if qspec.is_dynamic != key.is_dynamic:
+                continue
+            return val
+    return "UNREGISTRED_QSPEC"
```
rascani: Typo: UNREGISTERED_QSPEC

backends/cortex_m/quantizer/quantizer.py
```diff
@@ -128,20 +145,20 @@ class PatternQuantizer(Quantizer, QuantizerReporterUser):
...
+        self.quantization_config: QuantizationConfig | None = quantization_config
         self.node_finder: NodeFinder = node_finder
         self.pattern_matcher: PatternMatcher = pattern_matcher
 
     def get_quantizer_info(self):
         name = self.__class__.__name__
         targeted_nodes_description = str(self.node_finder)
         quantization_config_path = SUPPORTED_QCONFIGS.get(
-            self.quantization_config, "CUSTOM_QCONFIG"
+            self.quantization_config, "UNREGISTRED_QCONFIG"
```
rascani: Typo: "UNREGISTERED_QCONFIG"

backends/cortex_m/quantizer/quantizer_reporter.py
```diff
@@ -31,31 +31,66 @@
...
+                continue
+            if qspec.quant_min != key.quant_min:
+                continue
+            if qspec.quant_max != key.quant_max:
+                continue
+            if qspec.qscheme != key.qscheme:
+                continue
+            if qspec.is_dynamic != key.is_dynamic:
+                continue
+            return val
```
rascani, line 70: Should we add a check for qspec.ch_axis?

backends/arm/quantizer/TARGETS
```diff
@@ -17,9 +17,11 @@ runtime.python_library(
     deps = [
         ":arm_quantizer_utils",
         ":quantization_annotator",
+        ":quantizer_support",
         "//executorch/backends/arm:constants",
         "//executorch/backends/arm:ethosu",
         "//executorch/backends/arm:vgf",
+        "//executorch/backends/cortex_m/quantizer:quantizer",
```
rascani, line 23: Pulling this in exposed that the cortex_m/quantizer:quantizer is broken on main. Here's summary of what needs to change:

```
  backends/cortex_m/quantizer/TARGETS — add missing srcs and deps:
  python_library(
      name = "quantizer",
      srcs = [
          "__init__.py",
          "node_finders.py",       # MISSING
          "operator_configs.py", # NEEDS REMOVING
          "pattern_checkers.py",   # MISSING
          "pattern_matcher.py",    # MISSING
          "quantization_configs.py",
          "quantizer.py",
          "quantizer_reporter.py", # MISSING
          "quantizer_support.py",  # MISSING
      ],
      deps = [
          "//caffe2:torch",
          "//executorch/backends/arm:common",                        # MISSING
          "//executorch/backends/arm:constants",                     # MISSING
          "//executorch/backends/arm/quantizer:quantization_annotator",  # MISSING
          "//executorch/backends/arm/quantizer:quantization_config",
          "//pytorch/ao:torchao",
          "fbsource//third-party/pypi/tabulate:tabulate",           # MISSING
      ],
  )
```

I'll put together a separate PR, but feel free to incorporate into this one to unblock.

rascani (approved):

digantdesai:
I like the overall direction, and the Cortex-M quantizer reuse. Also the priority for the composition.
I will let RJ review and stamp, as I am still catching up.

One thing I would say is, add an e2e test with both Ethos and Cortex-M quantizers and run the test on FVP.

backends/arm/quantizer/__init__.py
AdrianLundell: This is a workaround since the import situation is a bit messy with imports across the cortex-m and arm backend. The idea is to clean this up when things have stabilized, I didn't want to move things in this commit to make the diff cleaner, I hope this is OK.

backends/arm/quantizer/quantizer_support.py
AdrianLundell: Done

backends/cortex_m/quantizer/quantizer_reporter.py
AdrianLundell: Done

backends/cortex_m/quantizer/quantizer.py
AdrianLundell: Done

backends/cortex_m/quantizer/quantizer_reporter.py
AdrianLundell: I'm not sure if it is that interesting to differ between qspecs with different ch_axis values, it seems like a rather internal value to me. But I'll follow up with a better system for getting qspec/qconfig string representation anyways.

backends/arm/quantizer/TARGETS
AdrianLundell: Done

AdrianLundell:
I hope this is what you had in mind for the test @digantdesai, we have also run all tests internally with the new quantizer set to true sucessfully. The idea is that we can let you and others try this out and give feedback and then there should be minimal disruption as we flip the flag to True as default.

backends/arm/quantizer/TARGETS
```diff
@@ -43,11 +44,22 @@ runtime.python_library(
     name = "arm_quantizer_utils",
     srcs = ["arm_quantizer_utils.py"],
     deps = [
+        "//caffe2:torch",
```
rascani, line 47:
```
        "//executorch/backends/arm:common",
        "//executorch/backends/arm:constants",
```

backends/arm/quantizer/arm_quantizer.py
```diff
@@ -359,59 +444,67 @@ class TOSAQuantizer(Quantizer):
...
+    def tosa_spec(self):
+        return self.quantizer.tosa_spec
+
+    @property
+    def compile_spec(self):
+        return self.quantizer.compile_spec
+
+    @property
+    def global_config(self):
+        return self.quantizer.global_config
```
rascani, line 476:
```
    @global_config.setter
    def global_config(self, value: Optional[QuantizationConfig]) -> None:
        self.quantizer.global_config = value

    @property
    def io_config(self):
        return getattr(self.quantizer, "io_config", None)

    @io_config.setter
    def io_config(self, value: Optional[QuantizationConfig]) -> None:
        self.quantizer.io_config = value

    @property
    def module_type_config(self):
        return getattr(self.quantizer, "module_type_config", {})

    @module_type_config.setter
    def module_type_config(
        self, value: Dict[Callable, Optional[QuantizationConfig]]
    ) -> None:
        self.quantizer.module_type_config = value

    @property
    def module_name_config(self):
        return getattr(self.quantizer, "module_name_config", {})

    @module_name_config.setter
    def module_name_config(
        self, value: Dict[str, Optional[QuantizationConfig]]
    ) -> None:
        self.quantizer.module_name_config = value
```

backends/cortex_m/quantizer/TARGETS
```diff
@@ -12,24 +12,35 @@ python_library(
...
+        "//executorch/backends/arm:constants",
+        "//executorch/backends/arm/quantizer:arm_quantizer_utils",
+        "//executorch/backends/arm/quantizer:quantization_annotator",
         "//executorch/backends/arm/quantizer:quantization_config",
         "//pytorch/ao:torchao",
+        "fbsource//third-party/pypi/tabulate:tabulate",
     ],
 )
 
+
```
rascani, line 35:
```
python_library(
    name = "quantizer_reporter",
    srcs = [
        "quantizer_reporter.py",
    ],
    deps = [
        "//caffe2:torch",
        "//pytorch/ao:torchao",
        "fbsource//third-party/pypi/tabulate:tabulate",
    ],
)

```

backends/cortex_m/quantizer/TARGETS
```diff
@@ -12,24 +12,35 @@ python_library(
...
 python_library(
     name = "quantization_configs",
     srcs = [
         "quantization_configs.py",
     ],
     deps = [
         "//caffe2:torch",
```
rascani, line 42:
```
        ":quantizer_reporter",
```

rascani (commented):
Ok, finally gotten it all the rest of the way through. Thank you for sorting out the lazy imports, that resolved a lot of it. There were still a few more buck rules that needed fixing up, but there were also usages of some Quantizer APIs that no longer existed. 

I think this should do it.

digantdesai (approved):

PR #18102, opened 2026-03-11
feat: added checks for support of activations on target

novak-vaclav:
### Summary

Introduces stricter checks for activation support on the target.
The updated validation logic applies to: `clamp`, `hardtanh`, `leaky_relu`, `relu`, `sigmoid`, `tanh`.

edit after rework:
Introduces stricter checks for support in cases where there might be `relu` or `relu6` operator alone in a partition. In such cases, the Neutron converter fails to delegate the operator. Also fixed a small issue in `hardtanh_converter` related to reading out the arguments from the edge operator.
This additional check applies for `relu_converter`, `clamp_converter` and `hardtanh_converter`. Also checked whether the issues does not affect other activations, such as `leaky_relu`, `tanh` and `sigmoid`, but empirically the issue did not occur there.

### Test plan

cc @robert-kalmar @JakeStevens @digantdesai @MartinPavella @roman-janik-nxp

### DIFF
--- a/backends/nxp/backend/ir/converter/node_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converter.py
@@ -5,6 +5,7 @@
 
 import operator
 from abc import ABC, abstractmethod
+from typing import Callable
 
 import torch
 
@@ -185,6 +186,35 @@ class NodeConverter(ABC):
 
         # Node not quantized
         return True
+
+    @staticmethod
+    def is_node_alone_in_partition(
+        node: Node, partition_list: list[Partition], filter_fn: Callable[[Node], bool]
+    ) -> bool:
+        """Return True if `node` is the only node in its partition for which `filter_fn`
+        returns True.
+
+        The function finds the unique partition containing `node` and applies
+        `filter_fn` to all nodes in that partition. If only one node passes the
+        predicate — and that node is `node` — the function returns True.
+
+        :param node: The torch.fx.Node to check.
+        :param partition_list: List of proposed partitions.
+        :param filter_fn: Predicate applied to nodes in the partition.
+                        `node` is considered alone if it is the only node
+                        for which this predicate returns True.
+        """
+        partitions = [p for p in partition_list if node in p.nodes]
+        if len(partitions) != 1:
+            raise ValueError(
+                "Cannot find a partition of a node in graph. This should not occur."
+            )
+
+        partition = partitions[0]
+        filtered_partition_nodes = list(filter(filter_fn, partition.nodes))
+        return (
+            len(filtered_partition_nodes) == 1 and filtered_partition_nodes[0] == node
+        )
 
     def assert_convertible(self, node):
         """Assert that the call `is_supported()` returns `True`. Otherwise, raise an exception and print an
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
@@ -12,6 +12,9 @@ from executorch.backends.nxp.backend.ir.converter.node_converter import (
 from executorch.backends.nxp.backend.ir.lib.tflite.BuiltinOperator import (
     BuiltinOperator,
 )
+from executorch.backends.nxp.backend.neutron_operator_support import (
+    activation_supported_on_target,
+)
 from executorch.backends.nxp.backend.neutron_target_spec import NeutronTargetSpec
 from torch.fx import Node
 from torch.fx.passes.infra.partitioner import Partition
@@ -76,19 +79,15 @@ class ClampConverter(NodeConverter):
     ) -> bool:
         bounds = cls._get_clamp_bounds(node)
 
+        # Neutron cannot delegate a partition where ReLU or ReLU6 is the only operator
+        # and at the same time the node does not satisfy delegation requirements.
+        # In contrast, ReLUN1To1 and ReLU0To1 are supported and delegated successfuly.
         if bounds in [cls.SUPPORTED_BOUNDS["Relu"], cls.SUPPORTED_BOUNDS["Relu6"]]:
-            # If this is the only operator in the partition, NeutronConverter will not create a NeutronNode for some
-            #  reason.
-            clamp_partitions = [p for p in partition_list if node in p.nodes]
-            if len(clamp_partitions) != 1:
-                return False  # Should never happen
-
-            clamp_partition = clamp_partitions[0]
-            non_q_dq_partition_nodes = list(
-                filter(is_not_qdq_node, clamp_partition.nodes)
+            is_alone_in_partition = cls.is_node_alone_in_partition(
+                node, partition_list, filter_fn=is_not_qdq_node
             )
-            if len(non_q_dq_partition_nodes) <= 1:
-                return False  # This would be the only node in the partition, which would cause a crash later on.
+            if is_alone_in_partition:
+                return activation_supported_on_target(node, neutron_target_spec)
 
         return True
 
--- a/backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
+++ b/backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
@@ -1,15 +1,21 @@
-# Copyright 2025 NXP
+# Copyright 2025-2026 NXP
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
 from executorch.backends.nxp.backend.ir.converter.node_converter import (
     CustomDelegationOptions,
+    is_not_qdq_node,
     NodeConverter,
+    Partition,
 )
 from executorch.backends.nxp.backend.ir.lib.tflite.BuiltinOperator import (
     BuiltinOperator,
 )
+from executorch.backends.nxp.backend.neutron_operator_support import (
+    activation_supported_on_target,
+)
+from executorch.backends.nxp.backend.neutron_target_spec import NeutronTargetSpec
 from torch.fx import Node
 from torch.nn import Parameter
 
@@ -17,12 +23,45 @@ from torch.nn import Parameter
 class HardTanhConverter(NodeConverter):
 
     # Maps possible input parameters of HardTanh to equivalent ReLU-based operators supported by TFLite.
-    supported_modes_map = {
+    SUPPORTED_MODES_MAP = {
         (0.0, 6.0): BuiltinOperator.RELU6,
         (-1.0, 1.0): BuiltinOperator.RELU_N1_TO_1,
         (0.0, 1.0): BuiltinOperator.RELU_0_TO_1,
         (0.0, float("inf")): BuiltinOperator.RELU,
     }
+
+    # Maps possible modes of HardTanh to equivalent ReLU bounds.
+    SUPPORTED_BOUNDS_MAP = {
+        "ReluN1To1": (-1.0, 1.0),
+        "Relu0To1": (0.0, 1.0),
+        "Relu6": (0.0, 6.0),
+        "Relu": (0.0, float("inf")),
+    }
+
+    @staticmethod
+    def _get_hardtanh_bounds(node: Node) -> tuple[float, float]:
+        args = node.args
+
+        match len(args):
+            case 1:
+                min_val = -1
+                max_val = 1
+
+            case 2:
+                min_val = args[1]
+                max_val = 1
+
+            case 3:
+                min_val = args[1]
+                max_val = args[2]
+
+            case _:
+                # should not occur
+                raise ValueError(
+                    f"Unexpected number of arguments for HardTanh node: {len(args)}"
+                )
+
+        return min_val, max_val
 
     @staticmethod
     def _is_supported_in_IR(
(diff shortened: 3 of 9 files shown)


backends/nxp/backend/neutron_operator_support.py
```diff
@@ -77,3 +82,24 @@ def transposition_is_supported_on_neutron(
...
+    input_shape = list(input_tensor(node, 0).shape)
+    if node.args[0].meta[NXP_NODE_FORMAT].is_channels_first():
+        input_shape = dims_to_channels_last(input_shape)
+
+    n = 1 if len(input_shape) == 1 else input_shape[0]
+    c = input_shape[-1]
+    num_macs = neutron_target_spec.get_num_macs()
+
+    # activations in Neutron are delegable only
+    # if `num_channels` % `num_macs` == 0 and `num_batches` == 1
```
MartinPavella: This is incorrect.

You call this function in every activation converter's `is_supported_on_target()` method. So this condition has to always be satisfied to delegate an activation. But operators which can use the `Custom Activation` feature of Neutron aren't restricted like this.

For example `test_leaky_relu_converter.py` contained tests with input shape (23, ), and it delegated successfully.

Your task was aimed only at solving the issue when the activations were the only ops in the model. But here you are introducing a check which effects every case (including models with multiple nodes), and it breaks existing working functionality.

I see you changed all existing tests to satisfy this condition. But the tests were passing, and they should still be passing if Neutron supports them.

Our goal is to delegate as many operators to Neutron as possible. Therefore, we cannot introduce arbitrary restrictions like this. Every single test that started failing when you introduced this change must still be passing. You cannot just modify the tests and pretend it didn't happen. If a test fails -> there is an issue. In this case, the issue was caused by your changes.

---
It appears you have misunderstood the task. In `clamp_converter.py`, you have removed the method `supports_partitioning_result()`. What was the motivation for this? Instead, you should have added this method to every "activation_converter" and used it to prohibit delegation if the activation was the only node in the partition (and perhaps consider some operator-specific conditions).

Unless I'm missing something, a complete rework is required.

MartinPavella (changes requested):

backends/nxp/backend/neutron_operator_support.py
novak-vaclav: As I told you and the team multiple times at the regular meetups, I noticed that the converters did NOT check whether the node is delegable. Such functionality was not implemented, when in my opinion it should have been. I also believe that the absence of such functionality was the root cause of `ReLU` not being delegated when there are no other nodes in the partition (aside from QDQ nodes of course).
In the Neutron Converter [documentation](https://bitbucket.sw.nxp.com/projects/AITEC/repos/neutron-software/browse/neutron-converter/docs/SupportedOperatorsC.md#Activation), there is a list of conditions, such as `num_channels % num_macs == 0` that are not being checked. That is the whole motivation behind all these changes.
The tests needed to be modified to correctly reflect the newly imposed restrictions. I wanted to adhere to the documentation the Neutron team provided, however if you think it is better to start checking the requirements only  when we encounter a situation when it stops working, that is fine by me. Just wanted to do the due dilligence of properly implementing the conversion of one of the most important operators there are, such as `ReLU` or `Sigmoid`.
As I am looking at it right now, some of the code is quite old, for example the functionality concerning `ReLU`. Maybe the restrictions did not exist back then.

MartinPavella: Let's not escalate this.

The bottom line is this:
- You wanted to adhere to the Neutron documentation, even in cases when we have working implementation which contradicts the documentation. So you prohibited these cases, which solved the issue of your ticket, but it also broke some supported cases. 
- I think we should support as many cases as possible. We know the Neutron documentation is not 100% accurate, so I think prohibiting working cases just because they contradict the docs is not the way to go. Based on this opinion, the solution should have been different.

We can discuss it tomorrow if you want.

P.S.
I didn't mean to sound aggressive in my first comment. I apologize if my phrasing was inappropriate.

novak-vaclav:
New PR description after major rework:

Introduces stricter checks for support in cases where there might be `relu` or `relu6` operator alone in a partition. In such cases, the Neutron converter fails to delegate the operator. Also fixed a small issue in `hardtanh_converter` related to reading out the arguments from the edge operator.
This additional check applies for `relu_converter`, `clamp_converter` and `hardtanh_converter`. Also checked whether the issues does not affect other activations, such as `leaky_relu`, `tanh` and `sigmoid`, but empirically the issue did not occur there.

backends/nxp/backend/neutron_operator_support.py
novak-vaclav: Did major rework, please review the changes.

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
```diff
@@ -1,47 +1,108 @@
...
+                max_val = 1
+
+            case 3:
+                min_val = args[1]
+                max_val = args[2]
+
+            case _:
+                # should not occur
+                min_val = 0
+                max_val = 1
```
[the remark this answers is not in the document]
MartinPavella: Good point

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
```diff
@@ -1,47 +1,108 @@
...
+    # Maps possible modes of HardTanh to equivalent ReLU bounds.
+    SUPPORTED_BOUNDS_MAP = {
+        "ReluN1To1": (-1.0, 1.0),
+        "Relu0To1": (0.0, 1.0),
+        "Relu6": (0.0, 6.0),
+        "Relu": (0.0, float("inf")),
+    }
+
+    @staticmethod
+    def _get_hardtanh_bounds(node: Node) -> tuple[int, int]:
```
MartinPavella: Nit: The return type should be `tuple[float, float]` as the bounds don't need to be strictly integers.

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
```diff
@@ -1,47 +1,108 @@
...
     @staticmethod
     def _is_supported_in_IR(
         node: Node,
         parameters_mapping: dict[str, Parameter],
         custom_delegation_options: CustomDelegationOptions,
     ) -> bool:
-        _, min_value, max_value = node.args
-        return (min_value, max_value) in HardTanhConverter.supported_modes_map.keys()
+        bounds = HardTanhConverter._get_hardtanh_bounds(node)
+        return bounds in HardTanhConverter.SUPPORTED_MODES_MAP.keys()
```
MartinPavella: Nit: The `.keys()` is redundant.

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
```diff
@@ -1,47 +1,108 @@
...
+        custom_delegation_options: CustomDelegationOptions,
+        neutron_target_spec: NeutronTargetSpec,
+        parameters_mapping: dict[str, Parameter],
+    ) -> bool:
+        bounds = HardTanhConverter._get_hardtanh_bounds(node)
+
+        if bounds in [
+            cls.SUPPORTED_BOUNDS_MAP["Relu"],
+            cls.SUPPORTED_BOUNDS_MAP["Relu6"],
+        ]:
```
MartinPavella, line 92: Nit: A comment explaining why only `Relu` and `Relu6`  would be useful.

backends/nxp/backend/ir/converter/node_converters/ops_converters/clamp_converter.py
```diff
@@ -77,18 +80,11 @@ def supports_partitioning_result(
         bounds = cls._get_clamp_bounds(node)
 
         if bounds in [cls.SUPPORTED_BOUNDS["Relu"], cls.SUPPORTED_BOUNDS["Relu6"]]:
```
MartinPavella, line 85: Nit: A comment explaining why only `Relu` and `Relu6`  would be useful.

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -186,6 +182,33 @@ def _has_shared_q_params_if_quantized(node: Node) -> bool:
...
+        node: Node, partition_list: list[Partition], filter_fn: Callable
+    ):
+        """Return True if `node` is the only node in its partition for which `filter_fn`
+        returns True.
+
+        The function finds the unique partition containing `node` and applies
+        `filter_fn` to all nodes in that partition. If only one node passes the
+        predicate — and that node is `node` — the function returns True.
+
+        :param node: The torch.Node to check.
```
MartinPavella: Nit: `torch.fx.Node`

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -186,6 +182,33 @@ def _has_shared_q_params_if_quantized(node: Node) -> bool:
...
+        :param partition_list: List of proposed partitions.
+        :param filter_fn: Predicate applied to nodes in the partition.
+                        `node` is considered alone if it is the only node
+                        for which this predicate returns True.
+        """
+        partitions = [p for p in partition_list if node in p.nodes]
+        if len(partitions) != 1:
+            return False  # Should never happen
```
MartinPavella: When this code was inside the `supports_partitioning_result`, returning `False` would result in not delegating the node. Here, it would have the opposite effect. Perhaps we should raise an exception here instead. What do you think?

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -181,6 +182,33 @@ def _has_shared_q_params_if_quantized(node: Node) -> bool:
...
+
+        partition = partitions[0]
+        filtered_partition_nodes = list(filter(filter_fn, partition.nodes))
+        return (
+            len(filtered_partition_nodes) == 1 and filtered_partition_nodes[0] == node
+        )
```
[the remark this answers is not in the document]
MartinPavella: Good point

backends/nxp/backend/ir/converter/node_converter.py
```diff
@@ -186,6 +182,33 @@ def _has_shared_q_params_if_quantized(node: Node) -> bool:
         # Node not quantized
         return True
 
+    @staticmethod
+    def is_node_alone_in_partition(
```
MartinPavella, line 191: Nice!👏🏻

MartinPavella (commented):
Very nice solution 👍🏻 
I pretty much only have cosmetic suggestions, and Copilot had a couple of good points. With those details implemented, LGTM.

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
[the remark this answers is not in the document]
MartinPavella: Good point

backends/nxp/backend/ir/converter/node_converter.py
novak-vaclav: Yes that true, I fixed it. Thank you

backends/nxp/backend/ir/converter/node_converter.py
novak-vaclav: Thank you 👍

novak-vaclav:
Fixed all the issues mentioned by @MartinPavella and Copilot. Please review. Thank you for your feedback! :D

MartinPavella (approved):
LGTM 👍🏻 
Once tests pass, I'll merge

MartinPavella:
@novak-vaclav the linting error is preventing the merge. Can you please fix it?

novak-vaclav:
Linting errors should fixed @MartinPavella and performed a rebase

novak-vaclav:
also fixed a small type issue found by copilot

backends/nxp/backend/ir/converter/node_converters/ops_converters/hardtanh_converter.py
```diff
@@ -1,47 +1,112 @@
...
+                min_val = args[1]
+                max_val = args[2]
+
+            case _:
+                # should not occur
+                raise ValueError(
+                    f"Unexpected number of arguments for HardTanh node: {len(args)}"
+                )
+
+        return min_val, max_val
```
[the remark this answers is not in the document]
novak-vaclav: fixed

MartinPavella:
@novak-vaclav unfortunately the rebase introduce some failures, so I cannot merge. Please take a loot at them.

novak-vaclav:
Pushed the correct changes and rebased onto main, please run the tests now @MartinPavella

MartinPavella (approved):

novak-vaclav:
Please @MartinPavella run the non-passing checks again, looking at the logs, all tests seem to be passing but for some reason the job timed out.
Otherwise the [internal tests](https://bamboo3.sw.nxp.com/browse/MLTECE-EXIGH109-8) passed.

novak-vaclav:
The `unittest` checks are still timing out. Should I wait before it is fixed (the issue with the checks timing out also appears in other PRs) or should it be merged? @MartinPavella

MartinPavella:
@novak-vaclav as the Linux unittests are passing and you didn't modify any non-NXP files, I don't see the need to wait for the windows unittest time-out fix.
I'm merging this PR.

PR #17821, opened 2026-03-03
NXP backend: Add support for 1D MaxPool and remove dependency on the non-edge dialect aten.max_pool2d.default.

MartinPavella:
### Summary
Add MaxPool1D support and remove dependency on non-edge dialect `aten.max_pool2d.default`.

### Test plan
Unit-tests provided.

cc @robert-kalmar @JakeStevens @digantdesai

PR #18117, opened 2026-03-12
NXP backend: Remove neutron_pass_manager.py from BUCK file

manuelcandales:
Differential Revision: D96291473

### DIFF
--- a/backends/nxp/BUCK
+++ b/backends/nxp/BUCK
@@ -75,7 +75,6 @@ fbcode_target(_kind = runtime.python_library,
         "nxp_backend.py",
         "neutron_partitioner.py",
         "neutron_node_extraction.py",
-        "neutron_pass_manager.py",
     ],
     deps = [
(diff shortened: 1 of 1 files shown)


robert-kalmar:
@manuelcandales , can you please specify the reason? 
CC @JakeStevens

JakeStevens:
> @manuelcandales , can you please specify the reason? CC @JakeStevens

looks like that file was removed in:

https://github.com/pytorch/executorch/pull/17821

OSS does not use BUCK so this only gets caught internal

robert-kalmar:
> > @manuelcandales , can you please specify the reason? CC @JakeStevens
> 
> looks like that file was removed in:
> 
> #17821
> 
> OSS does not use BUCK so this only gets caught internal

Ou, right, it is our recent change. Now it makes sense.

robert-kalmar (approved):

PR #18123, opened 2026-03-12
Arm backend: Add T5 model tests

YufengShi-dudu:
- Re-add TOSA and VGF tests for T5ForConditionalGeneration
- INT/quantize path currently shows poor numerical accuracy with atol=5

Change-Id: I0bc5b4482fcdc09146b6a394f944503f1617dd9e


YufengShi-dudu:
This PR only adds a new model test file. So it should be fine I think.

zingo (approved):
Added file is part of test that are not covered by buck2 e.g. should be OK to merge.

PR #18124, opened 2026-03-12
Arm backend: Add support for StaticCache in TOSA+INT

tom-arm:
* Create pass to decompose index_copy to index_put
* Enable INT tests for all targets except U55
* Adjust insert cast tests as the graph should delegate

Change-Id: I139728a78884ff1cbdc72568938f7b43f0b12279


tom-arm:
Failures are unrelated

zingo (approved):
Ok to merge, added file is under _passes that uses a glob() so buck2 should be OK.

PR #18128, opened 2026-03-12
Unified error reporting for Android ExecuTorch JNI layers

psiddh:
All four Android JNI modules (Generic, LLM, ASR, Training) previously used inconsistent exception types — a mix of RuntimeException, IllegalStateException, IllegalArgumentException, and the structured ExecutorchRuntimeException. This unifies them so every JNI error path throws ExecutorchRuntimeException with a proper error code, giving callers a single type to catch and programmatic access to the underlying runtime error.

Key changes per module:

LLM (jni_layer_llama.cpp) — generate() now captures the Error return from runner_->generate(), wraps the call in try/catch, reports failures via a new onError(errorCode, message) callback, and returns the actual error code instead of always returning 0.

ASR (jni_layer_asr.cpp) — replaced six env->ThrowNew(...) calls with setExecutorchPendingException (for pure JNI path)

Training (jni_layer_training.cpp) — added jni_helper.h include; replaced five throwNewJavaException("java/lang/Exception", ...) calls with throwExecutorchException, preserving the actual error codes from the failed operations.

Additionally: added default onError callbacks to LlmCallback (Java) and AsrCallback (Kotlin); 

cc @kirklandsign @cbilgin

### DIFF
--- a/extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/asr/AsrCallback.kt
+++ b/extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/asr/AsrCallback.kt
@@ -25,4 +25,12 @@ interface AsrCallback {
    * @param token The decoded text token
    */
   fun onToken(token: String)
+
+  /**
+   * Called when an error occurs during transcription.
+   *
+   * @param errorCode Error code from the ExecuTorch runtime
+   * @param message Human-readable error description
+   */
+  fun onError(errorCode: Int, message: String) {}
 }
--- a/extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/llm/LlmCallback.java
+++ b/extension/android/executorch_android/src/main/java/org/pytorch/executorch/extension/llm/LlmCallback.java
@@ -37,4 +37,14 @@ public interface LlmCallback {
    */
   @DoNotStrip
   default void onStats(String stats) {}
+
+  /**
+   * Called when an error occurs during generate().
+   *
+   * @param errorCode Error code from the ExecuTorch runtime (see {@link
+   *     org.pytorch.executorch.ExecutorchRuntimeException})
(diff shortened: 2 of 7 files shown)


extension/android/executorch_android/src/main/java/org/pytorch/executorch/ExecuTorchRuntime.java
```diff
@@ -45,6 +47,38 @@ public static void validateFilePath(String path, String description) {
...
+   * files).
+   *
+   * @param context Android context for accessing system services (must not be null)
+   * @param modelPath Path to the model file (must not be null or empty)
+   * @return true if reported available memory exceeds the file size, false otherwise
+   * @throws IllegalArgumentException if context is null or modelPath is null/empty
+   * @throws RuntimeException if the file does not exist or is not readable
+   * @throws IllegalStateException if the ActivityManager system service is unavailable
+   */
+  public static boolean checkMemoryFit(Context context, String modelPath) {
```
GregoryComer: I do worry a little bit about providing an API to check fit without being able to accurately guess peak memory - KV cache / load time peaks can be large. But I'll defer to your judgement. It might be interesting to test a few key models and check load / inference time peaks. I've used ~2x model size as a heuristic for small models, but transformers can get complicated.

GregoryComer (approved):
Thanks for cleaning this up. I don't have a ton of context on JNI error handling, but it looks reasonable to me.

extension/android/executorch_android/src/main/java/org/pytorch/executorch/ExecuTorchRuntime.java
psiddh: Yes fair point, I was thinking abt it too, I think dropping this change from the PR is a good idea, as this PR solely focusses on unifying error reporting

PR #18131, opened 2026-03-12
Update XNNPACK dependency to 1adaa7c70

GregoryComer:
Update the XNNPACK dependency, as well as cpuinfo and pthreadpool, which are needed for the newer XNNPACK.


digantdesai (approved):
Thanks @GregoryComer for the update.

PR #18132, opened 2026-03-12
Arm backend: Decompose where.ScalarOther for quantization support

YufengShi-dudu:
Arm quantizer does not annotate aten.where.ScalarOther, so where remains unquantized after it is rewritten to aten.where.self pre-QUANTIZE.

Decompose where.ScalarOther(cond, x, s) into where.self(cond, x, full_like(x, s)) before QUANTIZE so existing where.self annotation can handle it.

Change-Id: I82a7ffec5e3e037219ab7185d71371c7c14a5b4f


zingo (approved):
Good to merge, added file in _passes is handled by a glob in buck2

PR #18133, opened 2026-03-12
Fix unsafe deserialization in ETRecord and export_serialize

lucylq:
Add weights_only=True to torch.load() in deserialize_torch_artifact()
to prevent arbitrary code execution via malicious serialized artifacts.

Replace usages of pickle.dumps and pickle.loads with torch.save and torch.load

As generating and reading etrecord should be done as debugging steps and usually on the same version of et, BC is not a big concern.

cc @JacobSzwejbka @angelayi

### DIFF
--- a/exir/serde/export_serialize.py
+++ b/exir/serde/export_serialize.py
@@ -333,7 +333,7 @@ def deserialize_torch_artifact(
         return {}
     buffer = io.BytesIO(serialized)
     buffer.seek(0)
-    artifact = torch.load(buffer)
+    artifact = torch.load(buffer, weights_only=True)
     assert isinstance(artifact, (tuple, dict))
     return artifact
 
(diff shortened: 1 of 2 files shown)


Gasoonjia:
can you trigger import to pay attention to the internal ci?

exir/serde/export_serialize.py
```diff
@@ -333,7 +333,7 @@ def deserialize_torch_artifact(
         return {}
     buffer = io.BytesIO(serialized)
     buffer.seek(0)
-    artifact = torch.load(buffer)
+    artifact = torch.load(buffer, weights_only=True)
```
angelayi, line 336: I guess we forked the serialization/deserialization code here from export's... but in export's case we cant change the load to weights=True because then we can't load torchbind objects. In ET's case you already don't support torchbind objs so it doesnt matter here, but just to note this difference!

angelayi (approved):

exir/serde/export_serialize.py
lucylq: thanks!

PR #18134, opened 2026-03-12
Remove deprecated XNNPACK capture utilities and migrate tests (#18134)

JacobSzwejbka:
Summary:

Delete capture_graph_for_xnnpack() and get_xnnpack_capture_config() which
were only used in test files and relied on the deprecated exir.capture API.
Migrate test_xnnpack_utils.py to use inline to_edge(export(...)) calls.
Remove the deprecated exports from xnnpack/__init__.py.

Differential Revision: D95605468


digantdesai:
Looks OK to me. @GregoryComer  can you take a look as well?

GregoryComer (approved):

PR #18136, opened 2026-03-12
Arm backend: Fix the issue on conv->relu->permute->reshape(5D)

bdemirb:


zingo (approved):
OK to merge after Arm-Test are OK I see this fail:
[pull / unittest-arm-backend-with-no-deps (test_pytest_models_tosa) / linux-job (pull_request)](https://github.com/pytorch/executorch/actions/runs/23028987977/job/66883164940?pr=18136)

Also also see the cortex-m fail BUT that in unrelated and OK to ignore, a fix for that tests is on it's way, see https://github.com/pytorch/executorch/pull/18194

